PR #19, opened 2026-07-24
Fix/semantic search deadline

buger:
## Summary

This contribution is stacked on the cancellation-safe ONNX initialization change in #18 Merge that first.

Semantic and exact-vector searches now use one absolute deadline from request entry through single-flight waiting, global and per-index admission, query embedding, hybrid and multi-kNN composition, and cooperative

The ES handler now owns its spawned search task. Dropping the request aborts the child, while a child panic still becomes the established JSON HTTP 500 response.

The existing public Rust signatures for `Index::run_knn_brute_force`, `Index::run_semantic`, and `Index::run_hybrid` remain source-compatible. Private deadline-aware variants carry the shared request deadline on t

## Root cause

Semantic/vector queries ran under a non-preemptible `block_in_place` wrapper, several layers created fresh timeout budgets, identical-query followers could wait without the request deadline, and the HTTP handler d

Under retained neural-search load this produced individual requests beyond 30 seconds and post-cancellation drain beyond 120 seconds. Client disconnects did not reliably stop queued or scanning work.

## Behavior after this change

- One request-entry deadline bounds single-flight, both semaphore waits, embedding, recursion, and exact scanning.
- Exact scans yield and check between segments, document batches, SQ8 batches, and passage groups.
- Partial exact results set `timed_out: true` and `hits.total.relation: "gte"`.
- Multi-kNN and hybrid propagate child timeout state and stop scheduling later children.
- A cancelled single-flight leader removes its in-flight entry; subscribed followers recheck cache and recompute.
- Timed-out partials may wake current followers but are never inserted into the result cache.
- Dropping the HTTP route future aborts and drains its owned child task.
- Search-task panics return a JSON HTTP 500 rather than unwinding through Axum.

## Measured validation

Retained real four-PDF neural index: 2,190 records, query cache disabled for unique-query measurements.

- `timeout: 250ms`: HTTP 200 in 251 ms, `timed_out: true`, empty partial hits, `total.relation: "gte"`.
- Warm unique queries, 20 requests per concurrency:
  - c1: p50 260 ms, p95 361 ms
  - c2: p50 293 ms, p95 358 ms
  - c4: p50 315 ms, p95 428 ms
  - c8: p50 355 ms, p95 662 ms
- Eight unique clients aborted at 100 ms: phase logs stopped before segment-load, scored, and complete events.
- Cache-enabled identical c8: one leader executed and seven followers were served through single-flight.

## Known soft-deadline regions

Already-running ONNX blocking inference cannot be cancelled. One cold stored-segment decode/parse, full memtable source collection, SQ8 parameter fitting/code refresh, final candidate sorting, and response serialization are also not internally interruptible. Deadline checks surround the cold segment load, but a single large phase can overrun the soft wall.

## Tests

- Real future 1/10/250 ms deadlines with a nonzero partial scan.
- Mutation-sensitive one-child multi-kNN timeout propagation.
- Separate multi-kNN later-child scheduling suppression.
- Hybrid child timeout propagation.
- Deterministic subscribed-follower leader cancellation and recompute.
- Timed-out result exclusion from query cache.
- Concurrent unique semantic requests reaching multiple scan checkpoints.
- Public executor signature compatibility.
- Real Axum route panic to JSON HTTP 500.
- Real Axum route-future drop aborting and draining its child.

Validation completed:

- Scoped engine deadline tests: 7 passed.
- Scoped kNN and public-API integration tests: passed.
- API lifetime and route tests: 4 passed.
- `xerj-server` check and clippy with `-D warnings`: passed.
- Formatting and diff checks: passed.
- Rebuilt-release ES-YAML hard gate: 1,360 passed, 0 failed, 3 skipped.

## Manual reproduction

Run XERJ with an active neural or ONNX embedding backend and a flushed semantic index, then issue:

```bash
curl -sS -H 'content-type: application/json' \
  http://127.0.0.1:9200/my-index/_search \
  -d '{"query":{"semantic":{"field":"body","query":"financial risks and revenue trends","k":10}},"size":10,"timeout":"250ms"}'
```

The response should return near the requested budget with `timed_out: true` and `hits.total.relation: "gte"` when the work cannot finish in time. Aborting several unique requests with `curl --max-time 0.10` should stop cooperative scan phases instead of leaving them running after the clients exit.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -5386,6 +5386,168 @@ pub(crate) fn top_level_query_type(body_query: &Option<serde_json::Value>) -> &'
     }
 }
 
+/// Request-owned child task: dropping the HTTP handler aborts its search
+/// instead of leaving detached work running after a client disconnect.
+struct AbortOnDrop<T>(tokio::task::JoinHandle<T>);
+
+impl<T> Drop for AbortOnDrop<T> {
+    fn drop(&mut self) {
+        self.0.abort();
+    }
+}
+
+#[cfg(test)]
+type PendingSearchSignals = (
+    tokio::sync::oneshot::Sender<()>,
+    tokio::sync::oneshot::Sender<()>,
+);
+#[cfg(test)]
+static TEST_PENDING_SEARCH_SIGNALS: std::sync::OnceLock<
+    std::sync::Mutex<Option<PendingSearchSignals>>,
+> = std::sync::OnceLock::new();
+
+#[cfg(test)]
+struct SignalOnDrop(Option<tokio::sync::oneshot::Sender<()>>);
+#[cfg(test)]
+impl Drop for SignalOnDrop {
+    fn drop(&mut self) {
+        if let Some(sender) = self.0.take() {
+            let _ = sender.send(());
+        }
+    }
+}
+
+fn search_task_panic_response(join_err: tokio::task::JoinError, index: &str) -> Response {
+    tracing::error!(error = %join_err, index, "search task panicked");
+    ApiError::new(xerj_common::XerjError::internal(
+        "search panicked; check server logs for details",
+    ))
+    .into_response()
+}
+
+#[cfg(test)]
+mod search_task_lifetime_tests {
+    use super::{search_task_panic_response, AbortOnDrop, TEST_PENDING_SEARCH_SIGNALS};
+    use axum::http::{Request, StatusCode};
+    use tower::ServiceExt;
+
+    fn test_state() -> crate::state::AppState {
+        let dir = tempfile::tempdir().expect("tempdir").keep();
+        let mut config = xerj_common::config::Config::default();
+        config.server.data_dir = dir.to_string_lossy().into_owned();
+        config.storage.wal_sync = xerj_common::config::WalSync::Async;
+        let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+        let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+        crate::state::AppState::new(config, engine, metrics)
+    }
+
+    #[tokio::test]
+    async fn executor_panic_becomes_http_500() {
+        let join_err = tokio::spawn(async { panic!("injected search panic") })
+            .await
+            .expect_err("task must panic");
+        let response = search_task_panic_response(join_err, "panic-test");
+        assert_eq!(response.status(), StatusCode::INTERNAL_SERVER_ERROR);
+    }
+
+    #[tokio::test]
+    async fn es_search_route_maps_child_panic_to_json_http_500() {
+        let state = test_state();
+        state
+            .engine
+            .create_index(
+                "test-panic-search-task",
+                xerj_common::types::Schema::empty(),
+            )
+            .unwrap();
+        let app = crate::router::build_es_compat_router(state);
+        let response = app
+            .oneshot(
+                Request::post("/test-panic-search-task/_search")
+                    .header("content-type", "application/json")
+                    .body(axum::body::Body::from(r#"{"query":{"match_all":{}}}"#))
+                    .unwrap(),
+            )
+            .await
+            .expect("route response");
+        assert_eq!(response.status(), StatusCode::INTERNAL_SERVER_ERROR);
+        let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let body: serde_json::Value = serde_json::from_slice(&bytes).unwrap();
+        assert_eq!(body["status"], 500);
+        assert!(
+            body["error"]["reason"]
+                .as_str()
+                .is_some_and(|reason| reason.contains("search panicked")),
+            "unexpected error body: {body}"
+        );
+    }
+
+    #[tokio::test]
+    async fn dropping_es_search_route_future_aborts_owned_child() {
+        let state = test_state();
+        state
+            .engine
+            .create_index(
+                "test-pending-search-task",
+                xerj_common::types::Schema::empty(),
+            )
+            .unwrap();
+        let app = crate::router::build_es_compat_router(state);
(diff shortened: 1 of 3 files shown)


xerj-org (changes requested):
The deadline threading, cooperative checkpoints, and partial-results semantics are solid, and the CancellationSafeInit wait loop is fine as-is (tokio ≥1.23 guarantees `notify_waiters()` wakes a created-but-unpolled `Notified`; verified against tokio 1.52 source — no missed-wakeup hang).

One change needed before merge:

**Cold-segment loads now run unwrapped on async workers.** kNN/multi-kNN/semantic queries bypass the M5.21 `block_in_place` wrapper (index.rs ~7052-7068), but `stored_values_for` (index.rs ~14007) does synchronous segment open + decompress + whole-segment `serde_json` parse (~100MB/segment) on cache miss. The stored-value cache misses recur after every merge publish, so several concurrent cold vector queries under write load can pin all runtime workers for seconds — the exact starvation M5.21 fixed, reintroduced on the vector path. Please wrap the cold-load in `spawn_blocking` (or `block_in_place` scoped to the load), keeping the scan loop cooperative.

Minor, non-blocking:
- Nested kNN and hybrid aren't in `cooperative_vector_query`, so `timeout` still can't preempt them — worth documenting the asymmetry or covering in a follow-up.
- `timed_out_result()` early-returns skip `metric_query_count`, so admission timeouts are invisible in query metrics.

buger:
Addressed in the updated branch.

The blocking cold-load issue was valid. Exact kNN, semantic, nested-kNN, and HNSW-repair paths now use a cancellation-safe asynchronous stored-value loader:

- A cache miss runs segment open, decompression, and whole-section JSON parsing in `spawn_blocking`, never on a Tokio worker.
- Concurrent cold requests for the same immutable segment share one detached producer through a watch channel, so eight cold requests do not parse the same 100 MB segment eight times.
- Distinct-segment cold parsing is bounded to four simultaneous producers so a cold multi-segment storm cannot multiply whole-segment parse memory without limit.
- Cancelling or timing out every requester does not abort the non-cancellable blocking task. It finishes and warms the cache for a later request.
- Cache insertion and merge eviction share a short lifecycle lock. A load that publishes before retirement is removed by merge eviction; one that reaches publication after retirement sees the authoritative snapshot and does not insert. The same transaction protects the synchronous fast-aggregation loader.
- If a merge retires the segment while the detached load is running, the load can still satisfy callers holding the old snapshot but cannot retain the retired segment in the cache. Its in-flight key is removed when the producer completes.
- Cache hits remain the original cheap DashMap lookup plus `Arc` clone.

The regression test uses a one-worker Tokio runtime and an injected 250 ms cold load. Eight concurrent vector queries leave a 10 ms async heartbeat responsive and execute exactly one parse. A separate 20 ms request returns `timed_out: true` before that load completes, then verifies that the detached producer warms the cache. Another deterministic test pauses a detached producer immediately before publication, retires and evicts the segment, releases the producer, and proves the old-snapshot caller receives its result without a cache resurrection. A failure-path test proves all followers close and the in-flight key is removed.

I also addressed the two observability/deadline points:

- Nested kNN now receives the request's original absolute deadline and checks it during segment collection, parent scoring, and large nested-array scans. Partial responses set `timed_out: true` and `hits.total.relation: "gte"`.
- Timeouts while waiting for single-flight, global admission, or per-index admission now increment both query count and total query latency.

I did not remove the M5.21 blocking boundary from all hybrid queries. A hybrid may contain a normal lexical child, and the generic lexical scan is not yet cooperative. Running that shape directly on Tokio workers would trade deadline overrun for the worker-starvation regression M5.21 fixed. Vector children inside a hybrid already receive the same absolute deadline and stop at their cooperative checkpoints. Fully preemptible lexical-hybrid execution requires a cooperative lexical executor and should be a focused follow-up rather than a classifier-only change.

The branch was rebased cleanly onto current `main`. Validation on the rebased head:

- `cargo fmt --all --check`
- `cargo clippy -p xerj-engine --all-targets -- -D warnings`
- deadline regression module: 13 passed
- engine library excluding the three pre-existing process-aborting Painless stress tests: 140 passed, 0 failed, 3 filtered
- engine integration suite: 94 passed, 0 failed
- remaining named engine integration targets: 91 passed, 0 failed, 17 intentionally ignored battle/performance cases
- API suite: 55 passed, 0 failed, plus 2 doc tests passed and 2 ignored
- scoped release build: `xerj-server` and `es-yaml-runner`
- isolated rebuilt-release ES-YAML hard gate: 1360 passed, 0 failed, 3 skipped, 1363 total

The three filtered Painless tests are unchanged from `main` and process-abort this environment through their existing default-stack overflow; the PR does not touch the Painless parser or evaluator. The same named integration-target command initially reproduced that known abort because Cargo also invoked the library harness, after which every integration-test binary was run explicitly and passed.

xerj-org:
Merged — the rework is exactly what the review asked for and then some: the single-flighted, semaphore-bounded `spawn_blocking` cold loader with the lifecycle guard against merge-eviction races is verified sound, nested kNN got full deadline coverage, hybrid's retained `block_in_place` boundary is a defensible documented trade-off, and the regression tests (1-worker starvation, deadline drop, retired-segment reinsert, failed-load followers) pin all of it.

Non-blocking follow-ups worth filing:
1. `sender.send(Some(value))` in the detached producer (index.rs ~14570) can fail if every receiver dropped, and the error is discarded — a follower subscribing in the window before `loads.remove` then observes channel-close and silently skips a live segment. `send_replace` cannot fail and closes the hole.
2. The producer never re-checks `stored_value_cache` after winning the Vacant single-flight entry — a cache hit landing in that window does one redundant parse.
3. `stored_value_load_semaphore` is per-Index; N indexes × 4 permits can still stack cold parses under fan-out load — consider a process-wide bound.
4. `run_nested_knn_brute_force_with_deadline` clones the whole memtable before its first deadline check.
5. The `run_knn_hnsw` → `get_document_uncounted` cold path can still do a small synchronous section decode on a worker (bounded, ~30ms class) — same class as the fixed defect, much smaller magnitude.

PR #21, opened 2026-07-24
fix(autoindex): deduplicate content with crash-safe replacement

buger:
## Summary

This change makes `xerj autoindex` deduplicate byte-identical source files and replaces their indexed generation with a crash-safe, resumable transaction.

The original failure was not cosmetic accounting drift. Autoindex could process the same content through multiple paths, derive the same deterministic document IDs, and silently overwrite the live documents while journaling both files as complete. In the AMCOR PDF reproduction:

| Measurement | Before | After |
|---|---:|---:|
| Source paths | 2 | 2 |
| Canonical files extracted | 2 | 1 |
| `file_done` rows | 2 | 1 |
| Journal record sum | 2,442 | 1,221 |
| Refreshed live records | 1,221 | 1,221 |
| Duplicate aliases | 0 | 1 |

The previous result claimed 2,442 indexed records while only 1,221 existed. The new result indexes one canonical copy, preserves the second path as an alias, and keeps journal and live cardinality equal.

## User-visible behavior

Usage does not change:

```bash
xerj autoindex <folder>
xerj autoindex map
```

When identical files are present, autoindex now reports that their content will be indexed once:

```text
autoindex: 1 byte-identical duplicate path(s) will reuse canonical content
  duplicate: report-copy.pdf → report-original.pdf
```

Canonical records include every current name in `_source.ax_paths`. `autoindex map` exposes aliases in a separate `duplicate_files` section rather than classifying them as junk.

## Durable content identity

Every discovered file receives a streaming full-content XXH3-128 digest in `content.rs`. Equal digests are candidates, not proof: peers are byte-compared before being collapsed.

Additional identity rules:

- canonical selection is deterministic;
- real in-tree paths are preferred over followed symlinks;
- hardlinks are recognized without redundant byte comparison;
- non-UTF-8 paths retain a reversible path identity;
- byte-proven unequal files under a forced digest collision receive deterministic discriminators;
- full hashing and grouping are linear in corpus bytes rather than pairwise comparisons.

Legacy frozen plans keep their historical document key while persisting the new full digest. This avoids duplicating existing live documents while making same-size content mutations detectable on subsequent runs.

## Crash-safe replacement protocol

Replacing a completed file is now an explicit per-file generation transaction:

1. Extract the complete source into an on-disk NDJSON stage.
2. Hash the source again after extraction.
3. If the source changed, discard the stage without changing visibility.
4. Fsync `file_replace_start(file_key, generation)` before persisting the updated plan or deleting live records.
5. Delete all documents for the previous `ax_file` generation.
6. Publish the verified stage in ordinary bulk requests.
7. Fsync a matching-generation `file_done` record as the commit.

Replay removes the previous completion when it sees `file_replace_start`. A crash after plan persistence, delete, any intermediate bulk, or the final bulk before `file_done` therefore schedules repair. Resume deletes the zero or partial generation and republishes the full verified stage.

If generation B is pending and the source changes to C before repair, autoindex fsyncs a new `file_replace_start(C)` before C becomes visible. Only `file_done(C)` may clear the latest pending generation.

## Journal integrity

`state.rs` now provides:

- an exclusive state-directory lock;
- byte-accurate replay boundaries;
- truncation and fsync of a final torn record before any append;
- fail-closed handling for malformed newline-terminated middle records, including the corruption byte offset and remediation;
- generation-aware completion matching;
- a shared append transaction that records the original length, writes, fsyncs, and truncates plus fsyncs rollback after a partial write or sync error;
- propagation of commit failures to the top-level run result.

A `file_done` append or fsync error cannot be counted as success. Live records may already exist, but the pending transaction remains repairable on the next run.

## Staging lifecycle

Stages are direct children of the selected autoindex state directory and use the reserved `.autoindex-stage-` prefix.

Startup acquires the exclusive state lock before cleanup, then removes only direct-child, non-symlink regular files with that prefix. It never recurses, follows links, or applies a TTL. A second process using the same state directory is rejected before it can remove another process's active stage.

## Catalog lifecycle

Alias catalog publication removes prior duplicate entries by logical path before writing the current deterministic alias document. This cleans IDs produced by older identity schemes, including IDs that cannot be reconstructed from the current plan.

The AMCOR lifecycle was verified through:

- fresh deduplicated indexing;
- resume with no re-extraction;
- canonical removal and deterministic alias promotion;
- restoring the original canonical path;
- raw catalog lookup returning exactly one alias;
- map output rendering the duplicate section exactly once.

## Failure-injection coverage

The HTTP-level regression harness uses the production `run_index` path, real `Es` client, extractor, staging reader, bulk splitter, delete-by-query calls, and append-only journal.

It covers:

- source mutation after more than 5,000 staged records;
- failure immediately after the replacement plan;
- failure immediately after delete;
- partial application of the first bulk;
- one successful 5,000-document bulk followed by a partial final bulk;
- two successful bulks followed by a partial final bulk;
- failure after the final successful bulk but before `file_done`;
- `file_done` append failure;
- `file_done` fsync failure;
- a true 23-byte partial journal write followed by another worker's successful commit;
- pending generation B superseded by source generation C;
- torn `file_replace_start`, plan, and `file_done` tails;
- malformed middle-record rejection;
- stale-generation completion rejection;
- stage-orphan cleanup and concurrent state-directory rejection.

Every recovery case verifies exact live locators and values, with no old, zero, stale, or partial generation.

## Durability cost

The previous journal synced every 32 completions. This change syncs every `file_done`, because batching can lose completion records for up to 31 already-visible files and is incompatible with `file_done` being the replacement commit.

On the measured ext4 host, a journal-only microbenchmark wrote 8,192 realistic completion records:

| Policy | Median | Range | Throughput | Sync calls |
|---|---:|---:|---:|---:|
| Sync every file | 7.881 s | 7.316–8.493 s | ~1,039 files/s | 8,192 |
| Sync every 32 files | 0.398 s | 0.315–0.494 s | ~20,573 files/s | 256 |

This is a 19.79x journal-only ratio and approximately 0.913 ms of incremental latency per completed file. It is not a claim that end-to-end autoindex becomes 19.79x slower: extraction, PDF parsing, embeddings, and HTTP indexing dominate larger files. The cost is most visible for corpora containing very many tiny files.

### Benefits

- Journal counts match live document cardinality.
- Duplicate PDFs and copied folders are extracted and embedded once.
- Every current path remains discoverable.
- Canonical promotion and restoration are deterministic.
- Crashes and partial backend failures converge to an exact generation.
- Torn journal tails are repaired without accepting middle corruption.
- Legacy plans migrate without changing existing document IDs.

### Costs

- Every source is read once for full hashing before extraction.
- Replacement stages require temporary disk space up to the extracted NDJSON size of concurrently processed files.
- Per-file commit syncing adds approximately 0.913 ms/file on the measured filesystem.
- Adding `fs2` supplies the portable exclusive file lock.
- A state directory is intentionally single-writer.

## Key implementation files

- `engine/crates/xerj-autoindex/src/content.rs`: full-content identity, exact duplicate verification, canonical selection, mutation verification.
- `engine/crates/xerj-autoindex/src/state.rs`: state lock, journal replay and repair, generations, transactional append and rollback.
- `engine/crates/xerj-autoindex/src/lib.rs`: legacy migration, staged extraction, replacement ordering, error propagation.
- `engine/crates/xerj-autoindex/src/esclient.rs`: checked delete-by-query operation.
- `engine/crates/xerj-autoindex/src/catalog.rs`: alias schema, lifecycle, and rendering.
- `engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs`: end-to-end failure and resume coverage.

## Validation

```text
cargo test -p xerj-autoindex
53 passed; 0 failed

cargo clippy -p xerj-autoindex --all-targets -- -D warnings
clean

cargo fmt --all -- --check
clean

git diff --check
clean

ES-YAML conformance
1,360 passed; 0 failed; 3 skipped; 1,363 total
```

The conformance suite ran against the release server and runner built from this branch

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -1516,6 +1516,16 @@ dependencies = [
 ]
 
 [[package]]
+name = "fs2"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "9564fc758e15025b46aa6643b1b77d047d1a56a1aea6e01002ac0c7026876213"
+dependencies = [
+ "libc",
+ "winapi",
+]
+
+[[package]]
 name = "fst"
 version = "0.4.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -5863,6 +5873,7 @@ dependencies = [
  "csv",
  "encoding_rs",
  "flate2",
+ "fs2",
  "libc",
  "memchr",
  "pdf_oxide",
--- a/engine/crates/xerj-autoindex/Cargo.toml
+++ b/engine/crates/xerj-autoindex/Cargo.toml
@@ -18,6 +18,8 @@ xxhash-rust.workspace = true
 memchr.workspace = true
 rayon.workspace = true
 libc = "0.2"
+tempfile.workspace = true
+fs2 = "0.4"
 # ES-compat HTTP client. Blocking mode: all of autoindex is synchronous
 # (std threads), invoked from the server binary via spawn_blocking.
 reqwest = { workspace = true, features = ["blocking"] }
@@ -33,4 +35,3 @@ serde_yaml = "0.9"                               # YAML multi-doc
 rusqlite = { version = "0.32", features = ["bundled"] }
 
 [dev-dependencies]
-tempfile.workspace = true
--- a/engine/crates/xerj-autoindex/src/catalog.rs
+++ b/engine/crates/xerj-autoindex/src/catalog.rs
@@ -31,6 +31,7 @@ pub fn catalog_mapping() -> Value {
             "format": {"type": "keyword"},
             "status": {"type": "keyword"},
             "reason": {"type": "text"},
+            "duplicate_of": {"type": "keyword"},
             "records": {"type": "long"},
             "junk": {"type": "long"},
             "run_id": {"type": "keyword"},
@@ -59,6 +60,7 @@ pub const GOTCHAS: &[&str] = &[
     "all dates are normalized to RFC3339 UTC millis; mappings use strict_date_optional_time||epoch_millis",
     "query all data with the wildcard index pattern (e.g. ax-*) or comma lists — never multi-index aliases (they resolve to the first index only)",
     "documents were indexed with refresh at the end of the run; new writes need _refresh before they are searchable",
+    "byte-identical paths are indexed once; canonical records retain every current filename in _source.ax_paths, and exact alias resolution is available through the catalog's duplicate_files entries",
 ];
 
 pub struct DatasetDocInput<'a> {
@@ -127,6 +129,41 @@ pub fn file_doc(
             "bytes": bytes,
             "run_id": run_id,
         }),
+    )
+}
+
+pub fn duplicate_file_doc(
+    file_key: &str,
+    path: &str,
+    path_id: &str,
+    duplicate_of: &str,
+    bytes: u64,
+    run_id: &str,
+) -> (String, Value) {
+    let alias_id = duplicate_file_id(file_key, path, path_id);
+    (
+        alias_id,
+        json!({
+            "doc_kind": "file",
+            "file_key": file_key,
+            "path": path,
+            "format": "duplicate",
+            "status": "duplicate",
+            "reason": format!("byte-identical content already indexed from {duplicate_of}"),
+            "duplicate_of": duplicate_of,
+            "records": 0,
+            "junk": 0,
+            "bytes": bytes,
+            "run_id": run_id,
+        }),
+    )
+}
+
+pub fn duplicate_file_id(file_key: &str, path: &str, path_id: &str) -> String {
+    let identity = if path_id.is_empty() { path } else { path_id };
+    format!(
+        "file-alias:{}",
+        crate::ids::doc_id("duplicate-file", file_key, identity)
     )
 }
 
@@ -248,6 +285,7 @@ pub fn render_map(
     datasets: &[Value],
     correlations: &[Value],
     junk_files: &[Value],
+    duplicate_files: &[Value],
     junk_total: u64,
 ) -> String {
     let mut s = String::new();
@@ -255,10 +293,12 @@ pub fn render_map(
     if let Some(r) = run {
         let g = |k: &str| r.get(k).map(pretty_val).unwrap_or_default();
         s.push_str(&format!(
-            "run `{}` — root `{}` — {} files, {} records indexed, {} junk records, wall {}s\n\n",
+            "run `{}` — root `{}` — {} paths ({} unique content, {} duplicate aliases), {} records indexed, {} junk records, wall {}s\n\n",
             g("run_id"),
             g("root"),
             g("files_total"),
+            g("unique_content_files"),
+            g("duplicate_files"),
             g("records_total"),
             g("junk_records_total"),
             g("wall_seconds"),
@@ -400,6 +440,21 @@ pub fn render_map(
         }
     }
 
+    if !duplicate_files.is_empty() {
+        s.push_str(&format!(
+            "## Duplicate aliases ({} paths; content indexed once)\n\n",
+            duplicate_files.len()
+        ));
+        for file in duplicate_files.iter().take(30) {
+            let g = |key: &str| file.get(key).map(pretty_val).unwrap_or_default();
+            s.push_str(&format!("- `{}` → `{}`\n", g("path"), g("duplicate_of")));
+        }
+        if duplicate_files.len() > 30 {
+            s.push_str(&format!("- … and {} more\n", duplicate_files.len() - 30));
+        }
+        s.push('\n');
+    }
+
     if !correlations.is_empty() {
         s.push_str("## Cross-dataset correlations\n\n");
         for c in correlations {
@@ -488,5 +543,24 @@ fn trunc_f(s: &str) -> String {
     match s.parse::<f64>() {
         Ok(f) => format!("{f:.2}"),
         Err(_) => s.to_string(),
+    }
+}
+
+#[cfg(test)]
+mod duplicate_tests {
+    use super::*;
+
+    #[test]
+    fn duplicate_aliases_render_separately_from_junk() {
+        let duplicate = json!({
+            "path": "copy.pdf",
+            "duplicate_of": "report.pdf",
+            "status": "duplicate"
+        });
+        let rendered = render_map(None, &[], &[], &[], &[duplicate], 0);
+        assert!(rendered.contains("## Duplicate aliases"));
+        assert_eq!(rendered.matches("## Duplicate aliases").count(), 1);
+        assert!(rendered.contains("`copy.pdf` → `report.pdf`"));
+        assert!(!rendered.contains("## Junk / skipped"));
     }
 }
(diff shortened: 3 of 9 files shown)


xerj-org (changes requested):
The replacement-transaction protocol itself is sound (intent-before-visibility, fsynced commit records, honest non-atomicity) and the failure-injection tests are genuinely strong. Three things block merge:

1. **Legacy-key collision can double-assign one key to two files** (lib.rs ~311-335): two distinct files sharing a legacy `file_key` (xxh3_64 of first 64KB + size — realistic for fixed-size machine-generated files with identical headers) both match `key.as_str() == legacy_key` and both get the same planned key. Two workers then run the replacement transaction on the same `ax_file` key concurrently — each `delete_by_query` wipes the other's docs and the final live set is nondeterministic. This is the same lost-write class the PR exists to fix. The `plan.files.iter().find(...)` over a HashMap also makes assignment nondeterministic across runs.

2. **Any alias-set change forces a full-corpus re-index** (lib.rs ~355-362): the alias diff and the one-time `alias_paths_indexed` migration insert *every* key into `content_changed`, so adding/removing one duplicate file re-extracts and re-embeds the entire corpus. Should be per-key (only keys whose `paths_by_key` entry changed).

3. **Resume/no-op-rerun perf cliffs**: resumed runs are O(files × plan.files) with a fresh 64KB read per file (lib.rs ~308-318), and `content::resolve` full-hashes every byte of the corpus on every run with no size+mtime short-circuit — a no-op rerun of a 1TB folder reads 1TB.

Also worth fixing while in there: `delete_by_query(refresh=true)` is issued unconditionally per processed file, including fresh files that can't have prior docs — a real first-run ingest regression; and `write_plan` appends the full plan JSON every run (unbounded journal growth).

buger:
Thanks for the detailed review. I agreed with the safety and amplification findings and rebuilt the branch on current `main` rather than patching around them.

1. **Legacy-key collision and nondeterministic ownership — fixed.** Resume reconciliation now builds indexed path/content lookups, reserves every planned key to one owner, and never chooses an owner through `HashMap` iteration. If two byte-distinct files collide under the historical prefix-derived key, resume fails before bulk or delete-by-query activity and tells the operator to rerun with `--fresh`. Unit and HTTP tests cover the historical owner sorting second and verify zero backend mutation on failure.

2. **Alias changes reindexing the full corpus — fixed.** Ordinary alias additions and removals invalidate only the affected content key. The one-time migration from the historical global `alias_paths_indexed=false` state still rewrites every canonical key because that old boolean cannot prove which individual documents already contain `ax_paths`. Tests cover per-key add/remove behavior and the unavoidable all-key migration.

3. **Resume complexity and corpus hashing — lookup fixed; full hashing intentionally retained.** Normal resume matching is now O(files + plan), and exact path/content matches avoid the extra legacy 64 KiB prefix read. I did not add a size/mtime metadata shortcut for full-content hashing. Size, mtime, inode, and ctime are not proof of byte identity across every supported local and network filesystem; a same-size rewrite with restored or stale metadata could reuse an old digest, skip the file as completed, and retain stale indexed data indefinitely. That would violate the crash-safe exact-generation contract. The 1 TB no-op read remains a real and explicitly acknowledged performance cost; a future optimization needs a stronger filesystem/content-version contract or an explicit best-effort mode.

4. **Unconditional delete-by-query — fixed without weakening recovery.** Every publication receives durable intent before visibility. A genuinely fresh plan skips its unnecessary first delete-by-query, while replacements, pending repairs, and every resumed legacy todo conservatively clean possible prior visibility. The legacy regression seeds a plan with no completion or intent plus an obsolete visible locator, then proves one delete and an exact final locator set.

5. **Full-plan journal growth — fixed for the reported amplification.** Identical effective plans are no longer appended on no-op resumes. The journal remains append-only for compact transaction and resume history, so bounded checkpoint compaction remains a separate lifecycle improvement; this change specifically removes the full-plan-JSON-per-run growth.

The branch was rebased from upstream `1a6307e` to current `main` at `5a0881d`. Conflict resolution preserves the merged PDF worker count/timeout setup, bulk request timeout/retry behavior, and timeout provenance in the stronger transactional journal. The final branch is one bisectable commit: `c529e6047e98b7c467269c49af0e92161b5b9eda`.

Post-rebase validation: 73 passed and 0 failed for `xerj-autoindex`; strict all-target Clippy passed; `cargo fmt --all -- --check` passed; `git diff --check` passed. An independent synthetic-diff review also approved the composed PDF, bulk-timeout, and dedupe contracts.

xerj-org (changes requested):
The rebuild fixed four of the five original findings well (collision-safe legacy migration, per-key alias diffing, conditional delete_by_query, journal growth policy) — but the re-review surfaced a fresh blocker in the rebuilt resume logic plus two majors, all adversarially confirmed:

1. **Blocker — resume key-claiming lets two distinct current files silently share one plan key** (lib.rs ~298): when a planned key is claimed by one file via rel/path_id, another current file resolving to the same key isn't diverted — two workers again run the replacement transaction on one `ax_file` key concurrently, each `delete_by_query` wiping the other's docs. Same lost-write class as the original finding #1, reintroduced via a different path.

2. **Major — `delete_by_query` is not complete removal** (esclient.rs ~216): the server implements `_delete_by_query` as a single search capped at `size:10000` with per-hit deletes. A key whose previous generation has >10k docs (large PDFs at small chunk size) leaves stale docs behind that the replacement transaction then treats as removed — permanent ghost hits. Needs a loop-until-zero (or search_after pagination) with a hit-count check.

3. **Major — two error paths tell the operator to "rerun with --fresh"** (lib.rs ~313, state.rs ~315) — which re-extracts and re-embeds the entire corpus, the exact amplification this PR exists to avoid; both conditions are recoverable with narrower guidance.

4. The resume-cost finding is only partially addressed: the O(files × plan.files) scan is gone, but `content::resolve` still full-hashes every byte of the corpus on every run — a no-op rerun of a 1TB folder still reads 1TB. A size+mtime short-circuit (with the documented mtime-swap caveat) remains needed.

Also worth a look while in there: alias-orphan keys entering `content_changed` when a whole dup-group is deleted (lib.rs ~676), upfront intent fsync for every todo before phase B (lib.rs ~757) versus per-file, and the unconditional alias-hygiene delete_by_query per run (lib.rs ~1182).

xerj-org:
Merged via main with the review close-out on top (your branch head plus one commit — your fork doesn't allow maintainer edits, so it landed as a main-side merge). What the extra commit does:

1. **Exclusive plan-key ownership** (the confirmed blocker): a current file whose digest resolves to a key already claimed via rel/path_id is now deterministically diverted to `"{key}-claimed-{xxh3_128(rel):032x}"` (your own collision-discriminator scheme) and journaled as a junk entry naming the contested key and owner — first claimant in stable walk order wins, resumes stay deterministic, exit 3.
2. **`delete_by_query` loops until a pass reports `deleted: 0`** (cap 1000 passes, hard error at the cap) — the server implements it as a single size:10000 search, so one call was never complete removal for >10k-doc generations.
3. **Recovery guidance is scoped and honest** — legacy-collision names the two files and the narrow remedy; journal corruption gives path + byte offset with restore/truncate options, and truthfully labels journal deletion as full re-embed (verified: the done-set is what prevents recompute).
4. Deleting an entire dup-group no longer strands an uncommittable `file_replace_start` pending replacement.

5 new regression tests; suite 78/0.

**Follow-up left for you** (deliberately not bolted on): the no-op-rerun full-rehash — a size+mtime short-circuit needs a persisted per-file mtime (new plan field) and plumbing the plan into `content::resolve`; your comment at lib.rs ~414 rejecting metadata-only shortcuts deserves a considered design rather than a drive-by.

PR #23, opened 2026-07-24
fix(api): resolve wildcard/comma index specs on 9 more endpoints

Vinz2168:
## Summary

Same bug class as the recent `field_caps` wildcard-resolution fix (#10): several endpoints take an `:index` path parameter that ES lets be a single name, comma list, `_all`, or a glob pattern, but the handler passed the raw string straight into `state.engine.get_index()` / `get_or_create_index()` with no resolution — so a wildcard either 404'd, or (for `get_or_create_index`) tripped its index-name validation and 400'd on the literal `*` character.

Note on scope: the task also referenced a prior PR (`fix/async-search-wildcard-resolution`) said to have already fixed this on `_async_search`. That PR doesn't exist anywhere in this repo — checked local branches, `origin`, and `upstream`, and confirmed by reading the handler directly, which still called `get_or_create_index` on the raw string. `async_search_submit` is included in this sweep too, verified broken the same way as everything else below.

### Verified broken (curl against a running instance, `test-wildcard*` matching 2 real indices, before writing any fix)

| Endpoint | Before | After |
|---|---|---|
| `POST /:index/_refresh` | **Silent lie**: `if let Ok(idx) = get_index(...)` had no `else` — returned `200 {"_shards":{"successful":1,...}}` while refreshing nothing | Flushes every matched index, `_shards.total` reflects the real count |
| `POST /:index/_close` | 404 | Closes every matched index, `indices` map lists them all |
| `POST /:index/_open` | 404 | Reopens every matched index (index-sort-on-reopen logic now runs per-index) |
| `POST /:index/_forcemerge` | 404 (already known broken) | Force-merges every matched index, totals summed |
| `POST /:index/_cache/clear` | 404 (already known broken) | `_shards.total` sized to what matched |
| `POST /:index/_admin/segments/fsck` | 404 | fsck's every matched index, reports merged |
| `POST /:index/_search/template` | 404 | Searches every matched index (see merge note below) |
| `POST /:index/_terms_enum` | 404 | Enumerates terms across every matched index (trivial merge — a `BTreeSet`) |
| `GET /_cat/count/:index` | **Silent wrong answer**: `Err(_) => 0` — returned `0` docs indistinguishable from a genuinely empty index | Sums real doc counts across every matched index |
| `POST /:index/_async_search` | 400 `mapper_parsing_exception` (get_or_create_index rejects `*` as an invalid index-name character) | Searches every matched index (see merge note below); a single literal name keeps its existing auto-create convenience, a wildcard/comma spec doesn't |

### Fix approach

A new `resolve_indices_for_op()` helper (next to the existing `resolve_index_selector()`, already used by the alias endpoints) resolves a spec to the `Index` handles it actually matches, mirroring ES's `allow_no_indices` default: a wildcard/comma/`_all` spec matching nothing is a valid (if unhelpful) result, but a literal name that doesn't exist is still a 404.

- **Side-effecting admin ops** (`_refresh`, `_close`, `_open`, `_forcemerge`, `_cache/clear`, `_admin/segments/fsck`) now loop over every resolved index instead of acting on at most one.
- **`terms_enum`/`cat_count`** merge trivially (a term set / a doc-count sum) — no new merge logic needed.
- **`search_template`/`async_search_submit`** are different in kind: real multi-index search means merging *scored hits*, and `search()`'s own merge path is ~1000 lines of sort/collapse/knn/PIT/closed-index handling that doesn't belong in an admin-endpoint wildcard fix. Both get a deliberately **minimal** merge instead: concatenate hits from every resolved index, sort by score descending, truncate to the requested size, sum totals. `async_search_submit`'s `aggregations` are **not** merged across multiple indices — a real, documented gap (aggregations are included only when the spec resolved to exactly one index) — but since a wildcard/comma spec 400'd or 404'd outright before this fix, "hits work, aggs omitted for multi-index" is strictly better than what existed for anyone who could reach that code path, not a regression.

### Endpoints checked and found NOT broken

None — all 9 named in the task, plus `async_search_submit`, were confirmed broken by direct testing before any fix was written.

## ⚠️ Pre-existing bug this fix exposes (not caused by it, not fixed by it)

The ES-compat YAML conformance suite's `search/` directory has **7 failures** on top of this fix that don't reproduce against unmodified `upstream/main` run normally — but **do** reproduce against unmodified `upstream/main` too, once you call a *literal* `POST /<index>/_refresh` (verified directly, zero code changes, on a clean `upstream/main` build):

- `310_match_bool_prefix.yml` — 4 cases (`minimum_should_match`, `analyzer`, and their `multi_match` equivalents)
- `390_doc_values_search.yml` — 1 case (case-insensitive wildcard query on an `index:false` doc-values-only keyword field)
- `395_binary_doc_values_search.yml` — 2 cases (same, binary doc values)

**Root cause**: xerj computes `match_bool_prefix`'s `minimum_should_match`/`analyzer` clause counting, and case-insensitive wildcard matching on `index:false` doc-values-only fields, correctly for memtable-resident docs but **not** for the same docs once flushed to an on-disk segment.

**Why this fix exposes it**: every one of those 7 failing test cases calls the YAML DSL's bare `indices.refresh: {}` in its setup step, which the test runner translates to `POST /*/_refresh`. On unfixed xerj, that call was a silent no-op (`get_index("*")` fails, swallowed) — so the docs stayed memtable-resident (accidentally correctly scored) for every test relying on it. This fix makes `_refresh` actually work, including via wildcard, which is unambiguously correct and necessary — but it also means docs now really do reach the buggy on-disk-segment code path that this suite never exercised before.

This is a real, separate, pre-existing bug in the flush/segment scoring path (`xerj-query`/`xerj-storage` territory, not `xerj-api` index-name resolution) that deserves its own fix and its own PR. Flagging it here for visibility rather than scope-creeping this one — happy to pick it up next if useful.

## Test plan
- [x] `cargo build --release -p xerj-api`
- [x] `cargo clippy -p xerj-api --no-deps` (no warnings)
- [x] `cargo fmt -p xerj-api -- --check`
- [x] Manual empirical verification (curl) of current (pre-fix) broken behavior for all 10 endpoints, **before** writing any fix
- [x] Manual end-to-end verification (curl) of fixed behavior for all 10 endpoints against 2 real indices matching a wildcard pattern, plus regression checks (literal nonexistent name still 404s; wildcard matching zero indices succeeds with an empty/zero result, not an error; single-index `async_search_submit` aggregations and auto-create-on-literal-name behavior unchanged)
- [x] `es-compat-yaml` conformance suite: `indices/` (137/137), `bulk/` (21/21), `smoke/` (30/30) — clean
- [x] `es-compat-yaml` conformance suite: `search/` (66 files) — 439 passed, 2 skipped, **7 failed**; root-caused all 7 to the pre-existing flush/segment scoring bug described above (reproduced identically on unmodified `upstream/main` via a literal refresh call), not to this fix

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -14462,7 +14462,43 @@ async fn resolve_indices_for_op(
     spec: &str,
 ) -> Result<Vec<(String, std::sync::Arc<xerj_engine::Index>)>, xerj_common::XerjError> {
     let is_selector = spec == "_all" || spec.contains('*') || spec.contains(',');
-    let names = resolve_index_selector(state, spec).await;
+    let all: Vec<String> = state
+        .engine
+        .list_indices()
+        .await
+        .into_iter()
+        .map(|i| i.name)
+        .collect();
+    let mut names: Vec<String> = Vec::new();
+    for part in spec.split(',').map(str::trim).filter(|s| !s.is_empty()) {
+        if part == "_all" || part.contains('*') {
+            for n in &all {
+                // ES excludes hidden (dot-prefixed) indices from default
+                // wildcard/_all expansion — `POST /*/_close` must never
+                // sweep in `.xerj_users`-class system indices. A pattern
+                // that itself starts with a dot opts back in, matching
+                // ES's hidden-index addressing.
+                if n.starts_with('.') && !part.starts_with('.') {
+                    continue;
+                }
+                let matches = part == "_all" || part == "*" || glob_match_simple(part, n);
+                if matches && !names.contains(n) {
+                    names.push(n.clone());
+                }
+            }
+        } else {
+            // Concrete name inside a comma list: ES's default
+            // `ignore_unavailable=false` 404s the whole request on a
+            // missing name instead of silently dropping it —
+            // `POST /real,typo/_refresh` must not report success.
+            if !all.iter().any(|n| n == part) {
+                return Err(xerj_common::XerjError::index_not_found(part));
+            }
+            if !names.iter().any(|n| n == part) {
+                names.push(part.to_string());
+            }
+        }
+    }
     let handles: Vec<(String, std::sync::Arc<xerj_engine::Index>)> = names
         .into_iter()
         .filter_map(|n| {
@@ -19099,6 +19135,25 @@ pub async fn close_index(
     State(state): State<AppState>,
     Path(index): Path<String>,
 ) -> impl IntoResponse {
+    // ES 8's `action.destructive_requires_name` defaults to true: closing
+    // by wildcard/_all is refused outright — only concrete names may close.
+    if index == "_all" || index.contains('*') {
+        return (
+            StatusCode::BAD_REQUEST,
+            Json(json!({
+                "error": {
+                    "root_cause": [{
+                        "type": "illegal_argument_exception",
+                        "reason": "Wildcard expressions or all indices are not allowed"
+                    }],
+                    "type": "illegal_argument_exception",
+                    "reason": "Wildcard expressions or all indices are not allowed"
+                },
+                "status": 400
+            })),
+        )
+            .into_response();
+    }
     let handles = match resolve_indices_for_op(&state, &index).await {
         Ok(h) => h,
         Err(e) => return ApiError::new(e).into_response(),
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -12989,12 +12989,18 @@ pub async fn refresh_index(
     State(state): State<AppState>,
     Path(index): Path<String>,
 ) -> impl IntoResponse {
-    if let Ok(idx) = state.engine.get_index(&index) {
+    let handles = match resolve_indices_for_op(&state, &index).await {
+        Ok(h) => h,
+        Err(e) => return ApiError::new(e).into_response(),
+    };
+    for (_, idx) in &handles {
         let _ = idx.flush().await;
     }
+    let n = handles.len() as u32;
     Json(json!({
-        "_shards": { "total": 1, "successful": 1, "failed": 0 }
+        "_shards": { "total": n, "successful": n, "failed": 0 }
     }))
+    .into_response()
 }
 
 // ─────────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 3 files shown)


xerj-org (changes requested):
The direction is right (`resolve_indices_for_op()` layered on the existing `resolve_index_selector()`, honest writeup), but this can't merge yet:

1. **Blocker — real conformance regression hidden by a broken CI gate.** The ES-compat YAML job on this PR actually ends `1345 passed · 15 failed` (baseline main: `1360 · 0`), but the "0 failures required" step doesn't propagate the runner's failure count, so the check shows green. Beyond the 7 disclosed search/doc-values failures, there are 8 undisclosed `vectors/135_knn_query_nested_search_ivf` failures — nested kNN returns 0 hits once `/*/_refresh` actually flushes. Repo policy is no known-failures list; the flush/nested-kNN bugs need to land first (or the ordering agreed explicitly). The CI gate itself is being fixed on main separately.

2. **Major — comma lists silently drop missing concrete names**: `POST /real-index,typo/_refresh` returns 200 with `_shards.total: 1`; ES 404s under default `ignore_unavailable=false`, and `delete_index`/`get_index_inner` on main already implement the per-part semantics — please reuse those.

3. **Major — wildcards sweep in system indices on destructive ops**: `resolve_index_selector` expands over `list_indices()` unfiltered, so `POST /*/_close` closes `.xerj_users`/`.xerj_passkeys`/`.xerj_magic_links`, and `/*/_terms_enum` can enumerate values out of them. ES 8 excludes hidden/dot indices from default expansion and rejects wildcard `_close` (`action.destructive_requires_name`). Pre-existing exposure on `DELETE /*`, but this extends it to 9 more surfaces.

Minor (fine as follow-ups): `expand_wildcards`/`ignore_unavailable`/`allow_no_indices` unsupported; multi-index `from` applied per-index instead of after the global merge; multi-index `aggs` silently omitted on async_search with no marker; `forcemerge` 404 body drifts from the ES error shape.

xerj-org:
Pushed the full close-out to this branch (maintainer edit). The review blocker is resolved the right way — by fixing the engine bugs the PR exposed, not by hiding them:

**Conformance: full suite now passes 1360/0/3 WITH this PR's real `/*/_refresh`** (previously 1345/15 — and main only shows 1360 because the broken no-index refresh kept these bugs confined to the memtable). Four segment-only engine bugs fixed:
1. Nested kNN returned 0 hits for segment-resident parents — the nested collector didn't unwrap the `{_id,_seq_no,_source}` stored-doc shape the way the top-level kNN collector does (all 8 `135_knn_query_nested_search_ivf` cases now pass).
2. The FTS projection dropped `minimum_should_match`, so flushed `match_bool_prefix` with mm returned every single-term match.
3. The FTS projection re-analyzed queries with hardcoded `standard`, so a `whitespace`-analyzer query was wrongly lowercased post-flush.
4. The scored-columnar keyword-wildcard eval matched case-sensitively while every other wildcard path folds — case-insensitive wildcards on DV-only fields returned 0 from segments.

**Review majors also fixed on the branch:**
- Comma lists now implement ES's default `ignore_unavailable=false`: `POST /real,typo/_refresh` 404s instead of silently dropping the missing name.
- Wildcard/`_all` expansion excludes hidden (dot) indices, so `.xerj_users`-class system indices can't be swept; a dot-pattern opts back in. `_close` additionally refuses wildcard/_all outright per ES 8's `action.destructive_requires_name`.

Merging once CI is green.

PR #24, opened 2026-07-25
fix(engine): dynamic mapping infers ISO date strings as date, not text

Vinz2168:
## Summary

The task that kicked off this investigation described a specific root cause — `get_or_create_index()` creating auto-indexes with `Schema::empty()` "without ever consulting registered templates" — discovered by noticing xerj's own internal `xerj-monitoring` index (auto-created when Kibana POSTs monitoring data to `/_monitoring/bulk`) has its `timestamp` field mapped `text` instead of `date`.

**That specific claim turned out to be false**, verified empirically before writing any code: `create_index()` (which `get_or_create_index()` calls) already resolves the highest-priority matching template and merges its mapping in — tested with a general wildcard `index_patterns` match and with the exact literal name `xerj-monitoring`, both correctly produced `"timestamp":{"type":"date"}`. Template-aware auto-create already works.

**The real cause of the reported symptom** is unrelated to templates: nobody registers an index template matching `xerj-monitoring` (it's xerj's own internal index — not something a real ES/Kibana deployment would template), so the field falls through to **dynamic type inference** from the first document's value. That inference — `xerj-engine/src/index.rs::infer_field_type()` — unconditionally mapped every JSON string to `Text`, with zero date recognition. Real Elasticsearch's dynamic mapping has `date_detection: true` by default: a string that fully parses as a standard date/time format infers as `date` before falling back to `text`. xerj had nothing like this at all.

### A trap worth flagging for anyone else who goes looking at this code

There's a *second*, better `infer_field_type` in `xerj-common/src/schema.rs` — it already does real date/IP detection. But it's dead code from real indexing's perspective: its only caller, `ManagedSchema::apply_document`, is itself only called from `schema.rs`'s own test module — nothing in the live indexing path uses it. It's not reused here on purpose: it also reclassifies every short whitespace-free string as bare `Keyword` instead of `Text`, which doesn't match real ES either (ES's dynamic mapping keeps such fields `text` with a `.keyword` *sub-field*, not a top-level type swap) — pulling it in would have bundled a second, unrelated, and arguably incorrect behavior change into this fix.

## Fix

`infer_field_type()`'s `String` arm now checks the value against a small set of standard date/time formats (RFC3339, plain date, naive date-time with/without fractional seconds), gated behind the same cheap byte-prefix pre-check (`does this start "DDDD-"?`) this file already uses for sort-value date normalization — so the parser only runs against strings that are already date-shaped, not every string field in every document. Numbers, bools, arrays, objects, and short-string handling are all untouched.

## Test plan
- [x] `cargo build --release` (full workspace — this touches `xerj-engine`, so it recompiles every downstream crate too)
- [x] `cargo fmt -p xerj-engine -- --check`
- [x] `cargo clippy -p xerj-engine --no-deps` (no warnings)
- [x] Reproduced the exact original bug via `POST /_monitoring/bulk` against a brand-new `xerj-monitoring` index — `timestamp` now maps `date`
- [x] An ISO timestamp (with/without trailing `Z`, with millis) and a plain date all infer as `date`, both via a matching template (already worked) and via pure dynamic inference (previously `text`)
- [x] No false positives: a string that merely starts with a date-shaped prefix but isn't a valid date (`"2026-invoice-number-not-a-date-string"`) still infers `text`
- [x] No accidental scope creep: an ordinary short string (`"ok"`) still infers `text`, not `keyword` — confirms the rejected `xerj-common` version's Keyword-reclassification behavior wasn't accidentally introduced
- [x] Full ES-compat YAML conformance suite: **1360 passed, 0 failed, 3 skipped** (clean — no regressions)

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -1433,6 +1433,9 @@ pub async fn put_mapping(
             if let Some(dyn_val) = body.get("dynamic") {
                 obj.insert("dynamic".into(), dyn_val.clone());
             }
+            if let Some(dd) = body.get("date_detection") {
+                obj.insert("date_detection".into(), dd.clone());
+            }
         } else {
             existing = body.clone();
         }
--- a/engine/crates/xerj-api/src/router.rs
+++ b/engine/crates/xerj-api/src/router.rs
@@ -276,9 +276,18 @@ pub fn build_es_compat_router(state: AppState) -> Router {
             "/_count",
             get(es_compat::count_docs_global).post(es_compat::count_docs_global),
         )
-        .route("/:index/_refresh", post(es_compat::refresh_index).get(es_compat::refresh_index))
-        .route("/:index/_analyze", post(es_compat::analyze_text).get(es_compat::analyze_text))
-        .route("/_analyze", post(es_compat::analyze_text_global).get(es_compat::analyze_text_global))
+        .route(
+            "/:index/_refresh",
+            post(es_compat::refresh_index).get(es_compat::refresh_index),
+        )
+        .route(
+            "/:index/_analyze",
+            post(es_compat::analyze_text).get(es_compat::analyze_text),
+        )
+        .route(
+            "/_analyze",
+            post(es_compat::analyze_text_global).get(es_compat::analyze_text_global),
+        )
         // ── Document operations ────────────────────────────────────────────
         .route("/:index/_doc", post(es_compat::index_doc_auto))
         // POST with an explicit id is the same index op as PUT (auto-create
@@ -342,7 +351,10 @@ pub fn build_es_compat_router(state: AppState) -> Router {
             get(es_compat::field_caps).post(es_compat::field_caps),
         )
         // ── Multi-search ───────────────────────────────────────────────────
-        .route("/_msearch", post(es_compat::msearch).get(es_compat::msearch))
+        .route(
+            "/_msearch",
+            post(es_compat::msearch).get(es_compat::msearch),
+        )
         // Index-scoped multi-search — the path index defaults header lines
         // that omit `index`.
         .route(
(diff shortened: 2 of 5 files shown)


xerj-org (changes requested):
The inference fix itself is correct and cheap (schema-miss prefilter + byte-prefix gate — no hot-path regression), and the root-cause writeup debunking the template theory is appreciated. Two ES-semantics gaps need closing before this merges, because both fail silently:

1. **Non-date values are silently accepted into an inferred date field.** Ingest-time date validation only runs for fields declared `type: date` with an explicit non-empty `format` in a user-supplied mapping; dynamically inferred date fields never pass through it. So doc1 `{"f":"2026-07-25"}` makes `f` a date, then doc2 `{"f":"n/a"}` returns 201 where real ES returns 400 `mapper_parsing_exception` — and doc2 is silently invisible/wrong for date-range queries, sorts, and Kibana time filters. This silent-failure mode didn't exist while everything mapped to text.

2. **`date_detection: false` is accepted-but-ignored** (put_mapping allowlists the key but nothing stores or consults it), so detection is now unconditionally on with no ES-compatible opt-out — the only escape is `dynamic: false`, which kills all dynamic mapping.

Minor: the doc comment claims ES date-detects `epoch_millis` by default — it doesn't (default `dynamic_date_formats` is `strict_date_optional_time` + `yyyy/MM/dd HH:mm:ss Z||yyyy/MM/dd Z`); the code is right, the comment will mislead. A small `#[cfg(test)]` block pinning the accept/reject envelope of `looks_like_date_string` would also lock the contract.

xerj-org:
Pushed the two blocking fixes from the review directly to this branch (maintainer edit):

1. **`date_detection: false` is now honored end-to-end** — the toggle propagates from the raw mapping blob to the engine on index create, boot/restore open, and `PUT /_mapping` (which previously dropped the key entirely in its merge branch), and both schema-evolution paths consult it before inferring `date`.
2. **Non-date values into a default-format date field now 400 with `mapper_parsing_exception`** — validated on the per-doc write funnel before any WAL/memtable write, covering single-doc and per-doc bulk paths (bulk items get status 400). Acceptance is deliberately lenient (year-only, year-month, `yyyy-MM-ddT*`, epoch numbers/strings, arrays, null) so nothing ES would accept gets rejected; fields with an explicit mapping `format` or `ignore_malformed` are excluded via an engine-propagated set, so custom-format values are untouched. Indexes without date fields skip the whole check via an atomic fast-bail.

Verified: 3 new unit tests pin the detection/acceptance envelopes; 9/9 live behavior checks (inference on/off via create and put_mapping, 400s on both write paths, custom-format/epoch/partial-time acceptance); full local ES-YAML conformance run **1360 passed · 0 failed** — identical to main baseline.

Known residual (documented in the commit): the raw-bytes auto-id turbo bulk path skips the validation — it never parses doc bodies, same envelope as the pre-existing explicit-format validation. Merging once CI is green.

PR #27, opened 2026-07-25
fix(engine): serialize single-document publication by id

buger:
## Motivation

Concurrent mutations of the same document could pass their current-state checks independently and then complete WAL, VersionMap, full-text, vector, and response publication in a different order. This made the acknowledged result depend on task scheduling rather than one coherent per-document order.

The observable failures were correctness failures, not merely stale intermediate reads: two concurrent create-only requests could both succeed, concurrent partial updates could lose fields, two operations could consume the same optimistic-concurrency precondition, a lower external version could overwrite the source published by a higher version, and a delete could race an in-flight write.

## Root cause

Sequence-number reservation and individual state checks existed, but there was no shared critical section spanning the complete publication interval for one document ID. Create, optimistic concurrency control, external versioning, update, and delete therefore checked one state and could publish against another.

Fixing only one stage would not establish ordering. For example, serializing VersionMap writes alone would still allow FTS or HNSW publication to complete in reverse order, while serializing only the initial CAS check would allow both operations to proceed after the lock was released.

## What changed

- Added an exact keyed `WritePublicationCoordinator` whose key is the document ID.
- The coordinator stores weak mutex references so historical IDs do not accumulate in a permanent lock registry.
- Same-ID mutations queue behind the same mutex, while unrelated IDs remain concurrent; this is not a fixed-stripe lock and does not add one global write mutex.
- Added sorted and deduplicated multi-key acquisition support so future callers can acquire several IDs without inverted-order deadlocks or self-deadlock from duplicate IDs.
- Moved ordinary/prepared/OCC PUT, create-only, `external` and `external_gte`, update/upsert/noop, and delete current-state checks inside the exact per-ID publication interval.
- Hold the guard from current-state/CAS/liveness checks through WAL, VersionMap, FTS, HNSW, and response construction.
- Keep embedding and `copy_to` preprocessing outside the guarded interval so expensive ID-independent work can still overlap.
- Publish an accepted external version only after the document write succeeds, preventing a failed write from poisoning retries.
- Reworked update to hold one non-reentrant guard across read, merge or script evaluation, noop detection, and publication, which prevents concurrent patches from silently replacing one another.
- Return the existing sequence number and version for a true noop instead of inventing a new acknowledged mutation.
- Added deterministic test hooks at lock acquisition, after WAL/VersionMap, after FTS, after HNSW, and before release; these hooks are test-only.

## Scope and non-goals

This is the first publication-ordering layer and intentionally covers the single-document ordinary/prepared/OCC, create, external-version, update/upsert/noop, and delete paths wired by this change.

It does not claim that every engine write path is now ordered. Parsed, raw, synchronous, and realtime turbo batch paths are not wired to this coordinator in this PR and require separate follow-up work.

It does not make a transaction or cancellation-atomicity claim. Dropping or cancelling a future releases its per-ID guard safely, but this PR does not add a commit marker or rollback for publication stages that already completed before cancellation or failure.

It does not add sequence identity to existing HNSW nodes and therefore does not by itself prove stale-vector suppression across every batch or recovery interleaving.

It does not change the ES wire protocol, API request shape, durability configuration, embedding output, or query ranking.

## Correctness tests

- Exact coordinator tests prove strict same-key ordering, concurrent progress for disjoint keys, sorted and deduplicated multi-key acquisition, bounded weak-entry cleanup across 10,000 historical IDs, cleanup after early error, and cleanup after cancelling a queued waiter.
- A deterministic stage-interleaving test pauses PUT after WAL/VersionMap, proves that delete is queued at the same exact key, resumes PUT, and verifies ordered versions, sequence numbers, and final deletion.
- A 32-round concurrent conditional PUT/delete test proves that exactly one operation can consume a shared `if_seq_no` precondition.
- Sixteen concurrent create-only calls prove that exactly one create succeeds.
- Sixteen concurrent partial updates prove that all independent fields survive rather than being lost.
- Concurrent missing-ID upserts prove that both patches survive.
- Noop coverage verifies that the response reports the existing sequence number and version.
- Concurrent external versions verify that the highest accepted version owns the final source and that a later lower version remains rejected.
- Put/delete/recreate coverage verifies monotonic sequence/version responses, immediate visibility, and the final source after flush.

## Validation

- Focused coordinator and publication-order regression tests passed.
- `xerj-engine` library tests passed `137/137` with only the known `origin/main` Painless flat-chain stack-overflow test excluded; this exclusion is unrelated to this change and is not presented as a full unfiltered library pass.
- Warnings-denied Clippy passed for `xerj-engine` library and tests.
- Rustfmt and `git diff --check` passed.
- The scoped release builds used `cargo build --release -j 32 -p xerj-server` and `cargo build --release -j 32 -p es-yaml-runner`.
- The isolated hard ES-YAML gate used a fresh data directory and the complete 199-file corpus: `1360 passed / 0 failed / 3 skipped / 1363 total`.
- The conformance summary was checked directly rather than inferred from runner exit status, the server was stopped with SIGINT, and no listener remained on the three test ports.

## Manual reproduction

The race can be reasoned about with two clients using the same document ID and the same current `if_seq_no`: client A performs conditional PUT while client B performs conditional DELETE. Before this change, both operations could pass the precondition before either completed publication; the correct result is exactly one success and one version conflict.

The deterministic in-tree reproduction avoids relying on timing or sleeps:

```bash
cd engine
cargo test -p xerj-engine --test integration test_conditional_put_and_delete_same_id_linearize_at_cas -- --exact --nocapture
cargo test -p xerj-engine write_publication_integration_tests::put_holds_publication_key_through_response_before_delete -- --exact --nocapture
```

The first command repeats the competing conditional PUT/delete operation for 32 document IDs and requires exactly one winner per ID. The second pauses PUT at a precise publication stage and proves DELETE cannot enter the same-ID interval until PUT has constructed its response and released the guard.

The broader concurrency cases can be run with:

```bash
cd engine
cargo test -p xerj-engine --test integration test_concurrent_create_and_updates_share_exact_publication_key -- --exact --nocapture
cargo test -p xerj-engine --test integration test_concurrent_external_versions_publish_highest_source -- --exact --nocapture
cargo test -p xerj-engine write_publication::tests -- --nocapture
```

## Risk and trade-offs

Same-ID mutations are now deliberately serialized across their publication stages, so a slow write for one ID increases latency for later writes to that same ID. Writes to different IDs remain independently schedulable.

Embedding and `copy_to` work remains outside the lock to avoid unnecessarily extending contention, which means concurrent requests may perform preprocessing before one later fails create, CAS, or external-version admission. That is wasted work in the losing request but does not publish incorrect state.

The weak-entry cleanup logic is concurrency-sensitive. Tests specifically cover holder release, queued-waiter cancellation, early error, 10,000 transient IDs, and pointer-identity-aware cleanup so a departing guard cannot remove a replacement mutex.

This PR establishes the shared primitive required by later batch work, but reviewers should evaluate it only against the single-document paths listed above rather than treating it as an end-to-end batching or FinanceBench performance change.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -12683,56 +12683,48 @@ pub async fn update_doc(
         if src.is_empty() {
             return update_script_bad_request("script source is required".to_string());
         }
-        match idx.get_document(&id).await {
-            Ok(Some(mut current)) => {
-                if let Err(e) = apply_painless_update(&mut current, &src, &params) {
-                    return update_script_bad_request(e);
-                }
-                return match idx.index_document(Some(id.clone()), current).await {
-                    Ok(resp) => {
-                        state.metrics.record_doc_indexed(&index);
-                        let er = crate::responses::EsDocResponse::updated(
-                            &index,
-                            &resp.id,
-                            resp.version,
-                            resp.seq_no.saturating_sub(1),
-                        );
-                        Json(er).into_response()
-                    }
-                    Err(e) => ApiError::new(xerj_common::XerjError::from(e)).into_response(),
-                };
+        let upsert_body = body.upsert.clone().or_else(|| {
+            if body.doc_as_upsert {
+                body.doc.clone()
+            } else {
+                None
             }
-            Ok(None) => {
-                // Document missing: honour `upsert` / `doc_as_upsert` by
-                // indexing the upsert body as a new document. The script is
-                // not run against the upsert body (matches ES default,
-                // scripted_upsert=false).
-                let upsert_body = body.upsert.clone().or_else(|| {
-                    if body.doc_as_upsert {
-                        body.doc.clone()
-                    } else {
-                        None
-                    }
-                });
-                if let Some(up) = upsert_body {
-                    return match idx.index_document(Some(id.clone()), up).await {
-                        Ok(resp) => {
-                            state.metrics.record_doc_indexed(&index);
-                            let er = crate::responses::EsDocResponse::created(
-                                &index,
-                                &resp.id,
-                                resp.version,
-                                resp.seq_no.saturating_sub(1),
-                            );
-                            Json(er).into_response()
-                        }
-                        Err(e) => ApiError::new(xerj_common::XerjError::from(e)).into_response(),
-                    };
-                }
+        });
+        match idx
+            .transform_document_serialized(&id, upsert_body, move |mut current| {
+                apply_painless_update(&mut current, &src, &params)?;
+                Ok::<_, String>(current)
+            })
+            .await
+        {
+            Ok(Ok(Some(outcome))) => {
+                state.metrics.record_doc_indexed(&index);
+                let resp = outcome.response;
+                let er = if outcome.created {
+                    crate::responses::EsDocResponse::created(
+                        &index,
+                        &resp.id,
+                        resp.version,
+                        resp.seq_no.saturating_sub(1),
+                    )
+                } else {
+                    crate::responses::EsDocResponse::updated(
+                        &index,
+                        &resp.id,
+                        resp.version,
+                        resp.seq_no.saturating_sub(1),
+                    )
+                };
+                return Json(er).into_response();
+            }
+            Ok(Ok(None)) => {
                 let e = xerj_common::XerjError::document_not_found(&id, &index);
                 return ApiError::new(e).into_response();
             }
-            Err(e) => return ApiError::new(xerj_common::XerjError::from(e)).into_response(),
+            Ok(Err(error)) => return update_script_bad_request(error),
+            Err(error) => {
+                return ApiError::new(xerj_common::XerjError::from(error)).into_response();
+            }
         }
     }
 
@@ -17354,27 +17346,42 @@ async fn run_update_by_query(
         if hit.source.is_null() {
             continue;
         }
-        let mut source = hit.source;
         if let Some((src, params)) = script.as_ref() {
-            if !src.is_empty() {
-                if let Err(e) = apply_painless_update(&mut source, src, params) {
-                    failures.push(json!({
-                        "id": hit.id,
-                        "cause": { "reason": e },
-                    }));
-                    continue;
-                }
-            }
-        }
-        // Re-index in place: same `_id`, mutated source → an update, not an
-        // append (verified: `index_document(Some(existing_id), source)`).
-        match idx.index_document(Some(hit.id.clone()), source).await {
-            Ok(_) => updated += 1,
-            Err(e) => {
+            if src.is_empty() {
                 failures.push(json!({
                     "id": hit.id,
-                    "cause": { "reason": e.to_string() },
+                    "cause": { "reason": "script source is required" },
                 }));
+                continue;
+            }
+            let result = idx
+                .transform_document_serialized(&hit.id, None, |mut current| {
+                    apply_painless_update(&mut current, src, params)?;
+                    Ok::<_, String>(current)
+                })
+                .await;
+            match result {
+                Ok(Ok(Some(_))) => updated += 1,
+                Ok(Ok(None)) => {}
+                Ok(Err(error)) => failures.push(json!({
+                    "id": hit.id,
+                    "cause": { "reason": error },
+                })),
+                Err(error) => failures.push(json!({
+                    "id": hit.id,
+                    "cause": { "reason": error.to_string() },
+                })),
+            }
+        } else {
+            // No script: preserve the historical re-index-in-place behavior.
+            match idx.index_document(Some(hit.id.clone()), hit.source).await {
+                Ok(_) => updated += 1,
+                Err(e) => {
+                    failures.push(json!({
+                        "id": hit.id,
+                        "cause": { "reason": e.to_string() },
+                    }));
+                }
             }
         }
     }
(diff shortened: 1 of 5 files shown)


xerj-org (changes requested):
The coordinator is carefully built — the Weak self-cleanup avoids the ABA trap, lock ordering is consistent, no self-deadlock (recursion passes the held guard), and the stage-paused/32-round-CAS/16-way-fan-in tests are excellent. One gap blocks merge:

**Scripted `_update` is still an unguarded read-modify-write.** The PR body claims one non-reentrant guard is held "across read, merge or script evaluation" — but script evaluation lives in the API layer: `update_doc` (es_compat.rs ~12601-12656) does `get_document` → `apply_painless_update` → `index_document` entirely outside the coordinator (which is `pub(crate)` to the engine and unreachable from there). Two concurrent `ctx._source.n += 1` scripts still both read n=0 and publish n=1 — the exact lost-update class the motivation section cites. `_update_by_query` (~17223) has the same window. Either wire an engine-level scripted-update method that acquires the guard (mirroring `update_document_with_upsert`), or correct the PR description and list both paths as explicit non-goals with a follow-up issue.

Smaller items:
- A waiter cancelled while awaiting `lock_owned()` after the holder released leaves a permanently dead Weak entry in the registry (no guard was constructed, so no Drop cleanup; holder saw strong_count == 2 and skipped removal). The cancel test masks it by re-locking the key before asserting. A sweep or cleanup-on-failed-upgrade in `lock_inner` closes it.
- Delete still bumps `dataset_version` before its OCC check while PUT was reworked to bump only after CAS success — inconsistent, and failing conditional deletes needlessly invalidate the response cache.
- Given the per-write DashMap + Arc<Mutex> costs ride the single-doc path in native.rs/binary_protocol.rs, a quick before/after single-doc ingest micro-bench would derisk the beat-ES-ingest posture.

buger:
Thank you. All three correctness findings were valid, and the branch now addresses them.

### Scripted read-modify-write

Added an engine-owned `transform_document_serialized` boundary. Its exact-ID publication guard now spans the authoritative source read, Painless transform, WAL/version-map update, FTS/HNSW publication, and response. Both scripted `POST /{index}/_update/{id}` and scripted `_update_by_query` use this boundary. Missing-document upserts retain Elasticsearch's default `scripted_upsert=false` behavior, and transform failures still return the existing script error shape.

Deterministic tests start 32 `_update` requests behind one barrier and verify that 32 `ctx._source.n += 1` operations finish at `n=32`. A separate 16-way `_update_by_query` test finishes at `n=16`. The engine boundary also has its own 32-way transform test.

### Cancelled waiter cleanup

The acquire path now owns a cancellation-cleanup value before awaiting `lock_owned()`. If that await is cancelled, it removes the matching idle Weak registry entry. The regression test no longer masks the leak by reacquiring the same key: after aborting the queued waiter and releasing the holder, it immediately asserts `registry_len() == 0`.

### Conditional delete cache epoch

`dataset_version` now advances after optimistic-concurrency admission. A rejected conditional delete leaves the response-cache epoch unchanged; a successful delete advances it exactly once. A deterministic test checks both outcomes.

### Single-document ingest measurement

The repository's existing `indexing/single_doc` Criterion benchmark currently panics on unmodified `main` because its setup calls `Engine::new` and `create_index` outside a Tokio reactor after those paths began spawning background tasks. I preserved that failed control log rather than presenting it as a product measurement. The before/after numbers below use an identical harness-only runtime-entry repair in both worktrees and otherwise compare exact current `main` against this branch:

| Arm | Median estimate | Throughput estimate | 95% interval |
|---|---:|---:|---:|
| Exact current `main` (`5a0881d`) | 57.766 µs/doc | 17.311 K docs/s | 45.809–71.699 µs/doc |
| This branch (`dc24130`) | 58.238 µs/doc | 17.171 K docs/s | 46.851–70.563 µs/doc |

The midpoint delta is +0.82% latency / -0.81% throughput. Criterion reports `p=0.88` and “No change in performance detected”; the intervals overlap heavily, so this is evidence that the uncontended coordinator cost is below this benchmark's noise floor, not a claim that the change improves throughput. Both arms used 40 samples, a 2-second warm-up, a 4-second measurement window, the same fat-LTO profile, the same host with other builds/benchmarks paused, and the same harness-only reactor repair.

### Branch and validation

- Rebased base: `5a0881d7c0a996fec7738cf947948e21d2568548`
- Rebased publication commit: `16350ac69d4319ba5d0cfb340934e5e8109a3bbf`
- Review-fix commit: `dc24130cc961ec662776ed5188e82c363b287cb0`
- Targeted publication tests: 10 passed, 0 failed
- Scripted API concurrency tests: 2 passed, 0 failed
- Format: `cargo fmt --all --check` passed
- Strict Clippy: `cargo clippy -p xerj-engine -p xerj-api --all-targets -- -D warnings` passed
- API tests: 53 passed, 0 failed; doc tests 2 passed, 2 ignored
- Engine tests: every suite passed with 323 tests passed and expected ignores after excluding the unchanged-main `painless::tests::long_flat_binary_chain_does_not_overflow_evaluator` stack-overflow test; the unfiltered failure log is preserved
- ES-YAML hard gate: 1,360 passed, 0 failed, 3 skipped, 1,363 total; runner exited 0

The benchmark-harness repair is not included in this PR; it was applied identically and locally only to make the requested A/B runnable. This PR provides same-ID concurrency serialization, not a durable transaction: cancellation or process failure after WAL publication remains outside its claim. Scripted `_update_by_query` also applies each script to the latest source under the guard rather than implementing Elasticsearch snapshot/version-conflict semantics; that pre-existing compatibility gap remains follow-up work.

xerj-org:
Merged — `transform_document_serialized` closes the scripted read-modify-write hole correctly (script evaluation verified inside the per-id guard, wired for both `_update` and `_update_by_query`), the cancelled-waiter registry leak fix holds, and the delete-path `dataset_version` ordering is now consistent with PUT.

Non-blocking follow-ups:
1. A residual TOCTOU dead-Weak leak remains in `remove_if_idle` (write_publication.rs ~142): the strong-count check can race a dropping owner's still-held reference. Much narrower than the fixed leak; a periodic sweep would finish it.
2. Scripted `_update_by_query` now inflates `indices.stats` get metrics (reads via the counted `get_document` path).
3. The cancelled-waiter regression test aborts the waiter before dropping the holder — the interleaving that never leaked; swapping the order would pin the actual fixed race.
4. The noop echo in `update_document_with_upsert` reads `version` from the version map directly, bypassing the `external_versions` overlay.

PR #28, opened 2026-07-25
feat(api): auto-sense OpenSearch vs Elasticsearch clients per request

Vinz2168:
## Summary

xerj advertises itself as Elasticsearch on `GET /` and `GET /_nodes` (the `version`/`distribution` block clients use to decide wire compatibility). An OpenSearch client — `opensearch-py`, `opensearch-js`, or OpenSearch Dashboards itself — talking to xerj sees an Elasticsearch identity it doesn't recognize as compatible, and fails its own version-compatibility check before ever issuing a real query.

Rather than a static `--compat opensearch` flag that forces the SAME behavior for every client, this **detects per-request** from the caller's `User-Agent` header which ecosystem it belongs to, and answers accordingly — the same running instance serves an OpenSearch client and an Elasticsearch client at once, each seeing a response shaped for their own tooling.

### User-Agent detection — verified empirically, not guessed

Before writing any detection logic: installed real `elasticsearch-py`/`opensearch-py` and ran them against a local capture server to see their actual `User-Agent` strings:
- `elasticsearch-py/9.4.1 (Python/3.14.6; elastic-transport/9.4.2)`
- `opensearch-py/3.2.0 (Python 3.14.6)`

Both — and the equivalent JS clients (`elasticsearch-js`, `opensearch-js`) — self-identify with an unambiguous `<name>-py/js/<version>` prefix, so a case-insensitive "contains `opensearch`" check on the whole header is sufficient and low-risk.

### Three configuration tiers — most explicit wins

On top of pure auto-detection, this also does what the follow-up ask specifically wanted: a way to pin deterministic behavior without recompiling.

1. **`--compat-distribution` / `XERJ_COMPAT_DISTRIBUTION`** and **`--compat-version` / `XERJ_COMPAT_VERSION`** (env or flag, flag wins) — forces the identity for **every** client regardless of `User-Agent`. Useful for a client that sends no recognizable `User-Agent`, or for fully deterministic behavior:
   ```
   xerj -d ./data --compat-distribution opensearch --compat-version 2.11.0
   # or, e.g. in docker-compose:
   XERJ_COMPAT_DISTRIBUTION=opensearch XERJ_COMPAT_VERSION=2.11.0 xerj -d ./data
   ```
2. **Neither set (default)** — per-request auto-detection from `User-Agent`:
   ```
   xerj -d ./data
   ```
3. **Neither resolves anything** — unchanged pre-existing behavior, byte-identical to before this change: plain Elasticsearch.

The two settings resolve **independently** — `--compat-distribution opensearch` alone still lets the version auto-resolve, and vice versa.

`--help` documents all three modes with a worked example each.

## Two things found empirically that changed the plan — flagging both for visibility

**1. `GET /` isn't the endpoint that gates OSD's own compatibility check.** Its `savedobjects-service` instead calls `GET /_nodes?filter_path=nodes.*.version,...`. A real OSD 2.11.1 container pointed at a build with only the `GET /` fix logged:
```
This version of OpenSearch Dashboards (v2.11.1) is incompatible with the following OpenSearch nodes in your cluster: v8.13.0 @ 0.0.0.0:9200 (127.0.0.1)
```
`nodes_info()` (`GET /_nodes`) now shares the same `resolve_compat_version()` logic as `es_info()`.

**2. Mirroring the calling client's own self-reported version — the original plan for the auto-detected `version.number` — is wrong, and a real OSD container proved it.** OSD's backend talks to the cluster via `opensearch-js`; its own `opensearch-js` version (`1.1.0` at time of writing) has no relationship to a server version OSD considers compatible. Mirroring it verbatim made a real OSD container log:
```
This version of OpenSearch Dashboards (v2.11.1) is incompatible with the following OpenSearch nodes in your cluster: v1.1.0 @ 0.0.0.0:9200 (127.0.0.1)
```
A client library's own version number just isn't a reliable stand-in for a compatible server version — confirmed, not assumed. Fixed by using a single, empirically-verified constant (`FALLBACK_OPENSEARCH_VERSION = "2.11.0"`, OSD's own version) for the auto-detected path instead. Re-tested end-to-end against the same real OSD container: the "incompatible" error is gone, OSD proceeds to "Starting saved objects migrations".

## Scope note

Getting OSD's saved-objects migrations to fully complete (a 404 on an as-yet-unidentified follow-up request) is a separate, much larger effort — the same shape as everything discovered tonight getting Kibana's own login flow working end-to-end (`_security/privilege`, `profile/_activate`, `_has_privileges`, Basic auth — each its own PR as it surfaced: #16, #17, #25, #26). Out of scope here; this PR is specifically about the **initial version-compatibility handshake**, which now passes via pure auto-detection with zero startup flags — verified against a real OSD 2.11.1 container.

## Test plan
- [x] `cargo build --release -p xerj-api -p xerj-server`
- [x] `cargo clippy -p xerj-api -p xerj-common -p xerj-server --no-deps` (no warnings)
- [x] `cargo fmt --check` (clean on touched files)
- [x] Verified real `User-Agent` strings by running actual `elasticsearch-py`/`opensearch-py` against a capture server (not guessed/assumed)
- [x] curl with spoofed `User-Agent` headers: `opensearch-py`/`opensearch-js` → `distribution: "opensearch"` + resolved version; `elasticsearch-py`/no header → unchanged plain-ES response (byte-identical to before)
- [x] curl with `--compat-distribution`/`--compat-version` (flag and env both): always wins over any `User-Agent`, including a contradicting one
- [x] curl with only `--compat-distribution` set: version still auto-resolves independently
- [x] Full ES-compat YAML conformance suite: 1345 passed, 15 failed, 3 skipped — the 15 are the same pre-existing, already-documented-in-#23/#24 flaky failures (`match_bool_prefix`/`doc_values`/nested-kNN, all pre-existing and timing-related, confirmed unrelated to this change — none involve `User-Agent` or version detection)
- [x] **Real OpenSearch Dashboards 2.11.1 container, pure auto-detection, zero flags on xerj**: the "incompatible" version error is gone; OSD proceeds past the initial handshake to "Starting saved objects migrations" (full OSD functional bring-up is a separate follow-up, see Scope note above)
- [x] Same real OSD container, explicit `--compat-distribution opensearch --compat-version 2.11.0`: identical result, confirming both configuration tiers work end-to-end against a real client, not just curl

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -24,7 +24,7 @@ use crate::{
     responses::{
         BulkItemError, EsBulkItem, EsBulkItemAction, EsBulkItemResult, EsBulkResponse,
         EsDeleteDocResponse, EsDeleteIndexResponse, EsDocResponse, EsGetResponse, EsHit, EsHits,
-        EsHitsTotal, EsIndexResponse, EsInfoResponse, EsSearchResponse,
+        EsHitsTotal, EsIndexResponse, EsInfoResponse, EsSearchResponse, EsVersion,
     },
     state::AppState,
 };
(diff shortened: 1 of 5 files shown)


Vinz2168:
**Follow-up compatibility check** on this: tested Filebeat 7.10.2 and Auditbeat 7.10.2 (both via their official Docker images) against a build of `main` with this PR merged, single-node, `--insecure`, default ES-compat port.

**Filebeat 7.10.2** (ships log file contents — tails log files and forwards each line as a document, the most common Beat for shipping application/server logs) — `output.elasticsearch` pointed straight at xerj:
- Connected cleanly, index template loaded, all 3 test log lines bulk-indexed with full ECS mapping (`@timestamp`, `message`, `agent`, `host`, `log.file.path`).
- Note: this worked without the autosense logic actually classifying the client — Filebeat's Go-based `esclientleg` doesn't send an `opensearch-*`-shaped User-Agent, so it got the default identity (`version.number: 8.13.0`). Filebeat 7.10 just didn't hard-fail on a newer reported ES version. Worth knowing if a stricter Beats-family client shows up later: autosense currently only branches on `opensearch-py`/`opensearch-js`-style UAs, not on Beats.

**Auditbeat 7.10.2** (host-level audit/security data shipper — file integrity monitoring, process/socket/login events, auditd log collection) — `file_integrity` module against a bind-mounted dir:
- Same clean connection + template load. Initial scan and both live fsnotify events (file modified, file created) landed correctly as separate documents with the expected `event.module: file_integrity` / `event.action` fields.

No errors/warnings in either beat's logs beyond an unrelated Docker platform-emulation notice. Not opening this as a separate issue since it's a clean pass — just recording it here since it's adjacent to the client-compatibility work in this PR.

PR #29, opened 2026-07-25
fix(api): complete OpenSearch Dashboards compatibility (product header + _cat/templates/{pattern})

Vinz2168:
## Summary

Follow-up to #28. Two fixes, discovered while root-causing the real OSD `.kibana` saved-objects-migration crash that #28 explicitly left open:

**1. `x-elastic-product` was sent unconditionally, even to detected-OpenSearch callers.** A real OpenSearch server never sends this header at all — Elastic clients use its presence as a product-verification check, so it should only go to callers we've identified as Elastic. Made `es_headers_middleware` distribution-aware, reusing #28's detection (extracted into a shared `is_opensearch_caller()`).

**2. The actual root cause of the crash: `GET /_cat/templates/{pattern}` wasn't routed at all.** OSD's `opensearch-js` client calls `GET /_cat/templates/opensearch_dashboards_index_template*?format=json` to check for its own index template before creating one. Only the bare `/_cat/templates` (no pattern) was routed — the pattern-scoped call fell through to the router's default fallback: **no body at all**, not even xerj's usual ES-shaped JSON error. `opensearch-js` can't parse an empty body and throws a fatal, unrecoverable error — exactly the empty-body symptom (`meta.body: ''`, `content-length: '0'`) reported in #28.

## How #1 was found to be insufficient, and how #2 was actually found

#28 (correctly) suspected the `.kibana` 404 as the crash trigger, since it was the last request logged before the crash. It turned out to be a red herring — `GET /.kibana` was already returning a perfectly well-formed ES-shaped 404 JSON body, confirmed twice (curl, and xerj's own access log).

To settle this properly, rather than trust an intermediate proxy tool again (a naive Python HTTP-parsing proxy used earlier gave an inconclusive, possibly-misleading result), a **raw, non-interpreting TCP relay** (`asyncio`-based, byte passthrough with no HTTP parsing at all) was placed between a real OSD 2.11.1 container and xerj, capturing the exact bytes exchanged. That capture showed the actual failing exchange clearly:
```
GET /_cat/templates/opensearch_dashboards_index_template*?format=json HTTP/1.1
...
HTTP/1.1 404 Not Found
content-length: 0
date: ...
(no body)
```
— not `.kibana` at all. Added `GET /_cat/templates/{pattern}`, mirroring the existing `/_cat/indices/{pattern}` selector semantics (comma-separated names/globs, wildcard-matching-nothing is an empty `200`, never a `404`) and the `format=json` convention already used elsewhere in the `_cat` family.

## Verified end-to-end against real containers — zero flags/config on xerj

- **OpenSearch Dashboards 2.11.1**: saved-objects migration completes (`Creating index .kibana_1`, `Pointing alias .kibana to .kibana_1`, `Finished in 77ms`), server starts, `GET /api/status` reports `status.overall.state: "green"`, `GET /app/home` returns a real page.
- **OpenSearch Dashboards 3.6.0**: same result — `status: "green"`.

Both were stuck in the exact same crash before this PR. This is the actual completion of the auto-sense OpenSearch compatibility work started in #28, not a new/separate feature.

## Test plan
- [x] `cargo build --release -p xerj-api -p xerj-server`
- [x] `cargo fmt --check` / `cargo clippy --no-deps` — clean
- [x] curl: `x-elastic-product`/`warning` correctly present/absent per detected distribution, unchanged for Elasticsearch/Kibana callers
- [x] curl: `GET /_cat/templates/{pattern}` — no match → `200 []`; match → real entries; plain-text `/_cat/templates` unchanged in shape
- [x] Real OpenSearch Dashboards 2.11.1 container: full startup, `status: green`, `/app/home` serves
- [x] Real OpenSearch Dashboards 3.6.0 container: full startup, `status: green`, `/app/home` serves

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -84,19 +84,8 @@ const FALLBACK_OPENSEARCH_VERSION: &str = "2.11.0";
 ///    `EsVersion::default()`.
 fn resolve_compat_version(state: &AppState, headers: &axum::http::HeaderMap) -> EsVersion {
     let mut version = EsVersion::default();
-
     let cfg = &state.config.compat;
-    let user_agent = headers
-        .get(axum::http::header::USER_AGENT)
-        .and_then(|v| v.to_str().ok());
-
-    let is_opensearch = if !cfg.distribution.is_empty() {
-        cfg.distribution == "opensearch"
-    } else {
-        user_agent
-            .map(|ua| ua.to_ascii_lowercase().contains("opensearch"))
-            .unwrap_or(false)
-    };
+    let is_opensearch = is_opensearch_caller(state, headers);
 
     if is_opensearch {
         version.distribution = Some("opensearch".to_string());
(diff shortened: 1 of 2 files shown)


xerj-org (approved):
Reviewed across regression-risk, `_cat/templates` semantics, and routing lenses; conformance verified from the raw job log at 1360/0, identical to main. The header-suppression logic is safe for existing ES clients in auto-detect mode (absent/empty/invalid UA keeps the header; Kibana/elasticsearch-js UAs never match; middleware state threading and layer order check out), and `/_cat/templates/:pattern` conflicts with nothing. Approving; follow-ups worth filing:

- **Forced `--compat-distribution=opensearch` now strips `X-Elastic-Product` for ALL callers**, which hard-breaks any Elastic client pointed at the same instance (`UnsupportedProductError`) — intended per the rustdoc, but the CLI help text and `CompatConfig` docs still describe the setting as shaping only the `GET /` version block. Needs a doc update; same for the CHANGELOG since the UA substring heuristic's false-positive blast radius escalates from cosmetic (version block) to hard client refusal.
- **`_cat/templates` still lists only composable templates** — `engine.legacy_templates` exists but is never included; real ES lists both. Pre-existing on main (not introduced here), but this PR is the natural place to have closed it — good candidate for the next one.
- `glob_match_simple` supports only a single leading/trailing `*`, so `a*b` or `*mid*` patterns silently match nothing vs ES's `simpleMatch`; `_all` is special-cased where ES treats it literally.
- `format=json` rows emit `"version": ""` where ES emits `null`, and omit `composed_of` (ES 7.8+); `format` matching is exact-lowercase and yaml/cbor fall through to text.
- No tests pin the header-suppression matrix (Elastic UA / OpenSearch UA / forced distribution) or the pattern route — a small unit block would lock both contracts.
- Nit: the `cat_templates_pattern` doc comment claims a non-matching concrete name 404s; the implementation always 200s (which matches ES — fix the comment, not the code). Config-file validation of `compat.distribution` (the #28 gap) also remains open.

PR #56, opened 2026-07-27
feat(cache): Bound all segment hydration caches with one process-wide budget

buger:
## Summary

This change replaces seven independent or unbounded segment hydration caches with one process-wide, cgroup-aware retained-memory budget. It covers stored slices, doc values, parsed stored values, sort shadows, ID positions, row sequences, and decoded stored bytes.

The budget is enforced atomically across every index in the process. Cache admission uses a compare-and-swap reservation, the reservation remains attached to the cached value while any reader still holds it, and the final reader refunds it through RAII. When admission is refused, the request keeps the locally built value and returns the same result without publishing it into the cache.

This is a retained-cache bound, not a claim that total server RSS or transient decode memory is solved. Whole-section decoding, allocator fragmentation, memory maps, merge scratch, memtables, vectors, models, and request-owned materialization remain outside this budget and are stated as such in the configuration and production documentation.

## Why this is needed

The existing cache model could retain several expanded representations of the same immutable segment at once. Some caches were completely unbounded, while stored-slice and decoded-byte caches had separate host-derived limits. Those independent limits were additive across cache families and indices, so the process had no authoritative ceiling for hydrated segment state.

PR #43 also made the parsed `stored_value_cache` a production single-flight path for full-corpus fallback aggregations. That correctly removes repeated concurrent decoding, but it increases the importance of bounding the retained parsed representation.

The measured 4,096-document diagnostic workload attributed roughly 206 MiB of peak RSS to publish warming, and a merge heap profile found 226.35 MiB of cumulative live allocations below `warm_segment_at_publish`. This PR addresses retention and lifecycle control for those cache families. It does not claim those profile numbers as an end-to-end memory reduction.

### One shared authority

`ResourceGovernor` owns a single `SegmentHydrationBudget` shared by every index. The default budget is one fifth of the effective memory limit. An explicit configuration value is accepted but clamped to 50% of the effective limit so cache configuration cannot consume all process headroom.

The effective cgroup-v2 limit is now the minimum finite `memory.max` across the leaf and every ancestor, combined with the host limit. Previously the nearest finite value won, which was wrong for a child with a larger limit than its parent.

### Atomic admission and last-reader accounting

Admission uses a checked CAS against the process-wide limit. Every admitted value is wrapped in `CacheResident<T>` with its `BudgetCharge`. Removing a map entry does not refund bytes while a query still holds the value; the charge is refunded only when the last `Arc` is dropped.

An occupied `DashMap` entry reuses the existing resident value without double charging. A refused admission returns an uncached resident to the caller, preserving correctness without increasing retained cache bytes.

### Complete seven-cache lifecycle

The shared authority covers:

- stored slices;
- doc-value columns;
- parsed stored values, including synchronous and cancellation-safe asynchronous loading;
- sort shadows, including cached `None`;
- ID-to-position maps;
- row-sequence maps;
- decoded stored bytes.

Retirement removes all seven families under one lifecycle writer. Failed merges and failed flush finalization also remove every output owner.

### Cancellation-safe prepublication warming

Merge and flush warming use a shared last-owner prepublication lease. The asynchronous owner and blocking worker each hold the lease. If the asynchronous task is cancelled while `spawn_blocking` continues, cleanup waits until the blocking worker exits, preventing a late worker from republishing a cache entry after cleanup.

The lease is committed only after authoritative publication. Merge shortcut-count seeding remains before `apply_merge`, and `apply_merge` plus lease commit are adjacent synchronous operations. A warming panic remains non-fatal: the valid merge output is still published with cold or partially warmed caches.

### Conservative retained-size estimation

Estimation includes map keys, resident wrappers, `Arc` ownership, vector capacities, doc-value structures, roaring containers, sort-shadow nested `Arc<Vec<_>>` allocations, strings, arrays, nested JSON, and insertion-only `serde_json`/`IndexMap` capacity.

The parsed-value estimator deliberately overcharges IndexMap metadata using the pinned insertion-growth bounds. Trace labels describe these values as `estimated`; the PR does not present them as allocator-exact RSS.

### Machine-readable runtime evidence

Every ingest-memory trace event includes the authoritative segment-hydration snapshot:

- configured/effective limit;
- current charged bytes;
- peak charged bytes;
- refusal count;
- accounting-error count.

The bounded-suite analyzer validates these fields and rejects missing or inconsistent evidence. The trace artifact remains included in the run manifest and SHA-256 inventory.

## Configuration

The default is derived automatically:

```toml
[resources]
segment_hydration_cache_bytes = 0
```

`0` means one fifth of the effective host/cgroup memory limit. A nonzero value requests an explicit byte limit and is clamped to half of the effective memory limit.

The active values are visible through resource-governor/node statistics and opt-in ingest-memory traces.

## Failure behavior

Budget pressure does not change query results. When a cache entry cannot be admitted:

1. the request keeps the value it already built;
2. the value is not inserted into the shared cache;
3. the refusal counter increments;
4. later requests may rebuild it or use another bounded path;
5. retirement and restart semantics remain unchanged.

Arithmetic overflow fails admission. Accounting underflow increments an explicit error counter and triggers a debug assertion instead of silently wrapping.

## Verification

The final candidate was rebased onto upstream `v1.0.0-rc.5` (`16d6df0e9da2c5bf908d6176ac0f41e68edfff95`) and independently reviewed.

- ES-YAML conformance: `1360 passed / 0 failed / 3 skipped`;
- `xerj-engine`: `212/212`;
- `xerj-storage`: `91/91`;
- `xerj-api`: `59/59`;
- bounded analyzer: `9/9`;
- formatting: passed;
- diff check: passed;
- all-target, all-feature Clippy with warnings denied for engine, storage, API, and server: passed.

The engine suite was run with `RUST_MIN_STACK=8388608` because current main contains an independent Painless flat-chain test that aborts on Rust's default test-thread stack. That bug is reproduced on unchanged main and has a separate fix; this PR does not modify `painless.rs`.

Targeted coverage includes:

- 64-thread CAS contention without exceeding the limit;
- a shared ceiling across two indices;
- refusal with correct request-local fallback;
- last-reader refund after map retirement;
- occupied/vacant publication races;
- retirement racing all seven cache builders without resurrection;
- asynchronous stored-value cancellation and followers;
- failed merge cleanup;
- failed and panicked flush finalization after warming;
- asynchronous owner cancellation while a detached blocking worker continues;
- non-fatal merge warm panic;
- tightest cgroup ancestor selection;
- conservative sort-shadow and nested parsed-object estimates;
- literal engine-emitted trace labels and budget snapshots.

## Manual verification

```bash
cd engine
cargo fmt --all -- --check
cargo clippy -p xerj-engine -p xerj-storage -p xerj-api -p xerj-server --all-targets --all-features -- -D warnings
RUST_MIN_STACK=8388608 cargo test -p xerj-engine --lib -- --test-threads=1
cargo test -p xerj-storage --lib
cargo test -p xerj-api --lib
cd ..
python3 -m unittest discover demo/usecases/autoindex/scale/bounded/tests
```

Run the repository's ES-YAML runner against a fresh release server and empty data directory. The required result is:

```text
1360 passed
0 failed
3 skipped
1363 total
```

For a pressure experiment, configure `resources.segment_hydration_cache_bytes` below the retained working set and enable ingest-memory tracing. The run is valid only when the trace shows a nonzero refusal count, `current <= peak <= limit`, zero accounting errors, correct query/result digests, and eventual charge release after retirement/shutdown.

## Tradeoffs and non-claims

- Cache misses under pressure may rebuild data and use more CPU or I/O; correctness is preserved.
- Estimation intentionally overcharges some metadata to keep admission conservative.
- This PR does not reserve memory before full stored-section decode, so it does not bound transient hydration peak.
- This PR does not establish the 8 GiB full-server target or the 368-PDF FinanceBench result.
- No throughput or end-to-end RSS improvement is claimed from unit tests or heap attribution.

The next dependent change is bounded selected-row and field projection from stored sections, integrated through this shared lifecycle. That removes the transient full-value materialization for late hydration and will be evaluated with matched 10K, FinanceBench 4, FinanceBench 20, and full FinanceBench runs.

### DIFF
--- a/demo/usecases/autoindex/scale/bounded/analyze.py
+++ b/demo/usecases/autoindex/scale/bounded/analyze.py
@@ -51,13 +51,13 @@ MEASUREMENTS = {
     "merge_json_buffer": "unavailable",
     "merge_parsed": "unavailable",
     "merge_encoded": "unavailable",
-    "cache_stored_slices": "unavailable",
-    "cache_doc_values": "unavailable",
-    "cache_stored_values": "unavailable",
-    "cache_sort_shadow": "unavailable",
-    "cache_id_positions": "unavailable",
-    "cache_row_sequences": "unavailable",
-    "cache_decoded_stored": "unavailable",
+    "cache_stored_slices": "estimated",
+    "cache_doc_values": "estimated",
+    "cache_stored_values": "estimated",
+    "cache_sort_shadow": "estimated",
+    "cache_id_positions": "estimated",
+    "cache_row_sequences": "estimated",
+    "cache_decoded_stored": "estimated",
 }
 
 
@@ -86,6 +86,7 @@ def read_ndjson(path: Path) -> list[dict]:
 def validate_trace(path: Path) -> dict:
     rows = read_ndjson(path)
     problems = []
+    hydration_snapshots = []
     if any(row.get("schema") != TRACE_SCHEMA for row in rows):
         problems.append("schema mismatch")
     seq = [row.get("seq") for row in rows]
@@ -103,6 +104,31 @@ def validate_trace(path: Path) -> dict:
     if max(int(row.get("dropped_events", 0)) for row in rows) != 0:
         problems.append("dropped_events is nonzero")
     for position, row in enumerate(rows, 1):
+        hydration = row.get("segment_hydration_cache")
+        required_hydration_fields = (
+            "limit",
+            "current",
+            "peak",
+            "refused",
+            "accounting_errors",
+        )
+        if not isinstance(hydration, dict) or any(
+            type(hydration.get(field)) is not int or hydration[field] < 0
+            for field in required_hydration_fields
+        ):
+            problems.append(
+                f"record {position}: missing or invalid segment hydration cache snapshot"
+            )
+        else:
+            hydration_snapshots.append(hydration)
+            if hydration["current"] > hydration["peak"] or hydration["peak"] > hydration["limit"]:
+                problems.append(
+                    f"record {position}: invalid segment hydration cache current/peak/limit"
+                )
+            if hydration["accounting_errors"] != 0:
+                problems.append(
+                    f"record {position}: segment hydration cache accounting_errors is nonzero"
+                )
         gauges = row.get("logical")
         if not isinstance(gauges, list) or [gauge.get("category") for gauge in gauges] != CATEGORIES:
             problems.append(f"record {position}: logical category vocabulary/order mismatch")
@@ -177,6 +203,19 @@ def validate_trace(path: Path) -> dict:
             for category in CATEGORIES
         }
         if not any("logical category vocabulary/order mismatch" in problem for problem in problems)
+        else None,
+        "segment_hydration_cache": {
+            "limit_bytes": hydration_snapshots[-1]["limit"],
+            "max_current_bytes": max(snapshot["current"] for snapshot in hydration_snapshots),
+            "peak_bytes": max(snapshot["peak"] for snapshot in hydration_snapshots),
+            "admission_refusals": max(
+                snapshot["refused"] for snapshot in hydration_snapshots
+            ),
+            "accounting_errors": max(
+                snapshot["accounting_errors"] for snapshot in hydration_snapshots
+            ),
+        }
+        if hydration_snapshots
         else None,
         "problems": problems,
     }
--- a/demo/usecases/autoindex/scale/bounded/tests/test_suite.py
+++ b/demo/usecases/autoindex/scale/bounded/tests/test_suite.py
@@ -42,14 +42,42 @@ class BoundedSuiteTests(unittest.TestCase):
         self.assertEqual(parsed["data_dir"], value)
 
     def test_trace_validator_rejects_gap_and_accounting_error(self):
+        # Literal engine-emitted vocabulary: keep this fixture independent of
+        # the analyzer's expected mapping so a stale expectation cannot make
+        # the test pass tautologically.
+        engine_measurements = [
+            "serialized_size",
+            "exact_capacity",
+            "serialized_size",
+            "estimated",
+            "exact_capacity",
+            "estimated",
+            "exact_capacity",
+            "estimated",
+            "estimated",
+            "unavailable",
+            "unavailable",
+            "unavailable",
+            "unavailable",
+            "unavailable",
+            "estimated",
+            "estimated",
+            "estimated",
+            "estimated",
+            "estimated",
+            "estimated",
+            "estimated",
+        ]
         gauges = [
             {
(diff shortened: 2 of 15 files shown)


xerj-org:
CI is red on **Build + Test** (all other checks pass). One test fails:

```
---- segment_cache_budget::tests::underflow_is_detected_without_wrapping_the_counter stdout ----
thread '...::underflow_is_detected_without_wrapping_the_counter' panicked at
crates/xerj-engine/src/segment_cache_budget.rs:271:9:
debug builds must surface double release

test result: FAILED. 211 passed; 1 failed
```

Root cause: CI runs `cargo test --release --workspace` (`.github/workflows/ci.yml`), and
`engine/Cargo.toml`'s `[profile.release]` does not set `debug-assertions`, so it defaults to
off. The `debug_assert!(false, "segment hydration cache accounting underflow")` inside
`checked_sub` therefore compiles out, `catch_unwind` returns `Ok`, and
`assert!(result.is_err(), ...)` fails. The test can only ever pass in a debug build.

The production logic looks fine — the refund path still saturates via `checked_sub` and bumps
`accounting_errors`, which is the behaviour that actually matters. It is just the panic
expectation that is profile-dependent.

Suggested fix (author's call on intended semantics):

- gate the `catch_unwind` block behind `#[cfg(debug_assertions)]` and keep the
  `accounting_errors == 1` / `current == 0` assertions unconditional, or
- drop the panic expectation entirely and assert only on the counters, or
- if a double-release really should be loud in release builds too, promote the `debug_assert!`
  to a real `assert!`/`panic!`.

Not merging until this is green. Holding the PR open.

xerj-org:
Addendum after reproducing locally (`cargo test --release -p xerj-engine --lib segment_cache_budget`
→ same single failure, deterministic): the profile dependency in that test is not limited to the
`assert!(result.is_err(), ...)` line, so the first suggested fix above is not sufficient on its own.

`release()` calls `checked_sub` twice — once for `category_current[category]`, once for `used`:

- debug build: the first `checked_sub` hits `debug_assert!(false, ...)` and unwinds, so `used` is
  never touched and `accounting_errors == 1`.
- release build: nothing panics, both `checked_sub` calls fail and each bumps the counter, so
  `accounting_errors == 2` and line 274's `assert_eq!(snapshot.accounting_errors, 1)` fails next.

So a double release is counted once in debug and twice in release — that is also a telemetry
question for production, not just for the test: `accounting_errors` currently reports 2 per
underflow event in the profile you actually ship. Either count once per `release()` (e.g. bump the
counter in `release()` and let `checked_sub` just saturate) or assert `>= 1` and document that the
counter is per-counter, not per-event.

Still holding the PR — no push access to the fork branch here (`maintainerCanModify` is false), so
this needs a push from your side.

PR #86, opened 2026-07-31
fix(api): treat OpenSearch's flat_object as an alias for flattened

Vinz2168:
## Summary
- OpenSearch's `flat_object` field type indexes without expanding sub-fields — the same semantics as `flattened`, just a different type name.
- xerj only recognized `flattened`, so any mapping using `flat_object` (as OpenSearch's own sample dashboards/UBI sample data do) was rejected as an unsupported field type, and lost the flattened reconstruction/highlighting/`ignore_above` handling that comes with it.
- Every `flattened` special case in `es_compat.rs` now also treats `flat_object` as an alias.

## Test plan
- [x] `cargo build -p xerj-api`
- [x] `cargo fmt --check`
- [x] `cargo test -p xerj-api --lib` (94 passed)
- [x] Verified live against a real OpenSearch Dashboards 3.6.0 instance importing the UBI sample dashboards/data (which use `flat_object`) — no more "unsupported field type" rejection.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -1031,6 +1031,10 @@ fn is_supported_field_type(t: &str) -> bool {
             | "object"
             | "nested"
             | "flattened"
+            // OpenSearch's name for the same "index without expanding
+            // sub-fields" object type — every "flattened" special case
+            // below also treats this as an alias.
+            | "flat_object"
             | "geo_point"
             | "geo_shape"
             | "point"
@@ -4422,7 +4426,7 @@ fn synthetic_transform_object_ext2(
         // Flattened fields reconstruct as a single-level dotted-key map
         // (`{host:{name:x}}` -> `{"host.name":x}`), not a re-nested object.
         // Collapse here and skip the normal object recursion.
-        if child_type == Some("flattened") {
+        if matches!(child_type, Some("flattened") | Some("flat_object")) {
             if let Some(cv) = target.get_mut(&name) {
                 match cv {
                     Value::Object(_) => {
@@ -9110,7 +9114,7 @@ pub async fn search(
                         .and_then(|p| p.get(field));
                     let ftype_static: &'static str = match fspec.and_then(|f| f.get("type")).and_then(Value::as_str) {
                         Some("keyword") => "keyword",
-                        Some("flattened") => "flattened",
+                        Some("flattened") | Some("flat_object") => "flattened",
                         _ => "other",
                     };
                     let explicit = fspec.and_then(|f| f.get("ignore_above")).and_then(Value::as_u64).map(|n| n as usize);
@@ -9452,12 +9456,14 @@ pub async fn search(
                         // include the root itself (leaf-only expand may
                         // have dropped it).
                         let is_flattened = |field: &str| -> bool {
-                            mapping_for_fields.as_ref()
-                                .and_then(|m| m.get("mappings").and_then(|mm| mm.get("properties")).or_else(|| m.get("properties")))
-                                .and_then(|p| p.get(field))
-                                .and_then(|f| f.get("type"))
-                                .and_then(Value::as_str)
-                                == Some("flattened")
+                            matches!(
+                                mapping_for_fields.as_ref()
+                                    .and_then(|m| m.get("mappings").and_then(|mm| mm.get("properties")).or_else(|| m.get("properties")))
+                                    .and_then(|p| p.get(field))
+                                    .and_then(|f| f.get("type"))
+                                    .and_then(Value::as_str),
+                                Some("flattened") | Some("flat_object")
+                            )
                         };
                         // First pass: identify declared flattened roots
                         // that match the pattern (either among current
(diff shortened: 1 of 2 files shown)


xerj-org:
Reviewed for **v1.0.0-rc.9**. This is the only one of #86–#89 with **no blocker** — the alias is faithful to `flattened` at every site it touches and introduces no regression. I'm still holding it out of rc.9, for a reason that's about completeness rather than safety, and I'd rather say why than merge it quietly.

### It only half-fixes the case it exists for

The commit names OpenSearch's sample dashboards/UBI data as the motivation. OpenSearch Dashboards discovers index-pattern fields through `_field_caps`, not `GET _mapping` — and the alias doesn't reach `_field_caps`.

Only `GET /{index}/_mapping` round-trips it, because `get_mapping` (es_compat.rs:1846-1859) returns the stored JSON verbatim. The other three introspection surfaces derive the type from the **native schema**, not the stored mapping:

- `field_caps` → `native_type_to_es_str(&field.field_type)` (:17070)
- `global_field_caps` → same (:25448)
- `get_mapping_field` → `schema_to_es_properties` (:25511, :25526-25529)

`es_type_to_native` sends both `flattened` and `flat_object` through the `_ => FieldType::Object` catch-all (:12995), and `native_type_to_es_str` maps `FieldType::Object => "object"` (:17207). `collect_multi_fields` only emits `fields`/`properties` children, never the root, so nothing overrides it.

Net effect: index creation now succeeds where it previously 400'd — a real improvement — but the field surfaces to the dashboard as `{"type": "object", "searchable": true, "aggregatable": true}`. The failure mode moved from a loud `mapper_parsing_exception` at create time to a silently mistyped field at visualisation time, which is arguably harder to debug.

To be fair: this is **identical pre-existing behaviour for ES `flattened`**, so it's not a regression this PR introduces. It's the reason the PR doesn't yet accomplish what it sets out to.

### No test exercises any of the new arms

The diff is 17 insertions in one file and zero test changes; `grep -rl flat_object engine/tests/` returns nothing. The repo has several `flattened` fixtures — `search/340_flattened.yml`, `600_flattened_ignore_above.yml`, `aggregations/terms_flattened_field.yml`, `indices.create/20_synthetic_source.yml` and more — and none has a `flat_object` counterpart. So all six new match arms (:1037, :4429, :9117, :9476, :9513, :11734) execute zero times in CI. Green checks here aren't weak evidence, they're no evidence.

Cloning two of those fixtures with `type: flat_object` would cover the synthetic-source collapse, the fields-expansion roots and the `ignore_above` pruning in one commit — and would also have surfaced the `_field_caps` gap above.

### Smaller things

- The shared validator was widened beyond the mapping path, so `runtime: {x: {type: flat_object}}` now returns 200.
- The commit message describes highlighting handling that isn't in the code.
- One of the six changed sites is unreachable.

### Suggestion

Add a `FieldType::Flattened` (or map both names to a shared native type that `native_type_to_es_str` can render back correctly), so `_field_caps` reports the field as `flattened`/`flat_object` rather than `object` — that fixes the alias *and* the pre-existing `flattened` gap in one go — plus the two cloned fixtures. With those, this is a clean merge for rc.10.

Holding it out of rc.9 only because that release is already merged, tagged-pending and fully verified, and adding an untested change at the end for a partial fix means re-running the whole gate without gaining the thing the PR is for.

Vinz2168:
Thanks — the `_field_caps` gap was the right thing to hold this on. Fixed the main issue plus two of the smaller ones; being upfront about the two I couldn't close.

- **`_field_caps` reports `object` instead of the real type**: added `declared_flattened_type`, which consults the stored mapping (already available for the multi-fields case `_field_caps` handles) and overrides the generic `object` string with the real declared `flattened`/`flat_object` — applied to `field_caps`, `global_field_caps`, and `GET /{index}/_mapping/field/{field}`. Scoped as a per-field override, not a wholesale swap of the properties source, so a dynamically-inferred field with no stored-mapping entry still falls back to the schema-derived type exactly as before. Verified live: `curl .../_field_caps?fields=attrs` now returns `{"attrs":{"flat_object":{...}}}` instead of `{"attrs":{"object":{...}}}`.
- **Runtime-field validator leak**: confirmed — `is_supported_field_type` is shared between the mapping path and `validate_runtime_fields`, so widening it for `flat_object` let it through for runtime fields too, where real ES doesn't allow it. Added an explicit exclusion scoped to `validate_runtime_fields` only.
- **Test coverage**: added a YAML fixture cloned from `340_flattened.yml`/`600_flattened_ignore_above.yml` covering exists-query, fields-option/synthetic-source, and ignore_above pruning with `type: flat_object` — ran it against a live built instance rather than just adding it untested (caught one broken case in the process: a `field_caps:` action, which turns out the project's YAML runner doesn't implement as a verb at all — dropped from the fixture, covered instead with a proper Rust end-to-end test that goes through the real HTTP route, including one that fails without the fix and passes with it).

Two things I did **not** resolve, flagging rather than guessing:
- The commit message on the original PR does describe highlighting handling that isn't in the diff — didn't rewrite the pushed/reviewed commit over it, noting it here instead.
- I couldn't conclusively identify which of the six original match arms you flagged as unreachable — traced each one I could match to current line numbers and each looked live to me, but I didn't have a precise enough reference into a ~28k-line function to be confident. Happy to take another pass if you can point at the specific one.

xerj-org:

PR #87, opened 2026-07-31
feat(query): add script query support

Vinz2168:
## Summary
- A standalone `script` query — a boolean filter (no scoring), distinct from `script_score` (a `function_score` scoring function) — was completely unparseable (`unknown query type \`script\``), breaking any client relying on it. OpenSearch's UBI sample dashboards use it to filter on a computed field via a Painless boolean predicate ("Top Searches Without Results" panel).
- Adds `QueryNode::Script` (`ast.rs`), its parser (`parser.rs`), planner wiring (`planner.rs` always doc-scans it, like `Percolate` — no index can back an arbitrary Painless predicate), and evaluation in `doc_matches_query` (`index.rs`).
- The script must evaluate to an actual boolean to match: erroring or returning any other type fails closed, matching real Elasticsearch's script-cast-exception-drops-doc behavior in filter contexts, rather than truthy-coercing a non-boolean result (e.g. a script that just returns a number no longer incorrectly matches whenever that number is non-zero).

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test -p xerj-query -p xerj-engine --lib` (254 passed; the 2 failures in `snapshot_path_security_tests` are pre-existing on `main`, unrelated to this change — reproduced on bare `main` too)
- [x] Verified live against a real OpenSearch Dashboards 3.6.0 instance importing the UBI sample dashboards: the "Top Searches Without Results" panel, which was previously 400ing, now renders real data with no errors.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -13197,6 +13197,93 @@ mod malformed_bulk_route_tests {
     }
 }
 
+#[cfg(test)]
+mod script_query_end_to_end_tests {
+    //! Regression coverage for a real bug: the `script` query's unit tests
+    //! (in xerj-engine's index.rs) all exercised `doc_matches_query`
+    //! directly, never a real `POST /{index}/_search` — so they passed
+    //! while the query silently returned zero hits for every real search,
+    //! because `is_doc_scan_query` never routed `QueryNode::Script` into
+    //! doc-scan. This test goes through the actual HTTP route so a
+    //! regression here fails loudly.
+    use super::*;
+    use axum::{
+        body::{to_bytes, Body},
+        http::Request,
+    };
+    use tower::ServiceExt;
+    use xerj_common::{
+        config::{Config, WalSync},
+        metrics::Metrics,
+    };
+    use xerj_engine::Engine;
+
+    fn test_state() -> AppState {
+        let dir = tempfile::tempdir().expect("tempdir").keep();
+        let mut config = Config::default();
+        config.server.data_dir = dir.to_string_lossy().into_owned();
+        config.storage.wal_sync = WalSync::Async;
+        let metrics = Metrics::new().expect("metrics");
+        let engine = Engine::new(config.clone()).expect("engine");
+        AppState::new(config, engine, metrics)
+    }
+
+    #[tokio::test]
+    async fn script_query_matches_an_indexed_document_via_real_search() {
+        let router = crate::router::build_es_compat_router(test_state());
+
+        let index_response = router
+            .clone()
+            .oneshot(
+                Request::post("/script-query-e2e/_doc/1")
+                    .header("content-type", "application/json")
+                    .body(Body::from(r#"{"x": 10}"#))
+                    .expect("request"),
+            )
+            .await
+            .expect("index response");
+        assert!(
+            index_response.status().is_success(),
+            "indexing failed: {:?}",
+            index_response.status()
+        );
+
+        let search_body = serde_json::to_vec(&json!({
+            "query": {
+                "script": {
+                    "script": {"source": "doc['x'].value > 5"}
+                }
+            }
+        }))
+        .expect("serialize query");
+
+        let search_response = router
+            .oneshot(
+                Request::post("/script-query-e2e/_search")
+                    .header("content-type", "application/json")
+                    .body(Body::from(search_body))
+                    .expect("request"),
+            )
+            .await
+            .expect("search response");
+        assert!(
+            search_response.status().is_success(),
+            "search failed: {:?}",
+            search_response.status()
+        );
+
+        let bytes = to_bytes(search_response.into_body(), usize::MAX)
+            .await
+            .expect("response bytes");
+        let response: Value = serde_json::from_slice(&bytes).expect("JSON response");
+        let total = response["hits"]["total"]["value"].as_u64().unwrap_or(0);
+        assert_eq!(
+            total, 1,
+            "script query should match the freshly indexed document via a real search, got: {response}"
+        );
+    }
+}
+
 // ─────────────────────────────────────────────────────────────────────────────
 // Schema conversion helpers
 // ─────────────────────────────────────────────────────────────────────────────
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -19211,7 +19211,7 @@ impl Index {
         // a malformed section bails early and leaves the vec incomplete.
         mut offsets_out: Option<&mut Vec<(u32, u32)>>,
         // Cooperative timeout: when `Some`, the walk polls the wall clock
-        // every 4 096 docs and aborts (setting `*timed_out`) once past the
+        // every 256 docs and aborts (setting `*timed_out`) once past the
         // deadline.  One giant merged segment is otherwise a multi-second
         // uninterruptible unit — the between-segment deadline check in
         // `search_inner` can't bound it.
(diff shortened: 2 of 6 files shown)


xerj-org:
Reviewed for inclusion in **v1.0.0-rc.9**. Holding it. Two independent lenses (correctness and security) both landed on do-not-ship, and the headline reason is not a security one — **the feature currently doesn't match any documents.**

Every finding below was sent to an independent verifier told to refute it; these are the ones that survived.

### Blocker — a top-level `script` query silently returns zero hits on every index

The PR's only routing change is `QueryNode::Script { .. } => ExecutionPlan::MatchAll` in planner.rs:593-595. But `ExecutionPlan` is dead code — `grep -c ExecutionPlan` over index.rs returns 0, and nothing in `xerj-engine` or `xerj-api` references `plan_query`/`ExecutionPlan` at all.

The predicate that actually gates the doc scan is `is_doc_scan_query` (index.rs:23411-23478), a non-exhaustive `matches!` list. `QueryNode::Percolate` is there at :23476; `QueryNode::Script` was never added. So both halves of search skip it:

- **memtable** — not `MatchAll`, not a doc-scan query, `extract_query_text` returns `None` (:23398-23408), `try_doc_values_query` has no Script arm → `MemSnapshot::Empty` (:11365), every unflushed doc invisible
- **segments** — same predicate, same outcome

The new tests pass because they exercise the evaluator arm directly rather than a real `_search`. Adding `QueryNode::Script { .. }` to `is_doc_scan_query` is a one-line fix, but it should come with an end-to-end test that indexes a document and retrieves it through `POST /{index}/_search` with a `script` query — that's what would have caught this.

### Blocker — a 341-byte script query aborts the process

Nothing bounds heap allocation in the interpreter. The four limits that exist (`MAX_SCRIPT_LEN` 64 KiB, `MAX_PARSE_DEPTH` 100, `MAX_EVAL_DEPTH` 500, and no loop grammar) cover source size, stack depth and iteration — not allocation. Assignment to an existing variable parses, and `apply_binary` concatenates strings with an unbounded `format!` (painless.rs:1214-1227), so repeated doubling grows a value exponentially in the number of *statements*. Verified: SIGABRT.

This is pre-existing in the evaluator, but this PR is what makes it reachable from the **query** surface, where cost is also multiplied per candidate document.

### High — the filter fails OPEN, contradicting its own comment

The new arm's comment says anything other than `Ok(Bool(true))` "fails closed the same way" as ES. That holds for a script that fails to parse and for a non-boolean *result* — which are exactly the two cases the new tests cover (index.rs:30975-30997).

The common failures never produce an error at all. `resolve_doc_member` (painless.rs:1330-1345) reads `_source`; for an absent field the fallthrough is `_ => Ok(PainlessValue::Number(0.0))`. Same for a referenced-but-absent `params.*`. So `doc['missing'].value > -1` matches every document instead of none. A filter that silently inverts on a typo'd field name is worse than one that errors.

### Also confirmed

- **No compiled-AST cache** — `eval_painless` re-tokenizes and re-parses per document. A legal 64 KiB script measures 2.03 ms/doc. `scan_stored_section_into` polls its deadline every 4096 docs on the stated assumption that "the parse below costs ~3 µs/doc" (index.rs:19311), so the uninterruptible quantum becomes ~8.3 s instead of ~12 ms — and the prefiltered path has no deadline poll at all.
- The request-time script limit pre-check isn't applied on `_delete_by_query`, `_update_by_query` or `_validate/query`.
- `boost` and `_name` are silently dropped.
- The string shorthand `{"script": {"script": "<source>"}}` is rejected with a 400 that ES accepts.

The `is_doc_scan_query` arm plus an end-to-end search test is the minimum for it to work at all. Beyond that: an allocation ceiling in the evaluator, an AST cache keyed on source so the parse is paid once per query rather than per document, and making missing fields/params an error rather than `0.0` so the filter fails closed as the comment intends.

Genuinely useful feature and the parser/AST work is clean — it's the integration point and the resource bounds that need another pass.

Vinz2168:
Thank you for the review — both blockers verified directly before fixing, not taken on faith:

- **Zero-hits bug**: reproduced live (indexed a doc, searched with a matching `script` query via real `POST /_search`, got 0 hits) before touching anything. Fixed by adding `QueryNode::Script` to `is_doc_scan_query`. Added an end-to-end test that goes through the real HTTP route rather than calling `doc_matches_query` directly — confirmed it fails without the fix and passes with it.
- **SIGABRT via string doubling**: added `MAX_PAINLESS_STRING_LEN` (1 MiB) checked on every `+` concatenation, plus a regression test with 30 doublings that would otherwise reach ~1 GiB.
- **Fail-open on missing fields**: `doc['missing'].value` and any arithmetic/comparison on a non-numeric-coercible value (most commonly a missing `doc`/`params` field) now errors instead of silently acting as `0`, in `apply_binary`, unary `-`, and `resolve_doc_member`. Verified no existing tests relied on the old silent-zero behavior.

Not addressed in this pass (agree these are real, lower severity, tracking as follow-ups): no compiled-AST cache, request-time limit not applied to `_delete_by_query`/`_update_by_query`/`_validate`, `boost`/`_name` dropped, and the `{"script": {"script": "<source>"}}` string-shorthand form isn't accepted yet.

Full suite green: `cargo fmt --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and `cargo test -p xerj-engine -p xerj-query -p xerj-api --lib` (only the 2 pre-existing, unrelated `snapshot_path_security_tests` failures, confirmed present on bare `main` too).

xerj-org:

xerj-org:
Rebased onto current `main` and reconciled with the two things that had accumulated around it: merged PR #88's closure guards, and the review follow-up on this PR. It now merges as one coherent unit.

Three commits: your two, authorship preserved, plus one reconciliation commit.

**The follow-up and this PR contradicted each other on `doc['missing'].value`.** Your commit made it error; the follow-up made it return null. Those cannot both hold, so it was settled against real Elasticsearch rather than by which was newer.

Since ES 7.0, reading `.value` off an empty `ScriptDocValues` throws `IllegalStateException` ("A document doesn't have a value for a field! Use `doc[<field>].size()==0` to check if a document is missing a field!"). ES 6.x returned a type default with a deprecation warning and 7.0 removed that fallback. So returning null matches neither version, and **your fail-closed choice is the correct one** — it is also the only one that keeps a typo'd field name from matching every document, which is the bug this PR set out to fix.

The follow-up's underlying concern was still valid: a guarded script has to be runnable. That is addressed without loosening `.value`, because ES's actual idiom is `doc['x'].size() == 0 ? <default> : doc['x'].value`. `.size()`, `.length` and `.empty` are now total on a missing field, and the ternary only evaluates the branch it takes, so the idiom works. Null-aware `==`/`!=` from the follow-up is kept, since `params` really is a `Map` in Painless and an unsupplied key really does read as null.

**The `exec_body` conflict turned out not to be a conflict.** The follow-up appeared to restructure the statement loop, but its base predates #88's extraction of `exec_body`, so that hunk was just the old inline body carried along as diff context. The only real change in it was `&stmts` -> `stmts` for the memoised AST. Main's shape is kept, and #88's `CallGuard`, `MAX_CALL_DEPTH` and `MAX_CALL_COUNT` are untouched and still trip closed.

**Also corrected before merge:** the comment and commit message claimed executing a max-size script is "~0.2 ms/doc" and that the 256-doc poll keeps the worst case "in the tens of milliseconds". Measured on a release build it is 385 us/doc for a benign 64 KiB script (~98 ms/poll) and **1.27 s/doc for a deliberately adversarial one (~325 s/poll)**. The tighter poll interval is a real improvement but not a bound, because nothing imposes a per-evaluation CPU budget. Both now say that rather than implying the gap is closed.

Verified after the rebase: `xerj-engine` 313 passed, `xerj-query` 166 passed, 0 failed; release build of engine, query and api green. Painless tests went 43 -> 54 with none removed.

Thanks for this one — the script-query support is a real gap closed, and the fail-closed instinct on missing fields was right.

PR #89, opened 2026-07-31
fix(query): resolve named ES date formats for ignore_malformed validation

Vinz2168:
## Summary
`is_date_value_valid_with_format` — the ingest-time check that decides whether a `date`/`date_nanos` field gets silently dropped under `ignore_malformed` — had two bugs that happened to compensate for each other on OpenSearch's UBI sample dashboards, which is why the breakage was easy to miss:

- **Named ES built-in formats weren't resolved.** `strict_date_time`, `date_time`, etc. are named shorthands ES expands internally to a real pattern (`strict_date_time` → `yyyy-MM-dd'T'HH:mm:ss.SSSXX`). The old code matched the format string as literal text via ad-hoc chrono-strftime substitution, never expanding named formats first — so a value that's genuinely valid under the format (e.g. `"2024-01-01T00:00:00.000Z"` under `strict_date_time`) could never match and was silently dropped every time.
- **A JSON number was accepted as valid for *any* date format**, including non-epoch ones — so a bare epoch-millis integer was wrongly accepted even under a format with no `epoch_millis`/`epoch_second` in its list.

**Confirmed live against a real OpenSearch 2.11.1 node**: recreated `opensearch_dashboards_sample_ubi_events`'s exact mapping (`{"type":"date","format":"strict_date_time","ignore_malformed":true}`) and reindexed its exact documents. Real OpenSearch does the *opposite* of what xerj was doing — it accepts the ISO strings and rejects the bare numbers. Verified the fix reproduces this exactly with a clean minimal repro (scalar ISO string, array-wrapped ISO string, bare epoch-millis number).

## Fix
Delegates to `xerj_query::dates::compile_formats` + a new `date_value_matches_formats`, reusing the same format compiler/parser already used (and already tested) for range-query date bounds, instead of maintaining a second, incomplete parallel implementation. Removes the now-dead `es_date_format_to_strftime`.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo test -p xerj-query -p xerj-api --lib` (235 passed, including 6 new `dates.rs` tests and 4 new `es_compat.rs` tests — one of which is an end-to-end reproduction of the exact UBI mapping)
- [x] Live repro against a real OpenSearch 2.11.1 node (identical mapping + documents) confirming xerj's fixed behavior now matches
- [x] Live repro directly against the built binary via HTTP (scalar/array ISO string accepted, bare epoch-millis number correctly ignored)

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -27338,109 +27338,106 @@ fn filter_keyword_length(val: &Value, max: usize) -> (Value, bool) {
     }
 }
 
-/// Translate an ES date-format pattern to a chrono strftime string and
-/// check whether `v` parses under it.
+/// Check whether `v` parses under the declared ES date `fmt`, delegating to
+/// the same format compiler/parser `xerj-query` uses for range-query date
+/// bounds (`compile_formats` resolves named formats like `strict_date_time`
+/// to their real pattern; `date_value_matches_formats` only accepts a bare
+/// number for an epoch format). An unparseable `fmt` string should already
+/// have been rejected when the mapping was created, so don't newly reject
+/// ingest over it here — treat it as permissive instead.
 fn is_date_value_valid_with_format(v: &Value, fmt: &str) -> bool {
-    // Accept epoch numbers for epoch_millis / epoch_second formats.
-    if fmt.contains("epoch_millis") {
-        return matches!(v, Value::Number(_))
-            || matches!(v, Value::String(s) if s.parse::<i64>().is_ok());
+    match xerj_query::dates::compile_formats(fmt) {
+        Ok(compiled) => xerj_query::dates::date_value_matches_formats(v, &compiled),
+        Err(_) => true,
     }
-    if fmt.contains("epoch_second") {
-        return matches!(v, Value::Number(_))
-            || matches!(v, Value::String(s) if s.parse::<i64>().is_ok());
-    }
-    let Some(s) = v.as_str() else {
-        return matches!(v, Value::Number(_));
-    };
-    // Split combined format ("||") — ES allows multiple fallback formats.
-    for single_fmt in fmt.split("||") {
-        let pat = es_date_format_to_strftime(single_fmt.trim());
-        // ES validates declared date formats STRICTLY: the input must match
-        // the pattern exactly (e.g. `dd-MM-yyyy` requires a 4-digit year, so
-        // "19-12-90" is malformed). chrono's parsers are lenient about field
-        // widths, so we additionally require the parsed value to round-trip
-        // back to the original string under the same pattern.
-        if let Ok(dt) = chrono::NaiveDateTime::parse_from_str(s, &pat) {
-            if dt.format(&pat).to_string() == s {
-                return true;
-            }
-        }
-        if let Ok(d) = chrono::NaiveDate::parse_from_str(s, &pat) {
-            if d.format(&pat).to_string() == s {
-                return true;
-            }
-        }
-        // Timezone-aware formats: accept a successful parse (round-tripping
-        // `%z`/`%:z` offset spellings is unreliable, so don't require it).
-        if chrono::DateTime::parse_from_str(s, &pat).is_ok() {
-            return true;
-        }
-    }
-    false
 }
 
-/// Map common ES date-pattern tokens (yyyy/MM/dd/HH/mm/ss/SSS/Z) to chrono
-/// strftime directives. Preserves literal chars between tokens.
-fn es_date_format_to_strftime(es_fmt: &str) -> String {
-    let mut out = String::with_capacity(es_fmt.len() + 8);
-    let bytes = es_fmt.as_bytes();
-    let mut i = 0;
-    while i < bytes.len() {
-        // Longest-match substitution for tokens we care about.
-        let rest = &es_fmt[i..];
-        if rest.starts_with("yyyy") {
-            out.push_str("%Y");
-            i += 4;
-        } else if rest.starts_with("yy") {
-            out.push_str("%y");
-            i += 2;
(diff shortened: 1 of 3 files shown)


xerj-org:
Reviewed for inclusion in **v1.0.0-rc.9**. Holding it — the named-format resolution is a real improvement, but as written it **silently drops documents** that are currently indexed, and unknown formats fail open.

Every finding was sent to an independent verifier told to refute it; these survived.

### Blocker — numeric values under numeric-shaped date formats are now silently dropped

The new `Value::Number` arm (dates.rs:183-188) accepts a JSON number only when some compiled format is `EpochMillis` or `EpochSecond`. The old code returned `true` for any `Value::Number` under any non-epoch format.

ES does neither. `DateFieldMapper` reads the value with `parser().text()`, which stringifies a `VALUE_NUMBER` token, then parses that text with the declared formatter. So under `basic_date` (`yyyyMMdd`), the JSON number `20240101` renders as `"20240101"` and parses cleanly.

Verified by running both implementations over the same matrix: `basic_date` + `20240101` is **accepted before this PR, rejected after**. With `ignore_malformed: true` that value is now dropped from `_source` and recorded in `_ignored` instead of being indexed — silently, on documents that indexed fine yesterday. The same applies to any numeric-shaped pattern (`basic_date`, `basic_date_time_no_millis`, `yyyyMMddHHmmss`, …).

### High — unknown named formats fail OPEN and accept arbitrary garbage

`compile_one_format` handles ~20 named formats and falls through to `other => other`, compiling the *name itself* as a Java pattern. Many real ES format names contain letters that aren't in `VALID_JAVA_PATTERN_LETTERS` — `b`, `o`, `t`, `i`, `r` — so they return `Err(UnknownPatternLetter)`, and `is_date_value_valid_with_format` swallows that into `true` (es_compat.rs:26741-26744).

Measured, all of these currently accept literally any value: `basic_time`, `basic_ordinal_date`, `basic_week_date` (`b`), `ordinal_date`, `hour_minute` (`o`), `time`, `week_date` (`t`).

Failing open is the wrong default here: an unsupported format name should be a mapping-time error, or at minimum should fall back to the previous permissive-but-narrow behaviour rather than disabling validation entirely for that field.

### Also confirmed

- **The validator now contradicts the `ignore_malformed: false` rejection path** — the stricter setting ends up *more* permissive than the lenient one, which is backwards.
- **The format string is recompiled per value on the ingest hot path**, amplifying an unvalidated user-supplied input by roughly 4× on a hot path — worth a compiled-format cache keyed on the mapping.
- Leading/trailing whitespace is now silently trimmed, accepting values ES treats as malformed.
- `epoch_second` overflow checking depends on JSON spelling — the number and the string form of the same value disagree.
- `null` in a date field no longer records `_ignored`, a behaviour change on existing indices.

Stringify numbers and run them through the declared formatter the way `DateFieldMapper` does, rather than gating on epoch formats; make an unresolvable format name fail closed (or error at mapping time); and add fixtures for the numeric-under-`basic_date` case and for at least one `b`/`o`/`t` named format, since neither is covered today.

The direction is right — resolving named formats properly is exactly what this needed, and the `dates.rs` extraction is a good place for it. It's the two failure directions that need inverting.

Vinz2168:
Thanks — the blocker's mechanism (`DateFieldMapper` stringifying numbers via `text()`) checked out.

- **Numeric-shaped format blocker**: replaced the epoch-only `Value::Number` branch with `n.to_string()` run through the exact same per-format parser a `String` value uses. `basic_date` + `20240101` (and any other numeric-shaped pattern) is now accepted; `strict_date_time` + a bare epoch-millis number is still correctly rejected (its stringified form doesn't match the structured `yyyy-MM-dd'T'...` pattern, same as before, just via the unified path now).
- **Fail-open on unresolvable formats**: `is_date_value_valid_with_format` now returns `false` on a `compile_formats` error instead of `true`. Went with "fail closed at validation time" rather than a mapping-time error, per your "at minimum" — a format this engine can't resolve now makes the field's values ignored (visible in `_ignored`) rather than silently unvalidated.
- **Whitespace trimming**: removed — my fix had added a `.trim()` on the string path that ES doesn't do; a padded value that only matches after trimming is now correctly rejected.
- **`null` handling**: checked this deliberately rather than changing it — ES treats JSON `null` as "no value" uniformly across field types (not malformed), so `null` continuing to always match (and not appear in `_ignored`) is correct, not a regression.

Not addressed (agreed, tracking separately): the compiled format isn't cached per mapping (still recompiled per value), and `ignore_malformed:false`'s hard-reject path in `bulk.rs` uses a different, more lenient check than this one — so the stricter setting can currently be more permissive than the lenient one for the same value. That inconsistency needs the two paths unified, which felt like a separate change from this fix.

xerj-org:

xerj-org:
Rebased onto current `main` and folded the review follow-up into the branch, so this now merges as one complete unit rather than landing a known regression and fixing it afterwards.

The branch is now three commits:

1. `resolve named ES date formats for ignore_malformed validation` (yours)
2. `stringify numbers and fail closed on unresolvable date formats` (yours, addressing the first review round)
3. `implement the real ES named formats instead of dropping their data`

Commit 3 exists because "fail closed on unresolvable formats" from round 2 was the right call for safety but had a cost worth removing: roughly seven named formats this engine could not resolve (`basic_time`, `basic_ordinal_date`, `basic_week_date`, `ordinal_date`, `hour_minute`, `time`, `week_date`) went from accepting anything to accepting nothing, which silently drops values on fields that indexed fine before. Implementing the formats properly removes the choice between failing open and dropping data.

Verified after the rebase: `cargo build --release -p xerj-query -p xerj-api`, then `cargo test --release -p xerj-query -p xerj-api` → **284 passed, 0 failed**.

Thanks for working through two rounds of review on this one. The named-format resolution was the right direction and the `dates.rs` extraction is what made commit 3 straightforward.

PR #91, opened 2026-08-01
fix(engine): parse_date_ms accepts no-colon numeric zone offsets

Vinz2168:
## Summary
`parse_date_ms` is the single, shared date-to-epoch-ms conversion used throughout `xerj-engine` — aggregations (min/max, date_histogram, terms script buckets), range queries, and sorting all go through it (verified: `index.rs` has no separate date parser, every call site delegates to this one function).

It only tried `chrono::DateTime::parse_from_rfc3339`, which requires a colon in the zone offset per the RFC3339 spec (`+00:00`) — but ES's own `strict_date_time`/`XX`-style date formats also accept the no-colon numeric form (`+0000`), which real clients emit (e.g. Java's default date formatters). No other branch in the function handled it either, so it fell through to the final "parse as raw integer" fallback and returned `None`.

## How this was found
While doing an end-to-end test of the UBI sample dashboards (unrelated to this fix — see #86/#87/#88/#89), one dashboard panel ("Most Common Search Result") showed "No results found". Using OSD's own "Inspect → Requests" panel to see the exact query, then tracing it: a `min`/`max` aggregation over `timestamp` for the subset of documents with `"timestamp": "2025-01-24T07:31:52.102+0000"`-shaped values returned `null` — the field was validly mapped (`strict_date_time`), successfully ingested (not `_ignored`), sitting right there in `_source`, but invisible to every date-based query or aggregation.

**Important caveat, so the fix isn't over-claimed**: fixing the parser does NOT make that specific panel populate — verified live before and after. The documents carrying this timestamp shape are genuinely dated January–March 2025, while the dashboard's saved view has a *fixed* December-2024-only time range (`timeRestore: true`). Once the panel's time filter is widened to include that range, the aggregation returns real buckets (confirmed: `B013UFPODY: 20, B07XLFYN6S: 20, ...`) — so the panel's emptiness within its saved December-only window is correct behavior, matching what real ES would also do. What the fix actually corrects: before it, `min`/`max` on the full index silently reported `2024-12-15` as the latest timestamp; after it, it correctly reports `2025-03-03` — i.e. every date-aware read path was silently blind to any document using this (valid, common) offset spelling, regardless of dashboard time filters.

## Fix
Added `%Y-%m-%dT%H:%M:%S%.f%z` (chrono's RFC 2822-style numeric-offset parser) as a fallback tried right after the RFC3339 attempt.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy -p xerj-engine -p xerj-query -p xerj-api --all-targets -- -D warnings`
- [x] `cargo test -p xerj-engine -p xerj-query -p xerj-api --lib` (356 passed; only the 2 pre-existing, unrelated `snapshot_path_security_tests` failures)
- [x] 5 new regression tests, confirmed to fail without the fix (temporarily reverted, reran, restored)
- [x] Verified live: rebuilt with the fix, re-ran `min`/`max` over the real live index before/after — `max_ts` went from `2024-12-15` (silently truncated) to the correct `2025-03-03`

### DIFF
--- a/engine/crates/xerj-engine/src/aggs.rs
+++ b/engine/crates/xerj-engine/src/aggs.rs
@@ -3969,6 +3969,17 @@ pub(crate) fn parse_date_ms(val: &Value) -> Option<i64> {
             if let Ok(dt) = chrono::DateTime::parse_from_rfc3339(s) {
                 return Some(dt.timestamp_millis());
             }
+            // RFC3339 requires a colon in the zone offset (`+00:00`); ES's
+            // own `strict_date_time`/`XX`-style formats also accept the
+            // no-colon numeric form (`+0000`), which real clients emit —
+            // e.g. Java's default `OffsetDateTime.toString()`-adjacent
+            // formatters. Without this, a validly-mapped, successfully
+            // ingested date like "2025-01-24T07:31:52.102+0000" silently
+            // drops out of every date aggregation/range query even though
+            // it's neither malformed nor `_ignored`.
+            if let Ok(dt) = chrono::DateTime::parse_from_str(s, "%Y-%m-%dT%H:%M:%S%.f%z") {
+                return Some(dt.timestamp_millis());
+            }
             // Try ISO-8601 without timezone (treat as UTC).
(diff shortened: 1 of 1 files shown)


xerj-org:

PR #93, opened 2026-08-01
legal: add CLA agreement and cla-bot configuration

xerj-org:
Configures the cla-bot installed on this repo, and adds the agreement it points at.

## Why

Contributions currently arrive with no explicit licence grant beyond Apache-2.0 §5's inbound=outbound clause, and no commit in the tree carries a `Signed-off-by`. For a project with a commercial surface that is thin coverage — §5 gives you a licence to *use* the contribution under the same terms, but no explicit patent grant, which is the part that matters if XERJ is ever relicensed or offered commercially.

## What

- **`CLA.md`** — adapted from the [Apache Software Foundation Individual CLA v2.0](https://www.apache.org/licenses/icla.pdf), the standard instrument for Apache-2.0 projects: copyright licence, patent licence with the usual litigation-termination clause, contributor representations, third-party-work carve-out. Adds a short **generative-AI clause** making explicit that using an assistant does not relieve the contributor of the representations — increasingly relevant and cheap to state now.
- **`.clabot`** — points `contributors` at `.contributors` via the GitHub contents API, so signing is a pull request from the signer's own account and the commit history is the durable record. Custom message links the agreement and the one-line signing step.
- **`.contributors`** — seeded with `xerj-org` only.
- **`.gitignore`** — allowlists `CLA.md`, which the blanket `*.md` rule hid.

## Two things the maintainer must decide

1. **`CLA.md` is unreviewed by counsel** and carries a visible note saying so. Two placeholders must be filled before it is authoritative: the legal entity named as **the Project Owner**, and a contact address in "Notifying us".

2. **Nobody but `xerj-org` is listed in `.contributors`, deliberately.** That file asserts a person *has signed*. That is not an assertion this repo may make on someone's behalf, so the existing contributors on open PRs (#85–#89, #91) will need to sign before those merge. Each signs once, not per PR.

Merging this activates the check on all open pull requests.


xerj-org:
Reviewed this today because four external PRs are in flight (#156, #160, #162, #167/#168 from two contributors) and leaving it unexamined has a real cost. CI is 9/9 green and it merges cleanly against `main`, so nothing mechanical is in the way. I am **not merging it**, for three specific reasons — two substantive, one that makes the whole thing inert.

**1. The bot this configures is not installed, so merging changes nothing.** `.clabot` is configuration for the hosted cla-bot GitHub App. I checked the head commit of every open PR (#168, #167, #166, #163) and there is no CLA check run and no commit status from cla-bot on any of them, and `grep -rn 'cla-bot\|clabot' .github/` finds no workflow wiring either. So as it stands this PR adds a config file that nothing reads. The gating this is supposed to provide would still not exist the moment after merge. Either install the App on `xerj-org/xerj` (and confirm it posts a check on a test PR), or replace the hosted-bot approach with a workflow in `.github/workflows/` that does the same job with `GITHUB_TOKEN`.

**2. `CLA.md` grants licences to a party it never defines.** Sections 2 and 3 grant copyright and patent licences to "the Project Owner", and section 1 Definitions defines **You**, **Contribution** and **Submitted** — but not **Project Owner**. No legal entity is named anywhere in the document. A grant to an undefined grantee is the one defect that would matter most if XERJ is ever relicensed or offered commercially, which is the scenario the document's own header calls out.

**3. The contact placeholder is unfilled.** Section 7 obliges the signer to notify us of facts that would make their representations inaccurate, then gives the address as `**[MAINTAINER CONTACT — fill in]**`.

The file carries its own maintainer note saying it has not been reviewed by a lawyer and that two placeholders must be filled before it is authoritative. That note is honest and correct, and I am taking it at its word rather than merging past it. Asking four external contributors to sign an instrument that names no grantee and no contact would be worse than having no CLA — and the `.clabot` message text ("we need a signed Contributor License Agreement from…") would be making a claim the repository cannot currently back.

**To unblock, in order of who has to do it:**

- Name the legal entity that is the Project Owner, as a definition in section 1, and fill in the section 7 contact. Both are owner decisions, not something I should guess at.
- Install the cla-bot App (or swap in a workflow) and verify a check actually appears on a test PR.
- Get counsel to confirm the text grants what you need, then delete the maintainer note at the top of `CLA.md` — while that note is present, the document says of itself that it should not be relied upon.

The Apache ICLA v2.0 adaptation underneath is a sound choice and the "signature is a PR from your own account adding your username to `.contributors`" mechanism is a good, low-friction design. The structure is right; it is the three blanks and the missing installation that stop it. Worth noting the timing cost too: signatures collected under an instrument that is later materially corrected may need re-collecting, so it is better to fill these in before the first external contributor signs than after.

PR #95, opened 2026-08-01
fix(engine): bucket_script resolves TSVB filter_ratio's _count + null/ternary script

Vinz2168:
## Summary
Fixes `bucket_script` so it correctly resolves the exact pattern Kibana/OpenSearch Dashboards' TSVB `filter_ratio` metric compiles to — found live while testing the sample "[Flights] Global Flight Dashboard": the "Delays & Cancellations" panel (a `filter_ratio` line chart) rendered as a flat, empty 0% line for every time bucket.

Two independent bugs in the same function, both required to fix the panel:

**1. `_count` buckets_path segment not resolved (`resolve_bucket_script`)**

TSVB's `filter_ratio` generates a `filter` sub-agg named "numerator" and another named "denominator", then a `bucket_script` referencing each one's doc count via `"numerator>_count"` / `"denominator>_count"`. `resolve_bucket_script`'s lookup only checked for a field literally named `_count` with a nested `.value` (`sib.get("_count").and_then(|x| x.get("value"))`) — but a `filter` (singular) aggregation's bucket is shaped `{"doc_count": N}`, not `{"_count": {"value": N}}`. The special `_count` → `doc_count` mapping already existed elsewhere in this file (`get_bucket_metric_value`, `extract_bucket_value_opt`) but was never applied here, so any `bucket_script` referencing a sibling `filter`'s count always resolved to `null`.

**2. Script evaluator had no `null` literal or ternary support**

The real script TSVB sends is:
```
params.numerator != null && params.denominator != null && params.denominator > 0
  ? params.numerator / params.denominator : 0
```
`eval_script_expr`'s tokenizer treated the bare word `null` as a `params.<name>` lookup (always failing — no such param — aborting the whole parse), and had no `?`/`:` support at all. So this exact script could never evaluate, independent of fix #1.

## How this was found
Confirmed via a temporary debug log of the real request body xerj received from a live OSD instance (removed before this PR) — the panel's actual query didn't match my first, simpler manual reproduction, which is what surfaced bug #2.

## Fix
- `resolve_bucket_script`: added the same `_count` → `doc_count` special case (+ `.as_f64()` fallback for plain-numeric siblings) already used elsewhere in this file.
- `eval_script_expr`: added a `null` literal token (meaningful only in `==`/`!=`, matching how TSVB actually uses it), and ternary `cond ? a : b` via a pre-pass that splits the token stream on the first top-level `?`/matching `:` and recurses — rather than folding it into the operator-precedence shunting-yard, which doesn't model ternary well as a plain binary op.

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy -p xerj-engine -p xerj-api --all-targets -- -D warnings`
- [x] `cargo test -p xerj-engine --lib` (287/289 — the 2 failures are the pre-existing, unrelated `snapshot_path_security_tests` failures also present on `upstream/main`)
- [x] 3 new regression tests: `bucket_script_resolves_sibling_filter_doc_count_via_underscore_count`, `bucket_script_supports_null_literal_and_ternary_like_tsvb_filter_ratio`, `bucket_script_ternary_false_branch_avoids_division_by_zero` (denominator=0 guard)
- [x] Live verification: rebuilt a combined test binary with this fix, redeployed against the real Kibana sample Flights dataset — `curl` against the exact captured live request now returns real ratios (was all-null before) for all 27 buckets, and the "[Flights] Delays & Cancellations" dashboard panel renders a populated area chart instead of a flat empty 0% line.

### DIFF
--- a/engine/crates/xerj-engine/src/aggs.rs
+++ b/engine/crates/xerj-engine/src/aggs.rs
@@ -1007,9 +1007,15 @@ fn resolve_bucket_script(
             None => return json!({"value": Value::Null}),
         };
         let v = if let Some(m) = metric {
-            sib.get(m)
-                .and_then(|x| x.get("value"))
-                .and_then(Value::as_f64)
+            if m == "_count" {
+                sib.get("doc_count").and_then(Value::as_f64)
+            } else {
+                sib.get(m).and_then(|x| {
+                    x.get("value")
+                        .and_then(Value::as_f64)
+                        .or_else(|| x.as_f64())
+                })
+            }
         } else {
             sib.get("value").and_then(Value::as_f64)
         };
@@ -12220,6 +12226,40 @@ mod tests {
         let agg = json!({ "uniq": { "cardinality": { "field": "tag" } } });
         let result = run_aggs(&agg, &docs());
         assert_eq!(result["uniq"]["value"], 2);
+    }
+
+    // Regression: `bucket_script` referencing a sibling `filter` agg's doc
+    // count via the special `_count` buckets_path segment (e.g.
+    // `"numerator": "numerator>_count"`) always resolved to null, because
+    // the lookup only checked for a field literally named `_count` with a
+    // nested `.value` — which doesn't exist on a `filter` agg's bucket
+    // (`{"doc_count": N}`). This is exactly the shape Kibana/OSD's TSVB
+    // `filter_ratio` metric compiles to, so every `filter_ratio` panel
+    // (e.g. the sample Flights dashboard's "Delays & Cancellations")
+    // silently rendered as all-null.
+    #[test]
+    fn bucket_script_resolves_sibling_filter_doc_count_via_underscore_count() {
+        let docs = vec![
+            json!({"status": "delayed"}),
+            json!({"status": "delayed"}),
+            json!({"status": "ok"}),
+            json!({"status": "ok"}),
+        ];
+        let agg = json!({
+            "numerator": { "filter": { "term": { "status": "delayed" } } },
+            "denominator": { "filter": { "match_all": {} } },
+            "ratio": {
+                "bucket_script": {
+                    "buckets_path": { "numerator": "numerator>_count", "denominator": "denominator>_count" },
+                    "script": "params.numerator / params.denominator"
+                }
+            }
+        });
+        let result = run_aggs(&agg, &docs);
+        let ratio = result["ratio"]["value"]
+            .as_f64()
+            .expect("bucket_script must resolve _count buckets_path, not stay null");
+        assert!((ratio - 0.5).abs() < 1e-9);
     }
 
     #[test]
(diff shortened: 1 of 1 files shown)


xerj-org:
Reviewed for merge. **Holding this one**, because the review found a way to kill the process from a single request. The `bucket_script` fix itself looks right; the problem is in the expression evaluator it leans on.

## Blocker: unauthenticated remote process kill

`eval_tokens` (`aggs.rs:1046`) recurses once per nested ternary with **no depth cap**, and `/_msearch` does not apply the request-time script guard that the single-search path applies. A single ~240 KB `/_msearch` request is enough to exhaust the native stack, and with `panic = "abort"` in the release profile that is a process abort rather than an error response.

This is the same shape the project has already shipped and fixed three times: nested SQL `WHERE` parens, nested `query_string` parens, and the Painless closure recursion in #88. Every new recursive construct over user-supplied input needs a depth bound, and every entry point that accepts scripts needs the same guard.

Two things to fix:

1. **Bound the recursion in `eval_tokens`.** A `MAX_*_DEPTH` constant checked on entry, returning a bounded error, in the style of `MAX_SQL_DEPTH` and `MAX_EVAL_DEPTH` elsewhere in the tree.
2. **Apply the request-time script guard on `/_msearch`.** Right now it is a hole around whatever the single-search path validates, which makes it a general bypass and not only an issue for this PR.

## Also

The diff deletes 78 lines in `aggs.rs`, which is enough surface to change existing `bucket_script` results silently. Worth a note in the PR body on what behaviour is intentionally different for existing users, particularly around `gap_policy` handling and what happens when a referenced path is missing.

## Merge mechanics

The commits carry `Co-Authored-By: Claude Sonnet 5` and a `Claude-Session:` trailer. This repo does not allow those on `main`, so when this does land it needs a squash with a rewritten message. Same applies to #92 and #87, so it is a general note rather than a comment on this PR.

Genuinely useful fix and a real Kibana/TSVB compatibility gap. It is the evaluator's missing bound that needs closing first.

xerj-org:
Closing as **already merged**, not as rejected.

#105's branch was stacked directly on this one — `git merge-base --is-ancestor` confirms this PR's head is an ancestor of #105's, and both your commits (`d3ad618`, `34bbf91`) appear verbatim in the merge. So merging #105 delivered 100% of this PR, and merging this separately would be a no-op that re-applies the same changes.

Landed as 5df22b7 on `main`, with your authorship preserved on both commits.

The `_count` resolution and the null/ternary support were the right calls, and TSVB's `filter_ratio` was a real Kibana compatibility gap. Three findings from review were closed on top of your work before it landed:

- `_count` now agrees with `_doc_count` on every aggregation type, not just `terms`
- parenthesised ternaries evaluate, where `find_ternary_split` had only looked at paren depth 0
- two equal infinities compare equal again (`inf - inf` is NaN, so an epsilon-only `==` reported identical values as unequal)

The expression-evaluator depth bound from #111 landed first, so the recursion `eval_tokens` introduces is bounded before it can be reached — that was the blocker on both this and #105.

Thanks for both of these.

PR #102, opened 2026-08-01
fix(engine): query_string with no default_field searches every text field

Vinz2168:
## Summary
`query_string` with no explicit `default_field` used to search only `text_fields[0]` — one arbitrary field, picked by schema declaration order — instead of ES/OpenSearch's actual default (`index.query.default_field` defaults to `"*"`, i.e. every mapped text field). Any match in every other text field was silently missed.

## How this was found
Root-causing an OpenSearch Dashboards Timeline (Timelion) panel that rendered empty. Timelion's `.opensearch(index, ...)` datasource selects which index's documents to count using a `filters` aggregation with a field-less `query_string` matching the literal index name — against a document field OSD's own sample data seeds with that exact value. Reproduced directly with `curl`: on an index whose schema didn't happen to put that field first, the same query matched **18 documents instead of the true ~14,074** — off by two orders of magnitude, with no error, no warning, just quietly wrong.

## Fix
When no `default_field` is given, the projection now builds the token-OR across **every** entry in `text_fields`, not just the first. An explicit `default_field` is unchanged — still single-field behavior, including the keyword whole-value-term shortcut for exact-typed fields. The genuine no-text-fields-mapped edge case still falls back to the `_all` sentinel, same as before.

## Note
This fix alone does not fully resolve the Timeline panel — a second, deeper bug was also found in this investigation: the same `filters`-aggregation-with-`query_string` pattern, combined with an *outer* query filter, drops to 0 matches specifically when the search spans **multiple indices** (`_all`) — even though `hits.total` is computed correctly across those same indices. That one is still open; will follow up with a separate PR once root-caused (it lives in cross-index aggregation dispatch, not this projection). Also related: #101 (a real-but-insufficient-alone fix for the same investigation).

## Test plan
- [x] `cargo fmt --check`
- [x] `cargo clippy -p xerj-engine --all-targets -- -D warnings`
- [x] `cargo test -p xerj-engine --lib` (275/277 — the 2 failures are the pre-existing, unrelated `snapshot_path_security_tests` failures also present on `upstream/main`)
- [x] New regression test: `query_string_with_no_default_field_searches_every_text_field`
- [x] Live verification: the minimal repro (`filters` agg + field-less `query_string`, single index) went from undercounting to matching every mapped text field.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -28191,14 +28191,40 @@ fn query_node_to_fts(
         } => {
             // Honor top-level `query_string.boost` like the Match arm above.
             let b = boost.unwrap_or(1.0);
-            let field = default_field
-                .as_deref()
-                .or_else(|| text_fields.first().map(|s| s.as_str()))
-                .unwrap_or("_all");
-            // Keyword default field: whole-value term (keyword analyzer).
-            if exact_fields.contains(field) {
-                return Some(FtsQuery::Term(FtsTerm::boosted(field, query.as_str(), b)));
+            // An explicit `default_field` searches exactly that one field —
+            // unchanged single-field behavior, including the keyword
+            // whole-value-term shortcut.
+            if let Some(field) = default_field.as_deref() {
+                if exact_fields.contains(field) {
+                    return Some(FtsQuery::Term(FtsTerm::boosted(field, query.as_str(), b)));
+                }
+                let registry = AnalyzerRegistry::default();
+                let analyzer = registry.get_analyzer("standard")?;
+                let tokens = analyzer.analyze(query);
+                if tokens.is_empty() {
+                    return None;
+                }
+                let mut bool_q = FtsBool::new().boost(b);
+                for token in &tokens {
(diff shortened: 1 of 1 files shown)


xerj-org:
Holding this one on a blocker plus a correctness gap.

**Blocker: uncapped tokens x fields cross-product in the search hot path**, uninterruptible by the request deadline. A query with many tokens against many fields multiplies out with no ceiling and no deadline poll, so a single request can occupy a worker indefinitely. This is the `more_like_this` cross-product shape that was already fixed once in this repo. It needs a cap on the product and a deadline check inside the loop.

**High: the fix does not apply to heterogeneous segments**, and there it silently drops documents that previously matched. Silent under-matching on a search path is worse than the original bug, because nothing surfaces it.

The default-field behaviour you are fixing is a real ES-compat gap and worth having. It is the unbounded expansion that has to be closed first.

xerj-org:
Superseded by #130, which contains your commit (authorship preserved) plus the review fixes — the uncapped cross-product deadline, the heterogeneous-segment document drop, a 66x zero-token hot-path regression, and a max_score correction. Merged to main. Closing as included; thank you for the original fix.

PR #105, opened 2026-08-01
fix(engine): bucket_script resolves a bare-string `_count` buckets_path

Vinz2168:
## Summary

**Stacked on #95** (needs the `_count`/null/ternary script-evaluator work from that PR; this branch is based on it, not `main`).

`resolve_bucket_script` required `buckets_path.as_object()` unconditionally, returning null immediately for ES's "simple form" — `buckets_path` as a bare **string** rather than an alias map. In that form ES binds the resolved value to the single script variable `_value`, and the bare string `"_count"` (no `sibling>` prefix at all) refers to the **containing bucket's own** `doc_count` — distinct from #95's fix, which only handled `"alias": "sibling>_count"` (a *named sibling's* count, inside the object form).

This is exactly what Kibana/OpenSearch Dashboards' Timelion `.opensearch()` datasource emits for its per-time-bucket count metric whenever a `split` is present:

```json
"count": {
  "bucket_script": {
    "buckets_path": "_count",
    "script": {"source": "_value", "lang": "expression"}
  }
}
```

nested inside each `date_histogram` bucket, itself nested inside a `terms` split, itself nested inside the `filters` index-selection wrapper.

## Found live

Root-causing `[Logs] Chart and Visualization demo`'s "(Timeline) Stacked extensions over time" panel: real per-bucket `doc_count` present throughout the response, but every `count.value` null — the chart rendered as an empty flat line at 0 despite its TSVB sibling panel showing the correct stacked series for the same field.

## Fix

Two parts, because fixing `resolve_bucket_script` alone wasn't sufficient:

1. **`resolve_bucket_script`**: added a string-form branch — binds the resolved value to `_value`, with `"_count"` resolving to the current bucket via `siblings.get("doc_count")`, and any other bare string resolving via the same sibling-lookup logic the object form already uses (`sibling` or `sibling>_count`/`sibling>metric`).

2. **`run_aggs_with_all`**: the brute-force per-bucket resolver only ever calls `resolve_sibling_pipelines` on the **top-level** aggs result map, which has no `doc_count` key. A `terms`/`date_histogram` bucket's sub-aggs are evaluated by a *recursive* `run_aggs_with_all(sub, bucket_docs, all_docs)` call whose returned map never includes the containing bucket's own `doc_count` (that gets merged in separately by the caller building the bucket wrapper) — so a nested bucket_script had no way to see it, at any nesting depth. Fixed by staging the current doc slice's own count (`docs.len()`) under a transient `doc_count` key immediately before `resolve_sibling_pipelines` runs, then stripping it back out — it isn't a real sibling agg and must not leak into the returned aggs tree. Guarded against the unlikely case of a user-defined agg literally named `doc_count`, which wins over the synthetic value.

`fast_aggs.rs`'s own `resolve_sibling_pipelines` call site was already unaffected — it passes the full bucket map, `doc_count` included, by construction.

## A note on PR #103

While root-causing this, I found that PR #103 (multi-index keyed bucket-agg merge) had never actually been included in the combined test build I was verifying fixes against this session — an integration-branch bookkeeping gap on my end, not a code issue. The "(Timeline) Avg bytes over time" panel I'd earlier reported as fixed and verified had, in fact, only been rendering correctly by coincidence of index-iteration order (whichever index xerj happened to process first for that particular query happened to be the right one). With #103 actually included, both panels are now confirmed working for the right reason rather than by luck — flagging this for the historical record since it affects how much weight to put on earlier "verified live" claims from panels sharing this multi-index `_all` + `filters` shape.

## Test plan
- [x] `cargo fmt --check -p xerj-engine`
- [x] `cargo clippy -p xerj-engine --all-targets -- -D warnings`
- [x] `cargo test -p xerj-engine --lib` (296/296; 2 pre-existing unrelated failures confirmed present on unmodified code — macOS `/tmp` symlink path resolution in `snapshot_path_security_tests`)
- [x] New regression test `bucket_script_resolves_bare_string_underscore_count_buckets_path`, nested inside a `terms` bucket to match the real shape (bare `_count` only means anything inside a containing bucket)
- [x] Live verification: rebuilt a combined binary with this fix + #95/#99/#101/#102/#103/#104, redeployed against the real long-running test environment. `curl` against the exact captured live Timelion query: every `time_buckets` entry's `count.value` now mirrors its `doc_count` (previously null throughout). Confirmed in the browser: "(Timeline) Stacked extensions over time" — flat empty line at 0 since before this investigation started — now renders the correct stacked area chart, matching its TSVB sibling panel's shape and per-extension proportions.

### DIFF
--- a/engine/crates/xerj-engine/src/aggs.rs
+++ b/engine/crates/xerj-engine/src/aggs.rs
@@ -1644,9 +1644,21 @@ fn evaluate_rpn(rpn: &[Tok]) -> Option<f64> {
                 // +inf, inf - inf is NaN): 0.0 before, 1.0 after. `==` is
                 // unchanged — `NaN == x` was 0.0 and still is. Pinned by
                 // `ne_against_a_nan_operand_is_true_not_false`.
+                //
+                // The `x == y` short-circuit is load-bearing, not an
+                // optimisation. Two EQUAL infinities are not NaN and must
+                // compare equal, but `inf - inf` IS NaN, so the epsilon test
+                // alone reports them unequal: measured, `params.a * params.a
+                // != params.a * params.a` over `n = 1e308` answered 1.0 with
+                // only the epsilon, where IEEE-754 and the JLS both say
+                // `inf == inf`. Exact equality first, epsilon second, keeps
+                // NaN reporting unequal (`NaN == NaN` is false and
+                // `(NaN).abs() < 1e-9` is false) while equal infinities and
+                // ordinary equal values report equal. Pinned by
+                // `eq_between_two_equal_infinities_is_true`.
                 let r = if *op == "==" || *op == "!=" {
                     let eq = match (a, b) {
-                        (EvalVal::Num(x), EvalVal::Num(y)) => (x - y).abs() < 1e-9,
+                        (EvalVal::Num(x), EvalVal::Num(y)) => x == y || (x - y).abs() < 1e-9,
                         (EvalVal::Null, EvalVal::Null) => true,
                         _ => false,
                     };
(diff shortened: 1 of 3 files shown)


xerj-org:
Holding this one. The `_count` resolution is right, but review found a **process kill** in the evaluator it uses, plus a semantics problem.

**Blocker: uncapped recursion in `eval_tokens`.** An ~80 KB `bucket_script` overflows the native stack, and with `panic = "abort"` in the release profile that is a process abort rather than an error response. This is the same finding as on #95, so the two share a root cause: the expression evaluator needs a depth bound in the style of `MAX_SQL_DEPTH` / `MAX_EVAL_DEPTH`, and `/_msearch` needs the request-time script guard the single-search path applies.

For context on why this gets a hard block: nested SQL `WHERE` parens, nested `query_string` parens, oversized Painless expressions, and Painless closure recursion have each shipped and been fixed here already. Every new recursive construct over user input needs its bound before it lands.

**High: bare `_count` reports the physical doc count**, which contradicts the bucket's own `_doc_count` weighting. So a weighted bucket and its `bucket_script` disagree about how many documents are in it.

Bound the recursion and the rest looks good.

xerj-org:
Rebased onto current `main` and updated to close the review findings. #95's two commits are contained in this branch verbatim, so merging this closes both PRs; #95 will be closed as included rather than merged separately.

**`_count` now agrees with `_doc_count` everywhere.** The original finding was that a `_doc_count`-weighted bucket and the `bucket_script` inside it disagreed about how many documents the bucket held. A first attempt fixed `terms` but inverted the same defect into `range` and `multi_terms`, which previously agreed. Verified per agg type on the brute path with a weighted fixture: `terms`, `range`, `multi_terms`, `histogram`, `date_histogram`, `auto_date_histogram`, `filter`, `filters`, `ip_range`, `geo_distance`, `nested` and the sampler family all now report `_count` equal to their own `doc_count`.

**Parenthesised ternaries evaluate.** `find_ternary_split` only looked at paren depth 0, so `(1?1:0)` — and by extension any nested ternary, and `(params.a > 0 ? 1 : 0) + 1` — silently resolved to a null bucket value. That undercut the point of adding ternary support.

**Two equal infinities compare equal again.** Deriving `==` from `(x - y).abs() < 1e-9` alone is wrong for infinities, because `inf - inf` is NaN and NaN fails the epsilon test, so two identical values compared *unequal*. Measured with one doc `{"n": 1e308}`: `params.a * params.a != params.a * params.a` answered 1.0 where IEEE-754 and the JLS both say `inf == inf`. Exact equality is now checked before the epsilon. NaN still reports unequal, which is the IEEE-correct behaviour this branch deliberately introduced and which is separately pinned.

**Behaviour changes, stated in full rather than summarised.** An earlier draft claimed "identical NaN behaviour" and a comment claimed the sub-agg-named-`_count` case was "the full extent of the silent-change surface". Both were wrong, and each is now pinned by a test rather than asserted in prose:

- `!=` against a NaN operand: 0.0 → 1.0. This is the IEEE-correct direction and is intentional.
- A sub-agg literally named `_count`: the bucket's doc count now wins, per ES's reserved-path rule.
- Object-form buckets_paths at a plain numeric sub-metric now resolve where they previously returned null.

Verified after the rebase: `cargo test -p xerj-engine --lib` → **346 passed, 0 failed**; `cargo fmt --check` and `cargo clippy -D warnings` both clean. The infinity fix is mutation-checked — removing the exact-equality short-circuit fails the new test and nothing else.

Thanks for the original fix and for #95 underneath it; the `_count` resolution was the right call and TSVB's `filter_ratio` is a real compatibility gap.

Issue #76, opened 2026-07-30
security: console-API hardening — x-forwarded-for trust (S5-4), magic-link TOCTOU (S5-3), field-limit bypass (S5-5), cluster/info disclosure (AUTHZ-2)


xerj-org:
**Severity:** Medium / Low / Info · **Status:** open · Found by the five-stage instruction/data self-audit. Grouped console-API hardening; full detail in `docs/case-studies/xerj-self-audit/FINDINGS-V2.md`.

**S5-4 (Medium, unauthenticated)** — `crates/xerj-console-api/src/auth/magic.rs:360` + `rate_limit.rs:58`. An attacker-supplied `x-forwarded-for` header is trusted as identity: it becomes the rate-limit bucket key and the audit-log source IP. Rotating the header resets the quota every request (unthrottled magic-link/bootstrap-token brute force) and spoofs audit attribution. Fix: derive the caller from `ConnectInfo<SocketAddr>` and honor `x-forwarded-for` only from an operator-configured trusted-proxy CIDR.

**S5-3 (Medium, needs a valid magic-link token)** — `magic.rs:221`. The single-use / bootstrap-claim check is TOCTOU: the "consume once" check (`:210`) and the commit (`:299`) are separated by several awaits with no exclusive lock, so two redeems of one token in a burst can both succeed. Fix: make consumption an atomic transition, first, under a per-token lock.

**S5-5 (Low, superuser-equivalent)** — `crates/xerj-engine/src/index.rs:14198`. `max_fields_per_index` is enforced on the dynamic-mapping paths (PR #69) but not on the explicit `Index::add_fields` path (`PUT /:index/_mapping`, `POST /v1/schema/:name/evolve`). Send fields explicitly and the mapping-explosion cap does not apply. Fix: enforce the limit at the mutation point (`add_field`/`ManagedSchema`).

**AUTHZ-2 (Info, unauthenticated)** — `crates/xerj-console-api/src/router.rs:49`. `cluster::info` takes no `AuthSession`, so `GET /_xerj-console/api/v1/cluster/info` returns `node_id`, exact version, and uptime with no credential on both listeners. Fix: drop `version`/`node_id` from the pre-auth response or gate it behind `AuthSession`.

xerj-org:
**Verified still open as of `v1.0.0-rc.8` (`7e050d0`)** — all four sub-items remain unfixed:

[...]

PR #107, opened 2026-08-01
security(console): key the auth rate limiter on the socket, not a header

xerj-org:
Closes the last open item on #76 (S5-4). The other three shipped in rc.9.

Client identity for the per-IP auth rate limiter and the audit-log source address now comes from the transport (`ConnectInfo<SocketAddr>`, threaded through both listener paths) instead of a caller-written `x-forwarded-for` header. Previously any caller could spoof the limiter key, which defeated the limiter for an attacker while still throttling honest clients.

`x-forwarded-for` is honoured only when the peer is in a configured trusted-proxy set, and the default trusts nothing, so a deployment behind a real reverse proxy configures it explicitly rather than getting it by accident.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,32 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **`x-forwarded-for` is no longer trusted for client identity** (#76 S5-4).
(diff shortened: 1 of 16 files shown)


xerj-org:
Superseded by a clean rebase. I damaged this branch with a scripted conflict resolution that dropped the commit instead of the conflict hunks. Reopening from the intact commit.

PR #108, opened 2026-08-01
security(cluster): authenticate Raft control frames, fail closed when unconfigured

### LINKED ISSUE #75: security(High, gated): unauthenticated Raft control messages on the cluster transport (DESER-EGRESS-01 / Raft-auth Phase 2)
**Severity:** High (operator-gated) · **Status:** open — tracked as the cluster Raft-auth Phase-2 item · **Reachability:** unauthenticated on the cluster port, only when `cluster.enabled=true` (default false; port then binds `0.0.0.0`)

Found by the five-stage instruction/data self-audit.

`crates/xerj-cluster/src/transport.rs` (`read_frame` → `handle_connection`) reads a length-prefixed frame off a raw TCP socket and `serde_json::from_slice`s it into a `RaftMessage` with **no authentication** before or after the parse. Any client that can reach the cluster port can forge consensus control messages (`AppendEntries`/`RequestVote`, term-bumps) and disrupt/steer the cluster. There is a 10 MiB frame cap (not an unbounded-alloc DoS) and serde_json is not a gadget deserializer (no RCE) — the risk is forged consensus.

This matches PR #69's deferred "cluster Raft auth" Phase-2 item and the gRPC path already has a `GrpcAuth` contrast to model on.

**Suggested fix:** require a shared cluster secret / per-frame HMAC, or mTLS peer-cert pinning to the configured peer set, before trusting a deserialized `RaftMessage`.

xerj-org:
Closes #75.

The cluster transport accepted a `[4-byte len][node_id]` sender header with no verification, then looped reading frames. With cluster mode enabled, anyone who could reach the port could send control frames.

Control frames are now authenticated, and the server refuses to start with cluster mode enabled and no secret configured rather than running unauthenticated. Failing closed is the substance of the issue, not a detail.

Wire compatibility: this changes the frame format, so a mixed-version cluster will not interoperate. Cluster mode is off by default, which is why the issue was gated rather than critical.


xerj-org:
Superseded by a clean rebase, same cause as #107.

PR #137, opened 2026-08-02
fix(engine) Fix transactional merge publication and rollback

buger:
DEPENDS ON https://github.com/xerj-org/xerj/pull/135

## Summary

This change makes the production engine merge path transactional across provisional version-map ownership and durable segment-snapshot publication.

Before this change, a merge worker repointed live document identities to its output segment before `apply_merge` made that output authoritative. An apply error, task cancellation, or panic could leave the global version map aimed at an unpublished segment. Reads could then reject the still-authoritative input rows as stale and fail to resolve the replacement, producing missing documents even though the input segment files still existed.

The fix adds an RAII rollback transaction for provisional repoints, defines explicit publication outcomes at the storage boundary, restores the exact input snapshot when durable publication fails, commits repoints at the durable manifest boundary, and keeps disjoint pending merge batches moving after one batch fails.

## Root cause

The merge implementation crossed two publication mechanisms without one owner for the commit boundary:

1. The blocking merge worker changed `VersionMap` entries from input segment IDs to the provisional output segment ID.
2. The async merge driver later called storage to replace the authoritative input snapshot with the output and persist the new manifest.
3. Nothing owned an exact rollback log across those steps.
4. Dropping the worker result on error, cancellation, or panic did not restore the repoints.
5. `apply_merge` changed the in-memory RCU snapshot before saving the manifest, but a save failure did not restore the exact prior input metadata.
6. If both publication persistence and compensating rollback persistence failed, the API had no honest state classification.
7. In the multi-batch driver, the apply-error `continue` bypassed the common queue top-up, so one failed batch could silently suppress a later disjoint batch from the same pass.

This was an ownership and commit-boundary defect, not a ranking algorithm defect. The change is a prerequisite for a later collection-wide BM25 publication snapshot, but it contains no scorer, global-statistics, query-planning, or hydration optimization.

### Conditional version-map transaction

`VersionRepointTransaction` reserves its complete rollback log before mutation, records the exact prior and installed `VersionEntry` for every accepted repoint, and rolls back on drop unless explicitly committed.

Rollback is conditional: an entry is restored only if it still exactly equals the value installed by this transaction. A newer concurrent PUT or DELETE therefore wins and is never overwritten by rollback. The transaction also restores `live_count` consistently and removes entries that did not exist before the provisional repoint.

### Tagged storage outcomes

Transactional merge publication now reports one of these states:

[...]

PR #138, opened 2026-08-02
fix(engine) ate document reads on stable collection publication

buger:
DEPENDS ON: https://github.com/xerj-org/xerj/pull/137

## Summary

This change prevents public `search` and `get_document` calls from returning or caching a mixture of collection states while supported document writes, flushes, or engine merges publish their components in stages.

It introduces a per-index collection-publication authority with an exact generation, active-writer count, and sticky poison state. Readers execute only against one stable generation and validate that generation again before returning. A generation change restarts the complete operation within the original request deadline. An uncertain publication fails the live index closed until reopen reconstructs it from the durable segment snapshot and WAL.

This is a correctness prerequisite for collection-wide ranking. It contains no BM25 scorer, collection-statistics, candidate-collection, FTS-ranking, or hydration optimization.

## Root cause

Search-visible state is distributed across the version map, lexical memtables, immutable segment snapshot, stored-source cache, query cache, and single-flight results. PUT, bulk ingest, DELETE, flush, and merge do not publish every component in one machine instruction.

Before this change, a reader could capture those components at different moments. For example, a PUT could advance version ownership before the replacement lexical row became visible. A query could reject the old segment row as stale while also missing the new memtable row, then cache that incomplete result as successful. Merge had the analogous provisional-repoint-to-snapshot window. Caller cancellation after the first durable mutation also left the live process unable to prove that its in-memory structures described one coherent generation.

The existing per-document write coordinator serializes same-ID updates, and HNSW has its own publication guard, but neither is a collection-wide read snapshot.

### Collection publication authority

Each `Index` owns an `Arc<CollectionPublication>` containing:

- a monotonically increasing publication generation;
- the number of admitted writers currently inside an authoritative publication interval;
- a sticky poison bit for an interrupted or panicked mutation whose in-process coherence cannot be proven;
- a notification primitive used by readers waiting for active publication to finish.

A writer begins immediately before its first authoritative mutation. A complete publication calls `commit()`. A verified no-op or fully rolled-back operation calls `cancel()`. Dropping a pending guard marks the index poisoned and wakes waiters.

Writers admitted while the index was healthy may overlap. If one later poisons the index, another already-admitted writer may complete, but poison remains sticky: later commits cannot clear it and no new reader or writer is admitted. Reopen is the only recovery boundary.

### Stable reader tokens

Public `search` and `get_document` admit a reader only when the collection has no active writer and is not poisoned. The token records the exact stable generation.

The operation validates that token before any successful result leaves the method. If the generation changed, the complete operation retries within the original absolute deadline. If the index became poisoned, the operation returns a retryable error explaining that the index must be reopened so snapshot and WAL recovery can reconstruct it.

The query-cache and single-flight key stores `body_hash`, `dataset_version`, and `publication_generation` as independent fields. It does not fold them with XOR, so distinct states cannot alias. Cache hits, followers, fresh results, and cache publication all remain subject to final token validation.

### Bounded publication intervals

Expensive or fallible preparation occurs before admission where possible. The guarded interval contains only the state transition that must be seen atomically by `search` and `get_document`:

- single PUT/create/update publication;
- parsed turbo ingest;
- async raw ingest after parsing, copy-to, and schema preparation;
- synchronous raw ingest after preflight;
- realtime no-WAL lexical publication;
- DELETE/tombstone publication;
- memtable drain through detached flush finalization or verified rollback;
- provisional merge repoints through tagged durable merge apply.

Schema evolution, HNSW work, metrics, and later flush scheduling remain outside the lexical publication interval when they have their own consistency mechanism. HNSW retains its existing full-operation RAII guard and exact fallback while graph mutation is incomplete.

### Flush ownership

Flush acquires the collection guard before draining searchable memtable rows and transfers ownership into the detached finalizer. Cancelling the waiting request cannot release the barrier early. A raced empty drain cancels cleanly; successful durable publication commits; verified prepublication restoration cancels; an unclassified unwind poisons.

### Merge ownership

The blocking merge builder now returns immutable repoint intents rather than mutating the version map before the barrier. The async driver begins collection publication, applies the conditional repoint transaction synchronously, and calls the tagged transactional storage apply from the prerequisite branch.

- `Published` commits the collection generation.
- `NotPublished` cancels after exact rollback.
- `Indeterminate` leaves the pending guard to poison the live index.
- Panic after durable publication also poisons the live process; reopen uses the durable output snapshot and committed repoints.

This keeps expensive merge construction outside the reader outage while covering the actual repoint-to-durable-snapshot interval.

## Supported scope

The proof covers stable public `search` and `get_document` behavior during:

- ordinary PUT/create/OCC document publication;
- parsed turbo, raw, synchronous raw, and realtime ingest publication;
- DELETE and the tested update/delete delegations;
- engine flush success, detached completion, and verified rollback;
- the supported production engine merge path and its `Published`, `NotPublished`, `Indeterminate`, cancellation, and postpublication-panic outcomes;
- cache and single-flight reuse across publication generations;
- sticky fail-closed behavior and exact recovery after reopen.

WAL replay is construction-time recovery before the new `Index` is published and therefore does not acquire a runtime collection guard.

## Explicit exclusions

This PR does not claim a whole-index transaction or protection for every callable helper. Mapping/schema lifecycle changes, `delete_all_data`, direct `lookup_seq_no`/`lookup_version`, snapshot and term-enumeration helpers, specialized internal vector/semantic/hybrid executors called outside public search routing, and the hidden legacy storage `MergeExecutor` are outside the proven API surface.

The public server query routes evaluated here ultimately use guarded `search` or `get_document`. A follow-up should centralize or explicitly guard any additional externally supported reader that bypasses those methods before claiming all-read consistency.

Poison is deliberately fail-closed rather than self-healing. Once poisoned, the affected live `Index` rejects guarded reads and new supported document publication. Reopen rebuilds the version map from the durable segment snapshot, reapplies tombstones, recovers completed orphan publication, and replays WAL into storage and lexical memtables. There is no runtime `clear_poison()` API.

Active readers may wait for the complete flush finalizer because destructive drain removes memtable visibility before segment publication. That availability cost is correctness-preserving and should be measured separately; this PR makes no latency or throughput claim.

## Deterministic manual tests

Use a shared target outside the worktree:

```bash
cd engine
export CARGO_TARGET_DIR=/workspace/targets/collection-publication-epoch
cargo test -j 32 -p xerj-engine collection_publication --lib -- --nocapture
cargo test -j 32 -p xerj-engine merge_publication_transaction --lib -- --nocapture
cargo test -j 32 -p xerj-engine flush_publication_recovery --lib -- --nocapture
```

The 16 collection-publication tests cover stable tokens, exact cache-generation identity, active reader blocking, create/OCC rejection before admission, normal/turbo/raw/sync/realtime/delete publication, cancellation before and after WAL, overlapping already-admitted writers, sticky poison, and exact WAL recovery after reopen.

The six merge tests cover immutable pre-guard intents, reader blocking after guarded repoint, cancellation and conditional rollback, safe failure, post-durable panic, exact reopen, concurrent ownership, retry, and partial multi-batch propagation.

The eight flush tests cover running and queued finalizers, detached completion after waiter cancellation, cache invalidation, vector publication/restart, exact restoration, and failure paths. The reused-ID HNSW and ANN generation-retry regressions additionally prove that public search blocks inside ordinary publication and safely retries after generation change.

Run the complete local gate:

```bash
cd engine
export CARGO_TARGET_DIR=/workspace/targets/collection-publication-epoch
cargo test -j 32 -p xerj-storage
cargo test -j 32 -p xerj-engine --lib
cargo clippy -j 32 -p xerj-storage -p xerj-engine --all-targets -- -D warnings
cargo fmt --all --check
git diff --check 05fd42770920aa116ede0e83e07e0440bd34be54..HEAD
cargo build -j 32 -p xerj-server -p es-yaml-runner
```

The server/runner build is intentionally development profile locally; release/fat-LTO confirmation belongs to CI.

## Verified gates

- Exact commit: `ba72d70388c1a42ed0dd46964df0921361770125`.
- Exact tree: `0e9333da13118f77bfb601235d1de6e9896ea90b`.
- `xerj-storage`: 140 passed, 0 failed.
- `xerj-engine --lib`: 376 passed, 0 failed.
- Strict scoped all-target Clippy for `xerj-storage` and `xerj-engine` with warnings denied: passed.
- `cargo fmt --all --check`: passed.
- `git diff --check`: passed.
- Ranking-exclusion audit: only `collection_publication.rs`, `index.rs`, and `lib.rs` changed; no `xerj-fts`/memtable changes and no BM25/global-ranking symbols.
- Scoped development build of `xerj-server` and `es-yaml-runner`: passed in approximately three minutes.
- Fresh ES-YAML against the exact preserved candidate binary: 1365 passed, 0 failed, 3 skipped, 1368 total.
- Cleanup: candidate server stopped, port 9200 released, runtime data and temporary target removed, source worktree clean.

## Sealed evidence

Evidence directory: `/workspace/north-star-evidence/collection-publication-epoch-ba72d70-v1`

- Evidence checksum manifest SHA-256: `86d4ca26af860283e122f8e168d5e80ab4bb338593115d5e732d480e64e81531`.
- Candidate `xerj` SHA-256: `f6fd728eacda28f943fb7a08d56cc1aad81d4af3aa9d8577f1000cf2862dd9bf`.
- Candidate `es-yaml-runner` SHA-256: `7ee056f0f6930cfb7920d202829013d557e4003b4147e22aefad30dde9e46859`.
- Runner log SHA-256: `15fed233486ba3056e8eee6deca2eb654a4265721657627a38bf20d82c49b1fb`.
- Server log SHA-256: `4841938dc16e8081d89ac78b7d417cf5683200f7a32ef9d1b761351adf209f50`.
- Provenance SHA-256: `4958a30460c3959b5e70f566d56cbdc4b61aecfca23324c0ed62be1a74ddb27b`.
- Seal SHA-256: `6e5b1add5db70c3e43fb1d066335062184b1be8096b4d06b92f00e69e69f2f5b`.

Verify the retained payloads with:

```bash
cd /workspace/north-star-evidence/collection-publication-epoch-ba72d70-v1
sha256sum -c SHA256SUMS
```

## Dependency stack

This is the third contribution in a deliberate correctness stack:

1. `ee8a2d739b585be69c4c5113f3bfe554c7b65d2b` — atomic flush publication and recovery.
2. `eecaf0a` + `34fa244` + `05fd42770920aa116ede0e83e07e0440bd34be54` — transactional merge publication and promotion suite.
3. `ba72d70388c1a42ed0dd46964df0921361770125` — stable collection publication for public search and GET.

Review this branch against `05fd427` while the first two prerequisites are pending. After each prerequisite merges, rebase the remaining stack onto current upstream `main` and rerun the hard gate on the exact rebased candidate.

## Files changed

- `engine/crates/xerj-engine/src/collection_publication.rs`: publication authority, writer guards, reader tokens, poison state, notification, and primitive tests.
- `engine/crates/xerj-engine/src/index.rs`: guarded writer boundaries, stable search/GET retry, exact cache identity, flush/merge ownership, HNSW-composition fixtures, and deterministic integration tests.
- `engine/crates/xerj-engine/src/lib.rs`: module registration.

## Review focus

Please focus on writer guard boundaries, reader-token validation before every successful return, sticky poison semantics, the exact cache/single-flight key, detached flush ownership, merge-intent timing, and the absence of an await inside the short raw lexical publication interval. The narrow scope and exclusions above are intentional.

### DIFF
--- a/engine/crates/xerj-engine/src/collection_publication.rs
+++ b/engine/crates/xerj-engine/src/collection_publication.rs
@@ -1,0 +1,219 @@
+use std::sync::atomic::{AtomicU64, Ordering};
+use std::sync::Arc;
+
+/// Per-index seqlock-style publication barrier for collection-wide captures.
+/// Writers may overlap, but a reader only accepts a capture made while no
+/// writer was active and the generation remained unchanged.
+pub(crate) struct CollectionPublication {
+    generation: AtomicU64,
+    in_flight: AtomicU64,
+    poisoned: std::sync::atomic::AtomicBool,
+    notify: tokio::sync::Notify,
+    #[cfg(test)]
+    reader_admission_attempts: AtomicU64,
+}
+
+pub(crate) struct CollectionPublicationGuard {
+    state: Arc<CollectionPublication>,
+    outcome: GuardOutcome,
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) struct CollectionReadToken {
+    generation: u64,
+}
+
+impl CollectionReadToken {
+    pub(crate) fn generation(self) -> u64 {
+        self.generation
+    }
+}
+
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub(crate) enum ReadAdmission {
+    Admitted(CollectionReadToken),
+    WriterActive,
+    Poisoned,
+}
+
+#[derive(Clone, Copy, PartialEq, Eq)]
+enum GuardOutcome {
+    Pending,
+    Committed,
+    Cancelled,
+}
+
+impl CollectionPublication {
+    pub(crate) fn new() -> Arc<Self> {
+        Arc::new(Self {
+            generation: AtomicU64::new(0),
+            in_flight: AtomicU64::new(0),
+            poisoned: std::sync::atomic::AtomicBool::new(false),
+            notify: tokio::sync::Notify::new(),
+            #[cfg(test)]
+            reader_admission_attempts: AtomicU64::new(0),
+        })
+    }
+
+    pub(crate) fn begin(
+        self: &Arc<Self>,
+    ) -> Result<CollectionPublicationGuard, CollectionPublicationPoisoned> {
+        if self.poisoned.load(Ordering::Acquire) {
+            return Err(CollectionPublicationPoisoned);
+        }
+        self.in_flight.fetch_add(1, Ordering::AcqRel);
+        self.generation.fetch_add(1, Ordering::AcqRel);
+        // A concurrent failed publisher may poison after our first check.
+        // This writer has not mutated yet, so withdraw cleanly and reject it.
+        if self.poisoned.load(Ordering::Acquire) {
+            self.generation.fetch_add(1, Ordering::Release);
+            let previous = self.in_flight.fetch_sub(1, Ordering::AcqRel);
+            debug_assert!(previous > 0, "collection publication counter underflow");
+            if previous == 1 {
+                self.notify.notify_waiters();
+            }
+            return Err(CollectionPublicationPoisoned);
+        }
+        Ok(CollectionPublicationGuard {
+            state: Arc::clone(self),
+            outcome: GuardOutcome::Pending,
+        })
+    }
+
+    pub(crate) fn state(&self) -> (u64, u64, bool) {
+        loop {
+            let before = self.generation.load(Ordering::Acquire);
+            let in_flight = self.in_flight.load(Ordering::Acquire);
+            let after = self.generation.load(Ordering::Acquire);
+            if before == after {
+                return (after, in_flight, self.poisoned.load(Ordering::Acquire));
+            }
+        }
+    }
+
+    pub(crate) fn notified(&self) -> tokio::sync::futures::Notified<'_> {
+        self.notify.notified()
+    }
+
+    pub(crate) fn try_admit_reader(&self) -> ReadAdmission {
+        #[cfg(test)]
+        self.reader_admission_attempts
+            .fetch_add(1, Ordering::Release);
+        let (generation, in_flight, poisoned) = self.state();
+        if poisoned {
+            ReadAdmission::Poisoned
+        } else if in_flight != 0 {
+            ReadAdmission::WriterActive
+        } else {
+            ReadAdmission::Admitted(CollectionReadToken { generation })
+        }
+    }
+
+    #[cfg(test)]
+    pub(crate) fn reader_admission_attempts(&self) -> u64 {
+        self.reader_admission_attempts.load(Ordering::Acquire)
+    }
+
+    pub(crate) fn validate_reader(&self, token: CollectionReadToken) -> ReadAdmission {
+        let (generation, in_flight, poisoned) = self.state();
+        if poisoned {
+            ReadAdmission::Poisoned
+        } else if in_flight != 0 || generation != token.generation {
+            ReadAdmission::WriterActive
+        } else {
+            ReadAdmission::Admitted(token)
+        }
+    }
+}
+
+#[derive(Debug, Clone, Copy)]
+pub(crate) struct CollectionPublicationPoisoned;
+
+impl CollectionPublicationGuard {
+    pub(crate) fn commit(&mut self) {
+        self.outcome = GuardOutcome::Committed;
+    }
+    pub(crate) fn cancel(&mut self) {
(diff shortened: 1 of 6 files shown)


xerj-org:
Superseded by the rebased version (your commit, authorship preserved) — #135/#137 landed as squashes so this fork branch could no longer be rebased in place, and maintainer-edit is off. Merged verification: 404+140 tests green, adversarial crash/cancel analysis clean; the two disclosed availability tradeoffs are carried into the new PR body. Closing as included; thank you.

PR #154, opened 2026-08-03
perf(onnx): add reproducible fused FP32 model recipe

buger:
# Summary

This adds a reproducible, fail-closed offline recipe for the exact transformer-fused FP32 MiniLM ONNX graph that XERJ already accepts through `--embed-mode onnx-experimental --onnx-model ...`. It does not add a new embedding backend, change scoring, bundle the approximately 90 MB model, download assets, or optimize a model during server startup.

# Why

The fused graph previously existed only as an external experiment artifact. The repository had no reviewable way to reproduce its bytes, identify the exact source and tokenizer, pin the optimizer toolchain, validate the complete graph topology, or explain that the fused graph has a distinct embedding identity. Generic ONNX Runtime Level-3 session optimization does not produce this same transformer-fused graph, and serialized runtime caches may contain hardware-specific transformations.

# What changes

- Adds a CPython 3.13.5 recipe with exact optimizer package versions and SHA-256-pinned GNU/Linux x86-64 wheels.
- Accepts only the recorded source model and tokenizer byte lengths and SHA-256 identities.
- Requires the reproduced candidate to match the exact expected byte length and SHA-256.
- Validates the complete ONNX IR version, opsets, input/output contract, initializer types, external-data policy, node count, and operator inventory.
- Loads `artifact-contract.json` as the runtime source of truth, compares the full normalized observation to it, and embeds that complete contract plus observed source/candidate inventories in the generated manifest.
- Stages and fsyncs model plus manifest inside one private sibling directory, reserves the destination through an `O_EXCL` owner lock, and publishes the complete directory in one Linux `renameat2(RENAME_NOREPLACE)` operation.
- Verifies the lock device/inode before cleanup so an exiting process cannot remove another process's replacement lock.
- Documents the supported CPU/runtime boundary and the mandatory fresh-index migration.

# Reproduce

```bash
python3.13 -m venv /path/to/xerj-onnx-recipe-venv
/path/to/xerj-onnx-recipe-venv/bin/python -m pip install --only-binary=:all: --require-hashes -r demo/playbooks/onnx-model-optimization/requirements-linux-x86_64-cp313.lock

/path/to/xerj-onnx-recipe-venv/bin/python demo/playbooks/onnx-model-optimization/optimize_minilm_fp32.py \
  --source /models/minilm-source/model.onnx \
  --tokenizer /models/minilm-source/tokenizer.json \
  --output-dir /models/minilm-fused-v1
```

The final directory contains:

- `model.bert-fused-fp32.onnx`
- `model.bert-fused-fp32.manifest.json`

# Use with XERJ

```bash
target/release/xerj \
  --insecure \
  --data-dir /var/lib/xerj-fused \
  --embed-mode onnx-experimental \
  --onnx-model /models/minilm-fused-v1/model.bert-fused-fp32.onnx \
  --onnx-tokenizer /models/minilm-source/tokenizer.json

target/release/xerj autoindex /path/to/corpus \
  --url http://localhost:9200 \
  --prefix finance-minilm-fused-v1 \
  --fresh
```

The fused graph is not bit-identical to its source graph and therefore has a different `embedding_identity.json` model hash. Do not replace the source model for an existing index. Retain the original asset for that index, or use the fused model with `autoindex --fresh` and a new prefix.

# Compatibility boundary

The offline generation environment is CPython 3.13.5 with ONNX 1.18.0 and ONNX Runtime transformer optimizer 1.22.1. Consumption is proven only through XERJ's `ort`/`ort-sys` 2.0.0-rc.12 API-24 CPU path on GNU/Linux glibc x86-64. The graph uses `com.microsoft` contrib operators. GNU/Linux AArch64, musl, other runtime/API lines, other execution providers, and non-ORT consumers remain unproven.

# Reproduction result

A fresh outside-repository run through the final atomic-directory transaction completed successfully with empty stderr. It generated a 90,276,689-byte model at SHA-256 `d45c738311abe1488b6d9e2862e1db08267fc31d6f301d8f26163e2fb4956b9c`, byte-identical to the original experiment artifact. The generated manifest embedded the checked-in artifact contract exactly, and the published directory contained exactly the model and manifest with no remaining lock or staging directory.

This PR deliberately makes no public speed or retrieval-quality claim. The motivating optimizer screen is not a checked-in sanitized ten-arm benchmark campaign, so quoting its private measurements here would violate the repository's evidence rules.

# Validation

- 12 focused Python tests passed.
- `py_compile` passed.
- Artifact contract JSON validation passed.
- Full generated-manifest contract equality passed.
- Exact candidate SHA-256 and byte equality passed.
- Atomic publication, collision rejection, `O_EXCL` conflict, replacement-lock ownership, every topology field, same-size wrong-hash rejection, toolchain admission, full contract drift, and injected prepublication failure/no-residue cases are covered.
- `git diff --check` passed.
- Independent blocker review passed after corrections.
- Independent final GO audit passed.

# Scope

No model or tokenizer bytes are committed. No server/runtime code changes. No automatic downloader or model manager. No runtime graph rewriting or cache serialization. No quantization. No provider expansion. No scoring, pooling, chunking, scheduling, or embedding-identity relaxation.

### DIFF
--- a/.github/workflows/onnx-model-optimization-recipe.yml
+++ b/.github/workflows/onnx-model-optimization-recipe.yml
@@ -1,0 +1,37 @@
+name: ONNX model optimization recipe
+
+on:
+  pull_request:
+    paths:
+      - ".github/workflows/onnx-model-optimization-recipe.yml"
+      - "demo/playbooks/onnx-model-optimization/**"
+  push:
+    branches:
+      - main
+    paths:
+      - ".github/workflows/onnx-model-optimization-recipe.yml"
+      - "demo/playbooks/onnx-model-optimization/**"
+
+permissions:
+  contents: read
+
+jobs:
+  locked-recipe-tests:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.13.5"
+          cache: pip
+          cache-dependency-path: demo/playbooks/onnx-model-optimization/requirements-linux-x86_64-cp313.lock
+      - name: Install checksum-locked optimizer environment
+        run: >-
+          python -m pip install
+          --only-binary=:all:
+          --require-hashes
+          -r demo/playbooks/onnx-model-optimization/requirements-linux-x86_64-cp313.lock
+      - name: Run recipe unit tests
+        run: >-
+          python -m unittest
+          demo/playbooks/onnx-model-optimization/test_optimize_minilm_fp32.py
--- a/demo/playbooks/onnx-model-optimization/.gitignore
+++ b/demo/playbooks/onnx-model-optimization/.gitignore
@@ -1,0 +1,2 @@
+__pycache__/
+*.pyc
--- a/demo/playbooks/onnx-model-optimization/README.md
+++ b/demo/playbooks/onnx-model-optimization/README.md
@@ -1,10 +1,16 @@
-# Reproduce the measured fused FP32 MiniLM graph
+# Transform the exact pinned MiniLM export into a fused FP32 graph
 
-This offline recipe transforms one exact
-`sentence-transformers/all-MiniLM-L6-v2`-compatible ONNX export into the
-transformer-fused FP32 graph measured by XERJ's FB20 optimizer-regression
-screen. It does **not** download, commit, install, or select a model. The
-approximately 90 MB source and output remain external runtime assets.
+This offline recipe transforms one exact, previously recorded
+`sentence-transformers/all-MiniLM-L6-v2`-compatible ONNX export into one exact
+transformer-fused FP32 graph. It does **not** download, commit, install, select,
+or recreate the source model. The approximately 90 MB source and output remain
+external runtime assets.
+
+This is a checksum-locked transformation recipe, not a from-upstream
+reproduction recipe. You must already possess the exact source model and
+tokenizer bytes listed below. The repository does not publish those assets or
+an immutable exporter revision, and a fresh generic `optimum-cli export` is not
+expected to reproduce them.
 
 The recipe is intentionally narrow and fail-closed:
 
@@ -40,7 +46,7 @@ python3.13 -m venv /path/to/xerj-onnx-recipe-venv
   -r demo/playbooks/onnx-model-optimization/requirements-linux-x86_64-cp313.lock
 ```
 
-## Reproduce
+## Transform the pinned source
 
 ```bash
 /path/to/xerj-onnx-recipe-venv/bin/python \
@@ -67,12 +73,17 @@ Required result:
 | fused model | 90,276,689 | `d45c738311abe1488b6d9e2862e1db08267fc31d6f301d8f26163e2fb4956b9c` |
 
 The manifest embeds the complete checked-in artifact contract and records the
-observed source and candidate graph inventories plus the exact toolchain. The
-recipe compares the complete normalized observation to the contract before it
-publishes anything. The contract also records that the
-original experiment did not preserve an immutable upstream export revision:
-the exact source bytes are pinned, but a stronger upstream provenance claim
-would be dishonest.
+observed source and candidate graph inventories plus the exact toolchain.
+Before publication, the recipe enforces the observed source and tokenizer
+bytes and hashes, candidate bytes, hash and complete graph inventory, package
+versions, interpreter, and the optimizer arguments that it actually executes.
+
+Other contract fields are recorded metadata, not observations made by this
+script: artifact name/version, source license and provenance, the XERJ
+embedding contract, and the compatibility support boundary. In particular,
+the original experiment did not preserve an immutable upstream export
+revision. Exact source bytes are pinned, but a stronger upstream provenance
+claim would be dishonest.
 
 [`artifact-contract.json`](artifact-contract.json) is the checked-in expected
 contract. The generated manifest adds the complete inspected source and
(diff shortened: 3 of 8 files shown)


xerj-org (dismissed):
🙏 Thank you for this, @buger. This is a serious piece of work, and the parts that are enforced are properly enforced: source and tokenizer bytes plus SHA-256, artifact bytes plus SHA-256, and the full graph inventory. The `renameat2` publication, the lock inode-ownership check and the failure-cleanup path all held up when we exercised them directly (11 of 12 tests pass here; the 12th is discussed below).

We reviewed it across correctness, supply chain and claim accuracy, and verified the significant findings by running the code rather than reading it. Five things need a change before this can land. None of them are about the engineering, and all of them are about the same thing: the docs currently promise more enforcement than the code performs, and this repo's first ground rule is that a public claim has to trace to something a reader can verify.

**1. The contract is not fully compared, but the README says it is.** `README.md:71` states the recipe "compares the complete normalized observation to the contract before it publishes anything." `observed_contract()` starts from `copy.deepcopy(CONTRACT)` and overwrites only `artifact.bytes/sha256`, `graph.*`, `optimizer.packages`, `source.model` and `source.tokenizer`. Everything else is compared against itself. We set `compatibility.proven.xerj_ort_api=999`, `optimizer.arguments.num_heads=4096`, `optimizer.python="CPython 2.7"` and `xerj_contract.dimensions=1`, then ran `observed_contract()` + `verify_full_contract()`. **It passed.** The most consequential decorative block is `optimizer.arguments`, which is the operative recipe configuration but is duplicated as hardcoded literals at `optimize_minilm_fp32.py:383-387`, so editing the JSON changes documentation and nothing else.

Preferred fix is to make the sentence true by driving the optimizer from the contract (`optimize_model(str(source), **CONTRACT["optimizer"]["arguments"])`, minus `use_external_data_format`) and parsing `optimizer.python` for the interpreter gate. Otherwise please reword `README.md:71` to enumerate exactly which fields are verified and mark the rest as recorded-but-unverified metadata.

**2. `test_any_full_contract_drift_fails` asserts a property the production path lacks.** It hand-builds `drifted` and passes it straight to `verify_full_contract()`, bypassing `observed_contract()`. Applying the identical mutation to a real projection passes, because that field is deep-copied verbatim. Either route the test through `observed_contract()` and assert what actually happens, or rename it and drop the "any". Fixing (1) would let it be true as written.

**3. Named evidence that is not in the repository.** `grep -rn` returns zero hits for "FB20", "optimizer-regression", "ten-arm" and the "sealed 368-PDF North Star", and none of the three pinned SHA-256 values appear anywhere else. `docs/EXPERIMENTAL_ONNX.md` is user-facing and now cites "the measured FB20 optimizer-regression screen" as though a reader could go and find it. The README's "Evidence boundary" section is a good-faith disclaimer, but it does not cover the forward references planted in the user-facing doc. Please either land the sanitized screen under `demo/playbooks/`, or say "an internal optimizer screen whose evidence is not published in this repository".

**4. Third parties cannot currently run this.** The recipe hard-fails unless `--source` is byte-identical to `6fd5d72f…` (90,405,214 B), but `immutable_export_revision` is `null` and nothing pins the exporter toolchain. The only documented way to obtain a source model is `optimum-cli export onnx`, whose output bytes are not stable across optimum/torch versions, so the realistic outcome for anyone else is an immediate SHA-256 mismatch with no documented recovery. That makes "reproducible" in the title stronger than what ships. Please pin an acquisition path, or state plainly that the exact file must be supplied.

**5. The value proposition is unstated, while the cost is real.** `engine/crates/xerj-ai/src/onnx.rs:216` already calls `.with_optimization_level(GraphOptimizationLevel::Level3)` at session creation, which performs the same Attention / EmbedLayerNormalization / SkipLayerNormalization / BiasGelu fusions in-process for every session. The PR never mentions this, so a reader cannot tell what the offline recipe buys, while the cost is stated and real: a new embedding identity and a mandatory `autoindex --fresh` reindex. Please name the specific benefit (avoided startup fusion cost, most likely), or say the steady-state difference is unpublished rather than leaving it to be inferred.

**Minor.** The suite is not wired into any workflow, and `test_current_locked_toolchain_is_exact` calls `verify_toolchain()` unguarded, so it errors on any interpreter that is not CPython 3.13.5 with all 11 pinned versions (11 passed / 1 error on Python 3.14.6 here). As it stands the only test exercising the lock file is permanently red outside your venv, so a resolver conflict inside it would never surface. Worth a skip guard plus CI wiring.

To be clear about where this stands: we want this in. The `renameat2` publication and the lock handling are better than a lot of what is already in this repo, and the embedding-identity warning is exactly the right thing to be loud about. It is the doc-versus-code gap that has to close first, because "Honest claims only" is a rule we have already had to enforce against ourselves by deleting our own benchmark claims. Happy to take (1) either way you prefer, and if the FB20 screen cannot be published, saying so in one sentence is a perfectly good outcome. 🚀

buger:
Addressed the requested review in commit `e97cca6`.

1. Operative optimizer arguments now come directly from the strictly validated contract. The CPython implementation and exact version are also parsed from and enforced by that contract. The former hard-coded optimizer arguments have been removed.
2. The misleading “any contract drift” test was replaced with tests that exercise the real observation and production dispatch paths. The suite also proves that contract values such as `num_heads` reach `optimize_model`.
3. Unpublished FB20, ten-arm, and North Star evidence references were removed or explicitly described as internal and unpublished.
4. The documentation now calls this a checksum-locked transformation of exact source assets—not a generally reproducible model-acquisition recipe. It explicitly states that the required source and tokenizer bytes must already be available, that their acquisition/export path is not published, and that a generic `optimum-cli export onnx` output is not expected to match.
5. The documentation now acknowledges that XERJ already requests ONNX Runtime Level3 optimization. It does not claim that runtime Level3 creates the same graph, nor does it claim a published incremental cold-start, steady-state throughput, memory, or retrieval-quality benefit. The distinct embedding identity and reindexing cost remain explicit.
6. Added the github workflow too

Enforced observations and operative configuration are now clearly separated from recorded metadata.

Validation:

- 15/15 focused tests passed in the pinned CPython 3.13.5 environment.
- Python byte-compilation passed.
- `git diff --check` passed.
- An independent re-review found no remaining blocker against the five requested changes.


One qualification regarding point 5: the source confirms that XERJ enables runtime Level3, but it does not prove that runtime Level3 performs exactly the same transformer fusions as the offline optimizer. The updated documentation therefore avoids making that equivalence claim.

PR #156, opened 2026-08-03
fix(autoindex): refuse unsupported corpus reconciliation

buger:
# Summary

This makes `xerj autoindex` fail closed when a rerun cannot safely reconcile the current folder with its durable resume plan.

The supported contract after this change is explicit:

- An interrupted run resumes completed work at whole-file granularity.
- An unchanged rerun reuses completed files after verifying full-content identities.
- Replacing the contents of an existing path is detected and republished under the planned identity.
- Adding or removing a canonical content group is detected and rejected before journal or destination mutation because complete membership reconciliation is not implemented yet.
- `--fresh` is accepted only when the selected state directory has no durable plan. It never resets or reconciles an existing destination.

This is a correctness and operator-truthfulness change. It does not implement incremental add/delete synchronization; that remains follow-up work.

# Root cause

The resume journal is more than a local progress cache. Its frozen plan carries the identities needed to reconcile published data records, duplicate aliases, graph edges, catalog entries, replacements, and stale records.

Discarding that plan cannot prove convergence of an existing destination. Likewise, silently accepting files outside the frozen plan can publish only part of the new corpus while leaving removed artifacts live.

`xerj brain` also previously treated an absent or zero live-node probe as evidence that the destination was empty and automatically retried with fresh state. An absent metadata document, a malformed response, a transport failure, and an authoritative zero are different states; none proves that every data, graph, and catalog namespace is empty.

# What changes

## One authoritative preflight

Autoindex now acquires the exclusive state lock before directory discovery and full hashing. The same lock covers journal inspection, inventory classification, torn-tail handling, and authoritative open. A waiting invocation cannot classify a path snapshot taken while another invocation is publishing or replacing the plan.

## Fail-closed inventory classification

The current canonical content groups are compared with the durable plan before any mapping creation, delete-by-query, bulk publication, refresh, graph write, catalog write, or journal append.

Added and vanished groups produce an actionable unsupported-delta error. Same-path content replacement remains supported: the existing planned key is retained, its full digest is compared, replacement intent invalidates completion, stale records are cleaned, and the file is republished.

Human output goes to stderr with a nonzero exit. Machine-readable mode emits structured JSON describing added and vanished groups, confirming that the attempt made no destination changes, and naming the supported recovery choices.

## Honest `--fresh`

`--fresh` no longer discards a durable plan. Even an apparently unchanged key set can have alias, path, graph, catalog, or partial-publication history that only the old plan can reconcile.

An isolated rebuild requires:

- a new `--state-dir`
- a new `--prefix`
- a new `--brain` when graph indexing is enabled, or `--no-graph`
- validation before switching readers
- explicit cleanup of old targets after validation

The global autoindex catalog remains shared, so the change does not describe this as a completely isolated destination.

## Honest `xerj brain` recovery

`xerj brain` no longer automatically retries with `fresh` when the journal and server disagree. The live-node probe preserves absence, authoritative zero, malformed metadata, and request failure as distinct outcomes. Probe errors propagate instead of being converted into “empty.”

The refusal identifies the journal, URL, prefix, and brain namespace and prints an executable isolated-rebuild command. API-key values are never echoed.

# Manual reproduction

Start XERJ:

```bash
target/debug/xerj --insecure --data-dir /workspace/xerj-rescan-demo/server
```

Create and index a folder:

```bash
mkdir -p /workspace/xerj-rescan-demo/corpus
printf 'quarter,revenue\nQ1,100\n' > /workspace/xerj-rescan-demo/corpus/report.csv

target/debug/xerj autoindex /workspace/xerj-rescan-demo/corpus \
  --url http://localhost:9200 \
  --state-dir /workspace/xerj-rescan-demo/state \
  --prefix rescan-demo \
  --brain rescan-demo
```

An unchanged rerun succeeds and reuses the completed file:

```bash
target/debug/xerj autoindex /workspace/xerj-rescan-demo/corpus \
  --url http://localhost:9200 \
  --state-dir /workspace/xerj-rescan-demo/state \
  --prefix rescan-demo \
  --brain rescan-demo
```

Add a new file and rerun:

```bash
printf 'quarter,revenue\nQ2,120\n' > /workspace/xerj-rescan-demo/corpus/report-q2.csv

target/debug/xerj autoindex /workspace/xerj-rescan-demo/corpus \
  --url http://localhost:9200 \
  --state-dir /workspace/xerj-rescan-demo/state \
  --prefix rescan-demo \
  --brain rescan-demo
```

Expected result: nonzero exit, the added canonical group is named, the error states that no destination mutation was made, and the journal remains byte-identical.

Remove the added file, replace `report.csv` at the same path, and rerun:

```bash
rm /workspace/xerj-rescan-demo/corpus/report-q2.csv
printf 'quarter,revenue\nQ1,110\n' > /workspace/xerj-rescan-demo/corpus/report.csv

target/debug/xerj autoindex /workspace/xerj-rescan-demo/corpus \
  --url http://localhost:9200 \
  --state-dir /workspace/xerj-rescan-demo/state \
  --prefix rescan-demo \
  --brain rescan-demo
```

Expected result: the full digest change is detected and the planned file is cleaned and republished.

Attempting `--fresh` with the existing state is refused:

```bash
target/debug/xerj autoindex /workspace/xerj-rescan-demo/corpus \
  --url http://localhost:9200 \
  --state-dir /workspace/xerj-rescan-demo/state \
  --prefix rescan-demo \
  --brain rescan-demo \
  --fresh
```

# Validation

- `xerj-autoindex`: 219 passed, 0 failed
- focused `xerj brain`: 5 passed, 0 failed
- ONNX embedding-identity recovery tests: 8 passed, 0 failed
- scoped Clippy for `xerj-autoindex`, `xerj-server`, and `xerj-engine`: passed with `-D warnings`
- `cargo fmt --all --check`: passed
- `git diff --check`: passed
- evaluation script syntax: passed
- ES-YAML conformance on the current 200-file suite: 1,365 passed, 0 failed, 3 skipped
- two independent code/documentation reviews: GO after fixing lock order, namespace isolation, shared-catalog wording, and executable recovery guidance

Full conformance evidence is recorded at `/workspace/north-star-evidence/route1-incremental-sync-esyaml-2026-08-03/runner.log` in the development environment.

# Scope and follow-up

This PR deliberately refuses unsupported membership changes rather than implementing a partial approximation.

The next contribution should preserve the durable plan and transactionally reconcile added, changed, and deleted content across data, alias, graph, and catalog artifacts. It should advance the plan only after durable publication and retain enough intent to replay safely after a crash.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,22 +7,33 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-### Changed — fail-closed autoindex restart semantics
+### Changed — autoindex refuses reruns that would strand documents
 
-- **`autoindex --fresh` no longer discards a durable resume plan.** A durable
-  plan contains the alias, path, graph, and stale-record knowledge needed for
-  safe reconciliation. `--fresh` is accepted only when the selected state
-  directory has no durable plan and never cleans the destination. An
-  independent rebuild must use a new state directory, new prefix, and new brain
-  namespace when graph detection is enabled (or disable graph writes), be
-  validated, and only then replace the old reader target. The shared catalog
-  and old target require explicit, validated cleanup. Membership additions and
-  removals are refused; same-path replacement remains supported.
+- **Deleting an indexed file and rerunning `xerj autoindex` now fails instead
+  of exiting 0.** Nothing in the pipeline removes the documents, aliases,
+  graph edges or catalog entries that the deleted file published, so a rerun
+  that ignored the deletion left them live and searchable with no source file
+  behind them. The rerun is now refused before any remote call other than the
+  endpoint-readiness ping: no mapping, delete-by-query, bulk, refresh, graph
+  or catalog write is attempted, and the journal is not appended to. The error
+  names every removed file and its content key and gives three recovery
+  routes — restore the files and rerun, rebuild in place by deleting the named
+  indices and the state directory, or rebuild under a new state directory,
+  prefix and brain. `--fresh` is refused for the same case, because it does
+  not delete those documents either. `--json` emits the same facts as
+  `xerj.autoindex.unsupported_sync_delta.v1` on stdout with exit 1.
+- **Files added after a plan was frozen are now called out on stderr.** They
+  are still not indexed by a rerun that resumes an existing plan — the plan is
+  a crash-resume boundary, not a folder-sync generation — but the run now
+  lists them, records them as skipped (exit 3, completed-with-junk) and points
+  at `--fresh`, which rebuilds the plan in place and picks them up. Adding or
+  changing files and rerunning keeps working; only removals are refused.
 - **`xerj brain` no longer turns an absent or zero node-count probe into an
   automatic reset.** Journal/server disagreement now fails with the journal,
-  URL, prefix, and brain identity plus recovery instructions. Probe transport
-  and malformed-response failures remain errors instead of being reported as
-  an empty destination.
+  URL, prefix, and brain identity plus recovery instructions, including the
+  explicit `--fresh` rerun for a genuinely wiped data directory. Probe
+  transport and malformed-response failures remain errors instead of being
+  reported as an empty destination.
 
 ## [1.0.0-rc.10] - 2026-08-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed — fail-closed autoindex restart semantics
+
+- **`autoindex --fresh` no longer discards a durable resume plan.** A durable
+  plan contains the alias, path, graph, and stale-record knowledge needed for
+  safe reconciliation. `--fresh` is accepted only when the selected state
+  directory has no durable plan and never cleans the destination. An
+  independent rebuild must use a new state directory, new prefix, and new brain
+  namespace when graph detection is enabled (or disable graph writes), be
+  validated, and only then replace the old reader target. The shared catalog
+  and old target require explicit, validated cleanup. Membership additions and
+  removals are refused; same-path replacement remains supported.
+- **`xerj brain` no longer turns an absent or zero node-count probe into an
+  automatic reset.** Journal/server disagreement now fails with the journal,
+  URL, prefix, and brain identity plus recovery instructions. Probe transport
+  and malformed-response failures remain errors instead of being reported as
+  an empty destination.
+
 ## [1.0.0-rc.10] - 2026-08-03
 
 ### Security
@@ -82,7 +99,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   reaching the HNSW graph. Each arrived with a standalone harness and honest
   caveats. They are tracked for the following release rather than rushed into
   this one.
-
 
 ### Changed — runtime-field types (can break existing mappings)
 
--- a/demo/usecases/autoindex/RECIPE_DRAFT.md
+++ b/demo/usecases/autoindex/RECIPE_DRAFT.md
@@ -203,13 +203,11 @@ done in 0.1s — 3 datasets, 5801 records live, 0 junk records, 1 junk/skipped f
 just polite re-runs: in the robustness evaluation, a `kill -9` at 18 s into a
 200 MB file followed by a re-run converged to **byte-identical counts across
 all 9 datasets** — no duplicates. `xerj autoindex status` shows the journal
-and live index counts. `--fresh` starts without resume state only when the
-selected state directory has no durable plan; it never removes stale
-destination records. For an independent rebuild, use a new `--state-dir`, new
-`--prefix`, and new `--brain` when graph detection is enabled (or add
-`--no-graph`). Validate before switching readers. The shared
-`autoindex-catalog` and old target require explicit, validated cleanup. Added
-or removed content groups are refused; same-path replacement remains supported.
+and live index counts; `--fresh` ignores the journal and rebuilds the plan in
+place (ids stay idempotent), which is how a rerun picks up files added since
+the last one. Deleting an indexed file is refused before any destination
+change — nothing here removes the documents it already published — and the
+error names what is gone plus the recovery routes.
 
 ## Reproduce it yourself
 
--- a/demo/usecases/autoindex/RECIPE_DRAFT.md
+++ b/demo/usecases/autoindex/RECIPE_DRAFT.md
@@ -203,8 +203,13 @@ done in 0.1s — 3 datasets, 5801 records live, 0 junk records, 1 junk/skipped f
 just polite re-runs: in the robustness evaluation, a `kill -9` at 18 s into a
 200 MB file followed by a re-run converged to **byte-identical counts across
 all 9 datasets** — no duplicates. `xerj autoindex status` shows the journal
-and live index counts; `--fresh` ignores the journal and restarts (ids stay
-idempotent).
+and live index counts. `--fresh` starts without resume state only when the
+selected state directory has no durable plan; it never removes stale
+destination records. For an independent rebuild, use a new `--state-dir`, new
+`--prefix`, and new `--brain` when graph detection is enabled (or add
+`--no-graph`). Validate before switching readers. The shared
+`autoindex-catalog` and old target require explicit, validated cleanup. Added
+or removed content groups are refused; same-path replacement remains supported.
 
 ## Reproduce it yourself
 
--- a/demo/usecases/autoindex/run-eval.sh
+++ b/demo/usecases/autoindex/run-eval.sh
@@ -2,9 +2,11 @@
 # End-to-end eval of `xerj autoindex` against a corpus folder.
 #
 # Boots a dedicated server on es_compat $PORT (rest PORT+100, grpc PORT+101),
-# runs autoindex over $CORPUS, proves same-state resume, then performs an
-# isolated data/graph rebuild under a new state directory, prefix, and brain.
-# The global autoindex-catalog remains shared and is excluded from comparison.
+# runs autoindex over $CORPUS, then re-runs twice to prove idempotency — the
+# resume path (all files done) and a full --fresh re-extract — and requires the
+# per-index doc counts to be byte-identical to run 1. Ends with the data map.
+# The shared autoindex-catalog is excluded from the comparison: every run
+# appends its own run record, so a growing catalog is the design, not drift.
 #
 # Nothing here is bound to one machine: every path is an env override with a
 # default under the repo or $TMPDIR, so the same script gates CI and drives a
@@ -33,20 +35,11 @@ CFG="$WORK/server-$PORT.toml"
 LOG="$WORK/server-$PORT.log"
 URL="http://localhost:$PORT"
 STATE="$WORK/state"
-REBUILD_STATE="$WORK/rebuild-state"
-PREFIX="ax-eval-a"
-REBUILD_PREFIX="ax-eval-b"
-BRAIN="eval-a"
-REBUILD_BRAIN="eval-b"
+PREFIX="ax-eval"
+BRAIN="eval"
 
 [ -x "$BIN" ] || { echo "xerj binary not found at $BIN"; exit 1; }
 [ -d "$CORPUS" ] || { echo "corpus folder not found at $CORPUS"; exit 1; }
-[ ! -e "$DATA" ] && [ ! -e "$STATE" ] && [ ! -e "$REBUILD_STATE" ] || {
-  echo "work directory already contains eval state: $WORK"
-  echo "choose a new XERJ_AUTOINDEX_WORK or explicitly remove the old isolated eval directory"
-  exit 1
-}
-
 mkdir -p "$WORK"
 cat >"$CFG" <<EOF
 [server]
@@ -85,10 +78,10 @@ if [ "$RC" != "0" ] && [ "$RC" != "3" ]; then
   FAIL=1
 fi
 
-# Name + doc count only. The _cat row also carries on-disk size, which an
-# isolated data/graph rebuild legitimately changes (segments are rewritten).
-# The shared global catalog is excluded because every run appends its own run
-# record — a growing catalog is the design, not drift. Columns of the _cat row:
+# Name + doc count only. The _cat row also carries on-disk size, which a
+# --fresh re-extract legitimately changes (segments are rewritten), and the
+# catalog is excluded outright because every run appends its own run record —
+# a growing catalog is the design, not drift. Columns of the _cat row:
 # green(1) open(2) name(3) uuid(4) pri(5) rep(6) docs(7).
 counts() { curl -s "$URL/_cat/indices" | awk -v prefix="$1-" '$3 ~ ("^" prefix) {sub("^" prefix, "", $3); print $3, $7}' | sort; }
 
@@ -104,16 +97,16 @@ if [ "$RC" != "0" ] && [ "$RC" != "3" ]; then
   FAIL=1
 fi
 
-echo "=== isolated data/graph rebuild reproducibility (new state, prefix, and brain; shared catalog excluded) ==="
-time "$BIN" autoindex "$CORPUS" --url "$URL" --state-dir "$REBUILD_STATE" --prefix "$REBUILD_PREFIX" --brain "$REBUILD_BRAIN"
+echo "=== idempotency: re-run with --fresh (full re-extract, idempotent ids) ==="
+time "$BIN" autoindex "$CORPUS" --url "$URL" --state-dir "$STATE" --prefix "$PREFIX" --brain "$BRAIN" --fresh
 RC=$?
 if [ "$RC" != "0" ] && [ "$RC" != "3" ]; then
-  echo "isolated data/graph rebuild exited $RC (expected 0 or 3)"
+  echo "--fresh re-extract exited $RC (expected 0 or 3)"
   FAIL=1
 fi
 
-counts "$REBUILD_PREFIX" > "$WORK/counts-run3.txt"
-echo "=== normalized data-index count diff original vs isolated data/graph rebuild (shared catalog excluded; must be empty) ==="
+counts "$PREFIX" > "$WORK/counts-run3.txt"
+echo "=== doc-count diff run1 vs run3 (must be empty) ==="
 cat "$WORK/counts-run3.txt"
 if diff "$WORK/counts-run1.txt" "$WORK/counts-run3.txt"; then
   echo "IDENTICAL COUNTS ✓"
--- a/demo/usecases/autoindex/run-eval.sh
+++ b/demo/usecases/autoindex/run-eval.sh
@@ -2,9 +2,9 @@
 # End-to-end eval of `xerj autoindex` against a corpus folder.
 #
 # Boots a dedicated server on es_compat $PORT (rest PORT+100, grpc PORT+101),
-# runs autoindex over $CORPUS, then re-runs twice to prove idempotency — the
-# resume path (all files done) and a full --fresh re-extract — and requires the
-# per-index doc counts to be byte-identical to run 1. Ends with the data map.
+# runs autoindex over $CORPUS, proves same-state resume, then performs an
+# isolated data/graph rebuild under a new state directory, prefix, and brain.
+# The global autoindex-catalog remains shared and is excluded from comparison.
 #
 # Nothing here is bound to one machine: every path is an env override with a
 # default under the repo or $TMPDIR, so the same script gates CI and drives a
@@ -33,9 +33,19 @@ CFG="$WORK/server-$PORT.toml"
 LOG="$WORK/server-$PORT.log"
 URL="http://localhost:$PORT"
 STATE="$WORK/state"
+REBUILD_STATE="$WORK/rebuild-state"
+PREFIX="ax-eval-a"
+REBUILD_PREFIX="ax-eval-b"
+BRAIN="eval-a"
+REBUILD_BRAIN="eval-b"
 
 [ -x "$BIN" ] || { echo "xerj binary not found at $BIN"; exit 1; }
 [ -d "$CORPUS" ] || { echo "corpus folder not found at $CORPUS"; exit 1; }
+[ ! -e "$DATA" ] && [ ! -e "$STATE" ] && [ ! -e "$REBUILD_STATE" ] || {
+  echo "work directory already contains eval state: $WORK"
+  echo "choose a new XERJ_AUTOINDEX_WORK or explicitly remove the old isolated eval directory"
+  exit 1
+}
 
 mkdir -p "$WORK"
 cat >"$CFG" <<EOF
@@ -66,7 +76,7 @@ curl -sf "$URL/" >/dev/null || { echo "server failed to boot"; tail -20 "$LOG"; 
 FAIL=0
 
 echo "=== autoindex run ($CORPUS) ==="
-time "$BIN" autoindex "$CORPUS" --url "$URL" --state-dir "$STATE" --fresh
+time "$BIN" autoindex "$CORPUS" --url "$URL" --state-dir "$STATE" --prefix "$PREFIX" --brain "$BRAIN"
 RC=$?
 echo "exit code: $RC"
 # 0 complete, 3 completed-with-junk (unreadable files recorded, never fatal).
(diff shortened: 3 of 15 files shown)


xerj-org (changes requested):
Thanks for this, @buger — and thanks in particular for going after the *right* bug. You are correct that `main` is silently wrong here: a rerun over a changed folder skips added files and leaves documents for deleted files live, with exit 0. That is a genuine correctness hole and it deserved to be found.

I am requesting changes because of what the fix costs, not because of what it fixes. As written it converts a partial-correctness bug into a total-availability one for the tool's primary workflow, and it closes every in-place recovery route at the same time.

## What I actually observed

I did not want to take either of our word for the blast radius, so I wrote one test against `origin/main`: index a folder containing `first.csv`, add `second.csv`, rerun plainly, then rerun with `--fresh`.

On `main` it passes. The plain rerun returns `Ok` (skipping the new file — the bug you are fixing), and `--fresh` absorbs it, ending with documents at both `first.csv` and `second.csv`.

Copied byte-for-byte onto this branch, it fails — and it fails **earlier than I expected**, at the plain rerun, before the `--fresh` leg is ever reached:

```
thread '...baseline_fresh_reindex_absorbs_an_added_file' panicked at
crates/xerj-autoindex/src/failure_resume_http_tests.rs:1086:5:
assertion failed: run_index(config.clone()).is_ok()
```

So after this PR, any rerun of `xerj autoindex` over a folder that gained or lost a single file exits 1. That is the documented headline workflow ("point it at any folder … zero config").

## Blocking items

**1. `xerj brain` is bricked on any note added after the first run.** `brain.rs`'s `index_cfg` passes `state_dir: None`, so the default `~/.xerj/autoindex/<hash>` plan always exists, and `fresh: cfg.fresh` feeds the same gate. Write one new note, rerun `xerj brain ~/notes`, hard error. The wiped-data-dir auto-recovery in `brain.rs` is also replaced by a `bail!` whose `!fresh` escape is unreachable, because `--fresh` is refused upstream.

**2. `--fresh` is functionally dead.** In `lib.rs` the delta is computed and then `if cfg.fresh { return Err(delta.into_fresh_error()) }` fires whenever a plan exists, regardless of whether the delta is empty. The help text concedes it is only accepted when there is nothing to be fresh from.

**3. No in-place recovery remains.** Every guidance string — help text, CLI errors, the JSON `recovery.exact_rebuild` field, CHANGELOG, recipes, `llms-full.txt`, the second-brain design doc — points at "new `--state-dir`, new `--prefix`, new `--brain`, validate, then switch readers." For one added file, the supported answer is re-extract and re-embed the whole corpus into a differently-named index and manually migrate every reader. Nothing mentions that deleting the state dir is an option; the docs steer away from it.

**4. Two existing tests were flipped to assert the regression.** `deleting_an_entire_duplicate_group_strands_no_pending_replacement` went from `assert_eq!(run_index(...), 0)` to asserting refusal, and `a_planned_key_never_gains_two_owners_when_old_content_moves_paths` went from publishing the rewritten content to asserting `value == "original"` is left stale on the server behind an error. When a fix requires flipping tests that encoded working behaviour, that is usually the signal to re-scope.

## Honesty items (repo rule: honest claims only)

- The body says "Replacing the contents of an existing path is detected and republished under the planned identity." That holds only when no duplicate of the old content exists elsewhere; the PR's own modified test shows a same-path replacement being refused with the old content left live.
- Neither the body nor the CHANGELOG states plainly that a plain, non-`--fresh` rerun over a changed folder now fails. The CHANGELOG frames it as `--fresh` semantics plus "membership additions and removals are refused," which reads as a narrow guard rather than the ordinary rerun path.
- The validation evidence cited ("two independent code/documentation reviews: GO", conformance log at `/workspace/north-star-evidence/…`) lives outside the repo, so I could not check it.

CI being 9/9 green is not reassurance here: no job reruns autoindex over a mutated folder, and `demo/usecases/autoindex/run-eval.sh` — which previously used the `--fresh` idempotency leg as its correctness proof — was rewritten in this same PR to stop exercising it. I did reproduce your `cargo test -p xerj-autoindex` result on the branch: 219 passed, 0 failed.

## What would make this mergeable

Any one of these works for me:

1. **Preferred:** land the transactional add/change/delete reconciliation this PR defers to follow-up, so the correct cases succeed instead of being refused.
2. **Minimum:** keep the fail-closed gate only for the genuinely unsafe case (vanished groups leaving stale live records), let additions through, and keep `--fresh` working as an in-place full rebuild — `main` demonstrably converges under it.
3. **Fastest path to landing something today:** split the honest `xerj brain` live-node probe into its own PR. The absent / authoritative-zero / malformed / transport-failure distinction is correct, well tested, and independently valuable, as is the typed `xerj.autoindex.unsupported_sync_delta.v1` error with its stdout/stderr routing. I would merge that on its own now.

Two smaller notes for whichever direction you take: the branch is behind main (base `76d8bb0`) and will need a rebase; and `live_node_docs` now returns `Err` when a brain meta doc exists but has no `nodes_index`, where it previously returned 0 — that hard-fails `xerj brain` for any brain not written by current autoindex.

The instinct here is right and the fail-closed machinery is well built. It just needs to fail closed on the unsafe case only.

xerj-org:
@buger — I took the review's "minimum" option and landed it on top of your commit rather than asking you to redo the work. Your commit is untouched; there is one commit on top of it.

**Where it is:** `maintainerCanModify` is off for this PR, so I could not push to `probelabs:fix/autoindex-incremental-sync-1254`. The result is pushed to this repo as [`pr156-review-followup`](https://github.com/xerj-org/xerj/tree/pr156-review-followup) — `76d8bb0` → your `2a55358` → `66e91d7`. If you want it in this PR, `git pull` that branch into yours (or flip "allow edits by maintainers" and I will push it here).

## What changed

**The gate now fires only on removals.** A rerun is refused when a canonical content group the plan published no longer exists in the folder — nothing in the pipeline deletes the documents, aliases, graph edges or catalog entries that file wrote, so ignoring the deletion leaves them live with no source behind them. That is the case your PR was right to fail closed on. `--fresh` is refused for it too, because discarding the plan does not delete those documents either.

**Additions are reported, not fatal.** They are still not absorbed by a frozen plan, but the run now names them on stderr, records them as skipped (exit 3), and points at `--fresh`. `--fresh` is a plain in-place plan rebuild again, so the add-then-rerun-then-`--fresh` workflow converges as it did on main. `xerj brain` follows from that: a note added after the first run no longer errors.

**The refusal message names what is unsupported and what to do.** It lists every removed file with its content key, states that no remote mutation was made, and gives three routes ordered by cost: restore the files and rerun; rebuild in place by deleting the named indices and the state directory (with the honest note that edges the removed file taught stay live in the named edges index); rebuild isolated under a new `--state-dir`/`--prefix`/`--brain`. `--json` keeps your typed `xerj.autoindex.unsupported_sync_delta.v1` payload with stdout/stderr routing — that part I kept as you wrote it, it is good.

**In-place recovery wording is back** in the journal-corruption, key-collision and journal-mismatch errors, in `--fresh` help, the recipes, `llms-full.txt` and the second-brain docs, and the `--fresh` idempotency leg is back in `demo/usecases/autoindex/run-eval.sh` (keeping the per-run exit-code checks you added).

**`xerj brain`:** the live-node probe keeps your absent / authoritative-zero / malformed / transport-failure distinction, which is the part of this PR I most wanted to keep. One fix: a brain meta doc with no `nodes_index` is treated as absent evidence (`Ok(None)`) instead of an error, so brains written before that field existed still work. The disagreement message now also names the `--fresh` rerun for a genuinely wiped data directory, next to the isolated rebuild.

## What I did not do

- **Full add/change/delete reconciliation** is still follow-up work, as your PR said. This only narrows the refusal.
- **`deleting_an_entire_duplicate_group_strands_no_pending_replacement` stays flipped** to assert refusal. That test deletes an entire content group, which is exactly the unsafe case, so the refusal is correct there; I kept its original invariants (no stranded replacement intent, no journal growth) asserted alongside. The other flipped test, `a_planned_key_never_gains_two_owners_when_old_content_moves_paths`, is an addition and is restored to its pre-PR assertions.
- **PR body / CHANGELOG honesty:** I rewrote the CHANGELOG entry to state plainly what fails and what does not. The PR body still describes the wider contract — worth updating if you pull this in.

## Verification

Measured on `66e91d7`:

- `cargo test -p xerj-autoindex`: 221 passed, 0 failed.
- `cargo test -p xerj-server --bins brain::`: 5 passed, 0 failed.
- `cargo clippy -p xerj-autoindex --all-targets -- -D warnings` and the same for `xerj-server`: clean. `cargo fmt --all` clean.
- Mutation check on the gate: with `UnsupportedInventoryDelta::refuses()` forced to `false`, `completed_plan_rejects_vanished_content_group_before_remote_mutation` fails with `called `Result::unwrap_err()` on an `Ok` value: 0` — i.e. the deletion rerun silently succeeds, which is the bug — and `only_a_removed_content_group_refuses_a_rerun` fails on `assertion failed: UnsupportedInventoryDelta::between(&[], &[], &plan).refuses()`. Both pass with the gate restored.
- Branch merges cleanly into `main` as of `2e6fccd`; no rebase needed.

Thanks again for finding this one — the silent-skip on rerun was a real hole and the fail-closed machinery you wrote is what made narrowing it a small change.

xerj-org:
@buger — thanks for pulling the follow-up in and for `bace26c`. Naming the new-note bypass as a predicate and asserting it directly is a better guard than the implicit `records_total == 0` check it replaces; that is the branch most likely to be broken by future probe hardening, and now it cannot be broken silently.

I re-reviewed the branch at `bace26c` against the original review. The four blocking items are addressed and I have no further objections to the code. One commit on top, for the last gap.

**Where it is:** `maintainerCanModify` is still off for this PR, so a push to `probelabs:fix/autoindex-incremental-sync-1254` is rejected. The commit is on this repo as [`pr156-rerun-gate-ci`](https://github.com/xerj-org/xerj/tree/pr156-rerun-gate-ci) — your three commits untouched, `87ee42c` on top. `git pull` it into your branch, or flip "allow edits by maintainers" and I will push it here.

## What the commit adds

Nothing in CI reruns either tool over a folder or a server that *changed* — which is exactly where the silent-skip bug lived. `run-eval.sh` reruns autoindex over an unchanged corpus, and the brain harness never wipes a data directory out from under a resume journal. The refusals have good unit and in-process HTTP tests; this is the missing end-to-end gate, two phases on the harnesses that already exist:

- **autoindex:** remove `logs/app.log` from the corpus `run-eval.sh` just indexed, rerun, require a failure that *names the removed file*. Then put the file back and rerun — recovery (1) from your own refusal message — and require it through again, so the gate is provably a stop and not a brick.
- **brain:** kill the demo brain's server, delete its data directory (the resume journal lives outside it, under `~/.xerj/autoindex/<hash>`, and survives), rerun. Require the journal/server disagreement, then require the `--fresh` that message names to actually bring the brain back.

Both assert on the refusal *text*, not on a nonzero exit. A run that fails for an unrelated reason also exits nonzero, and the pre-fix brain path exits 1 with "nothing indexable" — exit code alone cannot tell those apart from a refusal.

## Verification, measured on `87ee42c`

- `cargo test -p xerj-autoindex`: 221 passed, 0 failed.
- `cargo test -p xerj-server --bins brain::`: 5 passed, 0 failed.
- `cargo fmt --all -- --check`: clean.
- Full `.github/scripts/usecase-smoke.sh` against the release binaries: every phase PASS, exit 0 — including the new phase 4 (`a rerun after a removal refuses (exit 1) and names the removed file`) and phase 5 (`the --fresh rebuild the refusal names recovers the brain (exit 0)`).
- Mutation check, so the new phases are not decoration. With `UnsupportedInventoryDelta::refuses()` forced to `false` and the release binaries rebuilt, phase 4 fails with `::error::expected a nonzero refusal naming logs/app.log, got exit 0` and the script exits 1 — that exit 0 *is* the original silent skip. Phase 5 still passes under that mutation, correctly, since it gates the brain live-node probe rather than the inventory delta. The same mutation fails `completed_plan_rejects_vanished_content_group_before_remote_mutation` with `called Result::unwrap_err() on an Ok value: 0`, and `only_a_removed_content_group_refuses_a_rerun` on `assertion failed: UnsupportedInventoryDelta::between(&[], &[], &plan).refuses()`. Restored, everything is green again.
- The branch merges into `main` with no conflicts.

## One thing left, and it is not code

The PR title and body still describe the original shape of this change: `--fresh` "accepted only when the selected state directory has no durable plan", "adding or removing a canonical content group is detected and rejected", the single-preflight-lock section. None of that is what the branch does now — additions are reported and skipped rather than rejected, `--fresh` is a plain in-place plan rebuild, and only removals are refused. The Validation section has the same problem: the 219-test count and the ES-YAML run were measured on the first version of the tree.

I have not touched your description. It does need updating before this merges — "honest claims only" is the repo's first rule, and a squash merge would carry that text into `main`'s history as the record of what shipped. The CHANGELOG entry on the branch already says the accurate thing, so it is mostly a matter of bringing the body in line with it.

Thanks again for finding this one and for staying with it through two rescopes.

xerj-org:
Status at head `bace26c`: the code is in good shape and I have no further objections to it. Three things stand between it and a merge, and all three need you — I have no push access here (`maintainerCanModify` is off, and a push to `probelabs:fix/autoindex-incremental-sync-1254` returns 403).

**1. It no longer merges cleanly.** This is my doing, not yours — I merged #160 a few minutes ago (`96f6d1a`), and this branch now conflicts with `main` in two files:

- `CHANGELOG.md` — both sides added entries in the same region; a normal adjacent-entry conflict.
- `engine/crates/xerj-autoindex/src/state.rs` — overlaps with the embedding-identity work that came in with #160.

Everything else in the diff auto-merges. A rebase onto current `main` should be short. I deliberately did not force anything.

**2. The CI rerun-gate commit is still off the branch.** The earlier review's point that no job reruns either tool over a mutated folder is unaddressed on the branch itself — both refusals are still gated only by unit and in-process HTTP tests. The commit that fixes it is `87ee42c` on `xerj-org/xerj` branch `pr156-rerun-gate-ci` (`0bea6bb` → `b0ed1c8` → `bace26c` → `87ee42c`). It adds two phases to `.github/scripts/usecase-smoke.sh`: phase 4 removes `logs/app.log` from the corpus autoindex just indexed, reruns, requires a failure naming the removed file, then restores it and requires the rerun to pass — so the gate is provably a stop and not a brick; phase 5 does the equivalent for the brain and then requires the `--fresh` that the refusal names to actually rebuild. Both assert on the refusal text rather than a nonzero exit, because the pre-fix brain path also exits 1.

Please `git pull` it, or enable "allow edits by maintainers" and I will push it. The PR is mergeable without it — this is a coverage gap, not a correctness one.

**3. The PR title and body still describe the first version of this change.** They say `--fresh` is accepted only with no durable plan, that additions are rejected, and they include the single-preflight-lock section; the Validation section quotes counts measured on that tree. The branch no longer behaves that way — additions are reported and skipped, `--fresh` is a plain in-place plan rebuild, and only removals are refused. Since this will be squash-merged, that body becomes `main`'s record of what shipped, and under "honest claims only" it should not. The CHANGELOG on the branch is already accurate, so this is only a matter of bringing the description in line. I have deliberately not rewritten your prose.

On the substance, for the record — the refusal does fail closed with an actionable message and is genuinely covered. The autoindex gate runs before the journal is opened and before any remote call other than the readiness ping; it names each removed file with its content key, states that no remote mutation was made, and gives three recoveries in cost order, including the honest note that edges taught by a removed file stay live. `--json` emits the typed `xerj.autoindex.unsupported_sync_delta.v1` with exit 1. The brain refusal names the journal path, URL, prefix and brain, and names both the in-place `--fresh` rerun and the isolated rebuild.

The review state is still CHANGES_REQUESTED from the original round; I have left it as-is rather than clearing it for you, but treat the code objections as withdrawn.

xerj-org:
@buger — first, the good part: everything I raised as a code objection in the earlier rounds is closed. The gate fails closed before any remote mutation, the refusal is typed and exits 1, the new-note probe bypass is pinned as a predicate at `bace26c`, and the four original blockers are gone. Thanks for staying with this through two rescopes 🙏

Then the part I owe you an apology for. I re-measured `--fresh` on this branch with clean, separate `CARGO_TARGET_DIR`s, and my earlier all-clear was wrong — an earlier run shared a target dir and gave me a false reading, so I rebuilt from scratch to confirm. `--fresh` has two regressions on this branch that I signed off on last round. That is my miss, not a new demand. Verdict is back to **hold**, and here is everything, with what would make each one pass.

---

### 1. `--fresh` is refused after any same-path content *edit* (regression, introduced by `b0ed1c8`)

`engine/crates/xerj-autoindex/src/lib.rs:739-741` — under `cfg.fresh` the comparison uses `inventory.keys.clone()`, i.e. the raw post-edit content keys. Those are compared at `lib.rs:760` against a plan that permanently retains the *pre-edit* key: `lib.rs:865` writes `inventory.keys[index] = planned_key`, so the planned key never follows the edit. The pre-edit key is therefore "vanished", `refuses()` at `lib.rs:624` returns true, and the run dies.

Measured both ways, clean target dirs: edit a file in place, rerun with `--fresh` → `ERROR` at `bace26c3`; the same sequence on `main` at `96f6d1a` → `Ok(0)`. So this is a regression, not a pre-existing limitation.

What would make it pass: when `cfg.fresh` is set the delta has to be computed over *paths* that genuinely left the folder, not over content keys the plan is guaranteed to hold stale. An edited file is still present; it should never appear in `vanished_content_groups`.

There is also no test for edit-then-`--fresh`. `a_rerun_after_an_added_file_succeeds_and_fresh_absorbs_it` only adds a file, and `fresh_cannot_erase_the_plan_and_bypass_the_removal_gate` (`failure_resume_http_tests.rs:399`) only removes and adds. Both assert the new behaviour without ever exercising the path that broke — a test that edits a file's bytes at the same path and then runs `--fresh` is the one that fails today.

### 2. `--fresh` no longer clears a journal root/url/prefix mismatch (regression, introduced by `0bea6bb`, not repaired by `b0ed1c8`)

`engine/crates/xerj-autoindex/src/state.rs:296` calls `read_plan_for_preflight`, which hits `anyhow::ensure!(recorded_root == root && ...)` at `state.rs:238`. That fires *inside preflight* — before the fresh deletion at `state.rs:347` ever runs. So `--fresh` can no longer rescue a journal whose identity moved.

Measured: same `--state-dir` with a changed `--prefix` plus `--fresh` → `ERROR` at head, `Ok(0)` on `main`.

Two things compound it: the preflight message that now fires offers no recovery at all, and the message that *does* name `--fresh` (`state.rs:383-386`, "or `--fresh` to rebuild the plan in place") is now unreachable dead code.

What would make it pass: perform the fresh journal deletion before the identity check, or skip the identity `ensure!` when `fresh` is set — the recorded identity of a journal you are about to delete cannot be a precondition for deleting it.

### 3. The refusal message contradicts itself (repo rule: honest claims only)

`lib.rs:519-560`. In the measured edit-plus-add case, one path appears on both sides of the same message: `Removed content groups [a.csv (axf2-49c7...)]` and `Also present but not in the frozen resume plan ... [a.csv (axf2-443b...)]`. The surrounding prose then asserts the file `no longer exist[s] in the folder` and that `their documents are still live in the destination` — while `a.csv` is sitting right there and its documents were just republished.

What would make it pass: the message needs to separate "this content key is no longer present" from "this file is no longer present", and must not claim a file is gone when the same relative path is on both lists. Fixing item 1 removes most of these cases; the wording still needs to be safe for the ones that remain.

### 4. CHANGELOG claim is false for the changed-file case

`CHANGELOG.md` around line 29 states that "Adding or changing files and rerunning keeps working; only removals are refused", and that the run "points at `--fresh`, which rebuilds the plan in place and picks them up". The second half is false for a changed file, per the measurement in item 1. Flagging this specifically because the CHANGELOG was the one artifact I certified as accurate last round, so it should not be assumed still-good after item 1 is fixed.

### 5. `--fresh` advice is wrong for a legacy journal (inference — read, not executed)

`lib.rs:395` now advises `Deleting the journal at {} (or rerunning with --fresh) also clears the collision`. For a legacy journal the `--fresh` half looks wrong to me: legacy plan keys are `{16hex}-{sizehex}` (`ids.rs:25`) while inventory keys are `axf2-{32hex}-{sizehex}` (`content.rs:134`), and the plan permanently retains the legacy keys — so `--fresh` would report every file in the corpus as removed. The same reasoning applies to `use --fresh after verifying the existing index` at `lib.rs:345` and `lib.rs:355`. I derived this by reading rather than running it, so treat it as inference; worth a legacy-journal + `--fresh` test either way.

### 6. Still no CI job that reruns either tool over a *mutated* folder

Unchanged from the last round: `.github/scripts/usecase-smoke.sh` at head is phases 0-3 only, and `87ee42c` (phases 4 and 5, on `xerj-org/xerj` branch `pr156-rerun-gate-ci`) is still not on this branch. Both refusals remain gated only by unit and in-process HTTP tests — which is exactly the coverage shape that let items 1 and 2 through.

### 7. The branch no longer merges

`git merge-tree origin/main bace26c3` conflicts in `CHANGELOG.md` and `engine/crates/xerj-autoindex/src/state.rs`. Merge base is `2e6fccd`; `main` is now `96f6d1a` (#160). Still my doing, still a short rebase.

### 8. PR title and body still describe the first version

They still say `--fresh` is accepted only with no durable plan, that additions are rejected, they carry the single-preflight-lock section, and the Validation counts were measured on the 219-test tree. A squash merge writes that text into `main`'s history as the record of what shipped. Note that item 4 means the CHANGELOG can no longer just be copied across — it needs the same correction.

---

### Two notes, not objections

**`xerj brain` has no working in-place route left.** For a *removed* note the gate refuses (`lib.rs:760`) and also refuses `--fresh` for that same case (`lib.rs:755-762`), and `xerj brain` exposes no way to delete the `ax-*` indices the message tells you to delete. On `main` this exited 0. The prior review accepted that trade explicitly and I am not re-opening it as a code objection. I am only pointing out the combination: with the `--fresh` edit regression in item 1, `xerj brain` over a live notes folder currently has no working in-place route for either add-plus-edit or removal. Fixing item 1 restores one of the two.

**A pre-existing invariant hole this PR makes routine.** On `main` already: an added-but-skipped file gets a live catalog document written for it (`lib.rs:2052`, `all_junk.extend(new_unplanned.iter())`) yet never enters the durable plan. Delete that file later and the delta sees nothing vanished, so the stale catalog entry stays live forever — precisely the staleness the new gate refuses for *planned* junk (`failure_resume_http_tests.rs:468`, `deleted_planned_junk_fails_closed_before_catalog_can_stay_stale`). Not introduced by you; the newly-blessed add-then-rerun path just makes it the common case. Worth its own issue rather than more scope here.

---

On access: thank you for offering write access to `probelabs/xerj` — the invitation is still unaccepted on our side (our PAT gets a 403 accepting it), so we genuinely cannot push to your branch yet and I am not going to pretend otherwise. Please keep pushing as you have been, and the moment that access goes live we will happily pick up the rebase, the CI phases and the body rewrite ourselves so you do not have to carry all of it. 🚀

Items 1 and 2 are the blockers; the rest travel with them.

xerj-org:
Reviewed. The bug is real and the fail-closed gate is the right interim shape, but three things need fixing before this lands.

`--fresh` is no longer an escape hatch. `Journal::preflight` calls `read_plan_for_preflight` unconditionally, and that hard-errors on a malformed record, a root/url/prefix mismatch, or an undecodable plan — all before `open_after_preflight` reaches the `remove_file`. Before this PR, `--fresh` deleted the journal without parsing it. Two consequences: your own message at state.rs:224 advertises `--fresh` as the recovery from exactly the error that now blocks it, and `Journal::open` decodes the plan tolerantly while the preflight turns the same condition into a hard `?`. Make the preflight no more fatal than `open`.

Unbounded error rendering. The `render` closure joins every vanished entry into one String, no cap — even though this same PR caps its "appeared after the plan was frozen" block at `.take(10)`. On a real journal here (82,551 files) that's ~7 MB in one String plus the intermediate Vec. An unmounted bind mount makes every group vanish. Apply the same cap.

`xerj brain` takes the wrong branch on the case its new message is written for. `live_node_docs` propagates via `?` on a non-2xx, so if the meta doc survives but the nodes index was deleted, the probe 404s and never reaches `journal_server_disagreement()`. User gets a raw HTTP 404 instead of the recovery text.

Also the PR description doesn't match the code: additions are not rejected (lib.rs:624), and your own test `a_rerun_after_an_added_file_succeeds_and_fresh_absorbs_it` asserts that. And `--fresh` after an in-place edit is refused because the old digest key vanishes, which contradicts the CHANGELOG line.

Branch is 22 commits behind and conflicting on CHANGELOG.md and state.rs. Rebase and re-run CI, the green run is from 2026-08-04 against an old base.

Quickwit is the precedent worth reading for the follow-up you mention: mark-then-sweep with grace periods and bounded batches rather than refuse. Fail-closed is defensible until then.

xerj-org:
Sequencing note after today's sweep: we'd like to land this one FIRST, then have #162 rebase on top of it (both carry the same unsupported-reconciliation refusal machinery — landing both as-is would apply it twice through different lineages). What it needs: the three fixes from the 2026-08-07 review (notably: --fresh is no longer an escape hatch because Journal::preflight hard-errors before open_after_preflight can remove_file) plus a rebase onto current main (it went stale when #160 merged). Fork has maintainer edits off, so the push has to come from your side — happy to re-review the moment it's up.

xerj-org:
Status update, and one thing we need from you.

This branch is on a fork with `maintainerCanModify=false`, so nobody here can push the rebase to it — every attempt 403s. Rather than let the work go stale, we rebased it onto current `main`, resolved the conflicts, and addressed the outstanding review items (the `--fresh` escape hatch that `Journal::preflight` had made unreachable, and the unbounded error rendering in `UnsupportedInventoryDeltaError::fmt`). That is open as **#254**, and your original commits are carried across with their authorship intact.

**It cannot merge until you sign the CLA.** Because the commits are yours, the CLA check on #254 reports you as an unsigned contributor — and we are not going to re-author your work to route around that.

It is a one-time thing per contributor, not per pull request: read [CLA.md](https://github.com/xerj-org/xerj/blob/main/CLA.md), then open a small PR adding `buger` to [`.contributors`](https://github.com/xerj-org/xerj/blob/main/.contributors). That PR is your signature, since it comes from your own account. @Vinz2168 just did exactly this in #246 if you want a template.

Thanks for the patch — it is good work and we would like to land it.

Issue #157, opened 2026-08-03
Publication test barriers block tokio workers with unbounded waits, stalling CI


xerj-org:
## What happens

`Build + Test` intermittently stalls, with the runner reporting lines like:

```
test index::merge_publication_transaction_tests::cancellation_inside_guarded_repoint_fails_closed_and_recovery_restores_inputs has been running for over 60 seconds
```

The test is not slow. On a 32-core box it finishes in **0.07s**, and even pinned to two cores with `--test-threads=2` the whole module finishes in **0.04s**. What it is, is *unbounded*: under contention it can wait forever rather than fail, which on CI shows up as a job that hangs until the workflow times out and takes the release with it.

## Why

The publication test barrier blocks a tokio worker thread on purpose. `publication_test_point` (`engine/crates/xerj-engine/src/index.rs:5134`) is a synchronous function invoked from async code, and its own comment says the quiet part out loud:

```rust
// Clone the callback while holding the hook registry lock, then invoke
// it after releasing that lock. Test barriers may intentionally block.
```

The hooks then do exactly that:

```rust
index.set_publication_test_hook(Some(Arc::new(move |_, point| {
    if point == PublicationTestPoint::MergeAfterRepoint {
        entered_tx.lock().take().unwrap().send(()).unwrap();
        resume_rx.lock().take().unwrap().recv().unwrap();   // blocks a worker thread
    }
})));
```

Two consequences:

1. **Worker starvation.** These tests run on `#[tokio::test(flavor = "multi_thread", worker_threads = 4)]`. Each blocked barrier consumes one worker. CI runs `cargo test --release --workspace -- --test-threads=2` on a 2-core `ubuntu-latest` runner, so several multi-threaded runtimes with blocked workers contend for two real cores. That is where the minutes go.

2. **No bound anywhere.** Across `index.rs` there are **34 blocking `recv().unwrap()` calls against only 2 `recv_timeout`**, and `merge.await` / `read.await` in these tests have no timeout either. If a refactor ever stops the code reaching `MergeAfterRepoint`, `entered_rx.recv()` waits forever. The test does not fail, it hangs, and the failure surfaces as an unrelated CI timeout with no useful message.

10 tests install this style of hook.

## Why it matters beyond flakiness

A hanging test is strictly worse than a failing one. It costs a full CI slot, produces no diagnosis, and blocks a release for a reason nobody can read off the log. This one sat on the last rc.10 blocker (#151).

## Suggested direction

[...]

PR #160, opened 2026-08-03
Feat/embedding execution identity

buger:
## Summary

Semantic `autoindex` resumes now pin the embedding execution identity before creating an index or publishing a data bulk.

This prevents one autoindex namespace from silently mixing vectors produced by different embedding backends, models, tokenizers, dimensions, or semantic contracts after a server restart or configuration change.

There is no new setup step or separate command for users. The existing workflow stays:

```bash
xerj autoindex ./reports
```

When the generated plan contains semantic fields, `xerj autoindex` automatically calls an authenticated, privacy-safe server endpoint on the same configured XERJ URL:

```http
GET /v1/embedding/identity
```

This endpoint is an internal machine-readable contract between the existing autoindex client and XERJ server. Users do not need to invoke it. It remains directly inspectable for agents and diagnostics, but it is not part of the normal indexing procedure.

The response exposes only a versioned opaque identity, effective backend, dimensions, semantic contract, resumability, and a sanitized non-resumable reason. It does not expose credentials, remote provider URLs, model filenames, local paths, or raw configuration.

## User-visible behavior

For a semantic autoindex run, the existing command now performs these steps automatically:

1. Fetches `/v1/embedding/identity`.
2. Persists the identity in the autoindex journal before index creation.
3. Requires the same identity when resuming.
4. Refuses identity drift before another index-creation or data-bulk request.

For a non-semantic plan, autoindex skips the endpoint entirely. The only new behavior most users can observe is a safe, actionable failure when they attempt to resume semantic indexing against a server that would produce incompatible vectors.

Example drift behavior:

```text
embedding execution identity changed since this autoindex run started
```

Recovery guidance distinguishes the safe choices without asking the user to call the identity endpoint:

- restore the original server embedding identity and resume;
- rebuild into a new state directory and new prefix; or
- delete and recreate the old indices before reusing the same prefix.

An ONNX asset-read failure is cached for the lifetime of that Engine, so its error guidance explicitly instructs the operator to repair the assets and restart the server.

## Identity semantics

- **Lexical:** stable algorithm/version identity; resumable.
- **ONNX experimental:** SHA-256 identity derived from the exact model and tokenizer bytes consumed by lazy inference, plus dimensions, pooling, token limit, and semantic-contract data; resumable.
- **Neural or proxy:** fresh indexing remains supported, but resume is rejected unless the effective backend resolves to a stable built-in identity. Remote aliases are not presented as immutable content identities.
- **Fallbacks:** the endpoint reports the backend actually constructed. Slim neural builds and invalid proxy configurations that fall back to lexical report lexical, not the requested backend.

## Exact-byte and lifetime guarantees

The endpoint and lazy ONNX loader share one immutable `Arc` snapshot owned by the Engine configuration. This closes the prior path fingerprinting race:

- replacing a model with different bytes while preserving its path, size, and mtime cannot change what the current Engine later loads;
- a subsequent Engine rereads the path and observes the replacement;
- dropping the final Engine/config/handle owner releases the raw asset bytes;
- the process-global session registry stores only content hashes, scheduling settings, and weak session references;
- public caller-supplied descriptive hashes cannot make different bytes share a session.

First-use file reading and hashing runs through `spawn_blocking`, rather than blocking an async router worker.

## Failure and compatibility behavior

- The endpoint is additive and protected by cluster-read authorization.
- It is exposed on the native router and the default ES-compatible `:9200` router used by autoindex.
- Autoindex discovers and calls it automatically through the already configured `--url`; no extra endpoint configuration, credential, or command is required.
- Non-semantic autoindex plans do not require the endpoint.
- Completed semantic legacy journals without a pinned identity fail closed.
- Malformed or conflicting identity journal records fail closed.
- Non-resumable semantic backends may start fresh but cannot resume.
- Endpoint failures return a fixed sanitized message while full details remain in server logs.

## Tests

Focused coverage includes:

- same-path, same-size, restored-mtime ONNX replacement;
- identity and real lazy ONNX pool consuming bit-identical assets;
- concurrent identity and lazy initialization;
- sequential Engines observing replacement assets;
- release of model/tokenizer bytes after Engine lifetime;
- distinct bytes never sharing sessions;
- identical bytes sharing despite untrusted descriptive labels;
- forged identical labels not aliasing different bytes;
- native and ES-compatible endpoint parity;
- path sanitization for asset failures;
- neural/proxy fallback truthfulness;
- authorization;
- journal fresh/same/drift/non-resumable/legacy/malformed/conflicting cases;
- HTTP resume drift rejection before another bulk request.

The optional real-model tests pass with the frozen fused FP32 model and tokenizer used by the FinanceBench work.

## Validation

Exact head:

```text
4abf38254d520d14c6e4caf904edf89d078a3d8f
```

Exact parent:

```text
76d8bb02bc1f1aa9dffcf2eb41ccd8d4c9bee6d6
```

Scoped validation:

```text
xerj-ai ONNX suite: 37 passed, 0 failed, 1 explicit real-asset test ignored
ONNX handle focused suite: 14 passed
engine identity focused suite: 15 passed
API sanitized endpoint tests: 2 passed
autoindex identity/recovery HTTP tests: 4 passed
cargo fmt --all --check: passed
ONNX all-target Clippy with -D warnings: passed
```

Mandatory ES-YAML gate on the exact head:

```text
1365 passed
0 failed
3 skipped
1368 total
```

The local gate used a scoped non-fat release build:

```bash
CARGO_PROFILE_RELEASE_LTO=false \
CARGO_PROFILE_RELEASE_CODEGEN_UNITS=16 \
cargo build --release -j 32 -p xerj-server -p es-yaml-runner
```

The non-fat choice affects local build time, not conformance behavior. Release packaging may continue to use its distribution profile.

## Scope

This PR supplies the transparent server/client contract required by crash-safe incremental autoindex reconciliation. It does not change the normal `xerj autoindex <folder>` workflow and does not include the larger reconciliation executor itself; that remains a separate dependent contribution.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -78,6 +78,44 @@ jobs:
         working-directory: engine
         run: cargo test --release --workspace -- --test-threads=2
 
+  onnx-feature:
+    name: ONNX feature compile + tests
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@stable
+        with:
+          components: clippy
+
+      - name: Cache
+        uses: actions/cache@v4
+        with:
+          path: |
+            ~/.cargo/registry/cache
+            ~/.cargo/registry/index
+            ~/.cargo/git/db
+            engine/target
+          key: ci-onnx-cargo-${{ hashFiles('engine/Cargo.lock') }}
+          restore-keys: ci-onnx-cargo-
+
+      - name: Compile the server ONNX feature path
+        working-directory: engine
+        run: cargo check -p xerj-server --no-default-features --features onnx-experimental
+
+      - name: Test the ONNX embedder
+        working-directory: engine
+        run: cargo test -p xerj-ai --features onnx-experimental --lib -- --test-threads=2
+
+      - name: Test engine embedding identity with ONNX
+        working-directory: engine
+        run: cargo test -p xerj-engine --features onnx-experimental --lib index::embedding_identity_tests -- --test-threads=2
+
+      - name: Clippy the ONNX implementation
+        working-directory: engine
+        run: cargo clippy -p xerj-ai -p xerj-engine --all-targets --features onnx-experimental -- -D warnings
+
   api-checks:
     name: API smoke + liveness + benchmark
     runs-on: ubuntu-latest
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,17 +7,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- **`xerj autoindex` pins the server's embedding execution identity** (#160).
+  Vectors produced by two different embedding backends — or by two different
+  models, tokenizers, or vector widths within one backend — are not comparable,
+  so mixing them in one index silently degrades every similarity result rather
+  than failing. A semantic autoindex run now fetches
+  `GET /v1/embedding/identity`, records the returned opaque digest in its
+  journal before the index is created, and requires the same digest on every
+  resume; a mismatch aborts before the next index-create or `_bulk`. There is
+  no new command or setup step — `xerj autoindex ./reports` calls the endpoint
+  itself over the already-configured `--url`, and plans with no semantic fields
+  skip it entirely.
+
+  The endpoint is authenticated (cluster-read) and deliberately opaque: it
+  returns a version, the *effective* backend, a digest, the semantic contract,
+  whether that backend is resumable, and a sanitized reason when it is not. It
+  never returns credentials, provider URLs, model names, or local paths.
+  `dimensions` is reported only for the backends whose width the server
+  actually pins (`lexical`, and `onnx-experimental` at 384); it is omitted for
+  `neural`, whose width comes from the loaded model's `hidden_size`, and for
+  `proxy`, whose width is whatever the remote returns.
+
+  `lexical` and `onnx-experimental` are resumable. `neural` and `proxy` are
+  not: neither attests immutable model bytes, so a fresh semantic run is
+  allowed but a resume is refused.
+
 ### Changed
 
-- **Semantic autoindex resume now pins the server's embedding execution
-  identity.** Existing state directories that already contain completed
-  semantic files predate that identity marker and therefore fail closed rather
-  than risk mixing incompatible vector spaces. Rebuild with `--fresh` and a new
-  `--prefix`; restoring the exact original server identity is sufficient only
-  for state that already contains a marker. Neural and proxy backends remain
-  usable for fresh runs but are not resumable because their loaded assets are
-  not yet content-addressed. The identity endpoint omits `dimensions` for those
-  unpinned backends instead of guessing a model width.
+- **BREAKING: an existing semantic `autoindex` state directory cannot be
+  resumed** (#160). A journal written before this change carries no pinned
+  identity, so there is no way to prove the server still produces the same
+  vector space; rather than assume it, the resume fails closed with *"this
+  semantic autoindex journal predates embedding identity pinning and cannot be
+  resumed safely"*. The same path fires when a previous `--no-semantic` run is
+  followed by a semantic run over the same state directory. Recover by
+  rebuilding with `--fresh` and a new `--prefix`; before reusing the old
+  prefix, delete and recreate its prior autoindex indices. Non-semantic runs
+  and brand-new state directories are unaffected.
+
+- `EmbeddingProxy::new` now rejects an endpoint that is not `http(s)` at
+  construction time instead of at first request, so an invalid
+  `embedding.default_endpoint` falls back to lexical at startup and
+  `/v1/embedding/identity` reports `lexical` — the backend the server will
+  really use — rather than `proxy`.
 
 ## [1.0.0-rc.10] - 2026-08-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **Semantic autoindex resume now pins the server's embedding execution
+  identity.** Existing state directories that already contain completed
+  semantic files predate that identity marker and therefore fail closed rather
+  than risk mixing incompatible vector spaces. Rebuild with `--fresh` and a new
+  `--prefix`; restoring the exact original server identity is sufficient only
+  for state that already contains a marker. Neural and proxy backends remain
+  usable for fresh runs but are not resumable because their loaded assets are
+  not yet content-addressed. The identity endpoint omits `dimensions` for those
+  unpinned backends instead of guessing a model width.
+
 ## [1.0.0-rc.10] - 2026-08-03
 
 ### Security
--- a/engine/crates/xerj-ai/src/embed.rs
+++ b/engine/crates/xerj-ai/src/embed.rs
@@ -139,6 +139,13 @@ pub struct EmbeddingProxy {
 impl EmbeddingProxy {
     /// Create a new proxy with the given config.
     pub fn new(config: EmbeddingProxyConfig) -> Result<Self> {
+        let endpoint = reqwest::Url::parse(&config.endpoint)
+            .map_err(|e| XerjError::embedding(format!("invalid embedding endpoint: {e}")))?;
+        if !matches!(endpoint.scheme(), "http" | "https") {
+            return Err(XerjError::embedding(
+                "embedding endpoint must use http or https",
+            ));
+        }
         let client = reqwest::Client::builder()
             .timeout(Duration::from_secs(config.timeout_secs))
             .build()
@@ -311,6 +318,14 @@ mod tests {
     fn config_with_api_key() {
         let cfg = EmbeddingProxyConfig::new("http://localhost", "model").with_api_key("sk-test");
         assert_eq!(cfg.api_key.as_deref(), Some("sk-test"));
+    }
+
+    #[test]
+    fn proxy_rejects_invalid_or_non_http_endpoints_at_construction() {
+        for endpoint in ["not a url", "file:///private/model"] {
+            let cfg = EmbeddingProxyConfig::new(endpoint, "model");
+            assert!(EmbeddingProxy::new(cfg).is_err(), "{endpoint} was accepted");
+        }
     }
 
     #[tokio::test]
--- a/engine/crates/xerj-ai/src/embedder.rs
+++ b/engine/crates/xerj-ai/src/embedder.rs
@@ -313,37 +313,6 @@ impl std::fmt::Debug for OnnxConfig {
 }
 
 #[cfg(feature = "onnx-experimental")]
-impl PartialEq for OnnxConfig {
-    fn eq(&self, other: &Self) -> bool {
-        self.model_sha256 == other.model_sha256
-            && self.tokenizer_sha256 == other.tokenizer_sha256
-            && self.intra_threads == other.intra_threads
-            && self.session_pool_size == other.session_pool_size
-            && self.microbatch == other.microbatch
-            && self.max_inflight_calls == other.max_inflight_calls
-            && self.max_input_bytes_per_call == other.max_input_bytes_per_call
-            && self.max_inflight_input_bytes == other.max_inflight_input_bytes
-    }
-}
-
-#[cfg(feature = "onnx-experimental")]
-impl Eq for OnnxConfig {}
-
-#[cfg(feature = "onnx-experimental")]
-impl std::hash::Hash for OnnxConfig {
-    fn hash<H: std::hash::Hasher>(&self, state: &mut H) {
-        self.model_sha256.hash(state);
-        self.tokenizer_sha256.hash(state);
-        self.intra_threads.hash(state);
-        self.session_pool_size.hash(state);
-        self.microbatch.hash(state);
-        self.max_inflight_calls.hash(state);
-        self.max_input_bytes_per_call.hash(state);
-        self.max_inflight_input_bytes.hash(state);
-    }
-}
-
-#[cfg(feature = "onnx-experimental")]
 #[derive(Debug, thiserror::Error)]
 #[error("{reason}")]
 pub struct OnnxAdmissionError {
@@ -540,7 +509,7 @@ mod onnx_handle_tests {
 
     fn cfg(model_sha256: &str) -> OnnxConfig {
         OnnxConfig {
-            model_bytes: Arc::from(b"model bytes".as_slice()),
+            model_bytes: Arc::from(format!("model bytes for {model_sha256}").into_bytes()),
             tokenizer_bytes: Arc::from(b"tokenizer bytes".as_slice()),
             model_sha256: model_sha256.into(),
             tokenizer_sha256: "tokenizer-hash".into(),
@@ -554,10 +523,24 @@ mod onnx_handle_tests {
     }
 
     #[test]
-    fn same_paths_with_different_content_hashes_never_share_session_cell() {
-        let first = OnnxHandle::new(cfg("model-a"));
-        let second = OnnxHandle::new(cfg("model-b"));
+    fn different_actual_bytes_never_share_session_cell() {
+        let first = cfg("model-a");
+        let mut second = cfg("model-b");
+        second.model_bytes = Arc::from(b"different model bytes".as_slice());
+        let first = OnnxHandle::new(first);
+        let second = OnnxHandle::new(second);
         assert!(!Arc::ptr_eq(&first.shared, &second.shared));
+    }
+
+    #[test]
+    fn identical_bytes_share_even_when_descriptive_hash_fields_differ() {
+        let first_cfg = cfg("descriptive-a");
+        let mut second_cfg = cfg("descriptive-b");
+        second_cfg.model_bytes = first_cfg.model_bytes.clone();
+        second_cfg.tokenizer_bytes = first_cfg.tokenizer_bytes.clone();
+        let first = OnnxHandle::new(first_cfg);
+        let second = OnnxHandle::new(second_cfg);
+        assert!(Arc::ptr_eq(&first.shared, &second.shared));
     }
 
     #[test]
--- a/engine/crates/xerj-ai/src/embedder.rs
+++ b/engine/crates/xerj-ai/src/embedder.rs
@@ -357,16 +357,46 @@ pub struct OnnxHandle {
 }
 
 #[cfg(feature = "onnx-experimental")]
+#[derive(Clone, Debug, PartialEq, Eq, Hash)]
+struct OnnxCacheKey {
+    model_sha256: String,
+    tokenizer_sha256: String,
+    intra_threads: usize,
+    session_pool_size: usize,
+    microbatch: crate::onnx::MicrobatchConfig,
+    max_inflight_calls: usize,
+    max_input_bytes_per_call: usize,
+    max_inflight_input_bytes: usize,
+}
+
+#[cfg(feature = "onnx-experimental")]
+impl From<&OnnxConfig> for OnnxCacheKey {
+    fn from(cfg: &OnnxConfig) -> Self {
+        Self {
+            model_sha256: cfg.model_sha256.clone(),
+            tokenizer_sha256: cfg.tokenizer_sha256.clone(),
+            intra_threads: cfg.intra_threads,
+            session_pool_size: cfg.session_pool_size,
+            microbatch: cfg.microbatch,
+            max_inflight_calls: cfg.max_inflight_calls,
+            max_input_bytes_per_call: cfg.max_input_bytes_per_call,
+            max_inflight_input_bytes: cfg.max_inflight_input_bytes,
+        }
+    }
+}
+
+#[cfg(feature = "onnx-experimental")]
 impl OnnxHandle {
     fn new(cfg: OnnxConfig) -> Self {
         use std::collections::HashMap;
         use std::sync::{Mutex, OnceLock, Weak};
-        static CELLS: OnceLock<Mutex<HashMap<OnnxConfig, Weak<OnnxShared>>>> = OnceLock::new();
+        static CELLS: OnceLock<Mutex<HashMap<OnnxCacheKey, Weak<OnnxShared>>>> = OnceLock::new();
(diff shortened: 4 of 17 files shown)


xerj-org (changes requested):
Thanks @buger — this is good work and the mechanism is sound. I verified every regression test in this PR is real by reverting each source hunk in isolation; four separate tests fail for exactly the stated reason, which is more than most PRs can demonstrate:

- Removed the pin call in `xerj-autoindex/src/lib.rs` → `semantic_resume_rejects_embedding_identity_drift_before_another_bulk` failed at `failure_resume_http_tests.rs:256` (`left: 0 / right: 1`, no `embedding_identity` journal event).
- Neutered the drift bail in `state.rs` → the same test failed at line 259 (`unwrap_err()` on an `Ok`), plus `state::tests::semantic_identity_is_durable_and_drift_fails_closed` at `state.rs:98`.
- Reverted `OnnxCacheKey::from` to key off the descriptive hash fields → `forged_equal_hash_fields_cannot_share_sessions_for_different_bytes` and `identical_bytes_share_even_when_descriptive_hash_fields_differ` both failed.
- Replaced the `runtime_onnx_assets` `OnceLock` with a per-call re-read → `onnx_identity_and_lazy_loader_share_one_immutable_byte_snapshot` failed with the digest changing `8fb0b84f…` → `90992e2f…` across a same-path/same-size/restored-mtime swap.

I also checked the load-bearing property directly: the backend ladder in `embedding_execution_identity` agrees case-for-case with `make_embedder`, including the slim-build neural→lexical fallback and the invalid-endpoint proxy→lexical downgrade. So the endpoint reports the backend the server will actually use, not the one that was requested. Authz is fine too — `/v1/embedding/identity` classifies as `Target::Cluster`, passes `cluster_is_read`, and the router tests assert no `api_key`, endpoint, model path or OS error text leaks.

The named failure mode — mixing vectors from two embedding backends across an autoindex resume — **is** prevented. I am blocking on four cheap items rather than on any correctness bug in what you implemented.

## 1. `dimensions` is hardcoded and wrong for two of four backends

`engine/crates/xerj-engine/src/index.rs` returns `dimensions: xerj_ai::local::DEFAULT_DIMS` (384) unconditionally. But neural dims come from `config.hidden_size` of `cfg.neural_model` (`xerj-ai/src/neural.rs:85` — a 768-dim encoder still reports 384), and proxy dims are whatever the remote returns. The 384 constraint is only actually enforced for ONNX, by `validate_onnx_dimensions`.

It is not load-bearing today (`esclient.rs` only checks `dimensions != 0`), but this PR advertises the response as a machine-readable contract "directly inspectable for agents and diagnostics." A false field in an agent-facing contract is a real finding under this repo's honest-claims rule. Please derive it from the effective backend, or omit it for backends where it is not pinned.

## 2. The default backend has no real identity

The lexical material is the literal string `"lexical-feature-hash.v1;dimensions=384"`, and nothing ties it to `local_embed`. `xerj-ai/src/local.rs` has only a within-run determinism test (`dims_and_determinism`) — no golden vector. So changing the token splitter or the hash seed would change the vector space while `identity_sha256` stays `177b14d1bdff6343…`, and autoindex would happily resume straight across it.

That is precisely the failure this PR exists to prevent, on the backend most users run. The golden-sha assertion in `lexical_execution_identity_is_stable_and_sanitized` pins the *string*, not the algorithm, so it cannot catch this. A golden-vector test on `local_embed` would close it.

## 3. No CHANGELOG entry for a user-visible breaking change

`CHANGELOG.md` has a live `[Unreleased]` section and the diff touches no docs at all. Any existing semantic autoindex state directory with completed files now hard-errors on resume ("predates embedding identity pinning") and requires `--fresh` plus a new `--prefix`. The same fail-closed path fires for a previous `--no-semantic` run followed by a semantic run on the same state dir. Users need to read that somewhere other than the error text.

## 4. CI never compiles `onnx-experimental`

`grep` over `.github/workflows/` returns zero hits for `onnx-experimental`, so the majority of the engine and embedder diff — and every "exact-byte guarantee" in the body — is invisible to the green checks. I compiled and ran it locally (29 tests passing across `xerj-ai` and `xerj-engine`), but nothing protects it from regressing. Adding that feature to a CI job would be worth it given how much of this PR lives behind it.

## Non-blocking, but please note

**Undisclosed memory regression on the ONNX path.** Model and tokenizer bytes are now retained as `Arc<[u8]>` in `config.embedding.runtime_onnx_assets` for the whole `Engine` lifetime, on top of ORT's own copy. The removed `sha256_file` streamed in 1 MiB chunks and `OnnxHandle::init` dropped its buffer after load. So a single cluster-read `GET /v1/embedding/identity` now permanently materializes the entire model in RAM even if no embedding ever occurs. I understand it as a deliberate tradeoff for the TOCTOU fix — it just belongs in an otherwise exhaustive PR body.

**Recovery guidance is advisory, not enforced.** `--fresh` against the same prefix relies on idempotent `_id` overwrite, so docs whose source files were deleted keep their old-vector-space embeddings. And `validate_embedding_identity` writes its marker only for ONNX, so a direct `_bulk` writer against a server switched lexical→neural still mixes vector spaces silently. Your body scopes itself honestly to autoindex, so I am recording this as remaining risk, not an overclaim.

Also worth calling out in the body: `EmbeddingProxy::new` now rejects non-`http(s)` endpoints at construction, which is an unrelated behaviour change riding along.

CI is genuinely green (9/9 on `4abf3825`, Build + Test a real 57-minute pass, not the #157 hang). Happy to merge once (1), (2) and (3) land — (4) I would take as a follow-up if you would rather not grow this PR.

buger:
I reproduced and agree with all four blocking findings.

## Changes made

1. `dimensions` is now an optional field in the execution-identity contract.
   Lexical and ONNX return `Some(384)` because that width is pinned and
   enforced. Neural and proxy omit the field because a configured model name or
   remote alias does not prove the width that will actually load or answer.
   Autoindex accepts an omitted width only for those unpinned backends and still
   rejects zero or a missing width for lexical/ONNX.

2. The default lexical identity is now bound to executable behavior, not only
   to a manually maintained version string. The engine hashes the little-endian
   `f32` output of representative probes covering empty input, ASCII
   tokenization, numeric/punctuation input, Unicode, and path-like separators.
   That behavior digest is part of the execution identity, so changing the
   splitter, FNV seed, weights, normalization, default width, or covered
   behavior changes the resume identity automatically. A golden SHA-256 test
   pins the probe output (`e6240f3d…`) and the resulting public identity
   (`58a7576c…`).

3. `CHANGELOG.md` now documents the user-visible migration under `[Unreleased]`:
   legacy semantic state with completed files fails closed, recovery requires
   `--fresh` plus a new prefix, and fresh neural/proxy runs remain
   non-resumable.

## Non-blocking findings acknowledged

- The ONNX byte snapshot intentionally retains model and tokenizer bytes for
  the Engine lifetime in addition to ORT-owned state. This closes the
  identity/load time-of-check-time-of-use gap but increases resident memory,
  including when identity is queried before inference.
- Recovery is scoped to autoindex journals, not arbitrary direct `_bulk`
  writers. Reusing a prefix with `--fresh` cannot remove documents for deleted
  source files, so the safe guidance remains a new prefix followed by explicit
  cutover/cleanup.
- `EmbeddingProxy::new` now rejects non-HTTP(S) endpoints at construction; that
  behavior change should be called out in the PR body.

## Verification

- `cargo test -j 32 -p xerj-engine --lib`: 408 passed.
- `cargo test -j 32 -p xerj-autoindex --lib`: 214 passed.
- `cargo test -j 32 -p xerj-ai --features onnx-experimental`: 37 passed,
  1 explicitly ignored real-model test.
- `cargo test -j 32 -p xerj-engine --features onnx-experimental
  index::embedding_identity_tests --lib`: 15 passed, 1 explicitly ignored
  real-model test.
- Strict scoped Clippy for `xerj-ai`, `xerj-engine`, `xerj-api`, and
  `xerj-autoindex`, all targets with `onnx-experimental`: clean.
- `git diff --check`: clean.

xerj-org:
@buger — I've written the four blocking items. Because `maintainer_can_modify` is off on this PR and I have no push access to `probelabs/xerj`, I could not push onto your branch. The four commits are on `xerj-org/xerj` at **`review/pr160-embedding-identity-followups`**, branched from your exact head `4abf3825`, with your six commits untouched:

```
git remote add upstream https://github.com/xerj-org/xerj.git
git fetch upstream review/pr160-embedding-identity-followups
git merge upstream/review/pr160-embedding-identity-followups
```

(Flipping "Allow edits by maintainers" on works too, and I'll push them directly instead.)

---

### 1. `dimensions` — fixed (`298633e`)

Now `Option<usize>`, `#[serde(skip_serializing_if = "Option::is_none")]`, derived from the *effective* backend inside the same match that already produces `material`/`resumable`/`reason`:

- `lexical` → `Some(DEFAULT_DIMS)`. `Embedder::Lexical` is `local_embed(t, DEFAULT_DIMS)` with no per-field override, so this one is genuinely pinned.
- `onnx-experimental` → `Some(384)`, taken from the persisted identity rather than re-stated, since `validate_onnx_dimensions` is what enforces it.
- `neural`, `proxy` → omitted. A reader gets no field at all rather than a wrong one.

The lexical material also derives its `dimensions=` from `DEFAULT_DIMS` instead of repeating the literal, so the constant and the identity string cannot drift apart. The canonical string is byte-identical, so `identity_sha256` for lexical is still `177b14d1…` and your golden assertion is untouched.

`esclient` keeps rejecting `dimensions: 0` and now accepts an absent width. Proved by reverting: with the field back to `usize`, `embedding_identity_accepts_an_omitted_width_and_rejects_a_zero_one` fails with `Err(parse embedding identity / Caused by: missing field 'dimensions')`. And with `dimensions: Some(DEFAULT_DIMS)` restored unconditionally, `configured_backends_report_the_effective_runtime_backend` fails at `index.rs:23733` — `left: Some(384) / right: None`.

### 2. Lexical golden vector — added (`1af12f2`)

`local::tests::lexical_vector_space_is_frozen` in `xerj-ai/src/local.rs`: a whole-vector checksum over the f32 bit patterns for one fixed input at `DEFAULT_DIMS`, plus four exact bucket values so a failure says *what* moved. The checksum function is deliberately **not** `fnv1a64`, so changing the embedder's hash cannot drag the fingerprint along with it.

I confirmed the hole you described is real before closing it. Dropping the trigram weight from `0.35` to `0.30` in `local_embed`:

```
---- local::tests::lexical_vector_space_is_frozen stdout ----
assertion `left == right` failed: the lexical vector space changed; see this test's doc comment
  left: 9167875185330901667
 right: 14262738046721457820
```

and with that same mutation in place, `lexical_execution_identity_is_stable_and_sanitized` **still passes** — every vector moved and `identity_sha256` did not budge. On the default backend, this test is the only thing standing between that edit and a silent cross-space resume.

The doc comment states the fix order for a future editor: bump `lexical-feature-hash.v1` → `.v2` first (which changes the digest and makes resumes across the change fail closed), and only then re-record the constants.

### 3. CHANGELOG — added (`d1b0903`)

`[Unreleased]` gets an **Added** entry for the endpoint (what it returns, what it deliberately does not, which backends are resumable and why) and a **Changed** entry flagged BREAKING for the resume behaviour: a pre-existing semantic state directory fails closed, `--no-semantic` followed by semantic on the same state dir hits the same path, and recovery is `--fresh` + a new `--prefix` or deleting and recreating the prior indices. The `EmbeddingProxy::new` scheme rejection you noted is recorded there too, since it changes when an invalid endpoint is detected.

### 4. ONNX in CI — added (`6cad7de`)

New `onnx-experimental` job: `cargo clippy --all-targets -- -D warnings` and `cargo test --lib` over `xerj-ai`, `xerj-engine`, `xerj-api` with the feature on. No model download — the real-asset tests stay `#[ignore]`d and `ort` fetches its own prebuilt runtime during the build.

`--lib` rather than the full test set is deliberate, and I want to be upfront about why: with the feature on and the **integration** targets included, `xerj-engine`'s `painless_script_limits::call_depth_limit_survives_the_multi_thread_block_in_place_path` aborts with `fatal runtime error: stack overflow` in a debug build. That reproduces on your head without any of my changes and is unrelated to this PR — CI's existing Build + Test job runs `--release`, where it passes — but it would have made the new job red. Every feature-gated test lives in a lib unit-test module anyway, and the clippy step still type-checks the integration targets with `--all-targets`. Worth a separate issue.

---

### What I ran

All from the tip of the follow-up branch:

```
cargo fmt --all --check                                          passed
clippy -D warnings, default features, xerj-{ai,engine,api,autoindex}   passed
clippy -D warnings --all-targets, onnx-experimental on ai/engine/api   passed
cargo test -p xerj-autoindex                        214 passed, 0 failed
cargo test --lib -p xerj-ai      (onnx-experimental)  38 passed, 1 ignored
cargo test --lib -p xerj-api     (onnx-experimental) 155 passed, 0 failed
cargo test --lib -p xerj-engine  (onnx-experimental) 419 passed, 1 ignored
```

The two ignored are the real-MiniLM-asset tests. I did not re-run the ES-YAML conformance gate — nothing here touches a request path.

### On your two non-blocking notes

I left both alone deliberately. The `Arc<[u8]>` retention is the TOCTOU fix working as designed and belongs in the PR body, not in a code change I make on your branch. The direct-`_bulk` gap is real but out of scope for a PR that scopes itself to autoindex — I'd rather it got its own issue than a rushed marker write in here.

Thanks for a genuinely careful piece of work — the byte-snapshot design is the right shape, and the tests behind it hold up under reverting.

buger:
CI comment now addressed too

buger:
Thank you — I fetched `review/pr160-embedding-identity-followups` and adopted the three code/documentation commits without rewriting their authorship:

- `298633e` — report identity dimensions only where the server pins them
- `1af12f2` — freeze the lexical vector space with a golden vector
- `d1b0903` — document embedding identity pinning and the resume break

They are now on the fork as `5c11f48`, `ac20c81`, and `3dd95d4`, respectively. The commit objects changed because they were cherry-picked onto the fork's newer head, but all retain `xerj-org <git@xerj.org>` as author and `buger <leonsbox@gmail.com>` as committer.

I compared these changes point-by-point with my overlapping `0519c89` follow-up and dropped that local commit rather than retaining two implementations of the same fixes.

The maintainer version is the better contract:
- `dimensions` is truthful: lexical and ONNX report a pinned width; neural and proxy omit it.
- The lexical vector space has an independent whole-vector checksum and exact bucket assertions.
- A lexical implementation change now fails CI and requires an explicit `lexical-feature-hash.v2` identity bump before the golden vector is re-recorded. This makes an intentional compatibility decision visible in review instead of silently changing the public identity at runtime.
- The changelog explains that the endpoint is automatic, which backends are resumable, the legacy-journal break, and the safe rebuild procedure.
- The ONNX workflow covers `xerj-ai`, `xerj-engine`, and `xerj-api`, with strict all-target Clippy and library tests. Real-model tests remain explicitly ignored, so CI does not download model assets.

The superseded local implementation remains recoverable at `backup/pr160-before-maintainer-20260804`.

## Verification

- `xerj-ai` with `onnx-experimental`: 38 passed, 1 explicitly ignored real-model test.
- Focused `xerj-engine` ONNX identity suite: 15 passed, 1 explicitly ignored real-model test.
- `xerj-autoindex`: 214 passed.
- `xerj-engine`: 408 passed.
- Strict all-target ONNX Clippy for `xerj-ai`, `xerj-engine`, and `xerj-api`: passed.
- Strict all-target Clippy for `xerj-autoindex`: passed.
- `cargo fmt --all --check`: passed.
- `git diff --check`: passed.

xerj-org:
@buger — all four blocking items are addressed at `3dd95d4f`, and CI is green on the jobs that have reported. Before approving I re-read the merged result and found that **merging my follow-up branch reverted two of your own hardenings**, and that is my fault, not yours: I branched `review/pr160-embedding-identity-followups` from `4abf3825`, which predates your `0519c899`. Applying those three commits on top overwrote the parts of `0519c899` they touched.

Restored, on top of your current head, at **`review/pr160-restore-lexical-probe`** (branched from `3dd95d4f`, so it merges as a fast-forward):

```
git fetch upstream review/pr160-restore-lexical-probe
git merge upstream/review/pr160-restore-lexical-probe
```

### 1. The lexical algorithm probe (`507f952`)

Your `lexical_algorithm_probe_sha256` is gone at head. The lexical material is back to the literal `lexical-feature-hash.v1;dimensions={DEFAULT_DIMS}`, with nothing in it computed from `local_embed`. This is the finding you agreed with, reopened by my merge — on the default backend.

I re-proved it on the branch as it stands. Dropping the trigram weight in `local_embed` from `0.35` to `0.30`:

- at `3dd95d4f`, `lexical_execution_identity_is_stable_and_sanitized` **passes** — every lexical vector moved and `identity_sha256` is still `177b14d1bdff6343…`, so an autoindex resume proceeds across two vector spaces;
- with your probe restored, the same mutation moves the public identity:

```
---- index::embedding_identity_tests::lexical_execution_identity_is_stable_and_sanitized stdout ----
  left: "c6679f2c962d40d24bb924f3b916050e24c99b2d84ffb116cbdeb99422b8dcff"
 right: "58a7576ccc1497ee471418bb6e8918c7e5a837161792992e1e213a2dcb2ebdbe"
```

so the resume fails closed. Your function and material are restored verbatim, keeping the `DEFAULT_DIMS` derivation from `5c11f485` — the two changes now compose instead of one overwriting the other. Lexical `identity_sha256` is `58a7576c…` and the probe is `e6240f3d…` again, exactly the values in your comment, which were not true of head.

`lexical_vector_space_is_frozen` stays as the developer-facing signal that says *what* moved and what to do. The probe is what makes the server refuse the resume even when nobody reads that test.

### 2. The pinned-backend width check (`2f5b5dc`)

`0519c899` made the autoindex client reject an identity that omits `dimensions` when the backend is `lexical` or `onnx-experimental`, while still accepting an omitted width from `neural`/`proxy`. My `5c11f485` dropped that clause. Restored, plus two cases in the existing table test. Reverting the guard:

```
---- esclient::tests::embedding_identity_accepts_an_omitted_width_and_rejects_a_zero_one stdout ----
assertion `left == right` failed: Ok(EmbeddingExecutionIdentity { version: 1, backend: "lexical",
  identity_sha256: "aaaa…", dimensions: None, semantic_contract: "semantic_text-derived-vector.v1",
  resumable: true, non_resumable_reason: None })
  left: true
 right: false
```

### What I checked on the merged result

The load-bearing property holds. `pin_embedding_identity` runs before index creation and before any `_bulk` on every semantic run: an invalid digest, a non-resumable backend on a resumed journal, a digest that differs from the journalled one, and a resumed journal with completed files and no marker each `bail!` rather than proceed. `lib.rs` calls it only when the plan has a semantic field, so non-semantic runs are untouched. With `507f952` in, that pin is now bound to the embedder's real output on the default backend rather than to a hand-maintained version string.

I did not re-open the two non-blocking notes — the `Arc<[u8]>` retention and the direct-`_bulk` gap are still where we left them.

### What I ran

From the tip of the restore branch:

```
cargo fmt --all --check                                             passed
clippy --all-targets, xerj-{engine,autoindex,ai}                    0 warnings
clippy --all-targets, onnx-experimental, xerj-{ai,engine,api}       0 warnings
cargo test --lib -p xerj-autoindex                       214 passed, 0 failed
cargo test --lib -p xerj-engine                          408 passed, 0 failed
cargo test --lib -p xerj-ai                               20 passed, 0 failed
cargo test --lib -p xerj-engine --features onnx-experimental
  index::embedding_identity                       15 passed, 1 ignored (real assets)
```

Sorry for the churn — the revert was introduced by my own follow-up branch. Once this lands I have nothing else blocking.

buger:
Thank you for catching the regression introduced by the earlier follow-up
merge. I reproduced both reopened holes on the PR head and adopted
`review/pr160-restore-lexical-probe` as a clean fast-forward, preserving both
maintainer commits and their original `xerj-org <git@xerj.org>` authorship:

- `507f952` — restore the lexical algorithm probe lost in the merge
- `2f5b5dc` — reject an omitted vector width from a pinned backend

The new head is:

```text
2f5b5dcbce9a98fb11c071266ba65585223303b1
```

The fixes compose the two intended contracts instead of allowing one to
overwrite the other:

- The independent golden-vector test remains the developer-facing signal that
  lexical behavior moved.
- The runtime lexical execution identity again includes
  `lexical_algorithm_probe_sha256()`, so a splitter, hash, weight,
  normalization or width change moves the identity and makes resume fail
  closed.
- Lexical and ONNX identities must include `dimensions`, because those
  backends pin their width.
- Neural and proxy identities may omit `dimensions`, because the server cannot
  honestly pin those widths at this boundary.

The manually added ONNX workflow remains in
`.github/workflows/ci.yml`; the fast-forward did not remove or rewrite it.

## Verification

- `xerj-ai --features onnx-experimental --lib`: 38 passed, 1 explicitly
  ignored real-model test.
- Focused `xerj-engine` ONNX identity suite: 15 passed, 1 explicitly ignored
  real-model test.
- `xerj-autoindex --lib`: 214 passed.
- Strict all-target ONNX Clippy for `xerj-ai`, `xerj-engine`, and `xerj-api`:
  passed.
- Strict all-target Clippy for `xerj-autoindex`: passed.
- `cargo fmt --all --check`: passed.
- `git diff --check`: passed.

xerj-org:
Re-reviewed at head `2f5b5dc`. The four blocking items from the earlier review are now closed on the branch, and I verified each in the code rather than from the commit messages:

1. **`dimensions` hardcoded to 384** — closed by `5c11f48`. The width is now reported only for the backends the server actually pins.
2. **Lexical backend had no real identity** — closed. `2f5b5dc`/`507f952` restore `lexical_algorithm_probe_sha256()` and the `algorithm-probe=` term in the identity material (`engine/crates/xerj-engine/src/index.rs:23476`), so the lexical resume identity is now derived from `local_embed` output rather than from a literal version string. The published constants `e6240f3d…` (probe) and `58a7576c…` (identity) are asserted in-tree and green in CI, so they match what the code produces. Thank you for pulling those in — the revert was my follow-up branch's fault for being cut from a commit that predated your `0519c89`.
3. **No CHANGELOG entry for the breaking resume change** — closed by `3dd95d4`.
4. **CI never compiled `onnx-experimental`** — closed by `291785e`.

The `esclient` guard rejecting an identity that omits `dimensions` from a pinned backend is present at `engine/crates/xerj-autoindex/src/esclient.rs:158` with table coverage for both `lexical` and `onnx-experimental`.

CI is 10/10 green on `2f5b5dc` and the branch merges into current `main` with no conflicts. Merging.

Three non-blocking items stay open and do not gate this PR — they were recorded as accepted risk and are worth their own issues rather than more churn here: the `Arc<[u8]>` retention on the ONNX path, the direct-`_bulk` writer cross-space gap, and the debug-build stack overflow in `painless_script_limits::call_depth_limit_survives_the_multi_thread_block_in_place_path` under `onnx-experimental` with integration targets (which is why the new CI job is `--lib` only).

PR #163, opened 2026-08-03
fix(engine): memtable circuit breaker ignores cgroup memory limit

Vinz2168:
## Summary

Investigates the sustained-ingest memory-growth OOM documented in `demo/playbooks/ES_COMPATIBILITY.md §15` (`MemoryMax=8G` cgroup, 40k docs/s open-loop bulk writer, RSS climbs to the cap, kernel OOM-kills in ~3-4 minutes).

**This PR fixes one confirmed, real bug and rules out all three of §15's documented suspects with direct measurement — but does not fully resolve §15.** Read the "Honest status" section before treating this as closed.

## Repro method

`systemd-run --user --scope -p MemoryMax=8G` isn't available in this dev environment, so a Docker container with `--memory=<cap>` was used instead — same kernel cgroup-memory mechanism, same OOM-kill behavior, just via Docker's cgroup wrapper instead of systemd's. A `soak_mixed.sh` script was recreated per §15's description: a fresh index, hit with a 10k-doc `_bulk` every 250ms, fired on a timer without waiting for the previous request to finish (open-loop).

Built with `cargo build --profile profiling --features debug-profiling` per `demo/playbooks/debug-profiling/README.md` and profiled with the bundled `capture.py`/`inspect.py`/`pprof` toolkit (jemalloc heap profiles + `GET /_nodes/stats`'s `segment_hydration_cache` gauge + live thread-count sampling).

## The three documented suspects — all checked with direct evidence, all ruled out

**#1 "DV/doc-values cache has no byte budget, unlike stored-slices"** — false as stated. `dv_cache` and `stored_slices_cache` are both gated by the same `SegmentHydrationBudget` (`segment_cache_budget.rs`). Live-polled `GET /_nodes/stats` once a second during a soak run: `segment_hydration_cache.current_in_bytes` climbs and then plateaus exactly at its configured ~614 MB ceiling (`segment_hydration_cache_mb=614` in the startup log) while total process RSS keeps climbing past 2-3 GB in the same window. The cache is doing its job; it is not the leak.

**#2 "WAL retention across checkpoints"** — `WalWriter::prune_verified` (`xerj-storage/src/wal.rs`) is genuinely dead code (`grep` shows it's only called from its own unit tests). But the *actual* production pruning path, `IndexStore::wal_maintain_all_verified` (`xerj-storage/src/index_store.rs`), is a more sophisticated, cache-optimized superseding mechanism (per-generation verdict cache, re-verifies only the "unproven" residue against the version map on each tick) and **is** correctly wired into the flush path (`finalize_flush_with_publisher`'s hot-path tick, and `Index::flush()`'s `force_wal_maintenance` for the unconditional pass). This suspect does not appear to be live.

**#3 "Drained-memtable Arc retention on the flush path"** — the closest real mechanism in the codebase is that `drain_shard_inner` (`xerj-engine/src/memtable.rs`) hands the just-drained memtable's HashMaps (inverted index, doc-values, field-length stats) to a **freshly-spawned, uncapped OS thread** (`xerj-drain-free`) that drops them off the flush critical path — a deliberate, documented perf optimization with no cap on concurrent in-flight drain threads. Hypothesized this could pile up under sustained flush pressure and tested it directly: sampled live thread count every second via `ps -M` against the native macOS release binary during a 35s soak. Thread count ramped to ~123 in the first few seconds, then stayed **flat** at 123 for the rest of the run while RSS climbed to 2.7 GB+ over the same window. Threads are being scheduled and completing promptly — this is not the leak either.

## The bug this PR actually fixes

`governor.rs`'s auto-derived memtable byte ceiling — the "parent circuit breaker" this very module was built to add after a prior 112 GiB OOM incident (see the module's own doc comment) — computed itself from raw host RAM (`system_total_bytes()`, i.e. `/proc/meminfo`'s `MemTotal`) instead of the already-correct, cgroup-aware `effective_memory_limit_bytes()` that every sibling budget (RSS watermark, segment hydration cache) in the same function already uses. `/proc/meminfo` is not namespace-virtualized by default, so inside any container/cgroup smaller than the host's physical RAM this silently reads the HOST total, not the container's actual budget.

Confirmed live with the governor's own startup log line, both pre- and post-fix, under an identical 1024 MB cgroup cap on an 8126 MB-RAM host:

```
pre-fix:  memtable_budget_mb=2048  memory_limit_mb=1024   (the ceiling alone is 2x the ENTIRE container)
post-fix: memtable_budget_mb=512   memory_limit_mb=1024   (correctly clamped to 50% of the real limit)
```

With the ceiling effectively disabled pre-fix, the RSS watermark (95% of the effective limit) was left as the only real backstop — and it trips with almost no headroom left to absorb already-admitted in-flight work or one more flush's segment-cache warm, which matches the documented "the governor logs the trip, the process is OOM-killed moments later anyway" symptom.

### Fix

Derive the auto memtable budget from `effective_memory_limit_bytes()` (matching every sibling budget in `build()`), and clamp it to never exceed 50% of that limit — the pre-existing 2 GiB floor alone can reproduce the same more-than-the-container problem under a small enough cap, independent of the host-RAM bug, so the floor needed its own ceiling.

## What was actually verified

- `cargo test -p xerj-engine --lib governor::` — all 11 tests pass, including the new-shape assertions on `memtable_budget_trips_on_refresh` and the cgroup-resolution tests (no test needed changing; none hardcoded the old formula).
- `cargo clippy -p xerj-engine --lib -- -D warnings` — clean.
- `cargo fmt -p xerj-engine -- --check` — clean.
- `cd engine && cargo build --release` — succeeds (built natively on macOS/arm64 after the Linux Docker VM's fat-LTO link ran out of memory on this constrained dev box — a local resource limit, not a code issue; the `profiling`-profile Linux build used for all the soak/profiling work above compiled and ran fine in the same Docker environment).
- Governor startup log confirms the exact before/after numbers quoted above, reproduced twice (1024 MB and 3072 MB caps).
- Re-ran the same soak (3 GB cap) after the fix: see "Honest status" below — it still hit OOM at roughly the same point.

## Honest status — NOT fully resolved, needs another pass

Re-running the identical soak (3 GB cap) after the fix still hit the same OOM at roughly the same RSS/timing as before the fix. The fix closes a real, independently provable gap in the governor's own admission math (the circuit breaker literally could not trip before consuming the whole container, pre-fix) — but on its own it did not prevent the OOM in this downscaled local repro.

Two known limitations on why this isn't conclusive yet:

1. **Scale mismatch.** My test environment (a Docker Desktop VM with only ~8.3 GB total RAM) can't reproduce the host-RAM-vs-cgroup-cap *ratio* that makes the pre-fix bug most severe — the doc's own `MemoryMax=8G` repro implies a real host with meaningfully more RAM than 8 GB, where the old `system_total_bytes()/4` ceiling would have been far more oversized than in my 8.3 GB-host / 1-3 GB-cap tests (where the pre-existing 2 GiB floor already dominated the calculation either way, muting the fix's visible effect). This needs re-verification on a host with a realistic RAM-to-cap ratio.
2. **Root driver still unidentified.** All three documented suspects were ruled out with direct measurement (see above), and this fix addresses a fourth, independently-found bug — but the actual dominant driver of the RSS growth in the soak repro is still unknown. One unexplored lead: flush cadence during the soak ran ~2.5 flushes/sec (89 flushes in 35s, ~16k docs each) against a merge cycle that only sweeps every 5 seconds in fixed-size batches — a possible flush-rate-vs-merge-rate mismatch causing an un-merged segment backlog. On-disk store size didn't look dramatically bloated when checked once (132.5 MB for 1.36M docs), so this is a lead, not a finding — flagging it here so the next investigator doesn't have to rediscover it.

**Recommendation for review**: merge as a real, standalone correctness fix to the resource governor (it's independently justified regardless of whether it fully explains §15), but keep §15 open and pointed at the merge-backlog lead above for the next pass.

## Test plan

- [x] `cargo test -p xerj-engine --lib governor::` (11/11 pass)
- [x] `cargo clippy -p xerj-engine --lib -- -D warnings` (clean)
- [x] `cargo fmt -p xerj-engine -- --check` (clean)
- [x] `cargo build --release` (clean, native macOS/arm64)
- [x] Live governor-log verification of the exact metric change, at two different cgroup caps
- [x] Soak re-test after the fix (result: OOM still occurs in this downscaled repro — documented above, not hidden)
- [ ] Re-verification on a host with a production-realistic RAM-to-cgroup-cap ratio — not done, flagged as follow-up
- [ ] Root-cause identification of the actual dominant RSS driver — not done, flagged as follow-up

### DIFF
--- a/engine/crates/xerj-common/src/config.rs
+++ b/engine/crates/xerj-common/src/config.rs
@@ -982,8 +982,10 @@ pub struct LimitsConfig {
     /// past the cap.
     pub max_buckets: usize,
     /// Process-wide ceiling on the total bytes buffered across ALL indices'
-    /// memtables (default: `0` = auto-derive to 25% of system RAM, floored
-    /// at 2048 MiB). This is the parent circuit breaker for the ingest path:
+    /// memtables (default: `0` = auto-derive to 25% of the effective
+    /// cgroup/system memory limit, floored at 2048 MiB, capped at 50% of
+    /// that same limit). This is the parent circuit breaker for the ingest
+    /// path:
     /// per-index back-pressure only bounds one index at `~3×flush_size_mb`,
     /// so `N` indices could buffer `N × 1.5 GiB` with no global ceiling —
     /// the structural cause of the 112 GiB OOM. When the summed memtable
@@ -1029,7 +1031,7 @@ impl Default for LimitsConfig {
             max_result_window: 10_000,
             max_mget_docs: 10_000,
             max_buckets: 65_536,
-            max_total_memtable_mb: 0, // 0 = auto-derive (25% RAM, floor 2 GiB)
+            max_total_memtable_mb: 0, // 0 = auto-derive (25% effective limit, floor 2 GiB, cap 50%)
             max_segment_hydration_cache_mb: 0, // 0 = 20% effective memory, no floor
             memory_watermark_percent: 95,
             disk_flood_stage_percent: 95,
--- a/engine/crates/xerj-engine/src/governor.rs
+++ b/engine/crates/xerj-engine/src/governor.rs
@@ -651,6 +651,12 @@ fn effective_memory_limit(sys: u64, cgroup: Option<u64>) -> u64 {
 /// (`memory.max`, walking up to a parent when a leaf reads `max`) and falls
 /// back to cgroup v1 (`memory.limit_in_bytes`). Returns `None` when no
 /// finite limit applies.
+///
+/// A limit file that is missing, unreadable, malformed, `0`, or the v1/v2
+/// "no limit" sentinel (`max` / `9223372036854771712`) contributes nothing
+/// rather than failing: the caller then falls back to host RAM. Nothing on
+/// this path panics, so an unexpected cgroupfs layout degrades to the
+/// pre-cgroup behaviour instead of taking the process down at startup.
 #[cfg(target_os = "linux")]
 fn fold_cgroup_memory_limit(current: Option<u64>, raw: &str) -> Option<u64> {
     const UNLIMITED_V1: u64 = 9_223_372_036_854_771_712;
@@ -666,43 +672,97 @@ fn fold_cgroup_memory_limit(current: Option<u64>, raw: &str) -> Option<u64> {
     }
 }
 
+/// Walk from `rel` up to the root, folding the tightest finite limit found
+/// in `<root><rel>/<file>` at each level. cgroup memory limits are
+/// hierarchical in BOTH v1 and v2 — a finite child value does not override a
+/// smaller finite ancestor — so the effective limit is the minimum over the
+/// whole chain. Levels whose file is missing or unreadable are skipped, so a
+/// partially visible hierarchy still yields whatever it does expose.
+///
+/// `root` is a parameter rather than a constant so the hierarchy walk is
+/// testable against a temp-dir fixture without a real cgroupfs mount.
 #[cfg(target_os = "linux")]
-fn cgroup_memory_limit_bytes() -> Option<u64> {
-    // cgroup v2: /proc/self/cgroup has a single "0::<path>" line.
-    if let Ok(cg) = std::fs::read_to_string("/proc/self/cgroup") {
-        for line in cg.lines() {
-            if let Some(path) = line.strip_prefix("0::") {
-                // Walk from leaf to root and take the tightest finite limit.
-                // cgroup-v2 limits are hierarchical: a finite child value
-                // does not override a smaller finite ancestor.
-                let mut rel = path.trim().to_string();
-                let mut tightest = None;
-                loop {
-                    let full = format!("/sys/fs/cgroup{rel}/memory.max");
-                    if let Ok(s) = std::fs::read_to_string(&full) {
-                        tightest = fold_cgroup_memory_limit(tightest, &s);
-                    }
-                    if rel.is_empty() || rel == "/" {
-                        break;
-                    }
-                    match rel.rfind('/') {
-                        Some(0) => rel = String::new(), // step to root next iter
-                        Some(i) => rel.truncate(i),
-                        None => break,
-                    }
-                }
-                if tightest.is_some() {
-                    return tightest;
-                }
+fn tightest_limit_up_hierarchy(root: &str, rel: &str, file: &str) -> Option<u64> {
+    let mut rel = rel.trim().to_string();
+    let mut tightest = None;
+    loop {
+        let full = format!("{root}{rel}/{file}");
+        if let Ok(s) = std::fs::read_to_string(&full) {
+            tightest = fold_cgroup_memory_limit(tightest, &s);
+        }
+        if rel.is_empty() || rel == "/" {
+            break;
+        }
+        match rel.rfind('/') {
+            Some(0) => rel = String::new(), // step to root next iter
+            Some(i) => rel.truncate(i),
+            None => break,
+        }
+    }
+    tightest
+}
+
+/// The cgroup-relative path of the v1 `memory` controller from one
+/// `/proc/self/cgroup` line (`<hierarchy-id>:<controllers>:<path>`).
+/// Returns `None` for v2 lines (`0::<path>`, empty controller list) and for
+/// controller sets that do not include `memory`.
+#[cfg(target_os = "linux")]
+fn cgroup_v1_memory_path(line: &str) -> Option<&str> {
+    let mut parts = line.splitn(3, ':');
+    let _hierarchy_id = parts.next()?;
+    let controllers = parts.next()?;
+    let path = parts.next()?;
+    controllers
+        .split(',')
+        .any(|c| c == "memory")
+        .then_some(path.trim())
+}
+
+/// Resolve the memory limit from the contents of `/proc/self/cgroup`,
+/// against the v2 unified mount at `v2_root` and the v1 memory-controller
+/// mount at `v1_root`. Split out from [`cgroup_memory_limit_bytes`] so the
+/// whole resolution — v2 line, v1 line, and the mount-root fallback — is
+/// testable against a fixture instead of the host's real cgroupfs.
+#[cfg(target_os = "linux")]
+fn cgroup_memory_limit_from(self_cgroup: &str, v2_root: &str, v1_root: &str) -> Option<u64> {
+    // cgroup v2: /proc/self/cgroup has a single "0::<path>" line, and the
+    // limits live at <v2_root><path>/memory.max.
+    for line in self_cgroup.lines() {
+        if let Some(path) = line.strip_prefix("0::") {
+            let tightest = tightest_limit_up_hierarchy(v2_root, path.trim(), "memory.max");
+            if tightest.is_some() {
+                return tightest;
             }
         }
     }
 
-    // cgroup v1 fallback.
-    if let Ok(s) = std::fs::read_to_string("/sys/fs/cgroup/memory/memory.limit_in_bytes") {
-        return fold_cgroup_memory_limit(None, &s);
+    // cgroup v1: the memory controller is mounted separately and
+    // /proc/self/cgroup carries one "<id>:memory:<path>" line per hierarchy.
+    // Walk that path from leaf to root exactly as for v2. Under Docker the
+    // cgroupfs is bind-mounted at the container's own directory, so the leaf
+    // path from /proc/self/cgroup does not resolve and only the mount root
+    // reads — which is the container's real limit. Under a bare
+    // `systemd-run -p MemoryMax=` on a v1/hybrid host it is the other way
+    // round: the process sits in a leaf slice and the mount root reads
+    // unlimited. Folding the whole chain is correct in both cases.
+    for line in self_cgroup.lines() {
+        if let Some(path) = cgroup_v1_memory_path(line) {
+            let tightest = tightest_limit_up_hierarchy(v1_root, path, "memory.limit_in_bytes");
+            if tightest.is_some() {
+                return tightest;
+            }
+        }
     }
-    None
+
+    // Last resort: /proc/self/cgroup was unreadable or named no memory
+    // controller, so read the v1 mount root directly.
+    tightest_limit_up_hierarchy(v1_root, "", "memory.limit_in_bytes")
+}
+
+#[cfg(target_os = "linux")]
+fn cgroup_memory_limit_bytes() -> Option<u64> {
+    let self_cgroup = std::fs::read_to_string("/proc/self/cgroup").unwrap_or_default();
+    cgroup_memory_limit_from(&self_cgroup, "/sys/fs/cgroup", "/sys/fs/cgroup/memory")
 }
 
 #[cfg(not(target_os = "linux"))]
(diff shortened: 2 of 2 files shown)


xerj-org (changes requested):
Welcome, @Vinz2168 — and thank you, this is a real bug and a correct fix. I reproduced it live, and it is worse than your write-up claims.

`build()` in `engine/crates/xerj-engine/src/governor.rs` derived the auto memtable budget from `system_total_bytes()` (raw `/proc/meminfo` `MemTotal`) while every sibling budget in the same function — the RSS watermark, the segment hydration cache — already used the cgroup-aware `effective_memory_limit_bytes()`. Under a real cgroup-v2 scope on this 119 GiB host:

```
systemd-run --user --scope -p MemoryMax=1G   # memory.max = 1073741824 confirmed

unfixed: memtable_budget_mb=30527  memory_limit_mb=1024
fixed:   memtable_budget_mb=512    memory_limit_mb=1024
```

So the "parent circuit breaker" was ~30× the size of the entire container and could never trip. The three-line fix is minimal and correct, and the new `.min(limit/2)` is non-binding on any uncapped host with ≥ 4 GiB RAM — so the blast radius is confined to exactly the deployments where today's value is unsafe. I also confirmed `memtable_budget_bytes` is a comparison threshold only (nothing outside `governor.rs` sizes an allocation from it), so the tighter value can only cause an earlier HTTP 429, never a smaller buffer.

I want to merge this. Four cheap asks first, none of which are about the fix itself.

## 1. No regression test — and as written, one cannot be added

This is the blocker. A safety mechanism just silently regressed, and nothing would catch it regressing again. The auto-derive branch has no coverage at all: `memtable_budget_trips_on_refresh` sets `max_total_memtable_mb=100` explicitly and takes the *other* branch. And the formula cannot be unit-tested where it currently lives, because it is inline in `build()` and calls `effective_memory_limit_bytes()`, which reads `/proc` directly.

Ask: extract `fn auto_memtable_budget(effective_limit: u64) -> u64` and add a table test — 1 GiB → 512 MiB, 4 GiB → 2 GiB, 8 GiB → 2 GiB, 64 GiB → 16 GiB, asserting `budget <= limit/2` in every case.

I wrote that test myself to verify your fix. With the fix: `memtable_budget_mb=512 memory_limit_mb=1024`, 12/12 governor tests pass. With only the source hunk reverted and the test left in place:

```
REVIEW: memtable_budget_mb=30527 memory_limit_mb=1024
panicked at crates/xerj-engine/src/governor.rs:859:
auto memtable budget 30527MB exceeds 50% of the effective (cgroup-aware) memory limit 1024MB
```

That branch was scratch and I deleted it — the test needs to live in your PR. Note the sibling budgets in this same function have tests and this one did not, which is precisely why it drifted.

## 2. The config docs contradict the merged code

`engine/crates/xerj-common/src/config.rs:984-985` still documents `max_total_memtable_mb` as "auto-derive to 25% of system RAM, floored at 2048 MiB", and line 1032 repeats "0 = auto-derive (25% RAM, floor 2 GiB)". After this PR it is 25% of the *effective* limit, floored at 2 GiB, then capped at 50% of that limit. That docstring is the user-facing description of the knob, so please update it in the same PR.

## 3. The comment that lands overclaims

The 14-line comment above the change states as established fact:

> With this ceiling effectively disabled, the RSS watermark below is left as the only real backstop, tripping with almost no headroom to absorb already-in-flight ingest work or one more flush's segment-cache warm.

That is an unproven causal explanation of the §15 OOM — and your own "Honest status" section says the post-fix soak still OOMed at the same RSS and timing, so this story was explicitly *not* confirmed. It also hard-codes a personal anecdote ("confirmed live under a 1 GiB cgroup cap on an 8 GiB-RAM host") that will age badly in the tree.

Please trim it to what is provable: the budget must be derived from the effective limit because `/proc/meminfo` is not namespace-virtualized, and the 2 GiB floor needs its own ceiling. That is a complete and sufficient justification on its own.

## 4. Commit message contradicts the PR body

The commit body says suspect #3 (`Arc<Memtable>` retention) was "not conclusively investigated … should not be considered cleared", while the PR description says it was ruled out by direct `ps -M` thread-count sampling. The message also ends with "No PR opened per explicit instruction: … PR #93's CLA gate blocks merges regardless until signed" — on a commit that *is* the PR. Per `engine/CLAUDE.md` the git history is the session log, so please rewrite it to match reality.

## Follow-up, not a blocker on this diff

cgroup v1 is only half handled, and it bounds what this fix can claim. `cgroup_memory_limit_bytes()` parses only the v2 `0::<path>` line of `/proc/self/cgroup` and walks ancestors for `memory.max`; the v1 lines (`N:memory:/path`) are never parsed, and the v1 fallback blindly reads the mount root `/sys/fs/cgroup/memory/memory.limit_in_bytes`. Inside a Docker v1 container the cgroupfs is bind-mounted at the container's own directory so that read is correct — but on a v1 or hybrid host under `systemd-run -p MemoryMax=`, the process sits in a leaf slice while the root reads unlimited, `fold_cgroup_memory_limit` filters it, and the effective limit falls back to host RAM, making this entire fix a no-op. That may be the §15 repro configuration, if that host is v1. I verified the v2 path end-to-end here; the v1 path I could only read, since this host is pure v2. Worth its own issue.

## Credit where it is due

Your PR body is unusually honest for a first contribution: it says up front that the fix does not close §15, reports that the post-fix soak still OOMed, leaves two test-plan boxes unchecked, and hands the next investigator a concrete lead (flush cadence ~2.5/s against a 5s fixed-batch merge sweep). Nothing in it claims behaviour the diff does not implement, and there are no unmeasured performance claims. That is exactly the standard this repo asks for, and it is rarer than it should be.

CI status as of this review: 8/9 pass, Build + Test still in progress, no failures. It will need re-checking before merge. The branch is behind main but conflict-free — `governor.rs` is untouched on main.

Ping me when the test and the docstring land and I will take another look promptly.

Vinz2168:
Thanks for the thorough review — and for reproducing on a real host, that 30527 vs 2048 number makes the severity much clearer than my own 8 GB-RAM dev box could. Addressed all four:

1. **Regression test** — extracted `auto_memtable_budget(effective_limit: u64) -> u64` out of `build()`, added `auto_memtable_budget_is_cgroup_proportional_and_bounded` with your exact table (1 GiB → 512 MiB, 4 GiB → 2 GiB, 8 GiB → 2 GiB, 64 GiB → 16 GiB) plus a `<= limit/2` assertion per case. 12/12 governor tests pass.
2. **Config docs** — updated both spots in `xerj-common/src/config.rs` to describe the new (effective-limit-based, 50%-capped) behavior.
3. **Trimmed the comment** — cut it down to the provable justification (`/proc/meminfo` isn't namespace-virtualized; the 2 GiB floor needs its own ceiling) and dropped the unconfirmed causal story about the RSS watermark being "the only real backstop."
4. **Commit message rewritten** — reflects that a PR is in fact open, and that suspect #3 was tested (not "not conclusively investigated"). Amended in place and force-pushed rather than stacking a fixup, since the single commit is the PR.

Also folded in your cgroup-v1 finding as a known-limitation note in the commit body, credited to you, with a recommendation it get its own issue rather than block this one.

Re-verified after amending: `cargo test -p xerj-engine --lib governor::` (12/12), `clippy -D warnings` clean, `fmt --check` clean, `cargo build --release` clean.

xerj-org:
Thanks @Vinz2168 — all four asks landed cleanly, and the extracted `auto_memtable_budget` + table test is exactly what was missing. I've pushed one commit on top (`57d887f`) rather than ask you for another round, since it closes the cgroup-v1 follow-up I raised and that follow-up directly bounds what your fix can claim.

**What I changed, and why it belongs in this PR rather than its own issue**

I originally said the v1 gap was a separate issue. On a second look it isn't separable from *this* diff: your fix derives the memtable budget from `effective_memory_limit_bytes()`, so it is only as good as that limit. On a v1 or hybrid host the limit lookup missed the cap entirely and fell back to host RAM — which would have made the headline fix a silent no-op on exactly the kind of host it is meant to protect. Shipping the breaker fix without it would have meant claiming a fix that doesn't hold everywhere.

- `tightest_limit_up_hierarchy(root, rel, file)` — your v2 leaf-to-root walk, extracted and now shared by both cgroup versions. v1 limits are hierarchical too, so the minimum over the chain is the right answer for both. `root` is a parameter purely so the walk is testable without a real cgroupfs.
- `cgroup_v1_memory_path(line)` — parses the `<id>:<controllers>:<path>` form, handles co-mounted controllers (`cpu,cpuacct,memory`), rejects v2 lines.
- v1 now walks its own slice path before falling back to the mount root. Folding the whole chain keeps the Docker bind-mount case working (leaf unresolvable, root read is the real limit) *and* finds the `systemd-run -p MemoryMax=` leaf case that was previously invisible.
- `cgroup_memory_limit_from(self_cgroup, v2_root, v1_root)` — the whole resolution split from the `/proc` read, so the wiring is covered end-to-end and not just the helpers. That was the same drift risk you fixed for `auto_memtable_budget`.

Pure-v2 hosts are unaffected: the v2 branch is tried first and returns before any v1 code runs.

**On the regression test.** `cgroup_hierarchy_walk_handles_v1_v2_max_and_missing_files` builds a cgroupfs fixture in a tempdir, so it runs the same on any host. It covers the v2 leaf reading the literal `max`, a tighter leaf beating a looser ancestor, a v1 leaf slice limit, missing/unlimited chains resolving to `None` rather than 0 or a panic, and full resolution for pure-v2, pure-v1, hybrid and Docker-v1 layouts. I proved it bites by reverting each hunk:

```
v1 leaf walk removed:
  assertion `left == right` failed: v1: the leaf slice limit must be
  found, not just the mount root
    left: None
   right: Some(2147483648)

hierarchy walk cut to the leaf only:
  assertion `left == right` failed: a v2 leaf reading the literal "max"
  must inherit its ancestor's finite limit
    left: None
   right: Some(1073741824)
```

**One small thing in your commit.** I dropped a `1 * GIB` from your table test — CI runs `cargo clippy --workspace --all-targets -- -D warnings` and `identity_op` fails it. Worth knowing for next time: `cargo clippy --lib` doesn't compile the test target, so test-only lints don't show up locally. `--all-targets` is the one that matches CI.

**What I did not verify.** No cgroup-v1 host was available to me, so the v1 walk is proven by fixture and code reading only — not by a running v1 container. I'd rather state that plainly than imply a live repro I didn't do. The v2 path is unchanged and still has your live reproduction behind it.

Verified here: `cargo test -p xerj-engine --lib` 408/408, `governor::` 13/13, clippy `--all-targets` clean on `xerj-engine` and `xerj-common`, `fmt --check` clean.

Your history is untouched — one commit added on top, no force-push. Nothing further needed from you; I'll re-check CI and hand it to the merge queue. Genuinely good first contribution, and the honesty in the write-up is the part I'd most like to see repeated.

xerj-org:
Thanks @Vinz2168 — merging this. 🎉 A first contribution that lands a real correctness fix is a good one.

All four asks came back addressed in `7a961ee`, and the cgroup v1 follow-up landed in `57d887f`. I verified the regression tests actually have teeth rather than trusting them, by reverting each source hunk one at a time and confirming the matching test fails for the right reason:

- dropping `.min(effective_limit / 2)` → `auto_memtable_budget(1073741824) = 2147483648, expected 536870912`
- removing the v1 leaf walk → `v1: the leaf slice limit must be found, not just the mount root`
- reading only the v2 leaf → `a v2 leaf reading the literal "max" must inherit its ancestor's finite limit`
- dropping `*value > 0` from the fold filter → `left: Some(0), right: None`

All four restored green afterwards. 408/408 lib tests, fmt and clippy clean with `-D warnings --all-targets`, and all 9 CI checks green on `57d887f`.

Two limits stated plainly, so nobody reads more into this than it earns:

- **This does not close the sustained-ingest OOM (§15).** Your own PR body reports the post-fix soak still OOMing in the downscaled repro, and I did not re-run that soak. This lands as a standalone governor correctness fix and §15 stays open.
- The cgroup **v1** path is proven by tempdir fixture and code reading only — no v1 host was available here, so there is no live container reproduction behind it.

Nice catch on the limit being read from the host instead of the process's own slice. 👏

PR #168, opened 2026-08-04
docs: codify contribution review discipline

buger:
## Why this is worth merging

Recent contributions repeatedly reached review with strong focused tests but still required substantial follow-up for the same reasons: the effective stacked diff was different from the tip commit, a recovery command could not actually run, invocation-local values were written as durable metadata, a fresh-schema test missed an older dynamic mapping, an optional fast path became a correctness dependency, feature-gated code was absent from CI, or a benchmark claim depended on evidence reviewers could not inspect.

These were not isolated coding mistakes. They were gaps in the repository's pre-submission review contract. Leaving the lessons in individual PR threads makes every maintainer and every new agent rediscover them.

This PR adds the smallest durable fix: a concise mandatory pointer in `AGENTS.md`, a human-facing pointer in `CONTRIBUTING.md`, and one detailed protocol that can evolve independently. The agent guide stays short while the reasoning, checks, and reusable templates remain discoverable.

## How it is used

Before submitting a non-trivial change, a contributor or agent opens the protocol, marks each section applicable or explains why it is not, runs the copy-paste ten-step audit, and records each important claim in the evidence table. A reviewer can then inspect the effective diff, behavior contract, negative controls, recovery proof, and known gaps without reconstructing the author's reasoning from scratch.

## What this prevents

- Correct patches being reviewed against the wrong ancestry, missing stacked dependencies or interactions with newly merged work.
- Fixes preserving one field or path while silently breaking sibling durable metadata, legacy mappings, zero-work resume, failure atomicity, or executable recovery.
- Optional optimizations failing the whole workflow instead of falling back safely and observably.
- Machine-readable identities, telemetry, exit codes, and agent-facing fields claiming more certainty than the effective runtime can provide.
- Tests passing because they depend on clocks, disk headroom, descriptor limits, process-global state, build profile, or a mock that cannot reproduce the real failure.
- Feature-gated paths and public performance claims receiving credit without applicable CI, matched controls, correctness gates, and inspectable evidence.

The protocol also distinguishes deterministic identity from collision-resistant uniqueness, requires identity and execution to use the same immutable asset snapshot, forbids incidental tie-breaking between equally valid durable matches, and requires targeted mutation or negative-control evidence for important regressions.

## Files changed

- `docs/CONTRIBUTION_REVIEW.md`: the review protocol, copy-paste audit prompt, and claim-to-evidence template.
- `AGENTS.md`: the mandatory review rule and discovery link.
- `CONTRIBUTING.md`: the human contributor entry point.

`AGENTS.md` also corrects the verified ES-YAML baseline from 1,360/1,363 to 1,365/1,368 and explains that deliberate case additions must advance the recorded verified baseline rather than leave it stale.

## Why the content is credible

The protocol was derived from concrete findings in recent reviews of stateful autoindex reconciliation, embedding execution identity, optional PDF replay, and durable catalog metadata. It was then reviewed adversarially as a contribution in its own right. That review found eight material gaps, including incorrect uniqueness wording, an impossible preflight guarantee, missing immutable-asset lifetime rules, missing ambiguity handling, incomplete legacy-schema coverage, and omissions from the copyable prompt. All eight were corrected before the commit.

## Scope and validation

This is documentation only. It changes no runtime code, dependency, workflow, public API, or product behavior.

- `git diff --check`: pass.
- All new relative Markdown links resolve within the repository.
- The committed diff contains only `AGENTS.md`, `CONTRIBUTING.md`, and `docs/CONTRIBUTION_REVIEW.md`.
- No build or runtime suite was run because no executable source changed.

## What it deliberately does not mandate

The protocol is applicability-based, not a demand to run every repository test for every change. It does not require unrelated ES-YAML conformance for documentation-only or isolated work; the existing engine hard gate remains authoritative. It does not require every change to become one large PR, prescribe a single recovery design, or turn every experiment into a public benchmark. It requires contributors to state what applies, prove the claims they make, preserve known contracts, and disclose what they did not verify.


xerj-org:
Thanks @buger — landing this. 🙏

What I actually checked before merging:

- **Effective diff against `main`** (merge-base `2e6fccd`): exactly three files, +232/−0 — `AGENTS.md` (+2), `CONTRIBUTING.md` (+2), `docs/CONTRIBUTION_REVIEW.md` (new). No runtime code, no workflow, no dependency, no generated artifact.
- **Your tip commit `ac7cf6d` ("docs: keep review guidance scoped")** — the revert of the baseline edit from `1,365/1,368` back to `1,360/1,363`. I checked its premise rather than taking it on trust: `README.md:13` still carries the `ES%20conformance-1360%2F1363` badge and `README.md:112` still reads "passes **1360 of 1363**". With that revert in, `AGENTS.md` and the README agree; without it this PR would have left the hard gate quoting one number while the public badge quoted another. Good call scoping it out. 👍
- **Every relative Markdown link** in the three files resolves to a real path in the head tree — checked mechanically against `git archive` of `ac7cf6d`, not by eye.
- **No numeric claims** anywhere in `docs/CONTRIBUTION_REVIEW.md`, so nothing for the honest-claims rule to trip over.
- **CI: all 9 checks green** on `ac7cf6d`, including Build + Test, Format + Clippy, and ES-compat YAML conformance.

On the revert-to-red step we normally run to prove a fix: it doesn't apply here, and I want to be explicit rather than imply a test I didn't run. Nothing executable changed, so there is no behaviour to break and no failure output to show — the verification boundary is exactly the documentation-only one you stated in the PR body.

One stale detail for the record, not a blocker: the PR body still says `AGENTS.md` "corrects the verified ES-YAML baseline from 1,360/1,363 to 1,365/1,368". `ac7cf6d` undid that, so the merged diff doesn't. If the newer count is real, it deserves its own PR that moves the README prose, the badge and `AGENTS.md` together. 🚀

PR #169, opened 2026-08-04
feat(autoindex): sharedterm@1 — connect documents by the distinctive words they share (#164)

### LINKED ISSUE #164: Second brain cannot connect documents by topic, only by explicit citation
## What we observed

Building brains over fifteen ordinary folders (a student's course reading, a reader's ebook shelf, a home cook's recipes, a traveller's city guides, a patient's printouts) produced maps that draw **labelled groups with almost no lines between them**. The panels report why, in their own words:

| brain | files | links found | of which *inside a single document* | notes with no links |
|---|---:|---:|---:|---:|
| library (153 ebooks) | 153 | 8,543 | **8,401** | 44,619 |
| examprep (revision notes) | 118 | 2,616 | **2,516** | 0 |
| recipes (saved recipe pages) | 141 | ~130 | most | many |
| travel (city guides, HTML) | 100 | 5,269 after relinking | — | 152 |

The pattern is consistent: nearly every "link" is a `sequence` edge joining chunks *of the same file*. Cross-document edges only appear when a document literally cites another one.

`travel` is the control that proves it. Those pages carry real `href`s; after resolving them to the local copies the brain went from 2,622 to 5,269 edges and the map drew genuine inter-group lines. The corpora saved as plain text have no links in the source, so nothing connects them and the map correctly draws nothing.

## Why this is a gap, not a bug

Every shipped detector is structural:

```
cratecite   href   mdlink   pathcite   samedir   sequence   wikilink
```

All seven find an explicit citation or a filesystem relationship. None looks at what a document is *about*. The UI states this plainly — *"LINKS COME FROM STRUCTURE AND CITATIONS, NOT MEANING"* — and `samedir.rs` concedes the workaround in a comment: directory co-location *"often means 'same topic', but it is"* weak evidence, weight 0.3, the floor of the detector set. Folders are being used as a stand-in for topics because there is no topic detector.

The practical consequence: point XERJ at a recipe folder and sourdough never connects to kimchi, even though both are fermentation and any reader would draw that edge. A knowledge base whose documents do not happen to hyperlink each other produces a graph with no edges.

## This is the norm elsewhere

Topic-level connection is what people expect from this category, and every comparable tool provides some form of it:

- **Obsidian / Roam / Logseq** — explicit `[[wikilinks]]` *plus* tags, and unlinked-mention detection that surfaces documents sharing a phrase without a link.
- **Semantic / embedding knowledge graphs** — edges from vector similarity between documents.
- **Agentic memory systems** — retrieval by similarity, so related memories surface without anyone having authored a link.

XERJ currently implements only the first half of the first bullet. For a folder of authored notes that cross-reference, that is enough. For the far more common case — a pile of PDFs, saved pages, exported notes — it produces an empty graph.

## Proposed: a topic detector

A `sharedterm` detector that emits an edge when two documents share distinctive terms, with the shared phrases carried as the edge evidence.

The parts already exist. The engine indexes every document's terms and knows their document frequency, so "distinctive" is computable without new infrastructure and without an LLM. It would keep the properties the graph is built on:

- **Deterministic and versioned**, like the other seven. Same input, same edges.
- **Evidence-carrying** — the edge shows the shared phrases, so a wrong link is inspectable the way a wrong `mdlink` is today. This matters: an unexplained similarity edge is exactly the kind of thing the evidence panel exists to prevent.
- **No LLM**, preserving the "deterministic · versioned · no LLM" line the dashboard prints.

Two things need deciding before implementation:

1. **Weight and confidence.** A shared-topic edge is weaker than a citation but stronger than `samedir`'s 0.3/0.4. It needs to rank below authored links so the map still prefers real citations.
2. **Density control.** Tuned loosely, everything links to everything and the result is the hairball THE MAP was explicitly designed to avoid. It needs a cap on edges per document and a floor on term distinctiveness, and the "what the map did not show" panel should report what the cap dropped, consistent with how the map already reports its budget.

Worth noting the honest limit: the built-in embedder is lexical feature hashing, so this would connect documents that *share vocabulary*, not documents that *mean the same thing*. That is still the connection people want for a recipe folder, and it should be described as what it is rather than as semantic understanding.

xerj-org:
Closes #164.

Every shipped detector was structural: `wikilink`, `mdlink`, `href`, `pathcite`
and `cratecite` need an explicit citation, `sequence` and `samedir` need a
filesystem relationship. Point a brain at a folder of PDFs or saved pages and
the map draws labelled groups with almost no lines between them, because those
documents never link to each other.

This adds `sharedterm@1`: a `shared_term` edge between two documents that use
the same distinctive words, carrying those words as the edge evidence.

## The decisions the issue flagged, and why

**Weight 0.45, confidence 0.5.** Above `samedir@2` (0.3/0.4), which only knows
that a human filed two documents in one folder — that is often about topic and
often about nothing. Below every authored link (`pathcite@1` is 0.6/0.7,
`wikilink@2` is 1.0/0.95): a human who writes a citation *asserts* a
relationship; this detector only infers one. Confidence 0.5 says exactly that —
an inference with real evidence behind it and a real error rate.

**Density, which is what decides between a map and a hairball.** Three named
constants, each with its reasoning in the module doc:

- `MAX_DOC_FRACTION = 0.10` — a term in more than 10% of the corpus is that
  corpus's own vocabulary, not a topic ("model" in a folder of ML papers).
  Taken literally this means **a corpus of ten documents or fewer produces no
  `shared_term` edges at all**, because a term in two of ten is in 20% of them.
  That is the rule being honest rather than an exception, and it is why the
  contract's five-note fixture edge set is unchanged by this PR.
- `MIN_SHARED_TERMS = 2`, counted in word **families** — one word in common is
  a coincidence. There is no stemming here, so "trajectory" and "trajectories"
  are two terms and one word; the family rule stops a plural from being the
  second signal.
- `MAX_EDGES_PER_DOC = 5` — a hard degree budget, spent round-robin so a
  document early in path order cannot starve a later one of its first link.

**Evidence is mandatory.** The edge quote names the shared words, rarest first,
in the same `evidence.{quote,source,offset}` shape the other detectors use, so a
wrong `shared_term` is inspectable exactly like a wrong `mdlink`. Terms lead the
quote deliberately: it is clipped at 240 characters and two real corpus paths
would otherwise push the evidence off the end of its own field (187 of 558
quotes were clipped mid-list before that ordering).

**Deterministic.** No clock, no randomness, no model. Terms accumulate per file
in stream order inside one worker; every cross-document structure is a
`BTreeMap`/`BTreeSet`, so Phase B's worker interleaving cannot reach the output.
Two independent runs over the 240-PDF library produced identical counts and
identical evidence.

## Measured on a real corpus

240 arXiv PDFs in 24 topic folders (236 extractable), indexed by the same binary
built from `origin/main` and from this branch, each into a clean data directory
and a clean resume journal.

| | before | after |
|---|---:|---:|
| live edges | 11,854 | 12,316 |
| `sequence` (inside one document) | 11,642 | 11,642 |
| `same_dir` | 212 | 212 |
| `shared_term` | 0 | **462** |
| cross-document edges | 212 | 674 |
| edges crossing a folder | **0** | 377 |
| edges crossing a top-level topic | **0** | 296 |
| candidates refused by the cap (`edges_capped`) | 0 | 271 |

Not one edge crossed a folder before, because nothing in that corpus cites
anything. That is the gap in the issue, reproduced.

**How good are the new links?** The corpus is filed by topic, so the folder is a
rough ground truth. A random pair of these documents shares a top-level topic
17.0% of the time and a leaf folder 9.7%. `shared_term` links do so 35.9% and
18.4% — about twice chance. Links resting on three or more shared words reach
41.9% / 23.3%; the two-word floor keeps the weaker band because a cross-topic
link is often correct (the same benchmark discussed in two areas), so "same
folder" undercounts correctness. This is a real signal and a noisy one, which is
the reason every edge carries its words.

Sampled evidence, verbatim from the run:

```
6 distinctive terms in common: bid, bids, allocation, competition, games, strategic
3 distinctive terms in common: uav, uavs, trajectory
2 distinctive terms in common: player, games
2 distinctive terms in common: tsinghua, utilizes      ← a bad one, and visibly bad
```

Two tuning decisions came out of that measurement rather than taste, and both
are recorded in the constants' doc comments: ranking each document's terms by
global rarity picked its *accidents* (a stray "rss", an author surname, a token
the extractor split), so ranking is by in-document frequency; and 405 of the
first 558 candidate links rested on a single shared term and read as noise,
which is what `MIN_SHARED_TERMS` refuses.

## The honest limitation

**This connects documents that share VOCABULARY, not documents that mean the
same thing.** Terms are compared as strings and are not stemmed — "ferment",
"ferments" and "fermentation" are three different terms. There is no model in
this path: XERJ's built-in embedder is lexical feature hashing, and this
detector does not use even that. Two papers about one idea in different words
stay unlinked; two documents sharing a rare word by coincidence get linked. The
docs and the console now say "shared wording", never "meaning", and the console
renders the evidence as a generated rationale rather than a quote from the note.

Also worth knowing: only prose sections reach the detector, so CSV rows and log
lines contribute nothing; and an incremental run only compares the files it
re-extracts, so a document added today is compared against the whole corpus on a
run that re-reads the whole corpus.

## Tests

Each was verified failing before the code existed (`detect::sharedterm`):

- two documents sharing distinctive terms get an edge carrying those terms —
  fails with the detector disabled
- two documents sharing only common words get no edge — fails against a naive
  implementation with the distinctiveness filter and stoplist removed, which
  links them on "and, data, filed, notes, the, were"
- the per-document cap is enforced, no document is left isolated, and every
  refused candidate is counted (45-pair clique → capped, degree ≤ 5)
- output is deterministic across two runs
- a singular and its plural are one word, not two
- a corpus too small for distinctiveness emits nothing
- plus an end-to-end test through the real pipeline against a mock endpoint,
  proving the new `detect_corpus` phase is wired in and converges on re-run

`cargo test -p xerj-autoindex` is green (218 tests), `cargo fmt --all` and
`cargo clippy -p xerj-autoindex --all-targets` are clean, and the 23 xerj-ux
tests still pass.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -34,6 +34,40 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   not: neither attests immutable model bytes, so a fresh semantic run is
   allowed but a resume is refused.
 
+- **A detector that connects documents by what they say, not only by what they
+  cite** (#164). Every shipped detector was structural — `wikilink`, `mdlink`,
+  `href`, `pathcite` and `cratecite` need an explicit citation, `sequence` and
+  `samedir` need a filesystem relationship — so a folder of PDFs or saved pages,
+  where no document links to another, produced a graph whose only edges ran
+  inside single documents and along directory chains. `sharedterm@1` emits a
+  `shared_term` edge (weight 0.45, confidence 0.5) between two documents that
+  use the same distinctive words, and carries those words as the edge evidence
+  so a wrong link is inspectable exactly like a wrong `mdlink`.
+
+  Measured on 240 arXiv PDFs filed in 24 topic folders (236 extractable), same
+  binary, same corpus: before, 11,854 edges — 11,642 `sequence` inside single
+  documents plus 212 `same_dir`, with 212 cross-document edges and **none**
+  crossing a folder. After, 462 `shared_term` edges (271 further candidates
+  refused by the density cap), 377 of them joining documents in different
+  folders and 296 joining different top-level topics. Two independent runs
+  produced identical counts and identical evidence.
+
+  Density is capped rather than trusted: a term in more than 10% of the corpus
+  is that corpus's vocabulary and is ignored, a link needs at least two shared
+  word families (so a singular and its plural are not two signals), and no
+  document takes more than 5 `shared_term` edges. What the cap refused is
+  reported by the run as `edges_capped` instead of vanishing. A corpus of ten
+  documents or fewer therefore produces no `shared_term` edges at all, which is
+  the 10% rule being honest rather than an exception.
+
+  The honest limit, which the docs and the console now state: this links
+  documents that share VOCABULARY, not documents that share MEANING. Terms are
+  compared as strings with no stemming and no model — XERJ's built-in embedder
+  is lexical feature hashing and this detector does not use even that. Against
+  the corpus's own topic filing, a random pair of those documents shares a
+  top-level topic 17.0% of the time; `shared_term` links do so 35.9% of the
+  time. Real signal, and a noisy one.
+
 ### Changed
 
 - **BREAKING: an existing semantic `autoindex` state directory cannot be
--- a/README.md
+++ b/README.md
@@ -158,7 +158,7 @@ Reference pages for individual subsystems, written against the source and
 including the limits each one does not lift:
 
 - [Second brain](./docs/SECOND_BRAIN.md) for the relationship layer over
-  indexed documents: the `/_graph` routes, evidence on links, the seven
+  indexed documents: the `/_graph` routes, evidence on links, the eight
   detectors and the two-hop cap.
 - [Scripting](./docs/SCRIPTING.md) for the Painless subset, where scripts run,
   and the resource limits that bound them.
--- a/docs/SECOND_BRAIN.md
+++ b/docs/SECOND_BRAIN.md
@@ -329,13 +329,16 @@ the section text. The quote is the full trimmed line containing that offset,
 clipped to 240 characters (`detect::line_at`, `detect/mod.rs:416`). The same
 path is mirrored into the top-level `src_file` field.
 
-Two of the seven detectors do not have an author's sentence to quote, because
-they fire on file layout rather than on text. They record a generated rationale
-in the same field instead, with offset 0:
+Three of the eight detectors do not have an author's sentence to quote, because
+they fire on file layout or on statistics rather than on a sentence someone
+wrote. They record a generated rationale in the same field instead, with
+offset 0:
 
 - `samedir` writes `"<a> and <b> share directory <dir>"` (`detect/samedir.rs:54`)
 - `sequence` writes `"<label> precedes <label> of <file>"`, or
   `"<label> opens <file>"` for the first section (`detect/sequence.rs:40`)
+- `sharedterm` writes `"<a> and <b> share N distinctive terms: …"`, naming the
+  words that produced the link, rarest first (`detect/sharedterm.rs`)
 
 The console reads that distinction back out and renders it differently: a quote
 is shown in quotation marks, a structural rationale is shown prefixed with
@@ -352,11 +355,11 @@ fill the gap.
 The honest claim is "every link shows its evidence", not "every link has a
 quote".
 
-## The seven detectors, and what they do not do
+## The eight detectors, and what they do not do
 
-Seven detectors ship. They are the complete contents of
+Eight detectors ship. They are the complete contents of
 `engine/crates/xerj-autoindex/src/detect/`, registered in
-`detect::default_detectors` (`detect/mod.rs:383`):
+`detect::default_detectors` (`detect/mod.rs`):
 
 | Detector | Edge `type` | `detector` tag | Weight | Confidence | Fires on |
 |---|---|---|---|---|---|
@@ -367,6 +370,7 @@ Seven detectors ship. They are the complete contents of
 | `cratecite` | `cratecite` | `cratecite@1` | 0.5 | 0.6 | a crate directory name in prose, linking to that directory's `Cargo.toml` |
 | `sequence` | `sequence` | `sequence@2` | 0.8 | 0.99 | the file card, then each section preceding the next within one file |
 | `samedir` | `same_dir` | `samedir@2` | 0.3 | 0.4 | files sharing a directory, chained in path order, never a clique |
+| `sharedterm` | `shared_term` | `sharedterm@1` | 0.45 | 0.5 | two documents using the same distinctive words, capped at 5 links per document |
 
 The `@N` suffix on the tag is the detector version. It is bumped on any
 behaviour change, so old edges stay attributable to the rules that produced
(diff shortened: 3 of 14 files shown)


xerj-org:
Rebased onto current `main` (`96f6d1a`, i.e. after #163 and #160 landed) and force-pushed as `0935297`. Only my own commit was rewritten — there are no contributor commits on this branch.

The rebase had one conflict, in `CHANGELOG.md`, where #160's Unreleased entries met this PR's. Resolved by keeping both: #160's entry and the `### Changed` section are preserved untouched and the `sharedterm@1` entry sits after them under `### Added`. `git diff origin/main -- CHANGELOG.md` is a pure addition, so nothing of #160's text was altered. `engine/crates/xerj-autoindex/src/lib.rs` auto-merged.

Re-verified on the rebased tree, since #160 touched the same crate:

- `cargo test -p xerj-autoindex` — 225 passed, 0 failed (218 before the rebase; the extra 7 are #160's tests, now in this crate)

CI is running on `0935297`. `Format + Clippy` and `UX pipeline tests` are green; the other eight jobs — including `Build + Test`, which is the one that exercises `brain_cli.rs`'s five-note fixture — are still in flight, so **this is not yet confirmed green and I have not merged it**. The five-note fixture is the case worth watching: a corpus of ten documents or fewer emits no `shared_term` edges at all under the 10% density rule, so that fixture's expected edge set should be unchanged.

xerj-org:
## Review: hold — 13 blocking findings across three lenses

This PR is `MERGEABLE/CLEAN`, so flagging clearly up front: **it should not be merged in its current state.** Three independent reviews (design fidelity, determinism, cost/blast-radius) each returned `fix-first`. Grouped by what would have to change.

### 1. Determinism is not established, and the test cannot see the one order-dependent path

`DocTerms::prune` (`sharedterm.rs:203`) is order-sensitive: **700 of ~957 tracked terms differ between two arrival orders of the same document.** The determinism test at `sharedterm.rs:687` varies nothing and its fixture never reaches `PRUNE_AT`, so it cannot observe this.

Either make `DocTerms` order-insensitive (accumulate unpruned, prune once in `detect_corpus`), or add a test that feeds a `>PRUNE_AT` document's sections in shuffled order and asserts identical edges. `sharedterm.rs:76` also credits `BTreeMap` for a guarantee that actually rests on an unstated Phase B invariant.

Given "same corpus in, same edges out" is the headline property, this is the blocker that matters most.

### 2. Incremental runs produce a different graph than a full run

- Distinctiveness is computed over the **re-extracted subset** (`sharedterm.rs:292-305`), not the corpus. An ordinary incremental run draws edges on corpus-wide jargon, and a run touching ≤10 files draws nothing.
- Editing the rel-smaller endpoint invalidates an edge (`lib.rs:996`) that `detect_corpus` then refuses to rebuild (`sharedterm.rs:286`) — not idempotent.
- Edges are taught by both endpoints but invalidated only via `src_file` (`sharedterm.rs:455`), so editing the rel-larger document leaves a live edge quoting words that document no longer contains.
- The per-document cap of 5 is a **per-run** property, not per-brain (`sharedterm.rs:403` + `detect/mod.rs:668`), so live degree grows without bound across incremental runs.

### 3. Claims that are stated as measured fact and are not

Under the honest-claims rule these cannot ship as written:

- `sharedterm.rs:91` — "Edges from earlier runs stay live and nothing is lost". `lib.rs:1002-1006` invalidates by `src_file` and `sharedterm.rs:300-304` returns before regenerating, so an edited file loses its `shared_term` edges until `--fresh`.
- `sharedterm.rs:40` — "everything it refuses is counted". `MAX_PARTNERS_TRACKED` refusals are not.
- `CHANGELOG.md:58` and `sharedterm.rs:37-38` — the unqualified per-document cap guarantee, per finding 2.
- `docs/design/SECOND_BRAIN_SPEC.md:1025-1026` documents rarity-first term selection; the code (`sharedterm.rs:313-326`) selects by in-document frequency and its own comment says rarity-first was measured and **rejected**. The normative spec describes a different detector than the one shipped.
- Evidence quote format: spec (`:1031`) and `docs/SECOND_BRAIN.md:340` say paths-then-terms; `sharedterm.rs:513-516` emits terms-then-paths.

### 4. Default-on with no way to decline it

`detect/mod.rs:406` + `cli.rs:219`: the only way to refuse the inferred detector is `--no-graph`, which also drops all 7 structural detectors. An inferred-edge detector that existing brains gain silently on their next run needs `--no-sharedterm` or a detector allowlist before it lands.

### 5. Unmeasured cost on the one axis that carries the risk

No wall-clock or memory measurement exists for a feature whose entire risk is edge density and O(corpus) resident state. The one memory characterisation (`sharedterm.rs:150-152`) points at the smaller structure while `sharedterm.rs:81-84` understates the dominant one by roughly 4×. Measure peak RSS and wall-clock on a corpus at least an order of magnitude past 236 documents.

Related: THE MAP's percolation guard cannot shed `shared_term` (`brain-map-pipeline.js:303`), and `shared_term` is precisely the cross-folder, component-merging edge type. Either add it to the resplit exclusion set or show a measured before/after group composition proving the bounded-map UX survives.

---

The core idea is right and #164 is a real gap — a folder of PDFs genuinely draws a map with no lines. The determinism item (1) and the doc/code divergence in (3) are what stand between this and merge.

Issue #170, opened 2026-08-04
autoindex: Rust const/static data tables get no symbol extraction, making them unreachable by symbol search


xerj-org:
## `autoindex` extracts no symbols for Rust `const`/`static` data tables

**Impact:** files whose entire content is a constant lookup table are invisible to symbol search. These are exactly the files an agent most needs to retrieve, because their contents are empirical and cannot be derived or recalled — a 256-entry frequency table is the canonical case.

### Reproduce

Index the `aho-corasick` crate, then:

```
GET /<prefix>-*/_search  {"query":{"wildcard":{"ax_path":"*byte_frequencies*"}}}
```

The record exists and its `body` is complete (4,337 chars, contains `BYTE_FREQUENCIES`), but:

```
ax_path      : aho-corasick/src/util/byte_frequencies.rs
defs         : None
symbols      : None
symbol_count : None
```

The file's first line is `pub const BYTE_FREQUENCIES: [u8; 256] = [`.

Control — a file with `fn`/`struct` in the same corpus extracts fine:

```
ax_path      : memchr/src/arch/all/twoway.rs
defs         : "struct Finder\nstruct FinderRev\nstruct TwoWay\nfunction new\n..."
symbol_count : 39
```

### Scale

On a 3-crate Rust corpus (memchr, regex, aho-corasick — 324 indexed `.rs` records), **28 records have no `defs` at all**. The affected set is dominated by data tables and macro-bodied files:

- `aho-corasick/src/util/byte_frequencies.rs` — `pub const`
- `memchr/src/arch/all/packedpair/default_rank.rs` — `pub const`
- `memchr/fuzz/fuzz_targets/*.rs` (7 files) — `fuzz_target!` macro body

### Why it matters for retrieval

Because `defs` is empty, these files can only be reached through `body`, where BM25 works against them: the defining symbol appears **once** in 4KB of numeric literals, while files that merely *reference* it mention it repeatedly in prose and code. Measured, searching the exact symbol name:

| query | top hit |
|---|---|
| `BYTE_FREQUENCIES` | `prefilter.rs` (7.47) — a *consumer* of the constant |
| | `util/mod.rs` (7.43) — a re-export |
| | `hir/literal.rs` (5.07) |

`byte_frequencies.rs`, the file that **defines** the symbol, does not appear in the top 3 for its own name.

### Suggested fix

[...]

PR #175, opened 2026-08-04
[HELD on #178] fix(autoindex): extract a C header's API surface, stop indexing include guards

### LINKED ISSUE #172: autoindex: C header API surface not extracted (prototypes, extern) and include guards indexed as constants
## `autoindex`: a C header's API surface is not extracted, and include guards are indexed as constants

Three related defects in the C tree-sitter query (`engine/crates/xerj-autoindex/src/extract/code.rs`, `C_Q`). Together they mean a `.h` file — where a C library's whole public interface lives — is close to invisible to symbol search, while contributing noise.

### Reproduce

Two files, identical except for an include guard:

```c
/* plain.h */                        /* guarded.h */
                                     #ifndef GUARDED_H
                                     #define GUARDED_H
#define MAX_ITEMS 128                #define MAX_ITEMS 128
extern int global_counter;           extern int global_counter;
struct thing { int x; };             struct thing { int x; };
int do_thing(int a);                 int do_thing(int a);
                                     #endif
```

```sh
xerj autoindex <dir> --url http://localhost:9200 --prefix ctest --no-graph --fresh
curl -s "localhost:9200/ctest-*/_search" -H 'Content-Type: application/json' \
  -d '{"size":5,"_source":["ax_path","defs","symbol_count"],"query":{"match_all":{}}}'
```

Observed:

```
plain.h     symbols=2  defs='const MAX_ITEMS\nstruct thing'
guarded.h   symbols=3  defs='const GUARDED_H\nconst MAX_ITEMS\nstruct thing'
```

### 1. Function prototypes are never captured

`int do_thing(int a);` is absent from both. `C_Q` captures `function_definition`, but a prototype is a `declaration` whose declarator is a `function_declarator` — a different node. Headers are overwhelmingly prototypes, so **the callable API of a C library does not reach `defs` at all**. Searching `do_thing` cannot find the header that declares it.

### 2. `extern` declarations are never captured

`extern int global_counter;` is absent. The shipped pattern anchors on `(declaration declarator: (init_declarator declarator: (identifier)))`, and an `extern` declaration has no initialiser, so there is no `init_declarator` to match.

### 3. Include guards are captured as constants (noise, and it is per-file)

`guarded.h` yields `const GUARDED_H`. That is not a symbol anyone searches for, and it appears once per guarded header — so on a real C tree it is one junk `defs` entry in the majority of headers. Measured on a valkey + memcached corpus (2,268 files): the guard-shaped-name rate in headers was 298/4,192 = 7% of `preproc_def` captures.

This one is a precision regression from the `preproc_def` capture added for #170, so it is worth fixing in the same area rather than later.

### Suggested direction

- Prototypes: capture `(declaration declarator: (function_declarator declarator: (identifier) @function))`, anchored to `translation_unit` so function-local declarations stay out (the 96%-locals trap measured in #170 applies here too).
- `extern`: allow a bare `(declaration declarator: (identifier) @static)`, again `translation_unit`-anchored.
- Guards: skip a `preproc_def` whose name matches the file's own guard shape (`FOO_H`, `_FOO_H`, `FOO_H_INCLUDED`) or, more robustly, one whose body is empty — a guard `#define` has no replacement list, unlike `#define MAX_ITEMS 128`.

Each is testable with the existing `syms()` harness in the same file, and each should assert on captured names rather than on the query merely compiling — a wrong node name compiles fine and captures nothing, which is how #170 stayed hidden.

Found while building a reference-coding corpus from valkey + memcached.

xerj-org:
Closes #172.

A `.h` file is where a C library's public interface lives, and `C_Q` could not see most of it: prototypes and `extern`s were invisible, while every guarded header contributed one junk constant.

**Stacking note:** this branch is cut from the unmerged #170 fix, so it carries two commits — `5712f28` is that fix (const/static/`#define`), `addb9f8` is this one. Review the second commit; the first should land on its own or with this PR.

Everything below is measured over the valkey + memcached corpus at `/home/claude/.xerj-code/corpora/kv-oss` (547 `.c`, 360 `.h`). "in-body" means the capture has a `compound_statement` ancestor — the function-locals trap that made #170's bare `(declaration …)` unusable.

## Shipped

| Defect | Pattern | Captures | In-body | Precision |
|---|---|---|---|---|
| 1. prototypes | `(declaration declarator: (function_declarator declarator: (identifier) @function))` | **3871** (3285 in `.h`) | 14 | 99.6% |
| 1. prototypes returning a pointer | same, wrapped in `pointer_declarator` | **652** (546 in `.h`) | 0 | 100% |
| 2. `extern` object | `(declaration (storage_class_specifier "extern") declarator: (identifier) @static)` | **153** | 0 | 100% |
| 2. `extern` pointer / array | same, `pointer_declarator` / `array_declarator` | **28 / 20** | 2 / 0 | 96% / 100% |
| 3. include guards | `(preproc_def name: (identifier) @const value: (preproc_arg))` | **−564 of 5255** | — | 284 dropped names are `_H`-suffixed guards |
| (also) pointer-returning **definitions** | `(function_definition declarator: (pointer_declarator …))` | **1477** | 4 | 99.7% |

The 14 in-body prototype captures are 9 instances of a `JEMALLOC_CC_SILENCE_INIT` macro misparse plus genuine local prototypes; the 2 in-body `extern`s are real `extern` statements declaring globals inside a function; the 4 in-body definitions are macro misparses of real top-level functions. Corpus totals: **21611 → 27248 captures**.

The last row is not one of the issue's three defects. `char *foo(void) {…}` wraps its `function_declarator` in a `pointer_declarator`, so the pre-existing `function_definition` pattern was already missing 1477 function definitions in `.c` files — the same node-shape bug, one line away, so it is fixed here rather than filed. Called out so it is not an undisclosed behaviour change.

### On anchoring — this contradicts the issue's suggested direction, with numbers

The issue suggests anchoring the new patterns to `translation_unit`, as #170 did. Measured, that is the wrong call **for these two patterns**: real headers are guarded, and a guard puts the whole file under `preproc_ifdef`, one level too deep. Anchored, the prototype pattern captures **689** names (176 in headers). Unanchored it captures **3871** (3285 in headers) at 99.6% precision.

Unanchoring is safe here and nowhere else in C, because each pattern demands a shape a local cannot cheaply have — a `function_declarator`, or the `extern` keyword. The shape #170 rejected is untouched and still rejected: `(declaration declarator: (identifier))` measures **6666 captures, 6033 (90%) function locals**.

### On include guards — the grammar rule, verified before use

`#define GUARDED_H` parses to `(preproc_def name: (identifier))`; `#define MAX_ITEMS 128` parses to `(preproc_def name: (identifier) value: (preproc_arg))`. The empty-replacement-list distinction holds, so requiring `value:` needs no name heuristic.

**Its cost, measured:** 564 of 5255 `#define` captures go away — 284 `_H`-suffixed guards, the rest valueless build knobs (`LUA_CORE`, `_GNU_SOURCE`, `LTTNG_UST_TRACEPOINT_DEFINE`). **21 files lose their last symbol** and return to #170's zero-symbol state. In 14 of them the lost symbol was the file's own include guard, in the other 7 a build knob; no file lost a constant anyone would search for. Whole-corpus zero-symbol files: `.c` 9→17, `.h` 10→17, against 6201 newly captured API names.

The alternative — a `#not-match?` regex on guard-shaped names — would keep the build knobs, but it is a heuristic on spelling and it misses real guards like `MATH_C_` and `_CRC16_TABLE_H__`. The grammar-level rule won.

## Not shipped

- **Double-pointer returns** (`char **f(void);`): 10 captures corpus-wide. Not worth a third pattern; tree-sitter queries cannot express "any depth of wrapper".
- **File-scope pointer globals** (`static char *p = "x";` — the `init_declarator` sits inside a `pointer_declarator`): 104 captures measured, 0 in-body, so a clean win is available. It is #170's territory, not this bug, so it is left for a follow-up rather than smuggled in.
- **`static int x;` at file scope with no initialiser**: still missed. It needs an anchored pattern (no keyword to key on that a local lacks — `static` is legal on locals), which the guarded-header problem then defeats. Not attempted.
- **Function-like macros** (`preproc_function_def`): several of the 21 newly zero-symbol headers are macro-only (`ql.h`, `qr.h`, `typed_list.h`), and their macros ARE their API. Capturing them as `@macro` would restore those files, but it is a recall change of unmeasured volume and a new symbol kind, so it is a separate issue.

## Verification

Five new tests, each asserting on **captured names** — a tree-sitter query with a wrong node name compiles fine and captures nothing, which is how #170 stayed hidden. All five fail against the previous query (checked by reverting `C_Q` and re-running), and `all_queries_compile` alone would not have caught any of this:

- `c_header_api_surface` — the issue's unguarded probe; before: `[MAX_ITEMS, thing]`.
- `c_include_guard_is_not_a_constant` — the guarded probe; asserts `GUARDED_H` is absent *and* that the prototype/`extern`/`struct` are present through the guard.
- `c_locals_still_excluded` — `int local; char *ptr; int table[4];` in a function body stay out of `defs`.
- `c_pointer_returning_functions`, `c_valueless_define_is_dropped` (the last documents the deliberate cost).

`cargo test -p xerj-autoindex`: 225 passed. `cargo clippy --all-targets` and `cargo fmt --check` clean.

End to end against a live server (`xerj autoindex … --no-graph --fresh`), the issue's two-file probe now indexes identically whether or not it is guarded:

```
before:  plain.h    symbols=2  defs='const MAX_ITEMS\nstruct thing'
         guarded.h  symbols=3  defs='const GUARDED_H\nconst MAX_ITEMS\nstruct thing'

after:   plain.h    symbols=4  defs='const MAX_ITEMS\nstatic global_counter\nstruct thing\nfunction do_thing'
         guarded.h  symbols=4  defs='const MAX_ITEMS\nstatic global_counter\nstruct thing\nfunction do_thing'
```

And on real code — indexing memcached, `assoc.h` goes from **0 symbols to 11**, and a `function assoc_insert` symbol search returns `assoc.h` (5.06) ahead of `assoc.c` (4.26), which is the retrieval this bug was costing.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -3,7 +3,8 @@
 //! Source files used to fall through to the plain-text extractor, so a repo was
 //! searchable only as prose. This parses each file with the matching tree-sitter
 //! grammar and captures its DEFINITIONS (functions, classes, methods, structs,
-//! traits, interfaces, modules, …). Each file becomes one document carrying:
+//! traits, interfaces, modules, constants, …). Each file becomes one document
+//! carrying:
 //! - `language`  the detected language
 //! - `symbols`   a structured array of {name, kind, line}
 //! - `defs`      "kind name" per symbol, newline-joined so BM25 matches a query
@@ -246,6 +247,19 @@ const TS_Q: &str = r#"
 (variable_declarator name: (identifier) @function value: (arrow_function))
 "#;
 
+// `const_item` / `static_item` are captured because a file whose whole content
+// is a lookup table (`pub const BYTE_FREQUENCIES: [u8; 256] = [...]`) otherwise
+// extracts ZERO symbols and becomes unreachable by symbol search — see #170.
+// Those files matter disproportionately: their contents are empirical, so they
+// are exactly what cannot be recalled and must be retrieved.
+//
+// Deliberately NOT anchored to `source_file`. Measured over the 331-file
+// memchr/regex/aho-corasick corpus, the 959 const/static declarations sit at:
+// file-scope 720, impl-block associated consts 75, trait 4, function-local 160.
+// Anchoring would throw away the 79 associated consts — `impl Foo { const N }`
+// is idiomatic Rust — to suppress function-local ones, and a function-local
+// `const` in Rust is still a deliberately named constant (Rust `let` bindings
+// are a different node and are never captured), so it is signal, not noise.
 const RUST_Q: &str = r#"
 (function_item name: (identifier) @function)
 (struct_item name: (type_identifier) @struct)
@@ -254,14 +268,38 @@ const RUST_Q: &str = r#"
 (mod_item name: (identifier) @module)
 (macro_definition name: (identifier) @macro)
 (type_item name: (type_identifier) @type)
+(const_item name: (identifier) @const)
+(static_item name: (identifier) @static)
 "#;
 
+// Go's package-level `const`/`var` are anchored to `source_file` on purpose:
+// unanchored, `var_declaration` also matches every function-local `var`, which
+// fills `defs` with locals (verified — unanchoring makes `go_const_and_var`
+// fail on a captured `local`). Note the grammar is asymmetric: a grouped
+// `const ( A = 1 )` keeps `const_spec` as a direct child of `const_declaration`
+// (there is no `const_spec_list` node — a query naming one fails to compile),
+// but a grouped `var ( T = 1 )` wraps its specs in `var_spec_list`, so that
+// form needs its own pattern or package-level tables in `var (…)` are missed.
 const GO_Q: &str = r#"
 (function_declaration name: (identifier) @function)
 (method_declaration name: (field_identifier) @method)
 (type_declaration (type_spec name: (type_identifier) @type))
+(source_file (const_declaration (const_spec name: (identifier) @const)))
+(source_file (var_declaration (var_spec name: (identifier) @static)))
+(source_file (var_declaration (var_spec_list (var_spec name: (identifier) @static))))
 "#;
 
+// No constant capture here, deliberately. #170 is about files that extract zero
+// symbols and so become unreachable; Java barely has that failure mode, because
+// every file must declare a type. Measured: 7 of 1043 Java files in one corpus
+// and 1 of 72 in another extract zero symbols under this query (0.7%), against
+// 20 of 331 for Rust and 186 of 400 for C headers. A clean query does exist if
+// the recall is wanted later — `(field_declaration (modifiers "static")
+// declarator: (variable_declarator name: (identifier) @const))` captured 372
+// constants with 0 method-local false positives, where capturing all fields
+// would have pulled in 1311 including private instance state — but it is a
+// recall improvement, not this bug, and no Java corpus is wired into the
+// end-to-end check that proves it, so it is not shipped unverified.
 const JAVA_Q: &str = r#"
 (class_declaration name: (identifier) @class)
 (interface_declaration name: (identifier) @interface)
@@ -270,11 +308,36 @@ const JAVA_Q: &str = r#"
 (constructor_declaration name: (identifier) @method)
 "#;
 
+// C has #170's failure mode worse than anywhere else: 186 of 400 sampled headers
+// and 39 of 400 sampled .c files extracted zero symbols before this.
+//
+// `preproc_def` is the lever that matters — it rescued 148 of those 186 headers
+// and 13 of those 39 .c files, because a header of nothing but `#define`s is the
+// canonical C constant table. It needs no anchoring: only 319 of 7356 captures
+// (4%) sat inside a function body, and include-guard-shaped names were 298 of
+// 4192 in headers (7%). Function-like `#define F(x)` is a different node
+// (`preproc_function_def`) and is not captured.
+//
+// The object declarations ARE anchored to `translation_unit`. A bare
+// `(declaration ...)` is exactly the noisy query to avoid: measured over the
+// same 400 .c files it captured 11806 names of which 11320 (96%) were function
+// locals. Anchored, it captured 274 with zero function locals.
+//
+// Known limit of that anchoring, measured not assumed: an include guard parses
+// as `translation_unit > preproc_ifdef > declaration`, so guarded headers put
+// their declarations one level too deep and only `preproc_def` sees them. That
+// is why the anchored patterns rescued just 6 headers. Widening the anchor per
+// preprocessor form (`preproc_ifdef`, `preproc_if`, nesting) is unbounded and
+// each extra level re-admits in-function declarations, so it is left alone:
+// missing a symbol is recoverable, filling `defs` with locals is not.
 const C_Q: &str = r#"
 (function_definition declarator: (function_declarator declarator: (identifier) @function))
 (struct_specifier name: (type_identifier) @struct)
 (enum_specifier name: (type_identifier) @enum)
 (type_definition declarator: (type_identifier) @type)
+(preproc_def name: (identifier) @const)
+(translation_unit (declaration declarator: (init_declarator declarator: (identifier) @static)))
+(translation_unit (declaration declarator: (init_declarator declarator: (array_declarator declarator: (identifier) @static))))
 "#;
 
 const CPP_Q: &str = r#"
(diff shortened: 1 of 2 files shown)


xerj-org:
## ⛔ Held — duplicates an existing index on upgrade

Flagging because this is `MERGEABLE/CLEAN` with green CI, so nothing on the PR surface shows it.

Independent verification: re-indexing an existing prefix after this change **duplicates the corpus**. Live, 361-file folder, identical flags (`--no-graph --fresh`, same prefix, same state dir):

```
pre-fix binary  -> 1248 docs
this branch     -> 2295 docs   (+1047 stale, +84%)
```

Controls prove the transition is the cause, not the tool: `main->main`, `#170->#170` and `branch->branch` are all perfectly idempotent at 1248. Isolating further, the `#170 -> #172` step **alone** reproduces it, so it is commit `addb9f8` and not merely the stacked #170.

This is not specific to this PR — it is the root cause now filed as **#178**: `dataset::cluster()` groups by Jaccard similarity over the record's **field-name set**, so giving a previously symbol-less file `defs`/`symbols`/`symbol_count` moves it to a different index while the old copy survives. Every extractor improvement hits it.

### The extraction work itself is good and verified

The verifier reproduced the precision numbers almost exactly, and they are strong:

| pattern | captures | in-body false positives |
|---|---:|---:|
| prototypes | 3,871 | 14 (**99.6%**) |
| pointer-returning prototypes | 652 | 0 |
| extern identifier | 153 | 0 |
| extern array | 20 | 0 |
| pointer-returning definitions | 1,477 | 4 (**99.7%**) |

`preproc_def` 5,255 → 4,691, exactly 564 guards dropped. The issue probe now gives `plain.h` and `guarded.h` **identical** symbol sets, which is the point of the fix. One small correction: the `.h` split for prototypes is 3,290, not the 3,285 stated in the PR body.

So the C_Q work should land — it just cannot land until #178 is fixed, or it will duplicate every existing user's index on their next run.

Suggest holding this behind #178. Fixing #178 by excluding symbol fields from the clustering key would unblock both this and #170 with no deletion logic at all.

xerj-org:
## Unblocked on the duplication, pending a rebase onto #180

#178 now has a verified fix on `fix/issue-178-clustering-key` (PR #180): the autoindex
clustering key is the file-derived field names only, so an extractor that starts emitting new
names can no longer re-home a file into a new dataset. See
`engine/crates/xerj-autoindex/src/extract/mod.rs:42` (`FieldOrigin`),
`src/lib.rs:253` (only `FieldOrigin::Data` names enter the key), and `src/dataset.rs:53`
(`cluster()` takes its Jaccard over `Sketch::key_fields`).

**This was checked by actually re-indexing with this branch's extractor, not by inspection.**
`addb9f8` was built on top of the #178 fix and the `corpora/kv-oss` upgrade re-run:

| build | upgrade run | documents | datasets |
|---|---|---:|---:|
| main | main → main + `addb9f8` | 7,923 → **8,023** | 407 |
| #178 fix | fix → fix *(control)* | 7,923 → 7,923 | 405 |
| #178 fix | fix → fix + `addb9f8` | 7,923 → **7,923** | 405 |

The 100 documents main gains are the 100 C files whose headers this PR teaches the parser to
read; with the #178 fix in place those files gain their symbols **in place** and the totals do
not move. Independent verification of PR #180 re-ran this and returned no refutation.

`addb9f8` applies cleanly on top of `fix/issue-178-clustering-key` and is code-only
(`extract/code.rs`), so the rebase should be mechanical. One thing it will need: `RawRecord`
now has a required `origin` field, so any record this branch constructs has to declare
`FieldOrigin::Extractor` for its symbol fields — that is a compile error, not a silent
behaviour change, which is the point of making it required.

### What is still on this PR

Nothing beyond the rebase, on the evidence so far. The hold above was solely the duplication;
the extraction work itself was reproduced and is strong (prototypes 3,871 captures / 14
in-body false positives, pointer-returning definitions 1,477 / 4, `preproc_def` 5,255 → 4,691
= exactly 564 guards dropped, and `plain.h` / `guarded.h` giving identical symbol sets). The
one correction still outstanding is the `.h` prototype split in the PR body: 3,290, not 3,285.

### Order

#180 has to land first. Merging this before it re-introduces the 7,923 → 8,023 duplication on
every existing corpus's next re-plan.

Not merging anything here — flagging that the blocker is cleared.

xerj-org:
Unblocked: #178 is fixed and merged as `b2765153` (PR #180).

Dataset clustering no longer keys on extractor-origin fields, so gaining `defs`/`symbols`/`symbol_count` no longer re-homes a file between datasets. Verified against **this branch's own extractor commit**, not just in principle:

```
before:  main -> main + this extractor    duplicated
after:   fix  -> fix  + this extractor    idempotent
```

rust-text 53,873 → 53,873 and kv-oss 7,923 → 7,923, with the extraction still landing (code docs without `defs` 29 → 10 and 221 → 121 respectively).

Next step here is a rebase onto current main.

For #176 specifically: the migration sweep should now come **out** entirely. It was only needed to clean up the re-homing that #180 prevents at source, and it is the part that deleted a co-tenant corpus's records (569 → 467). Dropping it removes the data-loss blocker without losing anything the fix was for.

PR #176, opened 2026-08-04
[BLOCKED: data loss] fix(autoindex): extract const/static/#define, and sweep superseded records

### LINKED ISSUE #170: autoindex: Rust const/static data tables get no symbol extraction, making them unreachable by symbol search

xerj-org:
Closes #170.

Two commits. The first is the extractor fix for #170 itself; the second closes
the re-index hazard that an independent verifier found in it, which was the one
thing blocking this from being mergeable.

## 1. `5712f28` — extract const/static/#define

A source file whose entire content is a lookup table extracted ZERO symbols, so
it got no `defs`, and searching for the constant it defines returned the files
that merely *consume* it and never the file that declares it. Those files are
the worst ones to lose: a table's contents are empirical, exactly what cannot be
recalled from memory and must be retrieved.

`RUST_Q` now captures `const_item`/`static_item`, `GO_Q` captures package-level
`const`/`var` anchored to `source_file`, `C_Q` captures `preproc_def` plus
`translation_unit`-anchored declarations. Java and C# are deliberately left
alone. The anchoring decisions and the measurements behind each one are in the
commit body; the short version:

| | |
|---|---|
| rust records with no `defs` | 28 → 9 |
| `aho-corasick/src/util/byte_frequencies.rs` | no defs → `const BYTE_FREQUENCIES` |
| search "BYTE_FREQUENCIES" — defining file | rank 5 → top pair (ties `rank.rs`, which declares the same constant) |

## 2. `35b93b0` — the blocker: in-place re-index duplicated records

**Reproduced first, not taken on trust.** Corpus: memchr 2.8.3, regex 1.13.1,
regex-automata 0.4.16, regex-syntax 0.8.11, aho-corasick 1.1.4 — 298 paths, 284
unique-content files. Indexed with the pre-#170 extractor, then re-indexed into
the SAME prefix with `--fresh` and the #170 extractor:

```
live documents under the prefix     518 -> 538
the run's own summary still said    "518 records live"
```

`regex-syntax/src/unicode_tables/general_category.rs` — a pure lookup table, the
class of file #170 exists for — ended with two live copies, in two indices, from
two runs. The whole `regex-syntax-src` dataset dissolved and its index was left
live holding 20 orphaned documents that no dataset in the new plan pointed at.

Three parts had to line up:

1. `ids::doc_id` mixes the dataset slug in, so a file that moves schema group is
   written under a new `_id` in a new index; nothing overwrites the old copy.
2. Phase B's delete-before-replace clears only the file's CURRENT assignment, so
   it cannot see the dataset the file just left.
3. `--fresh` starts from an empty journal, so `cleanup_required` is empty and
   that delete does not run at all.

And `--fresh` was not avoidable. Measured on the same corpus: re-running WITHOUT
it is a complete no-op — the frozen resume plan is reused, every file is already
`file_done`, the upgraded extractor never runs, and the prefix keeps its 518
pre-fix records. **The only path that delivered #170's benefit on an existing
index was the one that duplicated.**

### The fix

One invariant, enforced once per run: no index under this prefix may hold
records of a file that this plan assigns to a different index. After every file
is published and every bulk confirmed, each index the run *inherited* — sampled
from `_cat/indices` before the run creates any of its own — gets a
`_delete_by_query` for the `ax_file` keys this plan homes elsewhere.

Same corpus, same two runs, with the fix:

```
live documents under the prefix     518 -> 518
the run reports                     "removed 20 superseded record(s)"
general_category.rs                 1 live copy
byte_frequencies.rs                 1 live copy, defs "const BYTE_FREQUENCIES"
3rd and 4th --fresh run             518, 518 — no drift
first run into an unused prefix     518, sweep skipped, nothing extra
```

Scale check, because the sweep batches keys 512 at a time and the corpus above
never crossed one batch. 700 single-record files whose dataset slug changes for
all 700 at once (`d/dee/*` renamed to `d/eff/*`, content untouched so every
`ax_file` key is unchanged), re-indexed in place with `--fresh`:

```
unpatched build   700 -> 1400 live documents   (an exact doubling; both indices live)
this branch       700 ->  700 live documents   ("removed 700 superseded record(s)")
                  whole run, sweep included:   0.64 s wall
```

Three decisions worth stating:

* **`ax_run` is not the discriminator**, though every document carries it. Run
  ids are `run-{UTC second}-{pid}`, so two runs from one process inside one
  second share one. The first cut of this sweep compared stamps and silently
  kept the stale copy; the new test caught it. The plan's own assignment cannot
  collide.
* **Scoped to this corpus's file keys**, not "everything that is not mine". The
  default prefix is `ax`, so unrelated corpora routinely share one prefix and an
  unscoped sweep would delete them. A test pins that.
* **Runs after publication, not before.** A run that dies mid-way then leaves a
  duplicate, which a rerun repairs, rather than a hole, which nothing does. And
  the sweep is not gated on the plan being freshly inferred, so the repair
  actually happens. Exercised for real — 4000 files, all migrating, `--fresh`
  run `kill -9`ed after publication but before the sweep, then resumed normally:

  ```
  after the kill      8000 live documents (4000 old + 4000 new, both indices)
  after the resume    "removed 4000 superseded record(s)" -> 4000 live
  ```

  Gating the sweep on a freshly inferred plan would have left those 4000
  duplicates permanently, since a resume reuses the frozen plan.

Cost: zero requests when nothing was inherited — a first publication into an
unused prefix pays nothing. Otherwise one delete-by-query per inherited index
per 512 keys, measured at **0.2 ms each** when it matches nothing (50 calls,
512-term filter, live server), so even a re-index with 50 inherited indices and
10k files is well under a second of added round trips. It cannot be collapsed
into a single wildcard call: verified against a live server, both `prefix-*` and
`a,b` return 404 `index_not_found_exception` from `_delete_by_query`.

### Tests

`crates/xerj-autoindex/src/reindex_migration_tests.rs`, against a mock endpoint
that keys documents by `(index, _id)` — a mock keyed by `_id` alone merges the
two copies and passes with the bug open.

* `in_place_reindex_after_a_dataset_migration_does_not_duplicate` — provokes the
  migration the way the real one happens (a second file with a disjoint field
  set splits the cluster and renames both slugs), re-indexes in place with
  `--fresh`, asserts the document count does not grow and that the file actually
  left its original index, so the test cannot pass vacuously. **Verified failing
  without the sweep: 4 live copies of a 2-record file instead of 2.**
* `reindex_leaves_another_corpus_under_the_same_prefix_alone` — a second corpus
  sharing the prefix keeps its records.

Full suite: `cargo test -p xerj-autoindex` 222 passed, 0 failed. `cargo fmt
--check` and `cargo clippy --all-targets -- -D warnings` clean. `cargo check -p
xerj-server` passes (`Es::delete_by_query` now returns the number of documents
removed).

### Known limits, disclosed rather than papered over

* `ax_file` is derived from CONTENT, so `--fresh` gives an EDITED file a new key
  and its pre-edit records are no longer anything the corpus claims. Measured
  with one file edited between two `--fresh` runs: **518 → 519, identical on the
  unpatched build** — pre-existing, not introduced here, and not fixed here.
  `autoindex` now prints this as a `note:` whenever `--fresh` runs against a
  prefix that already has indices, and points at deleting the `{prefix}-*`
  indices for a guaranteed-clean rebuild. Resuming (no `--fresh`) has no such
  gap: the resume plan carries the old key forward and the file replaces itself
  in place.
* Same shape, same reason: a file the new plan classifies as junk (oversized,
  unparseable) is not in `plan.files` either, so its earlier records also stay.
  Sweeping `plan.junk_files` keys instead would be actively unsafe — a
  key-collision divert records the *owner's* key in `junk_files`, so it would
  delete the legitimate owner's records.
* An index the sweep empties stays live at 0 documents. Autoindex has no
  delete-index path and adding one is a larger, destructive change.
* The catalog's `ds:{slug}` document for a dissolved dataset also survives with
  a stale `run_id`. It sits outside `{prefix}-*` so it affects no document
  count, and the catalog id is global across prefixes, so cleaning it needs its
  own design.

### DIFF
--- a/engine/crates/xerj-autoindex/src/esclient.rs
+++ b/engine/crates/xerj-autoindex/src/esclient.rs
@@ -318,12 +318,16 @@ impl Es {
     /// locator set alongside the replacement. The server executes ONE bounded
     /// search-and-delete pass per call (size-capped at 10k docs), so a single
     /// response is not complete removal: repeat until a pass deletes nothing.
-    pub fn delete_by_query(&self, index: &str, query: &Value) -> Result<()> {
+    /// Returns the total number of documents removed across the passes.
+    pub fn delete_by_query(&self, index: &str, query: &Value) -> Result<u64> {
         const MAX_PASSES: usize = 1_000;
+        let mut removed = 0u64;
         for _ in 0..MAX_PASSES {
-            if self.delete_by_query_pass(index, query)? == 0 {
-                return Ok(());
+            let pass = self.delete_by_query_pass(index, query)?;
+            if pass == 0 {
+                return Ok(removed);
             }
+            removed += pass;
         }
         Err(anyhow!(
             "POST /{index}/_delete_by_query still reported deletions after {MAX_PASSES} passes; \
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -3,7 +3,8 @@
 //! Source files used to fall through to the plain-text extractor, so a repo was
 //! searchable only as prose. This parses each file with the matching tree-sitter
 //! grammar and captures its DEFINITIONS (functions, classes, methods, structs,
-//! traits, interfaces, modules, …). Each file becomes one document carrying:
+//! traits, interfaces, modules, constants, …). Each file becomes one document
+//! carrying:
 //! - `language`  the detected language
 //! - `symbols`   a structured array of {name, kind, line}
 //! - `defs`      "kind name" per symbol, newline-joined so BM25 matches a query
@@ -242,6 +243,19 @@ const TS_Q: &str = r#"
 (variable_declarator name: (identifier) @function value: (arrow_function))
 "#;
 
+// `const_item` / `static_item` are captured because a file whose whole content
+// is a lookup table (`pub const BYTE_FREQUENCIES: [u8; 256] = [...]`) otherwise
+// extracts ZERO symbols and becomes unreachable by symbol search — see #170.
+// Those files matter disproportionately: their contents are empirical, so they
+// are exactly what cannot be recalled and must be retrieved.
+//
+// Deliberately NOT anchored to `source_file`. Measured over the 331-file
+// memchr/regex/aho-corasick corpus, the 959 const/static declarations sit at:
+// file-scope 720, impl-block associated consts 75, trait 4, function-local 160.
+// Anchoring would throw away the 79 associated consts — `impl Foo { const N }`
+// is idiomatic Rust — to suppress function-local ones, and a function-local
+// `const` in Rust is still a deliberately named constant (Rust `let` bindings
+// are a different node and are never captured), so it is signal, not noise.
 const RUST_Q: &str = r#"
 (function_item name: (identifier) @function)
 (struct_item name: (type_identifier) @struct)
@@ -250,14 +264,38 @@ const RUST_Q: &str = r#"
 (mod_item name: (identifier) @module)
 (macro_definition name: (identifier) @macro)
 (type_item name: (type_identifier) @type)
+(const_item name: (identifier) @const)
+(static_item name: (identifier) @static)
 "#;
 
+// Go's package-level `const`/`var` are anchored to `source_file` on purpose:
+// unanchored, `var_declaration` also matches every function-local `var`, which
+// fills `defs` with locals (verified — unanchoring makes `go_const_and_var`
+// fail on a captured `local`). Note the grammar is asymmetric: a grouped
+// `const ( A = 1 )` keeps `const_spec` as a direct child of `const_declaration`
+// (there is no `const_spec_list` node — a query naming one fails to compile),
+// but a grouped `var ( T = 1 )` wraps its specs in `var_spec_list`, so that
+// form needs its own pattern or package-level tables in `var (…)` are missed.
 const GO_Q: &str = r#"
 (function_declaration name: (identifier) @function)
 (method_declaration name: (field_identifier) @method)
 (type_declaration (type_spec name: (type_identifier) @type))
+(source_file (const_declaration (const_spec name: (identifier) @const)))
+(source_file (var_declaration (var_spec name: (identifier) @static)))
+(source_file (var_declaration (var_spec_list (var_spec name: (identifier) @static))))
 "#;
 
+// No constant capture here, deliberately. #170 is about files that extract zero
+// symbols and so become unreachable; Java barely has that failure mode, because
+// every file must declare a type. Measured: 7 of 1043 Java files in one corpus
+// and 1 of 72 in another extract zero symbols under this query (0.7%), against
+// 20 of 331 for Rust and 186 of 400 for C headers. A clean query does exist if
+// the recall is wanted later — `(field_declaration (modifiers "static")
+// declarator: (variable_declarator name: (identifier) @const))` captured 372
+// constants with 0 method-local false positives, where capturing all fields
+// would have pulled in 1311 including private instance state — but it is a
+// recall improvement, not this bug, and no Java corpus is wired into the
+// end-to-end check that proves it, so it is not shipped unverified.
 const JAVA_Q: &str = r#"
 (class_declaration name: (identifier) @class)
 (interface_declaration name: (identifier) @interface)
@@ -266,11 +304,36 @@ const JAVA_Q: &str = r#"
 (constructor_declaration name: (identifier) @method)
 "#;
 
+// C has #170's failure mode worse than anywhere else: 186 of 400 sampled headers
+// and 39 of 400 sampled .c files extracted zero symbols before this.
+//
+// `preproc_def` is the lever that matters — it rescued 148 of those 186 headers
+// and 13 of those 39 .c files, because a header of nothing but `#define`s is the
+// canonical C constant table. It needs no anchoring: only 319 of 7356 captures
+// (4%) sat inside a function body, and include-guard-shaped names were 298 of
+// 4192 in headers (7%). Function-like `#define F(x)` is a different node
+// (`preproc_function_def`) and is not captured.
+//
+// The object declarations ARE anchored to `translation_unit`. A bare
+// `(declaration ...)` is exactly the noisy query to avoid: measured over the
+// same 400 .c files it captured 11806 names of which 11320 (96%) were function
+// locals. Anchored, it captured 274 with zero function locals.
+//
+// Known limit of that anchoring, measured not assumed: an include guard parses
+// as `translation_unit > preproc_ifdef > declaration`, so guarded headers put
+// their declarations one level too deep and only `preproc_def` sees them. That
+// is why the anchored patterns rescued just 6 headers. Widening the anchor per
+// preprocessor form (`preproc_ifdef`, `preproc_if`, nesting) is unbounded and
+// each extra level re-admits in-function declarations, so it is left alone:
+// missing a symbol is recoverable, filling `defs` with locals is not.
 const C_Q: &str = r#"
 (function_definition declarator: (function_declarator declarator: (identifier) @function))
 (struct_specifier name: (type_identifier) @struct)
 (enum_specifier name: (type_identifier) @enum)
 (type_definition declarator: (type_identifier) @type)
+(preproc_def name: (identifier) @const)
+(translation_unit (declaration declarator: (init_declarator declarator: (identifier) @static)))
+(translation_unit (declaration declarator: (init_declarator declarator: (array_declarator declarator: (identifier) @static))))
 "#;
 
 const CPP_Q: &str = r#"
@@ -297,6 +360,11 @@ const PHP_Q: &str = r#"
 (trait_declaration name: (name) @trait)
 "#;
 
+// Also no constant capture, for a blunter reason than Java: there were 4 C#
+// files on the machine this was measured on, so there is no corpus to measure a
+// candidate query's precision against. Shipping one would be a guess, and the
+// C numbers above show how wrong that guess can be (96% locals from a node type
+// that looks perfectly reasonable on paper).
 const CSHARP_Q: &str = r#"
 (class_declaration name: (identifier) @class)
 (interface_declaration name: (identifier) @interface)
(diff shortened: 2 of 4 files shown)


xerj-org:
## ⛔ DO NOT MERGE — this loses user data

Flagging loudly because this PR is `MERGEABLE/CLEAN` with all 10 checks green, so nothing on the PR surface indicates the problem.

Independent verification found that the migration sweep added in `35b93b0` **silently deletes records belonging to a different corpus that shares the prefix**, whenever the two corpora contain any byte-identical file. Introduced by this PR, undisclosed, and reachable on ordinary runs with no `--fresh`.

### Mechanism

`ids::file_key` is **content-only** — `xxh3_64(first 64KB) ‖ size`, with the path deliberately not included (`engine/crates/xerj-autoindex/src/ids.rs:14`). So one file present in two corpora has **one** `ax_file` key. `sweep_migrated_records` builds `homes` from *this* plan's files, then for every inherited index deletes every plan key not homed there — which includes keys whose records in that index belong to somebody else's corpus.

### Live reproduction (real files, real server, no `--fresh` anywhere)

```
index the 5-crate tree (284 files)                 ->  518 live
index one of its subfolders under the SAME prefix  ->  518 live, "removed 51 superseded record(s)"
ordinary resume of the parent (the obvious repair) ->  467 live, "removed 51 superseded record(s)"
```

Net: **569 legitimate records → 467.** Fifty-one files end up searchable in neither index. Exit code 0, no warning, and the message blames a "dataset move" that never happened. The frozen resume plan marks every file `file_done`, so a resume cannot restore them — only `--fresh` can, which this branch documents as the only path that re-publishes.

Same three runs against a sweep-neutered build of this same commit: **518 / 569 / 569**. Nothing lost. That A/B is what establishes this PR as the cause.

### Why this is not a corner case

The default prefix is `ax` — which is this PR's own stated justification for scoping the sweep — and byte-identical files across corpora are normal. In the test corpus alone, `LICENSE-MIT`, `LICENSE-APACHE`, `COPYING`, `UNLICENSE`, `rustfmt.toml`, `perl_word.rs` and `macros.rs` are shared across crates; that is exactly what the run's own "14 duplicate aliases" is counting. Indexing a folder and later a subfolder of it under the default prefix is enough to trigger it.

### The existing test cannot catch it

`reindex_leaves_another_corpus_under_the_same_prefix_alone` uses corpora with **disjoint content**, so it is structurally incapable of failing on this. The stated safety property — "scoped to this corpus's own file keys" — does not hold, because a content-only key is not corpus-scoped.

A minimal reproducer was written and reverted: `probe_shared_prefix_identical_file_content`, two 1-file corpora whose single file is byte-identical in different directories, ordinary runs, shared prefix. It fails on this branch (corpus A's record deleted, its index emptied to 0 docs) and passes without the sweep.

### To unblock

The sweep needs a corpus discriminator the plan actually owns:

- stamp each document with the run's root/corpus identity and require it to match before deleting, **or**
- restrict the sweep to indices this `(root, prefix)` pair has itself written, per the catalog/journal,

plus a regression test whose two corpora **share** a file — the current one cannot fail.

### What is genuinely good here and worth keeping

The duplication fix itself works and is well evidenced: in-place re-index 518 → 518 with "removed 20 superseded record(s)" where it previously went 518 → 538; the 700-file slug-change case 700 → 700 instead of an exact doubling; and a crash case (`kill -9` mid-`--fresh`, 8000 live) repaired to 4000 by an ordinary resume. Sweep cost measured at 0.2 ms per no-op call. None of that is in question — only the missing corpus scoping.

xerj-org:
## The duplication is fixed upstream. The data loss on this PR is not — drop the sweep, then rebase onto #180.

Two separate problems have been conflated on this PR. One is now solved elsewhere; the other
is still here.

### 1. Duplication on upgrade — solved by #180, verified with *this branch's* extractor

#178 has a verified fix on `fix/issue-178-clustering-key` (PR #180): the clustering key is now
the file-derived field names only, so gaining `defs`/`symbols`/`symbol_count` can no longer
move a file into a different dataset. See
`engine/crates/xerj-autoindex/src/extract/mod.rs:42` (`FieldOrigin`), `src/lib.rs:253` (only
`FieldOrigin::Data` names enter the key), `src/dataset.rs:53` (`cluster()` compares
`Sketch::key_fields`).

**Checked by re-indexing with `5712f28` itself, not by inspection.** Independent verification
built `main + 5712f28` and `fix + 5712f28` and ran the upgrade both ways on `corpora/rust-text`
(846 files, `--no-semantic --no-graph`, `--fresh`):

| build | upgrade run | documents | datasets |
|---|---|---:|---:|
| main | main → main + `5712f28` | 53,873 → **53,902** | 9 → 10 |
| #178 fix | fix → fix *(control)* | 53,873 → 53,873 | 7 |
| #178 fix | fix → fix + `5712f28` | 53,873 → **53,873** | 7 |
| #178 fix | fix + `5712f28` → fix *(downgrade)* | 53,873 → 53,873 | 7 |

The 29 documents main gains are the 29 code files that had no `defs`: 19 joined the
symbol-bearing dataset, 10 formed a brand-new index (`r1-memchr-fuzz`), and all 29 originals
stayed behind. With #180 in place those files gain their symbols **in place** — code documents
with no `defs` go 29 → 10 of 328 while the totals and dataset counts above do not move — and
the downgrade direction is stable too.

### 2. The migration sweep still loses user data — and #180 does not fix it

`sweep_migrated_records` is unaffected by any of the above. The mechanism stands exactly as
flagged: `ids::file_key` is content-only (`engine/crates/xerj-autoindex/src/ids.rs:14`,
`xxh3_64(first 64KB) ‖ size`, path deliberately excluded), so one file present in two corpora
has one key, and deleting every plan key not homed in an inherited index deletes records
belonging to somebody else's corpus. Live, no `--fresh` anywhere: 569 legitimate records →
467, exit code 0, no warning. The sweep-neutered build of the same commit gave 518 / 569 / 569.

PR #180 makes this explicit and is worth quoting, because it is the reason a sweep cannot be
repaired rather than removed: a content-only key cannot tell "this record moved" from "this
record belongs to another corpus". **With the clustering stable there is nothing left to
sweep** — the sweep existed to clean up re-homing that #180 prevents from happening.

### What to do

1. Land #180 first.
2. **Drop `sweep_migrated_records` from this branch entirely**, along with the "removed N
   superseded record(s)" message and the documentation that names `--fresh` as the only
   re-publish path.
3. Rebase the remaining code-only `extract/code.rs` change onto
   `fix/issue-178-clustering-key`. It applies cleanly. It will need `origin:
   FieldOrigin::Extractor` on the records it constructs — a compile error, by design, not a
   silent change.

Until step 2 is done this stays **DO NOT MERGE**: #180 clears the duplication that held this
PR, it does not clear the data loss this PR introduces.

Not merging anything here.

xerj-org:
Unblocked: #178 is fixed and merged as `b2765153` (PR #180).

Dataset clustering no longer keys on extractor-origin fields, so gaining `defs`/`symbols`/`symbol_count` no longer re-homes a file between datasets. Verified against **this branch's own extractor commit**, not just in principle:

```
before:  main -> main + this extractor    duplicated
after:   fix  -> fix  + this extractor    idempotent
```

rust-text 53,873 → 53,873 and kv-oss 7,923 → 7,923, with the extraction still landing (code docs without `defs` 29 → 10 and 221 → 121 respectively).

Next step here is a rebase onto current main.

For #176 specifically: the migration sweep should now come **out** entirely. It was only needed to clean up the re-homing that #180 prevents at source, and it is the part that deleted a co-tenant corpus's records (569 → 467). Dropping it removes the data-loss blocker without losing anything the fix was for.

xerj-org:
Closing as superseded — resolved a better way than this branch's sweep.

The duplication this PR's sweep worked around (a file re-homing to a new dataset when the extractor starts capturing more symbols) is now prevented **at the source** by #180's `FieldOrigin` guard on `main`: extractor-invented fields (`defs`, `symbols`) are kept out of the clustering key, so a file's inferred dataset no longer shifts when #170's extractor captures a `const`/`#define`.

With #180 on main, this branch's const/static extractor rebases cleanly and re-indexes **count-stable** (measured: a C corpus of const/#define/extern/prototype, 3 records → 3 records on re-index, no `+stale`). So the `35b93b0` migration sweep is no longer needed — and its delete scope (`indices_under_prefix()`, a whole-prefix wildcard) was the cross-corpus data-loss hazard called out on #172, which we do **not** want on main.

The valuable half of this branch — the `#170` const/static extractor — landed via **#175** (rebased onto #180, tests green, no duplication, extraction verified). So #170 and #172 are both fixed on `main` now without the sweep.

Thanks for the careful root-cause work here; the sweep was the right instinct before #180 fixed re-homing structurally.

PR #179, opened 2026-08-04
[HELD: unbounded scan with deletes] fix(search): a highlight block must not change _score or hit order (#177)

### LINKED ISSUE #177: search: adding a highlight block changes _score and reorders hits (6/6 -> 1/6 top-1 on a labelled set)
## Adding a `highlight` block changes `_score` and reorders results

`highlight` is a presentation concern. It must not affect scoring or ranking. On this build it does both, deterministically.

### Reproduce

Index any corpus, then issue the **same query twice**, differing only by the presence of a `highlight` block.

```sh
Q='{"multi_match":{"query":"addReplyNull null bulk string reply","fields":["body","defs","title"]}}'

# A — no highlight
curl -s "$URL/$IDX/_search" -H 'Content-Type: application/json' \
  -d "{\"size\":3,\"_source\":[\"ax_path\"],\"query\":$Q}"

# B — identical query plus a highlight block
curl -s "$URL/$IDX/_search" -H 'Content-Type: application/json' \
  -d "{\"size\":3,\"_source\":[\"ax_path\"],\"query\":$Q,
       \"highlight\":{\"fields\":{\"body\":{\"fragment_size\":320}}}}"
```

Observed (valkey + memcached corpus, 7,923 records):

```
A — WITHOUT highlight            B — WITH highlight
  12.68  valkey/src/networking.c   20.43  valkey/src/t_list.c
  11.84  valkey/src/t_zset.c       19.96  valkey/src/t_hash.c
  11.42  valkey/src/module.c       19.84  valkey/src/geo.c
```

Not a tie-break wobble: **every score changes** (12.68 → 20.43) and the result set is disjoint. Identical across 3 consecutive runs, so it is deterministic rather than flaky.

Reproduced with `highlight.fields` naming a concrete field (`body`) and with the `"*"` wildcard form — both shift the ranking identically, so it is not specific to wildcard expansion.

Also reproduced with the query wrapped in `bool.must`, which on its own changes nothing: A and B above score identically to their `bool.must` equivalents, so the wrapper is not implicated. The only variable that moves the ranking is the presence of `highlight`.

### Impact

This is severe for any client that displays snippets, because asking for snippets silently changes which documents you get back.

Measured on a 6-query hand-labelled set over this corpus:

| query form | top-1 | top-3 |
|---|---:|---:|
| without `highlight` | **6/6** | **6/6** |
| with `highlight` | **1/6** | **2/6** |

A retrieval client that requests highlights — the obvious thing to do, since the point is to show the matching passage — gets materially worse results and has no way to know. It looks like a relevance problem in the corpus or the query, not a scoring bug.

### Expected

`highlight` should be applied to the already-selected, already-scored hits. Adding it must leave `hits.hits[*]._score` and `hits.hits[*]._id` unchanged.

### Note on where to look

The score inflation (12.68 → 20.43) suggests the highlighter's field resolution is feeding back into query rewriting or field selection rather than running as a post-pass over the result set — the "with highlight" ranking looks like a different field set was searched, not like the same search rendered differently.

Found while building a reference-coding retrieval client; the client had requested highlights from the start, so every retrieval measurement taken through it was affected.

xerj-org:
Fixes #177.

## Reproduction (before this PR)

rc.11 build, valkey+memcached corpus, index `xc-kv-oss-valkey-deps-3`, 616 matches:

```
Q = {"multi_match":{"query":"addReplyNull null bulk string reply",
                    "fields":["body","defs","title"]}}

A  size:5, no highlight        12.6832  11.8398  11.4218  11.3757  11.2564
B  size:5, + highlight.body    20.4350  19.9605  19.8379  19.1846  17.2884
T  size:700 (all 616), top-5   20.4350  19.9605  19.8379  19.1846  17.2884
```

Every score differs, the hit sets are disjoint, and `A` is not the global
top-5 — `B` is. So the *default* page was the wrong one, and asking for
snippets accidentally corrected it.

The same defect is visible with no highlight anywhere in the body: on that
corpus `size:100` reports a top hit of `12.6832` and `size:155` reports
`20.4350`.

## Mechanism

`engine/crates/xerj-engine/src/index.rs`.

1. `highlight` was one of the shapes excluded from the page-cap gate
   (`let materialisation_limit: usize = if mem_doc_count == 0 && …`), so a
   `size:5` body got a cap of `from+size = 5` without it and `256` with it.

2. That should not have been observable. It was, because the cross-segment
   merge was arrival-order. `search_bounded` returns each segment's top-k by
   score; the merge pushed them into `all_hits` with a plain
   `if all_hits.len() >= materialisation_limit { break }` and an outer
   `all_hits.len() < materialisation_limit` guard. The first segments visited
   filled the cap and every later segment was dropped whole, regardless of
   score. The page was "the best hits of whichever segments happened to fit".

So the cap decided which hits won, and `highlight` decided the cap.

## Fix

**The merge now selects the global top-k by score.** `page_worst` is the
lowest score currently held (defined only once the collector is at its cap —
below the cap nothing has to be displaced). A segment is opened when its best
hit beats that; its descending run is walked until a hit cannot; the overflow
is trimmed back to the cap with the comparator the final page sort already
uses (score DESC, `seq_no` ASC, `_id` ASC), so admission and final ordering
agree. Memory stays O(cap), 2×cap transiently. A segment that cannot beat the
worst kept hit is skipped exactly as before — the common "first segments
dominate" case does no extra work.

**`highlight` is removed from all three execution gates**, so a body now runs
the *same plan* with and without it rather than a different plan that happens
to agree:

| gate | why it was wrong |
|---|---|
| page-cap `materialisation_limit` | highlighting is a post-pass; it has no bearing on how many candidates the scan needs |
| `need_sources_for_post` | at `size:0` there is no page to highlight, so it should not force source materialisation |
| `function_score_fast_fvf` | `function_score_columnar` hydrates `_source` for the winners, which is all `apply_highlight` reads — the same argument `scored_fast_plan` already makes for admitting `highlight` |

## Tests

`engine/crates/xerj-engine/tests/highlight_does_not_change_ranking.rs`, three
tests. Each builds two segments where the second holds the genuinely
top-scoring documents and the first holds enough matches to fill a small page
on its own.

* `highlight_block_does_not_change_scores_or_order` — same body twice, once
  with a highlight block; asserts identical `(_id, _score)` pairs in identical
  order, identical `hits.total`, that the page equals the full-materialisation
  top-5, that the highlighted run returns fragments containing
  `<em>quicklist</em>`, and that the plain run carries no `highlight` key.
* `highlight_block_does_not_change_multi_match_ranking` — the issue's
  `multi_match` shape.
* `top_hit_is_stable_across_page_sizes` — `size:1/2/5/10/50` must all return
  the `size:200` top hit. This is the same defect seen from the other side.

All three FAIL on `main` and pass here:

```
main:  0 passed; 3 failed
  top_hit_is_stable_across_page_sizes
    left:  ["weak03", 0.07465832680463791]      right: ["best", 0.39556291699409485]
  highlight_block_does_not_change_multi_match_ranking
    left:  [("weak03",0.0873), ("weak04",0.0873), ("weak08",0.0873)]
    right: [("strong0",0.5503), ("strong1",0.5448), ("strong2",0.5311)]
  highlight_block_does_not_change_scores_or_order
    left:  [("weak55",0.2899), ("weak03",0.0877), …]
    right: [("strong0",0.5696), ("strong1",0.5622), …]

this branch:  3 passed; 0 failed
```

Rest of the suite: `cargo test -p xerj-engine --no-fail-fast` — 412 unit tests
and every integration binary green, 0 failures. `cargo fmt` clean,
`cargo clippy -p xerj-engine --all-targets` clean.

## Live verification

Two servers on the same corpus (a copy of the `xc-kv-oss-*` data dir): rc.11
`main` build vs this build.

```
                       A==B    A == size:700 top-5
before (main)          false   false
after  (this branch)   true    true
```

Across 9 query shapes — `multi_match`, `match`, `match_phrase`, `bool should`,
`bool must+filter`, `dis_max`, `prefix`, `query_string`, `match_all` —
`highlight` changed the page on **3 of 9 before, 0 of 9 after**. `hits.total`
is unchanged on all 9. Highlighting still works: 3 fragments returned, all 3
carrying `<em>`-marked matches, e.g.

```
"...* if we need to convert a listpack to a quicklist. Note that we only check
 <em>string</em>\n * encoded objects as their string length can be queried..."
```

Warm median latency for the issue's query (9 samples after 3 warm-ups):

| | before | after |
|---|---|---|
| no highlight | 163.1 ms | 32.5 ms |
| with highlight | 27.2 ms | 30.5 ms |

The plain path got faster because it no longer takes the narrow-cap route that
reopened segments to no purpose; both shapes now run the same plan, so they
land on the same number.

## Known limits — not fixed here

* This PR makes the `size:5` page equal the `size:700` top-5 for `multi_match`,
  `match` and `match_phrase` (it did not before). Five of the nine shapes —
  `bool should`, `bool must+filter`, `dis_max`, `prefix`, `query_string` —
  still do not, **both before and after** this change. Those go through other
  executor arms (stored-doc scan, columnar scored family) whose selection this
  PR does not touch. `match_all` was already correct.
* `bool should` additionally scores the *same document* differently at
  different page sizes (`5cfb36ea…` scores 1.2497 at `size:5` and 4.8065 at
  `size:700`, on the unmodified `main` build). That is a separate pre-existing
  defect — the bool IDF rescore runs over the materialised window — and since
  this PR changes which hits that window holds, that query's page moves. Its
  top-1 becomes the global top-1; the rest of its page is not comparable across
  cap values while the score itself depends on the cap.
* Which of several *exactly tied* documents survives a top-k truncation is
  still decided by arrival, not by `seq_no`. Ties are strictly not displaced
  (`sh.score > worst`), matching Lucene's "lowest doc id wins a tie" only when
  segments are visited in doc-id order. The new tests deliberately use
  tie-free rankings so they assert selection, not tie-break.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -13020,16 +13020,20 @@ impl Index {
 
         // Determine whether we need hit *sources* at all.
         // - size > 0          → sources for hits output
-        // - rescore/highlight → mutate source after collection
+        // - rescore           → mutates source after collection
         // - collapse          → needs source field
         // - sort on a field   → needs source
         // - aggs without DV fast path → needs sources
         // When none of these apply (classic `size:0 + match_all + track_total_hits`
         // counting query), we skip materialisation entirely and just count.
+        //
+        // `highlight` is deliberately NOT in this list (#177): it is a pure
+        // post-pass over the rendered page (`apply_highlight`), so at `size: 0`
+        // — where there is no page — it has nothing to do, and a `size: 0` body
+        // must execute identically with and without it.
         let need_hits_output: bool = size > 0;
         let need_sources_for_post: bool = need_hits_output
             || !request.rescore.is_empty()
-            || request.highlight.is_some()
             || request.collapse.is_some()
             || request
                 .sort
@@ -14088,11 +14092,20 @@ impl Index {
         // `from+size` stored-doc parses (live: the 256-doc floor was ~60% of
         // the range/term page cost).  Every excluded shape (sorts, aggs —
         // the brute agg fallback aggregates the materialised sources —
-        // rescore/collapse/highlight/explain, min_score, count-only,
+        // rescore/collapse/explain, min_score, count-only,
         // pinned's phase-4 overlay headroom) keeps the original slack
         // byte-for-byte.  Shadowed AFTER the memtable snapshot: the gate
         // requires the memtable be empty, so the earlier `fetch_limit` uses
         // are untouched.
+        //
+        // `highlight` was in the excluded list until #177.  It never belonged
+        // there — highlighting is a post-pass over the rendered page — and
+        // while the segment merge was arrival-order it made this cap
+        // observable: a body with `highlight` got 256 candidates, the same
+        // body without it got `from+size`, and the two pages carried
+        // different `_id`s and different `_score`s.  The merge now selects the
+        // global top-k by score (see the FTS admission below), so the cap
+        // controls only how much work is done, never which hits win.
         let materialisation_limit: usize = if mem_doc_count == 0
             && !deletes_present
             && size > 0
@@ -14101,7 +14114,6 @@ impl Index {
             && request.aggs.is_none()
             && request.rescore.is_empty()
             && request.collapse.is_none()
-            && request.highlight.is_none()
             && request.min_score.is_none()
             && !request.explain
             && !matches!(query, QueryNode::Pinned { .. })
@@ -14562,13 +14574,53 @@ impl Index {
                                     }
                                 }
                                 // Materialise sources for the top hits only.
-                                // `seg_hits` is score-sorted descending, so taking
-                                // the prefix that fits under `materialisation_limit`
-                                // preserves top-k ordering/scoring.
+                                // `seg_hits` is score-sorted descending, so the
+                                // prefix we take is this SEGMENT's top-k.
+                                //
+                                // #177 — across segments the collector used to be
+                                // a plain FIFO capped at `materialisation_limit`:
+                                // the first segments visited filled it and every
+                                // later segment was dropped whole, however well
+                                // its documents scored.  The page was therefore
+                                // "the best hits of whichever segments happened to
+                                // fit", not the global top-k, and it moved
+                                // whenever the cap moved — which is what made a
+                                // `highlight` block (which only widened the cap)
+                                // rewrite every `_score` and `_id`.
+                                //
+                                // `page_worst` is the lowest score currently kept,
+                                // defined only once the collector is at its cap
+                                // (below the cap nothing has to be displaced).  A
+                                // segment is opened when it can beat that, and its
+                                // descending run is walked until it can't; the
+                                // overflow is trimmed back to the cap by score
+                                // right after.  Bounded by 2×cap in the worst case.
+                                let page_worst: Option<f32> = if sort_topk.is_none()
+                                    && !count_only
+                                    && all_hits.len() >= materialisation_limit
+                                {
+                                    all_hits.iter().map(|h| h.score).fold(
+                                        None,
+                                        |acc: Option<f32>, s| {
+                                            Some(match acc {
+                                                Some(a) if a <= s => a,
+                                                _ => s,
+                                            })
+                                        },
+                                    )
+                                } else {
+                                    None
+                                };
+                                let seg_can_enter = sort_topk.is_some()
+                                    || all_hits.len() < materialisation_limit
+                                    || seg_hits
+                                        .first()
+                                        .zip(page_worst)
+                                        .is_some_and(|(sh, worst)| sh.score > worst);
                                 if !fts_sorted_handled
                                     && !count_only
                                     && !seg_hits.is_empty()
-                                    && all_hits.len() < materialisation_limit
+                                    && seg_can_enter
                                 {
                                     // F2 RANDOM-ACCESS source hydration.  The
                                     // scored FTS page needs `_source` for only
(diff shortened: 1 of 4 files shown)


xerj-org:
## ⛔ Held — undisclosed unbounded scan on any index with deletes

Flagging on the PR surface because this is `MERGEABLE/CLEAN` with all 10 checks green, so
nothing here shows the problem.

The reported defect is real and this branch does fix it: `A == B == size:700 top-5` on the
issue's query, and `highlight` changes the page on 0 of 9 query shapes after, 3 of 9 before.

What holds it is a behaviour change the PR does not mention. `deletes_present`
(`index.rs:14084`) is `any(has_tombstones) || version_map.ghost_events() > 0`, and
`ghost_events()` is monotonic — so it is permanently true on any index that has ever taken an
overwrite or a delete. On that path `fts_cap = usize::MAX` (`index.rs:14461`), `page_worst`
is `None` while the collector is below its cap (`index.rs:14598`), the new early-out
(`index.rs:14718`) fires only when `page_worst` is `Some`, and the trim (`index.rs:14792`)
runs only after the whole segment run is walked. The guard this replaced
(`all_hits.len() >= materialisation_limit`) bounded the walk regardless of `fts_cap`; the new
one does not, so a segment's entire match set is hydrated before anything is trimmed —
O(matches), not O(cap), and measured at roughly an 8× latency regression.

That falsifies "Memory stays O(cap), 2×cap transiently" in the PR body, the same sentence in
the commit body, and the new comment at `index.rs:14597`. None of the three new tests can
catch it: they never issue a delete or an overwrite, so `deletes_present` is false throughout.

A second, separate change: `seg_can_enter` at `index.rs:14614` is unconditionally true when
`sort_topk.is_some()`, where the replaced guard had no such exemption — see the issue for why
that matters on the field-sorted fall-back path.

Full detail and the four things that would clear the hold: https://github.com/xerj-org/xerj/issues/177#issuecomment-5185340236

Issue #191, opened 2026-08-07
Tied scores: a bounded page and a full page disagree on which tied document is returned


xerj-org:
Found while fixing #188. Splitting it out because it is independent and pre-existing.

When several documents score EXACTLY equal, a bounded page and a full materialisation disagree on which of the tied documents comes back.

It matters more after #188 than before. Making BM25 statistics index-wide removes per-segment normalisation artefacts, and one consequence is that genuinely-identical documents now genuinely tie instead of being separated by an accident of which segment they landed in. So exact ties get more common, not less.

Concrete example from the #188 work, in `highlight_does_not_change_ranking.rs`: four documents carried a constant `"title": "listpack"`. With index-wide stats the body IDF collapses to `ln(1 + 0.5/44.5) = 0.0112` (all 44 docs have the term in `body`) while the title IDF is `ln(10) = 2.303`, so `multi_match`'s `best_fields` max picks `title` — and all four docs land on exactly 2.859662, a 4-way tie straddling a `size:3` page boundary. Which three you get back depends on the collection path.

The fixture there was changed to vary the title so the test has the tie-free ordering it needs to test its own subject. That is a workaround in one test, not a fix.

What is needed is a total order: score first, then a deterministic tiebreak that both the bounded collector and the full path apply identically. Lucene ties by doc id within a leaf and by leaf order across leaves, so a tie is resolved the same way regardless of how many hits were collected. XERJ has the ingredients already, since the bounded collector sorts by `(score, seq_no, id)` in places, but the two paths do not agree.

Not blocking rc.12. Worth fixing before anyone relies on stable pagination over a corpus with duplicate or near-duplicate documents, which is exactly what a re-indexed or synced corpus looks like.

xerj-org:
Reopening: this issue was auto-closed by the merge of #267, which #267 explicitly said must not happen ("#191 must stay open until #270 closes"). The close was my mistake, and it was in prose, not in the code.

[...]

PR #192, opened 2026-08-07
fix(search): BM25 statistics must be index-wide, not per scoring arm (#188)

### LINKED ISSUE #188: BM25 length normalisation is per-arm: overwriting a document moves it from last to first
Overwriting a document changes its BM25 score enough to move it from near-last to first. Reproducible on a 2-core box, invisible on a 32-core one, which is why CI has been red while local runs are green.

## Repro

```sh
taskset -c 0,1 cargo test --release -j 2 -p xerj-engine \
  --test search_bounded_under_ghosts size5_returns_global_top5_under_ghosts_on_large_matchset
```

```
sanity: the 5 short docs must be the global top-5, got
[("weak001", 0.28768212), ("strong0", 0.17714293), ("strong1", 0.17271437),
 ("strong2", 0.16645661), ("strong3", 0.16120009)]
```

## Why that ordering is wrong

The corpus is built so the ranking is unambiguous. `weak_body(i)` is a single buried occurrence of `quicklist` in a long padded field, so weak docs are the *lowest* scoring by length normalisation. `strong{i}` repeats `quicklist` 10-i times in a short field, so they are the true top 5.

`weak001` should be near the bottom. It comes back first, at 1.6x the top strong doc.

The only thing that happened to it is an overwrite:

```rust
idx.index_document(Some("weak001".into()), json!({"body": weak_body(1)})).await
```

Same id, same body. Its live copy moves to the memtable and its segment copy becomes a ghost. Content unchanged, score jumps from lowest band to highest.

## Mechanism

The memtable arm scores with its own local statistics rather than the index's. A document sitting alone in the memtable is length-normalised against itself, so `dl/avgdl` is ~1 and the long-field penalty that made it weak disappears. Segment-resident docs are still normalised against the segment's average, where the short strong docs earn their boost.

So BM25 length normalisation is per-arm, not global.

## Isolation

Of the four tests in that file, only this one fails on 2 cores, and it is the only one that performs an overwrite. Delete-only ghosts are fine.

```
page_reaching_into_truncated_ghost_segment_is_exact ... ok
size5_returns_global_top5_under_ghosts_on_large_matchset ... FAILED
top_hit_stable_across_page_sizes_under_ghosts ... ok
large_single_segment_with_top_ghost_returns_exact_order ... ok
```

Core count matters because it changes shard fan-out and therefore whether the overwritten doc is still memtable-resident at query time. On 32 cores the run happens to avoid it. That makes this look flaky when it is deterministic given the shape.

## Why it matters beyond CI

Any client that re-indexes a document, a sync job, an upsert, a retry, gets that document ranked against different statistics than everything else until the next flush. It will drift up the results for no reason the user can see, then drop back after a merge. Two identical queries against identical data return different orders depending on flush timing.

Related: #173 notes BM25 statistics are per-index, so term rarity is estimated from too small a sample. This is the same family, one level down, between the memtable and the segments inside a single index.

## Status

Blocking every open PR, since main cannot go green. Not caused by #179, that test just exposed it. I would rather not paper over it with a tolerance in the assertion, because the assertion is right.

xerj-org:
Closes #188.

Overwriting a document with identical content moved it from last place to first. The arithmetic decomposes with no residue:

```
memtable N=1, df=1  ->  idf = ln(1 + 0.5/1.5) = 0.2876821
weak_body(1) = 87 tokens, memtable avgdl = 87/1 = 87  ->  dl/avgdl = 1.0
tf_norm = 2.2 / (1 + 1.2*(1-0.75+0.75*1.0)) = 1.0
score = 0.28768212        <- the literal failure value
```

It is not only the memtable. Every segment also scored against its own `FieldStats` (`xerj-fts/src/search.rs:1214` -> `reader.field_stats`), so a document's rank depended on which arm held it. The doc comment at `memtable.rs:996-1002` claiming scores are "segment-invariant" was wrong; it was shard-invariant.

Fix: fold one `CollectionStats` per search, the union of per-field `FieldStats` and per-(field,term) `doc_freq` over every segment plus the memtable, and feed it to both the memtable scorer and every segment searcher. That is Lucene's shape (`IndexSearcher.fieldStats` sums docCount over `reader.leaves()`, `TermStates.build` sums docFreq, one SimScorer scores every leaf) and already this repo's convention on the columnar path. `FieldStats` was already the right additive carrier, so there is no new arithmetic. Adds `open_stats_only` so gathering stats skips the postings decompress and the O(docs) norms load.

2-core reproduction, before and after:

```
taskset -c 0,1 cargo test --release -j 2 -p xerj-engine --test search_bounded_under_ghosts
before: size5_returns_global_top5_under_ghosts_on_large_matchset ... FAILED
after:  4 passed
```

## What this changes for users

**Absolute `_score` values change** on any index with more than one live scoring arm. That is the fix: scores become comparable across arms instead of depending on flush timing. Relative ordering within a single-arm index is unchanged, and there is a unit test pinning that identity.

## Known limits, not fixed here

- Multi-term expansions (prefix, wildcard, fuzzy, the trailing term of `match_phrase_prefix`) still take their local `doc_freq`, because their term sets resolve per arm.
- The union declines and falls back to per-arm statistics above `MAX_STATS_PROBES` ((fields + terms) x segments > 4096).
- The cost measurement is inconclusive rather than clean: the harness floor (~3.3 ms) and its run-to-run variance (0.46 ms) swamp the signal.
- Exact-score ties expose a separate pre-existing defect, filed as #191. `highlight_does_not_change_ranking.rs` has a one-line fixture change (varying title tf) so it keeps the tie-free ordering it needs to test its own subject; no assertion was touched.

The `search_bounded_under_ghosts.rs` reflow is `cargo fmt`, identical to #186.

Merge after #186 so the formatting hunks collapse to no-ops.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -13470,6 +13470,73 @@ impl Index {
         // paths, FTS, DocsForScan, and match_all.
         let mut mem_matches_known: Option<u64> = None;
         let mem_doc_count = self.memtable.doc_count();
+
+        // ── #188: INDEX-WIDE BM25 collection statistics ───────────────────
+        //
+        // BM25 only compares two documents when both were scored against the
+        // same `N`, `avgdl` and `doc_freq`.  Historically every arm used its
+        // own: each segment scored against its own `FieldStats`, and the
+        // memtable against the union over its shards only.  Two user-visible
+        // consequences:
+        //
+        //   * overwriting a document PROMOTED it — its live copy moves to the
+        //     memtable, and alone there it scores idf=ln(4/3), dl/avgdl=1,
+        //     i.e. a fixed 0.28768 that beats almost any correctly-normalised
+        //     segment hit, regardless of the document's real length;
+        //   * scores tracked segment TOPOLOGY — the same 365-doc corpus
+        //     flushed into 1 segment vs 16 scored `strong0` at 0.177 vs 0.570.
+        //
+        // Fix: fold one `CollectionStats` over every live arm (all segments
+        // plus the memtable) and hand it to BOTH the memtable scorer and every
+        // segment `FtsSearcher`.  This is Lucene's shape —
+        // `IndexSearcher.fieldStats` sums `docCount`/`sumTotalTermFreq` over
+        // `reader.leaves()` and `TermStates.build` sums `docFreq` over them,
+        // then ONE `SimScorer` scores every leaf — and tantivy's
+        // (`Bm25StatisticsProvider for Searcher` sums over `segment_readers()`,
+        // `Bm25Weight::for_terms` builds one weight from it).  It is also
+        // already this repo's convention on the columnar path:
+        // `leaf_constant_score` sums df + total_docs across every segment.
+        //
+        // SINGLE-ARM GATE.  With at most one live scoring arm the union is the
+        // identity, so we skip it and take the historical path verbatim.  That
+        // is not just an optimisation: for a memtable-only index the union
+        // would also redefine `N` from `doc_count()` (all memtable docs) to
+        // per-field docs-with-field, so the gate is what keeps every
+        // index → refresh → search shape — the ES-YAML suite's dominant one,
+        // and one segment per refresh on a 1-ingest-shard runner — byte-exact.
+        //
+        // `count_only` skips it too: a size:0 count never builds a scorer, and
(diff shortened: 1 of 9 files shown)


xerj-org:
Ran the conformance suite against this branch myself, since the implementer explicitly did not and this is the change that moves scores.

```
1365 passed · 0 failed · 3 skipped · 1368 total
```

Dedicated instance on a fresh data dir, binary built from 798baf2a. The gate holds.

Issue #204, opened 2026-08-07
Tracking: accepted-and-ignored is the dominant bug class, six instances found in one review


xerj-org:
Tonight's source review turned up the same bug six times in six unrelated subsystems. Filing it as a class because fixing them one at a time will miss the next one.

The shape: XERJ accepts an instruction, returns success, and does nothing. Or it hits a failure, degrades to a weaker behaviour, and reports normal.

Known instances:

| where | what happens |
|---|---|
| `doc_values: false` on a mapping | accepted, echoed back by GET _mapping, ignored (fixed in rc.12) |
| `index: false` on a mapping | same, still open |
| `xerj autoindex` under a write block | reports success, writes zero documents (#195) |
| ILM policies | stored and returned on GET, never executed (#199) |
| scroll | returns 10,000 documents and stops, no error (#198) |
| TLS cert failure | logs, then serves plain HTTP (#200) |
| unparseable `schema.json` | opens the index with an empty dynamic mapping (#202) |
| restart | templates, pipelines, data streams silently gone (#203) |

Each one in isolation looks like an oversight. Together they are a missing convention.

What they have in common is that the failure is invisible at the point where a user could act on it. The signal is either in a log on another machine, or nowhere. Every one of these leaves the system running and answering, which is why they survive to production.

Two things worth adopting before GA:

**Accepted means honoured.** If an API takes an option it does not implement, it should reject it, not store it. An endpoint that 404s tells the truth. An endpoint that accepts and echoes a retention policy it never runs produces an unbounded disk bill and a compliance problem. Applies to ILM, `index: false`, and anything else in the mapping and settings surface that is currently parsed and dropped.

**Degrade loudly or not at all.** Falling back is fine when the fallback is equivalent. It is not fine when the fallback is weaker: cleartext instead of TLS, dynamic instead of an explicit mapping, a partial export instead of a full one. Those should fail closed.

A useful sweep would be to grep the codebase for `unwrap_or_else`, `unwrap_or_default` and `let _ =` on paths that handle user-supplied configuration, and ask of each one whether the fallback is equivalent or weaker. The two in `xerj-engine/src/index.rs:5568` and `:5578` are the pattern in its purest form.

The storage layer already gets this right and is worth copying: segment headers carry a format version and readers fail loudly with `UnsupportedFormat` rather than guessing (`xerj-storage/src/segment.rs:243-246`).

Filed as a tracking issue. Individual fixes are #195, #198, #199, #200, #202, #203.

PR #231, opened 2026-08-08
fix(console-api): require If-Match on dashboard writes — 428 when missing

### LINKED ISSUE #210: Dashboard writes lose updates: If-Match is optional so two writers clobber silently
Two people editing the same dashboard can silently overwrite each other.

`enforce_etag` in the console API (`dashboards.rs:574-588`) treats `If-Match` as optional. Its own comment says "If-Match optional but recommended", and it returns `Ok` when the header is absent. So the 409 conflict protection only applies to clients that opt in by sending the header. Two writers that do not send it both succeed, and the second one wins.

Lost update, no error, no warning. The person whose edit vanished finds out later, if ever.

Optional concurrency control is not concurrency control. The two workable shapes are:

- require `If-Match` on write and reject with 428 Precondition Required when it is missing, or
- keep it optional for reads but make writes without it fail

Either is fine. What does not work is documenting it as protected while the default path is unprotected.

This is the same family as #204 (accepted-and-ignored): a safety mechanism that exists, is documented, and does nothing unless the caller happens to opt in.

Found while reviewing XERJ's source against the user-feedback corpus.

xerj-org:
Closes #210

## Problem

Two people editing the same dashboard could silently overwrite each other. `enforce_etag` in `xerj-console-api` treated `If-Match` as optional — its comment said "optional but recommended" and it returned `Ok(())` when the header was absent — so the 409 lost-update protection only applied to clients that opted in. Two writers that never sent the header both got 200 and the second one won. Same family as #204: a documented safety mechanism that does nothing unless the caller opts in.

## Fix

Issue #210 offered two workable shapes; this takes the first: **require `If-Match` on write, reject a missing header with 428 Precondition Required** (RFC 6585 §3).

- `enforce_etag` now returns a new `ConsoleApiError::PreconditionRequired` → **428** with a message carrying the current version to retry with. A present-but-stale header still **409**s exactly as before; weak (`W/"7"`) and strong (`"7"`) forms both still accepted.
- Applies to both write call sites (PUT replace, PATCH) — private dashboards included: one owner with two tabs open is still two writers. Module doc updated (it previously claimed private dashboards skip the check).
- Console client already sends `If-Match` built from `doc.version` on every PUT/PATCH; its one 409-recovery branch (`pushPatch`) now also retries on 428 via the same re-GET + reapply path.
- Unchanged, deliberately: POST create (no prior version to guard), the admin/owner-only `_bulk` whole-set overwrite (used by seeding), and DELETE (out of this issue's scope; the client already sends the header on `remove()` if the server later enforces it).

## Evidence

- New test `phase3_user_state::dashboard_write_without_if_match_is_428`: headerless PATCH and PUT must 428 with code `precondition_required`, the rejected write must not land, and the same PATCH with `If-Match: W/"1"` still succeeds. **Verified failing before the fix** — headerless PATCH returned 200 and `"Writer B clobber"` landed — and passing after.
- `cargo test --release -p xerj-console-api`: all suites green (108 tests, 0 failed).
- ES-YAML conformance gate against this build (fresh data dir, port 9710): **1365 passed / 0 failed / 3 skipped**.

## Retrieval

`xc.py xerj-search` queried twice (If-Match / precondition / version-conflict phrasings); top hits were S3 blob-store ETag fixtures and dense-vector mapper code — irrelevant to document-write concurrency, so no peer implementation was drawn on. The fix follows RFC 6585 §3 / RFC 9110 If-Match semantics and this crate's existing 409 path.

### DIFF
--- a/engine/crates/xerj-console-api/src/dashboards.rs
+++ b/engine/crates/xerj-console-api/src/dashboards.rs
@@ -2,9 +2,11 @@
 //!
 //! Replaces `localStorage["xerj.dashboards"]` and
 //! `localStorage["xerj.layout.<id>"]` with engine-backed durable state.
-//! Every shared dashboard write expects an `If-Match: W/"<version>"`
-//! etag for optimistic concurrency; private dashboards skip that check
-//! since only the owner ever writes them.
+//! Every dashboard write (PUT / PATCH) **requires** an
+//! `If-Match: W/"<version>"` etag for optimistic concurrency: a missing
+//! header is rejected with 428 Precondition Required, a stale one with
+//! 409 Conflict. This applies to private dashboards too — one owner
+//! with two tabs open is still two writers.
 //!
 //! Real-time push (SSE `/_stream`) is NOT implemented in this release:
 //! there is no `/_stream` endpoint on dashboards or views, and none is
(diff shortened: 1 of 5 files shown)


xerj-org:
Author identity fixed: 15082748 was authored as `ivan.w.a@icloud.com` (a concurrent session's git identity leaked through the shared repo config). Amended to `xerj-org <git@xerj.org>` and force-pushed as 1b4ee45b — tree, parent, message, and dates are byte-identical (verified: `git diff 15082748 1b4ee45b` is empty; raw commit objects differ only in the two identity lines).

@cla-bot check

PR #234, opened 2026-08-08
fix(api): support the legacy path-parameter form of scroll continuation

Vinz2168:
## Summary

OpenSearch Dashboards 3.7.0 crashes fatally against xerj — not just a broken page, the whole process exits — whenever it needs to migrate its saved-objects index (`.kibana_1` → `.kibana_2`, triggered on every startup once OSD detects a mapping change, which is normal/expected on a version bump). This blocks the entire Dashboards UI, discovered incidentally while verifying an unrelated ISM/ILM feature against a real OSD instance.

## Root cause, isolated precisely

Enabled `logging.access_log = true` (an existing config option — request-level tracing via tower-http) and reproduced with a real OSD 3.7.0 container against an xerj instance carrying a pre-existing `.kibana_1`. Confirmed a **fresh** instance with no prior `.kibana_1` boots 3.7.0 fine (`Creating index .kibana_1` → `Finished in 102ms`) — the crash is specifically in the *migration* path, not first-time setup.

The request trace showed the exact sequence up to the crash: `PUT /.kibana_2` (200), `POST /.kibana_1/_search?scroll=15m` (200 — scroll opened), `POST /_bulk` (200 — first page migrated into `.kibana_2`), then **nothing** — the next request OSD sent never reached any handler at all.

Confirmed directly with curl: continuing a real scroll via the documented body/query-param form (`POST /_search/scroll {"scroll_id":...}`) returns 200; continuing the *same* scroll via the documented path-parameter form (`GET /_search/scroll/{scroll_id}`) returns a bare 404 with an empty body — axum's own "no route matched" fallback, not an application error. OSD's opensearch-js client uses this path form for the migration's scroll continuation (and, per the verified fix below, `DELETE /_search/scroll/{scroll_id}` to close it), and OSD's migration code treats a raw 404 there as fatal instead of retrying.

Both forms are real, standing Elasticsearch/OpenSearch REST API variants — this was simply never implemented, not a client quirk to route around.

## Fix

Extracted the response logic already in `next_scroll` into a shared `scroll_page_response`, and added:
- `GET|POST /_search/scroll/{scroll_id}` → `next_scroll_path`
- `DELETE /_search/scroll/{scroll_id}` → `clear_scroll_path`

Both delegate to the exact same code path the existing body-form endpoints use (same expiry/keep-alive handling, same "unknown UUID → 400 Cannot parse scroll id" behavior matching ES) — this is the same feature reachable through a URL shape ES itself documents, not new behavior.

## What was verified

- New test `scroll_continuation_via_legacy_path_param_form_works`: builds a real scroll context, hits `GET /_search/scroll/{id}` through the actual router, asserts the same shape as the existing body-form test, and asserts an unknown id still gets ES's 400.
- `cargo test -p xerj-api --lib` — **162/162 pass**.
- `cargo clippy -p xerj-api --lib -- -D warnings` / `cargo fmt --check` — clean.
- `cargo build --release` — clean.

**Live end-to-end re-reproduction of the original crash, confirmed fixed**: started a fresh xerj instance, let a real OSD 3.6.0 create `.kibana_1` against it (recreating the exact pre-migration state that crashed 3.7.0), then swapped in OSD 3.7.0 pointed at the same instance to force the real migration.

- **Before this fix**: `FATAL` crash, OSD process exits, container status `exited`.
- **After this fix**: OSD's own log shows the complete successful sequence — `Migrating .kibana_1 saved objects to .kibana_2` → `Pointing alias .kibana to .kibana_2.` → `Finished in 98ms.` — and `GET /api/status` returns 200. xerj's access log confirms the previously-missing call now succeeds: `DELETE /_search/scroll/{id}` → 200 (OSD used the DELETE path-form to close the scroll once `.kibana_1`'s small dataset fit in a single page).

## Test plan

- [x] `cargo test -p xerj-api --lib` (162/162)
- [x] `cargo clippy -p xerj-api --lib -- -D warnings`
- [x] `cargo fmt -p xerj-api -- --check`
- [x] `cargo build --release`
- [x] Full live re-reproduction against real OSD 3.6.0 (create `.kibana_1`) → OSD 3.7.0 (force migration) — crash confirmed fixed end-to-end

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -17168,7 +17168,50 @@ pub async fn next_scroll(
             return ApiError::new(e).into_response();
         }
     };
+    let scroll_keep_alive = body.scroll.as_deref().or(params.scroll.as_deref());
+    scroll_page_response(
+        &state,
+        scroll_id,
+        scroll_keep_alive,
+        params.rest_total_hits_as_int.as_deref() == Some("true"),
+    )
+    .await
+}
 
+/// `GET|POST /_search/scroll/{scroll_id}` — the legacy path-parameter form
+/// of scroll continuation. A real, documented ES/OpenSearch REST API
+/// variant alongside the body/query-param form above (`next_scroll`), not
+/// a client mistake: found missing while investigating why OpenSearch
+/// Dashboards 3.7.0 crashes fatally partway through its saved-objects
+/// migration (`.kibana_1` → `.kibana_2`) against xerj. Traced via request
+/// tracing (`logging.access_log = true`) to the exact call: the migration
+/// opens a scroll (`POST /.kibana_1/_search?scroll=15m`, 200), bulk-indexes
+/// the first page into the new index (200), then continues the scroll via
+/// `GET /_search/scroll/{scroll_id}` — which had no matching route at all,
+/// so axum's default fallback returned a bare 404 with an empty body. OSD's
+/// migration code treats that as fatal and aborts the whole process rather
+/// than retrying, so ANY client using this form — not just OSD 3.7.0's
+/// migrator — would have hit the same wall.
+pub async fn next_scroll_path(
+    State(state): State<AppState>,
+    Path(scroll_id): Path<String>,
+    Query(params): Query<ScrollQueryParamsFull>,
+) -> impl IntoResponse {
+    scroll_page_response(
+        &state,
+        scroll_id,
+        params.scroll.as_deref(),
+        params.rest_total_hits_as_int.as_deref() == Some("true"),
+    )
+    .await
+}
+
+async fn scroll_page_response(
+    state: &AppState,
+    scroll_id: String,
+    scroll_keep_alive: Option<&str>,
+    rest_total_hits_as_int: bool,
+) -> Response {
     // An id xerj never issued (its scroll ids are v4 UUIDs) cannot name a
     // context: ES answers 400 `Cannot parse scroll id`, not a 404 (item 4e).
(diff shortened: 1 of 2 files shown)


xerj-org:
Thanks for this — a real fix for the legacy path-parameter scroll form, and it lines up with the loud-truncation work that just landed in #225.

One thing before we can merge: the `verification/cla-signed` check is asking for a CLA signature from your account. That gate is deliberate (it keeps provenance clean for an Apache-2.0 project), and it's the only thing blocking this PR — every other check is green. Once you've signed, comment `@cla-bot check` on the PR and it'll re-run.

We'll review properly on this side in the meantime.

xerj-org:
@cla-bot check

xerj-org:
Good news: your CLA is signed — #246 is merged and `Vinz2168` is in `.contributors`. The remaining red check is not a CLA problem, it is an identity-resolution one.

The commit here is authored `Vincenzo Lombardo <vincenzo.lombardo@seacom.it>`, and GitHub cannot map that address to your account, so cla-bot has no GitHub username to look up and reports "could not parse the GitHub identity" rather than checking the list.

One minute to fix, either way:

- **Easiest:** add `vincenzo.lombardo@seacom.it` to your account at https://github.com/settings/emails (it can stay private). Then comment `@cla-bot check` here and the check resolves.
- **Or** amend the commit to use an email already on your account and force-push:
  `git commit --amend --author="Vinz2168 <your-github-email>" && git push --force-with-lease`

The code itself is done and needs no changes — this is the only thing standing between it and merge.

xerj-org:
@cla-bot check

PR #246, opened 2026-08-09
Sign CLA: add Vinz2168

Vinz2168:
Signing the CLA per CLA.md — this PR adds my username to .contributors.

Issue #199, opened 2026-08-07
ILM policies are accepted and stored but never executed, so retention silently does nothing


xerj-org:
The ES-compat ILM API accepts policies, stores them, and returns them on GET. Nothing ever runs them.

`xerj-api/src/es_compat.rs:22720-22764` handles the ILM endpoints as storage only. The engine's background sweepers (`xerj-engine/src/engine.rs:610-708`) cover PIT, scroll and async-search cleanup. There is no delete-phase executor, no rollover executor, no retention timer of any kind.

So a user migrating from ES does this:

```
PUT _ilm/policy/logs-30d   -> 200 OK
GET _ilm/policy/logs-30d   -> their policy, exactly as written
```

and concludes retention is configured. It is not. The index grows forever.

This is worse than not implementing ILM at all. An endpoint that 404s tells the truth. An endpoint that accepts and echoes a retention policy while ignoring it produces an unbounded disk bill and a compliance problem, and the user has no signal until the disk fills.

There is also a GDPR angle. Right-to-be-forgotten works mechanically (merges purge tombstoned docs, `xerj-storage/src/merge.rs:25-29`) but there is nothing that expires data on a schedule, so "we delete logs after 30 days" cannot be honoured by configuration.

Options, in order of honesty:

1. Implement a retention executor. Honour the ILM delete phase, and add a simpler native `retention_days` on the index for people who do not want ES policy syntax.
2. Until then, reject ILM policy writes with a clear error saying retention is not automated yet.

Option 2 is a few lines and can ship immediately. Silently accepting is the one option that should not survive to GA.

This is the same failure class as #195 (autoindex reports success and writes nothing) and the `doc_values` bug fixed in rc.12: an option is accepted, echoed back, and ignored. Worth looking for others.

Found during a source review against the user-feedback corpus, where log retention cost is one of the loudest complaints ($60K/year for 30 days of logs).

xerj-org:
Another instance of the same thing, and this one is in the config file rather than an API.

`xerj-common/src/config.rs:806-822` defines a native `[logs]` section with `retention_days`, documented as "How long to retain log data before automatic deletion (default: 90 days)", plus `time_partition`. Nothing reads either value.

So there are now two independent surfaces that promise retention and deliver none: the ES-compat ILM API and XERJ's own config. An operator who avoids ES policy syntax and uses the native knob instead gets the same silent nothing.

Same fix applies. Either wire an executor, or remove the knob. Tracked in the class issue #204.

[...]

PR #247, opened 2026-08-09
feat(engine): real ISM/ILM index lifecycle management — one engine, two REST surfaces

Vinz2168:
## Summary

Real index lifecycle management, execution not just CRUD: a single internal
engine modeled on OpenSearch's ISM state machine (named states, ordered
actions, ordered transitions), exposed through two REST surfaces:

- **Native ISM** (`_plugins/_ism/*`) — matches OpenSearch's own API shape.
- **ES-shape ILM** (`_ilm/*`) — translated into the same internal model, not
  a second engine. `_ilm/policy` PUT parses ES's fixed hot/warm/cold/delete
  phases into ISM states (one state per phase, in that order); `_ilm/explain`
  and index.lifecycle.name auto-attach are ES-native entry points into the
  identical execution path ISM policies use.

A background job (default every 5 minutes, `[lifecycle] tick_interval_secs`)
walks every policy-attached index, executes the current state's pending
actions, then evaluates transitions in order and moves to the first one
whose condition is satisfied.

## What's implemented

**Actions**: `rollover` (reuses the existing, already-real
`rollover_data_stream()` — not reimplemented), `delete`, `read_only` (a
write-block flag). `replica_count` is accepted and stored (real ILM/ISM
policies commonly include it) but **errors explicitly at execution time**
rather than faking success — xerj has no per-index replica concept to honor
it against. Same principle as an earlier CCR fix in this codebase: an honest
error beats a silent no-op that looks like it worked.

**Conditions**: `min_index_age`, `min_size`, `min_doc_count` — evaluated
against real index stats (segment byte sum + memtable size for size,
`state_entered_at_ms` for age).

**REST surface** (`_plugins/_ism/*`): policy CRUD, `add`/`remove`/
`change_policy`/`retry` on managed indices, `explain` (single + list-all).
The list-all forms (`GET .../policies`, `GET .../explain`) and the
`remove`/`change_policy`/`retry` actions were not in the original minimum
spec — added after finding, via the real OSD UI, that OSD's own backend
proxy calls them and that its Managed Indexes table renders buttons for all
three with no backend otherwise. `change_policy` honors the `include` safety
condition (refuses the change if the index's current state doesn't match
the caller's allow-list) the same way real ISM does.

**ES surface** (`_ilm/*`): policy CRUD (including the bare `GET _ilm/policy`
list-all form Kibana's own UI calls), `{index}/_ilm/explain`, and
`index.lifecycle.name` auto-attach on both index creation and
`PUT {index}/_settings`.

## Verified

- **10 unit tests** in `xerj_engine::lifecycle` (duration/size parsing,
  condition evaluation, policy validation, action JSON round-trip, ILM
  phase-to-state translation including the unsupported-action error path).
- **Real OSD 3.6.0 UI**, not just curl: created a policy from the visual
  Policies UI, attached it to a real index, confirmed the Managed Indexes
  table shows live `policy_id`/`state`/`step` — then clicked "Remove
  policy", "Change policy", and "Retry policy" from the actual table and
  confirmed each against the real REST calls those buttons make
  (`POST /api/ism/removePolicy` → 200, etc.), not just the underlying
  endpoints in isolation.
- **Real Kibana 8.13.0** for the ILM side: created an ILM policy, attached
  it via `index.lifecycle.name`, confirmed `_ilm/explain` reports the
  correct phase — and found + fixed the bare `GET _ilm/policy` gap that was
  making Kibana's own Policies list page render empty.
- **`load_persisted_ism_policies` validates on read**: the two REST
  insertion paths (native PUT, and ILM translation) already validate before
  storing a policy; the disk-reload path didn't, so `change_policy` could
  in principle end up referencing a policy that was never checked if the
  persistence file were hand-edited or corrupted. Not REST-reachable, but
  closed for defense in depth — a bad single entry is now skipped and
  logged rather than silently trusted.
- **Rebased clean against upstream/main** immediately before opening this
  PR (the repo moves fast) — one real merge conflict in `es_compat.rs`
  against `PUT {index}/_settings` (an unrelated `index.blocks.*` forwarding
  change landed in the same function), resolved by keeping both blocks;
  confirmed with the `index.blocks.*` integration test suite (6/6) plus the
  full xerj-api/xerj-engine suites after resolving.
- `cargo test -p xerj-api --lib`: 169/169. `cargo test -p xerj-engine
  --lib`: 434/436 (the 2 failures are pre-existing, macOS-only,
  `snapshot_path_security_tests` — confirmed via `git stash` to fail
  identically on clean `upstream/main`, unrelated to this change).
  `clippy -D warnings` / `cargo fmt --check`: clean on both crates.

## Known limitations (flagging for review, not hidden)

- **ILM's `min_age` is measured from state-entry time, not from rollover
  time.** Real Elasticsearch measures a phase's `min_age` from the index's
  rollover time once the hot phase has a rollover action; this
  implementation measures every state's age from when the index entered
  that state. Documented in `xerj_engine::lifecycle`'s module doc comment.
  Simple ILM policies (age measured from index creation, no rollover) are
  unaffected.
- **No UI-facing "retry failed step" state exists to retry, by design**:
  xerj's engine re-attempts a failing action on every tick automatically
  (see `ManagedIndexState::failed`'s doc comment) rather than halting like
  real ISM does. The `retry` endpoint still exists because OSD's UI button
  needs somewhere to post to, but it's mostly a formality here — the engine
  was already self-healing.
- Single-node: `replica_count` can only ever error, never succeed, until/
  unless xerj gains a real per-index replica concept.

## Test plan

- [x] `cargo test -p xerj-engine --lib` (434/436, 2 pre-existing unrelated
      macOS failures)
- [x] `cargo test -p xerj-api --lib` (169/169)
- [x] `cargo clippy -p xerj-api -p xerj-engine --lib -- -D warnings`
- [x] `cargo fmt -p xerj-api -p xerj-engine -- --check`
- [x] Live verification against real OSD 3.6.0 UI (policy create, attach,
      explain, remove, change_policy, retry — all through the actual
      buttons, not just curl)
- [x] Live verification against real Kibana 8.13.0 UI (ILM policy create,
      attach, explain, Policies list page)
- [x] Rebased clean against `upstream/main` immediately before opening,
      one conflict resolved and re-verified (build + full test suites +
      the `index.blocks.*` integration tests specifically)


xerj-org:
Thanks for this — real execution rather than CRUD-only is exactly what #199 asks for, and modelling it on OpenSearch's ISM state machine is a sensible shape.

Same note as on #234: your CLA is fine (#246 merged, `Vinz2168` is in `.contributors`). The red check here is identity resolution — these commits are authored `Vincenzo Lombardo <vincenzo.lombardo@seacom.it>`, and GitHub cannot map that address to your account, so cla-bot has no username to look up. Add that email at https://github.com/settings/emails (it can stay private) and comment `@cla-bot check`, and both PRs go green together.

Heads-up on overlap: we had an agent starting on #199 in parallel, not knowing this existed. Yours came first and is further along, so we will review this one rather than land a competing implementation — if ours produces anything useful it will come to you as review comments, not a rival PR.

Issue #173, opened 2026-08-04
autoindex: 2-repo tree fragments into 407 indices with 1.2% semantic_text coverage, breaking vector search and BM25 IDF


xerj-org:
## `autoindex`: a 2-repo source tree fragments into 407 indices, and only 1.2% of them get `semantic_text`

Indexing two ordinary C repositories produces a corpus that is badly shaped for retrieval: hundreds of tiny indices, and vector search effectively disabled across almost all of it.

### Reproduce

```sh
git clone --depth 1 https://github.com/valkey-io/valkey
git clone --depth 1 https://github.com/memcached/memcached
xerj autoindex <dir-containing-both> --url http://localhost:9200 --prefix kv --no-graph
```

Measured (2,268 files in, 7,923 records out):

```
indices                    407
records                  7,923
mean records per index    19.5
singleton indices            0
indices with body: semantic_text   5 of 407  (1.2%)
```

### Why this hurts

**Vector search is dead on 98.8% of the corpus.** A `semantic` query against an index whose `body` is plain `text` does not return fewer results — it fails the whole request with a 400 and takes every other index in the wildcard down with it. A client must therefore read the mapping first and aim the vector arm only at capable indices, which here is 5 of 407. Hybrid retrieval degenerates to lexical-only for almost every query.

**Scoring is fragmented.** BM25 statistics are per-index. Splitting one logical corpus across 407 indices means IDF is computed over ~19 documents at a time, so term rarity — the thing BM25 ranks on — is estimated from a sample far too small to be meaningful.

The fragmentation is not caused by singletons (there are none); it is many small groups, which suggests schema inference is splitting on incidental field differences between source files rather than on anything a user would call a dataset.

### What I could not determine

Whether the `semantic_text` decision is made per inferred group by sampling content (a group of numeric-heavy or short files not "looking like" prose), or whether it follows from group composition some other way. Both fit the data I have. A related observation from #170: 28 records that got no `defs` had also landed in the one group whose `body` was mapped plain `text`, and giving them `defs` dissolved that group — so grouping and mapping appear coupled, but I did not isolate the mechanism.

### Suggested direction

Worth deciding explicitly what a "dataset" should mean for a source tree. Options, in rough order of effort: group by repository or top-level directory rather than inferred schema; or keep inference but merge groups whose mappings are compatible; or make `semantic_text` the default for anything detected as code/prose regardless of group size, so the vector arm is not silently unavailable.

[...]

PR #248, opened 2026-08-09
perf(autoindex): parse each PDF once and replay it in phase B (supersedes #166)

xerj-org:
> [!IMPORTANT]
> **Do not merge until the CLA question is answered — owner decision, see [the CLA status comment](https://github.com/xerj-org/xerj/pull/248#issuecomment-5234659799).**
> The green `cla-signed` check here means every *commit author* is `xerj-org`;
> it does **not** mean @buger signed. `.contributors` does not contain `buger`,
> and the `Co-authored-by:` trailer that carries the real attribution is
> invisible to cla-bot. Settle it either by buger's own one-line signature PR
> (asked on #166) or by an explicit written waiver here. Held in draft so it is
> not merged by accident; `gh pr ready 248` once decided. Class-level gap
> tracked in #269.

Supersedes #166 by @buger.

#166's head is on `probelabs/xerj` with `maintainerCanModify: false`, so nobody
in xerj-org could push the rebase to it — every attempt 403s. The seven commits
`e73eda74..237d6df9` are re-landed here on an xerj-org branch, rebased onto
current `main` (about 100 commits of drift), with the outstanding review
blockers fixed. Two commits: the contributor's design, then our rebase and
review work, so the two stay legible. Both carry
`Co-authored-by: Leonid Bugaev <leonsbox@gmail.com>`.

**Do not close #166 from here** — it closes when this merges.

## What it does

Every accepted PDF was parsed twice per fresh run. Phase A spawns the isolated
worker to sniff and sample, and for PDFs that is a *complete* parse:
`extract::extract` routes `Family::Pdf` to `pdf::extract(path, sink)` and drops
the sampling limit entirely (`extract/mod.rs:162`) — only delivery stops early.
Phase B then spawned the worker again for the same bytes.

This retains each validated phase-A worker response in an anonymous temporary
file under `--state-dir` and lets phase B replay it. Reuse is an optional
accelerator, never correctness-critical state:

- **Bounded twice**: 384 MiB of retained-plus-in-flight response bytes, and an
  artifact-handle cap derived from live `RLIMIT_NOFILE` and `/proc` open-fd
  counts. Both snapshots are re-taken at every admission.
- **Correctness-critical resources come first**: a 4 GiB-or-half-free
  filesystem floor (plus `workers × bulk_mb`) is reserved for the journal and
  phase-B staging before any artifact is admitted.
- **Linux only, by design**: other platforms have no live descriptor evidence,
  so they decline the optimization and keep the ordinary parse path. This is a
  precondition of the feature, not a bug — it is in the CLI help and stated
  here rather than only in the help text.
- **Verified before any record is published**: physical length, content digest,
  JSON decode, and worker-protocol identity. A failure publishes no record and
  no `file_done`; phase B reparses.
- **Consumed exactly once**, with descriptor and bytes refunded before
  unbounded phase-B staging begins. A file phase A junks, or that the final
  todo set excludes, refunds immediately instead of holding a descriptor across
  the phase boundary.
- **Restart-safe by not persisting**: a resumed run has a frozen plan but no
  trusted handle, so it parses again. Extraction reuse is deliberately not
  coupled to the journal.
- **Observable**: `--json` gains a `pdf_extraction_reuse` block (parser
  responses, verified replays, fallback categories, exact retained bytes,
  current/peak ownership, at most three path examples).

## Measured here, on this branch

The 17–21 % in #166 describes a base ~100 commits behind current `main` and is
**not** carried over. Everything below is a run I did on this branch's own
release binary (`sha256:c25c1af6e4fb36ad…`, `-C lto=fat`), x86_64 Linux.

Corpus: the 60 largest PDFs in this repository (471 MB; 49 yield text, 11 are
image-only and land as junk in every arm). Each run gets a fresh data dir,
fresh state dir and its own port; `--workers 1 --pdf-workers 4 --bulk-mb 1
--no-graph`, default (lexical) embedding. Counterbalanced order: on, off, off,
on.

Both arms are the **same binary**. The "off" arm runs under `ulimit -n 96`,
which drives the descriptor probe below the reserved headroom, so admission is
refused outright (`capacity_status: "disabled"`, `capacity_reason:
"descriptor_headroom"`) and every PDF takes the pre-change phase-A-parse +
phase-B-reparse path. Using one binary removes codegen differences from the
comparison; the cost is that the off arm carries an artificial descriptor
limit, which no code path other than spool admission scales with. A calibration
run at `-n 104` came back `"limited"` (partial reuse), which is why 96 was used.

| arm | external wall | product `wall_seconds` | phase-B PDF parses | verified replays |
|---|---:|---:|---:|---:|
| reuse **on** 1  | 121.94 s | 121.8 s | 0  | 49 |
| reuse **off** 1 | 138.54 s | 138.4 s | 49 | 0  |
| reuse **off** 2 | 136.15 s | 136.0 s | 49 | 0  |
| reuse **on** 2  | 127.02 s | 126.9 s | 0  | 49 |

Paired improvement: **11.99 %** and **6.71 %**; arm means 124.48 s vs 137.35 s,
a **9.37 %** reduction in end-to-end wall time (12.9 s). Product-reported wall
gives the same 9.37 %. n=2 pairs — a signal, not a distribution.

Every arm produced identical output: 60 files walked, 49 indexed, 11 junk,
1,813 records, exit code 3 in all four. The on arms retained 2,431,155 bytes of
artifacts, peaked at ~35.9 MB retained-plus-reserved, and finished with
`current_live_artifacts: 0` — every descriptor and byte refunded.

Two honest caveats. This corpus is smaller and more image-heavy than
FinanceBench-20, so the parse share of the run is lower and the win is smaller
than #166's. And the win is corpus-mix dependent by construction: it is bounded
by how much of the run is PDF parsing.

## Rebase — what had to be decided, not just merged

- **`build_phase_a`**. Main moved the whole phase-A body out of
  `run_index_report` into `build_phase_a` after the branch was written, so the
  inline spool logic moved inside it; it now returns a `PhaseA` struct instead
  of a tuple.
- **Argument growth** tripped `clippy::too_many_arguments` (9/7). The three
  spool parameters are grouped into
  `PdfSpoolContext { state_dir, budget, capacity_warning }`.
- **Phase B precedence — a real decision, not a textual merge.** Main added the
  `fa.as_document` branch (#173, demoted config files) exactly where this
  branch added `Family::Pdf -> spool.replay(...)`. `as_document` deliberately
  wins: it decides the record *shape*, and phase A built that dataset's mapping
  by re-sampling through `extract_as_document`, so replaying PDF page records
  there would publish records the frozen plan does not describe. No PDF can
  reach that branch today (`demotable_family` is JSON/YAML/XML only), so it is
  a guard for if that set widens; the artifact is refunded, not replayed.
- **A defect the rebase itself created and this PR fixes.** Main added the
  required `origin: FieldOrigin` field to `RawRecord`. The six `#!/bin/sh` PDF
  worker stubs in `failure_resume_http_tests.rs` still emitted pre-`origin`
  protocol JSON, so `validate_response_identity` rejected every response and
  each test PDF became junk (`files_junk: 1`, exit 3,
  `phase_a_pdf_parser_responses: 0`). The whole PDF suite was silently
  measuring nothing until the stubs were updated.

## The three review blockers from 2026-08-04 are closed

1. **Inert verification.** The `source_size`/`source_digest` `ensure!` was a
   self-comparison at the only production caller — phase B passes `f.size` and
   `digests[i]`, exactly what the spool captured in phase A. It is kept as a
   *binding* assertion (it still catches an artifact routed to the wrong file)
   and is now documented as one. The count is corrected everywhere: **four**
   artifact verifications can fail, plus one binding assertion. Source mutation
   between phases is caught earlier and at full strength by phase B's
   `content::verify` (`lib.rs:1781`).
2. **Test flake.**
   `failed_worker_protocol_call_is_not_reported_as_a_completed_phase_a_parse`
   raced on the process-global `XERJ_PDF_WORKER_BIN` that the `run_index` tests
   set for whole runs in the same binary; it passed in CI only because of
   `--test-threads=2` plus name ordering. A shared `WORKER_BIN_ENV_LOCK` now
   serializes every test that touches that variable, and the unit test points
   the variable at a worker that does not exist so its outcome no longer
   depends on who else is running.
3. **Coverage regression.** The phase-A source-generation-changed arm of
   `scan_file` had no test — its only candidate runs with
   `ExtractionSpoolBudget::new(0, 0)`, so no artifact is created and the
   `content::verify` comparison is never reached. Added
   `source_changed_during_phase_a_discards_the_artifact_and_names_the_reason`,
   which gives the run real capacity, lets the worker rewrite one byte of its
   own input, and asserts the artifact is created, discarded, refunded and
   reported as `source_generation_changed`.

## Counter semantics (the numbers quoted as evidence)

- `phase_a_pdf_parser_calls` → **`phase_a_pdf_parser_responses`**. It is
  incremented only after `spawn_worker` returns a response, so it never counted
  invocations that spawned and then failed or timed out. The name now says what
  the number is.
- `record_replay_fallback` no longer increments **`io_fallbacks`**. A digest,
  JSON or protocol-identity mismatch is an integrity outcome, already counted by
  `replay_integrity_failures`; counting it as I/O overstated disk trouble. The
  corrupted-replay test now asserts `io_fallbacks == 0`.
- Dropped the dead `_observe_global_replay` parameter.
- Reverted an `is_globally_disabled()` early return that suppressed every
  per-file fallback diagnostic when capacity was refused — those recorded
  examples are the only place `--json` explains *why* nothing was reused, and
  they are already capped at three.

## The second reviewer's design questions, answered

- **"Up to 512 spool fds held at the phase A→B boundary."** True, and that is
  the ceiling, not the expectation: the live cap is
  `RLIMIT_NOFILE − open − (64 + 4×workers + 4×pdf_workers)`, re-probed at every
  admission, and reaching 512 also requires the retained artifacts to be small
  enough to fit under the 384 MiB byte ceiling. CI's Autoindex FD smoke job runs
  on ubuntu, macos and windows.
- **"`fs2::available_space` statvfs per admission on a hot path."** One `statvfs`
  per PDF, on a path that has just spawned a process and parsed a whole
  document. It is not measurable next to that, and dropping it would mean
  admitting artifacts against a stale free-space snapshot.
- **"Phase A now pays a full parse for every spooled PDF."** It does not: phase
  A already paid a full parse for every PDF before this change
  (`extract/mod.rs:162` ignores the sampling limit for `Family::Pdf`). The
  change removes the second parse, it does not add a first one.
- **"Up to 384 MiB of transient artifacts land on the destination filesystem."**
  Bounded, and admitted only above the 4 GiB-or-half-free floor that protects
  journal and staging writes; every artifact is unlinked and refunded before
  phase-B staging begins.

## Not fixed, deliberately

- The nit to skip building per-file `SpoolFallback` messages when the budget is
  globally disabled is **not** taken. The suppression the previous attempt used
  also removed the only operator-visible explanation of why nothing was reused —
  which `refused_pdf_spool_reparses_and_reports_fallback_without_weakening_publication`
  asserts on. One small `format!` per PDF is not measurable next to a PDF parse,
  and the stored examples are already capped at three. The transient cost is one
  short `String` per PDF held for the duration of phase A.
- No `Fixes #N`: no open issue tracks this work.

## Gate (run locally on this branch, x86_64 Linux)

```
cargo fmt --all -- --check                                    pass
cargo clippy -p xerj-autoindex --all-targets -- -D warnings    pass
cargo test -p xerj-autoindex                                   283 passed, 0 failed
```

The suite was run five times — at `--test-threads=1`, `2` and the default — to
prove the env-var flake is actually gone rather than reordered.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,28 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **`autoindex` parses each PDF once per fresh run instead of twice.** Phase A
+  already paid a *complete* parse for every PDF — `extract::extract` routes
+  `Family::Pdf` straight to the isolated worker and drops the sampling limit,
+  so only delivery ever stopped early — and phase B then spawned the worker
+  again for the same bytes. Phase A now retains each validated worker response
+  in an anonymous temporary file under `--state-dir` and phase B replays it.
+  Reuse is an optional accelerator, never correctness-critical state: it is
+  bounded to 384 MiB of retained-plus-in-flight bytes and to an artifact-handle
+  cap derived from live `RLIMIT_NOFILE` and open-descriptor counts, both
+  re-probed at every admission; a 4 GiB-or-half-free filesystem floor is
+  reserved for the journal and phase-B staging first; and every artifact is
+  verified (physical length, content digest, JSON decode, worker-protocol
+  identity) before a single record is published. Anything that cannot be
+  measured conservatively declines and reparses — which is why the optimization
+  is **Linux-only** today: other platforms have no live descriptor evidence.
+  A restart has a frozen plan but no trusted handle, so it parses again; the
+  spool is deliberately not journal state. `--json` gains a
+  `pdf_extraction_reuse` block so the behaviour is observable rather than
+  inferred from timing. Original work by Leonid Bugaev (@buger).
+
 ## [1.0.0-rc.14] - 2026-08-10
 
 ### Changed
--- a/engine/crates/xerj-autoindex/src/cli.rs
+++ b/engine/crates/xerj-autoindex/src/cli.rs
@@ -130,9 +130,14 @@ pub fn print_help() {
              Image-only pages need OCR. Page parse failures reject the whole PDF instead\n\
              of silently creating a partial index. XERJ_PDF_WORKER_BIN is a trusted,\n\
              developer-only executable override.\n\
-             Current cost: phase-A sampling parses/materializes each complete PDF, and\n\
-             phase-B indexing parses it again. A framed, early-stop protocol is planned;\n\
-             use fewer --pdf-workers when parent memory is constrained.\n\
+             On Linux, fresh runs attempt to retain each validated PDF extraction in\n\
+             bounded anonymous storage under --state-dir. Live admission-time checks\n\
+             attempt to preserve disk and descriptor headroom; refused artifacts are\n\
+             parsed again in phase B and reported under pdf_extraction_reuse. Other\n\
+             platforms currently disable this optimization. Frozen-plan resumes skip\n\
+             phase A and parse unfinished PDFs once.\n\
+             --pdf-workers bounds both phase-A parser processes and phase-B replay\n\
+             materialization; --workers does not widen that PDF memory gate.\n\
          \n\
          EMBEDDINGS:\n\
              autoindex sends semantic_text to the running server; it does not choose the\n\
--- a/engine/crates/xerj-autoindex/src/extract/pdf.rs
+++ b/engine/crates/xerj-autoindex/src/extract/pdf.rs
@@ -45,6 +45,16 @@ const JOURNAL_FILESYSTEM_HEADROOM: u64 = 64 << 20;
 const MIN_DESCRIPTOR_HEADROOM: u64 = 64;
 const DESCRIPTOR_HEADROOM_PER_WORKER: u64 = 4;
 
+/// Serializes every test that mutates the process-global
+/// `XERJ_PDF_WORKER_BIN`/`XERJ_TEST_PDF_COUNT` pair. The variables are read
+/// inside `spawn_worker`, so a test that sets them changes the behaviour of
+/// every other test running concurrently in this binary — including tests in
+/// sibling modules. Take this before the first `set_var` and hold it until the
+/// restoring guard drops. Poison is ignored on purpose: a panicking test must
+/// not cascade into unrelated failures here.
+#[cfg(test)]
+pub(crate) static WORKER_BIN_ENV_LOCK: Mutex<()> = Mutex::new(());
+
 #[cfg(test)]
 static CORRUPT_REPLAY_SOURCE_SIZE: AtomicU64 = AtomicU64::new(u64::MAX);
 #[cfg(test)]
@@ -84,8 +94,11 @@ pub(crate) fn extract_and_spool(
 ) -> Result<(ExtractStats, Option<ExtractionSpool>, Option<SpoolFallback>)> {
     let _permit = worker_gate().acquire();
     let response = spawn_worker(path)?;
-    // Count completed parser protocol calls, not attempts that failed before
-    // producing a response which Phase A could use.
+    // Counted only once the worker has produced a usable protocol response.
+    // An invocation that spawned and then failed, timed out, or was killed is
+    // deliberately excluded — hence `..._responses` and not `..._calls` in the
+    // report. Total parser *invocations* are not a number this counter can
+    // honestly supply.
     budget.record_phase_a_parse();
     let (spool, fallback) =
         try_spool_response(state_dir, source_size, source_digest, budget, &response);
@@ -222,7 +235,7 @@ pub(crate) struct ExtractionSpoolBudget {
     exact_artifact_bytes: AtomicU64,
     peak_retained_or_reserved_bytes: AtomicU64,
     peak_live_artifacts: AtomicU64,
-    phase_a_pdf_parser_calls: AtomicU64,
+    phase_a_pdf_parser_responses: AtomicU64,
     capacity_status: &'static str,
     capacity_reason: &'static str,
     fallback_examples: Mutex<Vec<serde_json::Value>>,
@@ -272,7 +285,7 @@ impl ExtractionSpoolBudget {
             exact_artifact_bytes: AtomicU64::new(0),
             peak_retained_or_reserved_bytes: AtomicU64::new(0),
             peak_live_artifacts: AtomicU64::new(0),
-            phase_a_pdf_parser_calls: AtomicU64::new(0),
+            phase_a_pdf_parser_responses: AtomicU64::new(0),
             capacity_status: "enabled",
             capacity_reason: "explicit_budget",
             fallback_examples: Mutex::new(Vec::new()),
@@ -388,7 +401,7 @@ impl ExtractionSpoolBudget {
             exact_artifact_bytes: AtomicU64::new(0),
             peak_retained_or_reserved_bytes: AtomicU64::new(0),
             peak_live_artifacts: AtomicU64::new(0),
-            phase_a_pdf_parser_calls: AtomicU64::new(0),
+            phase_a_pdf_parser_responses: AtomicU64::new(0),
             capacity_status,
             capacity_reason,
             fallback_examples: Mutex::new(Vec::new()),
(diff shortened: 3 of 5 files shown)


xerj-org:
### Merge-result verification (base drift since CI)

The green run above built against base `12d5cc32`. `main` has since moved 6
commits, one of which — `b7441f83` "make section splitting linear" — rewrote
`engine/crates/xerj-autoindex/src/extract/mod.rs` (+349/-56), the module this
PR's phase-B replay path calls into. That combination had never been compiled,
so the green was not evidence about what would merge.

I merged current `main` into this branch locally and ran the gate on the
merged tree (x86_64 Linux, scoped build):

```
cargo fmt --all -- --check                                   pass
cargo clippy -p xerj-autoindex --all-targets -- -D warnings  pass
cargo test -p xerj-autoindex                                 288 passed, 0 failed
```

288 rather than the 283 in the description: the extra five are `b7441f83`'s own
section-splitting tests, which is the expected count for the merged tree. No
textual conflict (`git merge-tree` clean) and no semantic interaction — the
spool captures and replays worker responses within a single run, so both arms
always execute the same `extract/mod.rs`.

The branch is deliberately **not** rebased: GitHub reports `MERGEABLE` /
`CLEAN`, and rewriting the head would discard an hour of valid CI to chase a
base that keeps moving. The merged result is verified above instead.

Also re-confirmed against the tree, not the description: the three 2026-08-04
review blockers are genuinely closed (`WORKER_BIN_ENV_LOCK` at
`extract/pdf.rs:56` with `PdfWorkerEnvGuard::acquire()` at all seven env-var
call sites; the `source_size`/`source_digest` `ensure!` at `extract/pdf.rs:813`
now documented as a binding assertion with the count corrected to four
artifact checks at `:772-781`; and
`source_changed_during_phase_a_discards_the_artifact_and_names_the_reason`
present and passing).

xerj-org:
## CLA status — the one thing left, and it is not code

An adversarial re-review of this branch found the change sound, the regression
tests genuinely failing without the fix, and replay byte-identical to a direct
parse on real repository PDFs. It raised exactly one blocker, and it is a
decision rather than a defect. Recording it here so the decision is made
deliberately instead of by a green check.

**The green `cla-signed` check on this PR does not mean Leonid Bugaev signed.**
It means every *commit author* is `xerj-org`. Verified mechanically today:

- `gh api repos/xerj-org/xerj/pulls/248/commits` → `author_login: xerj-org` for
  all three commits (`c337eb8f`, `81c64883`, `9eb7b6b6`).
- `.clabot` resolves signatures from `.contributors`, currently
  `["xerj-org","dependabot[bot]","Vinz2168"]`. `buger` is not in it.
- The `Co-authored-by: Leonid Bugaev <leonsbox@gmail.com>` trailer that carries
  the real attribution is invisible to cla-bot, which only reads commit authors.
  That is why this PR is labelled `cla-signed` while #166 — the same work under
  buger's own authorship — carries no label at all.

Merging as-is would land a third party's design while the project's own CLA
gate reports green for a reason unrelated to them.

### Two ways to settle it

1. **buger signs.** One small pull request from their account adding `buger` to
   `.contributors`, per CLA.md. Asked directly on #166 — the earlier ask there
   framed signing as an optional authorship-credit choice, which was wrong and
   is corrected in that thread.
2. **The owner records an explicit waiver here** — a deliberate decision to
   accept the contribution on the `Co-authored-by` trailer without a signature.

### Context that belongs in that decision

Found while verifying the finding, not part of the review: this exposure is not
new. 57 commits authored `buger` / `Leonid Bugaev <leonsbox@gmail.com>` are
already on `main`, via #135, #154, #160, #161 and #168 (merged 2026-08-02 …
08-04). No `verification/cla-signed` check ran on any of them — cla-bot's checks
first appear around #186 — so those predate enforcement rather than evade it.
They are still unsigned, and a waiver scoped to this PR alone would not cover
them. `main` also has no branch protection, so the CLA check is advisory on
every PR. Tracked in #269, together with the suggestion to teach the *CLA gate
config* job to read `Co-authored-by:` trailers so this cannot recur silently.

### Held in draft

Not because anything is wrong with the branch — so that nothing merges it
before the question above is answered. `gh pr ready 248` once it is.

Local gate re-run on this head (`9eb7b6b6`) today, x86_64 Linux:

```
cargo fmt --all -- --check                                    pass
cargo clippy -p xerj-autoindex --all-targets -- -D warnings   pass
cargo test -p xerj-autoindex                                  283 passed, 0 failed
```

No code changed in this repair. The finding was correct, and the correct
response to it is a decision, not a commit. In particular `buger` was **not**
added to `.contributors` from this side: per CLA.md the signature is a pull
request from the contributor's own account, so adding the name here would forge
it rather than obtain it.

PR #253, opened 2026-08-09
chore(release): verify the published rc.13 artifacts, and make the check repeatable

### LINKED ISSUE #197: rc.12 post-release check: checksum, version string and search all verified on the published artifact
Post-release check on the published rc.12 artifact, not on a local build.

Downloaded `xerj-1.0.0-rc.12-x86_64-unknown-linux-gnu.tar.gz` from the release page.

```
size      19,627,395 bytes
sha256    d611a3929beecd180d8c345132f959af9cd03a41b41fe2e0806841e6db1627b5
          matches the published .sha256
--version xerj v1.0.0-rc.12
```

Started it on a clean data dir, indexed one document, searched for it:

```
cluster status  green
hits.total      1
max_score       0.28768212
took            0 ms
```

The version string is the part worth calling out. v1.0.0-rc.10 shipped binaries that reported rc.9, because the tag was cut at a commit where `engine/Cargo.toml` still held the old version. Asset names come from the git tag, the banner comes from `CARGO_PKG_VERSION`, so the two can disagree silently. rc.12 bumped the version before tagging and this confirms it survived to the artifact a user actually downloads.

Nothing to fix here. Filing it so the check is on the record and someone can repeat it for the next cut.

xerj-org:
Fixes #197

## What #197 asked for

Issue #197 recorded a manual post-release check of the **rc.12** tarball — checksum, version string, one indexed doc searched back — and closed with *"filing it so the check is on the record and someone can repeat it for the next cut."*

rc.13 has since shipped. This PR does two things: **re-runs that check against the current latest release**, and **turns the ritual into one command** so the next cut does not depend on someone remembering.

## Evidence: v1.0.0-rc.13 verified against the published artifacts

Not a local build — downloaded from the release page.

```
xerj-1.0.0-rc.13-x86_64-unknown-linux-gnu.tar.gz
size      19,396,449 bytes
sha256    b2ad2276b3386aeb9d509fa3e1131d9b57d57a59e5137adf33754658632f1b4c
          matches the published .sha256
--version xerj v1.0.0-rc.13
```

Booted on a clean data dir, indexed one document, searched it back:

```
cluster status  green
hits.total      1
max_score       0.57536423
took            0 ms
```

Widened past rc.12's single tarball — **all 8 published targets**:

| check | result |
|---|---|
| every target release.yml publishes is present on the release | 8/8 |
| `.sha256` companion published for every archive | 8/8 |
| checksum matches the download | 8/8 |
| archive carries binary + LICENSE + README | 8/8 |
| binary reports `1.0.0-rc.13` (incl. macOS + Windows) | 8/8 |
| `curl -fsSL https://xerj.org/get \| sh` resolves the current *latest* release, verifies sha256, installs, `--version` agrees | pass — resolved rc.13 when first measured; re-run 2026-08-10 it resolves **rc.14** and still agrees |

The version string is the part worth calling out, exactly as #197 said. Re-confirmed at the tag: `engine/Cargo.toml` at `v1.0.0-rc.13` (`de13be53`) holds `1.0.0-rc.13`, and that value survived to the binary a user downloads on every one of the eight targets.

**No defect found in the rc.13 release artifacts.**

## Why a script, and not just a comment on the issue

The bug class #197 exists to watch for is still live and still reproducible **today**, from the published release page:

```
$ ./scripts/verify-release.sh v1.0.0-rc.10
  FAIL  aarch64-apple-darwin — tag says 1.0.0-rc.10, binary says: 1.0.0-rc.9
  ... all 8 targets ...
  FAIL  --version  ->  'xerj v1.0.0-rc.9', expected 'xerj v1.0.0-rc.10'
  FAIL  v1.0.0-rc.10: 9 check(s) failed — do not announce this release       exit 1

$ ./scripts/verify-release.sh v1.0.0-rc.13
  PASS  v1.0.0-rc.13: checksums and versions verified on 8/8 targets,
        plus a live search                                                   exit 0
```

That is the regression test: **fails on the known-bad release, passes on the current one**, against real published bytes rather than a fixture. rc.10's CI was green and every test passed — asset filenames come from the git tag, the banner comes from `CARGO_PKG_VERSION`, and nothing made the two agree. Only running the artifact shows it.

## What the script checks

1. **every target we ship is present.** The eight targets are declared in the script and asserted, not read off the release page — see below
2. every archive has a `.sha256` companion, and matches it
3. every archive contains binary + LICENSE + README. An archive that matched its published checksum and still will not extract is a defect in the *release*, so it **fails** — it is not rounded down to "this host could not read it"
4. every binary reports the tag's version — **including the macOS and Windows binaries a Linux runner cannot execute**, by reading the startup-banner bytes out of the binary instead of running it. Drift is just as likely to land in a target the checker cannot boot, so checking only the native one would have been the weaker check
5. the host-native binary: `--version`, boot on a clean data dir, cluster green, index a doc, search it back, terms aggregation. **Linux and macOS hosts, both architectures.** On a host where it cannot run one — a Windows host, or any platform we do not ship a binary for — it says so and the run **fails**; it does not report a PASS for a release nobody executed

```
scripts/verify-release.sh                  # latest release
scripts/verify-release.sh v1.0.0-rc.13     # a specific tag
scripts/verify-release.sh --no-smoke       # checksums + versions only
```

**Honesty behaviours, deliberately.** A gate whose failure text is *"do not announce this release"* must never print an unqualified PASS for a run that checked less than it claims:

- with `--no-smoke` the verdict says *"NO binary was executed (`--no-smoke`), so nothing here says it runs"* instead of claiming a live search it never ran. That exits 0 — it is a narrowing of scope the operator asked for, and it is the only one that does.
- a host where **no** binary could be executed and nobody asked for that — a Windows host, an unsupported platform — is counted like a skipped target and **fails closed with exit 1**, naming the host and the remedy. An earlier revision of this PR exited 0 there; see the second repair table below.
- a target that **could not** be checked (e.g. a Windows `.zip` on a host with no `unzip`) is counted, printed as `SKIP`, and **fails closed with exit 1**: nobody asked for that reduction in scope, so it is not rounded up into a PASS. The verdict names the count and the remedy.
- a broken archive **fails and still reaches a verdict**: the extractors are guarded, so `set -e` cannot end the run mid-step with no verdict at all. See the third repair round below.
- PASS is only ever printed as `verified on 8/8 targets`.
- ports are allocated free rather than hardcoded, so it will not collide with a xerj already listening on 9200.

## Review repairs (commit `dbfcabaf`)

Adversarial review of the first revision found the gate could still greenlight a release that was **missing platforms**. Three connected holes, all fixed, each with the fault re-injected and re-run:

| fault injected | before | after |
|---|---|---|
| two `windows-msvc` archives absent from the release | `6 archives`, 6/6 PASS every step, verdict "checksums and versions verified", **exit 0** | 2× `FAIL … NO archive published for this target`, **exit 1** |
| host without `unzip` (all 8 archives present) | step 4 prints 6 lines, both Windows binaries never version-checked, nothing marks it, **exit 0** | 2× `SKIP … version NOT checked`, verdict `2 of 8 target(s) NOT verified`, **exit 1** |

Root cause of the first: the target set was whatever `gh release download` happened to fetch, so one failed leg of the `release.yml` matrix shrank the gate instead of failing it. The eight targets are now a declared constant checked before anything else. Root cause of the second: a bare `continue` dropped an unreadable archive out of the section headed *"on every target"* with no output at all.

Re-verified after the repair, against the real published bytes: **rc.13 PASS 8/8 + live search, exit 0 (13s)**; **rc.10 FAIL 9 checks, exit 1** — the new step adds no spurious failure, rc.10 did publish all eight targets, they just all carried the rc.9 banner.

## Second repair round (commit `218f3fda`) — the verifier no-op'd on Apple Silicon

The merge gate re-ran every measurement above against the real published bytes and found one hole the first repair left open, on the host a maintainer is *most* likely to verify a release from.

`verify-release.sh` built its host glob out of a raw `uname -m`. An Apple Silicon Mac reports `arm64`; the artifact is named `aarch64-apple-darwin`. They never matched. So the Mac ARM archive was downloaded, unpacked and version-checked as PASS at step 5 — and then step 6 printed `SKIP no artifact for Darwin/arm64 — nothing to execute here` and the run exited **0 PASS**.

That is three defects in one line:

- **the check this script exists for never ran, and the verdict still said PASS.** Everything green, nobody executed the binary — the rc.10 shape, reappearing inside the tool written to catch it, and the repo's accepted-and-silently-ignored class (#204).
- **the message was false on its face.** The Mac ARM artifact *was* published, and this very run had already unpacked it. An operator reading that line would conclude we ship no Mac ARM build.
- it was **the only `warn` in the script incrementing neither `FAILURES` nor `SKIPPED_TARGETS`**, contradicting the rule the first repair wrote 74 lines below it: *"a skipped target is not something anyone asked for, so it fails closed."*

| fault | before (`dbfcabaf`) | after (`218f3fda`) |
|---|---|---|
| Apple Silicon host (`uname` shim `Darwin`/`arm64`, real rc.13 assets) | 8/8 PASS on steps 2–5 including `aarch64-apple-darwin reports 1.0.0-rc.13`, then `SKIP no artifact for Darwin/arm64`, verdict **PASS, exit 0** | glob normalises to `aarch64`, `unpack/…-aarch64-apple-darwin/xerj` is selected and executed; step 6 no longer skips, **exit 1** on this Linux host (see the honest caveat below) |
| host that genuinely cannot run any artifact (`uname` shim `FreeBSD/amd64`) | verdict PASS, exit 0 | `SKIP this script does not boot a release binary on FreeBSD/amd64 (Linux and macOS hosts only) …`, verdict `nothing was executed — no binary ran on this host and nobody asked for that`, **exit 1**, remedy named |

Also fixed while in here: `x=$(find … | head -1)` appears five times under `set -o pipefail`. When `head` exits first, `find` takes SIGPIPE, the pipeline reports 141, and `set -e` aborts the whole verifier mid-run with no verdict at all. `|| true` on those five assignments. Latent, never observed to fire — fixed because a release gate that can vanish silently is the same bug class as the one above.

Everything re-measured after the fix, against the real published bytes:

| run | result |
|---|---|
| real Linux x86_64, rc.13 | 8/8 + live search, `max_score 0.57536423`, **PASS exit 0** (unchanged) |
| `--no-smoke` | `NO binary was executed (--no-smoke)`, **PASS exit 0** |
| no `unzip` on PATH | `2 of 8 target(s) NOT verified`, **exit 1** (unchanged) |
| two `windows-msvc` archives absent | 2× `NO archive published for this target`, **exit 1** (unchanged) |
| `v1.0.0-rc.10` | 9 check(s) failed, **exit 1** (unchanged) |

## Third repair round (commit `f8472018`) — independent re-measurement, and one more hole closed

Every row above was re-run from scratch against the **real published bytes**, not inferred from the diff, and the fault injections were repeated rather than trusted. The Apple Silicon hole is confirmed closed.

Re-measuring turned up one last way the verifier could stop short of its own header promise (*"anything this run could not check is counted and reported… the run then exits non-zero"*): **an archive that matched its published checksum and still would not extract killed the script mid-run.** `tar -xzf` / `unzip` ran unguarded, so `set -e` ended the run with the extractor's own error and **no verdict line at all** — fault-injected with a garbage `…-aarch64-unknown-linux-musl.tar.gz` plus a matching `.sha256`, observed `exit 2`, output stopping after three PASS lines in step 4.

It could never become a false PASS, but it was the one partial check that was neither counted nor reported. Now: a checksum-clean archive that will not extract is treated as a defect in the **release**, so it `FAIL`s (not `SKIP`s), the run reaches a verdict, and step 5 deliberately does not *also* count it as a skipped target — that would attach the "install unzip" remedy to what is actually a broken artifact.

| fault | before | after |
|---|---|---|
| `.tar.gz` matches its `.sha256` but is not a gzip stream | script aborts inside step 4, **no verdict**, `exit 2` | `FAIL … matched its published checksum but did not extract`, `SKIP … version NOT checked (archive did not extract, see step 4)`, `FAIL v1.0.0-rc.13: 1 check(s) failed — do not announce this release`, **exit 1** |
| `.zip` matches its `.sha256` but has no central directory | same abort, **no verdict** | same `FAIL` pair and verdict, **exit 1** |

Full matrix at the head of this branch — every row an actual run, exit codes as printed:

| run | observed | exit |
|---|---|---|
| Linux x86_64, `v1.0.0-rc.13` | steps 2–5 8/8, step 6 boots green, indexes, `max_score 0.57536423`, terms agg; `PASS … 8/8 targets, plus a live search` | 0 |
| `uname` shim `Darwin`/`arm64`, rc.13 | step 6 **no longer skips** — it selects and exec's `unpack/…-aarch64-apple-darwin/…/xerj`, which on a Linux host cannot run (`Exec format error`) → `FAIL --version`, `FAIL server never answered`, `2 check(s) failed` | 1 |
| `--no-smoke`, rc.13 | `SKIP --no-smoke — no binary was executed`; `PASS … NO binary was executed (--no-smoke), so nothing here says it runs` | 0 |
| `uname` shim `FreeBSD`/`amd64`, rc.13 | `SKIP this script does not boot a release binary on FreeBSD/amd64 …`; `FAIL … nothing was executed — no binary ran on this host and nobody asked for that` | 1 |
| PATH mirrored without `unzip`, rc.13 | 2× `SKIP … contents NOT checked`, 2× `SKIP … version NOT checked`, `FAIL … 2 of 8 target(s) NOT verified` | 1 |
| both `windows-msvc` archives removed from the asset set | 2× `FAIL … NO archive published for this target`, `FAIL … 2 check(s) failed` | 1 |
| `v1.0.0-rc.10` (the known-bad release) | 8× `tag says 1.0.0-rc.10, binary says: 1.0.0-rc.9` + `FAIL --version -> 'xerj v1.0.0-rc.9'`, `9 check(s) failed` | 1 |
| checksum-clean `.tar.gz` that will not extract | `FAIL … did not extract`, verdict `1 check(s) failed` | 1 |
| checksum-clean `.zip` that will not extract | `FAIL … did not extract`, verdict `1 check(s) failed` | 1 |
| **`v1.0.0-rc.14` — the release that is *currently* latest** | 8/8 on every step + live search, `PASS v1.0.0-rc.14: checksums and versions verified on 8/8 targets, plus a live search` | 0 |

The last row is the check #197 asked to be repeatable, run against the release that shipped while this PR was open: **no defect found in the rc.14 release artifacts either.**

Also re-confirmed at head: `EXPECTED_TARGETS` still matches `release.yml`'s matrix 8-for-8; `bash -n` clean; `cargo fmt --all -- --check` clean.

### Known gaps, stated plainly

- **The Apple Silicon path is proven only up to selecting and exec'ing the right binary.** No macOS hardware was available; the measurement used a `uname` shim on a Linux host, where the Mach-O then fails with `Exec format error`. That is the shim's limit, not the script's — what it proves is that step 6 now resolves `…-aarch64-apple-darwin/xerj` and runs it instead of skipping. The boot/index/search half of step 6 on a real Mac is unverified by this PR.
- **The Windows `.exe` is never booted.** Windows archives are checksum- and version-checked like every other target; on a Windows host the script says so and exits 1 unless `--no-smoke` is passed. Booting the `.exe` from MSYS/Git-Bash is not attempted (path translation into the config file is its own problem), so a Windows-host smoke remains a gap.
- **`EXPECTED_TARGETS` is a hand-maintained constant.** It is asserted, not discovered — that is the point — but nothing mechanically ties it to `release.yml`'s matrix. It matched 8-for-8 when written; a future matrix change has to update both.
- **Not wired into CI.** This is a command a human or agent runs after a release, not a gate that runs itself. See Scope.
- **`--no-smoke` is still a real reduction in scope, and it exits 0.** It is the only narrowing that does, it is opt-in, and the verdict text says in full that no binary ran — but an operator who habitually passes it gets exactly the rc.10 blind spot back. There is no guard against that beyond the wording.
- **The first commit's message quotes an older verdict string.** `e1011255`'s body says `PASS v1.0.0-rc.13: checksums, versions and a live search all verified`; the wording the code actually prints at head is `checksums and versions verified on 8/8 targets, plus a live search`. Superseded by the two repair commits — recorded here so the history log is not read as a claim about current behaviour.

## Scope

No engine code touched. `cargo fmt --all -- --check` clean; no Rust crate changed, so there is no clippy/test surface in this diff. `.github/workflows/` deliberately untouched — wiring this into the release workflow as a post-publish gate is a judgement call for the owner, and it is the obvious follow-up.

## Separate finding, NOT fixed here

While checking version drift across shipped surfaces I found the same class of drift in the Helm chart, verified live:

- `deploy/helm/xerj/Chart.yaml` still pins `version: 1.0.0-rc.1` and `appVersion: "1.0.0-rc.1"`, 13 releases stale as of rc.14 — and `values.yaml` defaults the image tag to `.Chart.AppVersion`, so a default `helm install` asks for tag `1.0.0-rc.1`.
- `image.repository: xerj/xerj` returns **404 on Docker Hub** (`hub.docker.com/v2/repositories/xerj/xerj/` → 404). A `Dockerfile` does exist — `engine/Dockerfile`, tracked on `main` since `cbd19714` alongside `engine/docker-compose.yml`, and its header documents building `xerj/xerj`, the very repository `values.yaml:5` points at. What is missing is the **publishing** half: no workflow in `.github/workflows/` references `docker/build-push-action`, `docker buildx`, `ghcr.io` or `docker push`, so no tag of that image is ever pushed and the chart cannot pull one.
- `home:` and `sources:` point at `github.com/xerj-ai/xerj`, which **404s** (the repo is `xerj-org/xerj`).

rc.13's release notes advertise a Helm fix (`3e2a0a14`, "helm secure default"), so this is a surface we ship and claim. A default install fails with `ImagePullBackOff` — accepted-and-ignored input, the #204 class. Left out of this PR on purpose: it needs a product decision (publish the image the Dockerfile already describes, or mark the chart unsupported), not a silent edit.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,6 +61,7 @@ cannot push at all: [.github/AI_CONTRIBUTIONS.md](./.github/AI_CONTRIBUTIONS.md)
 | Architecture map | [docs/ARCHITECTURE.md](./docs/ARCHITECTURE.md) |
 | Pre-submission and review protocol | [docs/CONTRIBUTION_REVIEW.md](./docs/CONTRIBUTION_REVIEW.md) |
 | How to send a bug or a fix back upstream (written for agents) | [.github/AI_CONTRIBUTIONS.md](./.github/AI_CONTRIBUTIONS.md) |
+| Checking a published release is actually good | `scripts/verify-release.sh [tag]` — downloads every asset, verifies checksums, asserts each binary reports its tag's version, boots the host-native one and runs a search |
 
 ## How to evaluate this project fairly
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,7 +61,7 @@ cannot push at all: [.github/AI_CONTRIBUTIONS.md](./.github/AI_CONTRIBUTIONS.md)
 | Architecture map | [docs/ARCHITECTURE.md](./docs/ARCHITECTURE.md) |
 | Pre-submission and review protocol | [docs/CONTRIBUTION_REVIEW.md](./docs/CONTRIBUTION_REVIEW.md) |
 | How to send a bug or a fix back upstream (written for agents) | [.github/AI_CONTRIBUTIONS.md](./.github/AI_CONTRIBUTIONS.md) |
-| Checking a published release is actually good | `scripts/verify-release.sh [tag]` — downloads every asset, verifies checksums, asserts each binary reports its tag's version, boots the host-native one and runs a search |
+| Checking a published release is actually good | `scripts/verify-release.sh [tag]` — downloads every asset, asserts every target we ship is present, verifies checksums, asserts each binary reports its tag's version, boots the host-native one and runs a search |
 
 ## How to evaluate this project fairly
 
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -61,7 +61,7 @@ cannot push at all: [.github/AI_CONTRIBUTIONS.md](./.github/AI_CONTRIBUTIONS.md)
 | Architecture map | [docs/ARCHITECTURE.md](./docs/ARCHITECTURE.md) |
 | Pre-submission and review protocol | [docs/CONTRIBUTION_REVIEW.md](./docs/CONTRIBUTION_REVIEW.md) |
 | How to send a bug or a fix back upstream (written for agents) | [.github/AI_CONTRIBUTIONS.md](./.github/AI_CONTRIBUTIONS.md) |
-| Checking a published release is actually good | `scripts/verify-release.sh [tag]` — downloads every asset, asserts every target we ship is present, verifies checksums, asserts each binary reports its tag's version, boots the host-native one and runs a search |
+| Checking a published release is actually good | `scripts/verify-release.sh [tag]` — downloads every asset, asserts every target we ship is present, verifies checksums, asserts each binary reports its tag's version, then boots the host-native binary and runs a real search (Linux and macOS hosts; on a host where it cannot run one it says so and exits non-zero rather than passing) |
 
 ## How to evaluate this project fairly
 
--- a/scripts/verify-release.sh
+++ b/scripts/verify-release.sh
@@ -21,7 +21,9 @@
 #      the release page, so a release that lost a matrix leg fails instead of
 #      quietly verifying the targets that did make it
 #   2. every archive has a .sha256 companion, and matches it
-#   3. every archive contains the binary + LICENSE + README
+#   3. every archive contains the binary + LICENSE + README. An archive that
+#      matched its published checksum and still will not extract is a defect in
+#      the release, so it FAILS rather than being skipped
 #   4. every binary — including the ones this host cannot execute — carries the
 #      tag's version in its startup banner, and carries no other version
 #   5. the host-native binary: --version, boot on a clean data dir, health
@@ -96,6 +98,10 @@ SKIPPED_TARGETS=0
 # Set when step 6 did not execute anything for a reason the operator did NOT
 # ask for (no runnable artifact for this host). --no-smoke does not set it.
 SKIPPED_SMOKE=0
+# Archives that were present and checksum-clean but would not extract. Already
+# counted in FAILURES at step 4; tracked so step 5 does not ALSO count them as
+# skipped targets and hand the operator the wrong remedy.
+UNPACKABLE=""
 pass() { printf '  %sPASS%s  %s\n' "$GRN" "$RST" "$1"; }
 fail() { printf '  %sFAIL%s  %s\n' "$RED" "$RST" "$1"; FAILURES=$((FAILURES + 1)); }
 warn() { printf '  %sSKIP%s  %s\n' "$YEL" "$RST" "$1"; }
@@ -186,11 +192,27 @@ step "4. archive contents"
 for a in $ARCHIVES; do
   d="unpack/${a%.tar.gz}"; d="${d%.zip}"
   mkdir -p "$d"
+  # An archive that matched its published checksum and still will not extract is
+  # a defect in the RELEASE, not a gap in this host's tooling — so it is a FAIL,
+  # not a SKIP. Guarding the extractor also keeps `set -e` from ending the run
+  # right here with the extractor's own error and no verdict at all, which is
+  # the one way this script could stop short of the promise in its header that
+  # everything it could not check is counted and reported.
   case "$a" in
-    *.tar.gz) tar -xzf "$a" -C "$d" ;;
+    *.tar.gz)
+      if ! tar -xzf "$a" -C "$d"; then
+        fail "$a — matched its published checksum but did not extract"
+        UNPACKABLE="$UNPACKABLE $a"
+        continue
+      fi ;;
     *.zip)
-      if command -v unzip >/dev/null 2>&1; then unzip -q -o "$a" -d "$d"
-      else warn "$a — unzip not on PATH, contents NOT checked"; continue; fi ;;
+      if ! command -v unzip >/dev/null 2>&1; then
+        warn "$a — unzip not on PATH, contents NOT checked"; continue
+      elif ! unzip -q -o "$a" -d "$d"; then
+        fail "$a — matched its published checksum but did not extract"
+        UNPACKABLE="$UNPACKABLE $a"
+        continue
+      fi ;;
   esac
   # `|| true`: under `set -o pipefail`, find killed by SIGPIPE once head exits
   # returns 141 for the whole pipeline and `set -e` would abort the verifier
@@ -211,11 +233,20 @@ for a in $ARCHIVES; do
   target=$(printf '%s' "$a" | sed "s/^xerj-$EXPECTED-//; s/\.tar\.gz$//; s/\.zip$//")
   bin=$(find "$d" -type f \( -name xerj -o -name xerj.exe \) 2>/dev/null | head -1 || true)
   if [ -z "$bin" ]; then
-    # Step 4 could not unpack this archive (typically: no unzip for a Windows
-    # .zip). Dropping it here would leave this section headed "on every target"
-    # quietly checking fewer targets than it names.
-    warn "$target — no binary unpacked, version NOT checked"
-    SKIPPED_TARGETS=$((SKIPPED_TARGETS + 1))
+    case " $UNPACKABLE " in
+      *" $a "*)
+        # Step 4 already counted this one as a FAIL: the archive is broken, and
+        # the run is already failing closed for the right reason. Counting it a
+        # second time as a skipped target would add the "install unzip" remedy
+        # to the verdict, which is the wrong advice for a defective release.
+        warn "$target — version NOT checked (archive did not extract, see step 4)" ;;
+      *)
+        # Step 4 could not unpack this archive (typically: no unzip for a
+        # Windows .zip). Dropping it here would leave this section headed "on
+        # every target" quietly checking fewer targets than it names.
+        warn "$target — no binary unpacked, version NOT checked"
+        SKIPPED_TARGETS=$((SKIPPED_TARGETS + 1)) ;;
+    esac
     continue
   fi
   found=$(banner_versions "$bin" | tr '\n' ' ' | sed 's/ $//')
--- a/scripts/verify-release.sh
+++ b/scripts/verify-release.sh
@@ -25,11 +25,17 @@
 #   4. every binary — including the ones this host cannot execute — carries the
 #      tag's version in its startup banner, and carries no other version
 #   5. the host-native binary: --version, boot on a clean data dir, health
-#      green, index a doc, search it back, run a terms aggregation
+#      green, index a doc, search it back, run a terms aggregation. Runs on
+#      Linux and macOS hosts (both architectures). It is NOT run on a Windows
+#      host — those archives are checksum- and version-checked like every
+#      other target, but this script never boots the .exe.
 #
-# A target this host cannot unpack (no unzip for the Windows .zip) is counted
-# and reported, never silently dropped: the run then exits non-zero, because a
-# partially-checked release is not a verified release.
+# Anything this run could not check is counted and reported, never silently
+# dropped, and the run then exits non-zero — a partially-checked release is not
+# a verified release. That covers a target this host cannot unpack (no unzip
+# for the Windows .zip) and a check-5 smoke (printed as "step 6" at run time)
+# that never executed because this host has no runnable artifact. The one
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent re-verification (second agent, fresh machine state)

Everything below was re-run from scratch against the published bytes, not copied from the PR body. Every number agrees.

**Manual check, the way #197 did it** — `xerj-1.0.0-rc.13-x86_64-unknown-linux-gnu.tar.gz` downloaded from the release page:

```
size      19,396,449 bytes
sha256    b2ad2276b3386aeb9d509fa3e1131d9b57d57a59e5137adf33754658632f1b4c
          sha256sum -c  ->  OK   (matches the published .sha256)
--version xerj v1.0.0-rc.13
banner    xerj v1.0.0-rc.13 starting
```

Booted on a clean data dir, one doc indexed, searched back, aggregated:

```
cluster status  green
hits.total      1
max_score       0.57536423
took            0 ms
terms agg       {"key":"release verification","doc_count":1}
```

**The script, both directions.** This is the part that makes it a regression check rather than a report:

```
scripts/verify-release.sh v1.0.0-rc.13   ->  8/8 checksums, 8/8 contents, 8/8 versions,
                                             --version + boot + search + agg     exit 0
scripts/verify-release.sh v1.0.0-rc.10   ->  9 FAIL: all 8 targets say 1.0.0-rc.9,
                                             --version says rc.9                 exit 1
```

**Root cause of the rc.10 failure confirmed at the tag**, which is what makes the negative control meaningful rather than incidental:

```
git show v1.0.0-rc.10:engine/Cargo.toml | grep version   ->  version = "1.0.0-rc.9"
git show v1.0.0-rc.13:engine/Cargo.toml | grep version   ->  version = "1.0.0-rc.13"
```

**Installer path, end to end**, `curl -fsSL https://xerj.org/get | sh` into a scratch dir:

```
resolving latest release of xerj-org/xerj…
installing xerj v1.0.0-rc.13 for x86_64-unknown-linux-musl
ok  sha256 verified
ok  installed .../xerj
$ xerj --version   ->  xerj v1.0.0-rc.13
```

`cargo fmt --all -- --check` clean. No Rust crate in the diff, so no clippy/test surface. `bash -n` clean.

**No defect found in the rc.13 release artifacts.**

### One gap this PR does not close, flagged deliberately

`scripts/verify-release.sh` catches tag/`Cargo.toml` version drift *after* the release is published. Nothing prevents publishing it: `.github/workflows/release.yml` contains no assertion that the pushed tag matches `workspace.package.version` (`grep -n 'CARGO_PKG_VERSION\|Cargo.toml' .github/workflows/release.yml` returns nothing). The rc.10 hole is therefore still open at the point of cutting — this PR makes it *detectable in one command* rather than *impossible*. Closing it properly is a workflow change and belongs to whoever owns the release job.

Issue #238, opened 2026-08-09
autoindex: added-but-skipped files leave immortal live catalog entries


xerj-org:
## What happens

`xerj autoindex` writes a live catalog document for every junk/skipped file it
reports. Nothing ever deletes one. For files that are in the durable resume
plan that is at least *recoverable* — the plan remembers them, so a
reconciliation pass has something to work from. For a file **added after the
plan was frozen** there is no record at all:

`engine/crates/xerj-autoindex/src/lib.rs:1250-1270` builds `new_unplanned` — the
per-run list of files that appeared after the plan was frozen — and
`lib.rs:2220-2222` writes them into the catalog beside the planned junk:

```rust
let mut all_junk: Vec<&JunkFile> = plan.junk_files.iter().collect();
all_junk.extend(extra.iter());
all_junk.extend(new_unplanned.iter());   // ← live doc, no durable record
```

`new_unplanned` is a local `Vec` that dies with the run. The `file:{key}`
catalog document it produced outlives it forever: delete the file from the
corpus and the next run has no memory that the document was ever written, so
nothing can ever remove it. The catalog — which is the data map every
`xerj autoindex map`, `status` and agent query reads — keeps advertising a file
that is not there.

Reported by the maintainer during the review of #156
(<https://github.com/xerj-org/xerj/pull/156#issuecomment-3162>, "A pre-existing
invariant hole this PR makes routine"), with the request that it be filed
separately rather than widening that PR's scope. It is not introduced by #156;
that PR's newly-blessed add-then-rerun path just makes it the common case.

## Reproduce

```sh
mkdir -p /tmp/junkcat && printf 'id,v\n1,a\n' > /tmp/junkcat/planned.csv
xerj autoindex /tmp/junkcat --url http://localhost:9200 --prefix junkcat --no-graph

printf 'id,v\n2,b\n' > /tmp/junkcat/added.csv     # appears after the plan froze
xerj autoindex /tmp/junkcat --url http://localhost:9200 --prefix junkcat --no-graph
curl -s 'localhost:9200/autoindex-catalog/_search?q=path:added.csv' # 1 hit, status "skipped"

rm /tmp/junkcat/added.csv
xerj autoindex /tmp/junkcat --url http://localhost:9200 --prefix junkcat --no-graph
curl -s 'localhost:9200/autoindex-catalog/_search?q=path:added.csv' # STILL 1 hit — forever
```

## Why it is safe to fix properly

[...]

Issue #195, opened 2026-08-07
autoindex reports success and indexes nothing when a write block is active


xerj-org:
`xerj autoindex` reports success and indexes nothing when the server has a write block on.

## What happened

Ran autoindex over 262 files. It printed a normal progress trace:

```
autoindex: 262 files (9 MB) under /home/claude/ai/xerj/engine/crates
phase A: 4 datasets inferred, 0 junk/skipped files
phase B: indexing 262 files with 8 workers -> http://localhost:9200
```

Four indices were created. All four hold zero documents.

```
green open selfaudit-xerj                 1 0 0 0 4.2kb
green open selfaudit-xerj-2               1 0 0 0 4.9kb
green open selfaudit-xerj-wasm-examples   1 0 0 0 5.2kb
green open selfaudit-xerj-vector-tests    1 0 0 0 3.4kb
```

The cause was in the server log, not the client output:

```
WARN xerj_engine::governor: disk flood-stage watermark crossed
  - engaging read_only_allow_delete write block used_pct=95 flood_pct=95
```

## The block is right, the silence is not

Engaging a write block at the flood-stage watermark is correct and matches ES semantics. The problem is that autoindex swallowed every rejected bulk and still walked to completion. Nothing in the client output says a single document failed.

A user gets an index that exists, has a mapping, reports green health, and returns nothing for every query. Then they go looking for a bug in their query.

This is worse in the agent case, which is the case we care most about. An agent has no server log in front of it. It sees "indexed", queries, gets zero hits, and has no way to tell "the corpus does not contain that" from "nothing was ever written".

## Expected

Autoindex should fail loudly on rejected bulks. At minimum:

- non-zero exit when the indexed count is zero but files were submitted
- the rejection reason surfaced from the bulk response, not just a count
- a final line stating documents indexed versus files submitted, so a mismatch is visible without reading a log on another machine

## Repro

Fill a disk past 95%, or set the flood-stage watermark low, then run autoindex at any folder. Check `_count` after.

Found while indexing XERJ's own source to review it with XERJ.

xerj-org:
Narrowing this after another run.

Bulk failures are not uniformly silent. With the disk block lifted but the server under load, the same command failed loudly and correctly:

```
error: autoindex stopped with bulk/backend failures: _bulk: HTTP 429 Too Many Requests
  (x5). Failed source files were not journaled complete; fix the reported server or
  embedding configuration and rerun the same command to resume safely
```

[...]

PR #254, opened 2026-08-09
fix(autoindex): refuse reruns that would strand documents (supersedes #156)

xerj-org:
Supersedes #156 by @buger.

His work, our rebase. #156's head is on a fork with `maintainerCanModify=false`, so nobody in xerj-org can push to it — every attempt 403s. This branch carries his commits across with authorship preserved (`buger <leonsbox@gmail.com>` on two of them), rebased onto current `main`, with the outstanding review blockers fixed. Please close #156 when this merges.

## What it does

A durable resume plan is a crash-resume boundary, not a folder-sync generation. Deleting an indexed file and rerunning `xerj autoindex` used to exit 0 while the deleted file's documents, aliases, graph edges and catalog entries stayed live and searchable with no source behind them. That rerun is now refused before any remote call other than the endpoint-readiness ping, with a message that names every removed file and three recovery routes. `xerj brain` stops turning an absent or zero node-count probe into an automatic reset.

## Review blockers fixed on top of #156

**1. `--fresh` had stopped being an escape hatch.** `Journal::preflight` read the durable plan unconditionally and propagated every failure, so a malformed journal record or a root/url/prefix mismatch aborted the run before the `fresh` flag was consulted in `open_after_preflight` — which deletes that journal unread. The refusal text recommends `--fresh` as the recovery from exactly the error that had begun to block it, leaving a state directory no supported flag could repair.

`fresh` is now threaded into `preflight`; under `--fresh` an unreadable plan becomes `JournalPreflight::unreadable_plan` instead of an error. The preflight is never more fatal than the open it precedes. Without `--fresh` the refusal is byte-for-byte unchanged. Because the removal gate compares against `preflight.plan`, discarding an unreadable plan also discards its comparison basis — the run says so on stderr rather than proceeding quietly.

Peer precedent for separating verification from opening once the caller has opted into a rebuild: redb `Database::do_repair` (`redb/src/db.rs:994`, Apache-2.0/MIT) enters a repair session on a failed checksum instead of failing the open, and still refuses when repair is genuinely impossible — which is why only the `--fresh` path is softened here.

**2. Unbounded error rendering.** `UnsupportedInventoryDeltaError::fmt` rendered every entry into one `String` with no cap, while the duplicate and unplanned-file listings 400 lines away already capped at ten. Unmounting a bind mount under an indexed root makes every content group vanish at once, so on an 82k-file journal the refusal was megabytes of stderr — in the one code path whose entire job is to be read by a person. All three listings now share `REFUSAL_LIST_CAP` and the same "… and N more" tail. `--json` still carries every entry.

**3. `xerj brain` took the wrong branch for the case its new message exists for.** `live_node_docs` correctly reported absence for a missing meta doc or a missing `nodes_index` field, then called `es.search(...)?`, which propagates on non-2xx. A meta doc that outlives its nodes index — a wiped data dir, the exact condition the probe was added to catch — 404s, and `?` short-circuited before `journal_server_disagreement()` could print the recovery text. The operator got a raw HTTP 404 instead. New `Es::search_present` returns `Ok(None)` for 404 and behaves like `search` for every other status, so absence is never laundered out of a real failure.

**4. PR body accuracy.** #156's description still claimed "adding or removing a canonical content group is detected and rejected". After the rescope in `b0ed1c81`, `InventoryDelta::refuses()` is `!vanished_content_groups.is_empty()` — additions are reported and skipped, not fatal. This body describes what the code does.

## Rebase: four semantic conflicts, not just textual ones

`main` moved 110 commits since #156's merge base. Four conflicts needed decisions, not just hunk picking. The first two were found before the first push; the last two were found by review + CI on the pushed head and fixed in `420c5834` and `8c1667d3`:

**#238's junk sweep vs. the removal gate.** The gate built its vanished set from `plan.files` **and** `plan.junk_files`, on the reasoning that a deleted junk file left an immortal `file:{key}` catalog row nobody would clean up. True when the gate was written; no longer true — main's stale junk-catalog sweep deletes that row, then drops the plan entry only after the delete lands. A junk file publishes no documents, aliases or graph edges, so its removal now strands nothing. Merging the two unchanged made the gate refuse the very rerun the sweep exists to perform: main's own #238 regression test `an_added_then_deleted_file_leaves_no_immortal_catalog_entry` failed with the removal refusal naming `added.csv`.

The vanished set is now `plan.files` only — the files that actually published documents. Junk keys stay in `durable_keys`, so an unchanged skipped file is still not classified as an addition. `deleted_planned_junk_fails_closed_before_catalog_can_stay_stale` encoded the superseded rule and is replaced by `deleted_planned_junk_is_swept_rather_than_refused`, which pins both halves of the new boundary. The removal gate is also placed *before* the sweep computation, so a refused rerun computes nothing and mutates nothing.

**#195's counter redefinition.** `main` redefined `records_total` to the live server-side count and added `records_submitted_this_run` for the run-local one. `a_rerun_after_an_added_file_succeeds_and_fresh_absorbs_it` asserted the old run-local semantics on the new name. It now pins both: the rerun submits zero files and zero records, and the record the first run published is still live.

`MockState` keeps both main's `catalog_docs` (data counts are verified against the journal since #195, and catalog rows are not data rows) and this branch's `requests` recorder, which is what the refusal tests assert absence with.

**#195's zero-live gate vs. the `--fresh` escape hatch** (`420c5834`). Under `--fresh` the removal gate compared against `inventory.keys` — the CURRENT content keys — because there is no plan to map through. An in-place EDIT changes a file's content key, so the plan's key read as vanished and the new key read as added: the same path listed as both removed and added, and the run refused. `--fresh` is the route the refusal itself recommends for changed files, so the gate was blocking its own documented recovery, and it named a file that was sitting in the folder. Differential, both run here: `origin/main` returns `Ok(0)` for run → edit → `--fresh` (it carries no membership gate at all — `UnsupportedInventoryDelta` does not exist there), the pushed head returned the refusal above with `first.csv` named as both removed and added.

`UnsupportedInventoryDelta::between` now treats a plan entry whose `rel` or `path_id` is still present in the folder as a same-path replacement, never as a vanished group. That is fixed in the classifier rather than by calling `select_resume_plan_keys` under `--fresh`, because that helper bails on plan collisions and `--fresh` exists to get past exactly those — the gate must not be more fatal than the open it precedes. Genuine removals are unaffected: a removed file's `rel` and `path_id` are both gone, and all three removal tests still refuse.

**#195's zero-live gate vs. `xerj brain`'s refusal** (`8c1667d3`). Blocker 3 above was unreachable on the path it was written for, and this branch's own new CI phase caught it: run 31325755326 / job 93275739339 at head `5b0a6c7d` — *"expected a nonzero refusal naming the disagreement, got exit 1"*. Main's #195 verification lives INSIDE `run_index_report`, and on a wiped data dir it fires first: the run resumes, skips every already-done file, finds zero live documents and bails there, so `brain.rs`'s `?` propagated it and `needs_live_node_probe` / `journal_server_disagrees` / `journal_server_disagreement` never ran.

Reproduced by hand on this tree, then fixed and re-reproduced:

```
$ xerj autoindex <vault> --url http://localhost:9561 --brain zlvault   # wiped destination
error: autoindex verification failed: the resume journal records 2 record(s) from 2
completed file(s), but 0 documents are live across the 1 dataset index(es) …   (exit 1)
```

That is verbatim what `xerj brain` was printing. The condition is now the typed `ZeroLiveVerificationError { journal_records, files_done_journaled, dataset_indices }` — Display text unchanged, so `xerj autoindex` prints exactly what it printed before — and `brain.rs` downcasts it, converting it to `journal_server_disagreement()` only when the live-node probe agrees nothing is there. A write-blocked server, an unreachable server, the removal refusal, `--fresh`, and a probe that cannot answer all keep their own error. Peer precedent for typing a recoverable condition instead of folding it into a generic error: redb's `Database::do_repair` returns `DatabaseError::RepairAborted` as its own variant (`redb/src/db.rs:994`, Apache-2.0/MIT).

```
$ xerj brain <vault> --url http://localhost:9561 --data-dir <wiped> --no-open
error: the resume journal and server disagree: journal ~/.xerj/autoindex/765f64a0…
says <vault> is already indexed, but http://localhost:9561 has no confirmed live
node documents …                                                              (exit 1)
$ xerj brain <vault> … --fresh
✓ your second brain is ready — 2 files, 4 links                               (exit 0)
```

## Verification

Re-run on the current head after a second rebase onto `main` (`1ee31ecd`) — not carried over from an earlier measurement, which is the mistake that let the CI failure through the first time:

- `cargo fmt --all -- --check` — clean
- `cargo clippy -p xerj-autoindex -p xerj-server -p xerj-engine --all-targets -- -D warnings` — clean
- `cargo test -p xerj-autoindex` — **282 passed, 0 failed**, at default thread count *and* `--test-threads=1`
- `cargo test -p xerj-server` — **44 passed, 0 failed** across 7 targets
- `cargo test -p xerj-engine --lib embedding_identity` — 4 passed, 0 failed

New coverage: `fresh_recovers_journals_that_the_preflight_refuses_to_resume` (both fatal shapes; refused without `--fresh`, rebuilt and re-resumable with it), `fresh_over_a_readable_journal_reports_nothing_unreadable`, `refusal_prose_caps_its_listings_while_json_keeps_every_entry`, `refusal_prose_below_the_cap_has_no_more_tail`, `search_present_reports_a_missing_index_as_absence_and_nothing_else` (404 → absent, 403 → error, 200 → parsed), `deleted_planned_junk_is_swept_rather_than_refused`.

Repair coverage added on top: `fresh_absorbs_an_edited_file_instead_of_calling_it_a_removal`, `fresh_absorbs_an_addition_and_an_edit_after_a_reported_rerun` (the composite workflow the run's own stderr recommends), `an_ordinary_rerun_replaces_an_edited_file_without_stranding_records`, `a_zero_live_verification_failure_is_read_as_the_wiped_destination`, `zero_live_verification_prose_is_not_the_recovery_prose`.

Plus the end-to-end CI gate from #156: `usecase-smoke.sh` reruns autoindex over a folder a file was removed from and requires the refusal *by text*, then restores the file and requires the rerun through again — proving the gate is a stop, not a brick. The brain half wipes a data directory out from under a surviving resume journal and requires both the refusal and the `--fresh` recovery it names.

**`usecase-smoke.sh` was re-run locally on this head**, with release binaries built from it — all five phases, which is the thing the previous head's body implied but had not done:

```
════ 4. autoindex rerun over a mutated folder ════
  PASS: a rerun after a removal refuses (exit 1) and names the removed file
  PASS: restoring the file — recovery (1) in the refusal — lets the rerun through again

════ 5. brain resume safety (journal survives a wiped data dir) ════
  PASS: a journal/server disagreement refuses (exit 1) and names a recovery
  PASS: the --fresh rebuild the refusal names recovers the brain (exit 0)

USE-CASE SMOKE PASSED (second brain · MCP · autoindex)
```

## Not verified locally

- The two modified assertions in `xerj-engine` (`changed_onnx_assets_are_rejected`, `markerless_existing_index_fails_closed_for_onnx`) sit behind `#[cfg(feature = "onnx-experimental")]`, which this sandbox cannot build. Both were checked by inspection against the new message text; CI covers them if the feature is enabled there.
- The ES-YAML conformance suite was not run locally (it needs a booted server). The engine changes on this branch are error-string edits in `validate_embedding_identity`, which the suite does not exercise, but the required CI check is the authority.

## Known gaps, deliberately not fixed here

- **Added-but-skipped files still leave a live catalog entry until they are deleted** (#238's sweep handles the deletion, not the addition). This PR reports additions on stderr and records them as skipped; it does not index them. `--fresh` remains the way to absorb them.
- **`run_id` uniqueness** (`state.rs`) is closed only in-process and wraps at 65536 — carried over from #156, unchanged.
- **`--fresh` does not delete the records a same-path EDIT supersedes.** Document ids derive from the content key (`ids::doc_id`), so a `--fresh` rerun over an edited file publishes the new records and leaves the pre-edit ones live. That is `--fresh`'s behaviour on `main` too — it rebuilds the plan, it does not reconcile the destination — and an ordinary (non-fresh) rerun is the clean route for an edit: it keeps the planned key and runs a delete-before-replace transaction on it. Both halves are now asserted (`fresh_absorbs_an_edited_file_instead_of_calling_it_a_removal` vs `an_ordinary_rerun_replaces_an_edited_file_without_stranding_records`) rather than assumed. Closing it properly needs a durable "supersede these keys" intent in the journal, which is a design change, not a repair.
- **Nothing here reconciles removals.** The refusal is a fail-closed stop with documented recovery, not a sync implementation. Removing a file from an indexed folder still requires one of the three named routes.

## Before merge: the CLA check is a real blocker, not a flake

`verification/cla-signed` is `error` on this head because cla-bot resolves the commit authors and two commits are authored `buger <leonsbox@gmail.com>` — authorship preserved on purpose, since this is his work. `.contributors` currently holds only `xerj-org` and `dependabot[bot]`, so the bot's verdict is correct, not broken:

> Before we can, we need a signed Contributor License Agreement from: @buger

Per `CLA.md` the signature *is* a pull request from his own account adding his username, so nobody else can supply it: adding `buger` to `.contributors` from this branch would fake a signature and has deliberately not been done. Resolution is a maintainer decision — ask @buger to sign and comment `@cla-bot check`, or decide how his authorship is carried. Please do not merge past the check.

### DIFF
--- a/.github/scripts/usecase-smoke.sh
+++ b/.github/scripts/usecase-smoke.sh
@@ -170,6 +170,74 @@ else
   bad "expected >=3 hits for 'zanzibarite' across the discovered datasets, got ${HITS:-0}"
 fi
 
+# ── 4. autoindex: a removed file is refused, and the refusal is reversible ─
+phase "4. autoindex rerun over a mutated folder"
+# Nothing above reruns autoindex over a folder that CHANGED, which is where the
+# silent-skip bug lived: a rerun used to exit 0 while the removed file's
+# documents stayed live with no source behind them. The refusal has unit and
+# in-process HTTP tests; this is the end-to-end gate — and it also proves the
+# cheapest recovery the message names (put the file back) actually works.
+AX_STATE="$AX_WORK/state"
+ax_run() { "$XERJ_BIN" autoindex "$CORPUS" --url "$AX_URL" --state-dir "$AX_STATE" \
+             --prefix ax-eval --brain eval "$@" 2>&1; }
+
+mv "$CORPUS/logs/app.log" "$AX_WORK/app.log.away"
+REMOVED="$(ax_run)"; REMOVED_RC=$?
+if [ "$REMOVED_RC" != 0 ] \
+   && grep -q "no longer exist in the folder" <<<"$REMOVED" \
+   && grep -q "app.log" <<<"$REMOVED"; then
+  ok "a rerun after a removal refuses (exit $REMOVED_RC) and names the removed file"
+else
+  bad "expected a nonzero refusal naming logs/app.log, got exit $REMOVED_RC"
+  echo "$REMOVED" | tail -20
+fi
+
+mv "$AX_WORK/app.log.away" "$CORPUS/logs/app.log"
+RESTORED="$(ax_run)"; RESTORED_RC=$?
+if [ "$RESTORED_RC" = 0 ] || [ "$RESTORED_RC" = 3 ]; then
+  ok "restoring the file — recovery (1) in the refusal — lets the rerun through again"
+else
+  bad "expected the restored folder to rerun cleanly, got exit $RESTORED_RC"
+  echo "$RESTORED" | tail -20
+fi
+
+# ── 5. brain: a wiped data dir refuses, and the --fresh it names recovers ──
+phase "5. brain resume safety (journal survives a wiped data dir)"
+# The resume journal lives outside the server's data directory (default
+# ~/.xerj/autoindex/<hash>), so wiping the data dir leaves a journal claiming
+# the vault is fully indexed while the server holds none of it. `xerj brain`
+# used to resolve that ambiguity by resetting on the operator's behalf; it must
+# now refuse, and the in-place rebuild it names has to actually work.
+BRAIN_DATA="$BRAIN_ROOT/data"
+BRAIN_URL="http://localhost:$BRAIN_PORT"
+brain_run() { "$XERJ_BIN" brain "$BRAIN_ROOT/vault" --brain vault --url "$BRAIN_URL" \
+                --data-dir "$BRAIN_DATA" --no-open "$@" 2>&1; }
+
+[ -s "$BRAIN_DATA/server.pid" ] && kill "$(cat "$BRAIN_DATA/server.pid")" 2>/dev/null
+for _ in $(seq 1 60); do
+  curl -s -o /dev/null -m 5 "$BRAIN_URL/health/ready" || break
+  sleep 1
+done
+rm -rf "$BRAIN_DATA"
+
+REFUSAL="$(brain_run)"; REFUSAL_RC=$?
+if [ "$REFUSAL_RC" != 0 ] \
+   && grep -q "resume journal and server disagree" <<<"$REFUSAL" \
+   && grep -q -- "--fresh" <<<"$REFUSAL"; then
+  ok "a journal/server disagreement refuses (exit $REFUSAL_RC) and names a recovery"
+else
+  bad "expected a nonzero refusal naming the disagreement, got exit $REFUSAL_RC"
+  echo "$REFUSAL" | tail -20
+fi
+
+RECOVERY="$(brain_run --fresh)"; RECOVERY_RC=$?
+if [ "$RECOVERY_RC" = 0 ] || [ "$RECOVERY_RC" = 3 ]; then
+  ok "the --fresh rebuild the refusal names recovers the brain (exit $RECOVERY_RC)"
+else
+  bad "the --fresh recovery named by the refusal failed with exit $RECOVERY_RC"
+  echo "$RECOVERY" | tail -20
+fi
+
 # ── summary ───────────────────────────────────────────────────────────────
 echo
 if [ "$FAILED" = 0 ]; then
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -555,7 +555,23 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   URL, prefix, and brain identity plus recovery instructions, including the
   explicit `--fresh` rerun for a genuinely wiped data directory. Probe
   transport and malformed-response failures remain errors instead of being
-  reported as an empty destination.
+  reported as an empty destination. A nodes index that was deleted out from
+  under a surviving brain meta doc reads as absence and reaches that recovery
+  text, rather than surfacing as a raw HTTP 404.
+- **`--fresh` still recovers a state directory whose journal cannot be
+  parsed.** The rerun gate reads the durable plan before the run starts, and a
+  plan that is malformed, or recorded for a different root/URL/prefix, is fatal
+  to a resume — but not to `--fresh`, which deletes that journal unread. The
+  preflight is now no more fatal than the open it precedes: under `--fresh` the
+  unreadable plan is reported on stderr and rebuilt from the current folder.
+  Without `--fresh` the refusal is unchanged. Because the removal gate has no
+  comparison basis in that case, the warning says so: documents already
+  published for files that are now gone cannot be identified from an unreadable
+  plan and are not deleted.
+- **Refusal and skip listings are capped at ten entries plus an "and N more"
+  tail.** Unmounting a bind mount under an indexed root vanishes every content
+  group at once, so the uncapped listing was one rendered entry per journalled
+  file. `--json` still carries every entry.
 
 ## [1.0.0-rc.10] - 2026-08-03
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,56 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   `pdf_extraction_reuse` block so the behaviour is observable rather than
   inferred from timing. Original work by Leonid Bugaev (@buger).
 
+### Changed — autoindex refuses reruns that would strand documents
+
+- **Deleting an indexed file and rerunning `xerj autoindex` now fails instead
+  of exiting 0.** Nothing in the pipeline removes the documents, aliases,
+  graph edges or catalog entries that the deleted file published, so a rerun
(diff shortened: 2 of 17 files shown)


xerj-org:
Repair pass on head `5e7b0797`. Both review blockers are fixed, and the branch is rebased onto `1ee31ecd`.

**1. `--fresh` after an in-place edit was refused as a removal** (`c7a53029`). Under `--fresh` the gate compared against the current content keys, so an edited file's superseded key read as vanished and its new key as added — the same path listed as both, naming a file that was sitting in the folder. `UnsupportedInventoryDelta::between` now treats a plan entry whose `rel`/`path_id` still exists as a same-path replacement. Differential, both run locally: `origin/main` → `Ok(0)`, previous head → refusal. Regression tests: run → edit → `--fresh`, and the composite run → add+edit → rerun → `--fresh`.

**2. Blocker 3's fix was unreachable, and this PR's own CI phase 5 proved it** (`5e7b0797`). #195's zero-live verification fires inside `run_index_report`, so `brain.rs`'s `?` propagated it and the probe below never ran. It is now the typed `ZeroLiveVerificationError` (Display text unchanged), which `brain` downcasts into the recovery message only when the live-node probe agrees. Reproduced by hand before and after on a wiped destination — `xerj autoindex` still prints the verification text, `xerj brain` now prints the disagreement and the `--fresh` it names, which recovers at exit 0.

**CI is green** on this head — all 11 required checks, including `Use-case harnesses`, whose phase 5 now reads:

```
  PASS: a journal/server disagreement refuses (exit 1) and names a recovery
  PASS: the --fresh rebuild the refusal names recovers the brain (exit 0)
USE-CASE SMOKE PASSED (second brain · MCP · autoindex)
```

`verification/cla-signed` is still failing, correctly: two commits are authored `buger <leonsbox@gmail.com>` and `.contributors` does not list him. Per CLA.md only he can sign, so nothing on this branch can clear it — please resolve it rather than merge past it. Details in the PR body.

xerj-org:
@cla-bot check

PR #257, opened 2026-08-09
fix(resource): one documented core/memory policy — live worker knobs, a bounded phase A, and a governor that can see macOS

### LINKED ISSUE #240: autoindex/engine: no coherent core or memory policy — arbitrary worker caps, dead config knobs, macOS nice() misapplied, governor blind on Darwin
## User report

> "Macs are really slow while autoindexing a large code base."

This issue covers the **resource policy**: XERJ has no coherent story for how many cores it uses,
how much memory it may consume, or how it behaves on a laptop the user is also trying to work on.
The single biggest *slowness* cause is tracked separately as the `split_sections` quadratic; this
issue is everything else, and it is what makes the tool a bad citizen on a personal machine.

## 1. The autoindex worker caps are arbitrary, and raising them alone fixes nothing

`engine/crates/xerj-autoindex/src/cli.rs:124-131`

```rust
let mut workers = std::thread::available_parallelism().map(|n| n.get()).unwrap_or(8).min(8);
let mut pdf_workers = std::thread::available_parallelism().map(|n| n.get()).unwrap_or(4).min(4);
```

`git log -L 124,132:crates/xerj-autoindex/src/cli.rs` shows `workers` arrives whole in `0f3ef60d`
("feat(autoindex): zero-config folder discovery") and its commit body never mentions the cap.
`pdf_workers` arrives in `19f0dd40`, whose body **does** state a reason ("Cap concurrency at four",
alongside a 1536 MiB `RLIMIT_AS` per isolated PDF process) — so that one is real and should stay.

The thing people assume `workers` guards — file descriptors — is guarded elsewhere and differently:
`.github/scripts/autoindex-fd-smoke.sh:4-16` states the budget is per-index WAL fds, and the fix is
`esclient.rs:88-98` pinning every created index to `index.xerj_ingest_shards = 1`, plus
`raise_nofile_limit()` at `xerj-server/src/main.rs:1424`. A client worker costs ~1 tempfile + ~1
socket. So the cap **can** be raised safely.

**But raising it buys almost nothing.** The client never becomes CPU-bound. Measured sweep, 2,013
files / 52 MB of source, sampling `/proc` every 200 ms:

```
workers=1    phase B 35.09s   client 0.15 cores   server 0.62 cores
workers=8    phase B 13.69s   client 0.67 cores   server 2.07 cores
workers=32   phase B 11.67s   client 0.69 cores   server 2.25 cores
```

Clean `/usr/bin/time` run at `workers=8`: wall 10.07 s, user 7.93 s, sys 1.28 s = **0.91 cores
average over the whole process**. 8 → 32 workers moved phase B ~15% (inside this box's noise) while
raising client peak RSS from 353 MB to 483 MB. Shipping only a cap change would be a visible no-op
to the users who complained.

## 2. `--workers` does not control phase A at all

`lib.rs:595` (`files.par_iter().map(scan_file)`) and `content.rs:28` (`full_digest`) both run on
rayon's **default global pool**. `grep -rn 'ThreadPoolBuilder|num_threads|build_global'
crates/xerj-autoindex/src` returns nothing — the pool is never sized. `cfg.workers` only bounds
phase B's `std::thread::scope`.

Measured: with `--workers 1`, phase A still used **7.43 client cores**.

The help text at `cli.rs:68` reads `--workers <N>  extract workers (default min(cores,8))`, implying
it governs extraction. A user who sets `--workers 2` to get their Mac back still gets a
full-machine digest+sniff storm. This is also why lowering `--workers` does not reduce the
quadratic's memory: the knob users reach for is disconnected from the phase that hurts them.

## 3. Server-side thread demand, not the client, fills the machine

Per-run demand at HEAD:

| pool | sizing | file:line |
|---|---|---|
| tokio runtime | `(cores * 8).clamp(64, 512)` | `xerj-server/src/main.rs:1430-1437` |
| ingest_pool | N cores | `xerj-engine/src/lib.rs:104-108` |
| background_pool | N cores | `xerj-engine/src/lib.rs:148-152` |
| merge_pool | `max(2, N/8)` | `xerj-engine/src/lib.rs:188-192` |
| flush_finalize_gate | `max(2, N/8)` | `xerj-engine/src/lib.rs:236` |
| rayon global (search fan-out) | N, never explicitly built | — |
| onnx_intra_threads | full core count | `xerj-common/src/config.rs:927` |

Note the **clamp floor of 64**: a 4-core MacBook Air still gets 64 tokio worker threads.

"USE ALL THE CORES" already happens — on the server side, unbudgeted, while the client idles
waiting on it. A policy that only touches `--workers` is aimed at the wrong process.

## 4. Three documented `EngineConfig` knobs are dead

`flush_workers`, `search_workers`, `merge_workers` are declared, defaulted and validated at
`xerj-common/src/config.rs:1117-1133` (+ a validation rule at `:1152-1155`) — and read by
**nothing**. `grep -rn 'flush_workers|search_workers|merge_workers' --include=*.rs --include=*.toml
crates/` returns hits in `config.rs` only. Real merge sizing is hardcoded at `xerj-engine/src/lib.rs:188-190`.

A user following the config docs to make XERJ gentler on their laptop (`engine.search_workers = 2`)
changes nothing and gets **no warning**. This is the accepted-and-ignored bug class from #204.

## 5. macOS: the nice ladder is Linux semantics applied under `#[cfg(unix)]`

`xerj-engine/src/lib.rs:126` `libc::nice(5)`, `:157` `nice(10)`, `:197` `nice(15)` — each inside a
rayon `start_handler` under `#[cfg(unix)]`, which **includes macOS**.

On Linux the nice value is a per-task attribute, so this deprioritizes only that thread. On Darwin
it is a **per-process** attribute — which is exactly why Darwin adds a separate thread-scoped knob:
`PRIO_DARWIN_THREAD = 3` / `PRIO_DARWIN_BG = 0x1000` (present in the pinned libc). `nice()` is also
incremental, and these handlers run once per pool thread.

If the documented semantics hold, then on macOS the ladder does not separate the three pools at all
and instead ratchets the **whole server process** toward nice 19 as pool threads start —
deprioritizing search and bulk along with merges. The carefully measured Linux tuning described at
`lib.rs:109-121` would be actively wrong on the platform users are complaining about.

**This is UNVERIFIED on hardware** — reasoned from documented POSIX/Darwin semantics plus the libc
surface. Nothing in CI could catch it: the only macOS job (`ci.yml:307`) is the FD smoke test, which
does not measure scheduling.

## 6. Zero QoS awareness, and `available_parallelism()` counts E-cores

`grep -rn 'qos|QOS|setpriority|SCHED_|thread_policy' --include=*.rs crates/` returns **nothing**.

Rust std's Apple branch is plain `sysconf(_SC_NPROCESSORS_ONLN)` — the cgroup + `sched_getaffinity`
refinement is gated to Linux/Android. An M3 Pro 12-core (6P+6E) reports 12 with no distinction.
Thread placement on Darwin is decided by QoS class; with none set, indexing work competes with the
user's foreground apps and spreads onto E-cores at full priority.

The mechanism is already available: the pinned libc exposes `pthread_set_qos_class_self_np`, and
rayon's `start_handler` hook is already in use at `xerj-engine/src/lib.rs:122`. This is a few lines
inside handlers that already exist.

## 7. The server's memory awareness is real but blind on macOS

`xerj-engine/src/governor.rs` is a 1,206-line process-wide governor (memtable budget, RSS admission
watermark, search/bulk semaphores, cgroup-v2/v1 walker) that boots on every run. Three defects:

* `governor.rs:586-603` — `current_rss_bytes()` is `#[cfg(not(target_os = "linux"))] { 0 }`.
  **On macOS the RSS watermark can never trip.**
* `governor.rs:617-633` — `system_total_bytes()` reads `/proc/meminfo` on Linux and otherwise
  returns a hardcoded `8 * 1024 * 1024 * 1024`. **On a 16/32/64/128 GB Mac every derived budget is
  sized off a fictional 8 GiB.**
* It reads **MemTotal, never MemAvailable**. (`xerj-api/src/es_compat.rs:32175` does parse
  MemAvailable — but only to render `_cat/nodes` ram.percent; it feeds no budget.)

Derived budgets: `auto_memtable_budget` = `clamp(limit/4, ≥2 GiB, ≤limit/2)` (`governor.rs:391-395`);
`memory_watermark_percent` default 95 (`config.rs:1051`); segment hydration = limit/5 (`:417`).

The server is also the bigger consumer: VmHWM went 12.25 GB → 25.26 GB indexing the 136 MB lucene
corpus (+13 GB, ~95× corpus size).

Autoindex, by contrast, has **no** memory awareness at all — fixed `bulk_mb = 8`, `--workers`
unbounded (`cli.rs:154-159` applies only `.max(1)`, unlike `--pdf-workers` which is clamped to 1..=4).

## 8. Client backpressure exists but is blind

`esclient.rs:186-210` retries a 429 six times with 250 ms..8 s backoff, then fails. Nothing in the
crate adjusts `cfg.workers`, `bulk_mb`, or the rayon pool in response. The transport half of
"backpressure, never OOM, never drop" is correct; the missing piece is the **feedback edge** — a 429
should shrink concurrency for the rest of the run, not just sleep and re-offer the same load.

## What peers do (design precedent, nothing copied)

* **tantivy** (MIT): `MAX_NUM_THREAD = 8` for the *writer* (`indexer/index_writer.rs:35`), and
  `index.rs:656-660` further **reduces thread count when `memory_budget / num_threads` falls below
  `MEMORY_BUDGET_NUM_BYTES_MIN`** — threads derived from the budget. Its *search* executor gets full
  `available_parallelism()` (`index.rs:353`).
* **meilisearch** (MIT core): indexing pool = `num_cpus::get() / 2` (`option.rs:885`) with the stated
  reason "the indexer avoids using more than half of a machine's total processing units. This ensures
  Meilisearch is always ready to perform searches, even while you are updating an index"
  (`option.rs:818-821`).
* **quickwit** (Apache-2.0): `num_cpus()` honours `QW_NUM_CPUS`/`KUBERNETES_LIMITS_CPU` before the OS
  (`quickwit-common/src/cpus.rs:29-62`); explicit headroom by machine size — `num_cpus - 2` for ≥7
  cores, `-1` for 4..6 (`runtimes.rs:89-110`).

All three reconcile "use the machine" with "stay usable" the same way: **full parallelism for the
latency-critical path, a named reserve for the background path.** That shape is what to adopt — not
the specific constants.

Honest negative from retrieval: **no peer search engine sets macOS QoS classes** (grepped every
corpus for `pthread_set_qos_class_self_np` / `QOS_CLASS_` / `thread_policy_set`). There is no
precedent to copy for the Apple Silicon part.

## Measurement provenance

AMD Ryzen AI MAX+ 395, 16 cores / 32 threads, 119 GB RAM, Linux 7.0.0-27-generic, xerj v1.0.0-rc.13.
The box is shared with other agents' servers and cargo builds, so **wall times carry real noise**
(an unrelated repeat of `workers=8` took 57 s). The client-cores figure is the robust quantity and it
is below 1.0 in every run. **No macOS measurement was taken.**

xerj-org:
Fixes #240

XERJ had no single answer to "how much of this machine may I take?", and the
scattered answers it did have disagreed with each other — and, on Darwin, with
the OS. This adds one documented policy and wires the existing knobs to it.

## Reproduced first, at HEAD

| # | Defect | How it was reproduced |
|---|---|---|
| §2 | `--workers` does not control phase A | A unit test that reads `rayon::current_num_threads()` from inside the digest closure reported **32** on this 32-core box regardless of the requested worker count — phase A ran on rayon's default global pool (`content.rs:28`, `lib.rs:595`; the crate never built a pool). |
| §4 | Three dead `EngineConfig` knobs | `engine.search_workers = 2` → the global pool search fans out on still had **32** threads. The new integration test asserts `2`; with the wiring commented out it fails with `left: 32, right: 2`. |
| §5/§7 | macOS nice + blind governor | Source-level: `libc::nice()` under `#[cfg(unix)]` (per-process on Darwin), `current_rss_bytes()` hardcoded `0` and `system_total_bytes()` hardcoded `8 GiB` off Linux. **Not reproducible here — no Mac.** |
| §8 | 429 backpressure has no feedback edge | With the feedback edge stubbed back out, the two new HTTP-level tests fail where they should: a transport 429 leaves the window at `left: 8, right: 4`, a per-item 429 at `left: 4, right: 2` — i.e. at HEAD the client slept and re-offered the identical load. |

## What changed

* **`xerj-common/src/resource.rs`** — the policy, with its reasoning written
  down: cores (`XERJ_NUM_CPUS` → cgroup CPU quota → OS → warned default),
  total/available RAM, process RSS, the memory safe zone, threads per workload,
  and the platform-correct priority ladder. Every rule has a pure function and a
  test.
* **`xerj-engine/src/pools.rs`** — config → pool widths. `flush_workers` → the
  flush finalize gate, `merge_workers` → the merge pool, `search_workers` →
  rayon's global pool (built explicitly, since that is what search fan-out uses).
  **Defaults reproduce the widths the pools already used**, so the measured
  ingest/read tuning is unchanged; what changes is that the knobs work.
* **Loud instead of silent** — out-of-range worker counts are refused at startup
  (`engine.merge_workers = 0` used to pass validation); `--workers 0` /
  `--workers 100000` / `--bulk-mb 512` are refused with a range instead of being
  rewritten; a configured RSS watermark that this platform cannot enforce says
  so; an unusable `XERJ_NUM_CPUS` says so on stderr *and* in the log, because
  the CLI has no tracing subscriber.
* **macOS** — Linux keeps `nice(5/10/15)`. Darwin applies
  `setpriority(PRIO_DARWIN_THREAD, 0, PRIO_DARWIN_BG)` to the **merge pool
  only** (the one pool nobody waits on) and leaves ingest/flush at normal
  priority: Darwin's background tier throttles I/O too, and there is no partial
  per-thread nice. Nothing process-wide is touched. Real probes replace the
  fictions: `sysctlbyname("hw.memsize")` and `proc_pidinfo(PROC_PIDTASKINFO)`.
* **autoindex** — one plan per run: phase A gets every core by default *and is
  confined to a real pool*, phase B's **default** is capped by what the memory
  safe zone can pay for at the chosen `--bulk-mb`, the PDF **default** by what
  it can pay for at 1536 MiB each. The run prints what it decided; the JSON
  summary records it. Two rules keep that arithmetic from doing damage of its
  own (`resources.rs`):
  * **A cap the machine cannot substantiate is not applied.** The safe zone is
    an `Option`, and `None` — Windows, FreeBSD, anywhere `total_memory_bytes()`
    has no probe — means *no memory-derived cap*, not *a 1 GiB machine*. The
    first revision of this branch used a 1 GiB fallback there, which bought
    `1 GiB / 1536 MiB = 1` PDF worker and so cut PDF extraction from
    `min(cores, 4)` to 1 on every Windows run — charged against an `RLIMIT_AS`
    that is `#[cfg(unix)]`-only and therefore never enforced on that platform
    (`extract/pdf.rs` `containment_description()`: "no OS memory cap on this
    platform"). `plan_for` now takes `Option<u64>` so unknown cannot masquerade
    as small, and `an_unprobeable_machine_is_not_treated_as_a_tiny_one` fails
    (`pdf_workers: 4 != 1`) if it ever does again.
  * **The cap applies to the default, not to the operator.** An explicit
    `--workers 32` used to become 3 on a machine whose safe zone had dipped to
    its 256 MiB floor — the same accepted-and-quietly-replaced shape as #204,
    and easy to trigger, since the safe zone subtracts `max(1 GiB, total/8)`
    from a live `MemAvailable` reading. A number the operator typed is now
    honoured and argued with in a note; only hard range errors (`--workers 0`,
    `--workers 100000`) are still refused outright, which is the same split
    tantivy draws (`index_writer.rs:2831` rejects an out-of-range writer option
    rather than clamping it).
* **Backpressure now has a feedback edge** (`esclient::BulkAdmission`, §8) —
  every bulk passes through one window `--workers` wide, shared by all workers.
  A 429 halves it on the spot (floor 1); further 429s inside a 1 s cooldown are
  the *same* congestion event seen by the other in-flight workers and do not
  compound, so one momentary stall cannot collapse the run to a single worker.
  Recovery is a probe, not a reset: one worker back per 8 consecutive clean
  bulks, never above what the operator asked for. A permit is held across the
  retries, which is what makes a shrunken window reduce offered load rather
  than rename it. HTTP 200 with items rejected `status: 429` counts as
  congestion; 5xx does not. Off by default for probes and one-shot clients.

## Measured (32 cores / 119 GB; 273 files, 9 MB; `--dry-run` = phase A only)

```
--workers 1     6.04 s wall    5.95 s cpu   → 0.99 client cores
--workers 4     1.31 s wall    2.27 s cpu   → 1.73 client cores
default (32)    5.07 s wall   43.86 s cpu   → 8.65 client cores
```

Phase A now scales with the flag; before this change every one of those runs
took the whole machine. End-to-end, default workers: 286 docs, exit 0, peak
client RSS 528 MB vs 238 MB at `--workers 2` — the cost of more in-flight bulk
buffers, which is what the safe zone bounds on a small machine. Wall times on
this box are noisy (it is shared with other builds); the cores figure is the
robust quantity.

### Backpressure, end to end

Release build of this branch against a live server, `xerj autoindex docs/`:
165 files → 7 datasets, **1175 records live, exit 0**, and the run summary
carries `bulk_concurrency_final: 32`, `bulk_congestion_events: 0` — an
uncongested run keeps its full window, which is the behaviour that must not
change. The shrink path is covered by five deterministic tests rather than by
trying to overload a real server.

## Gates

* ES-YAML conformance against a release build of this branch:
  **1366 passed · 0 failed · 3 skipped**.
* `cargo fmt --all -- --check`; `clippy -D warnings` and the test suites for
  `xerj-common`, `xerj-autoindex`, `xerj-engine`, `xerj-server` — clean.
* `cargo check --target aarch64-apple-darwin -p xerj-common` — the macOS paths
  compile.
* Pre-existing, unrelated: `xerj-engine --test painless_script_limits` overflows
  its stack in debug builds; it does so identically on unmodified `origin/main`
  (verified by stashing this branch's changes).

## Honest limits

* The macOS behaviour is reasoned from documented POSIX/Darwin semantics plus
  the pinned libc surface and cross-compiles clean. It has **not** been measured
  on Apple hardware, and nothing in CI measures scheduling.
* Windows (and any other platform without a RAM probe) still has no memory
  probe — `GlobalMemoryStatusEx` is the follow-up. Two *different* things happen
  there, and an earlier version of this bullet wrongly claimed one number
  covered both:
  * the **server governor** keeps its pre-existing 8 GiB assumption
    (`ASSUMED_TOTAL_BYTES`, `xerj-engine/src/governor.rs:623`), unchanged so
    that no derived budget is re-tuned on a platform nobody has measured — the
    only difference is that the assumption is now announced;
  * the **autoindex client plan** applies no memory-derived cap at all
    (`memory_safe_zone_bytes()` is `None`, `xerj-common/src/resource.rs`), so
    scan/index/PDF counts stay core-derived. It does *not* inherit the 8 GiB
    figure, and it must not substitute a small one either.
* One part of #240 is deliberately out of scope: macOS QoS classes
  (`pthread_set_qos_class_self_np` is not in the pinned libc, and no peer engine
  sets them).
* The 429 feedback edge is AIMD, not a permanent shrink: a run that meets
  sustained pressure stays small, but a run that met one stall climbs back to
  its ceiling. That is a deliberate reading of "for the rest of the run" — a
  permanent collapse would punish a run for a single momentary spike — and it
  is measurable after the fact via `bulk_congestion_events`.
* The shrink path has not been exercised against a *real* overloaded server,
  only against mock endpoints that answer 429 on demand.
* Behaviour change worth knowing: a config that already sets
  `engine.{flush,merge,search}_workers` now takes effect. On macOS the governor
  budgets are derived from real RAM instead of 8 GiB, so memtable budget and
  watermark grow to match the machine.

## Reference-coding

Retrieved before writing (cited, never copied): quickwit
`quickwit-common/src/cpus.rs:23-63` and `runtimes.rs:89-110`;
`quickwit-config/src/node_config/mod.rs` `default_merge_concurrency()`;
meilisearch `option.rs:818-821,885,1103-1147`; tantivy
`index_writer.rs:32-36,285-291`. The `xerj-storage` corpus had nothing relevant
— sled/redb/fjall size neither pools nor budgets from the machine — so it is
reported as a miss rather than forced in.

For §8 the retrieval result was an honest **miss**, and is documented as one at
the type: no peer does this. quickwit's rest client sleeps 500 ms and retries
the same batch (`quickwit-rest-client/src/rest_client.rs:371-380`),
meilisearch's REST embedder classifies 429 as retry-later
(`crates/milli/src/vector/embedder/rest.rs:550`), its S3 path treats it as a
transient backoff (`.../enterprise_edition/s3.rs:447`) — all three are the
transport half XERJ already had. The AIMD shape is adapted from congestion
control, not from a search engine.


xerj-org:
Review repairs pushed as `295cd50a`. All three must-fix items were real; here is what changed and how each was verified.

**1. Windows/FreeBSD PDF throttling (confirmed, fixed).** `memory_safe_zone_bytes()` fell back to a flat 1 GiB where `total_memory_bytes()` has no probe, and `plan_for` charges 1536 MiB per PDF worker against it, so `(1 GiB / 1536 MiB).max(1) = 1` — `min(cores, 4) = 4` became 1, charged against an `RLIMIT_AS` that is `#[cfg(unix)]`-only and never applied there. Fixed at the type: `memory_safe_zone_bytes() -> Option<u64>`, `plan_for(cores, Option<u64>, ...)`, and `None` disables every memory-derived cap instead of standing in for a tiny machine. The 1 GiB constant is gone, so there is nothing left to fall back to.

**2. PR body claim (confirmed, corrected).** The 8 GiB `ASSUMED_TOTAL_BYTES` is the *server governor's* fallback only; the autoindex plan never inherited it. The "Honest limits" bullet now states the two halves separately.

**3. Explicit `--workers` silently reduced (confirmed, fixed).** The memory cap now bounds the **default** only. A number the operator typed is honoured and the disagreement is printed (`...allows 3 index workers at --bulk-mb 24; honouring --workers 32 anyway`). Hard range errors are still refused at the CLI — the same split tantivy draws between an invalid writer option it rejects (`src/indexer/index_writer.rs:2831`) and a tuning heuristic it does not impose.

**Regression tests, verified by reverting only the source hunk** (`if requested_workers.is_some()` → `if false`, plus a re-introduced 1 GiB fallback):

```
an_unprobeable_machine_is_not_treated_as_a_tiny_one   left: 1  right: 4    (pdf_workers)
an_explicit_count_survives_a_small_safe_zone_...      left: 3  right: 32   (index_workers)
```

Both reproduce the reviewer's measured numbers exactly, and both pass with the fix in place. The same test also re-asserts that the *unflagged* default on that 256 MiB machine still bows to the machine (3 / 1) — the cap was weakened for the operator, not removed.

Also on the way through: the `--json` summary reports `memory_safe_zone_mb: null` rather than a stand-in 1024, `resource::describe()` prints `safe_zone_mib=unknown`, and `--help` plus the two zero-config recipes say the reduction applies to the default.

Local gate on the pushed tree: `cargo fmt --all -- --check` clean; `clippy -p xerj-common -p xerj-autoindex --all-targets -D warnings` clean; `cargo check -p xerj-server` clean (it consumes `resources::plan`); `cargo test -p xerj-common` 65+3 passed; `cargo test -p xerj-autoindex` 270+1 passed, 0 failed.

Reference-coding: meilisearch keeps the same `Option` shape for the same reason — `total_memory_bytes() -> Option<u64>` is `None` on an unsupported system (`crates/meilisearch/src/option.rs:1135`) and the derived budget stays an `Option` where `None` means unbounded, not minimal (`crates/milli/src/update/index_documents/helpers/grenad_helpers.rs:122`). Adapted, not copied.

Issue #241, opened 2026-08-09
autoindex: no progress, percent or ETA — 47-64% of every run is silent, and agents cannot tell running from hung


xerj-org:
## User report

> "No progress bar in their AI agents" — users drive XERJ through Claude Code and similar agents via
> our prompt-based install (`xerj autoindex .`). The agent sees long silences and cannot tell running
> from hung, and neither can the human watching the agent.

**Reproduced and fully explained.** This needs no algorithmic work — every counter already exists.

## 1. 47–64% of every run is one unbroken silence at the end

Corpus: 2,013 walked files / 55.5 MB of tantivy+quickwit source. stdout **and** stderr both pipes
(asserted not-a-tty in the harness). Three runs:

```
run 1 (--brain gaptest)  exit=3 wall=32.1s lines=168   15.06s of silence [16.97s -> 32.04s]  = 47%
run 2 (--fresh, warm)    exit=3 wall=14.1s lines=168    9.04s of silence [ 5.08s -> 14.11s]  = 64%
run 3 (--no-graph)       exit=3 wall=57.6s lines=166   33.17s of silence [24.36s -> 57.53s]  = 58%
```

The silence is not a slow trickle — it is **one block at the end**, precisely when the user expects
completion, so it reads as a hang.

Run 3 (`--no-graph`) had the **longest** tail, so the graph is not the cause. The cost is the
`finalize: refresh, verify, correlate, catalog` block at `lib.rs:1988`, which emits nothing between
`lib.rs:1918` (last phase-B line) and `lib.rs:2352` (final summary).

## 2. Nothing in the workspace is TTY-aware — proven empirically

`grep -rn "is_terminal|IsTerminal" crates/` over the whole workspace returns **zero** hits. No
indicatif, no is-terminal, in any `Cargo.toml`.

Empirically: the identical command via python `pty.spawn` vs `subprocess.PIPE` emitted exactly 168
lines each; after normalizing the brain name and timings, the only diffs were which file randomly
finished 200th (worker race). Control-character census of the PTY capture:

```
ESC (0x1b) count: 0
CR  (0x0d) count: 168
CR not followed by LF (in-place redraw): 0
```

All 168 CRs are CRLF from the pty line discipline. There is **one** output path for humans, agents,
and CI — no TTY mode to degrade from and no non-TTY mode to degrade to. Both halves of a dual-mode
contract must be built from scratch.

## 3. Progress fires on file COUNT, so silence has no upper bound

[...]

Issue #202, opened 2026-08-07
Unparseable schema.json silently opens the index with an empty dynamic mapping


xerj-org:
If `schema.json` cannot be parsed at open time, XERJ does not refuse to open the index. It falls back to an empty dynamic mapping.

`xerj-engine/src/index.rs:5578`:

```rust
load_schema(...).unwrap_or_else(|_| ManagedSchema::dynamic())
```

Same shape one line up for settings, `xerj-engine/src/index.rs:5568`:

```rust
load_settings(...).unwrap_or(Value::Null)
```

So a truncated write, a partial disk, a bad downgrade, or any corruption of that one file and the index opens green with every explicit mapping gone. Field types then get re-inferred from whatever documents arrive next. A `keyword` that was deliberately not analyzed becomes whatever the first value looks like. Nothing in the response tells the user their mapping is gone.

This is worse than failing to start. A failed start is obvious and recoverable from backup. This one keeps serving, silently changes the meaning of the data, and the damage compounds with every write until someone notices their queries stopped matching.

The storage layer already does this correctly, which is what makes the contrast clear. Segment headers carry `format_version` and `schema_version` and readers fail loudly with `UnsupportedFormat` on a mismatch (`xerj-storage/src/segment.rs:243-246`). The sidecar path should behave the same way.

Fix: refuse to open the index when `schema.json` exists but does not parse. Falling back to dynamic is only correct when the file is genuinely absent, which is a different condition from unparseable.

Related, found in the same pass: whether an index actually survives a cross-version binary swap has never been tested. The version check at `segment.rs:243-246` is an exact `!=`, and no second format version has ever existed to exercise it. Worth a golden data-dir fixture before GA, since "replace the binary and restart" is the upgrade story.

Found during a source review against the user-feedback corpus, where upgrade traps and mapping surprises are recurring complaints.

Issue #239, opened 2026-08-09
autoindex: split_sections is quadratic in time and memory — one 16 MiB file drives peak RSS to 64 GiB


xerj-org:
## Summary

`split_sections` has an accidental quadratic in both **time and memory**. Peak RSS grows as
`(paragraph bytes)² / (2 × SECTION_CHARS)`. A single 16 MiB text file with no blank line drives
the autoindex CLI to **64 GiB RSS**.

This is the strongest available explanation for the user report *"Macs are really slow while
autoindexing a large code base"*: the arithmetic is platform-independent, and a 16–36 GB Mac
cannot hold the working set, so it goes to memory compression and swap and the run appears to hang.

## Root cause

`engine/crates/xerj-autoindex/src/extract/mod.rs:293-302`

```rust
while cur.len() > 2 * SECTION_CHARS {
    let cut = /* char boundary at SECTION_CHARS */;
    let rest = cur.split_off(cut);
    out.push(std::mem::replace(&mut cur, rest));
}
```

Two separate amplifiers in three lines:

1. **Time.** `String::split_off(at)` allocates a new `String` and copies the entire tail into it.
   That copy runs once per ~2 KB section, so a paragraph of length `L` memcpys `L²/(2·SECTION_CHARS)`
   bytes.
2. **Memory.** `split_off` does **not** shrink `self`'s capacity, and `mem::replace` then pushes
   that old buffer into `out`. Every section in `out` therefore retains the capacity of the whole
   remaining paragraph — N strings of ~2 KB *length* and up to 16 MB *capacity*.

`SECTION_CHARS = 2 << 10` at `extract/mod.rs:86`.

## Evidence

Isolated proof (`split_sections` copied verbatim vs a slice-based rewrite producing **identical**
section counts and identical total bytes):

```
current 2MB: sections=1024 sum_len=2097172 sum_capacity=1025.0MB  elapsed=186.5ms  peakRSS=1032.1MB
fixed   2MB: sections=1024 sum_len=2097172 sum_capacity=2.0MB     elapsed=0.97ms   peakRSS=6.4MB
current 4MB: sections=2048 sum_len=4194344 sum_capacity=4098.1MB  elapsed=709.3ms  peakRSS=4110.2MB
fixed   4MB: sections=2048 sum_len=4194344 sum_capacity=4.0MB     elapsed=1.65ms   peakRSS=10.4MB
[...]

PR #258, opened 2026-08-09
fix(engine,api,wasm): issue #204 sweep — fail closed, or fail loudly, on config xerj cannot honour

### LINKED ISSUE #204: Tracking: accepted-and-ignored is the dominant bug class, six instances found in one review

xerj-org:
Part of #204.

> **This pull request was written by an AI agent (Claude).** Code, tests,
> CHANGELOG and this body were all agent-authored, across four review rounds
> plus a landing pass. Every number in it comes from a command that was run;
> nothing is estimated, and the split between what was *measured* and what is
> *assumed* is stated below and again per round.
>
> It deliberately carries no closing keyword. It is `Part of #204`, not
> `Closes #204` — the sweep's remaining gaps are listed under
> [*Known gaps*](#known-gaps-stated-rather-than-closed) and the issue should
> outlive this merge.

## Verified vs assumed — landing summary

The per-round detail is further down, and each round names the commit it was
measured at. This is the consolidated view. The head being merged is
`a5d3f06` (= `d68b5e0` + the `de1e5e0`/#356 merge); where a number was last
measured on an earlier head of this same branch, that head is named rather
than glossed as "current".

**Verified — commands that were run, output that was read.**

Re-run on `a5d3f06` itself:

- `cargo fmt --all --check` — exit 0.
- `cargo clippy -p xerj-engine -p xerj-api -p xerj-fts --all-targets -- -D warnings` — exit 0, zero warnings.
- `cargo test -p xerj-api -p xerj-engine -p xerj-fts --profile ci-test --no-fail-fast` — **68 binaries `ok`, zero failures**.

Measured on `b4306a1` (the previous merge of this branch with `origin/main`;
the only change since is the `CHANGELOG.md`-only #356 merge plus `2483418`):

- Dev profile, overflow checks ON — which `--profile ci-test` does **not**
  have, since it inherits release: `fail_closed_config_http` **36 passed**,
  `xerj-engine --test integration` **120 passed · 1 ignored**, `xerj-fts
  --lib` **76 passed** — all 0 failed.
- `cargo test -p xerj-engine --test es_compat_tests --profile ci-test` —
  **65 passed · 0 failed**.
- **ES-compat YAML conformance**, own release binary from this tree, private
  port, throwaway data dir — **1366 passed · 0 failed · 3 skipped**,
  `RUNNER_EXIT=0`.

CI re-runs fmt, clippy, the full test matrix, the release (fat-LTO) build and
the ES-compat YAML gate on the merge commit, so those are checked
independently of the numbers above rather than taken on trust.
- **Fail-before for every behavioural fix**: each source hunk was reverted on
  its own and the suite re-run, so the test that covers it was observed to
  fail. The three fail-before tables list the reverted hunk, the test, and the
  assertion output.
- The two blocker repros (the restart-resurrects-a-refused-pipeline sequence,
  and the 40-concurrent-`PUT` lost-update) were reproduced **at a live node
  over HTTP**, then pinned as tests that fail before and pass after.

**Assumed — reasoned about, not demonstrated.**

- That `#372`'s `partial_multi_batch_failure_is_reported_without_losing_either_batch`
  flake stays confined to full-parallelism runs. It is briefed as pre-existing,
  it passes in isolation, and `git diff origin/main..HEAD` touches its module
  zero times — but no bisect was run.
- That no ES client sends a non-string `description` on a pipeline body. The
  translation used to coerce it to `""` and now carries it through, so such a
  body would become a compile error rather than a silently blanked
  description. No YAML conformance case sends one.
- That carrying ES's `version` / `_meta` / `deprecated` into the stored config
  is harmless because `PipelineConfig` is not `deny_unknown_fields` — read, not
  fuzzed.
- That a corrupt refusal-marker sidecar is rare enough that failing closed for
  every pipeline is the right trade. It is loud and the reason names the file,
  but it is a cluster-wide ingest stop until repaired.
- The file-ordering fix in `put_pipeline` (`2483418`) is argued **from the code
  path, not demonstrated**: the failure it protects against is a filesystem
  error and there is no injection point for one at this layer.

**Not claimed.** The breaking analysis-settings gate is **not** covered by the
ES-YAML suite — the suite reads 1366/0/3 either way, so CI being green is not
evidence that any given index still creates. That is why the breaking shapes
are enumerated by measurement in the CHANGELOG rather than inferred from a
green gate.

---

Issue #204 asks for a specific sweep: grep for `unwrap_or_else`,
`unwrap_or_default` and `let _ =` on paths that handle user-supplied
configuration or user data, and ask of each one whether the fallback is
**equivalent** or **weaker**. 682 raw hits across `engine/crates`; 41 sit on
user configuration or user data; the weaker ones are fixed here.

Every site shares one shape: xerj accepted an instruction, answered success,
and then did something smaller than what was asked — with the only evidence in
a log line, or nowhere at all.

The two purest instances the issue names (`index.rs:5568`/`:5578`) are
deliberately untouched: they are #202, fixed separately.

## What changed

**Ingest pipelines: `200 acknowledged` now means "we will run this".**

- `PUT /_ingest/pipeline/{id}` used to store the raw body *first* and only
  warn-log a compile failure, so a pipeline that never ran was acknowledged and
  echoed back by `GET`. It now compiles first and has three outcomes: compiled
  → store + acknowledge; config we refuse → 400; a stage or option this build
  cannot honour → accept as ES does, but record *why* it has no compiled form
  so the ingest paths listed below refuse loudly rather than indexing an
  untransformed document.
- **Every ES `rename` processor had always failed to compile.** The handler
  translated `{field, target_field}` into `{from, to}` while the `field_rename`
  plugin requires `{mappings: {old: new}}`. Invisible under warn-and-store:
  200, echoed by `GET`, renamed nothing. Fixed, with a regression test that
  asserts the *document*, not the status.
- Single-document write paths used to fall back to indexing the ORIGINAL
  document when the pipeline did not run — `201 Created` for redaction that
  never happened.
- `DELETE /_ingest/pipeline/{id}` removed only the definition and left the
  compiled pipeline live for anything reaching it by name.
- Stage config fails closed: unknown `pii_redaction` type (built an EMPTY
  pattern set and redacted nothing), unknown `grok` pattern name (silently
  became the catch-all `^(?P<message>.+)$`), unknown `convert` target
  (`_ => None` per document, field never converted), and array-typed keys given
  as a bare string (dropped by `and_then(as_array)`, stage behaved as if
  unconfigured).

**Which ingest paths enforce it — measured, not asserted.**

| path | before | now |
|---|---|---|
| `PUT/POST /{index}/_doc[/{id}]?pipeline=` | 400 (already) | 400 |
| `index.default_pipeline`, single-doc write | 400 (already) | 400 |
| `POST /_bulk?pipeline=` | **201, document stored untransformed** | per-item 400 |
| `index.default_pipeline` under `_bulk` | **201, document stored untransformed** | per-item 400 |
| `POST /_reindex` `dest.pipeline` | **200 `created:1`, copied verbatim** | 400 before anything is copied |
| `POST /{index}/_update_by_query?pipeline=` | **200 `updated:1`, document unchanged** | 400 before anything is written |

The last four were the previous round's blocker: the body claimed "every
ingest path refuses loudly" while `BulkOpts` had no `pipeline` field,
`ReindexDest` had only `index`, and `_update_by_query` never read the
parameter — so a *runnable* pipeline was ignored on those paths too
(`_bulk?pipeline=setp` left `{"a":1}` unchanged). They are wired now:

- `BulkOpts::default_pipeline` carries `?pipeline=`, and the per-item
  effective pipeline is resolved as ES resolves it — action metadata, then
  `?pipeline=`, then the target index's `index.default_pipeline`, with `_none`
  disabling at either of the first two levels. Resolution happens before the
  existence check and before the turbo-path partition, so the rest of `_bulk`
  needed no change.
- `ReindexDest::pipeline` and `_update_by_query?pipeline=` are validated
  **before the first document is touched** — an unknown or unrunnable name
  fails the whole request rather than half of it — and then applied to every
  document, with per-document failures reported in `failures[]` rather than
  passed through untransformed.
- One `Engine::default_pipeline_for` is now the single lookup for
  `index.default_pipeline`, so the bulk and single-document forms of the same
  write cannot disagree about which pipeline applies.

**Analysis settings: refuse a block we can only pretend to honour.**
`AnalyzerRegistry::apply_settings` is total by design — it also runs on the
index-*open* path, where a hard failure would brick an existing index — so an
unresolvable tokenizer silently became `standard` and an unresolvable filter
was dropped. `unsupported_analysis` reports every such construct and
`Index::create_with_settings` turns it into a 400 at the door.

> **This is a breaking ES-compat change on index creation, and the ES-YAML
> suite does not cover it** — the suite stays at 1366/0/3 either way, so CI
> being green is not evidence that a given index still creates. The
> [CHANGELOG entry](CHANGELOG.md) lists, from measurement against this branch,
> every settings shape that returned `200` on `main` and returns `400` here:
> a `keyword` `normalizer`; any `char_filter` (`html_strip`, `mapping`, at
> either level); a custom-*named* token filter of type `lowercase`, `stop`,
> `stemmer`, `snowball`, `kstem`, `porter_stem`, `ngram`, `edge_ngram`,
> `word_delimiter_graph`, `trim`, `unique`, `elision`, `decimal_digit`,
> `apostrophe`, `limit`, `truncate`, `pattern_replace`, `keyword_marker`,
> `uppercase`, `reverse`; a custom-*named* tokenizer of type `standard`,
> `whitespace`, `keyword`, `classic`, `uax_url_email`, `path_hierarchy`,
> `char_group`, `letter`, `lowercase`, `simple_pattern`; an analyzer whose
> non-custom `type` is `simple`, `stop`, `pattern`, `fingerprint`, or a
> language other than `english`. There is **no opt-out flag.** Declarations
> whose *name* resolves to a built-in (`english_stop`, `lowercase`,
> `stemmer`/`english_stemmer` with default options — the canonical
> `rebuilt_english` block) are unaffected, as is naming a built-in analyzer on
> a field without any `analysis` block.

**Other user-data paths:** `search_template` with an object `source` that
rendered to invalid JSON used to bolt an `"error"` key onto the *unrendered*
object and search with it; `admin_backup` and `prune_response` swallowed
failures; the ML registry logged nothing.

## Fixes applied after adversarial review

The first review ran the branch rather than reading it, and found that the
sweep had not been applied to the sweep's own new surface. Nine of its eleven
items were real; two are refuted below with evidence.

| # | Finding | Fix |
|---|---|---|
| 1 | `unsupported_analysis` judged a declared filter/tokenizer by its `type`, ignoring that `apply_settings` resolves it by **name**. The canonical Elasticsearch-docs block `english_stop: {type: stop, stopwords: _english_}` was reported unsupported although the built registry really does strip English stopwords — so `PUT /{index}` 400'd a settings block xerj had always served correctly. | The type complaint is skipped when the declared name resolves to a built-in that does exactly what the declaration asks for. When the name resolves but the *options* diverge (a custom `stopwords` list under the name `english_stop`), it is still reported — with a message about the options rather than a misleading "unsupported type". |
| 2 | ES's `grok` takes `patterns` (an array of grok expressions); xerj read `pattern` (a name) and defaulted to `SYSLOG`. `{"patterns": ["%{IP:client} %{WORD:method}"]}` compiled clean and left the document untouched. | `patterns` / `pattern_definitions` are refused as an unsupported *option*: the definition is accepted (ES accepts it) and the pipeline is registered unrunnable, so the write is refused instead of silently mis-ingested. |
| 3 | `build_plugin` silently discarded processor-level `if`, `on_failure` and `ignore_failure`. A `set` with `"if": "ctx.foo == 'never'"` compiled and then ran on the documents the guard excludes. | Same mechanism as (2) — an unhonourable control key makes the pipeline unrunnable rather than half-honoured. `tag` and `description` stay accepted: see the refutations below. |
| 4 | `set.override` was `unwrap_or(false)`; Elasticsearch defaults it to **true** (`SetProcessor.java:135`, read for semantics only). `{"set": {"field": "env", "value": "prod"}}` left an existing `{"env": "dev"}` untouched. | Defaults to true. A present-but-non-boolean `override` is refused rather than defaulted. |
| 5 | `"append" \| "set" => "set"` in `map_proc_name` executed ES `append` as `set` — it *replaced* the field — while the `_simulate` interpreter implemented a real append. One pipeline, two behaviours. | New `AppendPlugin` with ES semantics (scalar widens to a list, a list value extends rather than nests, a missing field becomes a one-element list). Ingest and `_simulate` now agree. `allow_duplicates: false` is refused rather than ignored. |
| 6 | `convert` hard-refused ES `long`, `double` and `auto`, although `ConvertTypePlugin`'s `integer` is i64 and `float` is f64 — equivalent targets, which #204 explicitly permits. | `long` → `integer` and `double` → `float` in the ES→xerj translation. `auto`, `ip`, `short` and `byte` have no equivalent here and stay refused. |
| 7 | The validator's doc-comment claimed it reported *every* construct `apply_settings` accepts and does not honour. False twice: `char_filter` was neither built nor reported, and a non-array `filter` was silently dropped and unreported. `normalizer` was likewise unhandled. | All three are now reported (plus a non-string `tokenizer` and non-string `filter` entries), and the doc-comment enumerates exactly what is checked **and what is not**. |
| 8 | `PUT /{index}/_settings` accepted an `analysis` block, merged it into the display copy, validated nothing and rebuilt nothing — one request walked straight around the new create-time gate. | Refused with a 400. The registry is built once at create/open and there is no rebuild path, so this could only ever change what `GET /_settings` echoes back. ES also refuses analysis updates on an open index, so this narrows towards ES rather than away from it. |
| 9 | `clone_index_to` (backing `_clone` / `_shrink` / `_split`) copied at most 10,000 documents (`size: 10000`) and swallowed every per-document write with `let _ =`, then answered success. | Flush + `search_after` keyset paging over the whole source, and a failed document write fails the request with the id and the count copied so far. |
| 10 | The PR body was the #241 autoindex writeup, opening with a closing keyword aimed at #241. | Replaced; this body. |
| 11 | `schema_persist_failures()` was defined and read by nothing. No HTTP-level test covered the new analysis-block 400. | Exposed on `GET /{index}/_stats` and `GET /_all/_stats`, and covered by HTTP-level tests (below). |

## Fixes applied after the merge gate

The merge gate re-ran every finding against a binary built from the repaired
head and found three of them still live, plus one undisclosed break. All four
are addressed here.

| Finding | Fix |
|---|---|
| **The repair introduced a new silent swallow.** `let _ = src_idx.flush().await;` (`es_compat.rs`) sat inside `clone_index_to` — the one function item 9 was about — on the operation the comment three lines above says the paging depends on. A failed flush yielded a silently partial or duplicated clone still answering `{"acknowledged": true}`. | Propagated with a message naming the source, the target and why the flush is required. `_reindex` carried the identical `let _ = source_idx.flush().await;` (pre-existing) and is fixed the same way. **These were the only two `let _ =` on a user-data path left in this diff.** |
| **"Every ingest path refuses loudly" was measurably false** on `_bulk?pipeline=`, `_bulk` + `index.default_pipeline`, `_reindex` `dest.pipeline` and `_update_by_query?pipeline=` — all four accepted an unrunnable pipeline and stored the untransformed document under 200/201. | Wired, not narrowed. See the table under *What changed*; five new HTTP tests cover the four paths and `_none`. |
| **Pipeline-level (top-level body) `on_failure` was still accepted and silently dropped** — 200 at `PUT`, echoed by `GET`, and the pipeline then ran with no recovery chain. Inconsistent with the same PR's processor-level refusal. | `UNHONOURED_PIPELINE_KEYS` in `put_ingest_pipeline`: the definition is accepted (ES accepts it) and registered unrunnable, exactly as one level down. `description`, `version`, `_meta` and `deprecated` stay accepted — identification metadata, the same line drawn for processor-level `tag`/`description`. |
| **`set` + `copy_from`** (valid ES 7.x+) was 400'd at `PUT` as `mapper_parsing_exception: missing value` — a xerj capability gap classified as a caller error, which is what the `UnsupportedOption` variant exists to prevent, and a hard rejection where `main` returned 200. | Routed through `UnsupportedOption` → accepted at `PUT`, refused at the write, with a message naming `copy_from`. |
| **The breaking analysis gate was in the diff and in no release note.** | A `CHANGELOG.md` hunk, marked BREAKING, enumerating every measured settings shape that stops working, the name-vs-type rule that decides it, and the fact that there is no opt-out. Summarised in the block quote above. |
| `GET /_all/_stats` carried no `mappings` key at all while `Index::schema_persist_failures`'s doc-comment named that endpoint. | The counter is summed across indices and reported at `_all.primaries.mappings.schema_persist_failures`; the doc-comment now says per-index vs aggregate. Covered by a test. |
| `analysis.filter.<x>.synonyms` given as a bare string was still silently dropped — a synonym filter with zero rules, registered, referenced, expanding nothing. | Reported by `unsupported_analysis` (and therefore 400 at create), the same wrong-shape class the repair fixed for an analyzer's `filter`/`tokenizer`. |
| Found while fixing the above: a **failed** `_bulk` item rendered `result: item.result.unwrap_or_else(\|\| "deleted".to_string())`, so every 400'd item reported `"result": "deleted"` next to its own `status: 400` — a claim that a document had been deleted when nothing was written or removed. | `EsBulkItemResult::result` is now `Option<String>` and is omitted on a failed item, as ES omits it. Successful items are unchanged. |
| Found while fixing the above: four `Ok(_) =>` arms treated "the pipeline returned no document" as *index the original body unchanged* — the same silent pass-through, one branch over. Unreachable today (`process_through_pipeline` zips one action per input) but the wrong default. | All four refuse: 400 on the single-document paths, a failed item in `_bulk`, both naming the pipeline. |

### Refuted

- **`tag` and `description` are not silently-discarded behaviour.** They are
  identification metadata: they change no document and no decision on the
  compiled ingest path. They are accepted deliberately, and a test pins that.
  (`if` / `on_failure` / `ignore_failure`, which *do* change behaviour, are
  refused — item 3.)
- ~~**The branch did not fail the ES-compat YAML conformance gate.**~~
  **RETRACTED — this was wrong, and the gate was red at `18e47ad`.** The
  refutation answered the wrong question: it checked whether the *analysis*
  change broke the suite (it did not — the only analysis case,
  `indices.create/10_basic.yml` "Create index with hunspell missing dict",
  expects a 400 and still gets one) and concluded the branch was clean. The
  failure was somewhere else entirely, in the *ingest* change:
  `smoke/smoke_test_multinode/40_simulate.yml`, "Test verbose simulate with
  error in pipeline", `docs.0.processor_results.1.processor_type` — expected
  `rename`, got `field_rename`. `_simulate` walks the STORED pipeline, and
  this branch is what made ES `rename` compile, so the stored form became
  xerj's internal `stages` shape for the first time and the internal stage
  name leaked into the response. It is fixed in the round-2 commit below and
  the suite is green again; the numbers are in **Round 2** under *Gate*.

## Round 4 — the round-2 defect was not fixed, only made less frequent (commits `4854af0`, `2483418`)

Round-3 verification fetched `4d0d645`, re-ran every gate independently, and
confirmed most of the round-3 work — and then reproduced the round-2 blocker
**verbatim on that head**, over HTTP, at a live node. It also found two more
instances of the defect class this PR exists to eliminate, newly created by
round 3. All seven items are addressed here.

**BLOCKER — the refusal marker did not survive CONCURRENT refusals, so the
headline claim was false under load.** `Engine::flush_pipeline_refusals`
rewrote `<data_dir>/ingest-pipeline-refusals.json` whole, snapshotting and
writing with **no lock**, while `flush_cluster_state` — called on the very next
line at both call sites — takes `cluster_state_write` for exactly this reason
and says so in its own comment. Two refused `PUT`s in flight were therefore a
lost update, not a merge: each snapshotted the map, and whichever wrote second
with the older snapshot erased the other's marker, while both callers were told
`{"acknowledged": true}` — contradicting the function's own doc contract
("`Err` means the marker did not reach disk, and the caller MUST NOT
acknowledge").

Measured by the reviewer on `4d0d645` at a live node: 40 concurrent refused
`PUT`s, **40 acknowledged, 37 markers on disk**, plain restart, and the three
lost ones answered **`201`** with `secret` dropped from a document their `if`
guard **excludes**. Every test for this marker in round 3 was sequential, which
is why a fully green suite sat on top of a mechanism that did not hold.

The snapshot now happens **inside** the lock. Measured on this head, same
script, same shape: 40 concurrent refused `PUT`s → **40 acknowledged, 40/40
markers on disk**, plain restart, **40/40 still `400`**.

Fail-before, with the single `let _writing = self.cluster_state_write.lock();`
line removed and nothing else changed — 6 runs, **6 failed**:
`15`, `5`, `10`, `5`, `1`, `9` of 64 markers acknowledged and never written.
(The new test deliberately does **not** barrier its threads: a barrier makes
every insert land before the first snapshot is taken, which closes the very
window the bug lives in. It runs three bursts of 64 so the detector is
reliable rather than 5-in-6.)

**The other six:**

| Finding | Fix |
|---|---|
| **A corrupt sidecar degraded to fail-OPEN after one `PUT` and a restart.** The file is written whole from the in-memory map, which a failed load leaves empty, and the flush was not gated on `pipeline_refusals_unreadable`. One refused `PUT` left the file holding exactly that one marker, erasing every prior refusal; the next boot then parsed it cleanly, read those definitions as "no provenance", and replayed them at `ReplayPersisted`. The advertised fail-closed behaviour lasted exactly one process lifetime. | `flush_pipeline_refusals` refuses while the flag is set, naming the file and the repair; `PUT` answers `500` rather than `{"acknowledged": true}`, and the file is left byte-for-byte as found. The markers are also loaded at boot **unconditionally** now, not from inside `apply_cluster_state`, which returns early when there is no `cluster_state.json` — on that path the flag was never set at all. |
| **The create-time analysis gate covered two of the three spellings this surface accepts.** `put_settings` got `declares_analysis` (all four spellings, with the comment "a gate that tests only the nested one is not a gate"), but `Index::create` routed through `AnalyzerRegistry::unsupported_analysis` → `analysis_block`, which resolves `/analysis` and `/index/analysis` only. Measured on `4d0d645`: nested → `400`, namespaced → `400`, **flat dotted → `200 {"acknowledged":true}`**, with `GET /_settings` echoing the analyzer back. By this CHANGELOG's own words such a declaration "never lowercased anything". | The spelling set moved into `xerj-fts`, next to the resolver that decides which spellings are actually read (`AnalyzerRegistry::dotted_analysis_keys` / `::declares_analysis`), and **both** gates use it — `es_compat::declares_analysis` is now a one-line delegation, so they cannot drift again. Measured on this head, same declaration in four spellings: **`400`, `400`, `400`, `400`**; a settings body with no analysis in it (`index.number_of_shards`, `index.sort.field`) still **`200`**. Refused, not newly honoured: no index already on disk changes what it analyses on reopen. |
| **The new ES-vocabulary `GET` broke native pipelines in both directions, one of them a silent acknowledged downgrade.** `GET /_ingest/pipeline` is the only read endpoint for a pipeline created via `PUT /v1/pipelines/{name}` (the native surface registers no `GET`), and `put_ingest_pipeline` rebuilt the definition as `{description, stages}`. PUTting `GET`'s own output back therefore **dropped `on_error` and `timeout_ms`** under a `200` — the error policy reverting from `pass` (keep the document) to the `Drop` default (**discard** it). The same body sent to the native endpoint answered `500 internal error: missing field \`stages\``. | The translation now **edits** the body instead of rebuilding it — the same rule already applied to `remove`/`rename` one level down — so nothing is dropped, and ES's own `version` / `_meta` / `deprecated` ride along and are echoed back. Measured on this head: `{"description":"n3","on_error":"pass","timeout_ms":250,"stages":[…]}` reads back with both keys, and PUTting that body back leaves both keys, before **and** after a restart. `PUT /v1/pipelines/{name}` given an ES-shaped body now answers **`400`** naming `/_ingest/pipeline/{name}`, not `500`. |
| **`register_unrunnable_pipeline`'s docstring still asserted the reason was in-memory only** "because the boot replay fails to recompile the stored definition and re-records why" — both halves false, in the docstring of the exact function round 3 changed to fix it. | Rewritten to say what the code does, and to name the shape that falsified it (a `stages`-shaped body recompiles softly, which is precisely how the round-2 repro reached `201`). |
| **A fourth undisclosed wire change.** `xerj-console-api`'s `delete_passkey` went from `let _ = idx.delete_document(..)` to `.await?`, so `DELETE /_xerj-console/api/v1/auth/passkeys/{id}` can now answer an error where it always answered `204`. `grep -i passkey CHANGELOG.md` returned nothing. | Added to the CHANGELOG entry, which now says **four** rather than three, with the behaviour spelled out (an already-absent credential is still `204`; only a genuine failure surfaces). |
| **`put_pipeline`'s `keep_if_uncompilable: Some(doc)` arm had no caller in the workspace** — every call site passed `None` — yet it held its own copy of the refusal-marker bookkeeping in a security-relevant path, with zero coverage. | The parameter is gone and the branch with it; the ES surface's "store an uncompilable definition" case goes through `register_unrunnable_pipeline`, which is the path that is actually tested. Six call sites updated. |

**One follow-up found while re-reading the round-4 diff (`2483418`).** Deleting
the dead `keep_if_uncompilable` arm left `put_pipeline` able to make exactly one
kind of marker change — a REMOVAL, via `compile_pipeline` clearing the marker on
a successful compile — and it inherited the flush order written for the INSERT
case. Marker-file-first is the safe order for an insert and the unsafe one for a
removal: a `flush_cluster_state` failure in between left the marker cleared on
disk while the OLD, refused definition was still stored, so the next boot would
replay it softly. The definition now lands first and the clear follows
(best-effort — a stale marker only makes a pipeline more refused, and the boot
self-heal clears it), while `register_unrunnable_pipeline` keeps the opposite
order because it inserts. Both comments now name the direction they protect.
Argued from the code path, not demonstrated: the failure is a filesystem error
and there is no injection point for one at this layer.

### Round 4 — measured at `4854af0`, re-measured at `b4306a1`, `2483418` and `a5d3f06`

`origin/main` moved to `e02a726` (#378, the unity autoindex reland) while this
round was in flight. It is merged in at `b4306a1`; the only conflict was
`CHANGELOG.md` (both sides additive — `### Fixed` from #378, `### Changed` /
`### Fixed` from #204 — kept whole, one `### Fixed` heading). Everything below
was re-run on the merge.

- `cargo fmt --all --check` — exit 0.
- `cargo clippy -p xerj-engine -p xerj-api -p xerj-fts --all-targets -- -D warnings` — exit 0, zero warnings.
- `cargo test -p xerj-api -p xerj-engine -p xerj-fts --profile ci-test --no-fail-fast` — **67 test binaries `ok`**, one failure: `index::merge_publication_transaction_tests::partial_multi_batch_failure_is_reported_without_losing_either_batch` (`556 passed · 1 failed` in `--lib`). Pre-existing **#372**: it passes in isolation (`9 passed · 548 filtered out` with `--test-threads=1`) and `git diff e02a726..HEAD -- engine/crates/xerj-engine/src/index.rs` matches that module **zero** times.
- `cargo test -p xerj-api --test fail_closed_config_http` (dev profile, overflow checks ON) — **36 passed · 0 failed**.
- `cargo test -p xerj-engine --test integration` (dev profile) — **120 passed · 0 failed · 1 ignored**.
- `cargo test -p xerj-fts --lib` (dev profile) — **76 passed · 0 failed**.
- `cargo test -p xerj-engine --test es_compat_tests --profile ci-test` — **65 passed · 0 failed**.
- **ES-compat YAML conformance**, own release binary from this tree, private port `19891`, throwaway data dir — **`1366 passed · 0 failed · 3 skipped · 1369 total`**, `RUNNER_EXIT=0`. Re-run after the `origin/main` merge: identical.

**CI history on this head, stated rather than glossed.** The first run at
`d68b5e0` had **Build + Test red** on
`xerj-fts::search::tests::bool_deadline_never_broadens_a_must`:
`precondition: the must clauses must have intersected to the document before
the budget (22.459885ms) ran out — left: 0, right: 1`. That test asserts a
**wall-clock** budget (`budget = full_walk / 2`), it was added by #130 on
`2026-08-01` and is on `main`, and `git diff origin/main..HEAD --
engine/crates/xerj-fts/src/search.rs` is **empty** — this branch does not touch
that file. It passed 8 out of 8 local runs of the same binary at full
parallelism (`76 passed` each time), and the re-run of the same job on the same
commit went green. Every other job in that run was green first time, including
**ES-compat YAML conformance**, **Format + Clippy** and **Release build (fat
LTO)**. `origin/main` then moved twice more (`d17ab91`, then `de1e5e0` = #356);
both are merged in at `a5d3f06`, the second one conflicting only in
`CHANGELOG.md` again (#356's `### Performance` and #204's `### Changed` are
disjoint sections, both kept whole). Re-run on that merge:
`cargo test -p xerj-api -p xerj-engine -p xerj-fts --profile ci-test
--no-fail-fast` → **68 binaries `ok`, zero failures** (the #372 flake did not
trigger on that run), clippy still zero warnings, `cargo fmt --all --check`
clean.

**Fail-before, verified by reverting only the source hunk and re-running:**

| reverted hunk | test that then FAILED | observed |
|---|---|---|
| `cluster_state_write.lock()` in `flush_pipeline_refusals` | `concurrent_definition_time_refusals_all_reach_disk` | 6 runs, 6 failed: `15`/`5`/`10`/`5`/`1`/`9` of 64 markers "acknowledged and never reached disk" |
| the `pipeline_refusals_unreadable` guard | `a_refusal_is_not_acknowledged_while_the_sidecar_is_unreadable` (engine) | `this must NOT be acknowledgeable while the sidecar is unreadable: ()` |
| same hunk | `a_refused_put_is_not_acknowledged_while_the_sidecar_is_unreadable` (HTTP) | `left: 200, right: 500`, body `{"acknowledged":true}` |
| the dotted-spelling report in `unsupported_analysis` | `every_spelling_of_an_unhonourable_analysis_block_is_refused_at_create` | `left: 200, right: 400`, body `{"acknowledged":true,…,"index":"spelling_2"}` |
| same hunk | `the_dotted_spellings_of_an_analysis_declaration_are_reported` (fts unit) | FAILED |
| the edit-don't-rebuild translation | `a_pipelines_error_policy_survives_a_round_trip_through_the_es_endpoint` | `left: Null, right: "pass"`, body `{"n3":{"description":"n3","processors":[…]}}` — `on_error` and `timeout_ms` gone |
| same hunk | `es_pipeline_metadata_is_not_dropped_by_the_translation` | `left: Null, right: 3` — `version` dropped |
| the native ES-body guard | `the_native_endpoint_answers_400_and_names_the_endpoint_for_an_es_body` | `left: 500, right: 400`, body `internal error: missing field \`stages\`` |

**Verified vs assumed, round 4.** *Verified:* every command and every number
above, run on this head; the concurrency repro at a live node, both directions;
the four-spelling probe at a live node; the native round trip at a live node.
*Assumed:* that the `#372` flake stays confined to full-parallelism runs (it is
briefed as pre-existing, its module is not in this diff, and it passes in
isolation, but no bisect was run); that no ES client sends a non-string
`description` on a pipeline body — the translation used to coerce it to `""`
and now carries it through, so such a body would become a compile error rather
than a silently blanked description (no YAML conformance case sends one); and
that carrying ES's `version`/`_meta`/`deprecated` into the stored config is
harmless because `PipelineConfig` is not `deny_unknown_fields` (read, not
fuzzed). The live `/v1/pipelines/{name}` probe returned `404` on the ES-compat
port because the native routes are not mounted there; that endpoint's `400` is
covered by a test that builds `router::build_native_router` directly.

## Round 3 — the blocker round 2's own fix introduced (commit `88e9068`)

Round-2 verification fetched `2e13f9f`, re-ran every gate, reverted each source
hunk individually, and confirmed almost everything the branch claimed — and then
found a defect **round 2 itself created**. It is fixed here, together with the
four secondary items from the same review.

**BLOCKER — a pipeline THIS build refused became runnable again after a plain
restart, with the refused option ignored, writing wrong documents under a
`201`.** Demonstrated live with nothing but xerj's own output:

1. `PUT /_ingest/pipeline/redact {"processors":[{"remove":{"field":"secret"}}]}` → acknowledged
2. `GET /_ingest/pipeline/redact` → `{"redact":{"description":"redact","stages":[{"type":"drop_field","config":{"fields":["secret"]}}]}}`
3. the operator edits **that** body, adds `"if":"ctx.tenant == 'a'"`, PUTs it back → `{"acknowledged":true}`
4. `PUT /docs/_doc/1?pipeline=redact {"tenant":"b","secret":"ssn-111"}` → **400**, correctly refused
5. restart the node — no PUT, no operator action
6. the identical write → **201**
7. `GET /docs/_doc/2` → `{"_source":{"tenant":"b"}}` — `secret` dropped from a document the guard **excludes**

Root cause: `CompileMode::ReplayPersisted` (round 2's fix for "an upgrade must
not flip 201 into 400") had no way to distinguish *persisted by an older, laxer
build* from *persisted by THIS build and explicitly recorded as unrunnable*.
`put_ingest_pipeline`'s `else { body.clone() }` branch stores a stages-shaped
body verbatim, `register_unrunnable_pipeline` persists it, and
`Engine::unrunnable_pipelines` was **in-memory only** — at boot `soften!`
downgraded the `if` refusal to a degradation and
`compile_pipeline_with_mode`'s `unrunnable_pipelines.remove(name)` erased the
PUT-time refusal. It covered every check the sweep added (`if` / `on_failure` /
`ignore_failure` / `ignore_missing` / `ignore_empty_value` / grok `patterns` /
convert target / `copy_from` / `allow_duplicates`); only an unimplemented
*stage* survived a restart, because the raw ES `processors` body it stores fails
to deserialise.

**Fix — provenance, the same pattern this branch already used for
`<index>/analysis-binding.json`.** A refusal recorded at definition time is
persisted in `<data_dir>/ingest-pipeline-refusals.json`, and its **absence** is
the load-bearing signal:

| on disk | replayed at | why |
|---|---|---|
| marked refused | `CompileMode::Strict` | the operator asked, was told 400; a restart must not answer differently |
| unmarked | `CompileMode::ReplayPersisted` | written by a build with no gate — the upgrade door stays open for the case it was cut for |

The marker is provenance, not a tombstone: a marked definition that compiles
cleanly on the new build clears its own marker at boot and runs (INFO), so a
capability upgrade needs no re-PUT; a successful re-PUT or a `DELETE` clears it
too. If the file exists but cannot be read or parsed, **no** pipeline's
provenance is known, so every persisted pipeline replays at definition
strictness and each failure names the file in its reason — fail closed, loudly,
rather than silently reopening the hole.

It is a **sidecar**, not an eighth field in `cluster_state.json`, because that
envelope is `#[serde(deny_unknown_fields)]` (`cluster_state.rs:90`): widening it
would make every file this build writes unreadable to an older binary — a
downgrade break traded for an upgrade fix. The marker reaches disk **before**
the definition; a marker without its definition makes the *previous* definition
replay strictly (it compiled once, so it compiles again, and the boot self-heal
clears the marker), whereas a definition without its marker is the defect
itself.

*Reference coding.* Lucene records the same kind of fact for the same reason:
`SegmentInfos.indexCreatedVersionMajor`
(`lucene/core/src/java/org/apache/lucene/index/SegmentInfos.java:172` — "The
Lucene version major that was used to create the index") is written once into
the commit (`SegmentInfos.java:633`, `out.writeVInt(indexCreatedVersionMajor)`)
and then **decides how the bytes are read back** rather than being re-derived:
`readCommit` throws `CorruptIndexException` when a segment predates the recorded
creation version (`SegmentInfos.java:491-501`), and `readLatestCommit` refuses a
commit whose recorded creation major is below the caller's floor
(`SegmentInfos.java:572-578`). The shape is what is taken, not the code: what a
*later* reader may assume about persisted bytes must be recorded **with** them
at write time, because it cannot be recovered afterwards. Lucene is Apache-2.0,
the same licence as XERJ; no code is reproduced.

**`GET /_ingest/pipeline[/{id}]` now speaks Elasticsearch.** It returned xerj's
internal `stages` shape for every pipeline that compiled — measured before this
commit:
`{"p_rename":{"description":"d","stages":[{"type":"field_rename","config":{"mappings":{"a":"b"}}}]}}`.
`_simulate` was taught this inverse in round 2 (`stages_to_es_processors`); GET
was not. It is a wire-visible fix in its own right, **and** it is what made the
blocker reachable by construction, because GET's output is what an operator
edits and PUTs back, and a stages-shaped body takes the handler's pass-through
branch. `set` is now rendered with an explicit `override`, always — it is the
one key whose default differs between the two vocabularies (ES `true`, xerj's
native `set` stage `false`, deliberately), so a bare ES `set` would claim a
native stage overwrites when it preserves. Stating it is exact in both
directions and round-trips. Relatedly, `_simulate`'s `set` interpreter honoured
no `override` at all and always overwrote, reporting a document ingest would
never produce; it now applies ES's rule.

**Three wire-visible changes that had no CHANGELOG entry** now have one:
`POST /v1/admin/backup` answering `400` on a present-but-unparseable body where
it previously ran a *default* backup under a success response; the added
non-Elasticsearch key `primaries.mappings.schema_persist_failures` inside
`GET /{index}/_stats` and `GET /_all/_stats`; and authz `prune_response`
answering `500` instead of falling back to the unfiltered bytes.

**`ignore_missing: false` — decided, not deferred.** Kept as a refusal, and the
CHANGELOG now states the decision and why it lands on the *opposite* side of the
rule from `ignore_failure: false`, instead of filing it under known gaps.
`ignore_failure: false` means "a failing processor is not ignored", which is
exactly what xerj does → accepted. `ignore_missing: false` means "**fail** the
document when the field is absent", which xerj cannot do at all
(`ProcessAction` has no error variant) → refused, because accepting it would be
precisely the silent lie this PR exists to remove. The asymmetry (omission
accepted, the ES default written out refused) is deliberate and is now called
out as a hard `400` for ES-generated pipelines, with the one-line repair.

**Also corrected in the CHANGELOG:** the door's stated *scope* ("a definition
already in `cluster_state.json` was accepted by whichever build wrote it") and
its stated *repair path* ("a re-PUT answers 400 with the same reason") were both
falsified by the repro above — the operator had already made that PUT and the
restart undid the answer.

### Round 3 — measured at `88e9068`, re-measured at `4d0d645`

`origin/main` moved to `a20dd82` (#322, the WAL tap) while round 3 was in
flight. It is merged in at `4d0d645`; the only conflict was `CHANGELOG.md`
(both sides additive — `### Added` from #320, `### Changed`/`### Fixed` from
#204 — kept whole, in Keep-a-Changelog order). Everything below was re-run on
the merge; where the counts differ from the pre-merge run it is because #322
added a test binary and four cases:

- `cargo fmt --all --check` — exit 0 · `cargo clippy -p xerj-engine -p xerj-api --all-targets -- -D warnings` — exit 0, zero warnings.
- `cargo test -p xerj-api --profile ci-test` — **25 binaries · 400 passed · 0 failed · 2 ignored**, exit 0 (was 24 / 396 pre-merge).
- `cargo test -p xerj-engine --profile ci-test` — **41 binaries · 938 passed · 1 failed · 23 ignored**; the one failure is the same pre-existing #372 flake described below.
- `cargo test -p xerj-api --test fail_closed_config_http` (dev profile) — **30 passed · 0 failed**, exit 0.
- `cargo test -p xerj-engine --test integration` (dev profile) — **118 passed · 0 failed · 1 ignored**, exit 0.
- **ES-compat YAML conformance** on the merged tree, own binary, private port `19855`, throwaway data dir — **`1366 passed · 0 failed · 3 skipped · 1369 total`**, exit 0.

Pre-merge numbers at `88e9068`, for the record

On a 20-core x86_64 Linux host, in a **private** `CARGO_TARGET_DIR` — the shared
one was found to be raced by another worktree mid-session (it resolved
`xerj-wasm`/`xerj-fts` from a different checkout), so every number below was
re-measured in isolation.

- `cargo fmt --all --check` — exit 0.
- `cargo clippy -p xerj-engine -p xerj-api --all-targets -- -D warnings` — exit 0, zero warning lines.
- `cargo test -p xerj-api --profile ci-test` — **24 binaries · 396 passed · 0 failed · 2 ignored**, exit 0.
- `cargo test -p xerj-api` (dev profile, overflow checks ON) — **24 binaries · 396 passed · 0 failed · 2 ignored**, exit 0.
- `cargo test -p xerj-engine --profile ci-test` — **40 binaries · 908 passed · 1 failed · 23 ignored**. The one failure is the briefed pre-existing #372 full-parallelism flake, `index::merge_publication_transaction_tests::partial_multi_batch_failure_is_reported_without_losing_either_batch`; it passes in isolation (`1 passed · 547 filtered out`) and its module is untouched by the entire PR (`git diff ca4d75a..HEAD -- index.rs` matches it zero times).
- `cargo test -p xerj-engine --test integration` (dev profile) — **118 passed · 0 failed · 1 ignored**, exit 0.
- **ES-compat YAML conformance**, own binary built from this tree, booted on private port `19855` with a throwaway data dir (the runner issues `DELETE /_all` before every case, so it must never see a shared node): **`1366 passed · 0 failed · 3 skipped · 1369 total`**, exit 0.

**Fail-before verified by reverting only the source hunk and re-running** — not asserted:

| reverted hunk | test that then FAILED | observed |
|---|---|---|
| boot replay always `ReplayPersisted` | `a_refusal_survives_a_restart_for_a_stages_shaped_definition_too` | `left: 201, right: 400` — the reviewer's live repro, reproduced |
| `register_unrunnable_pipeline` does not record the marker | `a_definition_time_refusal_is_recorded_on_disk_and_replayed_strictly` | "a restart must not resurrect a definition this build refused" |
| GET returns the stored doc verbatim | `get_ingest_pipeline_answers_in_elasticsearch_vocabulary`, `get_states_the_override_the_stage_actually_applies`, `a_refusal_recorded_at_put_time_survives_a_restart` | `left: Null, right: "a"`, body shown: `{"p_rename":{"description":"d","stages":[…]}}` |
| `stage_to_es_processors` does not state `override` | `set_renders_the_override_it_actually_applies` | `left: Null, right: false` |
| `_simulate` ignores `override` | `simulate_honours_the_set_override_flag` | `left: "prod", right: "dev"` |

The compatibility door is pinned in the other direction too:
`a_definition_with_no_refusal_marker_still_replays_softly` builds the exact
pre-sweep on-disk state — a definition that **compiled**, carrying an option no
build then checked, with no marker file — and asserts the write still succeeds
after the restart, then that a live re-PUT of the same thing is still refused.

**Verified vs assumed, round 3.** *Verified:* every command and every number
above; the seven-step repro, reproduced as a test that fails before and passes
after; all four Lucene citations, read line-by-line from the checkout at
`/home/claude/ref/lucene`. *Assumed:* that the `#372` flake stays confined to
full-parallelism runs (it is briefed as pre-existing and its module is not in
this diff, but no bisect was run); that no operator depends on `GET
/_ingest/pipeline` returning the internal `stages` shape (no YAML case asserts
that body, and speaking ES is the documented contract of the endpoint); and that
a corrupt refusal-marker file is rare enough that failing closed for every
pipeline is the right trade — it is loud, and the reason names the file, but it
is a cluster-wide ingest stop until repaired.

*One thing worth flagging for whoever reads this next:* the `refsym` reference
index this repo queries for reference coding was **empty** when checked during
this round (the node on `:9200` held one unrelated index). The likeliest cause
is another agent pointing the ES-YAML runner at the shared node — it issues
`DELETE /_all` before every case. The citations above were therefore read from
the on-disk Lucene checkout instead. That is a tooling accident, not a claim
about the corpus, and the index needs rebuilding.

## Known gaps, stated rather than closed

- **The analysis gate has two holes.** An **index template** carrying
  `template.settings.analysis` is accepted with 200 and the block is dropped
  wholesale at index creation (measured: `GET /tplidx1/_settings` shows no
  analysis), so the create-time gate never sees it — the template analogue of
  the `_settings` bypass this PR did close. And **option-level divergence
  inside a supported type** is not checked at all: `synonym.synonyms_path`
  (file-backed rule lists are never read), `length.min` (defaults to 2 here, 0
  in ES), `asciifolding.preserve_original`. Both are pre-existing; neither is
  fixed here.
- **`GET /_ingest/pipeline/{id}` carries no runnability signal.** An
  unrunnable pipeline is echoed back exactly as ES would echo it, so an
  operator only discovers the gap at write time — despite this PR saying the
  reason is "recorded". Adding a xerj-specific key to that response would
  diverge from the ES wire shape, so it is left alone and named here.
- **A write addressed to an alias does not pick up the backing index's
  `index.default_pipeline`.** Settings are looked up by literal index name.
  Identical on every write path, single-document and bulk; pre-existing.
- **`_simulate` and the compiled ingest path can disagree, and that is
  deliberate.** `_simulate` interprets the stored ES body directly and covers
  processors the compiled path does not — the ES `pipeline` processor is the
  canonical case, and the conformance suite depends on it. So a pipeline
  recorded as *unrunnable* will still show a transformation under `_simulate`
  that ingest refuses to perform. The disagreement is always in the safe
  direction — `_simulate` is optimistic, the write refuses — but it is a
  disagreement.
- **`_simulate`'s own `if` support is literal-only.** It skips a processor for
  `"if": "false"` and runs it otherwise; an arbitrary painless expression is not
  evaluated there either, and the verbose entry reports
  `{"result": cond == "true"}` regardless. Pre-existing, on the simulate
  interpreter, not touched here — and the reason this PR refuses `if` at
  definition time rather than claiming parity with simulate.
- **`_update_by_query` reports a `drop` processor as a per-document failure**
  rather than deleting the document. It updates in place and has no delete
  path, so honouring `drop` there is not something this endpoint can do; saying
  so per document beats writing the document as if the processor had not asked.

## Tests

- `xerj-fts` — `unsupported_analysis` cases for the canonical ES block
  (asserting both halves of the reviewer's measurement: no complaint, **and**
  the built registry emits `["quick","brown","fox"]` from "The Quick Brown
  Fox"), for a name that resolves with divergent options, for `char_filter`,
  `normalizer`, and wrong-shaped `filter` / `tokenizer`.
- `xerj-wasm` — `if` / `on_failure` / `ignore_failure` refusal, metadata keys
  still compiling, ES `grok` `patterns`, `set` override default (both
  directions plus a non-boolean), and `append` semantics.
- `xerj-api/tests/fail_closed_config_http.rs` — **36** HTTP-level tests at the
  merged head (`cargo test -p xerj-api --test fail_closed_config_http` →
  36 passed · 0 failed). The count grew each round; the figure of 18 that
  earlier revisions of this body quoted was the round-3 total and is corrected
  here. They accumulate as:
  - *Round 1/2* — analysis 400 with its ES error envelope, the canonical block
    still 200, the `_settings` bypass and its dotted spelling, grok/`if`
    pipelines accepted at PUT and refusing the write, `set` override, `append`,
    `convert long|double`, `ignore_missing`, the `remove`/`rename` guard not
    being dropped by the translation, and a `_clone` of 10,001 documents that
    also asserts the pre-fix `size: 10000` search really does top out at
    10,000 — so it fails on the old code rather than passing vacuously.
  - *Round 3* — `_bulk?pipeline=` running and refusing,
    `index.default_pipeline` under `_bulk` including `_none`, `_reindex`
    `dest.pipeline` (validated up front, applied per document),
    `_update_by_query?pipeline=` (same), pipeline-level `on_failure`, `set`
    `copy_from`, `_all/_stats` reporting the counter, the bare-string
    `synonyms`, the failed-bulk-item `result` assertion, the refusal marker
    surviving a restart (including the stages-shaped definition), a marker
    cleared by a definition that compiles, an unmarked definition still
    replaying softly, and `GET` answering in ES vocabulary.
  - *Round 4* — every spelling of an unhonourable analysis block refused at
    create, a pipeline's error policy surviving a round trip through the ES
    endpoint, ES metadata not dropped by the translation, concurrently refused
    pipelines all still refused after a restart, a refused `PUT` not
    acknowledged while the sidecar is unreadable, and the native endpoint
    answering 400 (not 500) for an ES-shaped body.
- `xerj-api/tests/reindex_preserves_generated_vectors.rs` —
  `clone_preserves_generated_embedding_companions`. `_clone`/`_shrink`/`_split`
  funnel through the same `clone_index_to` this PR rewrote from a single
  `size: 10000` search into a keyset-paged loop; this pins that the per-page
  body keeps its explicit `"_source": true`, so the rewrite cannot trade a
  silently truncated copy for a silently lossy one.

## Round 2 — the sweep's own blind spots (commits `934c026`, `2e13f9f`)

Nine defects found by review of the round-1 branch, all fixed here. One turned
the conformance gate red; four were "the gate does not apply to itself"; four
were upgrade breaks the sweep itself introduced.

**Red gate.** `_simulate`'s stages→processors conversion emitted the internal
stage name (`field_rename` for ES `rename`), which only became reachable once
this branch made `rename` compile. `stage_to_es_processors` now inverts both
halves of the translation — name and config — at both call sites.

**The gate did not apply to itself.**

1. `ignore_failure: false` — Elasticsearch's own default, which ES clients emit
   verbatim — was REFUSED, because `UNHONOURED_PROCESSOR_KEYS` tested key
   *presence*. That pipeline registered and then 400'd every write through it,
   while the byte-identical processor without the key indexed fine. The check
   now reads the VALUE: accept what this build implements (and absence, which
   is not configuration), refuse any other written value. Lucene's
   consume-then-complain shape, not a presence check —
   `AbstractAnalysisFactory.getBoolean` at
   `lucene/core/src/java/org/apache/lucene/analysis/AbstractAnalysisFactory.java:213-217`
   removes the key and returns the default only when it was absent;
   `ArabicStemFilterFactory` at
   `lucene/analysis/common/src/java/org/apache/lucene/analysis/ar/ArabicStemFilterFactory.java:44-51`
   throws on whatever is left.
2. `remove` and `rename` bypassed the gate: their translation built a fresh
   config holding only the translated keys, so a processor-level
   `if`/`on_failure`/`ignore_failure`/`ignore_missing` was dropped before
   `build_plugin` saw it. `{"remove": {"field": "secret", "if": "ctx.tenant ==
   'a'"}}` compiled, was acknowledged, and dropped the field on every document
   — while the identical `if` on `set` was refused. Every branch now edits the
   caller's config instead of rebuilding it.
3. `ignore_missing` / `ignore_empty_value` were neither honoured nor refused.
   `ignore_missing: true` is exactly what `remove`/`rename`/`convert`/`date`
   already do, so it is honoured; `false` asks for the document to be failed,
   which no stage can do (`ProcessAction` has no error variant), so it is
   refused. The remaining divergence — an OMITTED `ignore_missing`, where ES
   defaults to `false` — is now in the CHANGELOG's known gaps rather than
   silent.
4. The `PUT /{index}/_settings` analysis refusal missed the DOTTED spelling, so
   `{"index.analysis.analyzer.x.type": "custom"}` walked past it and
   `GET /_settings` echoed an analyzer that analysed nothing.

**Upgrade breaks the sweep introduced.**

5. Boot replay re-compiles every persisted pipeline, so an unknown grok pattern
   name, an unknown convert target, a bare string where an array is required,
   `set.copy_from` or a processor-level `if` flipped ALREADY-STORED working
   pipelines from 201 to 400 on a binary upgrade alone. New
   `CompileMode::ReplayPersisted` reproduces the previously-shipped behaviour
   for every check this sweep added, records each in `Pipeline::degradations` /
   `Engine::degraded_pipelines`, and logs it at ERROR; a live re-PUT still
   answers 400 with the same reason, which is the repair path. Definitions that
   did not compile before the upgrade either still fail at boot.
6. `set.override`'s default was flipped false→true in the plugin, which changed
   NATIVE pipelines too — they are recompiled at every boot, so a `set` stage
   that had preserved an existing value for the life of a cluster would start
   overwriting it. The native default is restored; ES parity now lives in the
   translation, which writes `"override": true` explicitly.
7. `apply_settings` resolving `/index/analysis` is a fix at index CREATE and a
   data bug at index OPEN. `AnalysisBinding` makes the choice a property of the
   index, recorded at create in `<index>/analysis-binding.json`; a pre-existing
   index with documents and analyzers declared only under `index.analysis`
   keeps the legacy binding and gets an ERROR naming the reindex repair, while
   an empty one adopts the canonical binding and records that.
8. The CHANGELOG's ingest entry now carries a **BREAKING** marker (it converts
   201 into 400 for any pipeline using an ES processor xerj does not
   implement), the `set.override` claim is corrected, and every item above is
   documented.

## Gate

Every number in this section is from a command that was run, on the commit the
sub-heading names; nothing is estimated. Rounds are kept rather than collapsed
so that a claim can be traced to the head it was measured on — where a later
round re-measured something and got a different figure, the later round is the
one that holds.

### Round 2 — measured at `2e13f9f` (this branch merged with `origin/main` `ca4d75a`)

Commands run, on an `aarch64` Linux host:

- `cargo fmt --all --check` — clean (exit 0).
- `cargo clippy -p xerj-wasm -p xerj-fts --all-targets -- -D warnings` — clean.
- `cargo clippy -p xerj-engine -p xerj-api --all-targets -- -D warnings` — clean.
- `cargo test -p xerj-wasm -p xerj-fts --profile ci-test` — **74 + 45 + 12 + 1
  + 0 passed · 0 failed** (the round-2 body said "72 + 45 + 12 + 1"; the real
  counts are 74 for the `xerj-fts` lib and a fifth, empty binary — corrected
  here after an independent re-run measured them).
- `cargo test -p xerj-wasm -p xerj-fts` (dev profile, overflow checks ON) —
  same counts, 0 failed.
- `cargo test -p xerj-engine --profile ci-test --test integration` — **114
  passed · 0 failed · 1 ignored**.
- `cargo test -p xerj-api --profile ci-test` — **24 test binaries · 386 passed ·
  0 failed** (exit 0).
- **ES-compat YAML conformance**, against a binary built from this commit,
  booted on private port `19733` with a throwaway data dir (the runner calls
  `DELETE /_all` before every case, so it must never be pointed at a shared
  node): **`1366 passed · 0 failed · 3 skipped · 1369 total`, `RUNNER_EXIT=0`**
  — including `40_simulate.yml` "Test verbose simulate with error in pipeline",
  the case that was red at `18e47ad`.

**Fail-before was verified by reverting the source hunk and re-running**, for
every fix, not asserted:

| reverted hunk | test that then FAILED |
|---|---|
| presence check for `if`/`on_failure`/`ignore_failure` | `writing_the_honoured_default_is_the_same_as_omitting_it`, `writing_an_unhonoured_value_is_refused_even_when_it_is_the_es_default`, `spelling_out_the_elasticsearch_default_does_not_break_the_pipeline` |
| `set.override` default → `true` | `native_set_stage_keeps_its_non_overriding_default` (`left: "prod", right: "dev"`) |
| `ignore_missing` unchecked | `ignore_missing_is_honoured_or_refused_never_silently_accepted` |
| stage name emitted verbatim in `_simulate` | `simulate_reports_elasticsearch_processor_names_for_a_compiled_pipeline` (`left: "field_rename", right: "rename"`) |
| `remove`/`rename` rebuild the config | `a_guard_on_remove_or_rename_is_not_dropped_by_the_translation` |
| `declares_analysis` → nested pointer only | `the_dotted_analysis_spelling_does_not_walk_past_the_settings_gate` |
| boot replay uses `CompileMode::Strict` | `a_persisted_pipeline_the_new_checks_refuse_still_runs_after_a_restart` |
| `analysis_binding_for_open` → always canonical | `reopening_a_pre_sweep_index_does_not_silently_change_how_it_analyses` (`left: 1, right: 0`), `an_empty_pre_sweep_index_adopts_the_canonical_binding_and_records_it` |

### Round 1 (unchanged, measured at the earlier merge)

Re-run on the **merge of this branch with `main`** at `2f2c3986` (rc.14, the
autoindex PDF-parse-once change, and the autoindex rerun-refusal work), not on
the pre-merge head:

- `cargo clippy -p xerj-fts -p xerj-wasm -p xerj-engine -p xerj-api
  -p xerj-console-api --all-targets -- -D warnings` — clean.
- `cargo test -p xerj-fts -p xerj-wasm -p xerj-engine -p xerj-api
  -p xerj-console-api` — **62 test binaries · 1235 passed · 0 failed.**
- **ES-compat YAML conformance**, against a server on a private port with a
  throwaway data dir: **`1366 passed · 0 failed · 3 skipped · 1369 total`.**

Every claim the merge gate measured false was re-measured live against a binary
built from this merge — the four ingest paths (refusing an unrunnable pipeline
*and* running a runnable one, plus `_none`), pipeline-level `on_failure`,
`set` `copy_from`, `_all/_stats`, the bare-string `synonyms`, and the failed
bulk item's absent `result`. The CHANGELOG's breaking-change table was checked
in **both** directions: the six shapes it says now fail return 400, and the
shapes it says are still honoured (`synonym`/`length`/`shingle`/`asciifolding`
filters, `ngram`/`edge_ngram`/`pattern` tokenizers, the `standard`/`english`/
`whitespace`/`keyword` analyzers, and the canonical `rebuilt_english` block)
still return 200. The three Known gaps that a single request can demonstrate
were re-measured too, and each behaves exactly as stated: an index template's
`analysis` block is dropped (`GET /tplidx1/_settings` returns no `analysis`),
`GET /_ingest/pipeline/{id}` echoes an unrunnable pipeline with no runnability
signal, and a write addressed to an **alias** whose backing index carries
`index.default_pipeline` is stored untransformed under a `201`.

One caveat, stated rather than hidden: `xerj-engine --test
painless_script_limits` aborts with a stack overflow under this sandbox's
default test-thread stack size, and passes 9/9 with
`RUST_MIN_STACK=16777216` — which is the run counted above. It is a
deep-recursion probe of the Painless call-depth limit, it is untouched by this
diff (last changed by #116, already on `main`), and it is green in CI.


xerj-org:
## Full audit — every site kept, with a verdict

Method: `grep -rn "unwrap_or_else\|unwrap_or_default\|let _ ="` over
`engine/crates` excluding `tests/` → **682** hits. Discarded as not
user-configuration / not user-data: error-message construction
(`id.unwrap_or_default()` in bulk item results), `#[cfg(test)]` lifecycle hooks,
shutdown channels (`let _ = tx.send(())`), mutex-poison recovery, `OnceLock::set`
on already-terminal results, doc-comment examples, benches. **41** sites kept.

### Weaker → fixed in this PR (11)

| site | why weaker |
|---|---|
| `xerj-api/src/es_compat.rs` `put_ingest_pipeline` | compile failure warn-logged, body stored, `200 acknowledged`; `GET` echoed a pipeline that never ran |
| same, processors `filter_map` | a non-processor entry silently dropped → pipeline compiled with fewer stages than written |
| `xerj-wasm/src/pipeline.rs` `pii_redaction` types | unknown type → empty pattern set → redacts nothing, reports success |
| `xerj-wasm/src/pipeline.rs` `grok` pattern | unknown name → generic catch-all, 1 field instead of 10 |
| `xerj-wasm/src/pipeline.rs` `convert` type | unknown target → `_ => None` per doc, field never converted |
| `xerj-wasm/src/pipeline.rs` `formats`/`types` shape | bare string dropped by `and_then(as_array)` → stage acts unconfigured |
| `xerj-fts/src/analyzer.rs:1231-1234` tokenizer | unresolvable tokenizer → `StandardTokenizer`, silently, no log at all |
| `xerj-fts/src/analyzer.rs:1244-1255` filter | unresolvable filter dropped (warned, but the create still succeeded) |
| `xerj-fts/src/analyzer.rs` `analysis_block` | ES's canonical `settings.index.analysis.*` read by nobody → zero analyzers, silently |
| `xerj-engine/src/index.rs:17662,17736` `let _ = save_schema` | dynamic mapping evolved in memory, never reached disk, rolled back at restart |
| `xerj-api/src/authz.rs:1480` | pruning re-serialisation failure → **original unpruned** listing to a scoped principal |
| `xerj-api/src/state.rs` ML `save_all`/`load_all` (×4) | registry write failure and corrupt-file read both silent |
| `xerj-api/src/native.rs:828` `admin_backup` | malformed body → different backup (default path, all indices) under `201 Created` |

### Equivalent → left alone (30)

| site | verdict |
|---|---|
| `xerj-console-api/src/auth/store.rs:80,147,155,192,236,281` `let _ = delete_document` | **equivalent** — delete-then-create upsert; a genuinely failed delete is surfaced by the `create_document` that follows (`?`) |
| `xerj-console-api/src/auth/store.rs:212` `delete_passkey` | **weaker → FIXED** (returned 204 while the credential kept authenticating) |
| `xerj-console-api/src/auth/sessions.rs:220` `bump_last_seen` | equivalent — documented best-effort `last_seen_at`; the re-read guard above it is what protects revocation |
| `xerj-console-api/src/auth/passkey.rs:113` uuid fallback | equivalent-in-effect — the WebAuthn user handle is not the authority; lookup is by credential id from our own store. Documented in-line |
| `xerj-console-api/src/auth/passkey.rs:115,211,226` display-name/name defaults | equivalent — absent optional value |
| `xerj-console-api/src/indices.rs:72,235` `let _ = s.add_field` | equivalent — hard-coded internal system-index schemas; `add_field` can only fail on a duplicate name, which is visible at compile time |
| `xerj-engine/src/rbac.rs:179,189` empty names/privileges | equivalent, and fail-**closed**: an empty grant is skipped and `roles.is_empty()` stays an accurate "no usable grant" signal. An unrecognised privilege string narrows the grant, never widens it |
| `xerj-engine/src/engine.rs:796` `let _ = effective_schema.add_field` | equivalent — guarded by an explicit `has_field` check on the line above |
| `xerj-api/src/es_compat.rs:1619` `merge_settings_defaults` | equivalent — response-shape normalisation for `GET /_settings`, no behaviour attached |
| `xerj-api/src/es_compat.rs:13509` `es_properties_to_schema` | equivalent — duplicate-name only; unknown mapper types are already rejected upstream by `validate_properties` at BOTH entry points (create :695, put_mapping :1761) |
| `xerj-api/src/es_compat.rs` `es_type_to_native` `_ => Object` | equivalent — same guard as above; the arm is only reachable for types that ARE supported and legitimately map to Object |
| `xerj-api/src/es_compat.rs:12181` scroll snapshot | equivalent — empty snapshot for a scroll with no captured state |
| `xerj-api/src/graph_api.rs:257` `drain_json`, `:453` detector label | equivalent — internal handler composition; documented default label |
| `xerj-api/src/binary_protocol.rs:150,157,169` | equivalent — serialising our own response types (cannot fail in practice) and the fallback is still an error frame |
| `xerj-api/src/extract.rs:155` | equivalent — the documented "absent body means defaults" extractor |
| `xerj-storage/src/index_store.rs:634` | **already correct** and is a model for this PR: present-but-unparseable `snapshot.json` is refused (`IncompatibleDataDir`), only a genuinely absent one yields empty. See its doc comment for the data-loss incident that produced it |
| `xerj-storage/src/backend.rs:151,338` parent-less prefix | equivalent — falls back to the root it is already scoped to; cannot escape |
| `xerj-storage/src/stored_codec.rs:301-305,514-516,2061` | equivalent — internal codec buffers writing into a `Vec` (infallible) with a byte-identical uncompressed fallback |
| `xerj-fts/src/index.rs:597` field-config default | equivalent **now** — the cause (an analyzer that failed to build) is what §3 of this PR rejects at create time; the open path warns once per index rather than once per field per flush |
| `xerj-fts/src/analyzer.rs:1100,1241` filter/rule arrays | equivalent — absent optional lists |
| `xerj-logs/src/ingest.rs:85` `from_timestamp().unwrap_or_default()` | equivalent in practice — only unreachable for years beyond ±262000; recorded rather than fixed |
| `xerj-logs/src/ingest.rs:197,199` level/message | equivalent — documented normalisation of absent optional fields |
| `xerj-ai/src/embedder.rs:54,111,137` | equivalent, fail-**closed** — publishes an `Err` terminal result so a panicking init cannot look like success |
| `xerj-ai/src/embedder.rs:200,372` mutex poison | equivalent — standard poison recovery |
| `xerj-server/src/main.rs:678,699,701,922,926,1460` | equivalent — crypto-provider install (idempotent), log-filter default, `--help`-style arg errors that `exit(2)`, worker-thread default |
| `xerj-engine/src/painless.rs:1091,1120,1127` `let _ = match_punct(';')` | equivalent — optional statement separator |
| `xerj-cluster/src/replication.rs:218,246` | equivalent — oneshot send to a dropped receiver; payload serialisation of our own entry type |
| `xerj-mcp/src/main.rs:72,546,920` | equivalent — `XERJ_URL` default, absent-params default, error-body text |
| `xerj-common/src/fsio.rs:64,72` | equivalent — temp-file cleanup on an already-failed path |
| `xerj-common/src/config.rs:1598` | equivalent, fail-**closed** — an invalid env override panics at startup with the key and value named |

### Found but NOT fixed (recorded so #204 can decide)

1. **`date` / `timestamp_parse` `formats` are chrono strftime, ES's are Java.**
   A real ES pipeline carrying `"formats": ["yyyy-MM-dd"]` is accepted, never
   matches, and silently falls through to the epoch/RFC-3339 heuristics. Same
   defect class; fixing it means implementing Java date-pattern translation.
2. **`PUT /{index}/_settings` accepts an `analysis` block and applies none of
   it.** Only `index.blocks.*` is forwarded to the engine; everything else
   updates a display-only map and answers `acknowledged: true`. ES rejects
   non-dynamic settings updates on an open index.
3. **`xerj-autoindex`** was audited but deliberately not touched — issues #239,
   #240, #241 and PRs #156/#162/#166/#167 are live in that crate right now and a
   concurrent diff there would collide. Its `let _ =` sites are extraction-path
   (a file that fails to parse is skipped and counted), not configuration.

xerj-org:
**Babysitter: CI is red with a genuine regression introduced by this PR — holding the merge.**

**Failure** (run 31469561172, ES-compat YAML conformance):

```
FAIL tests/es-compat-yaml/yaml/smoke/smoke_test_multinode/40_simulate.yml
     · Test verbose simulate with error in pipeline
     — match docs.0.processor_results.1.processor_type:
       expected String("rename"), got String("field_rename")
1365 passed · 1 failed · 3 skipped · 1369 total
```

**Root cause chain**

1. Before this PR, the ES→xerj translation for `rename` emitted `{"from","to"}`, which the `field_rename` plugin rejects — so **every** ES `rename` pipeline failed to compile, and `engine.put_pipeline(..., Some(body))` stored the **ES body verbatim** (`processors: [{"rename": …}]`). The simulate path then walked ES-shaped processors and everything matched.
2. This PR (correctly) fixes the translation to `{"mappings": {from: to}}` — the pipeline now compiles, so `put_ingest_pipeline` stores the **xerj-shaped config** instead: `stages: [{"type": "field_rename", "config": {"mappings": …}}]` (es_compat.rs, "compiled → store the translated config").
3. That exposes a latent defect in the two `stages` fallbacks that convert a stored pipeline back for simulate — `expand_pipeline_processors()` (used by `POST /_ingest/pipeline/_simulate` when a `pipeline` processor references a stored pipeline; the failing test) and `simulate_ingest_pipeline()` (`POST /_ingest/pipeline/{id}/_simulate`). Both emit the internal stage `type` verbatim as the ES `processor_type` ("field_rename" instead of "rename") and hand the xerj config shape to `apply_single_processor`, which only knows ES shapes — `field_rename`/`mappings` falls through to the forgiving `_ => Ok(())` arm, so the simulated rename silently no-ops instead of erroring (the test's later `status: "error"` assertion would fail too).

**Suggested fix** (not applied — substantive ES-compat semantics, outside the babysitter's mechanical-fix license): reverse-translate stages→ES processors in both fallbacks — `field_rename`→`rename` (one processor per `mappings` entry, `{"field": k, "target_field": v}`), `drop_field`→`remove` (`fields: [x]`→`field`), `timestamp_parse`→`date`, `json_parse`→`json`, `copy_field`→`copy`; `convert` needs no type reversal (the simulate interpreter accepts both `integer|long` and `float|double`). Alternatively, keep the original ES `processors` alongside the compiled shape for the simulate/GET surfaces. Either way, re-run the full ES-YAML suite — the gate is 0 failed.

xerj-org:
**Status from a board-clearing pass: this one is not mergeable yet, and the blocker is a real behaviour bug rather than staleness.**

**The hard gate is red.** `1365 passed · 1 failed · 3 skipped` — one case, and it is caused by this PR's own (correct) fix:

```
FAIL tests/es-compat-yaml/yaml/smoke/smoke_test_multinode/40_simulate.yml
  Test verbose simulate with error in pipeline
  match docs.0.processor_results.1.processor_type:
    expected String("rename"), got String("field_rename")
```

The translation fix in this PR is right — ES `rename` → xerj `field_rename` with `{"mappings": {from: to}}`, replacing a `{"from":..,"to":..}` shape the plugin always rejected. What it exposed is that **`_ingest/pipeline/_simulate?verbose` reports the internal processor name back to the client**. ES reports the ES-level name (`rename`); we now report `field_rename`, because the simulate path reads the type off the translated pipeline rather than the request. Before this PR the case passed only because the ES name and the stored name happened to agree.

So the remaining work is the inverse mapping: when building `processor_results`, translate the internal type back to the ES-facing name — the mirror of the `"rename" => …` / `"convert" => …` arms this PR adds. Worth checking whether `convert` has the same asymmetry, since it gains a vocabulary mapping here too, and whether any other renamed processor is echoed anywhere else in the simulate/GET responses.

**Second blocker: the branch no longer merges.** It is `CONFLICTING` against `main`. I attempted a rebase and stopped rather than force through it — the branch carries a merge commit (`7cb4375b`) plus 9 commits, and the conflict lands in `es_compat.rs`, exactly the file whose semantics this PR is changing. Resolving that under a rebase risks silently reintroducing the behaviour the sweep removes. The safer shape is to merge `origin/main` into the branch once and resolve `es_compat.rs` by hand, with the conformance suite re-run afterwards as the check. Left untouched — the branch is exactly as its author left it.

No other checks are failing (11 pass). Once the simulate `processor_type` mapping and the merge are done this should go green.

xerj-org:
Reviewed by two independent agent passes (Claude Opus 5), the second tasked with refuting the first. It could not knock down any of the four blockers and found five more defects of the same class. The direction here is right and the individual arguments are unusually well made — this is a needs-work on specifics, not on approach.

Lucene reaches the same conclusion this PR does, for what it is worth: it throws `IllegalArgumentException("Unknown parameters: " + args)` on any analysis-factory option it did not consume, rather than substituting a default.

---

## Blocking

**1. The ES-compat YAML gate is RED at head `18e47ad`, and this PR causes it.**

```
FAIL tests/es-compat-yaml/yaml/smoke/smoke_test_multinode/40_simulate.yml
  Test verbose simulate with error in pipeline
  match docs.0.processor_results.1.processor_type: expected String("rename"), got String("field_rename")
1365 passed · 1 failed · 3 skipped · 1369 total
```

The same job on `main` (`7bada00`) is green, so it is this branch. The mechanism: `simulate_ingest_pipeline` (`es_compat.rs:25707-25723`) rebuilds the ES processor shape from `type` with no inverse of `map_proc_name`. On main every ES `rename` failed to compile, so the ES body was stored and the `processors` branch was taken. This PR fixes the translation, `rename` now compiles, and simulate falls into the `stages` branch and reports the internal name.

The PR body states **`1366 passed · 0 failed · 3 skipped`** and its `### Refuted` section says "The branch did not fail the ES-compat YAML conformance gate." That contradicts the pushed head's own CI, which unfortunately means the body's other measurements need re-running before they can be relied on.

**2. `ignore_failure: false` is refused — and that is ES's own default.** In `build_plugin` the `UNHONOURED_PROCESSOR_KEYS` loop tests `config.get(key).is_some()` *before* `match stage_type`, so `{"set":{"field":"env","value":"prod","ignore_failure":false}}` returns `UnsupportedOption` and every write through the pipeline 400s. Specifying the default explicitly is exactly equivalent to omitting it, which is the "equivalent fallback is fine" rule #204 is built on. The new test `on_failure_and_ignore_failure_are_not_silently_dropped` only ever constructs `ignore_failure: true`, which is why it slipped through. The same function already models this correctly for `append`: `match config.get("allow_duplicates") { None | Some(Value::Bool(true)) => {} … }`.

**3. The analysis gate now runs on the index-OPEN path, which silently breaks existing indices.** `apply_settings` moved from `settings.pointer("/analysis")` to `analysis_block()`, which also resolves `/index/analysis`. `build_registry_from_settings` is called from both create (`index.rs:5949`) and open (`index.rs:6145`), and settings.json persists the `index` nesting verbatim. So an index created before this change with the canonical ES shape `{"settings":{"index":{"analysis":{...}}}}` was written with `standard`, and after upgrade reopens with the declared analyzers live. Query-time analysis stops matching the on-disk postings. Silently, with no version gate.

**4. CHANGELOG understates the blast radius.** The ingest change sits under `### Fixed` with no BREAKING marker while converting `201` into `400` for any pipeline using an ES processor xerj does not implement. The BREAKING block covers only create-time analysis 400s.

---

## Found during verification, same class as what the PR is fixing

**5. `remove` and `rename` bypass the PR's own new gate.** Those two branches build config from scratch (`json!({"fields":[s]})`, `json!({"mappings":{from:to}})`), so a processor-level `if` / `on_failure` / `ignore_failure` on them is dropped before `build_plugin` ever sees it. `{"remove":{"field":"secret","if":"ctx.tenant == 'a'"}}` compiles, is acknowledged, and drops the field on **every** document. That is precisely the "compiled implies honoured is a lie" defect this PR exists to eliminate, surviving inside its own gate — and it is self-inconsistent, because the identical `if` on `set` is refused. Conditional `remove`/`rename` is one of the commonest shapes in real Elastic pipelines.

**6. The boot-replay path turns working pipelines into write-refusing ones on restart alone.** `restore_cluster_state` (`engine.rs:2156-2179`) re-runs `compile_pipeline` over every persisted pipeline at startup, and a failure now lands in `unrunnable_pipelines`. So an unknown `grok` pattern name, an unknown `convert` target, a non-array `formats`/`types`, or `set.copy_from` flips previously-accepted pipelines from `201` to `400` with no PUT and no operator action — just a binary upgrade.

**7. `set.override` default flip reaches native pipelines too, not only the ES-translated path.** Native definitions are recompiled at every boot, so a `set` stage that has been *preserving* an existing field value for the life of a cluster starts *overwriting* it after upgrade. Listed only as an ES-parity fix under `### Fixed`.

**8. The `put_settings` analysis refusal misses the dotted spelling.** It tests `pointer("/analysis")` and `pointer("/index/analysis")` only. `{"index.analysis.analyzer.x.type":"custom"}` — which ES clients routinely send, and which this same handler supports for `blocks` — walks past the gate into `index_settings`, so `GET /_settings` echoes an analyzer that analyses nothing.

**9. `ignore_missing` and `ignore_empty_value` are still silently accepted** on `rename`/`remove`/`convert`/`date` and `set`. Neither honoured nor refused, and not listed among the known gaps.

---

## Checked and clear

Merges cleanly into `origin/main`, zero conflicts, not a draft. Touches no `Cargo.toml`/`Cargo.lock`, so the #352 tree-sitter link hazard does not apply here. Of the other open PRs only #356 and #322 share files; #322's diff has zero hits for `EsBulkItemResult`, `ReindexDest`, `put_pipeline`, `apply_settings` or `process_through_pipeline`, so the signature changes in this PR do not break the merged workspace.

Fix 1, 2 and 5 and re-run the conformance gate, and most of the value here can land. 3, 6 and 7 mainly need an explicit upgrade note and a version gate rather than a redesign.

PR #259, opened 2026-08-09
feat(autoindex): incremental --no-graph corpus reconciliation (supersedes #162)

xerj-org:
Supersedes #162 by @buger.

His work, our rebase. #162's head is on a fork with `maintainerCanModify=false`, so nobody in xerj-org could push the rebase or the review fixes to it — every attempt 403s. This branch carries his ten commits across onto current `main` with his authorship preserved (four `feat` commits and one `fix` are authored by him; the `Co-authored-by:` trailer on the new fix commit credits him for the code it modifies).

Do not close #162 by merging this — the maintainer closes it manually.

> ### ⚠️ Merge blocker outside this branch: `verification/cla-signed` is red
>
> The CLA gate landed after #162's last push, so #162 was never checked by it. Five of the ten commits here are authored by @buger, who is not in [`.contributors`](https://github.com/xerj-org/xerj/blob/main/.contributors), so cla-bot fails.
>
> **I did not and will not resolve this by adding his username myself.** The bot's own instructions say the signing pull request "is your signature, since it is made from your own account" — adding `buger` to `.contributors` from an xerj-org branch would forge that signature, and a CLA is a legal attestation only the contributor can make.
>
> Resolving it needs @buger to sign once (small PR adding his username, from his account), then `@cla-bot check` here. Everything else in this PR is ready.

## What it does

`xerj autoindex --no-graph` now **reconciles a changed folder instead of re-deriving it**.

A `--no-graph` run commits a durable *corpus generation*: a manifest of content groups (stable group identity, content digest and byte length, canonical path plus aliases, dataset assignment, validated output counts) plus a sealed source snapshot of the prepared records. A later run projects the current inventory onto the committed manifest and publishes only the difference. A run interrupted mid-generation replays from its own sealed snapshot, so the result never depends on whether the source tree changed after the crash.

Scope, plainly: **only `--no-graph`**. Graph-enabled runs are unchanged. Each changed generation still copies and prepares the full corpus (O(N) staging) — the win is in what gets *published and verified*, not yet in what gets read.

## The five review blockers, and what I did about each

**1. It reverted merged #160.** The branch carried `dimensions: usize` and `identity.dimensions == 0`, silently undoing the merged embedding-execution-identity fix. Fixed: `sync::ExecutionIdentity::embedding_dimension` is now `Option<usize>`, matching main's `esclient::EmbeddingExecutionIdentity::dimensions`. `None` is the honest record for a backend whose width the server does not pin (`neural`, `proxy`) and for the synthetic `disabled` identity a no-semantic plan pins. `validate_manifest` rejects `Some(0)` rather than requiring a positive number, so "unknown" and 384 stay different execution spaces instead of one being coerced into the other. The six duplicated #160 embedding commits were dropped in the rebase.

**2. It broke `xerj brain` self-heal.** The old gate refused `--fresh` whenever `preflight.plan.is_some()`, which is pinned on brain's recovery path (`brain.rs:283-298` sets `fresh = true` and re-runs `run_index_report` to converge a wiped data dir). Documented self-heal became a hard abort and `xerj brain --fresh` a guaranteed error.

Fixed by scoping the refusal to durable *generation* state only — a pending sync, or a committed manifest past genesis. That is **structurally** unreachable from a legacy journal, not merely untested: `apply_legacy_plan` (`state.rs:355-377`) sets `committed = None` whenever `bootstrap_legacy` returns `MigrationRequired`, which every non-empty legacy plan does, and an empty one bootstraps to generation 0 with no groups, which `genesis_recovery` filters out.

**3. Unversioned forced migration.** `bootstrap_legacy` pushed one identical `"legacy plan has no complete manifest group"` line *per file key* — an O(files) refusal string, megabytes on a large corpus, that read as a pile of per-file validation failures rather than the single migration decision it is. It is now stated once, with `sync::GENERATION_FORMAT_VERSION` in it and a followable rebuild command, and recorded as an explicit **Migration** entry in the CHANGELOG.

Approach borrowed from redb, which keeps a named, versioned `DatabaseError::UpgradeRequired(version)` distinct from `Corrupted` (`src/tree_store/page_store/header.rs:480-492`, `src/error.rs:212-213, 244-252`; Apache-2.0/MIT — approach only, no code taken).

**4. Dropped `no_graph` crash/resume coverage.** `failure_resume_http_tests` ran its whole injected-failure matrix with `no_graph: true`; since `--no-graph` now takes the generated executor, that module keeps the legacy graph-enabled transaction, which left the dangerous shape untested for no-graph. The incremental suite now injects a bulk that **applies half its items and then reports `errors: true`**, and asserts the generation stays uncommitted and the retry converges on exactly the new content.

**5. The 7-file rebase.** Done, onto current `main`. Adapted to main's `GroupSketch` struct (was a tuple), to `FileAssignment::as_document` from #173, and to #195's live-count verification. `reconcile_plan` now re-samples a demoted config file through `extract::extract_as_document`, since its frozen mapping describes the document shape and the family-extractor sketches would be the wrong thing to compare against it.

Also fixed from the 2026-08-04 review: the `--fresh` path used to build `UnsupportedInventoryDelta::default()` and print `Added content groups []; vanished content groups []` — arrays it never computed. Replaced by `UnsafeFreshGenerationError`, which names the durable state that is in the way and points at the plain re-run that *does* reconcile the change.

## Verification

Local gate on the rebased head:

```
cargo fmt --all -- --check                                  clean
cargo clippy -p xerj-autoindex --all-targets -- -D warnings  clean
cargo clippy -p xerj-server    --all-targets -- -D warnings  clean
cargo test -p xerj-autoindex        336 passed / 0 failed
cargo test -p xerj-server        26+3+3+5+2+1 passed / 0 failed
```

Live end-to-end against a real `xerj` server (release binary, private port, temp data dir) — not mocks:

| step | result |
|---|---|
| initial `--no-graph` index of 2 CSVs | `generation 1 committed — 2 datasets, 3 records live` |
| no-op re-run | still generation 1 — **no new generation appended** |
| change a row + add a row + delete a whole file | `generation 2 committed`; live count 3, `ALPHA-CHANGED`/`beta`/`gamma` present, the deleted file's record **gone** |
| `--fresh` on that generation | refused, exit 1, names `committed corpus generation 2`, journal byte-identical afterwards |
| `--fresh --json` | typed `xerj.autoindex.unsafe_fresh_generation.v1`, `blocking_state.generation = 2`, no empty delta arrays |
| **graph-enabled** index, then `--fresh` | **exit 0, converged** — the brain self-heal path (blocker 2) works |

## What I deliberately did NOT fix

- **A new file whose field shape needs a *new* dataset is refused**, e.g. adding `id,tag` to a corpus frozen on `id,value` gives `no frozen dataset accepts family csv … use a new prefix`. This is the designed frozen-identity contract and it is documented in the CHANGELOG. I verified live that the refusal is **transient, not wedging**: removing the offending file let generation 2 commit normally. Widening it is a design change, not a rebase fix.
- **O(N) staging remains.** `sync_executor` still copies every source file per changed generation. The CHANGELOG now says so plainly instead of the original body's "unchanged content is not re-extracted".
- **`--fresh` is refused on a state directory a prior `--no-graph` run generated, even for a graph-enabled run.** `default_state_dir` keys on (root, url, prefix) and both `autoindex` and `brain` default to prefix `ax`, so a user who ran `autoindex <folder> --no-graph` and then `brain <folder> --fresh` on the same root/url hits this. Refusing is correct — there is a real generation to destroy — and the message names both recovery routes, but it is a behavioural edge worth knowing.
- **Not addressed from the older review:** `run_id` uniqueness (`state.rs`) is closed only in-process and wraps at 65536.

## Second review round — three defects on sibling paths, fixed

An adversarial review confirmed the five blocker fixes above but found three
further defects on paths this PR never exercised. All three were reproduced by
running code, and each fix landed with the regression test that fails without it.

**1. `--dry-run` was accepted, silently ignored, and mutated the destination.**
`cfg.dry_run` was consulted only *after* the pending-replay and reconcile
branches had already published and returned. Reproduced: gen-1 run, then a
`dry_run = true` run issued a real data bulk (`data_bulk_requests` 1 → 2),
published the changed record, appended `sync_commit` #2 and exited 0. It is now
decided before either branch, where the journal is still only preflighted and
`gc_snapshots` has not run; it prints the plan a real run would act on and
returns `Ok((0, None))`. The printed plan comes from the same
`project_reconcile_plan` the real path uses, so the two cannot drift.

**2. `--no-graph` hard-failed on any folder containing one junk/skipped file.**
`create_prepared_snapshot` iterates every inventory file, but `prepare_artifact`
demanded a `plan.files` assignment — and a junk file lives in `plan.junk_files`,
never in `plan.files`, by construction. Reproduced: `a.csv` plus a zero-byte
`empty.csv` made the *first* run return
`Err("plan has no assignment for empty.csv")`. Content absent from `plan.files`
is now snapshotted as a blob with `prepared: None`.

**3. A junk record aborted the whole generation.** `ensure!(stats.junk == 0)`
killed the run while the legacy path merely accumulates it — contradicting
`cli.rs` and `landing/llms-full.txt`, which both say *"3 completed-with-junk
(junk recorded, never fatal)"*. Reproduced with a CLF `access.log` whose first
line is `!!! rotated by logrotate, not a log line`. Junk is now counted:
`PreparedArtifact::junk` seals it, `ManifestGroup::expected_junk_records` carries
it into the generation, and the catalog reports per-file `junk` plus a run-level
`junk_records_total` summed from the manifest instead of the hardcoded
`junk_records: 0`. Totals are summed over groups, not datasets, so a file feeding
several datasets is not double counted and one feeding none is not lost.

**4. Exit code 3 was lost for `--no-graph`.** Every generated-path return was
`Ok((0, Some(summary)))`. With (2) and (3) fixed that state is reachable, so
`generated_exit_code` derives it from the committed generation's own run
document — which also makes a no-op re-run over a corpus that still holds junk
report 3 again instead of flapping to 0.

Local gate on the repaired head: `cargo fmt --all -- --check` clean;
`cargo clippy -p xerj-autoindex --all-targets -D warnings` clean;
`cargo clippy -p xerj-server --all-targets -D warnings` clean;
`cargo test -p xerj-autoindex` **336 passed / 0 failed** (was 333 — three new
regression tests); `cargo test -p xerj-server` 26+3+3+5+2+1 passed / 0 failed.

Known and deliberately not changed: `--fresh --dry-run` on a *legacy* journal
still deletes `journal.ndjson` before the dry-run check. That is pre-existing on
`main` (`Journal::open` at `lib.rs:944` runs before `cfg.dry_run` at
`lib.rs:1168`), not introduced here, and fixing it is a separate change.

## Notes

- No `Fixes #N`: the triage found no open issue this closes. #238 (added-but-skipped files leave immortal catalog entries) covers the *addition* side, which this does not fix, so it is deliberately not linked.
- CI on #162 was stale-green — it ran 2026-08-04 against a `main` ~100 commits behind and never built the merged tree. Treat this branch's run as the first real signal.


xerj-org:
Second-review repairs pushed as `a02f905b`. CI on that exact head, run [31349184144](https://github.com/xerj-org/xerj/actions/runs/31349184144):

| check | result |
|---|---|
| Build + Test | **pass** (1h08m45s) |
| ES-compat YAML conformance | pass |
| Use-case harnesses (brain + MCP + autoindex) | pass |
| Format + Clippy | pass |
| ONNX feature compile + tests | pass |
| API smoke + liveness + benchmark | pass |
| Autoindex FD smoke — ubuntu / macOS / windows | pass / pass / pass |
| UX pipeline tests | pass |
| CLA gate config | pass |
| `verification/cla-signed` | **fail — unchanged, and not fixable from this branch** |

The CLA gate is the one remaining blocker and it is the same one described at the top of the PR body: five of the eleven commits are authored by @buger, who is not in [`.contributors`](https://github.com/xerj-org/xerj/blob/main/.contributors). Adding his username from an xerj-org branch would forge a signature that only he can make, so it stays red until he signs from his own account and someone comments `@cla-bot check` here. Nothing in the code needs to change for it.

xerj-org:
@cla-bot check

PR #261, opened 2026-08-09
feat(security): make the cargo-audit and cargo-fuzz claims real, and fix what they found (#207)

### LINKED ISSUE #207: Public claims contradicted by the source: cargo-audit and fuzzing are documented but not in CI
A source review turned up several public claims in this repo that its own source contradicts. The project's stated rule is that every public number traces to a verified run, so these should be fixed or removed before GA.

**Security tooling that does not exist.** `user-feedback/09-security/cves-and-vulnerabilities.md:67-68` claims `cargo-audit` and "fuzz testing (cargo-fuzz) on all input parsing paths". Neither is in CI. `.github/workflows/ci.yml` runs lint, build-test, onnx, api-checks, conformance, ux, usecase and fd-smoke. There is no audit job, no fuzz job, and no fuzz target directory in the tree.

This is the worst one because it is a security claim. Someone evaluating XERJ against the CVE history of the incumbent reads that line and believes input parsers are fuzzed. They are not.

**Settings count is self-contradictory.** The config doc header says 38 settings, a mid-file comment says 56 fields, and the count test asserts 60 (`xerj-common/src/config.rs:49-92`, `xerj-engine/tests/product_experience.rs:780-784`). The test asserts a number the code no longer matches, so it passes tautologically rather than checking anything.

**Wire compatibility is not proven the way the test claims.** `xerj-engine/tests/product_experience.rs:429-555` is described as asserting identical curl contracts. It exercises the Rust engine API directly (`engine.create_index`, `idx.index_document`, `process_bulk`) and never makes an HTTP request. The ES response shapes it checks are real and worth having; the "same curl commands" framing is not what the test does.

**OpenAPI spec declares 1.0 while covering 7 routes of roughly 210.** `landing/openapi.json`. Either label it partial or generate it.

**Merge I/O rate limiting is claimed but dormant.** The `RateLimiter` and `io_rate_mb_per_sec` appear in `xerj-storage/src/merge.rs` inside a `MergeExecutor` that the production merge path does not use. Background merges are not rate limited in the code that actually runs.

None of these are hard to fix. Most are a line of text. The fuzz claim needs either a CI job or the sentence removed, and removing it is fine. The point is that a project whose pitch rests on honest measurement cannot ship claims its own repo refutes.

Found while reviewing XERJ's source against the user-feedback corpus.

xerj-org:
Fixes #207

The security docs answered Elasticsearch's CVE history with "`cargo-audit` for
dependency vulnerability scanning" and "fuzz testing (cargo-fuzz) on all input
parsing paths". Neither existed — no audit job, no fuzz job, no fuzz target
anywhere in the tree. Both are real now, and turning them on found four ways to
kill the process from an unauthenticated request.

## Reproduced first, at 12d5cc32

| claim | evidence at HEAD | after |
|---|---|---|
| cargo-audit runs | `cargo audit -f engine/Cargo.lock` → **3 vulnerabilities, exit 1** | **0 vulnerabilities, exit 0** |
| fuzzing runs | no `engine/fuzz`, no job in `ci.yml` | 7 harnesses, `fuzz` job, `fuzz-smoke.sh` ran all 7 locally, exit 0 |
| settings count | `config.rs` header 38, mid-file 56, one count test 60, another 61 — both by re-adding a constant | **104**, measured by serialising `Config::default()` (103 when this was written; rc.14 added one key — see "Merged main" below) |
| merge throttle | `run_merge_once` never reads `io_rate_mb_per_sec` | named as dormant in code, TOML, docs site, and at startup |
| wire-compat test | `journey_es_migration` makes no HTTP request | comment says what it does; points at the YAML conformance suite |
| OpenAPI 1.0 | 7 paths vs **212** `.route(` registrations in `router.rs` (216 after rc.14 — the spec and the CHANGELOG say "200-plus" rather than a number that drifts) | titled and versioned as a partial spec |

## cargo-audit

New CI job `security-audit`. A RUSTSEC *vulnerability* fails the build;
informational advisories (unmaintained / unsound / yanked) are printed and do
not — they live in transitive crates we cannot upgrade unilaterally, and a gate
that is red for reasons nobody can act on gets ignored, then disabled.

The first run was red:

* `quick-xml` 0.36.2 in `xerj-autoindex` — **RUSTSEC-2026-0195** (unbounded
  namespace-declaration allocation → memory-exhaustion DoS, CVSS 7.5) and
  **RUSTSEC-2026-0194** (quadratic run time on duplicate attribute names, CVSS
  7.5). That parser reads `.docx` / `.xlsx` / `.xml` belonging to whoever owns
  the folder being indexed. → 0.41, which also de-duplicates the crate (0.41 was
  already present via `pdf_oxide`).
* `crossbeam-epoch` 0.9.18 → 0.9.20 (**RUSTSEC-2026-0204**).

quick-xml 0.38 changed how character data arrives — entity references are their
own `Event::GeneralRef` rather than being folded into the neighbouring
`Event::Text` — so a reader that ignored the new event would silently drop every
`&`, `<`, `>` and `&#233;` while the words around them still arrived.
`xml_x.rs` / `docx.rs` handle it, reassemble a value's fragments before emitting,
and bound the accumulator at `MAX_LINE`. One case improves on 0.36: an
undeclared entity used to fail `unescape()` for the whole node and
`unwrap_or_default()` discarded it; now only the reference is lost.

## cargo-fuzz

New CI job `fuzz`, wrapping `.github/scripts/fuzz-smoke.sh` so a contributor
runs exactly what CI runs. Seven harnesses over the parsers an unauthenticated
request reaches: ES query DSL, Lucene `query_string`, date math + date formats,
index-name date math, SQL, Painless, and the aggregation engine's two separate
script tokenisers. Targets come from `cargo fuzz list`, so a
new harness cannot be silently left out; a target with **no seeds fails the
script** rather than passing by exploring from the empty input.

## The four aborts it found

The release profile sets `panic = "abort"`, so none of these was a 400 — each
was the process. Every one has a regression test that **aborts the test binary
before the fix**; I verified that by reverting each fix in place and running it:

```
TimeDelta::hours out of bounds                                          (dates.rs)
end byte index 18 is not a char boundary; it is inside '٠'              (parser.rs)
end byte index 24 is not a char boundary; it is inside 'ʋ'              (painless.rs)
byte range starts at 2 but ends at 0                                    (index.rs)
```

1. **`{"range":{"ts":{"gte":"now+33333333333333H"}}}`** — 20 s into the
   `date_math` target's first run. `chrono`'s `Duration::hours` and siblings
   **panic** on a count that does not fit a `TimeDelta`, and they run *before*
   the `checked_add_signed` that was supposed to make `add_unit` total. Fixed
   with the fallible `try_*` constructors. Grepping the pattern found **three** more
   copies in the index-name resolvers — `index.rs`, and *both* branches of
   `es_compat.rs` (`<logs-{now+9999999999999d}>` and the anchored
   `<logs-{2026-01-01||+9999999999999d}>`, handled by two different functions
   sixty lines apart). All three are fixed and their `n * 30` / `n * 365` are
   checked too.
2. **`query_string: "_٠"`** (also `?q=`). The tokeniser walks bytes but slices
   `&q[start..i]` as a `&str`, and broke on `is_whitespace()` — true for
   Latin-1 NBSP (0xA0), which in valid UTF-8 is only ever a *continuation* byte.
   Every delimiter test is `is_ascii_*` now, which makes "the scan can only stop
   on a character boundary" true by construction.
3. **A Painless script with one non-ASCII character.** Same defect in the script
   tokeniser, found in under a second. Identifier scanning is ASCII-only now;
   string literals still carry any Unicode.
4. **An index name of `<{}{}>`.** First `{`, last `}` — when the `}` comes
   first, `date_part[2..0]` panics. Both resolvers now require the closing brace
   to come after the opening one, and read the unit as one *character*.

**Approach reference** (no code taken — ES is AGPL/SSPL/Elastic-2.0,
approach-only here): `JavaDateMathParser.java:120-192` calls
`dateTime.plusHours(sign * num)` unguarded and lets `java.time` throw. On a JVM
that overflow is a catchable exception and the request becomes a parse error; in
a Rust binary with `panic = "abort"` the identical shape is a process kill. Same
algorithm, different failure mode — which is why the `try_*` constructors are
load-bearing here and absent there.

## The other claims in the issue

* **Settings count → 104, measured.** Two count tests were re-adding hardcoded
  sums (`38 == 38`, `61 == 61`) and could not fail whatever `Config` held. Both
  now serialise `Config::default()` and count leaf keys, section by section.
  38 / 50 / 56 / 60 / 61 / "<50" corrected across `config.rs`,
  `xerj-common/src/lib.rs`, `engine/README.md`, `xerj.default.toml`,
  `engine/tests/battle_sla.sh` and the shipped feedback responses. Closes item
  10 of `ROADMAP-TO-GA.md`.
* **`xerj.default.toml` disagreed with the defaults it documents** in four
  places (`max_segment_mb` 5120 vs 8192, `wal_max_size_mb` 512 vs 1024,
  `flush_size_mb` 256 vs 512, `default_quantization` `"scalar8"` vs `"none"`),
  so `cp xerj.default.toml xerj.toml` silently changed four behaviours including
  switching on 8-bit vector quantization. A new test diffs every leaf key
  against `Config::default()`. **The docs site had drifted the same way** and
  is corrected in the same commit range — see "Gate round 2" below for exactly
  which pages were still wrong and what now stops them drifting again.
* **Merge rate limiting is dormant — confirmed, and worse than reported.**
  `run_merge_once` reads only `max_segment_mb`, `tier_floor_mb`,
  `min_merge_count`, `max_merge_count`. `strategy` was unread, so
  `log_structured` ran size-tiered merging silently — now refused at startup
  like `storage.backend`. `io_rate_mb_per_sec`, `min_segments` and
  `max_concurrent` stay accepted but `MergeConfig::dormant_overrides` names any
  the operator moved off its default and the server logs each at WARN.
  **Judgement call worth a second opinion:** hard-erroring instead would refuse
  to start on `io_rate_mb_per_sec = 100`, which this project has shipped in
  `xerj.default.toml` since v0.1. Wiring a real throttle into `run_merge_once`
  is the fix that removes the entry.
* **Wire-compat framing** and **OpenAPI partial-spec labelling** as described in
  the issue.

`engine/crates/xerj-engine/tests/security_tooling_claims.rs` is the guard: it
fails if a documented tool stops running in CI, if a fuzz target ships without
seeds, or if the "all input parsing paths" overclaim reappears in any prose file
in the repo. All four of its tests fail at 12d5cc32 — verified by restoring the
old `ci.yml`/docs and moving `engine/fuzz` aside.

## Gate

```
cargo fmt --all -- --check                                  clean
cargo +nightly fmt --manifest-path fuzz/Cargo.toml --check  clean
cargo clippy -p xerj-{common,server,engine,api,autoindex,query} --all-targets -- -D warnings   clean
cargo test -p xerj-{common,query,server,engine,api,autoindex}   all green
cargo audit (branch lockfile)                               0 vulnerabilities, exit 0
FUZZ_SECONDS=25 bash .github/scripts/fuzz-smoke.sh          all 7 targets survived, exit 0
```

Two notes for the reviewer:

* `xerj-engine --test painless_script_limits` overflows the stack in the **debug**
  profile (`rescore_script_over_call_depth_reports_a_failure`). Pre-existing and
  unrelated: 10/10 pass under `--release`, which is what CI runs, and 9/9 pass in
  debug with `RUST_MIN_STACK=16M`. The suite's own test names
  ("…fits_a_2mib_stack") show the depth budget is calibrated for release frames.
* The `fuzz` job builds `xerj-engine` with ASan on a 2-core runner and will not
  be quick. That is the price of the claim being true; tuning its budget or
  moving it to push-only is an owner call.

---

## Review repairs (2026-08-10, commit `69d7170c`)

An adversarial review executed tests against this branch's HEAD and found the
crash fix incomplete on two sibling paths. Both are fixed here; each repair
carries a test that aborts when only its source hunk is reverted.

1. **`es_compat::apply_date_math_tail` was a third copy of the same defect** —
   the *anchored* branch (`<logs-{2026-01-01||+1d}>`) of the same resolver,
   sixty lines above the `now` branch that was patched. Eight panics through the
   public `resolve_date_math_index`, which runs on the index path parameter
   (`es_compat.rs:671`), alias resolution (`:825`) and the `_search` index list
   (`:6763`): `TimeDelta::days out of bounds`, `` `DateTime + TimeDelta`
   overflowed ``, and `byte index 1 is not a char boundary` on a multi-byte
   unit. Fixed the three ways `resolve_now_expr` was — one *character* for the
   unit, `Duration::try_*` + `checked_mul`, `checked_add_signed`.
   Test: `crates/xerj-api/tests/index_name_date_math_is_total.rs`, at the crate
   boundary because a same-file fix is what missed this.

2. **The `index_name_date_math` harness fuzzed a function nothing calls.**
   `xerj_engine::index::resolve_date_math` has no production callers — `grep`
   across `crates/` finds only `lib.rs`'s re-export, its own `#[cfg(test)]`
   module and the harness. That is precisely why the harness missed (1). The
   target now drives `xerj_api::es_compat::resolve_date_math_index` first (with
   `xerj-api` added to `engine/fuzz/Cargo.toml`), keeps the engine copy on the
   same input, and gains four seeds for the anchored form.

3. **`aggs::lex_script` and `aggs::tokenize_script` had the identical
   char-boundary defect** — two more Painless-subset tokenisers, reached from an
   ordinary `_search` body: `&src[i..i + 2]` at a byte index that can sit inside
   a multi-byte character, plus `is_whitespace()` on a raw byte.
   `{"m":{"scripted_metric":{"map_script":"中"}}}` was 16 distinct panics through
   `run_aggs`. Both now gate the two-character probe on both bytes being ASCII
   and use `is_ascii_whitespace`, and a new **`agg_script`** fuzz target covers
   all three entry points.
   Test: `crates/xerj-engine/tests/agg_script_tokenisers_are_total.rs`.

4. **Docs corrected to match the code**: the CHANGELOG's "two more copies … both
   are fixed too" (false — a third survived); `landing/docs/config.html:165`,
   which still told operators to switch to `log_structured` while
   `Config::validate` now refuses to start on it; the muddled `strategy` /
   dormant-keys sentence in `landing/docs/storage.html`; the `cargo-audit`
   bullet in `cves-and-vulnerabilities.md` that ended mid-sentence, and the
   final bullet that ended without a full stop; and the "Painless scripts"
   coverage claim in `SECURITY.md` + that file, which now names
   `xerj_engine::painless` *and* the two aggregation tokenisers.

**Judgement call kept, not changed:** `merge.strategy = "log_structured"` stays
a hard startup error rather than being downgraded to a WARN. `xerj.default.toml`
ships `strategy = "size_tiered"`, so no config this project handed out trips it,
and the alternative is silently running size-tiered merging for an operator who
picked a levelled policy for its read amplification. The docs now say so.

Gate re-run on the repair commit: `fmt` clean (both toolchains), `clippy -D
warnings` clean on `xerj-engine` + `xerj-api`, `cargo test --release -p
xerj-engine -p xerj-api` 39 suites / 775 passed / 0 failed, `cargo +nightly fuzz
build` clean, `FUZZ_SECONDS=25 fuzz-smoke.sh` all 7 targets survived.

---

## Gate round 2 (2026-08-10, commit `574b66b8`)

The merge gate measured this PR's own release note and found it false, which is
the one thing this PR cannot ship: **"The docs site carried the same four stale
defaults and is corrected too"** was true for one key out of four. Only the
DEFAULT table on `landing/docs/config.html` had been corrected. The
copy-pasteable blocks below it — the part a reader actually pastes — still said:

All line numbers in this table are at `69d7170c`, the commit the gate measured
(the rc.14 merge below moved `config.rs`; the defaults themselves did not
change):

| shipped docs | said | real default (`config.rs`) |
|---|---|---|
| `landing/docs/config.html:160` | `wal_max_size_mb = 512` | `1024` (`:641`) |
| `landing/docs/config.html:161` | `flush_size_mb = 256` | `512` (`:642`) |
| `landing/docs/config.html:195` | `default_quantization = "scalar8"` | `"none"` (`:917`) |

So the page contradicted itself on three keys, and pasting its `[vector]` block
switched on 8-bit quantization with its 1–2% recall loss — verbatim the harm
this PR reports for `xerj.default.toml`.

Grepping the four keys rather than working from the gate's list found two more
copies, one of them in a file this PR already edits:

* `landing/docs/storage.html:145` — WAL "(default **512 MiB**)", real 1024.
* `landing/docs/vectors.html:153` — `scalar8` "is the default", real `"none"`.
* `engine/README.md:239` — `flush_size_mb` "(default 256 MiB)", real 512.

`vectors.html` also sold a **`scalar4`** mode at "~8× smaller".
`VectorQuantization` is `None | Scalar8 | Binary`; the 4-bit quantizer in
`xerj-vector` is reachable from neither config (no such variant — the TOML
fails to parse) nor mapping (`es_compat.rs:13664` maps only
`int8_hnsw`/`int8_flat` → `scalar8`). The page says that now, and so does the
duplicated docs search-index blob in all 44 landing pages, which was still
listing `scalar4` as a selectable enum value.

### What stops it happening again

`shipped_default_config_documents_the_real_defaults` only reads the TOML —
which is precisely why this drift survived a review round. New guard, in the
same module:

`xerj-common::config::tests::the_docs_site_config_page_agrees_with_the_real_defaults`
reads `landing/docs/config.html` and diffs **both halves** of it against
`Config::default()`:

1. every `DEFAULT` cell in the reference table, and
2. every assignment in every `<pre class="code">` example block.

An example line that deliberately differs — the blocks whose whole point is
switching TLS or cluster mode *on* — must carry `# not a default` in its
comment, so the reader is told as well as the test. Twelve lines are annotated (ten from this round, plus the two rc.14 `[server]` lines the merge brought in).

Verified it fails on the defect rather than passing vacuously. Restoring only
`default_quantization = "scalar8"`:

```
example: [vector] default_quantization = "scalar8" contradicts the real default
"none" and does not say so — add `# not a default ("none")` to the line
```

and flipping only the table cell for `wal_max_size_mb` to 512:

```
table: [storage] wal_max_size_mb says 512, code says 1024
```

### Known gap, disclosed rather than fixed

`aggs::parse_time_zone_offset` (`engine/crates/xerj-engine/src/aggs.rs:5698`)
has the **same char-boundary defect** this PR fixed in the three tokenisers:
`rest.len() == 4` is a *byte* length and `rest[..2]` can land inside a
character. A `date_histogram` with `"time_zone": "+中a"` in an unauthenticated
`_search` body dies with `end byte index 2 is not a char boundary; it is inside
'中'` — and with `panic = "abort"` in the release profile that is the process,
not a 400.

It is **pre-existing on `main`, outside this diff, and not claimed by this PR**:
`SECURITY.md` and `cves-and-vulnerabilities.md` already say the seven harnesses
are the coverage and "not every parser in the engine". Rather than quietly let
"seven harnesses" read as completeness, it is now:

* filed as **[#272](https://github.com/xerj-org/xerj/issues/272)** with the
  repro, the one-line fix in the shape this PR used elsewhere
  (`rest.len() == 4 && rest.bytes().all(|b| b.is_ascii_digit())`, which makes
  the boundary true by construction and changes no input that parsed before),
  and a proposal for a harness over the aggregation date parsers; and
* stated as a **known gap in the CHANGELOG**, under the fuzz-coverage bullet.

A second one, also pre-existing and not touched here, now measured and filed as
**[#275](https://github.com/xerj-org/xerj/issues/275)** — see "Gate round 3"
below. `es_compat.rs` matches `"scalar8" | "int8" | "none"` on a mapping's
`quantization` and falls through `_ => {}`, so `"quantization": "scalar4"` is
accepted **and echoed back by `GET /_mapping`** while the field is stored at
full precision. Issue #204's bug class, in the same area this round corrected
the docs for.

### Gate

```
cargo fmt --all -- --check                                        clean
cargo clippy -p xerj-common --all-targets -- -D warnings          clean
cargo test -p xerj-common                            62 passed / 0 failed (+3 doc-tests)
cargo test -p xerj-engine --test security_tooling_claims   4 passed / 0 failed
```

The prose scanner is re-run because this round edits prose. Nothing outside
`landing/`, `CHANGELOG.md`, `engine/README.md` and one `#[cfg(test)]` module in
`xerj-common` changed, so the engine behaviour CI measured at `69d7170c` is
unchanged.

---

## Merged main (v1.0.0-rc.14), 2026-08-10

GitHub flipped the PR to `CONFLICTING` when rc.14 landed. Resolved on the
branch; four files needed hands:

* **`landing/docs/config.html`** — kept this branch's corrected key table and
  spliced in rc.14's `bind_address` default (`"0.0.0.0"` → `"127.0.0.1"`) and
  its new `allow_insecure_network_bind` row. rc.14's `[server]` example block
  is kept as written, with the `# not a default` annotations the new guard
  requires, so `bind_address = "0.0.0.0"` and
  `allow_insecure_network_bind = true` now say they are demonstrating an
  override rather than quoting a default.
* **`engine/crates/xerj-common/src/config.rs`** — kept this branch's measured
  `count_user_facing_settings` over main's `assert_eq!(total, 62)` hardcoded
  sum. That test then earned its keep immediately: it failed on the merge and
  named the drift.
* **`engine/crates/xerj-server/src/main.rs`** — both sides added a startup
  block in the same place; kept both.
* **`CHANGELOG.md`** — this branch's entries stay under `[Unreleased]`, rc.14's
  section follows.

**The settings count is now 104, not 103** — rc.14's
`server.allow_insecure_network_bind` is a real new leaf key, and the count is
measured, so it moved. Swept everywhere it is quoted: `config.rs` (header,
sub-config banner, `SETTINGS_BY_SECTION` server 6 → 7, the total assertion),
`xerj-common/src/lib.rs`, `engine/README.md`, `xerj.default.toml`'s header
("sets 54 of XERJ's 104"), `engine/tests/battle_sla.sh`, `EXPECTED_SETTINGS` in
`product_experience.rs`, the four feedback responses and `ROADMAP-TO-GA.md`.
The dated notes in `ROADMAP-TO-GA.md` and the CHANGELOG say 103 was the figure
when #207 was done and why it is 104 now, instead of rewriting the record.

Gate re-run after the merge:

```
cargo fmt --all -- --check                                        clean
cargo clippy -p xerj-common -p xerj-server --all-targets -- -D warnings   clean
cargo test -p xerj-common                            77 passed / 0 failed (+3 doc-tests)
cargo test -p xerj-server                            see PR checks
cargo test -p xerj-engine --test product_experience --test security_tooling_claims
```

---

## Gate round 3 (2026-08-10)

Two things: the branch went `CONFLICTING` again, and one sentence the round-2
docs sweep produced was still doing the thing this PR is about.

### Merged main twice

`main` moved twice while this round ran, and both times the only conflict was
`CHANGELOG.md`, because both sides append to `[Unreleased]`.

1. **past rc.14** — autoindex PDF parse-once, `.contributors`, release metrics.
   Main's `autoindex` entry was folded into this branch's existing
   `### Changed` rather than left as a second `### Changed` heading under one
   release.
2. **the autoindex rerun-refusal wave** (`#254` and the `#238`/`#240`/`#241`
   follow-ups) — main's `### Changed — autoindex refuses reruns…` section kept
   as its own heading, placed before this branch's `### Fixed` so the
   Changed-type sections stay together.

No conflict in code either time, but two seams needed testing rather than
assuming: `xerj-autoindex` (main's work lands on this branch's `quick-xml`
0.36 → 0.41 advisory upgrade) and `xerj-engine/src/index.rs` (main's autoindex
wave touched a file this branch patched for date math). Both are in the gate.

### The `scalar4` sentence implied a guard that does not exist

Round 2 rewrote `landing/docs/vectors.html` to:

> A 4-bit quantizer exists in the vector crate but no config or mapping value
> reaches it, so `scalar4` is not a mode you can select in this build.

Every clause is true and the sentence still misleads: a reader takes "not a
mode you can select" to mean the value is rejected. Measured through the real
ES-compat router at this head — `PUT /<index>` with
`{"properties":{"v":{"type":"dense_vector","dims":4,"quantization":"<q>"}}}`:

| `quantization` | create | `GET /_mapping` returns | actually stored |
|---|---|---|---|
| `scalar4` | `200 acknowledged` | `"quantization":"scalar4"` | f32, unquantized |
| `sq8` (typo) | `200 acknowledged` | `"quantization":"sq8"` | f32, unquantized |
| `binary` | `200 acknowledged` | `"quantization":"binary"` | f32, unquantized |
| `int4` | `200 acknowledged` | `"quantization":"int4"` | f32, unquantized |
| `scalar8` | `200 acknowledged` | `"quantization":"scalar8"` | SQ8 |

The echo is what makes it worse than a no-op: the operator asks for
quantization, gets a 200, reads the mapping back to confirm, and sees the value
they wrote. Someone sizing on a 4× memory reduction gets 1× and no signal.
Startup refuses an unimplemented `[vector] default_quantization` — that guard
is #207's — and the per-field mapping key has none.

It is **pre-existing on `main` and outside this diff**, so it is disclosed, not
fixed here (a `400` on a previously-accepted mapping value is a wire-behaviour
change that wants its own PR and its own conformance run):

* filed as **[#275](https://github.com/xerj-org/xerj/issues/275)** with the
  table, the repro and a proposed `400` — scoped to the XERJ-specific
  `quantization` key, because `index_options.type` families outside
  `int8_hnsw`/`int8_flat` (`hnsw`, `flat`, `bbq_*`, `int4_*`) are
  *deliberately* accepted-and-served-exact so a real Elasticsearch mapping is
  not rejected, and the fix has to distinguish the two;
* stated on the page, including that reading the mapping back is not a way to
  confirm quantization took; and
* stated in the CHANGELOG under the docs-drift entry — which is also where the
  new guard's honest scope now lives: `the_docs_site_config_page_agrees_with_the_real_defaults`
  diffs `config.html` against `Config::default()`, so it covers the **config**
  half and nothing checks a mapping.

Two known gaps are now disclosed rather than implied: this one and
[#272](https://github.com/xerj-org/xerj/issues/272) (`parse_time_zone_offset`).

### Gate

Re-run in the worktree after both merges. CI was green on all 13 checks at
`8a2cf833`, including ES-compat YAML conformance; everything this round adds on
top is prose, so the local run targets the **merge seams** and the claim
scanners.

```
cargo fmt --all -- --check                                      clean
cargo clippy -p xerj-common -p xerj-query -p xerj-engine \
             -p xerj-api -p xerj-autoindex -p xerj-server \
             --all-targets -- -D warnings                       clean
cargo test -p xerj-common                        77 passed / 0 failed
cargo test -p xerj-engine --test security_tooling_claims \
           --test product_experience \
           --test agg_script_tokenisers_are_total  16 passed / 0 failed
cargo test -p xerj-autoindex                    338 passed / 0 failed
cargo test -p xerj-api -p xerj-query            465 passed / 0 failed
cargo test -p xerj-engine                       741 passed / 0 failed
```

Re-run after the second merge, at the pushed head:

```
cargo fmt --all -- --check                                      clean
cargo test -p xerj-engine -p xerj-autoindex -p xerj-common
                                               1182 passed / 0 failed  (exit 0)
```

The four keys the gate measured were re-checked directly, not just via the new
guard: `wal_max_size_mb`, `flush_size_mb`, `max_segment_mb` and
`default_quantization` now agree with `Config::default()` in every assignment
and every prose "default" in `config.html`, `storage.html`, `vectors.html`,
`engine/README.md` and `xerj.default.toml`. Seven fuzz targets, each with a
seed corpus (`agg_script` 9, `date_math` 10, `index_name_date_math` 15,
`painless` 10, `query_dsl` 11, `query_string` 12, `sql` 7).

**One local-environment caveat, stated because it looks like a failure and is
not one.** `cargo test -p xerj-engine --test painless_script_limits` aborts in
this sandbox with `has overflowed its stack` on whichever call-depth test the
scheduler reaches first. It is **not** from this PR: the identical abort
reproduces with `main`'s `painless.rs` swapped in (this PR's change to that
file is the ASCII tokeniser gate, which cannot affect recursion depth), and the
test file is untouched by this diff. The tests deliberately recurse until the
interpreter's call-depth limit trips, and a debug build's frames do not fit
tokio's 2 MiB worker stack here. With `RUST_MIN_STACK=16777216` all 9 pass, and
CI passes them unset. The 741 figure above is the run with the larger stack.

### Not re-run locally

The **ES-compat YAML conformance suite**. This round changes no query, mapping
or search semantics — the diff since the CI-green head is two `main` merges
plus prose in `CHANGELOG.md` and `landing/docs/vectors.html`. The suite is a
required check and runs on this push.


xerj-org:
All six review findings are addressed in `69d7170c`. The two crash findings were real and I reproduced both before changing anything.

| # | finding | disposition |
|---|---|---|
| 1 | `apply_date_math_tail` is a third unpatched copy | **fixed** — one *character* for the unit, `Duration::try_*` + `checked_mul`, `checked_add_signed`. Test at the crate boundary: `crates/xerj-api/tests/index_name_date_math_is_total.rs` (20 names + assertions that anchored math still resolves). Reverting only the `es_compat.rs` hunk makes both tests fail, first message `TimeDelta::days out of bounds`. |
| 2 | the harness fuzzes a function with no production callers | **fixed** — target now drives `xerj_api::es_compat::resolve_date_math_index` first (`xerj-api` added to `engine/fuzz/Cargo.toml`), keeps the engine copy on the same input, + 4 anchored seeds incl. `2026-01-01\|\|+9999999999999d`. The README now says *fuzz the function the request reaches, and check that it is that function*. |
| 3 | the CHANGELOG sentence is false | **fixed by landing the fix** — corrected to three copies with the anchored form named. |
| 4 | `aggs::lex_script` / `aggs::tokenize_script` | **fixed**, not narrowed. Two-character probe gated on both bytes being ASCII; `is_whitespace()` on a raw byte → `is_ascii_whitespace()` (`aggs.rs:1501` too). New `agg_script` fuzz target over all three entry points. Test: `crates/xerj-engine/tests/agg_script_tokenisers_are_total.rs` — 16 panics before, 0 after, and ASCII scripts still evaluate (`bucket_script` with `>=` + ternary → 12.0). |
| 5 | `config.html:165` contradicts `Config::validate` | **fixed by rewriting the prose**, keeping the refusal. `xerj.default.toml` ships `strategy = "size_tiered"`, so no config this project handed out trips the error — the WARN alternative would mean silently running size-tiered merging for an operator who picked a levelled policy for its read amplification. Also fixed the `storage.html` sentence that called `strategy` one of "three keys that log a warning" and then listed four keys. |
| 6 | truncated bullets in `cves-and-vulnerabilities.md` | **fixed** — the `cargo-audit` bullet now finishes (and says *why* informational advisories do not fail the build); the last bullet ends in a full stop. |

Gate on `69d7170c`, all run locally before pushing: `cargo fmt --all --check` and `cargo +nightly fmt --manifest-path fuzz/Cargo.toml --check` clean; `cargo clippy -p xerj-engine -p xerj-api --all-targets -- -D warnings` clean; `cargo test --release -p xerj-engine -p xerj-api` 39 suites / 775 passed / 0 failed; `cargo +nightly fuzz build` clean; `FUZZ_SECONDS=25 bash .github/scripts/fuzz-smoke.sh` all 7 targets survived.

CI on this head: **14/14 green**, including `Fuzz harnesses` (10m42s, up from 7m48s — the ASan build now includes `xerj-api`, plus a seventh target), `cargo-audit` (3m11s), ES-compat YAML conformance, and `Build + Test`.

Issue #206, opened 2026-08-07
A failed index cannot be inspected, deleted or retried without stopping the server


xerj-org:
An index whose directory fails to open at boot gets quarantined into a failed set. There is no way to inspect it, delete it, or retry without stopping the server and editing the data directory by hand.

That is the stuck-recovery state the feedback corpus is full of. The 90-hour recovery stories, the death spirals, the "pain in the ass since day one" quotes are mostly this: something is wrong, the tool will not tell you what, and the only lever is offline surgery.

Everything around this is well built, which is what makes the gap stand out. The WAL truncates a torn tail on open. Mid-file corruption freezes the generation instead of guessing. A data-dir format marker is checked before any destructive step, so a directory written by a different build is refused rather than eaten. The engine clearly takes recovery seriously. Then one index fails to open and the operator has no move.

What is needed:

- an endpoint that lists failed indices with the actual reason, not just the name
- delete a failed index without a restart
- retry opening one after the operator fixes the underlying cause

Second, related problem. Readiness goes hard-fail on any failed index, so a pod with 200 healthy indices and one broken one is removed from service permanently and never comes back. That converts a single-index problem into a total outage. Readiness should reflect whether the node can serve, not whether every index is perfect.

Third, the three health surfaces disagree with each other. `/v1/health`, `/v1/cluster/health` and `/health/ready` use real engine health, while `_cat/health` reports a hardcoded value. An operator watching the ES-compatible endpoint, which is the one their existing dashboards point at, will not see a red cluster. Whatever the answer is, all four should give it.

Found during a source review against the user-feedback corpus, where recovery states and operational opacity are the number one complaint.

Issue #203, opened 2026-08-07
Templates, pipelines, data streams and ILM policies are lost on restart


xerj-org:
Restarting XERJ loses index templates, ingest pipelines, data streams, ILM policies and component templates.

`Engine::new` restores only index directories, `es_mapping.json`, API keys and aliases (`xerj-engine/src/engine.rs:473-547`, see the restore calls at 502, 522, 530). The other management state lives in in-memory maps declared at `xerj-engine/src/engine.rs:264-336` and is never written to disk or read back.

So the single-binary upgrade story holds for data and not for configuration:

```
PUT _index_template/logs   -> 200
restart
GET _index_template/logs   -> 404
```

The next index that should have matched that template gets created without it. Nobody gets an error. The data keeps flowing into an index shaped differently from the one the operator designed.

That is a bad shape for the two moments it matters most, which are a version upgrade and an unplanned restart. Both are exactly when an operator is least able to notice that configuration quietly reverted.

Fix is to persist these the way aliases and API keys already are, and reload them in the same boot path. The pattern is already in the codebase, it just was not applied to these five maps.

Worth deciding before GA, because "replace the binary and restart" is the headline operational claim and this makes it untrue for anything except raw documents.

Found during a source review against the user-feedback corpus, where upgrade failures and lost configuration are recurring complaints.

PR #262, opened 2026-08-09
fix(ilm): execute retention policies instead of accepting and ignoring them

### LINKED ISSUE #199: ILM policies are accepted and stored but never executed, so retention silently does nothing

xerj-org:
> ## ⛔ BLOCKED — superseded by #247, which landed on `main` today
>
> **This PR cannot be merged as it stands, and the blocker is not text.** While
> it was in the merge gate, `main` merged
> [#247](https://github.com/xerj-org/xerj/pull/247) (`1b903196`,
> *"feat(engine): real ISM/ILM index lifecycle management — one engine, two
> REST surfaces"*) — an **independent, complete implementation of the same
> feature**: `xerj-engine/src/lifecycle.rs` (848 lines),
> `xerj-api/src/ism_api.rs` (726 lines), and `state.engine
> .spawn_lifecycle_manager()` wired into `main.rs` at the *same numbered slot*
> this PR uses.
>
> **The premise of this PR is therefore no longer true of `main`.** The opening
> section below says "nothing in the engine ever read it". On `main` at
> `1b903196`, something does: `lifecycle::tick` runs on a timer and calls
> `engine.delete_index` and `engine.rollover_data_stream`. Those sentences must
> not ship as release notes describing `main`'s behaviour.
>
> **Mechanically the two cannot be concatenated.** `git merge origin/main`
> conflicts in four files — `es_compat.rs`, `router.rs`, `engine.rs`,
> `main.rs` — and the route tables register the *same path and the same
> method* three times over:
>
> | path | `main` (#247) | this PR |
> |---|---|---|
> | `GET /_ilm/policy` | `es_compat::list_ilm_policies` | `es_compat::get_all_ilm_policies` |
> | `GET /:index/_ilm/explain` | `es_compat::ilm_explain` | `es_compat::explain_ilm` |
> | `PUT`/`GET`/`DELETE /_ilm/policy/:name` | `put/get/delete_ilm_policy` | `put/get/delete_ilm_policy` |
>
> axum 0.7 rejects a duplicate method+path registration when the router is
> built, so a "keep both sides" resolution does not produce a node with two ILM
> engines — it produces a node that does not boot.
>
> **This needs a decision, not a patch**, and it is outside what this
> merge-gate pass is authorised to make:
>
> - `main`/#247 executes **`rollover` and `delete`** and adds an OpenSearch
>   `_plugins/_ism/*` surface. This PR deliberately **refuses `rollover`** with
>   a 400 and executes `delete` + `readonly`.
> - This PR carries safety and honesty work #247 does not appear to: the
>   persisted **detach tombstone** that makes an acknowledged
>   `index.lifecycle.name: null` authoritative (the data-loss defect found in
>   review), the fail-closed **action allowlist**, the delete rails (never a
>   dot-prefixed index, never a data stream's write index, never an index whose
>   age cannot be established), `_ilm/status` + `_ilm/start|stop`,
>   `POST /{index}/_ilm/remove`, and `ilm_state.json` persistence.
>
> Someone has to choose: **close this as superseded and port the safety rails
> and detach semantics onto #247's engine as a follow-up**, or **rebase this to
> replace #247's implementation** and re-review both. Either is a substantial
> change to a shipped feature, not a merge-gate repair.
>
> **State of the branch.** Left at `4f5c0bc7`, the last head that was fully
> verified: **CI 12 of 12 green** there (including `Build + Test`), ES-compat
> YAML `1366 passed · 0 failed · 3 skipped`, and every merge-gate finding below
> fixed and falsification-tested. No speculative merge of #247 has been pushed,
> because resolving it by guesswork is exactly how the defects below got in.
>
> *Everything below this line was accurate as of `4f5c0bc7`, i.e. before #247
> landed. It is kept for whoever makes the call above.*

---

## The defect

`PUT /_ilm/policy/{name}` stored a policy in a `DashMap`, `GET` echoed it back, and nothing in the engine ever read it. `Engine.ilm_policies` had one writer and no reader outside the GET handler and a stats counter — no executor, no notion of index age, no persistence.

Reproduced first, on a stock **v1.0.0-rc.11** binary, with a delete phase due immediately:

```
PUT /_ilm/policy/repro-199 {"policy":{"phases":{"delete":{"min_age":"0ms","actions":{"delete":{}}}}}}
  -> {"acknowledged":true}
PUT /logs-199 {"settings":{"index.lifecycle.name":"repro-199"}}   -> 200
POST /logs-199/_doc?refresh=true                                  -> created

...two hours later:
GET /logs-199/_count       -> {"count":1}
GET /logs-199              -> 200
GET /logs-199/_ilm/explain -> 404          (no way to even ask)
```

Two more defects the same run exposed: `GET` answered `{"repro-199":{"policy":{"policy":{…}}}}` (the stored body kept its own envelope, so the response was double-wrapped and unlike ES), and `PUT /_ilm/policy/x {"this":"is not a policy"}` was accepted with `{"acknowledged":true}`.

## The fix

Both honest halves — **execute what we can, refuse what we cannot.**

- **New `xerj-engine/src/ilm.rs`.** Policy model, strict `min_age` parsing, a fail-closed validator, and `Engine::run_ilm_once(now_ms)` — one pass that ages every managed index and applies the due phases. The clock is a parameter, so a 30-day transition is tested without sleeping.
- **Executes `delete` and `readonly`** — the two actions this engine can genuinely perform.
- **Rejects everything else.** `EXECUTABLE_ACTIONS` is an allowlist. The canonical ES logs policy is now a 400 naming `hot.rollover, warm.forcemerge` rather than a silent no-op. No new accepted-and-ignored input (#204's bug class).
- **Persists** policies and lifecycle attachments in `<data_dir>/ilm_state.json` (atomic temp+rename, mirroring `aliases.json`). Retention that forgets its policy on restart is retention that silently stops.
- **Observability**: `GET /{index}/_ilm/explain`, `GET /_ilm/status` with real counters, `POST /_ilm/start|stop` as an operator kill switch. Config gains `[ilm] enabled` (default true) and `poll_interval_secs` (default 600, ES's `indices.lifecycle.poll_interval`).
- `GET /_ilm/policy/{name}` answers in ES's shape; `GET /_ilm/policy` lists all; `DELETE` refuses while an index still points at the policy, naming it — decided by the **same resolver the executor uses**, so it cannot disagree with what the pass will actually do.
- **Detaching is authoritative and is a first-class verb.** `POST /{index}/_ilm/remove` (ES's own) and `PUT /{index}/_settings {"index.lifecycle.name": null}` both record a persisted tombstone that outranks the index's own `settings.json`, so an acknowledged "stop deleting my data" genuinely stops retention — including on an index upgraded from before `ilm_state.json` existed.
- **Data streams are managed.** Backing indices are `.ds-<stream>-NNNNNN`, which no `logs-*` template pattern matches by name, so the policy is resolved from the *stream* name at backing-index creation. The current write index is never deleted; older generations age out.

### Safety rails

- **Never deletes an index whose age cannot be established** — skipped with the reason logged and surfaced in `_ilm/explain`. Every age fallback errs late, never early.
- **Never deletes a dot-prefixed internal index**, so a wildcard policy cannot eat a second brain. `.ds-*` backing indices are exempt from the rail (they are what ILM is for) but a data stream's current write index is never deleted.
- **A template-attached policy is resolved at index creation, never re-read when the executor runs.** This was a real hazard in the first draft: re-reading the template at evaluation time makes a `logs-*` template written today retroactively manage — and past its `min_age`, *delete* — every `logs-*` index that already existed yesterday. ES applies template settings at creation and never retroactively. With the evaluation-time lookup restored, the new test fails with `deleted: ["logs-from-last-year", "logs-from-today"]`.
- **An unparsable `min_age` rejects the policy** rather than defaulting to 0 — a silently-zeroed `min_age` would delete the user's data on the next pass.

## The other half of the issue

The second comment reports the same failure in XERJ's own config: `logs.retention_days`, documented as "how long to retain log data before automatic deletion (default: 90 days)", is read by nothing. It cannot simply be removed — `LogsConfig` is `deny_unknown_fields`, so deleting the field turns every existing `xerj.toml` that sets it into a node that refuses to boot. The doc comment now says plainly that it is not implemented, and the server logs a WARN at boot when it is set away from its default, naming the surface that does work. Wiring or removing it stays tracked in #204.

## Verification

| check | result |
|---|---|
| `cargo clippy -p xerj-{engine,api,common,server} --all-targets -- -D warnings` | clean |
| `cargo test -p xerj-engine --test ilm_retention_executes` | 8 passed |
| `cargo test -p xerj-api --test ilm_policies_are_executed` | 7 passed |
| `cargo test -p xerj-common -p xerj-api -p xerj-server` | 21 targets, 0 failed |
| `cargo test -p xerj-engine` | 29 of 30 targets pass at the time of this run; `painless_script_limits` aborts — see note. Post-merge re-run: 31 of 32 |
| ES-YAML conformance (binary built from this branch) | **1366 passed · 0 failed · 3 skipped** |

The one engine failure is `painless_script_limits`, a stack overflow in a debug build on this machine. It is **pre-existing and unrelated**: it reproduces identically at `origin/main` (1ee31ecd) after a forced rebuild, and overflows a *different* test on each run (`rescore_script_over_call_depth_reports_a_failure`, then `call_depth_limit_survives_the_multi_thread_block_in_place_path`, then `script_failure_is_not_cached_or_replayed`) — a resource-dependent flake on a box running many concurrent cargo jobs, not a regression from this change.

The YAML suite ran against a server with `poll_interval_secs = 5`, so the executor ticked ~200 times during the run and disturbed nothing.

### Live, on a binary built from this branch

```
PUT /_ilm/policy/logs-full   (rollover + forcemerge + delete)
  -> 400 "xerj executes only these ILM actions: delete, readonly.
          This policy also asks for: hot.rollover, warm.forcemerge."
PUT /_ilm/policy/oops        (min_age "30", no unit)
  -> 400 "time value '30' has no unit — use one of ms, s, m, h, d"
GET /_ilm/policy/repro-199
  -> {"repro-199":{"version":1,"modified_date":…,"policy":{"phases":{…}}}}   (single envelope)

PUT /logs-199 {"settings":{"index.lifecycle.name":"repro-199"}} + one doc
GET /logs-199/_ilm/explain -> managed:true, phase:"delete", age_measured_from:"recorded_at_attach"
...one pass later:
  INFO ILM delete phase is deleting this index index="logs-199" policy="repro-199"
       age="416ms" age_from="recorded_at_attach"
GET /logs-199 -> 404
GET /_ilm/status -> {"operation_mode":"RUNNING","xerj":{"passes":6,"indices_deleted":1,…}}
```

## Prior art (reference-coding mandate)

quickwit's janitor, Apache-2.0, **adapted not copied** — verified by reading the corpus clone directly, since the retrieval server on `:9200` was down for this run:

- `quickwit-janitor/src/actors/retention_policy_executor.rs:35` — retention runs on a coarse periodic loop (`RUN_INTERVAL = 1 hour`), not per-index timers. Ours defaults to 10 minutes and is configurable.
- same file, `78-80` — *"Should not return an error to prevent the actor from crashing"*: the loop logs and continues. `run_ilm_once` returns a report and never propagates an error out of a tick.
- `quickwit-janitor/src/retention_policy_execution.rs:65-76` — splits that cannot be dated are **warned about and skipped, never deleted**. Same rule here.
- same file, `49-52` — cutoff is `now - retention_period`, one directional comparison.

Elasticsearch was consulted for **wire semantics only** (phase names, `min_age` origin, `_ilm/explain` and `_ilm/status` shapes). No ES code is copied — it is AGPL/SSPL/Elastic-2.0 and XERJ shares no code with it.

## Upgrade note

Policies were in-memory before this change, so they do not survive the upgrade — nothing can be deleted until an operator re-PUTs a policy, which is the explicit act that re-affirms intent. After that, an index older than its delete phase's `min_age` is deleted on the next pass.

## Scope

Touches `xerj-engine`, `xerj-api`, `xerj-common`, `xerj-server`, and two playbook docs. No overlap with #203's persistence work beyond `engine.rs` struct fields; `ilm_state.json` is a separate file and loading is idempotent, so if #203's general metadata persistence lands, this half can be folded into it.

Fixes #199

---

## Adversarial-review repairs (2026-08-10, commit `30e6e36f`)

Review of the first cut found that it shipped a data-loss defect of exactly the class this PR exists to eliminate, plus four smaller correctness gaps. All five are fixed on this branch; the full root-cause writeup is in `30e6e36f`'s commit body.

| # | Defect | Fix |
|---|---|---|
| 1 | **DATA LOSS** — an acknowledged `index.lifecycle.name: null` detach was ignored and the index deleted anyway. `set_index_lifecycle_policy` cleared only the in-memory entry while `ilm_policy_for_index` fell back to the index's persisted `settings.json`, which still carried the create-time policy. Silent no-op entirely when no entry existed. | A detach records a **persisted tombstone** (`IlmIndexState { policy: None }`); the resolver answers from recorded state *totally* and consults settings only when there is no entry at all. |
| 2 | `DELETE /_ilm/policy` decided "in use" by scanning the same map the broken detach cleared — so it returned 200 for a policy an upgraded index still pointed at, and 400 for a detached one. | Both now go through `Engine::ilm_indices_using_policy`, i.e. the executor's own resolver over the live index set. |
| 3 | **Accept-and-ignore (#204's class)** — a data stream under an ILM-carrying template was silently unmanaged: `evaluated: 0`, no error. | The template is resolved from the **stream name** at backing-index creation (`create_data_stream`, `rollover_data_stream`). Still creation-time only. |
| 4 | `spawn_ilm_executor` is a non-test `tokio::spawn` reaching `get_index`/`delete_index`, but `index_guard`'s contract — carry the guard or say why not, and the audit table lists them all — was satisfied by neither. | Audit-table row added plus a spawn-site paragraph: no guard *deliberately*, since scoping a retention tick to whoever booted the node would make retention apply to that principal's indices only. |
| 5 | `delete_data_stream` bypasses `delete_index`, so ILM state outlived the index — status reported a phantom managed index and the policy became permanently undeletable. | `forget_ilm_index(backing)` in the loop. |

Also found and fixed while in there: a settings body carrying `index.lifecycle.origination_date` **detached the index**, because the handler asked "does this body mention lifecycle?" and then read the absent name. Replaced by a three-valued directive (absent / explicit null / a name). And `POST /{index}/_ilm/remove` — ES's actual detach verb, previously not implemented at all — now exists, sharing one code path with the settings route.

### Verification of the repairs

**12 new regression tests. Reverting each fix falsifies the tests that cover it:**

| revert | what fails |
|---|---|
| resolver + detach | 4 new engine tests, 4 API tests — including the pre-existing `deleting_a_policy_still_in_use_is_refused`, which asserted only the 200 and is how this got through |
| data-stream template resolution | `ilm_policy_for_index(".ds-applogs-000001")` = `None` |
| the `forget_ilm_index` line alone | `managed_indices = 1` for an index that no longer exists |

**Live, on a release binary built from this branch** (real clock, `poll_interval_secs = 5`):

```
detach (nested)  logs-detached -> managed:false -> survives the pass, 200
detach (flat)    logs-flat     -> managed:false -> survives the pass, 200
POST _ilm/remove logs-remove   -> managed:false -> survives the pass, 200
control          logs-control  -> managed:true  -> DELETED, 404
GET /logs-detached/_settings   -> no lifecycle key, in any spelling
DELETE /_ilm/policy/inuse-p    -> 400 "in use by 1 index(es): inuse-idx"

data stream applogs, applogs* template carrying a policy:
  .ds-applogs-000001  managed:true, blocked "current write index of data stream 'applogs'"
  after POST /applogs/_rollover:  gen1 -> 404,  gen2 (write index) -> 200
DELETE /_data_stream/applogs   -> managed_indices 0, policy deletable again
```

| gate | result |
|---|---|
| `cargo clippy -p xerj-engine -p xerj-api --all-targets -- -D warnings` | clean |
| `cargo test -p xerj-api` | 13 targets, 0 failed (15/15 ILM) |
| `cargo test -p xerj-engine --no-fail-fast` | **29 of 30 targets pass, 1 fails**: `painless_script_limits` — see below. Post-merge re-run: 31 of 32 |
| ES-YAML conformance (binary from this branch) | **1366 passed · 0 failed · 3 skipped** |

**The one engine target that fails, stated exactly.** `cargo test -p xerj-engine --no-fail-fast` at this head is 29 targets passing and one aborting: `painless_script_limits` overflows the debug build's stack and the harness dies with `SIGABRT`. It is not a regression from this change and does not touch it — the file is untouched by this PR, it reproduces at `origin/main`, it **overflows a different test on each run** (`terms_set_min_should_match_script_over_call_depth_reports_a_failure` this time; `rescore_script_over_call_depth…` and `call_depth_limit_survives_the_multi_thread_block_in_place_path` on earlier runs), and the whole target passes 9/9 under `RUST_MIN_STACK=33554432`. It is a resource-dependent flake on a box running many concurrent cargo jobs, but it *is* a failure and this PR is not entitled to call the target green.

`field_less_query_string_cross_product_respects_the_request_deadline`, named as a second failure in the previous revision of this section, **passed** on this run. It asserts an absolute 3s wall-clock bound, so it fails when the box is loaded; calling it a standing failure was as wrong as calling the suite green.

### Prior art for the repairs

Elasticsearch's `DataStreamIT.testComposableTemplateOnlyMatchingWithDataStreamName` (`modules/data-streams/src/internalClusterTest/.../DataStreamIT.java:454`) pins that a data stream's template is resolved from the **stream name**, deliberately using a wildcard-free pattern *so the backing indices cannot match by name* — which is the bug in #3 and the shape of its fix. **Approach only**: Elasticsearch is AGPL/SSPL/Elastic-2.0 and no ES code is copied here.

---

## Merge-gate repairs (2026-08-10, commit `1ec34b0b`)

The merge gate held this at `30e6e36f` on two grounds, both correct. The
adversarial repair had introduced a fresh accept-and-ignore defect **on the
endpoint it added**, and three shipped sentences claimed things the code
contradicts. Both are fixed; the missing CHANGELOG entry is added.

### 1. The new accept-and-ignore defect (#204's class), fixed

`POST /ghost-0/_ilm/remove` answered `200 {"has_failures":false}` for a name
that is not an index, and wrote a **permanent detach tombstone** for it. The
handler's own `if targets.is_empty() { …404… }` guard was *dead* for the case
it was written for: `resolve_index_selector` documents that it returns a
literal name "whether or not it exists; the caller decides", so a nonexistent
literal never yields an empty result. `PUT /{index}/_settings` had the same
hole. Every call runs `flush_ilm_state`, which serialises and atomically
rewrites the whole file, so N distinct names on the public ES port were O(N)
unbounded on-disk growth with O(N²) write amplification.

| | at `30e6e36f` | at `1ec34b0b` |
|---|---|---|
| `POST /ghost-0/_ilm/remove` | `200 has_failures:false` | **`404 index_not_found_exception`** |
| `PUT /ghost-1/_settings {"index":{"lifecycle":{"name":null}}}` | `200 acknowledged` | **`404 index_not_found_exception`** |
| 200 calls to distinct phantom names | 200 entries, `ilm_state.json` 4129 B | **0 entries, 58 B** |

Measured both times on a release binary built from the branch. The fix is
`resolve_existing_index_targets`, used by both routes: a literal name (or a
member of a comma list) that is neither an index nor an alias is a 404, ES's
`ignore_unavailable=false` default.

Deliberately **not** changed, because they are different questions the previous
doc comment conflated: detaching an index that **exists but is unmanaged** is
still `200` (genuinely idempotent, as in ES), and a **wildcard matching
nothing** is still an empty `200` (`allow_no_indices=true`) that writes no
state. Both are asserted.

### 2. `GET /_ilm/status` `managed_indices` now asks the resolver

It counted `ilm_index_state` entries — the same map-vs-resolver split that
broke `DELETE /_ilm/policy` and was fixed *there*, left standing in the one
place it wasn't. An index upgraded from before `ilm_state.json` is managed
through its persisted `settings.json` alone, so status reported `0` for
exactly the population the executor was genuinely deleting. `ilm_status` is now
async and walks `list_index_names()` through `ilm_policy_for_index`.

### 3. The three false claims, corrected

| where | said | now |
|---|---|---|
| `ilm.rs:686` | `POST /{index}/_ilm/remove` "is not implemented" | it is implemented (`router.rs:473`) and is this PR's headline repair; the comment now describes both detach routes and the existence precondition |
| `es_compat.rs` `remove_ilm_policy_from_index` | "an index with no policy is not a failure — ES treats the call as idempotent" | separates "exists but unmanaged" (200, idempotent) from "not an index" (404), and records why the old guard was dead |
| gate table `cargo test -p xerj-engine` \| "all green" | contradicted three lines below by a paragraph naming failures | the row now names the failing target; the paragraph below states exactly what fails and what does not (see **Verification of the repairs**, corrected in place) |

Two more of the same shape, found while in there and not flagged by the gate:
`ES_COMPATIBILITY.md` and `config.rs` both said ILM left the accept-and-store
list **"in v1.0.0-rc.13"**. rc.13 and rc.14 are tagged and do **not** contain
this change; it is `[Unreleased]`, and both now say so.

### 4. CHANGELOG entry added

The diff turned XERJ from "stores ILM policies" into "a background task
deletes indices on a 600s timer" and touched no `CHANGELOG.md`. The
`[Unreleased]` entry now carries, in shipped form, the upgrade note that
previously lived only in this PR body — **policies were in-memory before this
change and do not survive the upgrade** — plus the safety rails, the kill
switch, and the known gaps in plain words: `rollover` is refused so a data
stream nobody rolls over never deletes anything; templates are read at index
creation and never retroactively; `logs.retention_days` is still not
implemented and is tracked in #204. The 404 behaviour change above is its own
entry, since it is a visible ES-compat change on two endpoints.

### 5. Merged `origin/main` (v1.0.0-rc.14)

main shipped rc.14 while this sat in the gate and the branch went
`CONFLICTING`. Merge commit `e8231c1a`, two conflicts, both resolved by keeping
**both** sides:

- `create_index_with_settings` — main's #206 failed-index guard runs first (it
  is a refusal), then this branch's create-time ILM attachment read.
- `forget_index_metadata` — main refactored `delete_index` so this shared
  helper now serves **both** the open-index and the failed-index delete paths.
  `forget_ilm_index(name)` moves into it, which is strictly better than where
  it was: a *failed* index that is deleted must stop being ILM-managed for the
  same reason an open one must, and before the merge it would not have.
- `CHANGELOG.md` — rc.14 shipped, so the old `[Unreleased]` body is now the
  `[1.0.0-rc.14]` section. This branch's two entries move up under the new,
  empty `[Unreleased]`; main's rc.14 entries are untouched.

### CI

**`e8231c1a`: 12 of 12 checks green**, including **`Build + Test`** — the
single job that compiles and runs the workspace test suite, and the one the
merge gate correctly refused to accept as pending on the previous head.
ES-compat YAML in CI: `1366 passed · 0 failed · 3 skipped`.

`55b00629` is `e8231c1a` plus a second `origin/main` merge (below). Its
incoming commits touch `xerj-autoindex`, the contributors list, and
`metrics/release-downloads.jsonl` (the automated daily download snapshot) —
no engine or API crate overlap. **CI has since completed on `55b00629`: 12 of
12 checks green**, ES-compat YAML `1366 passed · 0 failed · 3 skipped · 1369
total` (run 31377005616).

### Local gate on `e8231c1a` (post-merge, all re-run)

| gate | result |
|---|---|
| `cargo clippy -p xerj-engine -p xerj-api -p xerj-common -p xerj-server --all-targets -- -D warnings` | clean |
| `cargo test -p xerj-api -p xerj-common -p xerj-server --no-fail-fast` | 24 targets, **0 failed** (15/15 ILM) |
| `cargo test -p xerj-engine --no-fail-fast` | **31 of 32 targets pass, 1 aborts** — `painless_script_limits`, see below |
| ES-compat YAML conformance (release binary from this branch, private port 9271) | **1366 passed · 0 failed · 3 skipped · 1369 total** |

The phantom-index probes were re-run against a release binary built from the
merged head: `POST /ghost-0/_ilm/remove` → 404, `PUT /ghost-1/_settings
{"index":{"lifecycle":{"name":null}}}` → 404, 200 phantom calls → `ilm_state.json`
still 58 bytes / 0 entries, existing-but-unmanaged detach → 200, wildcard
matching nothing → 200.

**3 new regression tests.** Two are falsified by reverting the fix each
covers. The third,
`ilm_remove_on_an_existing_unmanaged_index_is_still_a_success`, pins the two
cases that must *stay* a 200 (an existing-but-unmanaged index, and a wildcard
matching nothing), so no revert makes it fail — it is a guard against
over-correcting, not a falsification of a fix.

| revert | what fails |
|---|---|
| the `_ilm/remove` existence guard | `ilm_remove_on_a_name_that_is_not_an_index_is_404_and_writes_nothing` — `left: 200, right: 404` |
| `managed_indices` back to the map scan | `status_counts_the_upgraded_index_the_executor_would_actually_delete` — `left: 0, right: 1` |

### Known gaps at this head, stated plainly

- **`painless_script_limits` aborts** in the debug build (`fatal runtime error:
  stack overflow` → `SIGABRT`), so `cargo test -p xerj-engine` is 31/32, not
  green. Directly measured here: this PR touches no scripting code (see the
  diffstat), the target overflows a **different test on each run**
  (`terms_set_min_should_match_script_over_call_depth_reports_a_failure`, then
  `rescore_script_over_call_depth_reports_a_failure`), and it passes **9/9**
  under `RUST_MIN_STACK=33554432`. That is a stack-budget flake on a box
  running many concurrent cargo jobs — but it is a failure, and this PR does
  not get to call the target green *locally*.

  **Settled by CI.** "Build + Test" compiles and runs the workspace suite in
  **release** mode (`target/release/deps/painless_script_limits`), where the
  frames are smaller and the overflow does not occur — the job concluded
  `SUCCESS` on `e8231c1a`. The failure is specific to a debug build on a
  loaded box, not to this change.
- **`rollover` is not executed.** On a data stream whose caller never issues
  `POST /<stream>/_rollover`, the write index is the only generation and the
  write index is never deleted, so a delete-only policy on such a stream
  deletes nothing, forever. Disclosed in `ES_COMPATIBILITY.md` and the
  CHANGELOG.
- **`logs.retention_days` is still not wired.** It is documented as
  unimplemented and WARNs at boot; wiring or removing it stays in #204.
- **Template policies are creation-time only** — deliberately, since the
  alternative retroactively deletes pre-existing data.

### 6. Second `origin/main` merge (`55b00629`)

main moved again while CI ran (#248 / #249 / #273 — autoindex PDF parse-once,
plus a contributors-list entry). One conflict, `CHANGELOG.md`'s `[Unreleased]`
section, resolved by keeping both sides: main's `### Changed` autoindex entry,
then this branch's `### Fixed` ILM entries. No engine or API crate overlap.
`cargo fmt --all -- --check` and `cargo clippy -p xerj-{engine,api,common,server}
--all-targets -- -D warnings` are clean on it, and both ILM targets are 15/15.

One sentence in `ES_COMPATIBILITY.md` was tightened in the same commit: it now
says the ILM fix is in **no tagged release up to and including v1.0.0-rc.14**,
rather than describing its own merge status, so it stays true whenever it is
read.

---

## Merge-gate repairs, round 2 (2026-08-10, commit `6bd3862d`)

The previous round's fixes were re-verified at this head rather than taken on
trust, and one more instance of the **same root cause** was found on a route
the previous round did not touch.

### 1. `_ilm/explain` had its two target cases inverted

Round 1 fixed the dead `targets.is_empty()` guard on the two ILM routes that
*write* state (`PUT /{index}/_settings`, `POST /{index}/_ilm/remove`). The
third route with the same guard — `GET /{index}/_ilm/explain` — was left as
it was, and had both cases backwards:

| | at `55b00629` | at `6bd3862d` | ES |
|---|---|---|---|
| `GET /typo-idx/_ilm/explain` (not an index) | `200 {"managed": false}` | **`404 index_not_found_exception`** | 404 |
| `GET /nomatch-*/_ilm/explain` (wildcard, no match) | `404` | **`200 {"indices": {}}`** | 200 |
| `GET /real-idx/_ilm/explain` | 200, explains | 200, explains | 200 |

Measured both times on a release binary built from this branch, private port
9271.

**Why the 200 was the serious half.** `_ilm/explain` is the observability
surface this PR added so an operator can answer "is retention running on this
index?". Answering `managed: false` for a mistyped name reads as *"nothing is
managing it"* — while the index they actually meant is being deleted on a
600s timer. It is a confident wrong answer on the endpoint whose whole job is
to prevent one. Nothing is persisted here, so this is not the unbounded
on-disk growth the gate found on `_ilm/remove`; it is the same root cause with
a smaller blast radius, in the one place round 1 did not look.

Root cause is identical: `resolve_index_selector` returns a literal name
"whether or not it exists; the caller decides", so `targets` was never empty
for a literal and the handler's own 404 was dead — while a wildcard matching
nothing *did* produce an empty vector and hit that 404, the inverse error.
Now uses `resolve_existing_index_targets`, the helper round 1 already added.

ES semantics were read from the corpus clone rather than recalled:
`RestExplainLifecycleAction.java:42` passes `IndicesOptions.strictExpandOpen()`
= `ERROR_WHEN_UNAVAILABLE_TARGETS` + `allowEmptyExpressions(true)`
(`IndicesOptions.java:561-563`). **Approach only** — Elasticsearch is
AGPL/SSPL/Elastic-2.0 and no ES code is copied.

The CHANGELOG's 404 entry named two endpoints; it now names all three and
states the `managed: false` misreport in plain words.

**New test**, falsified by reverting only the resolver line:
`explain_404s_a_name_that_is_not_an_index_and_200s_an_empty_wildcard` fails
`left: 200, right: 404`.

### 2. Round 1's claims, re-measured at this head

Each verified independently on a release binary from this branch, not read off
the previous commit message:

```
POST /ghost-0/_ilm/remove                                 -> 404 index_not_found_exception
PUT  /ghost-put/_settings {"index":{"lifecycle":{"name":null}}} -> 404
PUT  /ghost-flat/_settings {"index.lifecycle.name":null}  -> 404
200 POSTs to distinct phantom names -> no ilm_state.json written at all, managed_indices 0
POST /plain-idx2/_ilm/remove   (exists, unmanaged)        -> 200 has_failures:false
POST /nomatch-*/_ilm/remove    (wildcard, no match)       -> 200, no state written

detach (nested) detach-me -> managed:false -> survived the pass, 200
POST _ilm/remove remove-me -> managed:false -> survived the pass, 200
control keep-me/control-me/plain-idx -> managed:true -> DELETED, 404
PUT policy with hot.rollover -> 400 naming `hot.rollover`
```

**The upgrade case, re-proved across a real restart.** With `upgraded-idx`
attached only through its persisted `settings.json` (its `ilm_state.json`
entry removed while the node was down, simulating a pre-`ilm_state.json`
node), after restart: `managed_indices` = **1**, `_ilm/explain` reports
`managed: true, policy: p30d`, and `DELETE /_ilm/policy/p30d` is refused
*"in use by 1 index(es): upgraded-idx"*. The two detach tombstones survived
the same restart and both still report `managed: false`.

### 3. Local gate at `6bd3862d`

| gate | result |
|---|---|
| `cargo clippy -p xerj-engine -p xerj-api -p xerj-common -p xerj-server --all-targets -- -D warnings` | clean |
| `cargo test -p xerj-api -p xerj-common -p xerj-server --no-fail-fast` | 24 targets, **0 failed** (16/16 ILM) |
| `cargo test -p xerj-engine --no-fail-fast` | **31 of 32 targets pass, 1 aborts** — `painless_script_limits`, unchanged and documented below |
| ES-compat YAML conformance (release binary from this branch, private port 9271) | **1366 passed · 0 failed · 3 skipped · 1369 total** |

`painless_script_limits` was re-checked rather than assumed: it aborts on the
default debug stack under load and passes **9/9** here under
`RUST_MIN_STACK=33554432`, exactly as the section above describes. It remains
the one target this PR does not get to call green locally.

### Known gaps still open at this head

Unchanged from the list above, and none of them fixed by this round:

- **`painless_script_limits` aborts locally** in a debug build under load
  (green in CI's release build, green locally with a 32 MiB stack). Not a
  regression from this PR and not fixed by it.
- **`rollover` is not executed**, so a data stream nobody rolls over never
  deletes anything.
- **`logs.retention_days` is still not wired** — documented as unimplemented,
  WARNs at boot, tracked in #204.
- **Template policies are creation-time only**, deliberately.
- **A corrupt `ilm_state.json` is logged and skipped at boot.** Policies and
  attachments live in the same file, so a corrupt file means the executor
  manages nothing and deletes nothing — fail-safe — *until* an operator
  re-`PUT`s a policy, at which point any detach tombstone that file held is
  gone and an index attached through its `settings.json` becomes managed
  again. This is the documented upgrade path's behaviour, reached by a
  different route; it is not fixed here and is stated so it is not a surprise.

### 4. Third `origin/main` merge (`4f5c0bc7`)

main moved again while round 2 was in the gate (#254 — autoindex rerun
refusal, plus the brain zero-live classification). One conflict,
`CHANGELOG.md`'s `[Unreleased]`, resolved by keeping **both** sides: main's
`### Changed — autoindex refuses reruns that would strand documents` continues
the existing `### Changed` run, then this branch's `### Fixed` ILM entries
follow. No engine or API crate overlap — main's commits touch
`xerj-autoindex`, `xerj-server/src/brain.rs`, docs and CI scripts, plus
`xerj-engine/src/index.rs`, which auto-merged (this branch adds only
`settings_snapshot` there, in a different impl region).

**The whole gate was re-run on the merged head `4f5c0bc7`, not carried over:**

| gate | result on `4f5c0bc7` |
|---|---|
| `cargo clippy -p xerj-engine -p xerj-api -p xerj-common -p xerj-server --all-targets -- -D warnings` | clean |
| `cargo test -p xerj-api -p xerj-common -p xerj-server --no-fail-fast` | 24 targets, **0 failed** (16/16 ILM) |
| `cargo test -p xerj-engine --no-fail-fast` | **31 of 32 targets pass, 1 aborts** — `painless_script_limits`, the documented debug-stack flake |
| ES-compat YAML conformance (release binary from this head, private port 9271) | **1366 passed · 0 failed · 3 skipped · 1369 total** |

And the behaviour probes, re-run against a release binary built from
`4f5c0bc7`:

```
GET  /typo-idx/_ilm/explain    -> 404 index_not_found_exception
GET  /nomatch-*/_ilm/explain   -> 200 {"indices":{}}
GET  /rr-idx/_ilm/explain      -> 200, explains
POST /ghost-9/_ilm/remove      -> 404
POST /rr-idx/_ilm/remove       -> 200 has_failures:false
```


xerj-org:
Closing as superseded by #247, which landed on main first and is an independent complete implementation of the same feature.

This is not a judgement on the code here — it was 12/12 green at `4f5c0bc7` with every merge-gate finding fixed. The two simply cannot be concatenated: axum rejects a duplicate method+path registration, so a keep-both resolution produces a node that does not boot.

The work this PR carries that #247 does not — the persisted detach tombstone, the fail-closed action allowlist, the delete rails, `_ilm/status`/`start`/`stop`, `POST /{index}/_ilm/remove` and `ilm_state.json` persistence — is captured in https://github.com/xerj-org/xerj/issues/282 to be ported onto the shipped engine. The branch is left at `4f5c0bc7` to lift from.

Verified before closing that this does not leave a live defect on main: #247 measures age from a recorded state-entry timestamp rather than a filesystem date, so a freshly attached index cannot look instantly old, and `delete_index` carries the #79 `index_guard` backstop. The rails above are hardening, not a fix for an active data-loss path.

PR #263, opened 2026-08-09
fix(engine): persist templates, pipelines, data streams and ILM policies (#203)

### LINKED ISSUE #203: Templates, pipelines, data streams and ILM policies are lost on restart

xerj-org:
Fixes #203

## The defect, reproduced first

`PUT /_index_template/logs` answered `200 {"acknowledged": true}`, and after a restart `GET /_index_template/logs` answered 404. Verified at `origin/main` before writing anything, by running the new in-process restart test against the unmodified sources:

```
GET /_index_template/logs after restart:
  {"error":{"type":"index_not_found_exception",
            "reason":"index not found: index template [logs] missing"},"status":404}
left: 404  right: 200
```

The next index that should have matched that template is created without it, and nobody gets an error — so "replace the binary and restart" held for documents and not for anything the operator had configured.

## Root cause

`Engine::new` restored index directories, `es_mapping.json`, API keys and aliases, and nothing else. Index templates, legacy (v1) templates, component templates, ingest pipelines, data streams and ILM policies lived only in `DashMap`s on `Engine`. The persistence pattern was already in the tree (`flush_aliases` / `load_persisted_aliases`); it had simply never been applied to these six maps.

## The fix

`PersistedClusterState` → `<data_dir>/cluster_state.json`, one document for all six maps so a data stream and the template that shaped it can never come back from different generations. Written through `index::write_file_atomic` (write to `.tmp`, fsync, rename, fsync the parent directory). The rename is the commit point, which is what makes `kill -9` safe. Reloaded by `Engine::load_cluster_state` in `Engine::new`, after the index scan and before the listeners come up.

Everything else follows from making that the only write path:

- **Typed mutators** (`put_index_template`, `delete_ilm_policy`, `put_pipeline`, …) — every ES-compat and native handler goes through them, and each map's field doc says so, because a raw `.insert()` that skips the flush is how this bug class returns.
- **A failed write is reported, not swallowed.** The in-memory change is rolled back and the handler answers 500 instead of `acknowledged`. The operator learns at the write, not at the next boot.
- **Ingest pipelines are recompiled on boot, not just re-read.** Storing the definition without rebuilding the executable form would leave `?pipeline=x` accepted and quietly inert after a restart — the accepted-and-ignored shape tracked by #204.
- `Engine::create_pipeline` becomes the private `compile_pipeline` (it does not touch disk, so it must not be reachable from a request); the xerj-native `PUT /v1/pipelines/{name}` had the same loss and now uses `put_pipeline`.
- `DELETE /_ingest/pipeline/{id}` now drops the compiled form too — it previously kept running a pipeline `GET` reported as 404, until the next restart.

## Four further defects found while verifying, each reproduced before it was fixed

**1. Concurrent writers corrupted each other.** `write_file_atomic` stages through one fixed path, so two flushes in flight truncate and rename the *same* file. A provisioning script firing its PUTs in parallel was enough — 32 concurrent template PUTs gave `{"type":"store_exception","reason":"storage error: IO error"}` immediately. Serializing snapshot-and-write is also the correct shape rather than a workaround: because the snapshot is taken inside the lock, the last file written is always a superset of every mutation already applied, so no acknowledged write is lost to a stale snapshot landing after a fresher one.

**2. An interrupted rollover wedged the data stream permanently.** `rollover_data_stream` creates `.ds-<name>-00000N` before it records generation N, so a crash in that window left the index on disk and the document saying N-1. Every later rollover then recomputed the same name and got `409 resource_already_exists_exception: .ds-metrics-app-000002` — forever. Boot now reconciles each stream's generation against the backing indices actually present, adopts the higher one, warns, and persists the repair once.

**3. A `cluster_state.json` that does not load was destroyed by the next write.** The node boots with empty maps, so the first PUT rewrote the file and took whatever was still in it. See the round-2 section below for how this is handled now — the first pass fixed only the corrupt-parse half of it, and the half it missed was the worse one.

**4. `DELETE /_data_stream` recorded the removal before it destroyed the backing indices.** Found by the final merge gate, in the subsystem this PR is about. The function even carried a comment arguing that recording first is the recoverable ordering — the argument describes the *opposite* ordering to the one the code implemented. Measured on a release binary built from the PR head, by putting a data dir into exactly the state a crash in that window leaves (stream removed from `cluster_state.json`, `.ds-ds3-000001/2/3` still on disk) and rebooting:

```
GET    /_data_stream/ds3  -> 404
DELETE /_data_stream/ds3  -> 404
PUT    /_data_stream/ds3  -> 409 resource_already_exists_exception: .ds-ds3-000001
boot                      -> no warning of any kind
```

That is defect 2's permanent wedge, reintroduced on the delete path. The order is now reversed: the backing indices are destroyed first and the removal is flushed only once they are gone, so a crash in the window leaves a stream the next boot restores and warns about (`restored data stream references backing indices that are not present in the data dir`) and a re-issued DELETE finishes the job. `let _ = idx.delete_all_data().await` went with it — a backing index that cannot be destroyed now fails the DELETE with a 500 instead of answering `{"acknowledged": true}` about data still on disk (the #204 shape), and it is the swallow that would have made the reordering unsafe: a failed delete would still have recorded the removal. The stream and the index are put back, so the retry after the cause is fixed completes; `NotFound` from `remove_dir_all` counts as success so the retry is idempotent.

To be exact about what that 500 means, because the release note has to be: the loop does **not** stop at the first failure. Every backing index that can be destroyed is destroyed, and only the ones that could not are left — the caller asked for the whole stream to go, so the useful half is made and the 500 reports that the operation did not *finish*, not that nothing happened. That is now pinned by a test rather than only described (see Tests), and it is listed under Residual risk.

Boot also gained `warn_orphaned_backing_indices`, which names every `.ds-*` index no restored stream claims. It **reconciles nothing** — an orphan may hold the only copy of real data — but it removes the silence. On the same wedged data dir:

```
WARN backing indices on disk belong to no known data stream — they are unreachable
     through the data-stream API and are NOT deleted automatically; remove each with
     DELETE /<index> once you have confirmed the data is not wanted
     count=3 indices=".ds-ds3-000001,.ds-ds3-000002,.ds-ds3-000003"
```

`DELETE /.ds-ds3-000001` (and 2, 3) then clears it and `PUT /_data_stream/ds3` answers 200. Verified silent on a healthy boot and on a `chmod 000` boot (where the state did not load, so every stream is unknown and every backing index would look orphaned).

## Round 2 — the case the first pass missed (adversarial review)

The review of this PR found a hole this PR's own fix had left open, and it was the more likely half: a `cluster_state.json` that is **intact but unreadable**.

`load_cluster_state` salvaged a copy only in the `serde_json` parse-error arm. The `std::fs::read` error arm logged and returned. The maps then came up empty, and the first `PUT /_index_template/...` called `flush_cluster_state`, whose `write_file_atomic` renames `cluster_state.tmp` over the target — and `rename(2)` needs write permission on the **directory**, not on the file. So the write succeeded, answered `{"acknowledged": true}`, and unlinked a document whose bytes were perfectly good. Reproduced with `chmod 000` for the boot only, on a file holding one index template:

```
PUT /_index_template/fresh          -> 200 {"acknowledged":true}
cluster_state.corrupt.json exists   -> false
live file still contains "logs-*"   -> false      <-- the operator's template
```

The triggers are ordinary rather than exotic — a uid change on a container volume, a backup tool's chmod, `EIO`, `EMFILE` at boot — and in every one of them the file was fully recoverable *until* xerj overwrote it. It also falsified this PR body's earlier claim that "the bytes found are now copied to `cluster_state.corrupt.json` first"; that sentence has been corrected above.

Fixed by `cluster_state_loaded: AtomicBool`, set false in **both** error arms. While it is false, `ensure_cluster_state_writable` refuses every management mutation with a 500 that names the file, and the document on disk is never touched. It clears only on a boot that loads cleanly, so recovery is "fix the permissions (or move the file aside) and restart" — which the tests exercise.

Two details worth stating:

- The guard sits at the top of each mutator, not only inside `flush_cluster_state`. A `DELETE` consults the in-memory map first, and with that map empty an unguarded delete answers `404 not found` about an object sitting in the file on disk — the same lie in a different shape. `create_data_stream` / `rollover_data_stream` guard before `create_index`, so a refusal cannot strand a `.ds-*` directory with no stream in front of it.
- The corrupt-parse arm now refuses too. The salvage copy stays (un-wedging the node means moving the original aside by hand, so leave something behind that survives that step), but salvage-then-overwrite had a second hole of its own: the `!salvage.exists()` guard means a **second** corruption would be destroyed *and* not copied.

Prior art for the shape: redb latches an `AtomicBool` after a failed I/O and returns `StorageError::PreviousIo` — "Please close and re-open the database" — from every write after it (`redb/src/tree_store/page_store/cached_file.rs:125-145`, `redb/src/error.rs:65`). Apache-2.0/MIT; the shape is adapted, no code is copied.

## Prior art

`fjall/src/file.rs:17` (`fsync_directory` after every atomic rewrite) and `sled/src/metadata_store.rs:697` (leftover `*.tmp` from an interrupted rewrite is discarded on the next boot — `load_cluster_state` does the same), both from the `xerj-storage` corpus. Both Apache-2.0/MIT; neither is copied, the shape is adapted and the code is ours.

## Tests

`crates/xerj-api/tests/cluster_metadata_survives_restart.rs` — writes one of each object, drops the engine, reopens the same data dir, and compares **whole response bodies** before and after, so a partial restore cannot pass. Plus: a restored pipeline still transforms; a deleted template stays deleted; 32 concurrent PUTs all survive a restart with the right contents; a stream recovered from an interrupted rollover rolls to `-000003` instead of reissuing `-000002`; a corrupt document is preserved *and* further writes are refused until the operator removes it; and (round 2, unix) `an_unreadable_cluster_state_is_never_overwritten` drives the `chmod 000` reproduction above and asserts the file is byte-identical afterwards, that the refused PUT rolled its in-memory change back, that DELETE is refused too, and that a plain restart restores everything and re-enables writes.

Three more for defect 4: `a_backing_index_that_cannot_be_deleted_aborts_the_delete` injects the failure where it happens in the field (the directory cannot be unlinked) and asserts the 500, that the stream stays addressable, that the document on disk is byte-identical, and that the retry after the cause is fixed finishes the job — it fails on the previous commit with `left: 200, right: 500` and `{"acknowledged":true}`; `a_deleted_data_stream_leaves_no_backing_index_behind` pins the clean-path end state (nothing on disk, nothing in the document, name reusable after a restart); and `a_partly_failed_delete_still_removes_what_it_could` pins the *partial* outcome that the release note now states — a three-generation stream whose first generation cannot be unlinked loses `-000002` and `-000003`, keeps `-000001`, answers 500, stays addressable and recorded, and completes on the retry. Adding a `break` to that loop fails it, which is the point: the sentence in the CHANGELOG cannot quietly stop being true.

`crates/xerj-server/tests/cluster_metadata_survives_sigkill.rs` — the same property against `CARGO_BIN_EXE_xerj` with a real `SIGKILL`, which is the only way to prove the write happened at PUT time rather than in a shutdown hook. Unix-only.

## Gates (first pass — superseded by the re-run at the end)

| gate | result |
|---|---|
| `cargo clippy -D warnings` (xerj-engine, xerj-api, xerj-server) | clean |
| `cargo test -p xerj-engine` | lib 427/0; integration green apart from the wall-clock tests below |
| `cargo test -p xerj-api` | green (13 targets, incl. 7/7 cluster-metadata at that point; 9/9 now) |
| `cargo test -p xerj-server` | green (7 targets, incl. the SIGKILL test) |
| ES-compat YAML suite | **1366 passed / 0 failed / 3 skipped** (CI) |

Environment-dependent failures on the dev machine (load average ~190 while these ran) were checked against the unmodified branch head and flip there identically: `painless_cpu_budget` and `query_string_default_field` are wall-clock-deadline tests, `xerj-engine`'s `painless_script_limits` overflows the 2 MiB default test-thread stack in a debug build (passes under `RUST_MIN_STACK`). None of them touch cluster state; CI's Build + Test job is the authority.

## Residual risk / known gaps

Stated plainly, because these are the things this PR does **not** fix.

**Orphaned `.ds-*` indices are warned about, never cleaned up.** A data dir written by a build that predates `cluster_state.json` (data streams were not persisted at all then), or a `DELETE /_data_stream` interrupted by a build that recorded the removal first, leaves backing indices that no data-stream API can reach: `GET` and `DELETE` on the stream answer 404 while `PUT /_data_stream/<name>` answers `409 resource_already_exists_exception` naming the orphan. Boot now names them in a warning and the recovery is `DELETE /<backing-index>` per index — by hand, deliberately, because an orphan may hold the only copy of real data. Nothing is deleted or adopted automatically.

**A DELETE that fails part-way still destroys every backing index it can.** The loop does not stop at the first failure: if `.ds-orders-000001` cannot be unlinked but `-000002` and `-000003` can, the latter two are gone and the request answers 500. That is deliberate — the operator asked for the stream to be deleted, and the 500 says plainly that it did not finish — but it is *not* "nothing happened", and the release note should not be read that way. The stream stays addressable, the document on disk still names it, and the retry after the cause is fixed completes the delete.

**`aliases.json` still has both swallow shapes that `cluster_state.json` no longer has.** `load_persisted_aliases` swallows an unreadable file and `flush_aliases` will later overwrite it; and `flush_aliases` itself only `warn!`s when the write fails — it has no return value, so `delete_data_stream`'s final `self.aliases.remove(name); self.flush_aliases();` can leave the alias on disk after a 200, and the next boot resurrects an alias pointing at a stream that is gone. Pre-existing behaviour on `main`, not introduced or made worse here, and refusing there would reach index-creation paths rather than only management ones — worth its own issue rather than a silent widening of this PR.

**A `flush_cluster_state` failure during a rollover** leaves the new backing index created with the generation unpersisted. Self-healing on the next boot (defect 2 above) rather than permanent, and the operator still gets a 500 at the time of the write.

**A `flush_cluster_state` failure at the end of a DELETE** — after the backing indices are already destroyed — answers 500 and puts the stream back in memory, so memory and disk agree. The next boot restores a stream whose backing indices are gone, warns about exactly that, and a re-issued DELETE completes it. This is the recoverable half of the window by design; it is not silent and it is not permanent.

## CHANGELOG

Added under `## [Unreleased] / ### Fixed`, including the known gaps above in the release-note text itself rather than only here.

## Live re-verification of the delete-ordering fix

Re-measured against a release binary built from this branch's merged head (checked first that the binary actually contains the new code — `strings | grep "belong to no known data stream"` → 1), on a private port with a throwaway data dir. Not from a test harness: a real node, killed with `kill -9`, restarted.

**The old ordering's crash state, reproduced by hand** (stream removed from `cluster_state.json`, `.ds-ds3-000001/2/3` left on disk):

```
boot                       -> WARN backing indices on disk belong to no known data stream …
                                  count=3 indices=".ds-ds3-000001,.ds-ds3-000002,.ds-ds3-000003"
GET    /_data_stream/ds3   -> 404
DELETE /_data_stream/ds3   -> 404
PUT    /_data_stream/ds3   -> 409 resource_already_exists_exception: .ds-ds3-000001
DELETE /.ds-ds3-000001|2|3 -> 200, 200, 200
PUT    /_data_stream/ds3   -> 200
```

**The new ordering's crash state** (two backing indices already destroyed, the removal not yet recorded):

```
boot                       -> WARN restored data stream references backing indices that are
                                  not present in the data dir
                                  data_stream="ds3" missing=".ds-ds3-000001,.ds-ds3-000002"
GET    /_data_stream/ds3   -> 200
DELETE /_data_stream/ds3   -> 200        (finishes the job)
leftover .ds-ds3-* on disk -> none;  document still names ds3 -> false
```

**And the two silence claims.** A healthy boot logs no orphan warning (0 occurrences). A `chmod 000` boot logs the read failure, logs **no** orphan warning (0 — every stream would look unknown), refuses `PUT /_index_template/fresh` with a 500 naming the file, and leaves `cluster_state.json` byte-identical (sha256 before == after).

## Gate re-run after the delete-ordering fix and the merge with `main`

Measured on this head, not remembered.

| gate | result |
|---|---|
| `cargo clippy --all-targets -D warnings` (xerj-engine, xerj-api, xerj-server) | clean |
| `cargo test -p xerj-api` | 15 targets, **260 passed / 0 failed** / 2 ignored (incl. 10/10 `cluster_metadata_survives_restart`) |
| `cargo test -p xerj-server` | 8 targets, **54 passed / 0 failed**, incl. `cluster_metadata_survives_kill_dash_nine` |
| `cargo test -p xerj-engine` | 31 targets, **732 passed / 0 failed** / 23 ignored (`RUST_MIN_STACK=16777216`, for the debug-build stack overflow noted above) |
| ES-compat YAML suite | **1366 passed · 0 failed · 3 skipped · 1369 total** (local, private port `:9397`, throwaway data dir, release binary rebuilt from this merged head and checked to contain the new code) |

The conformance number is read off the runner's own banner, which prints the URL it connected to (`http://localhost:9397`) — an earlier attempt of mine reported a number from a log file a sibling job had overwritten, so the URL line is now part of the evidence rather than an assumption.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -456,6 +456,58 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   service and addressable and the operator can retry the delete once the cause
   is fixed.
 
+- **Index templates, ingest pipelines, data streams and ILM policies survive a
+  restart** ([#203](https://github.com/xerj-org/xerj/issues/203)). Index
+  templates, legacy (v1) templates, component templates, ingest pipelines, data
+  streams and ILM policies lived only in memory: `PUT /_index_template/logs`
+  answered `{"acknowledged": true}` and `GET` answered 404 after the next
+  restart, and the next index that should have matched the template was created
+  without it — with no error anywhere. All six are now persisted together in
+  `<data_dir>/cluster_state.json`, written atomically (tmp → fsync → rename →
+  fsync the parent directory) so the write is committed at the request, not in
+  a shutdown hook, and restored in `Engine::new` before the listeners come up.
+  A write that fails is rolled back in memory and answered with a 500 instead
+  of `acknowledged`. Restored pipelines are recompiled, not just re-read —
+  storing the definition alone would have left `?pipeline=x` accepted and
+  silently inert after a restart.
+
+  Four defects found while verifying it, each reproduced first:
+
+  - Concurrent management writes corrupted each other (all flushes staged
+    through one fixed `.tmp` path — 32 parallel template PUTs returned
+    `store_exception`). Snapshot-and-write is now serialized.
+  - A rollover interrupted between "backing index created" and "generation
+    persisted" wedged the data stream forever: every later rollover recomputed
+    the same name and got `409 resource_already_exists_exception`. Boot now
+    adopts the highest generation actually present on disk, warns, and persists
+    the repair once.
+  - A `cluster_state.json` that could not be **read** (EACCES after a uid
+    change on a container volume, a backup tool's chmod, EIO) came up as empty
+    maps, and the next management write renamed a snapshot of those empty maps
+    over a file whose bytes were perfectly good. The load failure now latches:
+    every management mutation is refused with a 500 naming the file, and the
+    file is not touched, until a boot loads it cleanly. The corrupt-parse arm
+    refuses too, and still keeps a `cluster_state.corrupt.json` copy.
+  - `DELETE /_data_stream/<name>` recorded the removal before destroying the
+    backing indices, so a crash in that window stranded `.ds-<name>-00000N`
+    directories that no data-stream API could reach — GET and DELETE answered
+    404 while `PUT /_data_stream/<name>` answered
+    `409 resource_already_exists_exception` permanently. The order is reversed:
+    the removal is recorded only once every backing index is gone, and a
+    backing index that cannot be deleted now aborts the DELETE with a 500
+    (it was previously swallowed and answered `acknowledged`) so the operator
+    can retry.
+
+  **Known gaps, deliberately not closed here.** `.ds-*` indices that no data
+  stream claims — left by a data dir written before this change, or by a
+  DELETE interrupted by an older build — are **not** cleaned up or adopted
+  automatically; boot now names them in a warning and the recovery is
+  `DELETE /<backing-index>` per index, because an orphan may hold the only copy
+  of real data. And `aliases.json` still has the read-error shape that was
+  fixed here for `cluster_state.json`: an unreadable aliases file is swallowed
+  at boot and overwritten by the next alias write. That is pre-existing
+  behaviour, not introduced here, and is tracked separately.
+
 - **`autoindex` no longer leaves immortal catalog entries for skipped files**
   ([#238](https://github.com/xerj-org/xerj/issues/238)). A file added after the
   resume plan was frozen is skipped and reported in the catalog — but that
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -496,17 +496,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
     the removal is recorded only once every backing index is gone, and a
     backing index that cannot be deleted now aborts the DELETE with a 500
     (it was previously swallowed and answered `acknowledged`) so the operator
-    can retry.
+    can retry. Read that 500 as "did not finish", not as "nothing happened":
+    a multi-generation stream still loses every backing index that *could* be
+    destroyed, the stream itself stays addressable, and the retry after the
+    cause is fixed completes the delete.
 
   **Known gaps, deliberately not closed here.** `.ds-*` indices that no data
   stream claims — left by a data dir written before this change, or by a
   DELETE interrupted by an older build — are **not** cleaned up or adopted
   automatically; boot now names them in a warning and the recovery is
   `DELETE /<backing-index>` per index, because an orphan may hold the only copy
-  of real data. And `aliases.json` still has the read-error shape that was
+  of real data. And `aliases.json` still has both swallow shapes that were
   fixed here for `cluster_state.json`: an unreadable aliases file is swallowed
-  at boot and overwritten by the next alias write. That is pre-existing
-  behaviour, not introduced here, and is tracked separately.
+  at boot and overwritten by the next alias write, and a *failed* alias write
+  is only logged, so an alias can survive on disk after the stream it named was
+  deleted with a 200. Both are pre-existing behaviour, not introduced here or
+  made worse here, and are tracked separately.
 
 - **`autoindex` no longer leaves immortal catalog entries for skipped files**
   ([#238](https://github.com/xerj-org/xerj/issues/238)). A file added after the
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -17374,8 +17374,10 @@ pub async fn put_index_template(
         mappings,
         priority: body.priority.unwrap_or(0),
     };
-    state.engine.templates.insert(name, tmpl);
-    Json(json!({ "acknowledged": true })).into_response()
+    match state.engine.put_index_template(name, tmpl) {
+        Ok(()) => Json(json!({ "acknowledged": true })).into_response(),
+        Err(e) => ApiError::new(xerj_common::XerjError::from(e)).into_response(),
+    }
 }
 
 pub async fn get_index_template(
@@ -17435,11 +17437,14 @@ pub async fn delete_index_template(
     State(state): State<AppState>,
     Path(name): Path<String>,
 ) -> impl IntoResponse {
-    if state.engine.templates.remove(&name).is_some() {
-        Json(json!({ "acknowledged": true })).into_response()
-    } else {
-        let e = xerj_common::XerjError::index_not_found(format!("index template [{name}] missing"));
-        ApiError::new(e).into_response()
+    match state.engine.delete_index_template(&name) {
+        Ok(true) => Json(json!({ "acknowledged": true })).into_response(),
+        Ok(false) => {
+            let e =
+                xerj_common::XerjError::index_not_found(format!("index template [{name}] missing"));
+            ApiError::new(e).into_response()
+        }
+        Err(e) => ApiError::new(xerj_common::XerjError::from(e)).into_response(),
     }
 }
 
@@ -22151,8 +22156,6 @@ pub async fn put_ingest_pipeline(
     Path(id): Path<String>,
     Json(body): Json<Value>,
 ) -> impl IntoResponse {
-    // Store the raw JSON for GET /_ingest/pipeline.
-    state.engine.pipelines.insert(id.clone(), body.clone());
     // Convert ES processor format → xerj stage format, then compile.
     // ES: {"processors": [{"set": {"field":"x","value":"y"}}]}
     // xerj: {"stages": [{"type": "set", "config": {"field":"x","value":"y"}}]}
@@ -22207,10 +22210,17 @@ pub async fn put_ingest_pipeline(
         } else {
             body.clone()
         };
-    if let Err(e) = state.engine.create_pipeline(&id, xerj_config) {
-        tracing::warn!(pipeline = %id, error = %e, "pipeline stored but failed to compile");
+    // `Some(body)` keeps this surface's long-standing behaviour: a definition
+    // that will not compile is still stored and still readable through
+    // `GET /_ingest/pipeline/{id}` (the engine logs the compile error and
+    // drops the executable form, so `?pipeline={id}` behaves identically
+    // before and after a restart). An `Err` means it did not reach disk —
+    // answering `acknowledged` there would promise a pipeline that is gone at
+    // the next boot, which is the whole of issue #203.
+    match state.engine.put_pipeline(&id, xerj_config, Some(body)) {
+        Ok(_) => Json(json!({ "acknowledged": true })).into_response(),
+        Err(e) => ApiError::new(xerj_common::XerjError::from(e)).into_response(),
     }
-    Json(json!({ "acknowledged": true })).into_response()
 }
 
 pub async fn get_ingest_pipeline(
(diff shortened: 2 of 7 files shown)


xerj-org:
Both review items addressed in 6ca0d23.

**must-fix 1 — unreadable `cluster_state.json` destroyed by the next write.** Confirmed, and reproduced before touching anything, exactly as described: `chmod 000` for the boot only, then `PUT /_index_template/fresh` →

```
PUT /_index_template/fresh          -> 200 {"acknowledged":true}
cluster_state.corrupt.json exists   -> false
live file still contains "logs-*"   -> false
```

Fixed the preferred way rather than the minimum: `cluster_state_loaded: AtomicBool`, set false in **both** error arms, and `ensure_cluster_state_writable` refuses every management mutation with a 500 that names the file. The document on disk is never touched, so a transient trigger (`EMFILE` at boot, a chmod that gets fixed) needs nothing but a restart — where renaming the file aside would have relocated a perfectly good config permanently.

Two things beyond the letter of the review:

- The guard is at the top of each mutator, not only inside `flush_cluster_state`. A `DELETE` reads the in-memory map first, and with that map empty it answered `404 not found` for an object sitting in the file on disk — the same lie in a different shape. `create_data_stream` / `rollover_data_stream` guard before `create_index` so a refusal cannot strand a `.ds-*` directory.
- The corrupt-parse arm refuses too. Salvage-then-overwrite had its own second hole: `!salvage.exists()` means a **second** corruption is destroyed *and* not copied.

Prior art for latching on a failed load: redb returns `StorageError::PreviousIo` ("Please close and re-open the database") from every write after an I/O failure — `redb/src/tree_store/page_store/cached_file.rs:125-145`, `redb/src/error.rs:65`. Apache-2.0/MIT, shape adapted, no code copied.

`an_unreadable_cluster_state_is_never_overwritten` (unix) drives the reproduction and asserts the file is byte-identical afterwards, the refused PUT rolled its in-memory change back, DELETE is refused too, and a plain restart restores everything and re-enables writes. `a_corrupt_cluster_state_is_preserved_before_it_is_overwritten` now asserts the original is left untouched and that removing it un-wedges the node.

**must-fix 2 — CI.** Both jobs that were pending at review time finished green on run 31338122854 (Build + Test 23:03Z, Autoindex FD smoke windows-latest 22:18Z), i.e. 12/12 on the reviewed commit. That run is now superseded by this push and CI is re-running from scratch; still not green until it says so.

Local gate on this commit: `cargo fmt --all -- --check` clean; `clippy -D warnings` clean on xerj-engine/xerj-api/xerj-server; `cargo test -p xerj-api` and `-p xerj-server` green (incl. 7/7 cluster-metadata and the SIGKILL test); `cargo test -p xerj-engine` lib 427/0. Four engine integration tests failed under a load average of ~190 on this box — `painless_cpu_budget`, `query_string_default_field`, `painless_script_limits` — all wall-clock/stack budget tests; each was re-run against the unmodified branch head and fails there too (baseline failed a *superset*: 4/8 vs 2/8 in `painless_cpu_budget`). None of them touch cluster state.

PR #264, opened 2026-08-09
feat(autoindex): real progress, honest percent and ETA — no run is silent (#241)

### LINKED ISSUE #241: autoindex: no progress, percent or ETA — 47-64% of every run is silent, and agents cannot tell running from hung

xerj-org:
Fixes #241

`xerj autoindex` looked hung to the AI agents that drive it, because it was silent for most of every run. This gives it a real progress contract — phase, items, bytes, percent, rate, ETA and the file it is actually waiting on — with a **time**-based liveness guarantee and separate surfaces for humans and machines.

## Reproduced first, at `origin/main` (1ee31ecd)

251 files / 35.3 MB, deliberately size-skewed the way real trees are: 250 small CSVs plus one 34.6 MB JSONL holding **98% of corpus bytes**. Local node, stdout **and** stderr both pipes (never a tty).

```
exit=0  wall=27.65s  lines=16
worst silence 16.39s [0.50 -> 16.89]  = 59.3% of wall
then a further      10.75s [16.89 -> 27.64]  across the finalize block
```

27.14 of 27.65 seconds produced no output after the first half second.

## Root cause — three defects, not one

1. **Liveness was item-driven, not time-driven.** The print fired on `dn % 200 == 0`, so silence had no upper bound: 200 files is milliseconds of source or minutes of PDFs. Worse, it lived *in the worker* — a worker blocked inside one large file cannot heartbeat itself.
2. **The `finalize` block emitted nothing**, despite every one of its loops having a bound known before it starts.
3. **Nothing in the workspace was TTY-aware** (`grep -rn IsTerminal crates/` → zero hits), so humans, agents and CI all shared one output path, and `--json` made it strictly worse: stdout stayed empty until the very end.

## The fix

A ticker thread owns the output surface; workers only bump atomics.

- **stdout = RESULT, stderr = PROGRESS.** `--json` stdout stays one clean parseable document.
- `--progress auto|plain|json|none` (default `auto`) + `--progress-interval SECS`. `auto` → live redrawn line when stderr is a terminal and `TERM != dumb` and `CI` unset; otherwise one parseable line per interval. `--quiet` is an alias for `--progress none`. `--json` and `--progress json` are orthogonal.
- **Twelve instrumented phases**, including the whole previously-silent finalize block.
- **Percent and ETA are bytes-based.** On this corpus `items=250/251` is `pct=1.9` — the correct answer when one file holds 98% of the bytes. A file-count percent would have claimed 99.6% and sat there.
- **Straggler naming**: `waiting_on=big-events.jsonl(34.6MB)`.
- **One terminal line in every progress mode except `none`.** Exit 3 is *success*; an agent reading a bare `3` off a silent stream reads failure. `--quiet` (= `--progress none`) opts out of the surface entirely and prints no terminal line — poll `autoindex status` or read the exit code there.

### Honesty rules, enforced by tests

`pct=unknown` (JSON `null`) when no denominator exists. An ETA appears only after 5 s in-phase **and** 2% of phase bytes, is `rough` below 10% and `good` above, and may not move more than 20% per tick.

Bytes are credited when a file *finishes*, so a worker grinding through one large file looks like zero throughput from outside. Folding those empty samples into the EWMA would drive the rate to zero and the ETA to infinity at exactly the moment a user is watching — so **nothing completing is treated as an absence of measurement, not a measurement of zero**. After three ticks (min 15 s) with no completion the ETA is withdrawn as `eta_quality=stalled` and the straggler is named instead.

### Contradictions are refused, not silently resolved (#204 discipline)

`--quiet --progress plain`, `--progress none --progress-interval 5`, an unknown mode, an out-of-range interval, and `--progress` on `map`/`status` are all usage errors — verified to exit 2 with an explanation rather than accepting a flag that would not be honoured.

## Result on the same corpus, machine and node

```
before: 16 lines, worst silence 16.39s (59.3% of wall)
after:  30 lines, worst silence  5.00s = exactly the configured default
        interval, tail silence 0.05s, terminal line present:
        xerj-done ok=true exit=0 reason=completed wall=17.8s
                  files=251 records=235251 datasets=2 junk_files=0
```

Both runs indexed identically — **235,251 records, 2 datasets, exit 0**. The wall times (27.65 s vs 17.89 s) differ because of cold vs warm node state on a heavily shared box; this change is about output and **no speedup is claimed** from those numbers.

Also verified by hand on the built binary:

| check | result |
|---|---|
| `--json --progress json` | stdout exactly **1** valid JSON document; stderr **34** valid JSON lines (6 note / 27 progress / 1 done) |
| `--quiet` | **0 bytes** on stderr |
| unreachable endpoint | closes with `xerj-done ok=false exit=1 reason=aborted`, then the error line, exit 1 |
| empty folder | closes with `reason=no-files` |
| under a pty | **84 in-place redraws**, where the issue's control-character census measured 0 |
| 6 malformed/contradictory flag forms | all exit 2 with an explanation |

## Honest scope

The docs state the terminal-line guarantee precisely, with both of its exceptions: **every run that reaches an exit** emits one, on success and on error — *except* under `--progress none` (which `--quiet` selects), where the surface is disabled and nothing is printed; and except when the process is killed by a signal or aborted by a panic (the release profile is `panic = "abort"`, so unwinding never reaches `Drop`), where a process that is gone with no terminal line died rather than finished. `landing/llms.txt` and `--help` both say exactly that, and point agents at `autoindex status --state-dir <dir>` or the exit code for quiet runs.

Verified against the binary: a successful `--quiet` run writes **0 bytes** to stderr; a failing one writes **only** its `error:` line; neither writes a terminal line.

## Tests

- `progress::tests` — 14 unit tests over the render/ETA/clamp/stall logic, driving injected clock state rather than sleeping.
- `failure_resume_http_tests::a_run_reports_progress_through_every_phase_and_closes_the_stream` — end-to-end through the **real** progress surface a production run uses: phase B reports, the previously-silent finalize block reports, every line carries a percent and an ETA field, and the run closes with a terminal line stating the outcome in words. **Before this change it fails on the first assertion — a run emitted no `xerj-progress` lines at all.**
- `failure_resume_http_tests::quiet_runs_emit_no_progress_stream_at_all` — the other half of the contract.
- `cli::tests` — the refusals above.

## Gate

`cargo fmt --all -- --check` clean · `cargo clippy -p xerj-autoindex -p xerj-server --all-targets -- -D warnings` clean · `cargo test -p xerj-autoindex` **280 passed / 0 failed**.

### One pre-existing flake, named rather than hidden

Under heavy load this crate's suite intermittently fails `esclient::tests::all_timeout_retry_path_is_bounded_to_six_attempts_without_final_sleep` (a wall-clock retry-timing test with a 50 ms bulk timeout; it counts 5 attempts instead of 6), which poisons a shared failpoint mutex and cascades into a dozen `PoisonError` victims. **This reproduces on `origin/main` without this branch** — 5 failing runs out of 14 under synthetic CPU load, including one 20-failure run with the identical signature. It is not introduced here and is left for a separate fix.

While chasing that I did find and fix a genuine flake **in this branch's own test**: `ticker_keeps_emitting_while_no_item_completes` originally slept 400 ms at a 40 ms cadence and asserted ≥4 ticks — a rate assertion that measures the scheduler, not the code. It failed once at load average 357 with 2 lines, where the ticker had in fact fired. It is now "wait until repeated emission is observed, fail on a 20 s deadline", and survives 40/40 runs at load average **1150**.

---

## Review repair (2026-08-10)

An adversarial review of this branch confirmed the diagnosis, the finalize instrumentation, the two peer-code citations and green CI — but found the branch introduced a **new hang on exactly the paths it claims to fix**, plus a doc claim the code does not honour. Both are fixed in `f436a6fc`.

### 1. Lost wakeup in `Progress::spawn_ticker`

The ticker loop acquired `stopped` and went straight into `wait_timeout` without ever testing the flag. A condvar notification reaches only threads already parked in `wait`, so a `stop()` landing while the ticker was elsewhere — above all **before its first wait** — was lost, and `Ticker::drop`'s `join()` blocked for a full `--progress-interval`.

Reachable on every early exit: `es.ping()` refused, the `no-files` return, a small `--dry-run`. `--progress-interval 3600` is accepted, so the ceiling was an hour of silence *after* `xerj-done` printed.

Measured on the identical loop shape extracted to a standalone `rustc -O` binary:

| loop | interval | work before drop | hangs | worst `join()` |
|---|---:|---:|---:|---:|
| old | 30 s | 0 ms | 4/5 | **30.0217 s** |
| old | 5 s | 0 ms | 11/12 | **5.0438 s** |
| old | 5 s | 1 ms | 6/12 | 5.0171 s |
| old | 5 s | 5 ms | 0/12 | 2.85 ms |
| new | 30 s | 0 ms | 0/5 | 12.37 ms |
| new | 5 s | 0 ms | 0/12 | 41.51 ms |

**Fix:** hold the guard across the loop and test the predicate before waiting. `stopped` is still released around `emit()`, so `stop()` can never block behind a write to a slow pipe.

End-to-end on the release binary at `--progress-interval 3600` (worst case):

| path | wall | terminal line |
|---|---:|---|
| unreachable endpoint | **0.01 s** | `xerj-done ok=false exit=1 reason=aborted` |
| `no-files` | **0.68 s** | `xerj-done ok=true exit=0 reason=no-files` |
| `--dry-run` | **0.24 s** | `xerj-done ok=true exit=0 reason=dry-run` |

Two regression tests, both **mutation-verified** — they fail against the old loop in 5.4 s and pass against the fix, and the whole progress suite still runs in 0.4 s:

- `a_stop_before_the_first_wait_is_not_lost` — 50 attempts at a 3600 s interval against a 5 s box, so no machine is slow enough to fail it spuriously and no scheduler lucky enough to pass it. Fifty attempts because the defect is a race (11/12, not 12/12).
- `a_stop_that_lands_while_the_ticker_is_emitting_is_not_lost` — the second window, held open deterministically: the capture sink takes the buffer mutex on every write, so owning it parks the ticker provably inside `emit()`. Notified repeatedly until it parks (one early notify would itself be lost — the whole subject), and the notifies stop before the sink is released, so nothing rescues a loop that re-parks after emitting.

Both use a time-boxed helper that deliberately does **not** join on the timeout path: joining a thread parked for an hour would make a failing test hang for an hour instead of reporting in five seconds.

### 2. "one terminal line in every mode" was false

`Progress::finish` returns early when `!self.enabled()`, so `--quiet` / `--progress none` emit no `xerj-done` — as this branch's own `quiet_runs_emit_no_progress_stream_at_all` asserts. `landing/llms.txt` then told agents to *"wait for it rather than inferring completion from silence"*, which under `--quiet` is a wait that never ends.

Corrected at every site that carried the claim — `landing/llms.txt:82`, `cli.rs` `--help`, the `Progress::finish` doc comment, the module doc, and the `lib.rs` call-site comment — to what the binary actually does, verified by running it.

### Gate

`cargo fmt --all -- --check` clean · `cargo clippy -p xerj-autoindex -p xerj-server --all-targets -- -D warnings` clean · `cargo test -p xerj-autoindex` **282 passed / 0 failed** (280 before, +2 new).


xerj-org:
## Rebased onto #257 — three semantic conflicts resolved, gate re-run

#257 (issue #240, the core/memory policy) landed first, so this branch no longer
applied. `git rebase origin/main` conflicts in `cli.rs`, `content.rs` and
`lib.rs`; `brain.rs` auto-merges. The conflicts are semantic — both PRs decide
how many workers a run gets — so each was resolved toward **both** intents
rather than toward one side.

**#257 stays authoritative.** No deleted fallback comes back, no magic number
replaces `resources::plan`, and no refusal turns back into a clamp. Re-verified
on the merged binary against a live node, not only in tests:

```
--workers 0       error: --workers needs a number from 1 to 1024
--workers 100000  error: --workers needs a number from 1 to 1024
--bulk-mb 512     error: --bulk-mb needs a number from 1 to 24
no flag           autoindex: 32 scan threads, 32 index workers   (32-core box)
```

**#264 keeps working, and now reports the number the policy actually chose.**
Three decisions worth reviewing:

1. The resource plan is announced through `Progress::note`, not a bare
   `eprintln!` gated on `!quiet`. A raw write would survive `--progress none`
   (which promises silence) and would inject a non-JSON line into
   `--progress json` (which promises one parseable stream).
2. `with_bulk_concurrency`'s announce flag is `pr.enabled()` instead of
   `!cfg.quiet` — the 429 shrink/recover lines are stderr, and stderr is the
   progress surface's now.
3. `pool::configure` is first-call-wins (rayon pools cannot be resized), so
   `cfg.scan_workers` is a *request*. In a process that already indexed once —
   `xerj-server`'s brain endpoint does exactly that — the request loses. The run
   header, the phase-A banner and `run_doc.scan_workers` therefore read
   `pool::scan_pool().current_num_threads()`, the width phase A is really
   running at.

`content.rs` carries both changes at once: `resolve_reporting` keeps the
per-file byte callback **and** runs inside the scan pool, so `--workers` bounds
phase A and the progress callback fires from the threads doing the work.

### New test, mutation-verified

`the_resource_plan_is_announced_on_the_progress_surface_with_the_width_it_got`
builds the scan pool, then asks for one thread more than it has, and asserts the
stream and the JSON summary both report the installed width. Reverting decision
3 fails it with `the run must announce the phase-A width it actually got, not 33`.
`quiet_runs_emit_no_progress_stream_at_all` is the other half: it fails if any
resource line escapes the surface.

### End-to-end on the merged binary (311 files, 9.7 MB, live node)

| check | result |
|---|---|
| `--workers 3` | header `3 scan threads, 3 index workers`; summary `workers: 3, scan_workers: 3, bulk_concurrency_final: 3` |
| `--progress plain --json` | stdout **1** valid JSON document; stderr 64 `xerj-progress` lines + `xerj-done ok=true exit=0 reason=completed wall=52.8s files=311` |
| first ETA | `phase_elapsed_s=5.1`, `eta_quality=rough` — the 5 s rule still holds |
| bytes-based percent | `items=310/311` prints `9.3%` and names `waiting on big-events.jsonl` (one file holds 90% of the corpus) |
| under a pty | 148 in-place redraws, 44 frames carrying a percent |
| `--quiet` | **0 bytes** on stderr across a full run |

### Gate (re-run on the rebased tree)

`cargo fmt --all -- --check` clean · `cargo clippy -p xerj-autoindex -p
xerj-server --all-targets -- -D warnings` clean · `cargo test -p xerj-autoindex`
**298 passed / 0 failed** (lib) plus **1 passed / 0 failed**
(`tests/scan_pool_width.rs`).

**Correction to the body above:** it claims 282 passed. That was true before
#257 merged. 298 is the number on the rebased tree — the difference is #257's
own tests arriving with the rebase, plus the one added here. The 282 figure
should be read as historical, not as the current count.

PR #274, opened 2026-08-10
[WIP] feat(autoindex): first-class Unity project indexing

gonchar:
Text-serialized scenes/prefabs/assets become one record per Unity object with guid joins and script_class/script_path on MonoBehaviours; .meta sidecars form the guid↔path table. Marker-gated walk pruning skips generated dirs (Library/, node_modules/, target/) at any nesting depth so pointing autoindex at a monorepo stays usable without per-corpus config. Binary-serialized assets stay junk (Force Text required).

## Evidence

```
```

## Checks

**Not run:** 

## Provenance

- Written by: 


xerj-org:
Reviewed this properly rather than letting it sit — thank you for it, and sorry it waited.

**First, the WIP question:** you marked this `[WIP]`. Do you want it reviewed for merge now, or would you rather keep iterating and have me leave it open? I've reviewed what's here either way, and everything below is actionable now. Nothing here is a request to change the design — the design is good.

I also approved the workflow run (fork PRs sit at `action_required` until a maintainer does), so you now have real CI output on the thread instead of nothing.

## What I ran

Worktree at 62b2b8fc, `CARGO_TARGET_DIR` on a warm shard, plus a synthetic Unity project (scene with 2 GameObjects / 1 Transform / 2 MonoBehaviours, 2 `.cs` + `.meta` pairs, a 1.2 MB BVH, `Library/`+`Logs/`+`Temp/`, a `ProjectSettings/ProjectVersion.txt` marker) indexed against a live server built from this branch.

```
cargo build --release -j 32 -p xerj-autoindex      OK
cargo build --release -j 32 -p xerj-server         OK
cargo test -p xerj-autoindex                       472 passed, 0 failed, 0 ignored
cargo fmt --all -- --check                         FAIL (11 diffs)
cargo clippy -p xerj-autoindex --all-targets -- -D warnings   FAIL (1 error)
xerj autoindex <unity-fixture> --fresh             exit 3, 5 datasets, 25 records
```

## What works, verified live

Worth saying explicitly, because most of this is right and it is a genuinely nice piece of work.

- **Marker-gated pruning does exactly what you describe.** On the fixture: `ignore rules: default:unity-generated — 3 directories pruned` (`Library/`, `Logs/`, `Temp/`), and it is named on the report rather than being invisible. Your tests correctly prove the two things that matter: a folder merely *named* `Library` without the sibling marker is left alone, and both `--no-ignore` and `--no-default-ignores` disable the rule. This is the right shape for the problem.
- **Per-object splitting works.** One dataset per Unity class — `ax-assets-scenes-gameobject` (3), `-monobehaviour`, `-transform` — with `unity_class`, `unity_class_id`, `file_id`, `name`, `ref_guids` all populated.
- **The GUID join works.** Straight out of the index:
  ```json
  {"unity_class":"MonoBehaviour","file_id":1003,"speed":7.25,
   "script_guid":"aaaa1111bbbb2222cccc3333dddd4444",
   "script_path":"Assets/Scripts/PlayerController.cs",
   "script_class":"PlayerController",
   "ax_path":"Assets/Scenes/Main.unity"}
  ```
  That is the "which scenes use this script?" query the description promises, and it is one query.
- **BVH does what the module header says.** A `walk.bvh` with a 1.2 MB MOTION block yielded exactly one record — `joints: ["Hips","LeftUpLeg"]`, `frames: 300`, `duration_s ≈ 10` — and the motion block was never pulled.
- **`--stub` works.** `--stub '*.pem.txt' --stub 'payload_*.txt'` turned each into a single name-card record.
- Owning the document boundaries yourself instead of leaning on serde_yaml's multi-doc iterator is the correct call, and the reasoning in the `unity.rs` header (anchor names are invisible through serde_yaml, so the `fileID` join key would be lost) is exactly right.

## Blockers

### 1. CLA — the one hard gate

`verification/cla-signed` is red because `@gonchar` is not in `.contributors`. One-time, one small PR adding your username, then `@cla-bot check` here. I can't merge past this one regardless of the code.

### 2. Lint is red

`Format + Clippy` fails in CI, and reproduces locally.

- `cargo fmt --all` — 11 diffs: `extract/bvh.rs:45`, `extract/unity.rs:33,86,140,179`, `sniff.rs:633,651,712,719,724`, `walk.rs:351`.
- `clippy -D warnings` — one error, `sniff.rs:726`:
  ```
  error: this `repeat().take()` can be written more concisely
      bytes.extend(std::iter::repeat(0u8).take(64));
  help: consider using `repeat_n()`: std::iter::repeat_n(0u8, 64)
  ```

### 3. The two new statistical binary heuristics drop real text files

This is the one I'd most want changed, and it is not about Unity — it changes classification for every user of `autoindex`.

Same fixture, same command, two builds (`main` @ dab6a013 vs this branch):

| file | main | this branch |
|---|---|---|
| `readme_sjis.txt` — Shift-JIS Japanese prose, 13.4 kB | 14 records | **0 — `binary content (unknown)`** |
| `readme_koi8.txt` — KOI8-R Russian prose, 13.2 kB | 11 records | **0 — `binary content (unknown)`** |
| `genome.txt` — FASTA nucleotide, 32 kB | 15 records | **0 — `binary content (unknown)`** |
| `payload_b64.txt` — base64 blob | 5 records | **0 — `binary content (unknown)`** |
| `notes_zh_utf8.txt` / `english.txt` / `server_cert.pem.txt` | unchanged | unchanged |

Two rules cause it:

- **`sniff.rs:223-231`** — `encoding != "utf-8"` and >30% non-ASCII → `Binary`. The comment's premise ("real windows-1252 text runs well under 30% non-ASCII") holds for Western European prose, but *every* legacy East-Asian and Cyrillic encoding is majority-high-byte by construction: Shift-JIS, EUC-JP, GB18030, Big5, KOI8-R. Those all decode lossily through the windows-1252 path and are now binary.
- **`sniff.rs:452-459`** (`txt_kind`) — ≥4096 chars and <5% whitespace → `Binary`. `nonblank` lines have already had their terminators stripped, so line breaks never count toward `ws`; any format whose only whitespace is the line break reads as ~0%. That is FASTA, PEM, base64, hex dumps, and anything minified. `server_cert.pem.txt` survived by luck — 4051 non-blank chars, 45 under the threshold. A slightly larger certificate is gone.

What makes this worth blocking on rather than filing: the file just becomes a junk count and the run still exits `ok`. The user pointed autoindex at a folder, the folder is now missing files, and nothing says which rule ate them. That is the pattern tracked in #204.

The magic-byte additions in the same hunk (PSD/TIFF/RIFF/Ogg/FLAC/MP3/FBX/EXR) are unambiguous and a clear win — please keep those. For the TGA-style case they don't cover, I'd suggest a discriminator that cannot be true of text at all — NUL/C0 control-byte density in the prefix, or a byte-histogram entropy check — rather than whitespace ratio and non-ASCII ratio. Whatever replaces them, a test asserting Shift-JIS and KOI8-R prose stay text would keep it honest.

### 4. A Unity document containing a >u64 decimal token is junked whole, silently

`emit_doc` (`extract/unity.rs:165`) and `extract_meta` (`extract/unity.rs:278`) both hand the body to `serde_yaml::from_str::<serde_yaml::Value>`, which refuses any unquoted decimal token that needs u128:

```
SERDE ERR: MonoBehaviour.m_Script.guid: invalid type: integer
  `99998888777766665555444433332222` as u128, expected any YAML value
  at line 2 column 38
```

A Unity GUID is 32 hex chars, so the all-decimal-digit ones land here. Measured on the fixture with the real binary:

```
Assets/Scenes/Main.unity          records=4 junk=1     ← MonoBehaviour &2002 dropped
Assets/Scripts/EnemyAI.cs.meta    records=0
    reason: no records extracted (unity-meta candidate family, 1 junk lines)
```

Probing the shapes directly:

```
GUID aaaa1111bbbb2222cccc3333dddd4444  records=1  guid=Some("aaaa1111…")
GUID 1234567890abcdef1234567890abcdef  records=1  guid=Some("1234567890abcdef…")
GUID 00000000000000000000000000000001  records=1  guid=Some("0000…0001")   ← leading zero keeps it a string
GUID 99998888777766665555444433332222  records=0  junk=1
GUID 12345678901234567890123456789012  records=0  junk=1
```

The knock-on is worse than the one lost record. `build_unity_guid_map` (`lib.rs:1598`) builds the guid→path table by re-running `extract_meta`, so a junked `.meta` silently removes that guid from the map — and then every MonoBehaviour referencing it loses `script_path`/`script_class` with no diagnostic at all. The headline feature degrades quietly rather than failing loudly.

Per-guid this is rare — `(10/16)^32 ≈ 8e-7` — so I'm not claiming it bites often. I'm flagging it because the failure is invisible, and because a GUID is textual by nature: reading it as a string (or not routing the whole document through a u128-hostile parser) fixes both the lost object and the lost enrichment.

### 5. Documented flags that don't exist

`demo/usecases/autoindex/README.md:117` says `--no-default-excludes` and `:120` says `--no-gitignore`. Neither flag exists — the CLI has `--no-ignore` and `--no-default-ignores`. Neither string is on main, so this text is new here. Also, `--stub` is a user-visible flag introduced by this PR and it isn't mentioned in either README, only in `print_help`.

## Non-blocking notes

- **`lib.rs:744`, `an_invalid_pattern_fails_loudly_at_startup`** asserts only that a *valid* pattern compiles. `glob_to_regex` (`lib.rs:666`) escapes every non-glob character, so `StubMatcher::compile` can't fail — which makes the `.expect("validated at startup")` calls safe, but leaves the test's name promising something it doesn't check. Separately: `globset` is already in the tree via `ignore`, and `GlobBuilder::new(g).literal_separator(true)` gives you these exact gitignore-flavoured semantics without the hand-rolled translation.
- **`build_unity_guid_map` (`lib.rs:1598`) is serial** on the main thread while the phases either side of it are rayon-parallel. The doc comment says "costs milliseconds"; on a real project with tens of thousands of `.meta` files that's seconds, single-threaded, with no progress line — worth a `par_iter`, particularly given the resource-aware progress work elsewhere in the tree.
- **`file_id` intent doesn't survive the mapping.** You deliberately emit it as a `String` so negative 64-bit anchors round-trip; inference types it `long` and `_source` comes back `"file_id": 1001`. Harmless today since Unity fileIDs are i64, just noting the comment and the behaviour have drifted apart.
- **Provenance.** The commits carry `Co-Authored-By: cursoragent` and the PR body's Provenance / Evidence / Checks sections are still the empty template. I'm not touching your authorship — that stays yours. But the template asks for the Provenance block (written-by / verified / assumed) when an agent wrote part of a change, and it also asks that `Co-Authored-By` trailers not be added to this repo, with the PR account as the record of accountability. Filling in what you verified by running vs. what you assumed would help a lot here, especially for the sniffer heuristics — those read like they were tuned against one corpus.

## What would make this mergeable

1. Sign the CLA (`.contributors` PR, then `@cla-bot check`).
2. `cargo fmt --all` and the one clippy fix.
3. Narrow or drop the two statistical binary heuristics so Shift-JIS/KOI8-R prose and FASTA/base64/PEM stay text; keep the magic-byte additions. A test for the legacy-encoding case.
4. Handle the u128 token case so a single GUID can't junk an object or silently break the script-link map.
5. Fix the two flag names in the demo README and mention `--stub`.
6. Fill in Evidence + Provenance in the PR body.

Items 1, 2, 5 and 6 are mechanical. Item 3 is the one I'd genuinely like your judgement on — you clearly hit a real problem there (a multi-MB PSD amplified ~100× through sectioning is a bad day) and I don't want to just delete your fix for it; I'd rather it keyed on something that text can't accidentally look like. Happy to keep going on this thread as you push.

xerj-org (commented):
Thanks for this, @gonchar — and first, a question, because you marked it **[WIP]**: **do you want this reviewed for merge now, or left open while you keep working?** I've reviewed what's on the branch either way, so you have the feedback regardless of which you pick.

Verdict on the current head (`62b2b8f`): **not mergeable yet.** There is one genuine behaviour regression underneath the red CI, and I think it is worth fixing carefully rather than papering over — details and a verified one-line candidate fix below. The Unity work itself is good; that part is not what is blocking.

## What I ran

Worktree at `62b2b8f`, `cargo` shard-local target dir:

```
cargo fmt --all -- --check         → 11 diffs
cargo clippy -p xerj-autoindex --all-targets -- -D warnings
                                   → 1 error (could not compile lib test)
cargo test  -p xerj-autoindex --lib
                                   → 471 passed; 18 failed
```

CI's Build + Test job on the same commit reports `465 passed; 24 failed` — same root cause, the extra six are scheduling-dependent (see below).

## Blocker 1 — 18 test failures, and only one of them is real

All 18 come from a single test. Run alone it is deterministic:

```
cargo test -p xerj-autoindex --lib \
  failure_resume_http_tests::legacy_key_collision_fails_before_visibility_with_scoped_guidance \
  -- --exact

panicked at crates/xerj-autoindex/src/failure_resume_http_tests.rs:1504:5:
  left: 3
 right: 0
```

`run_index` returns 3 (junk present) where the test expects 0. That test panics while holding `FAILPOINT_TEST_LOCK`, which poisons the mutex, and the other 17 (24 on CI, depending on how many reach the lock before the process winds down) then fail with `PoisonError { .. }`. So it is one defect wearing 18 hats — the cascade is not independent breakage.

## Blocker 2 — the defect behind it: CJK / Thai prose is now classified as binary

The fixture at that assertion is `b.txt`, 65,537 printable bytes with no whitespace. `sniff.rs:453-460` (new in this PR) reclassifies it:

```rust
let total_chars: usize = nonblank.iter().map(|l| l.chars().count()).sum();
if total_chars >= 4096 {
    let ws: usize = /* whitespace chars */;
    if ws * 20 < total_chars {
        return Family::Binary;
    }
}
```

The comment justifies 5% as "natural-language text runs ~15% whitespace". That holds for space-separated scripts. It does not hold for scripts that do not put spaces between words. I checked this directly by calling `sniff_bytes` on real samples:

| sample (>4096 chars) | whitespace | family on this branch |
|---|---:|---|
| Chinese prose, ~42 chars/line | 2.4% | **`Binary`** |
| Japanese prose, ~42 chars/line | 2.4% | **`Binary`** |
| Thai prose (no inter-word spaces at all) | 1.3% | **`Binary`** |
| base64 body in a `.pem` | 1.5% | **`Binary`** |
| minified CSS (`app.min.css`) | 0.0% | **`Binary`** |
| minified JS (`app.min.js`) | 0.0% | `Code` (saved by extension routing) |

A `Binary` verdict means the content is never extracted. So `xerj autoindex` over a folder of Chinese, Japanese, Thai, Lao, Khmer or Burmese documents indexes none of their text. That is the #204 shape the repo cares about — the user asked for the folder, the folder was accepted, and part of it quietly did not get indexed. This crate also just took a UTF-8 fix in `sqldump.rs` for the same family of assumption, so it feels worth getting right rather than adjusting the fixture.

**Candidate fix, verified.** Whitespace alone can't separate these, but whitespace *plus* alphabetic share can. Measured on the same samples: zh 97.6% alphabetic, th 88.2%, base64 86.2%, minified CSS 76.6%, your synthetic ASCII soup 55.9%, lossy-decoded pixel bytes 20.3%. Adding a second condition at `sniff.rs:458`:

```rust
let alpha: usize = nonblank
    .iter()
    .map(|l| l.chars().filter(|c| c.is_alphabetic()).count())
    .sum();
if ws * 20 < total_chars && alpha * 10 < total_chars * 7 {
    return Family::Binary;
}
```

I applied exactly that locally and re-ran the suite:

```
cargo test -p xerj-autoindex --lib → 489 passed; 0 failed
```

All 18 failures clear, **and** your two new regression tests stay green on their own merits — `whitespace_free_printable_soup_is_binary_not_prose` and `media_containers_are_binary_by_magic_not_heuristics` both still pass, because the ASCII soup is 55.9% alphabetic and the pixel soup 20.3%. Re-sniffing afterwards: zh/ja/th all come back `TxtProse`, `t.tga` stays `Binary`.

Take it or replace it with something you like better — the threshold is mine, not gospel, and I'd rather you own the call. But whatever lands, please add a test with non-space-separated script so this can't regress. (The magic-bytes additions and the `encoding != "utf-8"` high-byte rule above it are both sound and I'd keep them as-is.)

## Blocker 3 — fmt and clippy

`cargo fmt --all` fixes all 11 (`extract/bvh.rs:45`, `extract/unity.rs:33,86,140,179`, `sniff.rs:633,651,712,719,724`, `walk.rs:351`).

Clippy has one hard error, in the new test at `sniff.rs:726`:

```
error: this `repeat().take()` can be written more concisely
  = help: consider using `repeat_n()` instead: `std::iter::repeat_n(0u8, 64)`
  = note: `-D clippy::manual-repeat-n` implied by `-D warnings`
```

## Blocker 4 — CLA

`verification/cla-signed` is failing: the bot reports "@gonchar have not signed the CLA". That one is purely procedural and independent of the code — see [CLA.md](https://github.com/xerj-org/xerj/blob/main/CLA.md); it has to be green before anything can merge.

## Smaller things, not blockers

1. **`demo/usecases/autoindex/README.md:117,120` documents flags that don't exist.** `--no-default-excludes` and `--no-gitignore` appear nowhere in the tree; the real flags are `--no-ignore` and `--no-default-ignores` (`cli.rs:378-379`). The same paragraph says `node_modules/` is pruned "when a sibling marker proves them ... next to `package.json`" and `target/` next to `Cargo.toml`, but `MARKER_GENERATED_DIRS` (`walk.rs:35-61`) contains only the five Unity entries, all gated on `ProjectSettings/ProjectVersion.txt`. `node_modules/`/`target/` are pruned by name via the existing blunt `DEFAULT_IGNORE_PATTERNS` (`ignore_rules.rs:66,68`), which is a different mechanism with different behaviour. The PR body inherits the same claim. Worth trimming to what the code does.

2. **`lib.rs:744` `an_invalid_pattern_fails_loudly_at_startup` doesn't test an invalid pattern** — the body only asserts a *valid* one compiles. As written `glob_to_regex` escapes every metacharacter, so I don't think an invalid pattern is reachable at all; if that's right, the honest move is to rename the test to say so (and then `.expect("validated at startup")` at `lib.rs:188` and `lib.rs:1390` is provably safe rather than assumed).

3. **`--stub` is a second feature riding along in a Unity PR.** It's cleanly built and well tested, but it's a new user-facing flag that isn't in either README, and the help text says "contents are never opened" while the phase-A content digest still hashes the file. Splitting it into its own PR would make both easier to land — your call.

## What's good here

Worth saying, because most of this diff is solid:

- The splitter rationale in `extract/unity.rs:1-34` is the right call and the reason is documented properly — serde_yaml swallowing the anchor name is exactly why the `fileID` join key needs a hand-rolled boundary parse, and owning the boundaries is what makes `continue`-past-corrupt safe. The 14 tests around it cover the cases that matter (negative file IDs, `stripped`, CRLF, over-cap doc, sampling truncation vs. real junk, reserved-field collision, locator idempotence).
- Marker-gated pruning is the correct shape for names as common as `Library/` and `Logs/`: gated on a sibling marker, works at any depth, routed through the existing report accounting via `record_marker_dir` so the prune is *named* rather than silent, and honours `--no-ignore`/`--no-default-ignores`. `walk.rs:323-413` tests all four of those, including "a file named `obj` is not a directory".
- `grouped_family` including `UnityYaml` (`lib.rs:876-879`) is a real catch — without it the later Unity classes in a big scene would go unsampled and untyped.

Once 1-4 are cleared and CI is green I'm happy to take another pass and merge it.

xerj-org:
One addition to the review above, from digging into the red **Build + Test** job (24 failures in `xerj-autoindex`): they are all rooted in the same heuristic as blocker 3, not in the Unity work.

**Root cause** — `failure_resume_http_tests::legacy_key_collision_fails_before_visibility_with_scoped_guidance` uses a fixture of 65,537 `x` bytes (`b.txt`, built to exceed the legacy 64 KiB key window). The new `txt_kind` whitespace rule (`sniff.rs:452-459`: ≥4096 chars, <5% whitespace → `Binary`) now classifies that fixture as binary, so the first `run_index` returns exit 3 (unparseable files present) where the test asserts 0:

```
thread '...legacy_key_collision_fails_before_visibility_with_scoped_guidance' panicked at
crates/xerj-autoindex/src/failure_resume_http_tests.rs:1504:5:
  left: 3
 right: 0
```

That panic happens while the test holds `FAILPOINT_TEST_LOCK` and `FILE_DONE_IO_FAILPOINT_TEST_LOCK`, which poisons both global locks — the other 23 failures are all `PoisonError` cascades from that one test (verified: they pass single-threaded in isolation). Fix blocker 3's heuristic and this entire job goes green; no separate defect is hiding behind it.

xerj-org:
**Board-clearing status — this one is blocked on you, not on us.**

Three checks are red, and the order they need fixing in matters:

1. **`verification/cla-signed` — fail.** This is a required check, so nothing else matters until it passes. It is signed by opening a pull request that adds your GitHub username to [`.contributors`](https://github.com/xerj-org/xerj/blob/main/.contributors) **from your own account** — the account is the signature. See [CLA.md](https://github.com/xerj-org/xerj/blob/main/CLA.md).
2. **Format + Clippy — fail.** `cd engine && cargo fmt --all`; CI also runs `clippy -D warnings`.
3. **Build + Test — fail.** Needs a look once the above are clear.

The branch is also `CONFLICTING` against `main`, which has moved a fair way since this was opened.

Since the title still says `[WIP]`, I have left it open and touched nothing. First-class Unity project indexing is a genuinely useful direction and worth finishing. If you want a hand with the conflict or the clippy pass, say so here — but the CLA has to come from your account either way.

xerj-org:
Picked this up again on a board sweep. Nothing here reverses the earlier review — the design verdict stands, it is good work. What is new is that I did the **merge against current `main`** and ran the gate on the merge result, which surfaces three problems that no textual diff shows. That is the part worth your time.

Head is still `62b2b8fc` (nothing pushed since 2026-08-10), so the CI on the thread is stale relative to `main`.

## First: the WIP question, still open

You marked this `[WIP]` and I have asked twice without hearing back — that is completely fine, but it means I will not merge it, because merging someone's WIP without them saying it is ready is not mine to do. **So, plainly: do you want this reviewed for merge, or left open while you iterate?** Either answer is fine and neither closes it. I am not going to close this PR for inactivity; the direction is worth finishing and it will sit here until you say otherwise. If you have moved on, say that too and I will note it as up-for-grabs rather than leaving you on the hook.

## Are the red checks your fault or the `tree-sitter` base breakage?

Fair question to ask, because a lot of PRs on this repo went red for a reason that was not theirs: until 2026-08-13 `xerj-engine` pinned `tree-sitter` 0.25 while `xerj-autoindex` wanted 0.26, and since `tree-sitter` declares `links = "tree-sitter"` only one version may exist in the graph. That made every open PR red with a ~5-ok/10-fail signature regardless of content. PR #350 fixed it.

**That is not what is happening here.** On the merge result:

```
cargo tree -p xerj-autoindex -i tree-sitter   →  tree-sitter v0.26.12  (single version, resolves clean)
```

Both red checks are this branch's own, and rebasing will not clear either one. Details below, with the proof for the test job.

### `Format + Clippy` — fail, yours

```
cargo fmt --all -- --check      11 diffs, every one in a file this PR adds or touches
  extract/bvh.rs:45   extract/unity.rs:33,86,140,179
  sniff.rs:642,660,721,728,733   walk.rs:502
```

Clippy, after the merge, now has **two** errors rather than the one you had before:

```
error: this `repeat().take()` can be written more concisely
  --> crates/xerj-autoindex/src/sniff.rs:735
     bytes.extend(std::iter::repeat(0u8).take(64));   →  std::iter::repeat_n(0u8, 64)

error: large size difference between variants
  --> crates/xerj-autoindex/src/cli.rs:91   (enum Cmd)
     Index(IndexCfg) ≥ 328 bytes vs Map(MapCfg) ≥ 104
```

The second one is a genuine *merge interaction*, not something you did wrong in isolation: `main` added `api_key_file` / `max_minutes` / `approve` to `IndexCfg`, you added `stub_globs: Vec<String>`, and the sum crosses clippy's 200-byte spread threshold. Neither side trips it alone. `Box`ing the variant or `IndexCfg` is the usual fix. CI runs `cargo clippy --workspace --all-targets -- -D warnings`, so both will bite.

### `Build + Test` — fail, yours, and I proved which line

19 failures in `xerj-autoindex` on the merge (`591 passed; 19 failed`). All 19 trace to one root, exactly as I described in the earlier comment — and here is the experiment that settles it. I disabled **only** the whitespace-density rule in `txt_kind` (`sniff.rs:467`, `if ws * 20 < total_chars`), changed nothing else, and re-ran:

| | passed | failed |
|---|---:|---:|
| merge as-is | 591 | **19** |
| same tree, that one rule disabled | 609 | **1** |

The single remaining failure is your own `sniff::unity_sniff_tests::whitespace_free_printable_soup_is_binary_not_prose`, i.e. the test that asserts the rule works. Every other failure is downstream of it.

The mechanism, unchanged: `failure_resume_http_tests::legacy_key_collision_fails_before_visibility_with_scoped_guidance` (`failure_resume_http_tests.rs:1571`) writes a fixture of 65,537 `x` bytes to exceed the legacy 64 KiB key window. One line, zero whitespace, so `ws * 20 < total_chars` is `0 < 65537` and the file is classified `Binary`; `run_index` returns 3 where the test asserts 0 (`:1574`). It panics holding `FAILPOINT_TEST_LOCK` and `FILE_DONE_IO_FAILPOINT_TEST_LOCK`, poisoning both — 18 of the 19 failures are `PoisonError` cascades and pass in isolation.

So blocker 3 from the earlier review is not just a classification-quality concern; it is also literally the whole red test job. Fix the heuristic and both go away together.

## New: the merge is one textual conflict but **three** silent semantic ones

`mergeStateStatus` is `DIRTY`, but that undersells it. The textual conflict is trivial:

- `extract/mod.rs:168` — `main` changed `code::extract(path, sink)` to `code::extract(path, sn, sink)` (#294). Take `main`'s signature, keep your four new arms. That is the entire conflict.

Resolve that and the tree still does not compile, because three changes on `main` collide with your new `Family` variants without producing a conflict marker:

1. **`lib.rs:861`, `stub_sniffed()`** — `error[E0063]: missing field `logical_name` in initializer of `Sniffed``. `main` added `Sniffed::logical_name: Option<PathBuf>` in `f9d76b5b` (#294) — the logical source name, because durable prep reads content-addressed blob paths and name-derived decisions must not use the blob path. For a `--stub` file you almost certainly want `Some(path.file_name())` rather than `None`, since the whole record *is* the name card.

2. **`estimate.rs:111`, `exact_scan_bytes()`** — `error[E0004]: non-exhaustive patterns: UnityYaml, UnityMeta, Bvh and 1 more`. This file did not exist at your merge base; it arrived with `91c1f9ca` (estimate-on-the-user's-machine / ask-before-a-long-run). It needs an honest answer per family, and yours have interesting ones: `Bvh` deliberately never reads the MOTION block, so `size` is *not* what it scanned; `Stub` never opens the file at all. Returning `Some(size)` for either would inflate the user-facing estimate against the one thing that module exists to get right.

3. **`order.rs:183`, `band()`** — same `E0004`. Also from `91c1f9ca`; decides which band phase B works first.

And one that compiles but is probably wrong: **`order.rs:199`, `band_from_family_str()`** matches on the family *string* the durable plan persists, with `_ => Band::Bulk`. Your `Family::as_str()` values are `"unity"`, `"unity-meta"`, `"bvh"`, `"stub"`, none of which are listed — so on the resume path a Unity scene is ranked in the bulk band even if `band()` ranks it higher. Worth making those two agree; the divergence is invisible until someone resumes a large run.

I patched all four locally purely to get the suite to run, and I am deliberately not pushing that — the estimate/band answers are design calls that should be yours, not a reviewer's guess.

## Everything else from the earlier review still stands

Re-verified on the branch, briefly: `demo/usecases/autoindex/README.md:117` still says `--no-default-excludes` and `:120` still says `--no-gitignore`; neither flag exists (`--no-ignore` / `--no-default-ignores` do). `--stub` is still documented only in `print_help` and in neither README. The u128-GUID junking in `extract/unity.rs` is unchanged. `@gonchar` is still not in `.contributors` and I do not see a CLA PR — that one is a required check and blocks merge on its own no matter how green everything else gets.

## What would make this mergeable, in the order that unblocks the most

1. **CLA** — a one-line PR adding your username to `.contributors` from your own account, then `@cla-bot check` here. Hard gate, independent of code.
2. **Narrow or drop the two statistical binary heuristics.** This is one change that fixes the classification regression (Shift-JIS / KOI8-R prose, FASTA, base64, PEM) *and* turns the entire red test job green. Keep the magic-byte additions — those are unambiguous and a clear win. For the TGA case you were actually solving, key on something text cannot accidentally look like: NUL/C0 control-byte density in the prefix, or a byte-histogram entropy check.
3. **Merge `main` and answer the three semantic collisions** above, particularly the `estimate.rs` rates for `Bvh` and `Stub`.
4. `cargo fmt --all`, plus the two clippy fixes.
5. u128-safe GUID handling so one decimal-looking GUID cannot junk an object and silently drop it from the script-link map.
6. README flag names, `--stub` documented, and the Evidence / Provenance blocks in the PR body filled in.

Items 1, 4 and 6 are mechanical. Item 2 is the high-leverage one. Item 3 is new since you last looked and is the reason a plain rebase will not go green.

Happy to keep going on this thread as you push, and happy to take the fmt/clippy/README pass off your hands if you would rather only do the judgement calls — say the word. The CLA has to come from your account either way.

xerj-org:
Reviewed by two agent passes (Claude Opus 5). Still `[WIP]`, so this is feedback rather than a merge verdict.

First, credit where it is due: the Unity extractor itself is the strongest part of this and it is genuinely well built. It owns the document boundaries so a corrupt document cannot loop, ties the per-document cap to `MAX_LINE` with a const assert, has an explicit reserved-name collision policy, and distinguishes sampling truncation from real corruption. Almost every problem below is in how it is wired into the pipeline, or in the unrelated riders, not in the extractor.

## Blockers

**1. The merged workspace does not compile.** This is the #352 scenario exactly. The PR adds four `Family` variants to a non-`#[non_exhaustive]` enum, but two exhaustive `match family` expressions landed on main after this branch forked from `dab6a01`:

- `src/order.rs:183-193` (`order::band`) ends at `Family::Binary => Band::Vendored`, no `_` arm
- `src/estimate.rs:111-132` (`exact_scan_bytes`) ends at `Family::Binary => None`, no `_` arm

Neither file exists in this branch, so there was no way to see them. E0004 twice after merge. There is also a textual conflict in `extract/mod.rs`: main calls `code::extract(path, sn, sink)`, this branch calls `code::extract(path, sink)`. Rebase on current main and both surface immediately.

**2. The sniffer rider silently junks CJK text globally.** This is the one to look at first, because it is a Unity PR breaking non-Unity users. `sniff.rs:453-460`:

```rust
if total_chars >= 4096 { if ws * 20 < total_chars { return Family::Binary } }
```

`nonblank` is built from `text.lines()`, which strips newlines, so the only whitespace counted is intra-line. Chinese, Japanese, Korean, Thai, Lao, Khmer and Burmese prose have essentially none, so any such file over ~4 KB becomes `Family::Binary` and `scan_file` turns that into `junk: binary content (unknown)`. The justifying comment says natural-language text runs ~15% whitespace, which holds only for space-delimited scripts. Both accompanying tests use Latin prose and ASCII soup, so they pass either way. The sibling guard at `sniff.rs:222-232` does the same to Shift-JIS/GBK/Big5.

Related, smaller: `ID3` is a 3-byte magic checked before the text heuristics, so a CSV whose first header cell is `ID3` is junked as an mp3. `RIFF` / `8BPS` / `Kaydara FBX Binary` are long enough to be safe.

**3. Unity sampling is capped at 4 MiB, which drops records the feature is for.** `lib.rs:858-865` puts `UnityYaml` in the `SAMPLE_LIMIT_BYTES = 4 << 20` bucket (SqlDump gets 64 MiB), and `open_reader` implements that as a hard `.take(n)`. `unity.rs`'s own module doc says real scenes exceed 200 MB. A class whose first document starts past 4 MiB never becomes a cluster, so `fa.assignments` has no entry, and phase B falls through to `file_junk += 1` at `lib.rs:3617-3620`. The `grouped_family` flag added at `lib.rs:876-899` removes the per-group record stop, not the byte cap. The comment at `lib.rs:872-875` describes this failure and then does not prevent it.

**4. The GUID join fails silently, on the headline query.** `build_unity_guid_map` does `let _ = extract::unity::extract_meta(...)`. A `script_guid` that fails to resolve produces no counter, no warning, no report line. The record ships without `script_path`/`script_class`, so "which scenes use this script?" returns empty, indistinguishable from a script with no users. Given the README advertises this as a one-query answer, that needs at minimum a counter in the run summary.

**5. `script_path`/`script_class` are typed only if the sample happened to see them.** `lib.rs:1518` gates mapping registration on `c.fields.contains_key("script_guid")`, but `enrich_unity_fields` stamps the fields unconditionally at `lib.rs:3632`, after `coerce_record`, so they bypass coercion too. Combined with (3), a cluster whose sampled window had no `m_Script` gets these dynamically mapped, which feeds the field-budget overshoot in #312.

## Worth fixing before review, not blockers

- `demo/usecases/autoindex/README.md` names `--no-default-excludes` and `--no-gitignore`; the branch's `cli.rs` has `--no-ignore` and `--no-default-ignores`, and `cli::parse` hard-errors on unknown arguments (`cli.rs:432`). It also documents marker-gated pruning of `node_modules/` and `target/`, which commit `3849451` retired.
- `build_unity_guid_map` (`lib.rs:1598`) runs serially outside `crate::pool` and outside the progress meter, opening and parsing every `.meta` on every run including resumed no-ops. A real project has 10k to 500k sidecars. The doc comment says milliseconds. This is the unattributed-stretch pattern from #241.
- `ref_guids` and `joints` are multi-valued keyword arrays, and #332 says those flatten to one space-joined FTS token and stop matching after flush. Both headline queries depend on them and no test in the PR flushes.
- `stub_matcher_tests::an_invalid_pattern_fails_loudly_at_startup` never constructs an invalid pattern. `glob_to_regex` escapes every metacharacter, so no glob can produce an invalid regex, which makes the `Result`, the error message and both `.expect("validated at startup")` sites guard an impossible condition.
- No integration test covers the pipeline half. All ~22 new tests are unit tests of the extractors and matchers; nothing exercises `build_unity_guid_map` + `enrich_unity_fields` end to end, which is what would have caught (5).

## Suggestion

Split it. The BVH extractor, the `--stub` flag and the sniffer rewrite are three separable features riding a Unity PR, and the sniffer rewrite is the risky one. On its own it is easy to review, revert and bisect; buried in a 1,490-line Unity change it is not. The Unity work would land much faster without it.

xerj-org:
Thanks for this @gonchar — the Unity design here is good work, and it is now relanded as #378 on a xerj-org branch so maintainers can iterate on it (this PR targets a fork branch we cannot push to).

Your authorship is credited in #378's body and commit message. Leaving this open for you; #378 carries the same work rebased onto current `main` (it had drifted 118 commits and no longer compiled — `order.rs` and `estimate.rs` gained exhaustive `match family` arms after your fork point, and `Sniffed` gained `logical_name`), plus fixes for the sampling-cap junking, the discarded `extract_meta` `Result`, and the sample-conditional `script_path`/`script_class` mapping.

One change worth flagging explicitly: the whitespace-density binary heuristic was dropped. `nonblank` is built from `text.lines()`, so only intra-line whitespace is counted — which made CJK/Thai/Lao/Khmer/Burmese prose, base64 and FASTA all score 0% and get junked as binary. The repo's own `legacy_key_collision_fails_before_visibility_with_scoped_guidance` fixture (65,537 repeated ASCII bytes) turns that run from exit 0 into exit 3. Full reasoning in #378.

xerj-org:
Thank you @gonchar — this landed in #378, merged as e02a7265.

The Unity work here is good, and the extractor in particular is the strongest part: it owns the document boundaries so a corrupt document cannot loop, ties the per-document cap to `MAX_LINE` with a const assert, has an explicit reserved-name collision policy, and distinguishes sampling truncation from real corruption. Almost everything that needed fixing was in how it was wired into the pipeline, or in the riders, not in the extractor.

## Why a new PR rather than merging this one

This branch is on a fork and had drifted: it forked from `dab6a01`, and two exhaustive `match family` expressions landed on `main` afterwards (`order.rs::band`, `estimate.rs::exact_scan_bytes`). Adding four `Family` variants to a non-`#[non_exhaustive]` enum made the merged workspace fail to compile with E0004 twice — issue #352's exact class, and invisible from this branch because neither file exists in it. Re-landing on a `xerj-org` branch was the practical way to rebase, fix and re-verify.

## What changed on the way in

Every original finding was addressed in code, not waived:

- the two `match family` breaks, plus a third (`Sniffed::logical_name`, E0063) found during the rebase
- **the CJK regression** — both byte-statistics guards removed. A >4 KB Chinese/Japanese/Korean/Thai document was being classified `Family::Binary` and silently junked; that is now pinned by regression tests
- the 4 MiB Unity sampling cap raised to 512 MiB with a testable override, and unsampled groups are named in the report
- `build_unity_guid_map` no longer discards its `Result` — unreadable, no-guid and unresolvable are all counted and reported
- `script_path`/`script_class` now registered for every UnityYaml cluster, and enrichment moved before coercion
- the guid map runs on `crate::pool` under its own progress phase instead of serially off-meter
- README flags corrected and pinned by a test that every flag it names is a real flag
- `an_invalid_pattern_fails_loudly_at_startup` now constructs a genuinely invalid pattern
- `unity_pipeline_tests` covers the pipeline half end to end
- six printable-ASCII magic signatures gained structural qualifiers, and a real `looks_like_tga_header` replaced the byte-statistics guess

Two things you were right about that I initially got wrong in review: `an_invalid_pattern_fails_loudly_at_startup` **is** reachable (`?`×200000 and `**`×50000 do produce an invalid regex — measured), and the scope criticism was applied surgically rather than wholesale, dropping only the whitespace-density guard.

## Follow-ups filed rather than smuggled in

- #379 — `BM` and `GIF8` are unqualified printable magic on `main`; a CSV with a `BMW` column header is silently junked
- #380 — qualifying those two
- #381 — bounding what a magic-less binary costs (the record-count explosion, not memory: `for_each_section` streams)

Closing this in favour of #378. The commit history there credits this PR.

xerj-org:
@gonchar — one more note, because the record here is unfair to you. Your Unity extractor is on `main` (relanded as #378 with the wiring fixes), but the reland was committed by us, so GitHub's contributor graph does not show you — only a commit authored by your own account puts you there, and the CLA gate is what stopped us merging yours as-is.

The fix is small and on your side: open a pull request from your account adding `"gonchar"` to [`.contributors`](https://github.com/xerj-org/xerj/blob/main/.contributors) (that pull request is the CLA signature, per [CLA.md](https://github.com/xerj-org/xerj/blob/main/CLA.md)). We merge it as your commit, you appear on the front page, and any Unity follow-up — the whitespace-density heuristic revisited, `.asset` coverage, anything — lands under your name. No response needed if you'd rather not; the offer stands whenever you want it.

xerj-org:
Merged into `main` as history (7444861e): your four commits are now ancestors of main, so the contributor graph credits the Unity extractor to its author. The tree did not change — the code itself has been on main since #378. Nothing you need to do; the CLA note above applies to future work only.

PR #284, opened 2026-08-10
autoindex: recognise .mts/.cts as TypeScript

pradaev:

## Defect

`registry()` claims `ts` for TypeScript and `js`/`jsx`/`mjs`/`cjs` for JavaScript.
TypeScript's own ESM/CJS extensions, `.mts` and `.cts`, are absent — even though
they are the exact counterparts of the `.mjs`/`.cjs` already claimed.

The consequence is larger than a skipped AST pass. `is_code_ext` is what tells the
sniffer a file is code at all, so an unclaimed extension does not degrade to
"code without symbols" — it falls through to the prose extractor. A `.mts` file is
indexed with no `language`, no `defs`, no `symbols`, and is split into several
body-only records instead of one document. It is invisible to symbol search, and
anything that treats one hit as one file counts it several times.

Node-ESM TypeScript projects use `.mts` for entry points and test suites — the
modules most likely to be looked up by name.

## Fix

One registry row: `&["ts"]` → `&["ts", "mts", "cts"]`, plus a comment recording
why the omission was not harmless.

## Failing test first

`typescript_module_extensions`, run against unfixed `main`:

```
---- extract::code::tests::typescript_module_extensions stdout ----
thread 'extract::code::tests::typescript_module_extensions' panicked at
crates/xerj-autoindex/src/extract/code.rs:458:9:
mts must be recognised as code

```

(That run also contained the two failures fixed by the sibling PRs; this branch
carries only the extension test.)

- `cargo test -p xerj-autoindex --lib extract::code` on this branch → `ok. 25 passed; 0 failed`
- ES-YAML conformance suite -> `1366 passed · 0 failed · 3 skipped · 1369 total`,
  runner exit 0. Disclosure: I ran the gate once, on the `fix/php-enum-and-const`
  branch's tree, not separately on this one. All three branches change the same
  crate (`xerj-autoindex`) and touch no ES wire surface — this one is
  a single registry row.

## Assumed — NOT tested

- That `.cts` appears in real corpora at a rate worth claiming. I added it for
  symmetry with `.cjs` and because the grammar handles it identically; I measured
  `.mts` only. Falsified by a corpus where `.cts` parses differently under
  `LANGUAGE_TYPESCRIPT`.

## Evidence boundary

The behaviour was first observed while indexing a private TypeScript monorepo
(~1,600 first-party TS-family files), where 279 `.mts` files carried no
`language`, `defs` or `symbols`. That corpus is not shareable, so the committed
test — not the number — is the reproducible evidence here.

## Not run

- The crate's wider suite has a pre-existing, platform-specific failure on macOS,
  `content::tests::lossy_display_collisions_keep_distinct_alias_identities`
  (`Os { code: 92, "Illegal byte sequence" }` — APFS rejects the non-UTF-8
  filename the test creates). It reproduces on unmodified `main` in this
  environment and is unrelated to this change.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -66,9 +66,15 @@ fn registry() -> &'static [LangDef] {
                 tree_sitter_javascript::LANGUAGE.into(),
                 JS_Q,
             ),
+            // `.mts`/`.cts` are TypeScript's ESM/CJS extensions — the exact
+            // counterparts of the `.mjs`/`.cjs` that javascript already claims
+            // below. Omitting them did not merely skip the AST pass: an
+            // unclaimed extension is not code to the sniffer at all, so those
+            // files fell through to the prose extractor and were chunked into
+            // several body-only records with no `language`, `defs` or `symbols`.
             def(
                 "typescript",
-                &["ts"],
+                &["ts", "mts", "cts"],
                 tree_sitter_typescript::LANGUAGE_TYPESCRIPT.into(),
                 TS_Q,
             ),
@@ -583,6 +589,26 @@ mod tests {
         let s = syms("typescript", "export const Button = () => 1;\n");
         assert!(has(&s, "Button", "function"), "got {s:?}");
         assert!(has(&s, "Button", "const"), "got {s:?}");
+    }
+
+    /// `.mts`/`.cts` are TypeScript's ESM/CJS file extensions, the exact
+    /// counterparts of `.mjs`/`.cjs` which javascript already claims. Without
+    /// them a Node-ESM TypeScript file is not code to the sniffer at all: it
+    /// falls through to the prose extractor, so it yields no `language`, no
+    /// `defs`, no `symbols`, and gets chunked into several records instead of
+    /// one.
+    #[test]
+    fn typescript_module_extensions() {
+        assert!(is_code_ext("mts"), "mts must be recognised as code");
+        assert!(is_code_ext("cts"), "cts must be recognised as code");
+        for ext in ["mts", "cts"] {
+            let d = registry().iter().find(|d| d.exts.contains(&ext));
+            assert_eq!(
+                d.map(|d| d.name),
+                Some("typescript"),
+                "{ext} must route to the typescript grammar"
+            );
+        }
     }
     #[test]
     fn rust() {


pradaev:
@cla-bot check

xerj-org:
Reviewed and verified end-to-end. Thanks for a genuinely well-written first PR — the
failing-test-first section and the explicit **Assumed — NOT tested** boundary made this
fast to check, and both held up.

## What I ran

Worktree on your head `02e8422` (already sitting on current `main` `bc01dd7a`, no rebase
needed), release build of `xerj-server`, live server on a scratch data dir.

**The test binds.** I reverted your registry row back to `&["ts"]` and re-ran it — it
fails exactly as your body reports, same assertion, same message:

```
thread 'extract::code::tests::typescript_module_extensions' panicked at
crates/xerj-autoindex/src/extract/code.rs:602:9:
mts must be recognised as code
```

(line 602 rather than your 458 — the two sibling PRs landed ahead of it.) Restored, green
again. A test that only passes is not evidence; one that fails when the fix is removed is.

**The sniffer really routes them.** `xerj autoindex --dry-run` on a corpus of
`server.mts` / `legacy.cts` / `control.ts`:

```
"rel": "server.mts",  "family": "code"
"rel": "legacy.cts",  "family": "code"
"rel": "control.ts",  "family": "code"
```

I traced the gate to confirm there is no second list to drift out of sync: `walk.rs`
filters only on ignore rules (no extension allowlist), `sniff.rs:191` is the sole caller
of `is_code_ext`, and `code.rs:122` re-selects the same registry row. One source of truth.

**A `.mts` file ends up searchable.** Indexed the same corpus for real and searched by
symbol:

```
$ GET /mgraph-docs/_search  {"query":{"match":{"defs":"MtsGateway"}}}
hits: 1
  server.mts | typescript | 'interface StartOptions\nconst DEFAULT_PORT\n
                             function startServer\nclass MtsGateway\nmethod handle'
```

and `.cts` too — so your "Assumed — NOT tested" item is now tested, not assumed:

```
legacy.cts | typescript | 'const CTS_MARKER\nfunction ctsHelper\nclass CtsAdapter\nmethod run'
```

`language`, `defs` and `symbols` all populated, one code document per file. Falsification
attempt for your `.cts` assumption: it parses identically under `LANGUAGE_TYPESCRIPT`, so
the symmetry argument holds.

**The one collision I went looking for is safe.** `.mts` is also AVCHD/MPEG-TS video from
camcorders, so claiming the extension could in principle feed a binary blob to the
TypeScript parser. It does not — the binary guards in `sniff.rs` run at steps 1–2 and the
code-extension check only at 2b, so a synthetic 188-byte-packet MPEG-TS file lands as:

```
"rel": "clip.mts",  "format": "binary",  "reason": "binary content (unknown)"
```

Worth knowing the ordering is what protects that, since it is load-bearing for any future
extension claim too.

**Gate** (Linux, this environment):

- `cargo test -p xerj-autoindex` → **473 passed / 0 failed**

One run of the full suite gave 3 failures in `state::sync_journal_tests`; they pass in
isolation and passed on two consecutive re-runs (`--test-threads=1` included), so that is
parallelism flake on a loaded box, not your change. Your macOS
`lossy_display_collisions_...` note is a different, known platform issue — it does not
arise on Linux.

On your ES-YAML disclosure — you were right to flag that you ran it on the sibling
branch's tree rather than this one. No need to re-run it: CI executed that job against
this exact head `02e84229` and it reports the same figures you did, so the gap is closed
by evidence rather than by assumption:

```
```

## One correction to the body, and a bug your PR led me to

Your stated consequence — prose fallthrough, no `language`/`defs`/`symbols`, chunked into
body-only records — is right on the default (graph-enabled) path, which is what users hit.

It is **not** the whole story on `--no-graph`, and chasing that turned up a real defect
that is **not yours**: on the durable-generation path every code file — `.py`, `.rs`,
`.ts`, `.mts`, `.cts` alike — indexes as **zero documents** while the run reports success.
`sync_executor.rs:1445` sniffs with the logical path but `:1529` extracts with the
extensionless content-addressed blob, and `code::extract` is the one extractor that
recovers its grammar from `path.extension()`. Filed as #294 with a repro.

Worth knowing because it slightly narrows your claim: under `--no-graph` specifically, a
`.mts` goes from "one prose record holding the source text" to "no record", until #294 is
fixed. That is not a reason to hold this — the same hole already swallows every `.ts` and
`.rs` in that mode, and your change makes `.mts` consistent with `.ts` rather than
special-casing it. On the default path this is a strict improvement.

Merging. Please do open the next one — this was a good find, and the evidence discipline
in the write-up is exactly what makes a contribution cheap to review.

PR #285, opened 2026-08-10
autoindex: capture exported module const in TypeScript

pradaev:

**Rescoped after benchmarking.** The first revision of this branch also captured
module-private `const`. I measured it, found no retrieval gain, and narrowed the
change. Details and numbers below — the negative result is the reason for the
current scope.

## Defect

`TS_Q` captured a `const` only when its value was an `arrow_function`. That covers a
component or a handler and misses how modern TypeScript declares a module's public
surface:

```ts
export const users  = pgTable("users", { ... });   // schema / table
export const router = createRouter({ ... });       // router
```

Neither is an arrow function, so neither reached `defs`. A module whose whole public
surface is built by factory calls — a database schema, a route table, a config object —
extracted **zero** symbols and could not be found by symbol search at all. That is
#170's failure mode in the language where the idiom is most common.

## Fix

```
(program (export_statement declaration: (lexical_declaration "const" (variable_declarator name: (identifier) @const))))
```

Three boundaries, each pinned by an assertion:

- **Exported only.** `defs` is the cross-file retrieval surface: what another module can
  name is what another module can search for. A module-private constant is reachable by
  `body` search, by a caller already in the right file.
- **Anchored** to `program` through `export_statement`. `lexical_declaration` is the same
  node inside a function body, so an unanchored pattern would fill `defs` with every
  function-local `const` — the 90%-locals trap measured for C in #170.
- **`"const"` token matched**, so `let` stays out even when exported: a mutable module
  binding is state, not a constant.

An arrow-valued exported const is captured twice on purpose, `@function` and `@const`.
`defs` dedups on (kind, name), so the file answers both `function Button` and
`const Button`; `typescript_arrow_const_is_both_kinds` pins it.

## Why not the broader scope — the measurement that changed this PR

The first revision also captured module-private `const`. I benchmarked it end to end
against the unpatched engine: same 8 retrieval tasks, same corpus, same prompts, same
model, index rebuilt with the patched binary, one run per task.

| metric | unpatched index | patched (broad scope) |
|---|---|---|
| tokens | 347,221 | 353,521 (**+1.8%**) |
| tool calls | 69 | 74 (+7.2%) |
| wall time | 481 s | 510 s (+6.0%) |
| search round-trips | 53 | **43 (−18.9%)** |
| answer accuracy | 31.5 / 32 | 31.5 / 32 (unchanged) |

**No token improvement.** Per-task deltas ran −11.2% to +11.6%, so +1.8% is inside the
noise of a single run per task — the honest reading is "no effect detected", not
"regression". The one metric outside that spread favours the change (19% fewer search
round-trips at equal accuracy), which is consistent with symbols letting a caller land in
one hop; but each response now carries a larger `symbols` array, and that appears to eat
the saving.

The task I expected to benefit most — "list all 32 tables in the schema module", whose
target file went from zero symbols to 32 — got **7.9% more expensive**: the agent
received the symbols and still re-verified with grep.

Given no demonstrated retrieval gain, I cut the half of the capture that costs the most
and buys the least. Measured on the same corpus (1,634 TS-family files):

| scope | files gaining const symbols | const symbols added |
|---|---|---|
| private + exported | 918 | 2,382 |
| **exported only (this PR)** | **238** | **518** |

**78% less added weight, and the motivating case is untouched** — the schema module still
yields all 32 tables with exact line numbers.

## Failing test first

`typescript_module_const`, run against unfixed `main` — the extractor found only the
function and none of the module constants:

```
---- extract::code::tests::typescript_module_const stdout ----
thread 'extract::code::tests::typescript_module_const' panicked at
crates/xerj-autoindex/src/extract/code.rs:487:9:
got [("f", "function", 4)]

```

- `cargo test -p xerj-autoindex --lib extract::code` on this revision → `ok. 26 passed; 0 failed`
- ES-YAML conformance suite on this revision's tree:
  `./target/release/es-yaml-runner --dir tests/es-compat-yaml/yaml`
  → `1366 passed · 0 failed · 3 skipped · 1369 total`, runner exit 0
- Scope comparison (918/2,382 vs 238/518) produced by running `parse_symbols` over every
  `.ts`/`.tsx`/`.mts` file of the corpus under each query, in a throwaway test.
- Node shapes confirmed via `to_sexp()` before writing the pattern: an exported module
  const is `(program (export_statement declaration: (lexical_declaration (variable_declarator …))))`,
  a function-local one sits under `(statement_block …)`.

## Assumed — NOT tested

- That module-private constants are not worth capturing **in general**. I measured one
  corpus and 8 tasks; a codebase that keeps its tuning knobs private and greps for them by
  name would want the broader scope. Falsified by a corpus where private constants are the
  thing callers search for.
- Namespace/ambient blocks (`declare module { export const X = 1 }`) are not captured —
  not direct children of `program`. Frequency not measured.
- `var` at module scope is not captured (different node). Treated as legacy.

## Evidence boundary

The benchmark and the scope comparison ran against a private TypeScript monorepo that is
not shareable; the committed tests are the reproducible evidence, and the percentages are
context for the scoping decision rather than something you can re-run here.

## Not run

- Pre-existing macOS-only failure in the crate's wider suite
  (`content::tests::lossy_display_collisions_keep_distinct_alias_identities`,
  `Os { code: 92, "Illegal byte sequence" }`), reproduced on unmodified `main` in this
  environment and unrelated to this change.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -237,6 +237,42 @@ const JS_Q: &str = r#"
 (export_statement declaration: (lexical_declaration (variable_declarator name: (identifier) @function value: (arrow_function))))
 "#;
 
+// The last pattern is #170's failure mode in the language where it is most
+// common. Modern TypeScript declares most of a module's public surface as
+// `export const x = someBuilder(...)` — schema/table definitions, routers,
+// config objects. That is not an `arrow_function`, so the arrow-valued pattern
+// above missed it, and a module made entirely of such declarations extracted
+// ZERO symbols.
+//
+// Scoped to EXPORTED module constants, deliberately, and narrower than the
+// obvious fix:
+//
+//   * `program > export_statement` only. A bare `program > lexical_declaration`
+//     would also pick up module-private constants. That was measured and
+//     dropped — see the note below — because `defs` is a cross-file retrieval
+//     surface: what another module can name is what another module can search
+//     for. A private constant is found by `body` search, by a caller who is
+//     already in the right file.
+//   * Anchored, not free-floating: `lexical_declaration` is the same node
+//     inside a function body, so an unanchored pattern would fill `defs` with
+//     every function-local `const` — the 90%-locals trap measured for C in #170.
+//   * `"const"` token matched, so `let` stays out: `lexical_declaration` covers
+//     both, and a mutable module binding is state, not a constant.
+//
+// Measured, and the reason for the narrower scope: on a private TypeScript
+// monorepo (~1,600 TS-family files) the broad form was benchmarked against the
+// unpatched engine across 8 retrieval tasks and produced no token improvement
+// (+1.8%, inside a per-task spread of -11%..+12%), at unchanged answer accuracy.
+// It did cut search round-trips by 19%, but every extra symbol is paid for on
+// every response that carries the array. With no demonstrated retrieval gain,
+// the private half of the capture was not worth its weight; the exported half
+// is what closes the zero-symbol hole.
+//
+// An arrow-valued exported const is captured twice on purpose — `@function` by
+// the pattern above and `@const` here. `defs` dedups on (kind, name), so the
+// file becomes reachable by both `function Button` and `const Button`, which is
+// what a caller searching for either would expect; suppressing one would mean
+// choosing which of the two true statements to hide.
 const TS_Q: &str = r#"
 (function_declaration name: (identifier) @function)
 (class_declaration name: (type_identifier) @class)
@@ -245,6 +281,7 @@ const TS_Q: &str = r#"
 (type_alias_declaration name: (type_identifier) @type)
 (enum_declaration name: (identifier) @enum)
 (variable_declarator name: (identifier) @function value: (arrow_function))
+(program (export_statement declaration: (lexical_declaration "const" (variable_declarator name: (identifier) @const))))
 "#;
 
 // `const_item` / `static_item` are captured because a file whose whole content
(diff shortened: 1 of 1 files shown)


pradaev:
@cla-bot check

xerj-org:
Merging this. Welcome to XERJ, @pradaev — and thank you, this is an unusually good first contribution.

I want to be specific about *why* it's good, because the thing that made this easy to review is the thing most PRs get wrong: **you shipped a negative result and let it shrink your own change.** The first revision captured module-private `const` too, you measured it, it bought nothing, and you cut it. Reviewers can't verify a claim like "this improves retrieval" from a diff — but they can verify a scope boundary that's pinned by an assertion. You gave me the second thing. That's the right instinct for this repo.

## What I ran

Worktree off your head `11d5a519`, cargo shard 4.

I didn't trust the unit test — I ran the **real extractor entry point** (`extract::code::extract`, so `read_whole` → `decode_text` → `parse_symbols` → `emit_code_doc`) over real `.ts`/`.tsx`/`.js` files on disk, once with your pattern and once with that one line deleted, same files both times.

A realistic Drizzle schema module (`schema.ts` — 3 `pgTable` calls, 1 `relations`, one private const, one `export let`):

| | before | after |
|---|---|---|
| `symbol_count` | *(field absent)* | `4` |
| `defs` | *(field absent)* | `const users` / `const organizations` / `const memberships` / `const usersRelations` |

Not "fewer symbols" — the document carried **no `defs` and no `symbols` field at all**, so it was unreachable by symbol search entirely. That is #170's failure mode exactly as you described it, and this closes it. Line numbers were correct (L4/L10/L15/L20).

Every boundary you claimed, checked against the real parser rather than the test fixture:

- `INTERNAL_TABLE_PREFIX` (module-private) — excluded ✓
- `export let migrationCursor` — excluded ✓
- `const RAW` / `const localOnly` inside function bodies — excluded ✓ (the locals trap stays shut)
- `export const MAX_PAGE_SIZE = 200, DEFAULT_PAGE_SIZE = 25;` — **both** declarators captured, both on the right line ✓
- `export const { host, port } = parseEnv();` — not captured, and no panic or garbage symbol ✓ (`object_pattern` isn't an `identifier`)
- `export const enum Level { … }` — captured once as `enum Level`, **not** double-counted as a const ✓ (`const enum` is an `enum_declaration`, so your `lexical_declaration` pattern correctly doesn't reach it)
- arrow-valued `export const buildUrl = …` — both `const buildUrl` and `function buildUrl` ✓, as documented
- `.tsx` behaves identically ✓ (`TS_Q` is shared by both registry rows, and the query compiles against `LANGUAGE_TSX` as well)

Your two "Assumed — NOT tested" items I did test, and you called them right: `export namespace Legacy { export const NAMESPACED = 1 }` and `declare module "express" { export const AMBIENT }` are both excluded, because neither `export_statement` is a direct child of `program`. Flagging them as untested assumptions when they turn out to be correct is better practice than quietly asserting them.

## Gate

Rebuilt after **merging current `main` into your branch** — `main` had moved 68 commits and touched the same file (`PHP_Q` gained `enum_declaration`/`const_declaration`). The merge is clean and the two changes are independent query constants; I ran the gate on the merged tree, not just your head:

```
cargo fmt --all -- --check                             clean
cargo clippy -p xerj-autoindex --all-targets -D warnings   clean
cargo test -p xerj-autoindex --lib extract::code       26 passed; 0 failed
cargo test -p xerj-autoindex (full crate, merged)      472 passed; 0 failed  (+1 integration, 0 doc)
```

Your body says `ok. 26 passed; 0 failed` for the extractor module — that is the literal number I got. Everything else in the body matched measured behaviour too. CI is 12/12 green.

One note so it isn't mistaken for your fault later: an intermediate run of mine showed 3 failures in `state::sync_journal_tests` with `Resource temporarily unavailable (os error 11)` and `already in use by another process`. That's a `/tmp` flock collision with another job on this box, in tests that never touch the extractor; with a private `TMPDIR` they pass 12/12. Not yours.

I also checked the one thing that would have made this risky: nothing downstream of the extractor branches on symbol *kind* strings, and `const` was already an emitted kind for Rust/Go/C/PHP. So TypeScript joining that set can't surprise a consumer.

## Two things for later, neither blocking

1. **The same hole is still open in JavaScript.** I ran a `plain.js` containing exactly your motivating idiom — `export const users = pgTable("users", {})` — and it still extracts zero symbols, because `JS_Q` didn't get the equivalent pattern. Fair scoping for a PR titled "in TypeScript", and I'd rather merge this than widen it. But an ESM `.js`/`.jsx` schema module has the identical failure. Filed as #293, with the repro and the suggested pattern — I'd be glad to have you take it if you want a second one.
2. **Cosmetic:** the body mentions counting `.ts`/`.tsx`/`.mts` files, but `.mts`/`.cts` aren't in the registry at all today (`ts` and `tsx` only), so they aren't AST-parsed. Doesn't touch this change's correctness — just means your corpus count was over `.ts`/`.tsx`. Pre-existing, and arguably its own small fix.

Thanks again. The measurement discipline here — including publishing the run that argued *against* the change you'd already written — is exactly what we want in this repo.

pradaev:
Thanks for the thorough verification — and you're right about the `.mts` counting, that
was loose of me.

The scope comparison came from a throwaway test that walked the corpus and called
`parse_symbols` directly per file, choosing the grammar by extension itself rather than
going through `registry()`. So it *did* parse the `.mts` files, but only because it
bypassed the very lookup that would have rejected them — on that branch those files were
never AST-parsed by the real pipeline. The 918/2,382 vs 238/518 figures are still a valid
comparison of the two query scopes against the same file set, but describing that set as
"`.ts`/`.tsx`/`.mts` files" overstated what the engine would actually have seen. It was
`.ts`/`.tsx` in practice, exactly as you say.

Moot going forward now that #284 has landed — the same count would be accurate today —
but the body was wrong at the time it was written, and that's worth having on the record
rather than quietly correct-by-accident.

Picking up #293. The `function_expression` widening you flagged for the JS overlap is
noted; I'll assert it rather than assume it.

Issue #209, opened 2026-08-07
Dynamic mapping infers every string as text, never keyword


xerj-org:
Every dynamically inferred string field becomes `text`. There is no path to `keyword` without declaring the mapping by hand.

The production ingest path uses a local `infer_field_type` at `xerj-engine/src/index.rs:26749-26776`, reached from `evolve_schema_from_doc` (`:17422`) and `evolve_schema_from_docs` (`:17338`). At `:26751-26756` every non-date string maps to `Text`. No length or whitespace fork, no `Keyword` branch, no IP detection.

There is a smarter inference in `xerj-common/src/schema.rs:227-242` that does fork short whitespace-free strings to `Keyword`. It never runs on ingest. Two inference functions exist and the better one is dead on the path that matters.

Verified by hand:

```
POST /dyntest/_doc/1  {"category":"books","price":9.99}
GET  /dyntest/_mapping
  category -> text
  price    -> double
```

Numbers infer fine. Strings always land on `text`.

What that costs a user:

- `term` queries on a dynamic string field go through an analyzer, so exact match on `"New York"` does not behave the way anyone expects
- there is no `.keyword` subfield either, so the ES habit of `category.keyword` does not work
- the only fix is to know in advance and declare the mapping, which defeats the point of dynamic mapping

Aggregations and sort do still work on these fields today, which I checked directly, so this is not urgent breakage. It is a correctness-of-inference problem that shows up the first time someone tries an exact-match filter.

Two options: run the existing `schema.rs` inference on the ingest path, or emit a multi-field (`text` plus a `keyword` subfield) the way ES does by default. The second matches what people migrating will expect.

Credit where due: an earlier pass of this review claimed a JSON null pins the field as `Keyword`. That was wrong. It pins as `Text`, because the audit had cited the unused `schema.rs` function rather than the one ingest actually calls. Same user-visible breakage, wrong mechanism.

Found while reviewing XERJ's source against the user-feedback corpus, where wrong-type auto-detection is a recurring complaint.

PR #292, opened 2026-08-11
fix(engine,api): dynamic mapping opacity for nested objects (mapping never recurses, then _mapping/_field_caps never merge it in)

Vinz2168:
## Summary

A dynamically-discovered nested object field (e.g. `{"metadata": {"kind": "preference"}}`) is genuinely queryable (`{"term": {"metadata.kind": "..."}}` works — the query path resolves dotted paths against `_source` regardless of the mapping) but permanently invisible to `GET _mapping`/`_field_caps` — indistinguishable from not existing to any client that discovers fields that way instead of already knowing the field name in advance (Kibana/OSD's own field list included).

Three related, separately-fixable bugs turned out to be involved, all in `xerj-engine`/`xerj-api`'s dynamic-mapping path. This PR fixes all three.

## How this was found

Building an external MCP tool (an OpenSearch-Dashboards dashboard-authoring agent skill) that discovers a data index's fields purely through `_mapping`/`_field_caps` — the same way Kibana/OSD itself does — to build a visualization over a `/_memory` namespace's own backing index. `metadata.kind`/`metadata.topic` never appeared as discoverable fields, despite being real, queryable data.

## Root causes (traced in source, reproduced live at each step — not guessed)

**1. `infer_field_type`/`dynamic_field_config` (`xerj-engine/src/index.rs`) never recursed into `Value::Object`.** A nested object was always recorded as an opaque `{"type": "object"}` terminal during dynamic-mapping discovery, unlike real ES/OS default dynamic mapping, which recursively maps nested object properties (same `Text`→`.keyword` multi-field treatment as #209, one level deeper).

**2. `get_mapping`/`field_caps` (`xerj-api/src/es_compat.rs`) served the cached explicit-mapping blob (`state.engine.index_mappings`) verbatim, forever, whenever an index was created with *any* explicit mapping — even a partial one.** Not specific to nesting at all: isolated with a flat repro — `PUT idx {"mappings":{"properties":{"a":{"type":"text"}}}}`, then index a doc with an extra undeclared flat field `"b"` — `"b"` is permanently absent from `GET _mapping`, no matter how many more documents arrive. `field_caps`'s multi-field expansion (the `#209` mechanism, `collect_multi_fields`) inherited the same staleness since it walks the same cached `properties`.

**3. `field_caps`'s per-field loop only ever registered top-level field names.** Even with (1) and (2) fixed, a nested object's own children were still invisible: `collect_multi_fields` finds `metadata.kind.keyword` (a real multi-field, its designed job) but never `metadata.kind` itself (a `properties` child, a different concept the loop never walked into).

## Fix

- `dynamic_field_config` recurses into `Value::Object` via a depth-capped `dynamic_field_config_at_depth` helper (`DYNAMIC_MAPPING_DEPTH_LIMIT = 20`, mirrors ES's own default `index.mapping.depth.limit`), populating `FieldConfig.fields` with nested children. **No renderer changes needed** — `schema_to_es_properties` already branched correctly on `field.field_type` to emit `properties` vs `fields`; only the *discovery* path had never populated it for `Object`.
- New `merge_mapping_properties` recursively merges the stored explicit `properties` with the live schema-derived ones — the stored declaration wins for any field it names (preserves explicit settings type inference wouldn't reconstruct identically: `ignore_above`, analyzer, `dense_vector` options, ...), but a field only the live schema knows about is still included, at any depth. Wired into both `get_mapping` and `field_caps`.
- New `register_field_cap_entry` replaces `field_caps`'s top-level-only loop: registers each field dotted-path-qualified, then recurses into `Object`/`Nested`'s own `properties` children (leaving `.keyword`-style multi-field siblings to `collect_multi_fields`, avoiding double registration).
- `xerj-api/src/memory_api.rs`'s `ensure_backing_index`: `/_memory`'s own `metadata` field is no longer pre-declared as a bare `{"type": "object"}` — any explicit declaration, even a shallow one, permanently blocks a field from going through dynamic discovery again, so leaving it undeclared lets it flow through the (now-fixed) dynamic path on first write.

**Regression found and fixed before this was ready**: the live-schema merge above started leaking `semantic_text`'s own companion `dense_vector` field (e.g. `body` → `body_vector`, an engine-only implementation detail never declared by the caller under that name) into `_mapping` — caught by `api_semantic_bulk_builds_complete_hnsw_and_survives_restart`, which explicitly asserts the companion stays hidden. New `semantic_companion_field_names` scans the stored mapping for `semantic_text` declarations (the only place that type can legitimately appear — dynamic mapping never infers it) and excludes each one's companion name (via the existing `semantic_mapping_contract`) from what the live schema contributes to the merge.

## Verified

End-to-end against a live build, on the motivating real case (not just synthetic repros):

```
POST /_memory/repro {"text": "example", "metadata": {"kind": "preference", "topic": "language"}}

GET .xerj-memory-repro/_mapping
  -> metadata.properties.{kind,topic}, each with its own #209 .keyword multi-field

GET .xerj-memory-repro/_field_caps?fields=metadata.kind
  -> {"fields": {"metadata.kind": {"text": {...}}}}   (was: {"fields": {}})

GET .xerj-memory-repro/_field_caps?fields=*
  -> the complete correct set: metadata, metadata.kind, metadata.kind.keyword,
     metadata.topic, metadata.topic.keyword, text, text_vector, stored_at
```

- `cargo test -p xerj-api --lib`: 174/174 (includes 3 new/updated tests: `dynamic_object_recurses_into_nested_properties`, an updated `dynamic_string_gets_keyword_multi_field` — its non-string-leaf table was asserting the bug as expected behavior, `(json!({"a": 1}), 0)` — and the pre-existing `api_semantic_bulk_builds_complete_hnsw_and_survives_restart`, which is what caught the companion-field regression above).
- `cargo test -p xerj-engine --lib`: 463/465 (2 pre-existing, macOS-only `snapshot_path_security_tests` failures, unrelated — same ones called out as pre-existing in #290).
- `cargo clippy -p xerj-api -p xerj-engine --lib -- -D warnings`: clean.
- `cargo fmt -p xerj-api -p xerj-engine`: clean.
- Rebased onto `upstream/main` (rc.15) immediately before opening this PR — 0 commits behind, no conflicts on the dynamic-mapping-recursion commit; the mapping-merge commit auto-merged clean against the intervening `#202`/`#206` failed-index-recovery work in the same functions, verified by re-reading the merged `get_mapping`/`field_caps` bodies line by line rather than trusting a clean auto-merge.

## Scope

Two commits, independently reviewable:
1. `fix(engine): recurse dynamic mapping into nested object properties` — stands alone, correct and tested on its own merits, but **not sufficient by itself** to fix the real-world `/_memory` scenario (masked by bug 2 below) — flagging this so it doesn't read as "why doesn't this change anything I can observe" in isolation.
2. `fix(api): GET _mapping/_field_caps merge live schema instead of serving a stale index-creation blob` — the actual unblock for the motivating case, plus the field_caps nested-children fix and the companion-field regression fix.

**Deliberately left alone**: `field_caps` still exposes `text_vector`/`body_vector` via the separate top-level `schema.fields` loop (which reads the live schema directly and already showed it before this PR, confirmed against a pre-PR build — not a regression introduced here). No existing test requires `field_caps` to hide it (only the `_mapping` test does, and that one passes). Changing that would be scope creep beyond what this PR's repro and tests establish.

## Test plan

- [x] `cargo test -p xerj-api --lib` (174/174)
- [x] `cargo test -p xerj-engine --lib -- dynamic_` (2/2, new/updated tests)
- [x] `cargo test -p xerj-engine --lib` full suite (463/465, 2 pre-existing unrelated)
- [x] Reproduced each of the 3 bugs live before writing any fix, confirmed each is gone after
- [x] Reproduced the companion-field regression live, confirmed the fix closes it without reintroducing bug 2
- [x] `cargo clippy -p xerj-api -p xerj-engine --lib -- -D warnings`
- [x] `cargo fmt -p xerj-api -p xerj-engine -- --check`
- [x] Rebased clean against `upstream/main` immediately before opening

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -13,7 +13,7 @@ use axum::{
 use chrono::{Datelike, TimeZone, Timelike, Utc};
 use serde::Deserialize;
 use serde_json::{json, Value};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use uuid::Uuid;
 use xerj_common::types::{FieldConfig, FieldType, Schema};
 use xerj_query::parse_request;
@@ -2430,6 +2430,85 @@ fn merge_dotted_property(
     merge_dotted_property(sub_obj, &rest, leaf_value);
 }
 
+/// Merge a stored (explicit, cached-at-index-creation) mapping's
+/// `properties` with the live schema-derived `properties`
+/// (`schema_to_es_properties`).
+///
+/// The stored declaration wins for any field it names -- it may carry
+/// explicit settings (`ignore_above`, `analyzer`, `dense_vector` options,
+/// ...) the schema's own type inference wouldn't reconstruct identically.
+/// But a field the live schema knows about and the stored blob doesn't is
+/// still included, instead of being permanently invisible: without this,
+/// GET _mapping/_field_caps serve the ORIGINAL index-creation-time blob
+/// forever, and any field discovered dynamically since -- flat or nested,
+/// at any depth -- never appears in either, no matter how long the index
+/// has been receiving new documents.
+///
+/// Recurses into whichever nested-children key each side actually uses
+/// (`properties` for Object/Nested, `fields` for a leaf's multi-fields),
+/// so a field explicitly declared without its own nested declaration
+/// (e.g. a bare `{"type": "object"}`) still picks up children discovered
+/// dynamically since, rather than only ever getting top-level treatment.
+fn merge_mapping_properties(
+    stored_properties: &serde_json::Map<String, Value>,
+    live_properties: &serde_json::Map<String, Value>,
+) -> serde_json::Map<String, Value> {
+    let mut merged = stored_properties.clone();
+    for (key, live_val) in live_properties {
+        match merged.get_mut(key) {
+            None => {
+                merged.insert(key.clone(), live_val.clone());
+            }
+            Some(stored_val) => {
+                for children_key in ["properties", "fields"] {
+                    let Some(live_children) =
+                        live_val.get(children_key).and_then(|v| v.as_object())
+                    else {
+                        continue;
+                    };
+                    let stored_children = stored_val
+                        .get(children_key)
+                        .and_then(|v| v.as_object())
+                        .cloned()
+                        .unwrap_or_default();
+                    let merged_children = merge_mapping_properties(&stored_children, live_children);
+                    if let Some(obj) = stored_val.as_object_mut() {
+                        obj.insert(children_key.to_string(), Value::Object(merged_children));
+                    }
+                }
+            }
+        }
+    }
+    merged
+}
+
+/// Companion `dense_vector` fields the engine auto-creates for each
+/// `semantic_text` field (e.g. `body` -> `body_vector`, or `target_field`
+/// if the caller declared one) are an implementation detail, never meant
+/// to be user-visible -- nothing in the caller's own mapping ever names
+/// them. Once a document makes the live schema aware of a companion field,
+/// `merge_mapping_properties` above would otherwise happily surface it in
+/// `GET _mapping`/`_field_caps`, even though it was never declared under
+/// that name by anyone. Scans `stored_properties` (the only place a
+/// `semantic_text` field can legitimately be declared -- dynamic mapping
+/// never infers that type) for each one and returns the companion name
+/// `semantic_mapping_contract` computes for it, so callers can filter
+/// those out of what they feed into `merge_mapping_properties`.
+fn semantic_companion_field_names(
+    stored_properties: &serde_json::Map<String, Value>,
+) -> HashSet<String> {
+    stored_properties
+        .iter()
+        .filter_map(|(name, def)| {
+            let obj = def.as_object()?;
+            (obj.get("type").and_then(Value::as_str) == Some("semantic_text")).then(|| {
+                let (target, _dimensions, _similarity) = semantic_mapping_contract(name, obj);
+                target
+            })
+        })
+        .collect()
+}
+
 // ─────────────────────────────────────────────────────────────────────────────
 // GET /{index}/_mapping
 // ─────────────────────────────────────────────────────────────────────────────
@@ -2476,17 +2555,43 @@ pub async fn get_mapping(
             .get(name)
             .map(|v| v.clone())
             .unwrap_or(Value::Null);
-        let mut mappings = if stored.is_null() {
-            match &idx {
-                Some(i) => {
-                    let schema = i.schema().await;
-                    let properties = schema_to_es_properties(&schema);
-                    json!({ "properties": properties })
-                }
-                None => json!({ "properties": {} }),
+        let mut mappings = match (&idx, stored.is_null()) {
+            (Some(i), true) => {
+                let schema = i.schema().await;
+                let properties = schema_to_es_properties(&schema);
+                json!({ "properties": properties })
             }
-        } else {
-            stored
+            // A live index with a stored (explicit-at-creation) mapping:
+            // merge in anything the live schema has picked up dynamically
+            // since -- otherwise a field discovered after index creation
+            // (flat or nested, at any depth) is permanently invisible here,
+            // forever serving the exact blob index creation started with.
+            (Some(i), false) => {
+                let schema = i.schema().await;
+                let mut live_properties = schema_to_es_properties(&schema);
+                let stored_properties = stored
+                    .get("properties")
+                    .and_then(|p| p.as_object())
+                    .cloned()
+                    .unwrap_or_default();
+                // semantic_text's own companion dense_vector fields are an
+                // engine-internal implementation detail, never declared by
+                // the caller under that name -- exclude them from what the
+                // live schema contributes, same as they were never visible
+                // before this merge existed (see semantic_companion_field_names).
+                let companions = semantic_companion_field_names(&stored_properties);
+                live_properties.retain(|k, _| !companions.contains(k));
+                let merged = merge_mapping_properties(&stored_properties, &live_properties);
+                let mut m = stored.clone();
+                if let Some(obj) = m.as_object_mut() {
+                    obj.insert("properties".to_string(), Value::Object(merged));
+                }
+                m
+            }
+            // No live index (failed index, #206) -- nothing to merge with,
+            // serve whatever's recoverable as-is.
+            (None, false) => stored,
+            (None, true) => json!({ "properties": {} }),
         };
         // ES injects default `rescore_vector.oversample: 3.0` on any
         // dense_vector whose `index_options.type` is BBQ-quantised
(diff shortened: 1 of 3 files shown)


xerj-org (changes requested):
First: thank you for the writeup. You traced three distinct bugs to `file:line`, reproduced each live *before* fixing it, found and closed your own regression (the `semantic_text` companion leak), and flagged up front that commit 1 changes nothing observable on its own. That is the standard we want, and the central insight is right — `schema_to_es_properties` already branched correctly on `field_type`, and only the *discovery* path was failing to populate `fields` for `Object`.

I read the whole diff and traced each path in the tree at `5fa093c6`. **Honesty note on my own evidence:** I had a release build of your branch queued, but our shared build host is saturated by other review jobs and it never got the cargo lock, so everything below is *source-traced with line numbers*, not measured on a running server. I'm flagging that rather than dressing it up — but findings 1-4 are plain control flow (exact-name lookups and a `len()`), not timing or heuristics, so they should take you seconds to confirm. CI on the PR is green, including the ES-YAML conformance gate.

I am not merging it yet. The change is a real improvement, but it is narrower than the PR body claims, and the biggest gap is the *same* "queryable but invisible" bug you set out to close, one level down. Details below, each traced to source.

---

## 1. Nested keys are discovered only from the **first** document that introduces the object

The case to try (and the one I'd ask you to add as a test): store one memory `{"metadata": {"kind": "preference"}}`, then a second `{"metadata": {"project": "xerj"}}`. Reading the code, `metadata.kind` becomes visible and `metadata.project` never does — no matter how many more memories arrive — while staying perfectly queryable via `_source` dotted-path resolution. That is verbatim the symptom in your own opening paragraph.

**Why.** Both production dynamic-mapping entry points gate on the **top-level** JSON key only:

- `engine/crates/xerj-engine/src/index.rs:17995` — `!schema.schema.has_field(key)` in `evolve_schema_from_doc`
- `engine/crates/xerj-engine/src/index.rs:17913` — same test in `evolve_schema_from_docs`

and `Schema::has_field` is a flat scan by exact name (`engine/crates/xerj-common/src/types.rs:454-461`, `self.fields.iter().find(|f| f.name == name)`). Once `metadata` is in the schema, `dynamic_field_config` is never called for it again, so no later document's `metadata` sub-object is ever inspected.

There is also no second path that could repair this later: across `index.rs` and `types.rs`/`schema.rs`, the only `fields.push` onto a `FieldConfig`'s children are the two inside `dynamic_field_config_at_depth` itself (`index.rs:28145`, `:28156`), both on a freshly constructed `FieldConfig`. Nothing anywhere merges children into an already-registered field, so a `FieldConfig`'s `fields` vec is frozen the moment it is added.

ES does not work this way. It registers dynamic mappers keyed on the **full dotted path** and tests `mappingLookup.getMapper(fullPath)`, so a later document introducing `metadata.project` is mapped normally (`DocumentParserContext.java:791-800` in our reference corpus — read for semantics only, nothing copied).

**Why this specifically undercuts the motivating case.** Dropping the `metadata` pre-declaration in `memory_api.rs` was the right call, and for exactly the reason you give: the shape is caller-defined and varies per memory. But that same variability is what makes first-doc-wins the wrong resting place here. A `/_memory` namespace's discoverable field list becomes whatever the first memory happened to carry, and every later distinct key is exactly as invisible as before — while now *looking* authoritative to the OSD agent that motivated the PR. Before, a discovery client got a consistently empty object; after, it gets a confident partial answer whose content depends on ingest order.

**What would make it mergeable:** either make the walk resolve dotted paths (check `metadata.project`, not `metadata`, and merge children into the existing `FieldConfig`), or — if that is too large for this PR — keep the current behaviour but state the limitation in `dynamic_field_config`'s doc comment and the PR body, and open a follow-up issue. What it should not do is ship under a body that reads as a complete fix.

## 2. `GET /_field_caps` (cluster-wide) still has the unfixed code

`global_field_caps` (`es_compat.rs:29310`, routed at `router.rs:720`) was not touched. It still:

- serves the stale blob verbatim — `stored_mapping.get("properties")` at `es_compat.rs:29338-29345`, the exact pattern bug 2 is about;
- iterates only top-level `schema.fields` (`:29347`), so nested children never appear;
- never calls `collect_multi_fields`, so it does not even have #245's `.keyword` entries that `field_caps` has had since that PR.

So `GET /{index}/_field_caps` and `GET /_field_caps` now disagree about what fields exist. Since the PR's framing is "clients that discover fields through `_field_caps`", the cluster-wide form belongs in scope — it is the same fix, ~20 lines, in the same file.

## 3. Indices that already exist are not fixed, and this is not disclosed

Every `.xerj-memory-*` index created before this ships still has `metadata` declared as a bare `{"type": "object"}` in its stored mapping *and* `metadata` present in its live schema — so `has_field("metadata")` short-circuits forever and no amount of new writes will ever populate children.

`merge_mapping_properties` handles the *render* side of this correctly (a stored bare `{"type":"object"}` does pick up live children) — the block is upstream, in discovery. Worth saying explicitly in the PR body that the memory fix applies to newly created namespaces only, and worth deciding whether existing ones get a migration.

## 4. Recursively added children bypass `max_fields_per_index`

The mapping-explosion guard counts `schema.schema.field_count()` (`index.rs:17934` and `:18009`), and `field_count()` is `self.fields.len()` — top-level only (`types.rs:480-482`). Before this PR that was accurate, because nested keys never became `FieldConfig`s. Now one top-level object field can carry an unbounded number of nested `FieldConfig`s, none of them counted.

Concretely: a document of 600 distinct **top-level** keys is refused by the guard at `index.rs:18009` (default `max_fields_per_index` is 500, `config.rs:1197`) and maps nothing. Move the same 600 keys one level down — `{"blob": { ...600 keys... }}` — and the guard sees `new_fields.len() == 1`, waves it through, and `blob` is stored carrying 600 child `FieldConfig`s that `field_count()` will forever report as 1.

The guard's own comment describes precisely this ("An authenticated client could ingest documents with arbitrarily many distinct field names, bloating the schema unbounded"). Finding 1 happens to bound the damage to a single document's key count, which is why I am filing this as hardening rather than a security issue — but the two interact, so if you fix 1, please count nested children here at the same time.

## 5. The two new `es_compat.rs` functions have no tests

`merge_mapping_properties` and `register_field_cap_entry` are the highest-risk code in the diff (216 added lines, recursive, and where the one regression you hit actually lived) and the diff adds zero `#[test]` to `es_compat.rs`. The three tests the body lists are two engine-side `dynamic_field_config` tests plus a pre-existing semantic test that caught the companion leak by luck. Please add direct coverage — stored-wins-on-conflict, live-only field included, bare `{"type":"object"}` picking up children, and the companion exclusion.

---

### Smaller things

- **Depth limit diverges from the ES behaviour it cites.** `DYNAMIC_MAPPING_DEPTH_LIMIT` silently truncates; ES throws a `MapperParsingException` (`ObjectMapper.java:601`, `checkObjectDepthLimit`). The constant's comment says it "mirrors ES's own default `index.mapping.depth.limit`" — true of the number, not the behaviour. Silent truncation is the #204 class; please either reject or reword the comment.
- **`memory_api.rs:168`** still says "text and metadata are always mapped" — stale as of this PR.
- **No CHANGELOG entry.** #260 and #252, the two closest mapping changes, both added one under `[Unreleased]`.
- **`#[allow(clippy::too_many_arguments)]` on `register_field_cap_entry` is dead** — the function takes 6 arguments and clippy's threshold is 7.
- **The depth assertion is weak.** `assert!(levels < DYNAMIC_MAPPING_DEPTH_LIMIT + 5)` passes at 24 as readily as at the real 19; `assert_eq!(levels, DYNAMIC_MAPPING_DEPTH_LIMIT - 1)` would actually pin the cap.
- `dynamic_field_config_at_depth`'s array unwrap uses `as_object()` on the first non-null element, so an array-of-arrays-of-objects (`[[{"a":1}]]`) infers `Object` via `infer_field_type`'s recursion but discovers no children. Cosmetic, and arguably fine — just noting it since the doc comment claims consistency with `infer_field_type`.

---

Items 1-3 are what I would want resolved before merge; 1 can be resolved by scoping honestly rather than by more code, if you would rather keep this PR small. 4 and 5 I would like in this PR too since they are cheap. Happy to take this in stages if you would prefer to land the merge fix (bug 2, which is clean and clearly correct) separately from the recursion.

Thanks again — this is careful work and the diagnosis is genuinely good.

xerj-org:
Correcting one detail in my own review above, since this repo holds evidence claims to a standard and mine should meet it too.

I wrote that my release build of your branch "never got the cargo lock". That is not quite what happened: it did acquire the lock and compiled through `xerj-engine` and `xerj-api`, and I then killed it to stop it starving the other review jobs sharing that build host. I have restarted it and it is currently queued behind a sibling job again.

The substance of the note is unchanged and still accurate — **no live run happened, no measurement was taken, and none of findings 1-5 rests on one.** They are all static reads of control flow with the line numbers given, which is exactly why I was willing to post them without a server. But "never got the lock" was wrong, and I would rather say so than let it stand.

If the build host frees up I will add the measured confirmation here. It should not change the analysis; if it does, I will say that too.

Issue #311, opened 2026-08-11
Default `_source` arm deep-copies every hit source even when the index has no embedding fields (follow-up to #309)


xerj-org:
Found in the pre-merge review of #309. Perf-only, correctness unaffected.

The new `SourceFilter::Default` arm (`engine/crates/xerj-engine/src/index.rs` ~24471) calls `filter_object(&h.source, &[], &companions)` per hit; the non-dotted branch (~24540) rebuilds the top-level map with a deep `v.clone()` of every field value. Pre-#309 the default (`Enabled(true)`) was a zero-cost passthrough, so every default `_search` on any non-semantic index now pays an O(source-bytes) allocation+copy per hit for nothing. Constant-factor (response serialization is already O(source)), but it sits on the hottest read path.

Easy wins:
- early-return the hits unchanged when `generated_companion_fields` is empty;
- `map.remove()` the companions in place instead of rebuilding the map;
- the 4 call sites each take `schema.read().await` and rebuild the companion `HashSet` per request, and `apply_source_filter` clones the set into a `Vec<String>` per call — cache or pass by reference.

PR #321, opened 2026-08-12
Feat(search) Make code search answer the question, not hand back the file

buger:
An agent asking XERJ *"which function converts a time difference to human-readable text, and how is it implemented?"* against a mixed WordPress + Markdown corpus used to spend **6,018,540 bytes over 10 requests** to find out. It now spends **34,574 bytes over 4** — and it learns how from `--help` before making a single expensive mistake.

That number is not a micro-benchmark. It is a fresh agent, given nothing but a binary and a folder, measured end to end at the wire.

---

## Why

XERJ describes itself as the search engine for AI agents. But the default answer to a code search was the **whole file**, and the mechanisms to do better were effectively invisible.

Concretely, before this PR:

| what the user does | what happens |
|---|---|
| searches for a function | gets the entire 356 KB file it lives in |
| asks for `_passage` to get a snippet | gets a 512-character window that opens mid-token — `"\t}\n\n\tif ( isset( "`, a dangling brace |
| tries `highlight` instead | gets fragments from the **top of the file**, never the match |
| guesses a field name (`content`) | `0 hits`, no explanation |
| guesses a format value (`markdown`) | `0 hits`, no explanation |
| wants a second instance on another port | there is no `--port` flag |

None of this was discovered by reading the code. It came from **blind usability runs**: fresh agents, told nothing, given a binary and a corpus, with every HTTP response size recorded. One of them resorted to running `strings` on the binary to find the config keys for a port. Two independently gave up on highlighting and downloaded whole documents instead. Their transcripts drove every change here.

---

### 1. `_passage` returns the enclosing code block

For code documents, a passage is now the smallest syntactic node that is a real semantic unit — function, method, class — containing the match.

```
before:  "\t}\n\n\tif ( isset( $source_args['field'] ) ..."      ← dangling brace
after:   "function render_block_core_post_date( $attributes, ... ) {
              ...
          }"                                                      ← complete, balanced
```

**Resolved at response time, not index time.** This is the load-bearing design decision: it works on **every existing index** with no reindex and no migration, adds nothing to stored data, and leaves `_source` untouched. Parsed trees are cached in-process by content hash — **182 ms cold, 18 ms warm** on a 357 KB PHP file, against a 121 ms no-passage baseline. Non-code documents are untouched, and code falls back to the existing window whenever the language is unsupported, the parse fails, no enclosing node qualifies, or the block exceeds a size cap.

The per-language node-kind tables are adapted from [probelabs/probe](https://github.com/probelabs/probe)'s `is_acceptable_parent` policy (dual Apache-2.0/MIT) — the node-kind **data only**, re-expressed for tree-sitter 0.25, with probe's generic containers (`block`, `module`, `object`) deliberately dropped so "no enclosing node" falls back rather than returning a bare block. No probe code is depended on or copied.

### 2. Highlighting finds the match

Two compounding defects, both invisible to the five existing highlight tests because those use short documents where the earliest match is coincidentally correct:

- **A shared scan budget starved rare terms.** `match_positions.len() > num_frags * 4` is a global count checked inside a per-term loop, and terms arrive sorted — so for `function human_time_diff`, hundreds of early "function" hits consumed the budget. Now budgeted per term.
- **Candidates were ordered earliest-first.** For any multi-term query that resolves to "wherever the commonest word first appears" — the top of the file. Now ranked by how many *distinct* query terms fall in the fragment window, position only as a tiebreak.

### 3. Silent failures now explain themselves

Three ways a newcomer silently got nothing, all now answered with the real values and a **corrected, already-efficient** request:

- **unknown field** — names it, lists the searchable fields
- **unknown value** — names sampled values present, points at aggregation for the complete set
- **oversized payload** — on `_search` *and* `GET /_doc/{id}`

Hints are additive (`_xerj.hints`), emitted **before** `hits` in the JSON — agents truncate long output from the bottom, so a hint after a 400 KB array is never read. A test asserts the byte offset of the hint key is lower than `"hits"`.

They fire **only when actionable**: a well-formed, efficient query gets nothing.

### 4. The optimisation is taught, not hidden

`--help` gained a `SEARCHING A CODEBASE` block. The post-index line — the moment an agent turns from indexing to searching — now prints the code-aware query when code was indexed, and, when a corpus mixes code and prose, how to scope plus **the `ax_format` values actually present in that run** (computed from the data, so it cannot drift stale).

The recommended query projects `["ax_path","title"]` rather than `_source:false`. That costs **69 bytes** and removes the reason for a **5,975,751-byte** follow-up: a blind agent got the right function, could not tell which file it came from, and paid 6 MB recovering the path.

### 5. `--port`

`--bind` takes an IP only, and the three port keys live in a TOML file `--help` never mentioned. `--port N` sets the ES-compat port and derives `N+1`/`N+2`.

---

## Testing strategy

Three layers, deliberately, because each catches what the others cannot.

**1. Unit and integration tests** — necessary, not sufficient. The highlight defect survived five existing tests.

**2. Falsification.** Every regression test here was *proven to fail* with its fix reverted, not merely observed passing. This caught a bad test of my own: the first highlight regression test passed even with the fix reverted, because narrow filler let clustered matches collapse into one window. Widening the filler past `fragment_size` made it discriminating. A test never seen to fail is not evidence.

**3. Blind usability runs.** Fresh agents, no knowledge of the system, given a binary and a folder, every response size recorded at the wire. This layer found what the others structurally could not:

- the hint that fired only on *unprojected* `_source`, so an agent that projected early — a good instinct — silenced it permanently
- the recommended query that returned a passage with no path
- the highlighter returning the head of the file
- the `strings`-the-binary port workaround

An early version measured *agent tokens* and showed the branch looking **worse**. That was an artifact: the harness truncated a 3.1 MB response before it reached the agent's context, so the baseline was shielded from its own cost. Measuring `%{size_download}` at the wire corrected it — the same request was **3,173,713 bytes** against **8,023**.

---

## Results

**Cold blind run** — fresh agent, mixed corpus (257 PHP + 68 Markdown), two questions (one code, one prose), no knowledge of the system:

| | before | after |
|---|--:|--:|
| bytes to answer both | **6,018,540** | **34,574** |
| requests | 10 | 4 |
| whole files fetched | several | **0** |
| both answers correct | yes | yes |
| optimisation found from docs | no | **yes** |

**174× fewer bytes.** The agent's own note:

> *"The `xerj --help` / `autoindex --help` text is dense but genuinely useful — it told me up front how to search for a function vs a whole file … which is exactly what made query 7 return a clean, complete function body instead of a whole PHP file."*

**Component measurements**

| | before | after |
|---|--:|--:|
| recommended query (passage + path) | 3,185 B, no path | **3,254 B, complete answer** |
| passage resolution, warm | — | **18 ms** (182 ms cold) |
| definition vs call-site ranking | — | definition wins, 8.6 vs 8.5 |
| highlight on a 356 KB file | head of file | **the actual match** |

---

## Compatibility

Additive. `_xerj` is a new optional top-level field on successful responses; `hits`, `total`, `_score` and the ES `reason` strings are untouched. Non-code documents keep their existing passage behaviour, asserted by test. No stored data changes, no reindex, no migration.

---

**On this branch's current base (`803c85b8`), after rebase:**

- `cargo test -p xerj-engine -p xerj-api -p xerj-autoindex -p xerj-query`: **1,915 passed, 0 failed** across 67 suites

On the previous base (`18f85755`), before the rebase:

- `cargo test -p xerj-engine -p xerj-api -p xerj-query`: **1,347 passed, 0 failed** across 62 suites
- `cargo test -p xerj-autoindex -p xerj-api`: **849 passed, 0 failed**
- ES-YAML conformance: **0 failed** (1366 passed, 3 skipped) — re-run after the highlight change specifically, since that suite exercises highlighting
- `cargo fmt --all --check`, `cargo clippy --workspace --all-targets -- -D warnings`, and the ONNX-feature clippy job: clean
- Cold blind runs, byte totals measured with `curl -w '%{size_download}'`
- Falsification of each regression test, both directions

## Assumed — NOT tested

- That the node-kind tables are complete for languages with no blind-run coverage (Swift, Solidity, Haskell, Crystal are **not** in XERJ's registry; the 13 that are were checked against each grammar's `node-types.json`). **Falsified by:** a passage in an untested language opening mid-token.
- That the 16 KB block cap suits all deployments. It is env-overridable. **Falsified by:** a legitimate function exceeding it and silently falling back.

## Not run

- **ES-YAML conformance on this exact base.** The suite passed at 0 failed on `18f85755`, including a run made specifically after the highlight change. It is re-running on `803c85b8` now but had not reported when this was written. Main gained #311 ("make the default `_source` arm a passthrough") in between, which touches an adjacent path — so **CI on this PR is the authority for conformance**, not the pre-rebase figure. The full test suites *have* been re-run on this base (1,915 passed, 0 failed).
- **Behavioural measurements were taken on `18f85755`** — the passage output, latency, byte totals and cold-run figures. The rebase changed no code in this branch, but they have not been literally re-measured on the new base.
- Highlight behaviour on non-Latin scripts beyond the existing multibyte tests.

---

## Reviewer notes

- **`_passage` still has no path of its own.** The recommended query projects `ax_path`, which closes the gap in practice, but a passage that cannot say where it came from is arguably half an answer. Deliberately left as a separate change rather than smuggling a response-shape change into this one.
- **Vocabulary mismatch is not addressed and cannot be by this PR.** A blind agent searching the literal phrase *"exit code 3"* got zero hits because the docs say *"completed-with-junk"*. The query was well-formed against a real field, so no hint applies. Fixing that class needs semantic retrieval, and the default embedder is lexical feature-hashing.
- The block-size cap and tree-cache capacity are tunable via `XERJ_CODE_PASSAGE_MAX_BYTES` and `XERJ_CODE_PASSAGE_CACHE_CAP`.


xerj-org:
CI is red on **Format + Clippy** and **ONNX feature compile + tests**. Both come from the same two mechanical causes, and one of them is pointing at a real mistake rather than a style preference. I have the fix locally but this PR is from `probelabs/xerj` with `maintainerCanModify: false`, so I cannot push it to your branch — here it is to apply.

**1. `cargo fmt --all` has not been run.** rustfmt wants reflows in three files: `es_compat.rs` (a `matches!` that needs wrapping), `code_blocks.rs` (`PYTHON_KINDS` over the width limit), and `index.rs` (the `refine_code_passage(...)` call). One command fixes all of it:

```sh
cd engine && cargo fmt --all
```

**2. The clippy failure is a merged doc comment, not a formatting nit.** `doc_lazy_continuation` fires three times, denied by `-D warnings`. The cause: the doc block for `enclosing_block` and the doc block for `WRAPPER_KINDS` ended up as one comment sitting above the const, so `Syntactic wrappers…` reads as a lazy continuation of the `Returns None … when:` bullet list above it.

The lint is the small half. The real consequence is that **`enclosing_block` — the public entry point of the module — currently has no rustdoc at all**, while `WRAPPER_KINDS` carries the description of a function it is not. Indenting the paragraph to silence clippy would lock that wrong attachment in place, so the fix is to split them back:

```rust
// ── Public API ────────────────────────────────────────────────────────────

/// Syntactic wrappers that carry a declaration's modifiers rather than being
/// blocks in their own right. A resolved block is widened through these so the
/// passage shows the declaration as it is actually written — `export function
/// f()`, not a bare `function f()` that reads as un-exported.
///
/// Language-agnostic by construction: a kind absent from a grammar simply never
/// matches, so one table is safe for every language in the registry.
const WRAPPER_KINDS: &[&str] = &[
    // JavaScript / TypeScript
    "export_statement",
    "export_default_declaration",
    // Python
    "decorated_definition",
];

/// Find the smallest node enclosing byte range `[start, end)` in `body`
/// whose kind counts as an acceptable semantic block for `language`, and
/// return its own `[start_byte, end_byte)` span.
///
/// Returns `None` — meaning "keep the caller's existing window" — when:
/// - `language` has no registered grammar/kind table here;
/// ... (list unchanged) ...
/// - the smallest acceptable block exceeds [`max_block_bytes`] — since any
///   larger enclosing block is only bigger, this stops at the first
///   candidate rather than searching further up.
pub fn enclosing_block(
    language: &str,
    body: &str,
    start: usize,
    end: usize,
) -> Option<(usize, usize)> {
```

Verified locally on top of your `866e914d`: `cargo fmt --all` clean, and `cargo clippy -p xerj-engine --all-features` reports 0 doc-lint errors. Doc comments and whitespace only — no behaviour change.

Once that is pushed the two red jobs should go green and this is mergeable. Alternatively, tick **"Allow edits by maintainers"** on the PR and I will push the commit for you.

buger:
@xerj-org forgot to push the commit with CI fix, should be fixed now

Issue #318, opened 2026-08-12
[compression] config section is accepted and ignored: level / enabled / block_size_docs reach no encoder (RFC #148 lever 3)


xerj-org:
Follow-up from [**RFC #148 — per-field / per-type storage compression**](https://github.com/xerj-org/xerj/discussions/148) — discussion thread, where the measured answer to lever 3 lives.

## The user problem

The RFC's lever 3 asks for exactly one thing:

> Making the level a config knob (like ES's `index.codec: best_compression`) is the clean version.

and the reply in the thread states where that stands:

> The config schema carries a `compression.level` knob that nothing reads yet — an accepted-and-ignored setting of precisely the class #204 is sweeping.

So an operator who wants to trade CPU for `$/GB` on a cold index sets `[compression] level = "best"`, restarts, re-ingests, and gets byte-for-byte the same index — with no error and no warning. The docs tell them it works: xerj.org/docs/compression says "Three levels, controlled by `[compression] level`" (`landing/docs/compression.html:125`) and the config table describes `enabled` / `level` / `block_size_docs` as live settings (`landing/docs/config.html:180-183`) — unlike the `[merge]` and `[vector]` rows right above them, which are annotated "Accepted but not wired".

## What XERJ does today

`CompressionConfig` is defined and defaulted (`engine/crates/xerj-common/src/config.rs:909-928`) and never referenced again. `grep -rn CompressionConfig engine/crates` returns only its own definition site. There is no validation (`block_size_docs` is documented as 16–4096 and is not range-checked) and no dormant-knob warning — the mechanism that exists for `[merge]` (`config.rs:864` `dormant_overrides`, emitted at `crates/xerj-server/src/main.rs:440`) has no `[compression]` equivalent.

The real encoder sites take a hardcoded constant instead:

- `crates/xerj-storage/src/stored_codec.rs:168` — `const STORED_ZSTD_LEVEL: i32 = 3` (stored `_source`, ZBS2)
- `crates/xerj-fts/src/index.rs:323` — `const ZSTD_DURABLE_LEVEL: i32 = 3` (`.meta` ZFM4, `.post` ZPS1), with a comment recording the 19→3 revert and why flush must stay cheap
- `crates/xerj-storage/src/doc_values.rs:651` — level 3 literal, level 1 on fallback

[...]

PR #322, opened 2026-08-12
feat(replication): push a filtered subset of indices to an external ES-compatible target (single-node WAL tap)

### LINKED ISSUE #320: [feature]: push a filtered subset of indices to an external ES-compatible target (single-node WAL tap)
## The user problem

From @Vinz2168 in discussion [#171](https://github.com/xerj-org/xerj/discussions/171):

> My use case: XERJ as a lightweight edge node (NAS/homelab scale) that syncs a curated subset of
> data up to a bigger central cluster — closer to "Beats but push-based and XERJ-native" than to
> full CCR, which I know is explicitly out of scope for v1.0.

Concretely: a single-node WAL tap that pushes a **filtered subset of indices** to an external
ES-compatible cluster (OpenSearch, Elasticsearch, or another XERJ node) in near-real-time rather
than by polling — config-driven index allowlist that never touches system indices,
runtime-adjustable over REST without a restart, target-agnostic because it is just `_bulk`
underneath.

This is deliberately *not* CCR. It is one-directional, single-node, and does not depend on the
post-GA multi-node sharding/replication track (ROADMAP.md L56), which is what makes it worth
discussing separately from the CCR refusal.

## What XERJ does today

Nothing pushes data out of the engine. Verified in the source:

- **CCR is an honest 501, not a stub-success.** `ccr_follow` / `pause_follow` / `resume_follow` /
  `unfollow` and all three `_ccr/auto_follow/{name}` handlers return `501 Not Implemented`
  (`engine/crates/xerj-api/src/es_compat.rs:34580` onward; `stub::not_implemented_yet` →
  `StatusCode::NOT_IMPLEMENTED` at `engine/crates/xerj-api/src/stub.rs:67`). Routes are wired at
  `engine/crates/xerj-api/src/router.rs:818-835`. `engine.ccr_auto_follow` exists as a map with no
  consumer, which is why the handlers were made to fail loud.
- **Reindex-from-remote is rejected up front**, so there is no pull-side workaround either
  (`engine/crates/xerj-api/src/es_compat.rs:18831`, error text at `:18849`).
- **Snapshot/restore is the only export path** (`engine/crates/xerj-api/src/router.rs:566-577`) —
  scheduled, whole-index, not near-real-time.

What already exists that a tap could build on:

- **The WAL carries enough to synthesise `_bulk`.** `WalEntry` is `Index { doc_id, source }`,
  `Delete { doc_id }`, `UpdateMapping { schema }` (`engine/crates/xerj-storage/src/wal.rs:145-155`),
  and `ReplayEntry` attaches a `seq_no` plus file offset (`:160`).
- **Ordering and cursors already exist per index.** Each `IndexStore` owns its own `wal/` dir and a
  single shared `seq_counter` across its WAL shards
  (`engine/crates/xerj-storage/src/index_store.rs:481`, `:638-654`, counter at `:643`), and
  `WalCheckpoint` is `{generation, offset, max_seq_no}` (`wal.rs:182`). One store per index
  (`engine/crates/xerj-engine/src/index.rs:5692`), so an index allowlist maps cleanly onto WAL dirs.
- `reqwest` is already a workspace dependency of `xerj-server`, so no new HTTP client is needed.

## What would have to change

Smallest honest statement of the work:

1. A read/observer position on the WAL append path that does not add a lock holder to the write
   path or slow `append`/`append_index_raw_batch`.
2. Durable per-index cursor state for the tap, separate from the flush checkpoint. **This is the
   sharp edge:** rotated WAL generations are deleted once every entry in them is verified durable
   in a segment (`WalWriter::prune_verified`, `engine/crates/xerj-storage/src/wal.rs:1003`, and
   `delete_generation` at `:1048`). A tap that falls behind a flush loses its source data unless
   retention accounts for it — and making retention wait on a remote target is exactly the kind of
   coupling that turns a slow target into a disk-full incident.
3. An outbound `_bulk` sender with backpressure and a bounded, non-blocking behaviour when the
   target is down or slow.
4. Config + a REST surface for the allowlist, with system indices excluded.

## Open questions for the design (not answered here)

These are the maintainer's asks in [#171](https://github.com/xerj-org/xerj/discussions/171), restated so they are tracked:

- WAL hook point, and what it costs the write path.
- Ordering and dedup semantics across restarts — including whether at-least-once with `doc_id` as
  the `_bulk` `_id` is sufficient, and what happens to `UpdateMapping`.
- Backpressure when the target is down or slow, and its interaction with WAL retention above.
- What "near-real-time" degrades to when the tap falls behind — and whether that degradation is
  visible or silent.

**@Vinz2168 owns the design here.** This issue exists so there is something to attach it to; it is
not a design decision and carries no commitment that the feature lands.

## Also raised in [#171](https://github.com/xerj-org/xerj/discussions/171), not this issue

The discussion collected three other enterprise asks — engine-level encryption at rest,
data-lake/warehouse connectors, and audit-log hardening (current audit log covers only `_search`
and API-key operations, retains 4096 entries (`DEFAULT_AUDIT_CAPACITY`, `xerj-engine/src/audit.rs:69`), and its read endpoints are not privilege-gated — `Privilege::AuditRead` exists at `xerj-api/src/authz.rs:250` with no enforcement call site, and `audit_search`/`audit_verify` take no auth extractor at `xerj-api/src/native.rs:996`,`:1005`).
Their current state is answered in
[the maintainer's reply](https://github.com/xerj-org/xerj/discussions/171#discussioncomment-17975506).
They need separate issues; this one is scoped to the WAL tap because it is the only ask in the
thread with a named user, a concrete use case, and a proposed implementation.

Discussion: [#171](https://github.com/xerj-org/xerj/discussions/171)

xerj-org:
Push a filtered subset of indices to an external ES-compatible target — the
single-node WAL tap. Native `_xerj/wal_tap` surface, `_bulk` on the wire, not
CCR. Closes #320.

**This PR was written and prepared by an AI agent.** This body was last rewritten
by an agent against the diff at `d64b7f7`, to match what the code now does.

---

## What the diff does

23 files, +5,915 / −82, on top of `main` at `ca4d75a` (the branch's merge-base is
`main`'s tip, so this is a fast-forward).

**New feature surface**

- `engine/crates/xerj-engine/src/wal_tap.rs` (new, 1,516 lines) — the tap. A
  background task tails each allowlisted index's WAL files from disk on a timer
  and POSTs `{target_url}/_bulk`. It takes no lock the write path wants; a
  caught-up poll reads the 16-byte WAL header and stops.
- `engine/crates/xerj-api/src/wal_tap_api.rs` (new, 316 lines) +
  `router.rs` — `GET`/`PUT`/`DELETE /_xerj/wal_tap` and
  `GET /_xerj/wal_tap/_stats`. Runtime-adjustable, persisted next to the
  cursors, `0600` because the state file holds `target_auth`.
- `engine/crates/xerj-common/src/config.rs` — new `[wal_tap]` section:
  `enabled`, `target_url`, `target_auth`, `indices`, `poll_interval_ms`,
  `max_batch_docs`, `max_batch_bytes`, `request_timeout_secs`,
  `max_retry_backoff_secs`, `min_retained_generations`. All optional, off by
  default.
- `engine/crates/xerj-engine/Cargo.toml` / `Cargo.lock` — `reqwest` added to
  `xerj-engine`. Already in the workspace via `xerj-ai`, so no new external
  dependency enters the tree.

**Security**

- `engine/crates/xerj-api/src/authz.rs` — **the whole `/_xerj/*` namespace is
  superuser-only**, not just `wal_tap`. `PUT /_xerj/wal_tap` is a whole-node
  export primitive: `target_url` names any host, `indices: ["*"]` names every
  index, and the tap attaches an operator-supplied `Authorization` header —
  exfiltration and authenticated SSRF in one call. It previously reached
  `Target::Cluster` through the `_`-prefix arm with `BodyShape::None`, so a
  mutating cluster verb naming nothing fell through to
  `Principal::Unscoped => Ok(())` and every unscoped key on the node could use
  it. The reads are covered by the same arm — `GET /_xerj/wal_tap` echoes
  `target_url`, `_stats` enumerates the node's whole index inventory.
- `WalTapConfig::check_target_url` refuses userinfo in `target_url`:
  `https://user:pass@host` is `target_auth` by another spelling in the one field
  the API echoes and the boot log prints.

**Storage**

- `engine/crates/xerj-storage/src/wal.rs`, `index_store.rs` — a bounded WAL
  retention floor (`wal_tap.min_retained_generations`, default `0`), enforced
  inside `IndexStore::wal_maintain_all_verified` (the prune pass the engine
  actually runs), plus the tail reader the tap uses. Deliberately a **floor**,
  not an Elasticsearch-style retention lease: it holds at most
  `n × storage.wal_max_size_mb` per WAL shard per index however far behind a
  consumer is, so a dead target cannot wedge the local disk.
- `engine/crates/xerj-engine/src/engine.rs`, `index.rs` —
  `Engine::apply_wal_retention_floor` pushes the floor onto live `WalWriter`s,
  `Index::{set_,}wal_min_retained_generations` and `Index::wal_dir` expose it.

**Docs / config surface**

- `engine/xerj.default.toml`, `engine/README.md`, `landing/docs/config.html`,
  `CHANGELOG.md`.
- **This PR takes the documented config surface from 105 to 115 settings**
  (`journey_zero_config`'s `EXPECTED_SETTINGS`, counted by the test, not by
  hand) — the 10 `[wal_tap]` fields above.

**Tests / tooling**

- `xerj-engine/tests/wal_tap_push.rs` (new, 1,371 lines),
  `xerj-api/tests/wal_tap_retention_floor_is_applied.rs` (new, 250 lines), plus
  unit tests in `wal.rs`, `config.rs`, `wal_tap.rs` and `authz.rs`.
- `scripts/verify-wal-tap.sh` (new) — boots two real nodes end to end. Not
  wired into CI.

---

## Review history — the last round

Round 2 was rejected on three items; a fourth and two hardening items were found
while fixing them. Each fix has a test that fails before the source hunk and
passes after; every fail-before below was produced by reverting **only** the
source hunk, tests untouched.

### 1. `min_retained_generations` reached no `WalWriter`, ever

**(a) The floor was implemented in a function nothing calls.**
`WalWriter::prune_verified` — where the floor lived and was unit-tested — has no
production caller anywhere in the tree:

```
$ git grep -n "prune_verified(" d64b7f7 -- engine/crates
engine/crates/xerj-storage/src/wal.rs:1080   (the definition)
engine/crates/xerj-storage/src/wal.rs:2570,2583,3582,3589,3598   (its own unit tests)
```

The prune the engine actually runs is `IndexStore::wal_maintain_all_verified`
(`index_store.rs:2649`, deleting at `index_store.rs:2778`), which never consulted
the floor. So on a live node every rotated generation was still deleted the
moment a flush made it redundant.

The floor is now enforced in that loop, before the prune-cache lookup so a held
generation is never decoded. Default `0` ⇒ `keep = 0` ⇒ the loop is unchanged:
no extra syscall, no behaviour change.

**(b) A floor set at runtime reached nothing, and the API said otherwise.**
`store_config_from` reads `Engine.config`, an `Arc<Config>` written once at boot
from `xerj.toml` and never mutated; `WalTap::set_config` touched only the tap's
own `RwLock` and the state file. The value therefore applied to no open index,
and not after a restart either (the boot `Config` is the *file's* value, not the
persisted overlay) — while the `200` carried a `warning` promising it would
apply on restart.

`PUT` now pushes the floor onto the live `WalWriter`s
(`Engine::apply_wal_retention_floor`) and answers with
`retention_floor_applied_to_indices` instead of a warning. `DELETE` releases it
the same way. Every open/create site applies it. `Engine::new` folds the tap's
effective configuration into the boot `Config` so a store is seeded correctly
during WAL replay, and `WalTap` now owns the file config verbatim so `DELETE`
still reverts to `xerj.toml`.

### 2. The health counter charged per `_bulk` chunk, not per poll

At `max_batch_bytes = 1` one legitimate at-least-once redelivery of four
documents produced four all-conflict `_bulk` responses, tripped the
three-in-a-row threshold on its own, and reported `healthy: false` with
`last_error: "… nothing is being replicated"` about a poll in which every
document was already at the target — on the path the module documents as
expected and benign. The field is now `consecutive_unapplied_polls`
(`wal_tap.rs:316`), rises by at most one per poll, and three unapplied polls in
a row still reports the stall.

### 3. `arm_backoff` could overflow

`config.max_retry_backoff_secs.max(1) * 1000` — a plain `u64` multiply two lines
after a deliberate `saturating_mul` on the same expression. `Config::validate`
did not bound the field, so any `u64` in `xerj.toml` reached it. Every multiply
now saturates (`wal_tap.rs:683-690`).

### 4. (found while fixing 3) the config file was the way around the API's bounds

`PUT {"min_retained_generations": 100}` is a `400` — the knob costs
`n × storage.wal_max_size_mb` per WAL shard per index, which is exactly why the
endpoint refuses it. `xerj.toml` took the same `100` in silence and the node
then held 100 rotated generations per shard forever.

The six inline checks in `put_wal_tap` are now one shared
`WalTapConfig::check_limits` (`config.rs:1886`), called from both
`Config::validate` (`config.rs:206`) and the handler (`wal_tap_api.rs:187`) so
they cannot drift. Bounds unchanged from what the API already enforced.
Precedent is `compression.block_size_docs` twenty lines below in the same
function (#318). **This item added no setting** — the count stayed at 115 across
it; the 105 → 115 increase is the `[wal_tap]` section itself, described above.

### 5. hardening, not a fix: no DashMap guard held across a WAL mutex

`apply_wal_retention_floor` called `set_wal_min_retained_generations` — which
locks every `WalWriter` of an index — from inside a `DashMap::iter`, whose guard
lives for the length of the iterator. That introduced a lock-order edge between
the index registry and the storage layer where none existed. No caller can close
the cycle today (nothing holds a WAL shard mutex while mutating
`Engine.indices`; `wal_maintain_all_verified` is in xerj-storage and cannot
reach the engine), so this is not a live bug — but the edge bought one avoided
`Vec` allocation on a path only an operator's `PUT`/`DELETE` triggers. The
handles are snapshotted and the guards dropped before any WAL mutex is taken.

### 6. the new API tests could have passed vacuously

`wal_min_retained_generations()` returns one entry per WAL shard, and every
assertion in `wal_tap_retention_floor_is_applied.rs` was an `.iter().all(...)` —
true for an empty vec. If the accessor ever returned nothing, the tests would
have gone green while reading nothing. Same shape as a CI gate that is a libtest
name filter.

The two retention-floor tests now capture the observed shard count into a local
(`let shards = before.len()`), assert `shards > 0` **before** any `all(...)`
runs, and re-assert `.len() == shards` at every later read — including after the
restart. They deliberately do **not** hard-code a number and do not read
`engine.ingest_shards`: its default is `(cpus / 2).max(1).next_power_of_two()`,
so a literal would be machine-dependent. The third test in the file
(`an_absurd_retry_cap_is_refused_at_the_api`) reads no shards and needs no such
guard.

---

### By the agent that rewrote this body (against `d64b7f7`)

```
$ gh pr checks 322 --repo xerj-org/xerj
16/16 checks pass, 0 pending, 0 fail — including Build + Test,
Format + Clippy, ES-compat YAML conformance, Release build (fat LTO),
Fuzz harnesses, cargo-audit, and the three Autoindex FD smokes
(ubuntu/macos/windows).

$ git merge-base origin/main d64b7f7
ca4d75a99b45dd3d7095b92b4c0d4e37023f6a99      # == origin/main tip; no rebase needed

$ git grep -n "prune_verified(" d64b7f7 -- engine/crates
6 hits, all in xerj-storage/src/wal.rs (1 definition + 5 in its own unit tests)

$ git grep -n "EXPECTED_SETTINGS" ca4d75a -- .../product_experience.rs   → 105
$ git grep -n "EXPECTED_SETTINGS" d64b7f7 -- .../product_experience.rs   → 115
```

and read in full: `wal_tap_retention_floor_is_applied.rs`, the `authz.rs` diff,
`apply_wal_retention_floor`, the `wal_maintain_all_verified` prune loop,
`WalTapConfig` + `check_limits` + `check_target_url`, and the `CHANGELOG.md`
entry.

### By the agent that prepared the branch

Fail-before proofs (revert source hunk only, tests untouched):

**Blocker 1a** — forced `keep = 0` in `wal_maintain_all_verified`:
```
$ cargo test -p xerj-storage --profile ci-test --lib the_retention_floor_holds_generations_through_engine_wal_maintenance
assertion `left == right` failed: the floor must hold 2 rotated generations per shard
through the maintenance path the engine actually runs; files on disk: [16]
  left: 0
 right: 2
```
After: `test result: ok. 1 passed; 0 failed`.

**Blocker 1b** — removed `apply_wal_retention_floor` from the handlers, the
per-open applies, and the `Engine::new` fold:
```
$ cargo test -p xerj-api --profile ci-test --test wal_tap_retention_floor_is_applied
an index open BEFORE the PUT must hold the new floor on every WAL shard,
got [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0]
test result: FAILED. 1 passed; 2 failed
```
Narrower revert of **only** `config.wal_tap = wal_tap.config()` in `Engine::new`:
```
assertion `left == right` failed: the boot Config a store opens with must carry
the effective floor, not the file's
  left: 0
 right: 3
test result: FAILED. 2 passed; 1 failed
```
After both restored: `test result: ok. 3 passed; 0 failed`.

**Blocker 2** — restored `saturating_add(unapplied_chunks)`:
```
$ cargo test -p xerj-engine --test wal_tap_push --profile ci-test a_multi_chunk_redelivery_is_one_unapplied_poll_not_four
assertion `left == right` failed: one poll that applied nothing is ONE, however many chunks it took
  left: 4
 right: 1
```
After: `ok`.

**Blocker 3** — restored the plain `* 1000`. Fails in **both** profiles, by
construction (`18_446_744_073_709_552 * 1000 == 2^64 + 384`):
```
$ cargo test -p xerj-engine --profile ci-test --lib an_unbounded_retry_cap_saturates_instead_of_wrapping
a cap too large to express in ms must not become a SMALL cap; the wrapped value
gives at most 384ms. got 192ms
test result: FAILED

$ cargo test -p xerj-engine --lib an_unbounded_retry_cap_saturates_instead_of_wrapping   # dev
panicked at crates/xerj-engine/src/wal_tap.rs:661:19: attempt to multiply with overflow
test result: FAILED
```
After: `ok` in both.

**Item 4** — removed the `check_limits` call from `Config::validate`:
```
$ cargo test -p xerj-common --profile ci-test --lib wal_tap_numeric_knobs
panicked: [wal_tap] min_retained_generations = 100 must be refused at boot
```
After: `test result: ok. 91 passed; 0 failed` (whole `xerj-common` lib).

Gates, run locally before the final push:

```
$ cargo fmt --all -- --check                                          # clean
$ cargo clippy --workspace --all-targets -- -D warnings                # clean (CI's exact gate)
$ cargo test -p xerj-storage --profile ci-test    → 167 passed, 0 failed (+1 passed 1 ignored)
$ cargo test -p xerj-storage --lib                → 167 passed, 0 failed   (dev, overflow-checks ON)
$ cargo test -p xerj-api --profile ci-test        → 24 binaries, 0 failed
$ cargo test -p xerj-api -p xerj-common --profile ci-test --no-fail-fast
                                                  → 26 binaries, 0 failed (re-run after item 4)
$ cargo test -p xerj-engine --profile ci-test --test product_experience journey_zero_config
                                                  → ok (settings count 115)
$ cargo test -p xerj-engine -p xerj-common --profile ci-test --no-fail-fast
                                                  → all green except the known #372 flake
$ cargo test -p xerj-engine --profile ci-test --lib -- --test-threads=1
                                                  → 557 passed, 0 failed (re-run at final head)
$ ./target/ci-test/xerj --config ./xerj.default.toml
                                                  → shipped-config boot OK (HTTP 200)
$ cargo test -p xerj-engine --test wal_tap_push --profile ci-test → 21 passed
$ cargo test -p xerj-engine --test wal_tap_push                   → 21 passed  (dev)
$ cargo test -p xerj-engine --lib wal_tap:: (both profiles)       → 9 passed
$ cargo test -p xerj-api --test wal_tap_retention_floor_is_applied (both)  → 3 passed
$ cargo test -p xerj-storage --lib wal (both profiles)            → 39 passed
```

ES-compat conformance, locally against a live `ci-test` server on :9200
(CI's own `ES-compat YAML conformance` job also passes on this head):

```
$ ./target/ci-test/es-yaml-runner --dir tests/es-compat-yaml/yaml
```

`index::merge_publication_transaction_tests::partial_multi_batch_failure_is_reported_without_losing_either_batch`
is the pre-existing #372 full-parallelism flake; it passes at `--test-threads=1`
and is untouched by this PR.

### Reference coding

Retrieved from the `refsym` index during preparation; the quoted text is the
symbol body.

- `KeepLastNCommitsDeletionPolicy.onCommit`,
  `lucene/core/src/java/org/apache/lucene/index/KeepLastNCommitsDeletionPolicy.java:51-58`
  — "The commits list is already sorted from oldest to newest /
  `for (int i = 0; i < size - numCommitsToKeep; i++) commits.get(i).delete();`".
  Same arithmetic, and the same "enforce inside the deletion pass, not in its
  callers" placement, as the fix to `wal_maintain_all_verified`.
- `LiveIndexWriterConfig`,
  `lucene/core/src/java/org/apache/lucene/index/LiveIndexWriterConfig.java:39-126`
  — "few setters for settings that can be changed on an IndexWriter instance
  'live'", against `IndexWriterConfig.setCodec:297-305` "Only takes effect when
  IndexWriter is first created". A setting is one or the other and says which;
  this one claimed live and was implemented at create.
- `IndexFileDeleter.revisitPolicy`,
  `lucene/core/src/java/org/apache/lucene/index/IndexFileDeleter.java:516-543`
  — re-runs `policy.onCommit(commits)` over commits that already exist rather
  than waiting for the next one; the model for
  `Engine::apply_wal_retention_floor`.
- `BufferedIndexInput.refill`,
  `lucene/core/src/java/org/apache/lucene/store/BufferedIndexInput.java:289-317`
  — the sliding-window shape of the WAL tail reader.

Apache-2.0, same licence as XERJ; adapted with attribution, not copied.

---

## ASSUMED — not verified

- **`scripts/verify-wal-tap.sh` was not re-run** in the final round, and is not
  wired into CI. It boots two real xerj nodes; no code it touches changed in the
  last round, and XERJ's own `_bulk` still ignores per-action `version_type`
  (only the single-doc path honours it, `es_compat.rs:2990`), which is why the
  external-versioning stub in `wal_tap_push.rs` exists.
- **The `refsym` citations above were not re-retrieved when this body was
  rewritten** — the `refsym` index was not resolvable on the local node at that
  time. They are carried forward from the preparing agent's round as-run.
- **The CHANGELOG entry says the retention test reads `3` on "all 16 WAL
  shards".** 16 is what `engine.ingest_shards` resolved to on the machine that
  run happened on — its default is `(cpus / 2).max(1).next_power_of_two()`, so
  the figure is machine-dependent. The test derives the count rather than
  asserting 16, so it is correct on any host; only the prose figure is specific
  to that run.
- **Multi-shard retention cost under a large allowlist is reasoned, not
  measured.** The floor holds `n × wal_max_size_mb` per shard per index by
  construction, but no disk-usage run was done at `indices = ["*"]` on a
  thousand-index corpus.
- **`retention_floor_applied_to_indices` counts open indices, including any
  closed/failed ones present in the `indices` map**; the tests assert `1` on a
  single-index node, not the boundary cases.
- **The default path is asserted unchanged, not benchmarked.** At `floor = 0`
  the new prune-loop lines are `keep = 0` / `prunable_prefix = len` and the
  per-generation `continue` is never taken, so there is no extra syscall — but
  no before/after ingest benchmark was run to confirm no regression.
- **The deadlock analysis behind item 5 is by inspection, not by a stress
  test.** Every writer of `Engine.indices` and every holder of a WAL shard mutex
  was traced and no path holds one while wanting the other; no concurrent repro
  was built, which is part of why the guard was removed rather than argued
  about.

## Known residual, not fixed here

`WalWriter::prune_verified` is dead code: nothing outside `wal.rs` and its own
unit tests calls it. It is how this round's blocker happened — a fully
unit-tested retention floor that had no effect on a live node, because the
function it lived in is not the one the engine runs. Its floor is now redundant
with the enforced one in `wal_maintain_all_verified`. Deleting a public
`WalWriter` method mid-review is risk without benefit, so it is left in place
and should get its own issue: either wire it or remove it, because a
well-covered function that nothing calls will manufacture this class of bug
again.

Carried forward from round 2 and still agreed non-blocking: the narrow
offline delete-and-recreate window, `fill_to_end` on the corruption path, and
the sub-header-length `gaps` increment.


xerj-org:
Rebased onto `main` (past #316 — `Engine::new` gained the `cluster_state::preflight` block right where the tap's config clone sits; both kept). The PR was `CONFLICTING` before that, which is why no CI run appeared on the first push.

Re-verified on the rebased tree:

| step | result |
|---|---|
| `cargo fmt --all --check` | clean |
| `cargo build --release -p xerj-server` | clean |
| `cargo test -p xerj-storage --lib wal::` | 22 passed, 0 failed |
| `cargo test -p xerj-engine --test wal_tap_push` | 12 passed, 0 failed |
| `cargo test -p xerj-common --lib` | 80 passed, 0 failed |
| ES-compat YAML suite | **1366 passed · 0 failed · 3 skipped** |

xerj-org:
## Rebased onto current main, re-proved end to end — green, awaiting owner's merge

**The red CI was never this PR.** The 5-ok/10-fail signature was the
`tree-sitter` `links=` collision that made main itself unbuildable (fixed by #350).
Rebasing onto current main was the entire fix: **16/16 checks pass**, and the
feature files are byte-identical across the rebase — `git diff 2828135e cf83a93d`
is empty for `wal_tap.rs`, `wal_tap_api.rs`, `wal.rs`, `wal_tap_push.rs` and
`verify-wal-tap.sh`.

### One thing the rebase broke, now fixed (0dea96d5)

`config.rs` carried the `114`-settings count in a **merge commit**, which a rebase
cannot replay: main had added a second banner comment quoting the total, and the
old `Merge branch 'main'` resolved it to 114. Flattening reverted that line to
`105` while every other site says 114. Nothing fails on it — it is a comment —
which is exactly why it would have shipped. Restored, plus the
`journey_zero_config` sibling comment that was already stale, plus two places
where the docs wrote the guarded prefix as `.xerj-*` when the code guards every
`.xerj` name (`.xerj_users` and `.xerj_api_tokens` hold hashed credentials and do
*not* match `.xerj-*`).

### Feature re-proved with two real nodes

`scripts/verify-wal-tap.sh` — **9/9 PASS**. Plus an independent 20-check probe:

| what | result |
|---|---|
| allowlisted `edge-*` reaches the target | 3 docs |
| the source really holds the skipped index | `orders` = 2 docs locally |
| the target has no `orders` **at all** | HTTP 404 |
| skipped for the *filter*, not a failure | `_stats` omits `orders`; live index shows `bulk_failures=0 item_failures=0 gaps=0 last_error=null` |
| widening to `["*"]` makes it ship | `orders` becomes tracked and lands |
| `["*"]` still ships no system index | 0 `.xerj*` tracked |
| a `.xerj` pattern is refused with a reason | 400 `"only matches system indices, which are never shipped"` |

### The prior finding still holds after the rebase

A first look at an **established** index (history written un-allowlisted, then
`_flush` to prune generation 0, then allowlisted at runtime):

```
lag_seq 0 · docs_shipped 0 · last_shipped_seq 5   ← seeded from head, not left at 0
target: HTTP 404 for hist-1                       ← history skipped, not shipped
```

Without the seeding this reads `lag_seq 5`. A write *after* the first look then
ships normally and lag stays at 0 — so a genuinely-behind tap still reports real lag.

Worth stating because it surprised me first time: **"no backfill" has two
branches**, both behaving as documented. A never-pruned index joins the allowlist
and is shipped **whole** (`orders` sent all 3); a pruned/established one starts
from now (`hist-1` sent none).

### Gate

`cargo fmt --all --check` clean · `cargo clippy -p xerj-common -p xerj-storage
-p xerj-engine -p xerj-api -p xerj-server --all-targets -D warnings` clean ·
`cargo test --profile ci-test` on those five crates **1571 passed / 0 failed /
26 ignored** across 75 binaries · ES-compat YAML run locally against a private
port: **1366 passed / 0 failed / 3 skipped**.

### One unrelated observation, not a blocker

`_cat/indices?h=index` is **not honoured** — it returns the full column set. That
makes `verify-wal-tap.sh`'s cosmetic line report "the target bootstraps its own 0
system indices" when it had bootstrapped 14 (its `^\.xerj` grep can never match a
row starting with `green`). The load-bearing assertion next to it is a separate
`_count` and is unaffected, so the script's verdict stands. The `h=` gap looks
pre-existing and out of scope here.

Not merging — leaving that to the owner.

xerj-org:
Reviewed by two independent agent passes (Claude Opus 5), the second tasked with refuting the first. It tried to get this to "merge" and could not; it found two further defects instead, one of them a credential leak. Reference points below came from retrieval against an indexed `apache/lucene` checkout.

CI is genuinely green here (16/16 including the fat-LTO all-members-link job), it merges cleanly, and it is the only open PR touching `Cargo.lock` — but the new dep (`reqwest`, `default-features=false`, `rustls-tls`) is already a workspace dep linked by `xerj-server`, so the #352 hazard does not apply. The problems are in the feature, not the integration.

## Security

**1. `PUT /_xerj/wal_tap` is a whole-node export primitive reachable by any unscoped API key.** `classify` routes `/_xerj/...` to `Target::Cluster` via the `s.starts_with('_')` arm (`authz.rs:487`), `body_shape` returns `BodyShape::None` (`authz.rs:702`) so `body_targets` yields nothing, `cluster_is_read` is false for PUT, and the fall-through is `Principal::Unscoped { .. } => Ok(())` (`authz.rs:1391`).

So an unscoped key can set `target_url` to a host it controls with `indices: ["*"]` and stream every non-system index off the node. That is also an authenticated SSRF, with an operator-supplied `Authorization` header attached to the outbound request.

`authz.rs:1331-1337` already makes the opposite call for the only comparable endpoint, with the reasoning written out: *"Snapshot and restore that name no indices cover every index on the node, brains included. There is no per-index decision to make, so they are superuser-only."* The WAL tap is that exact shape.

Separately, `GET /_xerj/wal_tap/_stats` iterates `index_name_list()` with no principal filter (`wal_tap_api.rs:180`), and `_xerj` is absent from `prune_sites` (`authz.rs:1553-1600`), so a key scoped to one index gets the node's full index inventory. `verify-wal-tap.sh` runs with `[auth] enabled = false` and the PR adds no authz test, so none of this is covered.

**2. Credential leak that defeats this PR's own invariant.** `config_body` (`wal_tap_api.rs:158-170`) deliberately omits `target_auth`, with a comment that echoing it *"turns 'can call the admin API' into 'holds the target's credential'"*. But it returns `target_url` verbatim, and `https://elastic:changeme@central:9200` is an ordinary way to point at an ES cluster. `put_wal_tap` validates only the scheme prefix, and reqwest turns URL userinfo into a Basic `Authorization` header, so it functions exactly as `target_auth` does.

That credential is echoed by both `GET /_xerj/wal_tap` and `_stats`, both `cluster_is_read` so every principal including a one-index scoped key can read it, and `main.rs:2193` logs `target = %cfg.target_url` at startup. The verification script's check is `grep -c 'target_auth"'`, which cannot catch it. Reject userinfo in validation, or redact before echoing.

## Correctness

**3. Silent data loss, or a permanent silent stall, after delete-and-recreate.** `forget_deleted` (`wal_tap.rs:788-802`) only drops a stale cursor when a tick happens to observe the index missing. At the 500 ms default, a `DELETE` immediately followed by a `PUT` (deploy script, ISM rollover, delete-then-reindex) is never observed. The recreated index gets a new WAL at generation 0, so the `cursor.generation > newest` guard never trips. `tail_wal_file` then resumes at the old byte offset:

- new file shorter: clamps to EOF, `parse_raw_frame` returns `End`, every document before that poll is skipped with `gap = false`
- new file already past that offset: resume lands mid-frame, `Corrupt` on CRC mismatch, `allow_resync` is false because gen 0 *is* `newest`, and the shard holds that offset forever

In both cases `gaps` stays 0, `last_error` null, `bulk_failures` 0. That is verbatim the failure the PR's own test comment calls *"the worst failure mode a WAL consumer can have, because `_stats` would show a healthy tap shipping nothing"* (`wal_tap_push.rs:512-516`). The test passes only because it inserts a `tick()` between the delete and the recreate at line 529.

**4. `_stats` reports healthy and zero-lag while shipping nothing, whenever the target rejects items.** `poll_index` (`wal_tap.rs:766-775`) derives `docs_shipped` / `last_shipped_seq` from render-time values and unconditionally clears `consecutive_failures` and `last_error` on any 2xx. Nothing consults per-item outcomes. Combined with `version_conflict_engine_exception` being documented as benign (`:566-567`, `:168-171`), a target holding documents at versions >= the source's seq_no makes every action a rejected no-op while `_stats` shows `lag_seq: 0` and `docs_shipped` climbing.

**5. The central safety claim has zero test coverage.** The stub target answers a hardcoded `{"took":1,"errors":false,"items":[]}` (`wal_tap_push.rs:145`), so the entire `errors == true` branch of `send_bulk` is dead code under test. And `scripts/verify-wal-tap.sh` runs two real xerj nodes, but xerj's own `_bulk` ignores per-action `version`/`version_type` (only the single-doc path honours it, `es_compat.rs:2990`) — so the end-to-end script exercises precisely the target that ignores the mechanism the design rests on.

## Performance

**6. The tap re-reads the entire active WAL generation on every poll.** `tail_wal_file` (`wal.rs:1873-1874`) does an unconditional `fs::read(path)` then seeks inside the resulting `Vec`. Fully caught up, it still reads the whole active generation into a fresh allocation to discover EOF. Cost per poll is O(active WAL bytes) x `num_wal_shards` x allowlisted indices, and the defaults compound: `poll_interval_ms = 500`, `num_wal_shards` 16 on a 32-core box, generation rotated only at flush (up to 30 s). Within one flush window the tap reads the generation ~60 times, averaging half of it: roughly **30x read amplification of the node's entire ingest volume, per shard**. At the permitted floor of 50 ms it is 300x.

This repo already fixed this bug class and wrote down why — `index_store.rs:571-584` documents `wal_prune_cache` as the RC4 W1 #8 follow-up, where every 1 s tick re-decoded every retained generation. Same shape, no cache, twice the cadence.

Lucene's NRT replication does the incremental thing: `SimpleTransLog.replay(primary, start, end)` reads frame-by-frame from a `FileChannel` at `pos` via `readBytesFromChannel` (a 4-byte length, then exactly that many bytes) and never reads the whole file. A `seek`-based read from `from_offset` is both the fix and simpler than what is there.

## Also

- **No retention floor.** With `flush_interval_secs = 30` the WAL window is about one flush, so a minute of target downtime on a hot index loses data, and the only mitigation offered is a log line. Lucene solves the same problem with a bounded, releasable lease (`SnapshotDeletionPolicy.snapshot` / `PrimaryNode.releaseCopyState`). A `min_retained_generations` knob is the obvious middle path and is offered nowhere.
- **Runtime config is not durable.** `PUT /_xerj/wal_tap` config lives only in an in-memory `RwLock`. After a restart the tap silently reverts to the file config (typically `enabled = false`), cursors freeze, WAL pruning continues, and `_stats` will not even list the affected indices. The headline "runtime config, no restart" implies a durability that does not exist.
- **`persist()` rewrites the whole cursor file per index per poll** (`wal_tap.rs:327-342`), so O(N^2) bytes and 2N fsyncs per tick. `index_store.rs:806-813` records 9,382 indices producing 9,709 threads and 718-760% CPU while idle; `indices = ["*"]` on an autoindex node walks into that.

## Suggested order

(1) and (2) first, since they are reachable by any authenticated key. Then (3) — key the cursor on an index incarnation id rather than the name, or validate that the offset lands on a frame boundary. (6) is a contained change to `tail_wal_file`. (5) mostly needs a stub that actually implements external-version semantics and returns `errors: true`, which would then cover (4) as well.

Issue #294, opened 2026-08-11
autoindex: --no-graph durable path silently indexes ZERO documents for every code file (extensionless snapshot blob defeats grammar lookup)


xerj-org:
## Summary

On the durable-generation path (`--no-graph`), **every source-code file is silently
dropped**: it is walked, sniffed correctly as `family: code`, counted as a file, and
then indexed as zero documents. The run reports success (`generation N committed`,
exit 3 `completed-with-junk`) with no message naming the files that vanished.

This is the #204 class: the pipeline accepts the file and silently indexes nothing.

## Root cause

`sync_executor.rs` sniffs with the *logical* path but extracts with the *blob* path:

```rust
// crates/xerj-autoindex/src/sync_executor.rs:1445
let sniffed = crate::sniff::sniff_with_name(snapshot_blob, &source.path)   // logical name → Family::Code  ✅

// crates/xerj-autoindex/src/sync_executor.rs:1529
let stats = crate::extract::extract(snapshot_blob, &sniffed, None, &mut sink)  // blob only  ❌
```

`code::extract` is the **only** family extractor that recovers its parameter from the
path rather than from `Sniffed`:

```rust
// crates/xerj-autoindex/src/extract/code.rs:115-124
pub fn extract(path: &Path, sink: Sink) -> Result<ExtractStats> {
    let ext = path.extension().and_then(|e| e.to_str()).unwrap_or("")...;
    let Some(def) = registry().iter().find(|d| d.exts.contains(&ext.as_str())) else {
        stats.junk += 1;
        return Ok(stats);      // ← zero documents, never surfaced
    };
```

Snapshot blobs are content-addressed and **extensionless** — `sync-snapshots/<tx>/blobs/00000000`.
So `path.extension()` is `None`, no `LangDef` matches, and the file becomes junk.
Every other extractor (`txt`, `csv`, `json`, `html`, …) takes its config from `sn`, so
only the code family is affected — but it is affected for *every* language.

## Reproduction (verified, v1.0.0-rc.15 @ main bc01dd7a)

Two files with the same Python body, differing only in extension:

```
$ printf 'def alpha_helper():\n    return 1\n' > repro/app.py
$ printf 'def beta_helper():\n    return 2\n'  > repro/notes.txt
$ xerj autoindex repro --url http://localhost:9483 --prefix repro2 \
      --state-dir /tmp/st --no-graph --progress none
```

Result — the prose file indexes, the code file does not:

```
total: 1
[...]

PR #323, opened 2026-08-12
docs(llms): lead with reference coding, and state what we expect back from AI

xerj-org:
`llms.txt` is the file AI agents read before deciding what XERJ is and what to do with it. It was failing at both jobs: **reference coding sat at section 12 of 14** (below the recipe list) and the **contribution protocol at 13** — an agent that stopped reading after the endpoint map saw neither. The file had also grown to 50 KB, larger than `llms-full.txt`, the reference it is meant to be a short front door to.

Restructured so the first two things an agent reads are the two things we want from it.

**1. `## Start here — reference coding`** — the problem (unmemorised API → retry loops burning output tokens; a compiler leaks a *name*, never a *contract*; grep pulls whole files into context), the four-step loop, the session-activation prompt, and the evidence table from the published study:

| arm (8 tasks × 4 languages, 16 runs) | output tokens | cost | solved |
|---|---:|---:|---:|
| bare — tools off, memory only | 260,916 | $11.18 | 11/16 |
| native Claude Code — greps the source | 26,477 | $3.27 | 16/16 |
| XERJ — same agent, reference retrieved | 9,982 | $1.58 | 16/16 |

Plus the 21/21-vs-1/21 result on the seven contract-bearing domains, the memorisation ceiling (retrieval is neutral-to-harmful on code the model already knows), and licence discipline.

**2. `## Your side of the deal — file the issue, send the pull request`** — the ask stated as the price of a free Apache-2.0 engine, with the three-outcome rule: reproducible defect → ONE issue (exact curl, verbatim output, expected + why, version, clean data dir, closed issues searched first); patched AND ran → a pull request (a test watched to fail before the fix, `cargo fmt`, scoped release build, ES-YAML at 0 failed, root-caused commit body); unrun suspicion → send nothing. Provenance rules, the no-`Co-Authored-By` rule, CLA accountability and the sandboxed-agent fallback travel with it.

## Honesty structure preserved

Every scope stated with its number (2.7× = the 8-task multilang arm; 21/21 vs 1/21 = the seven contract-bearing domains); the ~5× end-to-end figure quoted only as "users report" with its provenance file linked and a measured number adjacent; all five capability caveats intact.

## Size

The restructure alone took `llms.txt` from 50,524 to 30,753 bytes (238 → 165 lines), with the two lead sections taking *more* space than before, not less. Two later commits on this branch — the mandatory field-report protocol, and PR #343's first-contact fixes (see below) — put content back: the file is **48,102 bytes** as it stands, still smaller than main's 52,315, with every added byte answering a failure a probe actually hit.

Nothing was dropped: the 68-line "Running an index for a human" protocol became a 12-line summary in `llms.txt`, and its full contract — decision-payload keys, verbatim `xerj-bar`/`xerj-progress` samples, estimate honesty rules, ignore rules, exit codes — moved verbatim into `llms-full.txt` as new §10. `AGENTS.md` now points there.

## Two defects found while editing

- `llms.txt` sent agents to `/docs/recipes/all-way-search.html` and `/docs/recipes/autoindex-semantic-search.html`. **Neither page exists under `landing/`** — those recipes only ever shipped as markdown under `docs/recipes/`, and the site has no generator that would build them. Both were 404s for every agent that followed them; they now point at the live markdown.
- `llms-full.txt` §2 still said neural embeddings required an external inference endpoint, predating the in-binary BERT encoder. It now documents `--embed-mode neural` / `--embed-mode proxy`, with the rule that output is only called neural when one of those is actually running.

`llms-full.txt` also gains reference coding as the first entry in §4 with the same measured numbers and limits, and its further-reading list now carries the case study and the agent contribution protocol.

## Verification

- Deployed to Cloudflare Pages ahead of merge (the accepted pattern for landing changes). Live and verified: `https://xerj.org/llms.txt` → HTTP 200, 30,753 bytes, byte-identical to this branch, reference coding as §1 and the contribution ask as §2; `https://xerj.org/llms-full.txt` → HTTP 200, 62,679 bytes, carrying §10 and the new §4 entry.
- Every `xerj.org/*` and `github.com/xerj-org/xerj/*` link in the new `llms.txt` was resolved against the repo — 0 missing after the two 404s above were repointed.
- Docs-only change: no engine code touched, so the ES-YAML conformance suite was **not** run.

---

## PR #343 is folded into this branch

`#343` (*docs(llms): close the first-contact defects a 22-run agent study found*) edited the same two files and was agreed to land on top of this restructure rather than beside it. Its thirteen fixes are now re-placed into the new layout in the final commit here, so **#343 can be closed once this merges**.

Two of its fixes were pure placement and are already fixed by the restructure — they are *not* duplicated: reference coding was buried at section 12 (now section 1), and the contribution protocol at 13 (now section 2).

Version scoping was re-checked against **v1.0.0-rc.16** (released 2026-08-13) rather than copied from #343, which was written when rc.15 was current:

- the estimate gate (`--approve` / `--max-minutes` / exit `4`) **is** in a release now — rc.16, not "in no release";
- the `--no-graph` code-junking defect (#294) is **fixed in rc.16** (#316, #338); the warning is kept and re-scoped to "up to and including v1.0.0-rc.15", with the `_count` verification query for anyone on an older binary;
- `--port` now exists on `main` but is in no release, so the doc says ports are TOML-only on rc.16 and earlier and names the flag as coming next;
- `xerj mcp` ships in rc.16, and `mcp-tools.json` is now 10 tools against the OpenAI/Anthropic files' 6 — the new schema section says so instead of claiming all three match.

Verified rather than assumed while re-placing: the code-document fields out of `extract/code.rs`; the worked reference-coding query is the shape `catalog.rs` actually emits (`bool.should` of `multi_match` over `["body","defs"]` plus a `match_phrase` on `defs` boosted 4 — *not* a `defs^N` multi_match, which that code comments against); the installer's checksum fallback chain out of `landing/get`; the release-asset naming off the rc.16 assets.

## CI note

This branch is rebased on current `main`. The red jobs are inherited: `main` itself currently fails at dependency resolution (`error: failed to select a version for tree-sitter`) in the same jobs, with the identical error, on a docs-only diff.


xerj-org:
Cross-reference: this PR and #343 both rewrite `landing/llms.txt` and `landing/llms-full.txt`, and they will conflict. Flagging it now rather than letting whichever merges second discover it.

They are complementary in intent, not competing:

- **#323 (this one) restructures.** Reference coding moves from section 12 of 14 to the front, the contribution protocol comes up with it, and the file shrinks from 50 KB — an agent that stops reading after the endpoint map currently sees neither.
- **#343 fixes content defects** found by a 22-run first-contact study (`demo/playbooks/FTX_PROBE_2026-08-12.md`), where 6 of 12 wave-1 runs hard-failed. Thirteen specific corrections: `--approve fast` documented as the safe option when it selected `--no-graph` and indexed no code; "semantic search" read as neural-by-default by 9 of 12 probes because the lexical caveat sits two screens from what triggers the belief; the node lifecycle (no `--port` flag, ports are TOML-only, bind address and auth-enable unstated); the three agent tool schemas claimed but never linked.

**Suggested order: land #323 first, then rebase #343 onto it.** Structure before content is the cheaper direction — #343 is thirteen localised edits that can be re-placed into new sections, whereas re-applying a whole-file restructure on top of them means redoing the restructure. I am happy to do that rebase.

Worth noting for whoever reviews both: two of #343's findings are about *placement* rather than wording, and #323 may already fix them — the lexical-embedder caveat being far from the text that triggers the misreading, and reference coding being buried. If so, those two edits should be dropped rather than duplicated.

xerj-org:
## Rebased onto current `main` — 16/16 checks green, `MERGEABLE` / `CLEAN`

Head is now `fcb1396`. The branch was `CONFLICTING` at 5 pass / 10 fail; it is now conflict-free and fully green. Not merging — leaving that to the owner.

### The 10 red jobs were the base breakage, not this PR

Confirmed rather than assumed. The old run's `Format + Clippy` log fails at dependency resolution before any Rust is compiled:

```
package `tree-sitter` links to the native library `tree-sitter`, but it conflicts
with a previous package which links to `tree-sitter` as well:
package `tree-sitter v0.26.3`
    ... which satisfies dependency `tree-sitter = "^0.26"` of package `xerj-autoindex`
```

Every one of the ten failed in 8–52 s with that same error, on a diff containing **zero** `.rs`, `.toml` or `.lock` changes. #350 fixed it on `main`; after the rebase `cargo metadata --locked` resolves clean with exactly one `tree-sitter 0.26.12` in `Cargo.lock`. Nothing in this PR was "fixed" to make CI pass.

### Rebase conflict — one, resolved by taking each side's own change

`landing/index.html` footer. `main` had bumped the version string; this branch had changed the slogan. Both kept:

```
  <span>XERJ.AI · V1.0.0-RC.15 · 2026</span>   <- main's bump
  <span>SEARCH FOR THE AGENT ERA</span>        <- this branch's fix
```

The slogan change survived the rebase intact: 16 pages carry `SEARCH FOR THE AGENT ERA`, 0 still say `STORAGE`.

### Overlap with #343 — total in both files, and already reconciled here

The two branches cannot be merged by git. A `git merge-tree` of this head against #343's `71c26bd8` conflicts in **both** files.

- **`llms.txt` — the whole file collides.** This branch's diff against `main` is a single hunk, `@@ -1,239 +1,272 @@`: the 239-line file is rewritten end to end. All ten of #343's hunks (`main` lines 1‑6, 9‑22, 25‑31, 40‑59, 62‑68, 74‑80, 98‑114, 132‑138, 210‑220, 232‑237) land inside it.
- **`llms-full.txt` — region-for-region.** The two branches touch nearly the same anchors: #343 at `main` lines 60, 95, 111, 137, 186, 245, 306, 679; this branch at 60, 90, 111, 153, 202, 273, 338, 705.

I checked #343's thirteen clusters against this branch's files individually rather than trusting the fold-in claim. Twelve are present: the `--no-graph` junking warning with the `{"exists":{"field":"language"}}` `_count` recipe, `--embed-mode` step 0 and the at-ingest rule, the inline lexical qualifiers, node lifecycle (`rest_port`, `Address already in use`, `SIGTERM`, `admin.key`), the `~/xerj-data` quick-start fix, the PATH consequence and `XERJ_INSECURE_SKIP_CHECKSUM` fail-closed correction, version scoping, the three linked tool schemas, the manual verified install, the shell-less `CLI-only` / no-`POST /_autoindex` branch, the index-naming and `--prefix`/`--state-dir` scoping, and the `all-MiniLM-L6-v2` / `local_model_dir` / 20.2 GB honesty patches.

**Where this branch is more correct than #343**, worth recording before #343 is closed:

- #343 taught `{"multi_match":{"fields":["defs^3","body","title"]}}`. The reference-coding toolkit measured that form and it is **worse**: flat `body, defs, title` scores top‑1 3/6 top‑3 6/6, while `defs^2` gives top‑3 4/6 and `defs^3` top‑1 2/6. `catalog.rs:282-295` makes the same argument at the site that emits the query — a boost amplifies incidental token overlap, whereas the phrase clause requires the whole sequence. This branch carries no `defs^N` anywhere.
- #343 was written against rc.15 and said the estimate gate is in no release. It shipped in rc.16, and this branch says so.
- #343 implied the three tool schemas match. They do not, and this branch states the real shapes.

### The one gap I found and fixed

`llms.txt` condensed the field list and closed with *"the full field contract is in llms-full.txt §3.1"*. It was not there — §3.1 covered `_passage` and the code query and mentioned `ax_path` in passing, so an agent following that pointer for `ax_locator`, `ax_dataset`, `ax_format` or `ax_run` hit a dead end. #343 had those inline; the restructure moved the detail out without landing it. Commit `fcb1396` puts the contract where the cross-reference already points, every field read off this branch's source:

- `title` / `language` / `defs` / `symbols` / `symbol_count`, and that the last three are absent when nothing parsed — `extract/code.rs:437-474`
- `ax_path`, `ax_paths`, `ax_file`, `ax_locator`, `ax_dataset`, `ax_run`, `ax_format` — `sync_executor.rs:1496-1513`
- `ax_locator == "code"` — `extract/code.rs:472`

`ax_paths` was in neither file; it is written on every document.

### Claims re-verified against source, not carried forward

- The worked query is byte-for-byte the shape `catalog.rs:303-321` emits — `bool.should` of `multi_match` over `["body","defs"]` `most_fields`, plus `match_phrase` on `defs` `"boost": 4`, `_source: ["ax_path","title"]`, `fields: ["_passage"]`, `size: 3`.
- `--prefix` defaults to `ax` (`cli.rs:136`, `cli.rs:363`); `--state-dir` exists; index names are `<prefix>-<dataset-slug>` (`dataset.rs:348-359`).
- `--embed-mode` is a server-start flag taking `lexical | neural | proxy | auto` (`xerj-server/src/main.rs:149`, `:223`) — the at-ingest framing holds.
- `2.7×` = 26,477 / 9,982 output tokens, consistent with the table above it.
- Tool schemas: `mcp-tools.json` is an object with 10 tools; `openai-tools.json` and `anthropic-tools.json` are bare arrays of 6. Exactly as written.
- Link resolution: every `xerj.org/*` and every `github.com/xerj-org/xerj/{blob,tree}/main/*` target in both files resolves to a real path in the tree — **0 missing**.

### Gates

| gate | result |
|---|---|
| `cargo metadata --locked` | exit 0 — one `tree-sitter 0.26.12` |
| `cargo fmt --all -- --check` | exit 0, no diff |
| `cargo clippy` / `cargo test` | not run — the diff touches no crate (0 `.rs`/`.toml`/`.lock` files). CI's `Format + Clippy` and `Build + Test` both pass. |
| ES-YAML conformance (CI) | **1366 passed · 0 failed · 3 skipped · 1369 total** |
| PR checks | **16 pass, 0 fail** |

The `1,366/1,369` figure both files quote matches what CI just measured.

### For the owner

Ready to merge as-is. #343 can be closed unmerged once this lands — its content is here, three points of it corrected. The `CI note` at the bottom of the PR description is now stale: the reds are gone.

PR #333, opened 2026-08-12
fix(onboarding): make the documented setup work with auth on

buger:

# Following the documented setup and running `xerj autoindex` fails with a 401

A user on macOS downloaded XERJ, created a config file — the default one, exactly as the docs say — ran `xerj autoindex`, and got this:

```
error: semantic autoindex could not pin the server embedding execution identity:
GET /v1/embedding/identity failed: HTTP 401 Unauthorized;
semantic autoindex requires a XERJ server that exposes a resumable embedding identity
```

They concluded their server lacked a feature, edited the config to turn auth off, and moved on.

Reproduced byte-for-byte. The 401 is correct — the request had no credential. Everything around it was not.

---

## Why it happened

**The docs are split into two families that never intersect.**

| doc family | mentions `autoindex`? | passes `--insecure`? |
|---|---|---|
| pages that show `autoindex` | yes | **always** |
| `install.html`, `quickstart.html`, `cli.html` — "write a config" | **zero occurrences** | n/a |

A reader of the second family is never shown the key they need. `xerj.default.toml`'s own header tells you to copy it, and copying it enables auth. The server *does* generate a key on first start and prints the path — but no page in that family mentions it.

**The error blamed the wrong thing, at the wrong time.** `ping()` discarded the HTTP status, so the 401 on the very first round trip was swallowed. The run then printed ~15 lines of healthy-looking progress before dying at the embedding-identity probe with a message about the server's *capabilities*. That is why the reporter concluded the server was missing a feature rather than a credential.

**A zero-config server fails the same way.** `AuthConfig::default()` is `enabled: true`, so "zero config" does not mean "auth off". What actually separated working from broken was `--insecure`. `xerj brain` is the one command that works zero-config *with* auth, because it reads `<data_dir>/admin.key`. `autoindex` did not.

---

### Client
- **`ping()` fails fast and actionably** on 401/403 instead of discarding the status, and `embedding_execution_identity()` maps auth failures the same way rather than blaming the server.
- **`autoindex` gains the `admin.key` fallback `brain` already had** — loopback URLs only, and it announces which file it used.

### Server
- **`WWW-Authenticate: ApiKey realm="xerj"`** on 401 (RFC 7235 requires it; ES sends one). The scheme was verified against `extract_key` rather than assumed — the server also accepts `Bearer` and `Basic`, but advertising `Basic` would pop a browser password dialog for a credential with no username half.
- **The 401 body names the server's real data dir**, plus `XERJ_API_KEY`, `--api-key`, and `--insecure`. The ES envelope is byte-compatible: only `reason` changed.
- **`--insecure` help** now says it disables auth as well as TLS.
- **`brain` and `autoindex` completion banners** no longer print commands that 401, and reference `$XERJ_API_KEY` rather than echoing the secret.

### Docs
- **`install.html`** now bridges install → index a folder, with the `admin.key` export. Plus a macOS Gatekeeper note.
- **`quickstart.html`**: the "Drop a config next to it" dead end is gone. It leads with `--insecure` (its five `curl` steps are unauthenticated, so an auth-enabled config would have falsified them) and then shows the authenticated variants.
- **`cli.html`** claimed **"There are no subcommands"** — `xerj --help` has a `SUBCOMMANDS` block. `autoindex` and `--api-key` are now documented.
- **`security.html`** was wrong twice: it documented `Authorization: Bearer` where everything in the engine writes `ApiKey`, and a `.api-key` path that appears nowhere in the codebase (`grep -rn '\.api-key' engine/crates --include=*.rs` returns nothing; the real path is `<data_dir>/admin.key`). The example key was also the wrong shape — real keys are 64-char hex.
- **A documented `systemctl reload xerj` key-rotation procedure does not work**: there is no SIGHUP handler (only `ctrl_c`/`SIGTERM`), and the shipped unit has no `ExecReload`. Replaced with what the code actually implements.
- **`README.md`, `engine/README.md`, `xerj.default.toml`** — the auth-on path is documented alongside the `--insecure` one; `enabled = false` is marked as *not* the default.

---

## Verified by dogfooding this branch blind

An agent was given the built binary and the docs, forbidden from reading any Rust source or any internal notes, and told to follow the documentation literally and reproduce the reporter's journey.

| | result |
|---|---|
| **Ever had to disable auth?** | **No** — never passed `--insecure`, never touched `[auth]` |
| Time to first real search returning its own data | **under 5 minutes, 8 commands** |
| The tool's own `next:` hints and suggested `curl` | ran verbatim, worked |

### It also found three defects in this branch, fixed here

1. **The discovered key was printed verbatim in the completion banner.** The reader never typed that secret, and that banner is exactly what people paste into bug reports. Hints now render `$(cat <path>)`, which still pastes correctly.
2. **Discovery resolves `./data/admin.key` relative to the working directory**, so the same command succeeded in the tutorial directory and 401'd from `~/my-project` — the reporter's exact journey — with nothing explaining why. It is now announced on stderr and documented in the shipped config.
3. **The 401 listed `--insecure` as a co-equal third bullet** — one flag, no path to look up, by far the cheapest thing to try. A straight line back to "I turned auth off", the outcome the message exists to prevent. It is now ordered last and fenced as a last resort that disables auth *for every client*, not just the current command.

---

## A security defect in this change, caught in review before it shipped

The first version of `url_is_loopback` split on the last `:` to find `host:port`. That reads the **userinfo** in `http://localhost:9200@evil.com/` as a host, judges it loopback, and would have sent the local admin key to `evil.com` as `Authorization: ApiKey <key>`.

It now parses with `reqwest::Url` and reads `host_str()`. Schemeless input (`localhost:9200`) is retried *through* the parser rather than around it, so `localhost:9200@evil.com` still resolves to `evil.com`. Regression test included, and proven to fail against the old implementation:

```
http://localhost:9200@evil.com/ is NOT this machine - treating it as loopback leaks the admin key
```

---

| gate | result |
|---|---|
| `cargo test -p xerj-autoindex -p xerj-api -p xerj-server --lib` | **730 passed, 0 failed** |
| ES-YAML conformance | **1366 passed · 0 failed · 3 skipped** (200 files — the real suite, not the "No YAML test files found" path) |

Every behavioural test here was **proven to fail with its fix reverted**, and the failure output recorded. Two of them reproduce the original report exactly: reverting `ping`'s status check reproduces the swallowed 401, and reverting the identity mapping reproduces the reporter's sentence byte-for-byte.

## Assumed — NOT tested

- That no client depends on the previous `reason` string in the 401 body. The ES error *shape* is unchanged and asserted; only the human-readable text differs.

## Not run

- `cargo test -p xerj-engine --test painless_script_limits` aborts with a stack overflow. **Pre-existing and unrelated** — verified by reproducing it on clean `main`.
- macOS behaviour was not tested on a Mac. The Gatekeeper note is documentation only.

## Reviewer notes

- **The 401 body now discloses the server's data-dir path to an unauthenticated caller.** Judged acceptable — it is a path, `admin.key` is `0600`, and the default bind is loopback — but it is a deliberate trade and easy to revert to a placeholder if you disagree.
- Site search still cannot find `autoindex`: the `docs-index` blob is duplicated verbatim across ~30 pages with no entry for it. There is no site generator, so fixing it means editing ~30 files identically — better as its own mechanical change.
- Two pre-existing issues found and deliberately left alone: `xerj --help` renders every multi-line flag flush-left on `main` (Rust `\` continuation strips leading whitespace), and `clustering_key_tests` flakes under load because two tests share fixed `temp_dir()` paths.

### DIFF
--- a/README.md
+++ b/README.md
@@ -90,6 +90,17 @@ xerj --insecure --data-dir ./data &     # local dev: no TLS, no auth
 xerj autoindex ~/my-project
 ```
 
+If your server has auth on — which is the default for every start *without*
+`--insecure`, including any start from a config file — hand `autoindex` the same
+key. It never picks the key up from `xerj.toml`; pass `--api-key` or set
+`XERJ_API_KEY`, or every request comes back `401 Unauthorized`:
+
+```sh
+xerj --data-dir ./data &                          # auth on: key minted on first boot
+export XERJ_API_KEY="$(cat ./data/admin.key)"     # <data_dir>/admin.key
+xerj autoindex ~/my-project
+```
+
 That is the whole setup. There is no schema to write and no pipeline to
 configure. XERJ sniffs each file, works out what it is, and creates one index per
 dataset it finds:
--- a/engine/README.md
+++ b/engine/README.md
@@ -97,7 +97,19 @@ data_dir = "/var/lib/xerj"
 es_compat_port = 9200
 
 [auth]
-enabled = false   # set true and provide admin_api_key in production
+enabled = false   # NOT the default — the default is true
+```
+
+`auth.enabled` defaults to **true**; `enabled = false` above is an explicit
+opt-out, appropriate only for a trusted private network or local development.
+Leave it out (or set it to `true`) and the server mints an admin key on first
+boot, prints it, and writes it to `<data_dir>/admin.key`. Every client then has
+to send it — including `xerj autoindex`, which never reads the key out of
+`xerj.toml`; give it `--api-key` or `XERJ_API_KEY`:
+
+```bash
+export XERJ_API_KEY="$(cat /var/lib/xerj/admin.key)"
+xerj autoindex ~/my-project
 ```
 
 Pass a config file with `--config`:
--- a/engine/crates/xerj-api/src/auth.rs
+++ b/engine/crates/xerj-api/src/auth.rs
@@ -179,7 +179,7 @@ pub async fn auth_middleware(State(state): State<AppState>, req: Request, next: 
     if is_authorized(&state, auth_header) {
         next.run(req).await
     } else {
-        unauthorized_response()
+        unauthorized_response(&state.config.server.data_dir)
     }
 }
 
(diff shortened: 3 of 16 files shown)


xerj-org:
Review notes from a parallel session that was independently tasked with this same bug report and has stood down — this PR reaches the same root cause and goes further on the server side, so the duplicate work is being dropped rather than competing. Four findings from that effort, all re-verified against this diff and the tree before posting.

**1. `discover_local_admin_key()` will hand the admin key to any process listening on a loopback port.** The candidate list is walked in fixed order and the first non-empty file is returned, with nothing associating that key with the server actually being addressed, and no shape check on the file. `url_is_loopback` then gates only on the *host*, not the port. So on a shared machine, an unprivileged local user who binds an unused loopback port and persuades (or waits for) a run pointed at it — `xerj autoindex --url http://localhost:9999` — receives the admin key of a server they have no access to.

The blast radius is narrow (it needs the victim to target the wrong port) but the failure mode is credential disclosure, which is worth closing before merge rather than after. Two cheap mitigations, either of which breaks the chain: resolve the target port against a locally readable `xerj.toml` (`server.es_compat_port` / `rest_port`) and only offer the key to a port that config actually declares; and shape-check the file against what `ensure_admin_key` writes (64-char lowercase hex) so an arbitrary file named `admin.key` is not put on the wire.

**2. `url_is_loopback` counts `0.0.0.0` as loopback.** Verified in the diff. Being precise about severity: on Linux a *connection* to `0.0.0.0` lands on localhost, so this is not a remote-exfiltration hole. It is still a wildcard bind address rather than a loopback destination, and given the function's own comment says a false positive here "is a credential leak, not a cosmetic bug", widening the trusted set for no gain seems against the intent.

**3. Discovery ignores the `data_dir` the docs tell users to set.** The PR body frames not reading `xerj.toml` as a design choice, but `install.html` and `config.html` both recommend an absolute `data_dir` such as `/var/lib/xerj` or `/srv/…`. Anything outside the five hardcoded paths still dead-ends in exactly the way this PR is fixing. Reading `server.data_dir` through `xerj_common::config::Config::load` also composes with mitigation (1) — the same file that tells you where the key is tells you which port it belongs to.

**4. A doc falsehood in the same first-run class that this PR does not touch.** `landing/docs/config.html:124` states the precedence is `--config` → `/etc/xerj/xerj.toml` → `./xerj.toml`. Verified against the code: `load_config` (`engine/crates/xerj-server/src/main.rs:420-431`) consults `--config`, then `XERJ_CONFIG`, then `Config::default()`. There is no `/etc/xerj` or `./xerj.toml` discovery anywhere in `engine/crates` — the only occurrence of that path is a doc-comment example in `xerj-common/src/config.rs:22`. So a user who drops `xerj.toml` next to the binary and runs bare `xerj` silently gets defaults, which is the same shape of failure as the 401: the documented path does not work and the software does not say so. `XERJ_CONFIG`, which does work, is not documented on that page at all.

**5. Nothing in CI will catch the next regression of this path.** `gh pr diff 333 --name-only` touches no `.github/` file; the proof here is in-process tests plus a manual dogfooding run. That is the class of test this repo lacks and the reason a user found this before CI did. The stood-down branch has a 230-line `documented-first-run-smoke.sh` written against the *observable contract* (so it applies to this PR's behaviour unchanged): it writes the quickstart config, boots the server, **asserts the server genuinely refuses unauthenticated requests so the gate cannot pass vacuously**, runs `autoindex` with no key in the environment and no flag, asserts the documents are searchable, then re-asserts that wrong and absent keys are still refused and the refusal is actionable. It passed end to end. I will offer it as a small follow-up PR against this branch's behaviour rather than bolting it on here — say the word if you would rather it came in with this one.

Nothing above blocks the substance of this PR, which is a better fix than the one being dropped — in particular `url_is_loopback` going through a real URL parser and catching the `localhost:9200@evil.com` userinfo trap is more robust than the hand-rolled split it is replacing.

PR #335, opened 2026-08-12
feat(help): ask for bug reports and UX feedback, on by default

buger:

# Ask for bug reports and UX feedback, from the surface agents actually read

XERJ's help is read far more often by coding agents than by people. An agent that hits a confusing flow has no obvious way to tell anyone — and the flows worth hearing about are exactly the ones nobody files, because "it worked, it was just confusing" doesn't feel like a bug.

This adds three lines near the top of every help screen:

```
xerj v1.0.0-rc.16 — the unified search engine for AI (Elasticsearch wire-compatible)

Hit a bug, or a flow that confused you? Both are worth reporting. If you have a
GitHub tool, file it yourself at https://github.com/xerj-org/xerj/issues — or open
a Discussion if you are unsure. Keep secrets, API keys and private data out of it.

USAGE:
```

On by default. `--disable-feedback` turns it off.

## Placement is the requirement, not decoration

It sits immediately before `USAGE:` — **line 3** of `xerj`, `xerj index` and `xerj autoindex`; line 8 of `xerj brain`, after its description paragraph.

Agents truncate long output **from the bottom**, so an invitation at the end of a 200-line help screen is an invitation nobody reads. The tests assert the **line number**, not merely presence, so a later refactor cannot quietly drift it down the page. That assertion is real: a mutation moving the block to the bottom passes a presence-only check and fails the position one with `invitation sits on line 227`.

`xerj index --help` is included because the top-level help advertises it; it would otherwise have been the one screen that never asks.

## The off-switch

`--disable-feedback`, or `XERJ_DISABLE_FEEDBACK` using the same truthiness convention as `XERJ_ALLOW_INSECURE_NETWORK_BIND` (`1|true|yes|on` / `0|false|no|off`).

**The flag is position-independent**, and this is not incidental: `--help` is dispatched *inside* each argument loop and exits there, so a naive implementation would silently ignore `xerj --help --disable-feedback`. It scans argv instead. Both orders and the env var produce byte-identical output, verified with `diff`.

An unparseable env value warns on **stderr** and keeps the invitation rather than aborting — refusing to print `--help` because of a typo in an environment variable would be a worse failure than one extra line.

## No config key, deliberately

`--help` prints before any config is loaded, and `load_config` reads a file *only* when `--config`/`XERJ_CONFIG` names one — there is no default path search, so with no config file a TOML key would have nothing to read from. `autoindex` and `brain` never load the server config at all, so the key would work on at most one of four surfaces.

Making it genuinely work would mean parsing TOML on the `--help` path in three binaries, letting a malformed config break the one command that must always work. A setting that silently does nothing is worse than no setting, and the gap it would fill — "I cannot change argv" — is already covered by the environment variable.

| gate | result |
|---|---|
| `cargo test -p xerj-common -p xerj-autoindex -p xerj-server --lib` | **620 passed, 0 failed** |

Real help output for all four surfaces, with and without the flag, was captured from the built binary — not asserted from source.

**15 tests, each observed failing under a targeted mutation.** Two worth calling out:

- Moving the block to the bottom passes presence-only assertions and fails the position ones.
- The `{{`/`}}` escaping mutation reproduces the build error that has broken this exact file before: `invalid format string: … you can escape it using {{`. These help bodies are single string literals, so an embedded JSON example needs `\"` as well — there is a test asserting any JSON in the help still parses.

## Not run

- `cargo test -p xerj-engine --test painless_script_limits` aborts with a stack overflow. **Pre-existing and unrelated** — reproduces on clean `main`.

## Reviewer notes

- The message is three lines on every `--help`. That is a deliberate cost: it appears often, so it is kept short and the off-switch is documented in each OPTIONS list rather than inside the block.
- If you would rather it not appear for interactive humans at all, the cleanest variant is to gate it on `!stdout.is_terminal()` — that would show it to agents and pipes only. Not done here because it makes the behaviour harder to explain and to test, but it is a one-line change if you prefer it.

### DIFF
--- a/engine/Cargo.lock
+++ b/engine/Cargo.lock
@@ -6933,6 +6933,7 @@ dependencies = [
  "reqwest",
  "serde_json",
  "tokio",
+ "xerj-common",
 ]
 
 [[package]]
--- a/engine/crates/xerj-autoindex/src/cli.rs
+++ b/engine/crates/xerj-autoindex/src/cli.rs
@@ -109,9 +109,16 @@ pub fn print_help() {
 /// The help text as a value, so tests can assert that a documented flag is
 /// still documented instead of trusting a `println!`.
 pub fn help_text() -> String {
+    help_text_with(xerj_common::feedback::enabled())
+}
+
+/// [`help_text`] with the feedback invitation forced on or off, so a test does
+/// not have to mutate the process environment to see both shapes.
+pub fn help_text_with(feedback: bool) -> String {
     format!(
         "xerj autoindex — point it at any folder and make the contents AI-searchable, zero config\n\
          \n\
+         {feedback_block}\
          USAGE:\n\
              xerj autoindex <folder> [OPTIONS]     discover + index a folder\n\
              xerj autoindex map [OPTIONS]          print the discovered data map\n\
@@ -178,6 +185,9 @@ pub fn help_text() -> String {
                                   exactly like an agent-driven run. See ESTIMATE + DECISION\n\
                                   GATE.\n\
              --dataset <SLUG>     (map) show a single dataset\n\
+             --disable-feedback   do not print the feedback invitation above; honoured in\n\
+                                  any position, including after --help (env\n\
+                                  XERJ_DISABLE_FEEDBACK=true)\n\
              --help, -h           this help\n\
          \n\
          IGNORE RULES:\n\
@@ -334,6 +344,7 @@ pub fn help_text() -> String {
                        nothing was indexed; a JSON decision request is on stdout;\n\
                      2 usage; 1 endpoint/journal failure, a refused corpus removal, or a\n\
                      refused unsafe state transition\n",
+        feedback_block = xerj_common::feedback::block(feedback),
         fresh_help = FRESH_HELP,
         resume_policy_help = RESUME_POLICY_HELP,
         defaults = crate::ignore_rules::DEFAULT_IGNORE_PATTERNS.join(" ")
@@ -552,6 +563,9 @@ pub fn parse(args: Vec<String>) -> Result<Cmd, String> {
             }
             "--quiet" => quiet = true,
             "--dataset" => dataset = it.next(),
+            // Read out of band by `xerj_common::feedback`, which scans the
+            // whole argument list; accepted here so it is not "unknown".
+            xerj_common::feedback::DISABLE_FLAG => {}
             "--help" | "-h" => return Ok(Cmd::Help),
             "map" if sub.is_none() && folder.is_none() => sub = Some("map".into()),
             "status" if sub.is_none() && folder.is_none() => sub = Some("status".into()),
@@ -1152,6 +1166,60 @@ mod tests {
             "prompt_not_offered_because",
         ] {
             assert!(help.contains(expected), "help is missing {expected:?}");
+        }
+    }
+
+    /// The invitation is on by default and near the top — at the bottom of a
+    /// 200-line help body it would never be read. Position is the requirement,
+    /// so position is what is asserted.
+    #[test]
+    fn the_help_invites_a_bug_report_near_the_top_by_default() {
+        let help = super::help_text_with(true);
+        let line = help
+            .lines()
+            .position(|l| l.contains("Hit a bug, or a flow that confused you?"))
+            .expect("no invitation in the help");
+        assert!(line < 10, "invitation sits on line {}", line + 1);
+        for expected in [
+            "https://github.com/xerj-org/xerj/issues",
+            "GitHub tool",
+            "Discussion",
+            "secrets, API keys and private data",
+            // The off-switch belongs on the same screen as the thing it turns
+            // off.
+            "--disable-feedback",
+        ] {
+            assert!(help.contains(expected), "help is missing {expected:?}");
+        }
+    }
+
+    #[test]
+    fn the_invitation_is_gone_when_it_is_turned_off() {
+        let help = super::help_text_with(false);
+        assert!(!help.contains("Hit a bug"), "invitation was not silenced");
+        assert!(
+            !help.contains("\n\n\n"),
+            "silencing it left a blank gap:\n{help}"
+        );
+        // Silencing the invitation must not silence the rest of the help.
+        assert!(help.contains("USAGE:") && help.contains("--disable-feedback"));
+    }
+
+    /// `--disable-feedback` is consumed wherever it appears, on every
+    /// subcommand — it must never be an "unknown argument".
+    #[test]
+    fn disable_feedback_is_accepted_in_any_position() {
+        for args in [
+            vec!["data", "--disable-feedback"],
+            vec!["--disable-feedback", "data"],
+            vec!["map", "--disable-feedback"],
+            vec!["--disable-feedback", "map"],
+            vec!["status", "--disable-feedback"],
+            vec!["--disable-feedback", "--help"],
+            vec!["--help", "--disable-feedback"],
+        ] {
+            parse(args.iter().map(|s| s.to_string()).collect())
+                .unwrap_or_else(|e| panic!("{args:?} was rejected: {e}"));
         }
     }
 
--- a/engine/crates/xerj-common/src/feedback.rs
+++ b/engine/crates/xerj-common/src/feedback.rs
@@ -1,0 +1,189 @@
+//! The feedback invitation printed near the top of every `--help` surface.
+//!
+//! Most of the people — and most of the *agents* — who hit a rough edge in
+//! xerj never tell anyone: they work around it locally and the next caller
+//! pays the same cost from scratch (AGENTS.md, "If XERJ broke something in
+//! your build, send it back"). `--help` is the one screen every caller
+//! actually reads, so the invitation lives there, near the top, in three
+//! lines.
+//!
+//! One string, one on/off decision, shared by `xerj --help`,
+//! `xerj index --help`, `xerj autoindex --help` and `xerj brain --help`, so
+//! the four surfaces cannot drift apart.
+//!
+//! Turning it off: [`DISABLE_FLAG`] anywhere in the argument list, or
+//! [`DISABLE_ENV`] in the environment.
+
+/// Where a defect goes. Verified against the README badges and clone URL.
+pub const ISSUES_URL: &str = "https://github.com/xerj-org/xerj/issues";
+
+/// Where an open-ended "is this even a bug?" goes instead.
+pub const DISCUSSIONS_URL: &str = "https://github.com/xerj-org/xerj/discussions";
+
+/// The off-switch. Recognised in any position (see [`enabled_from`]).
+pub const DISABLE_FLAG: &str = "--disable-feedback";
+
+/// The off-switch for callers that cannot change argv — CI, containers, an
+/// agent harness that owns the command line.
+pub const DISABLE_ENV: &str = "XERJ_DISABLE_FEEDBACK";
+
+/// The invitation itself, ending in a blank line so a caller can splice it
+/// straight in front of the next help section.
+///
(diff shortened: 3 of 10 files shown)


xerj-org:
Conflicts against current `main` are resolved, but **I cannot push to this PR's branch.** `probelabs/xerj` is an organization-owned fork, and GitHub does not honour maintainer edits on those even though `maintainerCanModify` reports `true`. Every credential available here is refused — `GH_TOKEN`/`GITHUB_TOKEN` (both resolve to `xerj-org`), and three SSH keys. Not a scope problem; merging your PRs works, pushing to them doesn't.

**The rebased result is on `xerj-org/xerj` at branch `pr335-conflict-resolution` (`5b7b055d`)** — exactly `origin/main` plus this PR's 2 commits, conflict-free. To update #335 in place and re-run CI, from an account with write on the fork:

```sh
git fetch origin pr335-conflict-resolution
git push --force git@github.com:probelabs/xerj.git \
  pr335-conflict-resolution:feat/feedback-invite
```

### One thing you should know before merging: a clean merge that would not compile

`main.rs:1714`, the test `insecure_help_documents_auth_not_just_tls` added by **#333**, landed in an auto-merged region calling `help_text()` with **no arguments** — while this PR changes the signature to `help_text(feedback: bool)`. Git merged it without complaint because the two sides touched different lines; it is a hard compile error in the test profile that would only have surfaced in CI. Fixed to `help_text(true)` (the shipped default, invitation on), which incidentally also proves the invitation doesn't reflow the `--insecure` paragraph it sits above.

That is the whole hazard of merge-order interactions: **#333 and this PR independently extracted the same `help_text()` refactor with different signatures.** Both were green alone.

### The two real conflicts, both from #333

1. **`print_help` / `help_text` signature** (`main.rs:183-191`) — took this PR's `fn help_text(feedback: bool) -> String`, since the caller passes `xerj_common::feedback::enabled()`, and **merged both doc comments** rather than discarding main's rationale about asserting documented behaviour.
2. **`format!` arguments** (`main.rs:~308`) — took this PR's two-argument form adding `feedback::block(feedback)`; the format string body auto-merged and does carry the `{}` placeholder.

### Cleanly-merged regions audited rather than trusted

- #333's reworded `--insecure` help survives intact (`main.rs:219-223`) — still documents that the flag disables TLS *and* auth, and still points at `<data_dir>/admin.key` and `XERJ_API_KEY`. The invitation splices in above `USAGE:`, clear of it.
- #337's `xerj mcp` SUBCOMMANDS block (`main.rs:268-271`) and its dispatch (`main.rs:1629`) are untouched.
- This PR never touches completion banners, so #333's corrected ones — the whole point of which was that they stopped printing commands that 401 — are unaffected in both `xerj-autoindex/src/cli.rs` and `xerj-server/src/brain.rs`.

`cargo fmt -p xerj-common -p xerj-server -p xerj-autoindex -- --check` exits 0.

### Follow-up, deliberately not done here

#337 added a **fifth** help surface, `xerj mcp --help`, after this PR was written. By this PR's own stated rule — `xerj index --help` is included because the top-level help advertises it, so it would otherwise be the one screen that never asks — `mcp` now qualifies and doesn't carry the invitation. Adding it needs a new `xerj-common` dependency in `xerj-mcp` plus an extra entry in the test's `SURFACES` array, which I didn't attempt without a build. Related minor wart: `xerj mcp --disable-feedback` currently errors as an unknown argument.

buger:
Conflicts resolved and pushed — the branch is now `MERGEABLE` and carries the tree-sitter fix, so it builds. I could push here because this runs with a token that has write on `probelabs/xerj`; thank you for staging `pr335-conflict-resolution`, and apologies that the fork ownership made it unpushable from your side.

Your resolution was better than mine on the point where we differed, and I've taken yours.

**`help_text(true)`, not `false`.** I had independently hit the same `insecure_help_documents_auth_not_just_tls` landmine and resolved it with `false`, reasoning that the test should be about the flag's paragraph alone. Your argument is stronger: asserting against the shipped default *also* proves the invitation — which splices in above `USAGE:` — does not reflow the paragraph the test reads. Yours tests more, so that's what's on the branch, credited in the commit.

That landmine is worth restating for anyone reading later, because it is nastier than a normal conflict: **#333 and this PR independently extracted the same `help_text()` refactor with different signatures**, git auto-merged the regions cleanly because they touched different lines, and the result does not compile. Worse, `cargo test --lib` still passes — the call site lives in a `#[cfg(test)]` block inside the *binary* target, so only `cargo clippy --workspace --all-targets` (or CI) compiles it. A branch verified with the lib suite alone would have shipped a red tree.

**I also closed the follow-up you flagged: `xerj mcp --help` now carries the invitation.** You were right that it qualifies under this PR's own rule, and that `xerj mcp --disable-feedback` was being rejected as an unknown argument on the one surface the flag is meant to control. Doing it needed a build, which is what I had:

- `xerj-mcp` takes the shared `xerj-common` block, so the five surfaces cannot drift apart.
- `HELP` became `HELP_BODY` plus `help_text(feedback)`, splicing the invitation above `USAGE:` in the same position as the other four.
- The off-switch is documented in mcp's own OPTIONS list — the test requires the way out to be discoverable from the same screen, and it failed until that was added.
- `SURFACES` grows to five, so a sixth surface added later fails this test rather than silently going unasked.

**One thing you should know that is bigger than this PR: `main` does not build.** `xerj-engine` pins `tree-sitter = "0.25"` and `xerj-autoindex` `"0.26"`, and Cargo permits only one crate to `links` a native library:

```
```

Same shape of failure as the `help_text` one: #327 brought `code_blocks.rs` verified against 0.25, #317 moved autoindex to 0.26 for the Kotlin/Swift/Scala/Dart grammars, both merged green, and together they do not compile. #351 is the one-line fix — safe because both crates already pin identical grammar versions (all `0.23`), so only the core binding moves and node-kind names come from the grammars. It is cherry-picked onto this branch so #335 can build; merging #351 first and dropping that commit here is equally fine.

**Verified on this branch:** build ✓ · `xerj-common`/`xerj-server`/`xerj-mcp` libs 115 passed, 0 failed · `xerj-server` bins 33 passed · `feedback_invite` 5 passed across all five surfaces · `cargo fmt --all --check` exit 0 · `cargo clippy --workspace --all-targets -D warnings` exit 0.

Real output, `xerj mcp --help`:

```
xerj mcp — Model Context Protocol (MCP) stdio server for XERJ

Hit a bug, or a flow that confused you? Both are worth reporting — file it at https://github.com/xerj-org/xerj/issues with a GitHub tool or the gh CLI, or open a Discussion if you are unsure. Agents: say the report was filed automatically on behalf of a human. Keep secrets, API keys and private data out of it.

USAGE:
```

xerj-org:
Merged as `c4942f01`. Reviewed the whole diff and drove the built binary rather than reading the source and believing it — two things were worth checking hard on an on-by-default change, and both hold.

### 1. What it prints, and where it cannot print

Built `--bin xerj` from this branch merged onto `d8be09cf`, then ran every surface with **stdout piped, not a tty**:

| command | invitation line | total help lines |
|---|---:|---:|
| `xerj --help` | 3 | 133 |
| `xerj index --help` | 3 | 23 |
| `xerj autoindex --help` | 3 | 233 |
| `xerj brain --help` | 8 | 33 |
| `xerj mcp --help` | 3 | 49 |

Exactly the placement the body claims, including line 8 on `brain` behind its description paragraph. Nothing changes between a tty and a pipe — there is no `is_terminal` branch, so the output is deterministic, which is the right call for something a test asserts a line number on.

Where it does **not** appear, checked by running each one:

- `xerj --version` → 0 occurrences
- top-level unknown-argument path (`xerj --nonsense` → `unknown argument: --nonsense. Use --help for usage.` on stderr, exit 1) → 0
- server boot and every non-help code path — `feedback::block`/`feedback::enabled` have 16 call sites and all 16 are inside a `help_text`/`print_help` function (`main.rs:190,330,1148,1175`, `brain.rs:62,100`, `cli.rs:112,347`, `xerj-mcp/src/lib.rs:144,161`), so there is no path that prints it during an indexing run or a long-lived process.

The one extra place it *does* show is `xerj autoindex`'s usage-error path, which dumps the full help — e.g. `xerj autoindex map --data-dir X` prints `error: unknown argument: --data-dir` and then the help with the invitation. That is the right moment to ask, and `--disable-feedback` silences it there too (verified).

So the CI/spam question is: one wrapped line plus a blank, once, only when a caller explicitly asked for help. That is not spam.

### 2. Nothing leaves the machine

`strace -f -e trace=socket,connect,sendto,sendmsg` on `xerj --help`: **0 `AF_INET` sockets, 0 connects** (261 traced lines, all `AF_UNIX`/`AF_NETLINK`/cgroup). `openat` shows only `/proc/self/maps`, `/proc/self/cgroup` and `cpu.max` — the pre-existing resource probe. There is no reporting client here at all; it prints a URL and asks the caller to go there. Opt-in in the strongest sense, since the caller has to act.

Also confirmed the destination is real rather than decorative: `hasDiscussionsEnabled: true` on this repo with Announcements/General/Ideas/Polls/Q&A, so "open a Discussion if you are unsure" is actionable advice and not a dead end.

### 3. The off-switch, and the silent-ignore check

For all five surfaces, three routes — `... --help --disable-feedback`, `... --disable-feedback --help`, and `XERJ_DISABLE_FEEDBACK=true` — produce **byte-identical** output with zero occurrences of the invitation and no `\n\n\n` hole where it stood. The position-independence matters exactly as the body says, because `--help` exits from inside each argument loop; scanning argv in `enabled_from` is the right shape.

Env truthiness: `1|true|yes|on` disable; `0|false|off|""` keep it on; `maybe` keeps it on **and** writes `XERJ_DISABLE_FEEDBACK="maybe" is not a boolean; use true or false` to stderr with stdout untouched. That is the part I was looking hardest at — an unreadable setting that is silently dropped is this repo's #204 class, and this one is neither dropped nor allowed to break the one command that must always work.

No surface accepts `--disable-feedback` and then prints anyway, and no surface rejects it as unknown — `xerj mcp --disable-feedback` in particular, which was the gap you closed on top of the maintainer's conflict resolution, is now accepted (`xerj-mcp/src/lib.rs:168`) on the one screen the flag exists to control.

### Gates, on this branch merged onto current `main`

| gate | literal result |
|---|---|
| `cargo clippy --release -p xerj-common -p xerj-mcp -p xerj-autoindex -p xerj-server --all-targets -- -D warnings` | exit 0, **0 warnings** |
| `cargo test --release -p xerj-common -p xerj-mcp -p xerj-autoindex --lib` | **579 + 90 + 25 = 694 passed, 0 failed** |
| `cargo test --release -p xerj-server --bins --test feedback_invite --test sigpipe` | **33 + 5 + 2 = 40 passed, 0 failed** |
| PR checks at `424cd6ab` | **16/16 SUCCESS**, `mergeStateStatus: CLEAN` |

`cargo clippy --all-targets` is the gate that matters here, since the `help_text(true)` landmine from #333 lives in a `#[cfg(test)]` block inside the *binary* target and `--lib` alone cannot see it. It compiles and passes.

The tree-sitter alignment commit (`37b60155`) is now redundant — #350 landed the same 0.26 pin on `main` — but it merges to the identical result (`xerj-engine` and `xerj-autoindex` both at `0.26` post-merge), so there was nothing to strip.

### Two nits, neither worth holding the PR for

1. `feedback.rs:21` declares `pub const DISCUSSIONS_URL` and nothing reads it — the invitation says "open a Discussion" in prose without carrying the link. It is `pub`, so no dead-code warning will ever remind anyone. Either splice it in or drop the const next time this file is touched.
2. All five help screens render continuation lines flush-left (`--port <PORT> ...` and its follow-on lines both start at column 0), because Rust's `\`-continuation strips leading whitespace. That is pre-existing house style, not something this PR introduced — flagging it only so nobody blames the invitation for the ragged block it sits in.

Thanks for the write-up on the `help_text()` merge landmine and the `links = "tree-sitter"` conflict — both were more useful than the diff itself.

PR #339, opened 2026-08-12
build: add a `quick` profile so verification stops paying shipping-grade LTO

xerj-org:
`[profile.release]` is `lto = "fat"` + `codegen-units = 1` — deliberately the slowest build Rust can produce, because it yields the fastest binary. Correct for an artifact users run for months; wrong for "does this fix compile?".

Measured on this workspace: three concurrent release builds drew **~1.16 cores each** on a 32-core machine, because fat LTO's whole-program pass is largely single-threaded and `codegen-units = 1` removes intra-crate parallelism. One agent's build took **196 minutes**. A cold build of the same target under the new profile reached **19 concurrent rustc processes at ~21 cores** within 45 seconds.

`[profile.quick]` inherits release and relaxes only optimisation. Everything affecting correctness is held identical, including `panic = "abort"` (inherited deliberately — it changes observable behaviour, so a profile differing there would not be a valid functional test).

`[profile.release]` is deliberately untouched. Released artifacts, benchmarks and any published number keep fat LTO; a quick-profile binary is materially slower, so quoting a performance figure from one would be a false claim.

`docs/FAST_BUILDS.md` documents the other levers with their blast radius: sccache, the `wild` linker already installed and unused, system zstd/sqlite instead of rebuilding vendored C, the nightly parallel frontend, and why `-j 32` is right for one build and wrong for three. **No wall-clock speedup is claimed** — the machine was saturated throughout, and timing a build on a contended box measures the contention.

### DIFF
--- a/docs/FAST_BUILDS.md
+++ b/docs/FAST_BUILDS.md
@@ -1,0 +1,143 @@
+# Building XERJ faster, and using less of the machine
+
+Written 2026-08-12 after three agents cold-building this workspace in parallel
+drove a 32-core box to load 92 and still hadn't produced a binary three hours
+later. Every lever below is ordered by measured or expected impact. Numbers
+marked ESTIMATED are not yet measured on this workspace — do not quote them
+anywhere public until they are (see *Measuring* at the end).
+
+## The one that matters: stop paying shipping-grade optimisation to verify a fix
+
+`[profile.release]` is `lto = "fat"` + `codegen-units = 1`. That is deliberately
+the slowest build Rust can produce, because it yields the fastest binary. Fat
+LTO runs largely **single-threaded** at link time and `codegen-units = 1`
+removes intra-crate parallelism — which is why you see one `rustc` pinned at
+100% for twenty minutes on a 32-core machine while nothing else progresses.
+
+Correct for an artifact users download. Wrong for "does my one-line fix
+compile?", where the binary's runtime speed is irrelevant to the question.
+
+Use the `quick` profile (defined in `engine/Cargo.toml`):
+
+```sh
+cargo build --profile quick -j 12 -p xerj-server
+cargo test  --profile quick -j 12 -p xerj-autoindex
+```
+
+It keeps everything affecting **correctness** identical to release —
+`panic = "abort"` (inherited; it changes observable behaviour so it must match),
+same features, same `target-cpu` — and relaxes only optimisation.
+
+- **Valid for:** does it compile, do tests pass, is ES-YAML still 0 failed, does
+  the fix work.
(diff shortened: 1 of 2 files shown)


xerj-org:
Overlap disclosure, and a correction to this PR that #331 exposes.

**#331 landed this ground first and reasons about it better.** It was opened at 13:50 today, hours before this branch, and adds `[profile.ci-test]` (`lto = false`, `codegen-units = 16`, `panic = "unwind"`) with a measured case: CI run 31571998506 spent **76m05s compiling for 1m52s of test execution**. I did not check open PRs before starting, which is my mistake — this PR should be read as a follow-on to #331, not an independent proposal, and #331 should merge first.

**The correction matters and is not cosmetic.** `[profile.quick]` here inherits `panic = "abort"` from release, and I justified that as necessary because panic behaviour is observable and a profile that differed would not be a valid functional test. #331 points out the consequence I missed: `panic = "abort"` forces cargo to rebuild the **entire dependency graph** with `panic = "unwind"` for test targets, so `cargo test --profile quick` would share almost nothing with a preceding `cargo build --profile quick`. For the test half of the verify loop, #331's `panic = "unwind"` is the better choice, and its comment makes that split explicit rather than accidental. My reasoning was right about correctness and wrong about cost.

**What is left here that #331 does not cover**, and what I suggest keeping:

- A profile for the **build** half of the loop, distinct from the CI test profile — an agent asking "does my fix compile" is not running `cargo test`.
- `docs/FAST_BUILDS.md`, which is the larger part of this PR: `sccache` (uninstalled, and the direct fix for N git worktrees each cold-building the same 666 dependency packages), the `wild` linker **already installed at `~/.cargo/bin/wild` and unused**, `ZSTD_SYS_USE_PKG_CONFIG` / `LIBSQLITE3_SYS_USE_PKG_CONFIG` to stop rebuilding C the system already provides at zstd 1.5.7 and sqlite3 3.46.1, the nightly parallel frontend for long-pole crates, and why `-j 32` is right for one build and wrong for three.

Reviewer's call on how to resolve the duplication: either fold the build-profile stanza into #331 so there is one place profiles are defined and keep only the doc here, or land #331 and rebase this on top with the panic reasoning corrected. Either is fine by me; two near-identical profiles in one manifest is not.

**Supporting measurement from this box**, complementing #331's CI numbers: three concurrent `--release` builds drew ~1.16 cores *each* on a 32-core machine, one agent's build took **196 minutes**, and a cold build under the relaxed profile reached **19 concurrent rustc processes at ~21 cores within 45 seconds**. The parallelism figure is robust; I am still not quoting a wall-clock speedup, because every timing here was taken on a contended box.

PR #340, opened 2026-08-12
feat(search): tell the caller what their query actually saved them

buger:

# Tell the caller what their query actually saved them

XERJ's pitch is that it returns answers, not payloads. When a request genuinely withholds bytes, say so — with a number the reader can check, and a sentence naming the mechanism that did it.

```json
"_savings": {"bytes": 309596, "measured": "sampled",
             "note": "_source narrowed to the 1 field you listed"}
```

On by default. Silent when nothing was saved. **87–101 bytes** on the wire. `?savings=exact` for a byte-for-byte count, `?savings=false` to turn it off.

Five mechanisms produce five different sentences — projection by includes, projection by excludes, `_source: false`, highlight-instead-of-documents, and fields re-materialised — because the point is to teach the reader which lever they pulled, not to print a counter.

---

## The baseline is the whole feature

**The baseline is the response the caller would otherwise have received**: same query, same page, no `_source` clause.

That sounds obvious. It was not what shipped in the first revision, and the difference is the reason this PR exists in its current form.

The first revision measured against a response containing raw embedding float arrays. Those bytes genuinely were not serialized — the measurement was accurate — but the counterfactual is one **nobody would ever request**, and one this engine has not returned by default since #309. A blind dogfood over 35 queries caught it:

| | first revision | now |
|---|---|---|
| 737-byte response, range+sort | claimed **8,360,064** saved | claims **901,693** |
| `size:20`, no projection at all | claimed **22,298,691** | **emits nothing** |
| ratio claimed / actual | **9.1–9.4×** across every shape | **1.00×** |

The second row is the important one: it fired on queries where the caller projected nothing, taking credit for the engine's own default. A reader sees `bytes` and reads *"what I would otherwise have received"* — and that reading was false.

Independently verified after the fix, against a live server:

```
claimed saving : 309,596
projected resp :     533
baseline resp  : 310,159   (same query, no _source clause)
actual saving  : 309,626   ->  ratio 1.000
```

A pleasant side effect: the corrected baseline sums the *non*-companion members, so the vectors are never serialized. Fixing the honesty problem deleted the expensive half of the measurement.

## A floor, so it stays quiet when it is not worth the words

The block renders itself, measures itself, and is dropped unless the saving is **at least 10× its own size**. Self-referential, so it stays true if the wording changes.

The same dogfood found the old version spending **135 bytes to announce a 19-byte saving** — net −116 — and occupying 33% of a small response. Emitted blocks now clear the floor by 23–3000×; the case above reports a 309 KB saving from a 101-byte block.

The block also shed `tokens_approx` (it was exactly `bytes / 4`, 30 times out of 30) and `bytes_per_token` (always 4). A third of its cost was information the reader could derive.

## Estimation, and a sampler that was nearly wrong

Exact mode serializes withheld values into a counting sink, using serde itself so the count cannot drift from the wire format.

The default samples — and the obvious sampler over-reported by **+11%**. XERJ's embedder emits **sparse** vectors: 320 of 384 elements are exactly `0.0` (3 bytes), the rest are 18–21. No sampling scheme survives that bimodality; head, stride, prime and random at N=32/64/128 ranged from −9.7% to +16.8%.

The fix counts elements whose width is free to know (zeros, integers, bools, null) exactly, and samples only the remainder. Measured error: **0.00%** on the real companion, −0.28%/−0.66% on dense vectors, **−1.50% worst case** on a deliberately adversarial mixed-magnitude array — always under-claiming. Every figure is asserted against a budget, so a regression fails the build. `"measured":"sampled"` rides in the payload; its absence means exact.

Cost on a 10-hit page: **+59–69 µs** for a typical 18 KB document.

## What it deliberately does not claim

- **Aggregation-only queries report nothing.** At `size: 0` the caller asked for zero documents and received zero. Claiming the documents they declined would mean picking a hypothetical alternative and reporting it as measurement. It was the best sentence in the feature and it could not be shipped honestly.
- `_source: false` claims are dropped when `fields` / `docvalue_fields` / `stored_fields` re-materialise the values.
- Highlight is netted, not gross: if fragments cost more than the document, nothing is emitted.
- `PayloadSavings::new` returns `None` on zero — a zero-valued block is unrepresentable, so "0 tokens saved!" cannot occur.

## Bugs found by dogfooding, not review

- **The query cache leaked the block onto stock, opted-out responses.** `SearchRequest.savings` is `serde(skip)`, so it never reached the cache-key hash. Fixed by mixing the mode into the key, with a regression test that warms the cache first.
- A note reading *"Sent the 0 fields you asked for"* when the caller asked for one the document lacked.
- A `not all M` denominator that wandered 8 → 11 → 12 → 15 on the **same query** at different `size`, because it described the widest document on the page. Removed.
- `_source: false` combined with `fields` — the strictly better query — made the block vanish entirely. Now netted.

| gate | result |
|---|---|
| `cargo test` (engine, api, query, server — lib) | **896 passed, 0 failed** |
| `savings_stat_is_honest` end-to-end suite | **12 passed, 0 failed** |
| ES-YAML conformance | **1366 passed · 0 failed · 3 skipped** (200 files), matching baseline exactly |

Six falsification mutations — feature reverted, zero-guard removed, gross counting, estimator biased +5%, baseline reverted (over-claim visible: 15,372 vs 4,887), floor removed (prints the exact 19-byte block). Every test in the suite fails under at least one.

## Not run / flagged rather than hidden

- **Conformance no longer exercises the key.** The YAML fixtures are tiny (longest line 141 bytes), so nothing clears the floor. The green run proves no regression; the compatibility evidence is the earlier revision's run, where the key *was* emitted on `10_source_filtering.yml`'s shape and the suite still passed 1369/0.
- **A separate pre-existing engine finding.** On a companion index, `_source: {"excludes":[...]}` makes responses **10.5× larger** (995 → 10,441 bytes), because an explicit filter re-admits the vectors. This change does not cause it and now correctly says nothing for those callers — which is how it became visible. Worth its own issue.
- `cargo test -p xerj-engine --test painless_script_limits` aborts with a stack overflow: pre-existing, reproduces on clean `main`.
- The dogfooder's most-wanted feature — *"you could have projected here"* — is deliberately absent. Quantifying it means inventing a projection the caller did not write, which is the exact hypothetical this feature refuses to report. A non-numeric nudge would be a reasonable separate change.

## Reviewer notes

- One plain-ASCII line at the end of a `brain` index run invites the reader to tell a teammate. Nothing asks anyone to seed praise into commit messages, code comments or PR descriptions — that would put promotional text into someone's repository under their name.
- The sibling PR #335 owns the `--help` feedback invitation; this deliberately does not touch it.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -6657,6 +6657,28 @@ pub struct EsSearchQueryParams {
     /// URL parameter or a body field; whichever is set wins (body
     /// takes precedence when both are provided).
     pub explain: Option<String>,
+    /// xerj extension, `?savings=`: report the `_source` payload this
+    /// response did NOT have to send, relative to the response the caller
+    /// would have received with no `_source` clause at all.
+    ///
+    /// * absent / `true` / `sampled` — **the default**: long arrays are
+    ///   extrapolated from a bounded sample, everything else counted exactly,
+    ///   and the block says `"measured": "sampled"`.
+    /// * `exact` — serialize and count every omitted byte.
+    /// * `false` / `off` / `0` — do not measure, emit nothing.
+    pub savings: Option<String>,
+}
+
+/// Parse `?savings=` into a measurement mode. Anything unrecognised falls
+/// back to the default rather than 400-ing: this is a reporting extension,
+/// and a typo in it must never fail a search.
+fn parse_savings_mode(param: Option<&str>) -> xerj_query::ast::SavingsMode {
+    use xerj_query::ast::SavingsMode;
+    match param {
+        Some("false") | Some("off") | Some("0") | Some("none") => SavingsMode::Off,
+        Some("exact") => SavingsMode::Exact,
+        _ => SavingsMode::Sampled,
+    }
 }
 
 pub async fn search_all(
@@ -8985,6 +9007,9 @@ async fn search_impl(
             // (see the injection block above) — internal field, never on
             // the wire.
             r.leaf_ts_field = implicit_leaf_ts;
+            // xerj extension `?savings=` — `_source` savings measurement, on
+            // by default. Internal field, never echoed on the wire.
+            r.savings = parse_savings_mode(params.savings.as_deref());
             r
         }
         Err(e) => return ApiError::new(e).into_response(),
@@ -8992,6 +9017,14 @@ async fn search_impl(
 
     // Execute search on each index and merge results.
     let mut merged_hits: Vec<(String, xerj_query::executor::Hit)> = Vec::new(); // (index_name, hit)
+
+    // Per-(index, doc id) omitted bytes, re-summed after the merge below over
+    // the hits this response ACTUALLY emits. The engine measured a per-index
+    // page; multi-index merging, PIT filtering, kNN capping and collapse can
+    // all drop hits from it afterwards, and counting a dropped hit's bytes
+    // would inflate the number.
+    let mut savings_by_hit: HashMap<(String, String), u64> = HashMap::new();
+    let mut savings_record: Option<xerj_query::executor::PayloadSavings> = None;
     let mut total_count: u64 = 0;
     let mut total_relation = "eq".to_string();
     let mut merged_aggs: Option<Value> = None;
@@ -9182,6 +9215,12 @@ async fn search_impl(
                             }
                         }
                     }
+                }
+                if let Some(savings) = result.savings {
+                    for (id, bytes) in &savings.per_hit {
+                        savings_by_hit.insert((idx_name.to_string(), id.clone()), *bytes);
+                    }
+                    savings_record.get_or_insert(savings);
                 }
                 for hit in result.hits {
                     merged_hits.push((idx_name.to_string(), hit));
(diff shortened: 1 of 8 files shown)


xerj-org:
Conflicts against current `main` are resolved and verified, but **I cannot push the fix to this PR's branch.** The head is `probelabs/xerj:feat/savings-stats`, and although `maintainerCanModify` reports `true`, GitHub does not honour maintainer edits on **organization-owned** forks. Confirmed empirically rather than assumed:

```
remote: Permission to probelabs/xerj.git denied to xerj-org.
GET /repos/probelabs/xerj -> permissions {push:false, maintain:false, admin:false}
```

Not a token-scope problem — the PAT has full `repo` scope; the account simply has no write bit on that fork. (Merging your PRs works fine; only pushing to them doesn't.)

**The resolved merge is waiting on `xerj-org/xerj` at branch `merge/pr340-conflicts-resolved` (commit `012fd09c`).** To update this PR in place and re-run its CI:

```sh
git fetch https://github.com/xerj-org/xerj merge/pr340-conflicts-resolved
git push origin FETCH_HEAD:feat/savings-stats
```

### What actually conflicted: one hunk

Only `engine/crates/xerj-server/src/brain.rs`, the `agents:` hint line. Both sides rewrote the same `println!`:

- **main** (#333, the onboarding 401 fix) changed `XERJ_URL={} xerj-mcp` to `{} xerj-mcp` via `mcp_env(&cfg.url, api_key.as_deref())`, so the printed command carries the credential the run actually used — the whole point of that fix was that banners stopped advertising commands which 401.
- **this PR** kept the old line and appended a "tell a teammate" note.

Resolved by keeping main's `mcp_env(...)` form and placing this PR's note below it. Both intents survive; the net change to `brain.rs` is +6 lines, matching the original diff.

### Clean merges audited anyway

A hunk can merge cleanly and still be wrong when the other side changed the surrounding facts, so the clean merges were checked rather than trusted:

- **#330's compression plumbing in `xerj-engine/src/index.rs` is intact** — all 84 `compress`-matching lines are byte-identical to `origin/main`. The two PRs' hunks don't overlap: main's sit around merge-publication and the doc-values sidecar, this PR's in the search-response path.
- The `apply_source_filter` rewrite looked alarming in the diff, but main hasn't touched that function since the merge base — it's this PR replacing its own code with `apply_source_filter_measured`. The `Default` / `Enabled(true)` / `Enabled(false)` branches and the dotted-companion path were re-read and behave as before.
- `WalSync::Async`, used by the new `savings_stat_is_honest.rs` harness, still exists after the #341 WAL fsync scheduler landed.

`cargo fmt --all -- --check` exits 0. Final diff versus main is 8 files, +2168/-32 — identical to this PR's stated diff, so the merge added main's work without dropping or altering the feature.

I did **not** open a replacement PR and close this one. That would move the PR identity and its CLA-gate state off a contributor's branch, which isn't mine to decide.

PR #343, opened 2026-08-13
docs(llms): close the first-contact defects a 22-run agent study found

xerj-org:
From the first-contact study in `demo/playbooks/FTX_PROBE_2026-08-12.md`: 22 simulated agents (4 personas × haiku/sonnet/opus/fable) were given only the canonical README prompt and `llms.txt`. **6 of 12 wave-1 runs hard-failed, all of them the two shell-less personas at every model tier.**

Closes thirteen defects in `llms.txt` / `llms-full.txt`, the highest-signal ones being:

- **`--approve fast` was documented as the safe cheap answer** to the estimate gate. It sets `--no-graph`, which on affected builds indexed no source code at all while exit 3 reported success. Warning added, version-scoped so it stays true now the binary is fixed.
- **"Semantic search" read as neural-by-default by 9 of 12 probes** — the lexical-embedder caveat sits two screens from the text that triggers the belief. Inline qualifiers added at the trigger points; the caveats block itself is untouched, since probes praised it.
- **The reference-coding section teaches mechanism-phrased queries**, exactly what the default feature-hashing embedder handles worst, with no cross-reference to that caveat and no statement that `--embed-mode` is an at-ingest decision requiring a full re-index to change.
- **Node lifecycle was undocumented** (7 probes): there is no `--port` flag, ports are TOML-only, and the bind address, stop procedure and auth-enable path were all unstated.
- **The three agent tool schemas were claimed but never linked** — the one sentence aimed at shell-less agents, and the only unlinked claim in a document that links everything else.
- A verified-manual-install path, which wave 2 showed flips the policy-constrained persona from blocked to success.

Verified rather than assumed while writing: the estimate gate is in **no** released binary (checked rc.14 and rc.15 source), the installer fails **closed** on checksums rather than warning, and the worked reference-coding query uses field names read out of `extract/code.rs`.


xerj-org:
Cross-reference: this PR and #343 both rewrite `landing/llms.txt` and `landing/llms-full.txt`, and they will conflict. Flagging it now rather than letting whichever merges second discover it.

They are complementary in intent, not competing:

- **#323 (this one) restructures.** Reference coding moves from section 12 of 14 to the front, the contribution protocol comes up with it, and the file shrinks from 50 KB — an agent that stops reading after the endpoint map currently sees neither.
- **#343 fixes content defects** found by a 22-run first-contact study (`demo/playbooks/FTX_PROBE_2026-08-12.md`), where 6 of 12 wave-1 runs hard-failed. Thirteen specific corrections: `--approve fast` documented as the safe option when it selected `--no-graph` and indexed no code; "semantic search" read as neural-by-default by 9 of 12 probes because the lexical caveat sits two screens from what triggers the belief; the node lifecycle (no `--port` flag, ports are TOML-only, bind address and auth-enable unstated); the three agent tool schemas claimed but never linked.

**Suggested order: land #323 first, then rebase #343 onto it.** Structure before content is the cheaper direction — #343 is thirteen localised edits that can be re-placed into new sections, whereas re-applying a whole-file restructure on top of them means redoing the restructure. I am happy to do that rebase.

Worth noting for whoever reviews both: two of #343's findings are about *placement* rather than wording, and #323 may already fix them — the lexical-embedder caveat being far from the text that triggers the misreading, and reference coding being buried. If so, those two edits should be dropped rather than duplicated.

xerj-org:
Folded into #323 — **please close this once #323 merges** (leaving it open until then, so nothing is lost if #323 slips).

#323 restructures the same two files (reference coding from section 12 to section 1, the contribution protocol to section 2, and `llms.txt` cut from 50 KB). Landing the structure first and re-placing this PR's content into it was the cheaper direction, so #323 is now rebased on current `main` and carries all of this PR's work in its final commit.

**Eleven of the thirteen fixes are re-placed** — `--approve fast` / `--no-graph` warning, inline "semantic is lexical by default" qualifiers at the trigger points, the reference-coding ↔ lexical-embedder cross-reference plus `--embed-mode` as an at-ingest decision, node lifecycle (no `--port`, bind address, stop, restart-preserves-indices, auth-on path, background+readiness-poll), data dir outside the indexed tree, the PATH consequence, released version + `--help is the authority`, all three schema URLs, the manual verified-install path, the shell-less handoff block plus `autoindex` labelled CLI-only, the worked reference-coding query with real field names and per-corpus scoping, the reconciled llms-full gate contract, and the honesty patches (HuggingFace Hub download with no XERJ-pinned checksum, RSS sizing).

**Two were dropped as already fixed by the restructure, not duplicated:** both were placement rather than wording — the lexical caveat sitting far from what triggers the misreading, and reference coding being buried below the recipe list. #323 moves reference coding to the top of the file and the caveats immediately above Core agent operations, so re-adding a positional fix would have duplicated content instead of fixing anything. The wording half of the "semantic" fix (inline qualifiers) *is* carried, since that part is not positional.

**Version scoping was re-checked against v1.0.0-rc.16, released today**, rather than carried over — this PR was written when rc.15 was current and several of its claims have since flipped:

- the estimate gate (`--approve`/`--max-minutes`/exit `4`) **is** in a release now: rc.16, not "in no release";
- the `--no-graph` code-junking defect (#294) is **fixed in rc.16** (#316, #338), which also reports `code_files`/`code_files_indexed`/`code_files_junked` — the warning is kept and re-scoped to "up to and including v1.0.0-rc.15", with the `_count` verification query for anyone on an older binary;
- a `--port` flag now exists on `main` (it did not when this was written) but is in no release, so the text says ports are TOML-only on rc.16 and earlier and names the flag as next;
- `xerj mcp` ships in rc.16, and `mcp-tools.json` is now 10 tools against the OpenAI/Anthropic files' 6 — the schema section states that instead of "all three define the same six".

One correction to the worked query: it now uses the shape `catalog.rs` actually emits (`bool.should` of `multi_match` over `["body","defs"]` plus a `match_phrase` on `defs` boosted 4) rather than a `defs^3` multi_match — that file comments explicitly against boosting `defs` inside the multi_match, because a conceptual query then scores on incidental symbol-word overlap.

Thanks for the study — the FTX playbook did the work here; #323 just gave it somewhere to live.

xerj-org:
Cross-reference from the #323 rebase pass.

#323 is now rebased on current `main`, conflict-free, and **16/16 green**. It carries this PR's content: I checked all thirteen clusters against its files individually and twelve are present verbatim in substance (the thirteenth pair was pure placement, which the restructure fixes by construction).

The two branches **cannot both merge** — `git merge-tree` conflicts in both files. `llms.txt` collides completely: #323 rewrites all 239 lines in one hunk, so every hunk here lands inside it. `llms-full.txt` collides region-for-region (this PR at `main` lines 60/95/111/137/186/245/306/679; #323 at 60/90/111/153/202/273/338/705).

Three points where #323 supersedes rather than copies, so nothing is lost by closing this one:

- The `["defs^3","body","title"]` multi_match example here measures **worse** than flat `body, defs, title` in the reference-coding toolkit (top-1 3/6 top-3 6/6 flat, vs `defs^3` top-1 2/6). `catalog.rs:282-295` argues the same point at the site that emits the query: a boost amplifies incidental token overlap, while the phrase clause requires the whole sequence. #323 uses the phrase form and carries no `defs^N`.
- The estimate gate is described here as being in no release — correct against rc.15, but it shipped in **rc.16**. #323 is scoped to rc.16.
- The three tool schemas do not match: `mcp-tools.json` is an object with **10** tools, `openai-tools.json` and `anthropic-tools.json` are bare arrays of **6**. #323 states that instead of implying parity.

One gap the restructure opened, now fixed on #323 (`fcb1396`): `llms.txt` pointed at `llms-full.txt` §3.1 for "the full field contract" and it was not there. The `ax_*` field list this PR had inline is now in §3.1, verified against `extract/code.rs:437-474` and `sync_executor.rs:1496-1513`, plus `ax_paths`, which neither PR mentioned.

Suggest closing this unmerged once #323 lands. Owner's call.

xerj-org:
Rebased onto current `main` (`d8be09cf`) — **CLEAN / MERGEABLE, 16/16 checks pass**. Three llms.txt conflicts, all reconciled by keeping both sides; nothing `main` added was dropped:

| conflict | `main` side | this branch | resolution |
|---|---|---|---|
| Caveats block | `_passage` bullet (#327) | RSS-sizing sub-bullet under the ingest-scale bullet | sizing sub-bullet stays attached to its parent, `_passage` bullet follows it |
| Core agent operations | `_passage` sentence on `xerj_search` | CLI-only note on `xerj_autoindex`, lexical note on `xerj_semantic_search` | all three kept |
| Reference coding step 3 | `defs` + `multi_match`/`match_phrase^4` recipe | lexical-phrasing paragraph | recipe first, phrasing paragraph after |

## Claims that had gone stale — re-verified against the binary, not the old doc

`v1.0.0-rc.16` was published `2026-08-13T01:49Z` (16 assets, not a draft). The branch was written on 08-12 and scoped its warnings to "rc.15 is the latest release, rc.16 is not tagged". A second commit corrects that:

1. **The estimate gate is released.** `--approve`/`--max-minutes`/exit 4 are in the rc.16 tag (13 matches for `"--approve"` in `cli.rs`; 0 on rc.15 and rc.14). Was documented as "NOT in a release yet".
2. **The `--no-graph` zero-code defect is fixed in a release.** `f9d76b5b` (#294, PR #338) is an ancestor of `v1.0.0-rc.16` and absent from rc.15. Warning re-scoped to "rc.15 and older" + "fixed in rc.16"; the verify-your-own-binary `_count` recipe stays, because stale binaries are the common case.
3. **`--port` exists.** `ef5bf4ad` is on `main`, so the bolded "There is no `--port` flag" was false. It sets `es_compat_port = N` and derives `N+1`/`N+2` (`main.rs:521-539`). It is *not* in rc.16, so the bullet documents the flag, dates it after rc.16, and keeps TOML as the answer for every released binary.
4. **"All three define the same six operations" was wrong even on the branch's base.** `mcp-tools.json` has carried **ten** tools since `6725da55`. `scripts/mcp-schema-check.sh` exists precisely because this drifted once before — its comment says "six tools published against ten served". Now states 10/6/6, plus the fact the old phrasing hid: `xerj mcp` is a real stdio MCP server in the binary since rc.16.

Verified and left alone: installer checksum chain + fail-closed refusal (`landing/get:131-156`), `--bind` default `127.0.0.1`, `"xerj v{} stopped. Goodbye."`, `admin.key` at `0o600`, `--insecure` disabling TLS *and* auth, `--prefix ax` / `--state-dir`, `MAX_MAX_MINUTES = 10080`, `--pdf-workers` max 4, every flag in the llms-full synopsis, the HF-Hub model triple + `embedding.local_model_dir`, and `data/` genuinely absent from `DEFAULT_IGNORE_PATTERNS`.

## Overlap with #323

#323 rewrites `landing/llms.txt` wholesale (one hunk, `-1,239 +1,272`), so it overlaps this PR **everywhere in that file** — there is no line-range subset to carve out. It is not a competing account, though: its `ea73b356` ("fold PR #343's first-contact fixes into the new structure") already absorbed this branch's content, and it independently reached the same four corrections above (rc.16 as current release at its line 115; `--port` on `main` but not rc.16 at 156; the gate shipping in rc.16 at 208; #294 fixed in rc.16 at 210; ten-tool MCP + `xerj mcp` at 265-267). One difference worth keeping from #323: it scopes the six-operation sentence to "The OpenAI and Anthropic files", which is the cleaner phrasing.

So the two are consistent on the facts and differ only in structure. Suggested order: **merge this one first** (narrower, green, CLEAN), then rebase #323 — its restructure is the superset and should win on layout. Merging #323 first instead would strand this PR entirely.

Docs-only, no crate touched. `cargo fmt --all -- --check`: clean.

xerj-org:
Reviewed by two independent agent passes (Claude Opus 5), the second tasked with refuting the first. Both landed on needs-work, and the decisive reason turned out to be governance rather than content.

**Closing this in favour of #365**, which lands the good part.

### Why not merge as-is

The thirteen FTX fixes this PR is titled and described for already landed via #323 (merged as `5bfe3e2`). What is actually left here is 4 hunks, +15/-8, across `landing/llms.txt` and `landing/llms-full.txt`. The body does not merely overstate that — it is refuted by its own diff: it asserts *"there is no `--port` flag"* while hunk 1 documents the flag and its `PORT+1`/`PORT+2` derivation. On a squash-merge that sentence becomes the permanent commit message for a commit saying the opposite, in a repo whose `docs_capability_lists.rs` exists specifically to end prose claims outliving source.

@xerj-org's own comment on this thread already called it: *"Folded into #323 — please close this once #323 merges."* #323 has merged. The later "merge this one first" comment predates that landing.

### Two defects in the residue, now fixed in #365

**The `:9200 is already taken` fence was not safe to paste.** It put the new `--port` line and the TOML route in one block, both on `--data-dir ~/xerj-data`. On a `main` build pasting it starts a node on 9280, then the TOML start dies on the **data-dir lock** — and the prose two lines up explains `already in use` as a *port* collision, so the reader goes chasing ports. In a PR about first-contact failures, that reintroduces the exact failure class.

**`count the array rather than assuming they match`** is wrong for the one file the sentence is about. Checked directly:

```
mcp-tools.json:       top-level=dict  jq 'length'=1   actual tools=10
openai-tools.json:    top-level=list  jq 'length'=6   actual tools=6
anthropic-tools.json: top-level=list  jq 'length'=6   actual tools=6
```

Follow that instruction literally and the MCP file looks like the *smallest* of the three.

Third, quieter one: hunk 1 fixes a dangling antecedent and then creates another. Inserting the `--port` paragraph between the `[server]` block and `All three must be distinct.` leaves that sentence reading as a constraint on `--port` — which derives its own two extra ports and cannot collide with itself.

### What was verified, not assumed

Merge test is genuinely clean against `224f84c`, 16/16 checks pass, `mergeStateStatus=CLEAN`, and `closingIssuesReferences` is `[]` (so "closes thirteen defects" closes nothing on the tracker). All four factual claims in the residual diff check out against source: `--port` derivation at `xerj-server/src/main.rs`, the fusion string at `parser.rs:2926`/`:2940` asserted by `test_hybrid_learned_rejected` at `:5638`, MCP framing at `xerj-mcp/src/lib.rs` `write_msg`, and the 10/6/6 counts. The content was never the problem.

Also affirmatively checked for the #352 class: no other open PR touches either file, and this diff contains no compilable code, so there is no merged-workspace risk here.

#365 carries the three correct hunks with those three problems fixed, under a title and body that match the diff. `docs_capability_lists` passes 15/15 there.

PR #347, opened 2026-08-13
test(autoindex): stop one timing flake cascading into 65 failures (#336)

### LINKED ISSUE #336: One timing flake cascades into 65 failures: panics poison FAILPOINT_TEST_LOCK / HTTP_E2E_LOCK and every later test fails with PoisonError
## Summary

One timing-sensitive test flaking turns an otherwise-green `cargo test -p xerj-autoindex` run into **65 failures**, because the panic poisons module-global `Mutex`es that ~63 unrelated tests then acquire. The run looks like a serious regression; it is one flake.

## What happens

Two tests assert on wall-clock behaviour and fail under machine load:

- `failure_resume_http_tests::a_run_reports_progress_through_every_phase_and_closes_the_stream` — asserts *"a sub-2s run relays exactly one bar"*; observed **left: 3, right: 1** on a loaded box
- `esclient::tests::all_timeout_retry_path_is_bounded_to_six_attempts_without_final_sleep`

Both hold `FAILPOINT_TEST_LOCK` / `HTTP_E2E_LOCK` when they panic. `Mutex::lock()` on a poisoned mutex returns `Err`, and the `.unwrap()` at every other call site turns that into a fresh panic — so every subsequent test serialised on those locks fails with `PoisonError`, regardless of whether the code under test is correct.

Verified as pre-existing and unrelated to any change: stashing a working branch and re-running on the clean tree reproduces the same two failures, and therefore the same cascade.

## Why it matters

- **It manufactures false regressions.** A contributor or agent sees 65 failures and reasonably concludes they broke something. Establishing otherwise costs a stash-and-rerun cycle on a full workspace build, which on this codebase is expensive.
- **It hides real ones.** With 63 cascading `PoisonError`s in the output, a genuine failure among them is invisible.
- **It is load-dependent**, so it fires exactly when CI is busy or a dev machine is doing something else — and see the note below about the runner fleet masking it.

## Fix

The poisoning cascade and the flake are separable, and both are worth fixing.

**1. Contain the cascade** — a panicking test should fail one test. The codebase already uses this idiom in a few places; make it the rule at every call site of these locks:

```rust
let _g = FAILPOINT_TEST_LOCK.lock().unwrap_or_else(|p| p.into_inner());
```

The mutex here is a *serialisation* device for tests that share global failpoint/HTTP state, not a guard over data whose invariants a panic could break, so recovering the guard is correct rather than a papering-over.

**2. Fix the underlying flake.** `"a sub-2s run relays exactly one bar"` asserts a wall-clock-derived count. Under load the run exceeds 2s and legitimately relays 3 bars. The assertion should be on the progress *contract* (bars are monotonic, a terminal record is emitted exactly once, no bar follows the terminal line) rather than on how many arrive in a window the test cannot control.

## Note for whoever picks this up

This is plausibly invisible on the current CI fleet for the same reason other timing bugs have been: 2-core runners exercise very different timing than a 32-core box, and this cascade was found on the latter under heavy parallel load. Reproducing it may need `taskset`, deliberate background load, or a runner with more cores — not just re-running the suite.

## Reproduction

```sh
cd engine
# with several other cargo builds running, or otherwise loaded:
cargo test -p xerj-autoindex -j 12
```

Expect the two wall-clock tests to fail, followed by ~63 `PoisonError` failures in tests that touch neither.

xerj-org:
Fixes #336.

Two wall-clock-sensitive tests fail under machine load. Both hold a module-global serialisation mutex when they panic, poisoning it — and every later test acquiring that lock panics again on `.unwrap()`. **One flake became a 65-failure run**: 2 real failures plus ~63 `PoisonError` casualties in tests that touch neither.

That manufactures false regressions. A contributor or agent reasonably concludes they broke something, and disproving it costs a stash-and-rerun on a workspace where a full build is tens of minutes. It also hides genuine failures inside the noise.

**All 137 call sites** on the `*_TEST_LOCK` / `*_E2E_LOCK` mutexes now recover the guard rather than unwrapping. These mutexes serialise tests sharing global failpoint and HTTP state; they do not guard data whose invariants a panic could break, so recovering from poisoning is correct here rather than a workaround.

Verified the rewrite stayed in scope: the diff adds `unwrap_or_else(|p| p.into_inner())` on 137 lines, **0 of them outside those two lock families**.

**This fixes the cascade only.** The underlying flake stands — asserting a wall-clock-derived bar count inside a window the test cannot control is wrong under load, and should assert the progress *contract* instead (bars monotonic, exactly one terminal record, none after it). Kept separate so this change stays mechanical and reviewable.

Worth knowing for review: the cascade may be invisible on the current CI fleet, for the same reason other timing bugs have been — 2-core runners exercise very different timing than the 32-core box where this was found under heavy parallel load.

### DIFF
--- a/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
@@ -658,9 +658,15 @@ fn assert_unsupported_delta_without_remote_mutation(
 /// between them is visible in the output rather than only in the index.
 #[test]
 fn the_legacy_terminal_line_and_run_document_report_code_coverage() {
-    let _guard = FAILPOINT_TEST_LOCK.lock().unwrap();
-    let _io_guard = state::FILE_DONE_IO_FAILPOINT_TEST_LOCK.lock().unwrap();
-    let _sink_guard = crate::progress::SINK_TEST_LOCK.lock().unwrap();
+    let _guard = FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let _io_guard = state::FILE_DONE_IO_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let _sink_guard = crate::progress::SINK_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
     let corpus = tempfile::tempdir().unwrap();
     let state_dir = tempfile::tempdir().unwrap();
     fs::write(
@@ -714,8 +720,12 @@ fn the_legacy_terminal_line_and_run_document_report_code_coverage() {
 
 #[test]
 fn a_rerun_after_an_added_file_succeeds_and_fresh_absorbs_it() {
-    let _guard = FAILPOINT_TEST_LOCK.lock().unwrap();
-    let _io_guard = state::FILE_DONE_IO_FAILPOINT_TEST_LOCK.lock().unwrap();
+    let _guard = FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let _io_guard = state::FILE_DONE_IO_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
     let corpus = tempfile::tempdir().unwrap();
     let state_dir = tempfile::tempdir().unwrap();
     fs::write(corpus.path().join("first.csv"), "id,value\n1,first\n").unwrap();
@@ -770,8 +780,12 @@ fn a_rerun_after_an_added_file_succeeds_and_fresh_absorbs_it() {
 
 #[test]
 fn completed_plan_rejects_vanished_content_group_before_remote_mutation() {
-    let _guard = FAILPOINT_TEST_LOCK.lock().unwrap();
-    let _io_guard = state::FILE_DONE_IO_FAILPOINT_TEST_LOCK.lock().unwrap();
+    let _guard = FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let _io_guard = state::FILE_DONE_IO_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
     let corpus = tempfile::tempdir().unwrap();
     let state_dir = tempfile::tempdir().unwrap();
     fs::write(corpus.path().join("first.csv"), "id,value\n1,first\n").unwrap();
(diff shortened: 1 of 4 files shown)


xerj-org:
## Rebased onto current `main`, and the cascade claim re-proved by measurement

**Why it was red.** The 10 failing checks were not about this PR. The branch was based on
`a9013f7a`, where `xerj-engine` pinned `tree-sitter = "0.25"` and `xerj-autoindex` wanted
`"0.26"`; `tree-sitter` declares `links = "tree-sitter"`, so exactly one version may exist in
the graph. That is a hard cargo resolution failure — every job died in 9–43s regardless of
content. #350 fixed it on `main` (both crates now `0.26`). Rebased onto `d8be09cf`; no
conflicts, no content change (`git range-diff` clean: 137 removed lines, 137 added, same set).

**Re-proving the actual claim.** A test-hygiene PR is worth nothing if the cascade still
happens, so rather than trust the original load-dependent observation I made the flake
deterministic and measured both sides. The injection is the real failure site, not a
synthetic panic — the wall-clock assertion in
`failure_resume_http_tests::a_run_reports_progress_through_every_phase_and_closes_the_stream`
flipped from `bars.len() == 1` to `== 3`, so it panics at exactly the line it panics at under
load, while holding `FAILPOINT_TEST_LOCK`.

`cargo test -p xerj-autoindex -j 32`, all three runs on the same rebased tree:

| run | tree | passed | failed | `PoisonError` |
|---|---|---:|---:|---:|
| A | rebased PR, no injection | **578** | **0** | 0 |
| B | rebased PR + injected flake | **575** | **1** | **0** |
| C | PR's lock changes reverted + same injected flake | **516** | **60** | **59** |

Run C is the cascade, reproduced on demand: one flake, 59 casualties in tests that touch
neither the failpoint nor the HTTP mock. Run B is the same flake on the same tree with this
PR applied — it fails alone. That is the claim, and it holds. (60 rather than the issue's 65
because I injected one flake; the original run had two.)

Run A's 578 is 576 lib unittests + 1 `journal_lock_inheritance` + 1 `scan_pool_width`.

**Scope re-checked after the rebase.** All 137 removed lines are
`*_TEST_LOCK`/`*_E2E_LOCK` `.lock().unwrap()`; 0 removed lines are anything else. A
multiline-aware scan of all seven module-global test mutexes in the crate finds **148
recovering acquisitions and 0 plain unwraps** — 137 from this diff plus the 11 in
`sync_executor.rs` that already used the idiom.

**Gates**, `cargo fmt --all -- --check` → exit 0; `cargo clippy -p xerj-autoindex
--all-targets -- -D warnings` → exit 0; `cargo test -p xerj-autoindex` → 578 passed, 0 failed.

### Two things a reviewer should know

1. **The underlying flake is still live**, exactly as this PR says. `assert_eq!(bars.len(),
   1, "a sub-2s run relays exactly one bar")` still asserts a wall-clock-derived count. It
   will still go red under load — it just costs one failure now instead of sixty. Worth a
   follow-up that asserts the progress contract instead.

2. **One detail in #336 does not hold up.** The issue names
   `esclient::tests::all_timeout_retry_path_is_bounded_to_six_attempts_without_final_sleep`
   as a second test that "holds `FAILPOINT_TEST_LOCK` / `HTTP_E2E_LOCK` when it panics".
   `src/esclient.rs` references no module-global test lock at all — that test uses only
   local mutexes, so its flake never poisoned anything and never contributed to the cascade.
   It is a plain flake. This does not change what this PR does or whether it is correct; it
   means the issue's "two poisoners" framing is really one poisoner and one independent
   flake, and closing #336 should not be read as having addressed the `esclient` timing
   assertion.

PR #349, opened 2026-08-13
feat(help): ask for bug reports and UX feedback, on by default (rebased #335)

xerj-org:
Supersedes #335, which could not be updated in place: its head is on `probelabs/xerj`, an **organization-owned** fork, and GitHub does not honour maintainer edits on those even though `maintainerCanModify` reports `true`. Every credential available was refused — both tokens (which resolve to `xerj-org`) and three SSH keys. Not a scope problem; merging that PR works, pushing to it does not.

This branch is `origin/main` plus #335 two commits, conflict-free. **@buger authorship is preserved on both feature commits** — only the PR identity moves.

### The one thing worth reviewing: a clean merge that would not have compiled

`main.rs:1714`, the test `insecure_help_documents_auth_not_just_tls` added by **#333**, sat in an auto-merged region calling `help_text()` with **no arguments** — while this work changes the signature to `help_text(feedback: bool)`. Git merged it silently because the two sides touched different lines. It is a hard compile error in the test profile that would only have surfaced in CI. Fixed to `help_text(true)` (the shipped default), which incidentally also proves the invitation does not reflow the `--insecure` paragraph above it.

Root cause of the collision: **#333 and #335 independently extracted the same `help_text()` refactor with different signatures**, and both were green in isolation.

### The two real conflicts, both against #333

1. `print_help` / `help_text` signature (`main.rs:183-191`) — took this branch signature, since the caller passes `xerj_common::feedback::enabled()`, and **merged both doc comments** rather than discarding main rationale.
2. `format!` arguments (`main.rs:~308`) — took the two-argument form adding `feedback::block(feedback)`.

### Cleanly-merged regions audited rather than trusted

- #333 reworded `--insecure` help survives intact (`main.rs:219-223`), still documenting that the flag disables TLS *and* auth and pointing at `<data_dir>/admin.key`.
- #337 `xerj mcp` SUBCOMMANDS block and its dispatch are untouched.
- This work never touches completion banners, so #333 corrected ones — which stopped printing commands that 401 — are unaffected.

`cargo fmt` clean on the three touched crates.

### Follow-up not done here

#337 added a fifth help surface, `xerj mcp --help`, after this work was written. By its own stated rule it now qualifies for the invitation and does not carry it. Also `xerj mcp --disable-feedback` currently errors as unknown.

### DIFF
--- a/engine/crates/xerj-autoindex/src/cli.rs
+++ b/engine/crates/xerj-autoindex/src/cli.rs
@@ -109,9 +109,16 @@ pub fn print_help() {
 /// The help text as a value, so tests can assert that a documented flag is
 /// still documented instead of trusting a `println!`.
 pub fn help_text() -> String {
+    help_text_with(xerj_common::feedback::enabled())
+}
+
+/// [`help_text`] with the feedback invitation forced on or off, so a test does
+/// not have to mutate the process environment to see both shapes.
+pub fn help_text_with(feedback: bool) -> String {
     format!(
         "xerj autoindex — point it at any folder and make the contents AI-searchable, zero config\n\
          \n\
+         {feedback_block}\
          USAGE:\n\
              xerj autoindex <folder> [OPTIONS]     discover + index a folder\n\
              xerj autoindex map [OPTIONS]          print the discovered data map\n\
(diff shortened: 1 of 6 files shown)


xerj-org:
Closing as no longer needed. This was a workaround for not being able to push the conflict resolution to #335, whose head is on the organization-owned fork `probelabs/xerj`. #335 has since been updated directly (head `f741015a`, 05:45), so the original PR is live again and this replacement would only duplicate it.

The resolution work is preserved on branch `pr335-conflict-resolution` if any of it is still wanted. One finding from it is worth carrying over regardless, because it is a **compile error that git merges cleanly**: `main.rs:1714`, the test `insecure_help_documents_auth_not_just_tls` added by #333, calls `help_text()` with no arguments, while #335 changes that signature to `help_text(feedback: bool)`. The two sides touch different lines so git does not flag it, but it fails to build in the test profile. #333 and #335 had independently extracted the same `help_text()` refactor with different signatures, and both were green in isolation.

PR #351, opened 2026-08-13
fix(deps): align tree-sitter so the workspace links one native library (main is red)

buger:

# `main` does not build

```
note: only one package in the dependency graph may specify the same links value
      to ensure that only one copy of a native library is linked in the final binary
```

`xerj-engine` asks for `tree-sitter = "0.25"`; `xerj-autoindex` asks for `"0.26"`. Cargo permits exactly one crate to `links` a given native library, so the workspace cannot resolve — **every build, every test, every CI job**. Reproduced on a clean checkout of `b36b86fe`.

## Why it slipped through

Neither PR is at fault on its own:

- **#327** brought `code_blocks.rs`, written and verified against **0.25**.
- **#317** moved autoindex to **0.26** for the Kotlin, Swift, Scala and Dart grammars.

Each merged green. Together they do not compile. A PR that only builds its own branch cannot see this class of collision — it appears at the merge, not in either diff.

## The fix

One line: align `xerj-engine` to `0.26`.

That direction is the safe one because **both crates already pin identical grammar versions** — every `tree-sitter-*` grammar is `0.23` on both sides. Only the core binding moves, and node-kind names come from the grammars, not the core. That is what makes this a one-line change rather than a re-verification of every table in `code_blocks.rs`.

`Cargo.lock` is unchanged: `0.26.12` was already locked for autoindex, so the engine now resolves to the copy that was being compiled anyway.

## Verified

| gate | result |
|---|---|
| `cargo check -p xerj-server` | **builds** — was a hard resolution failure |
| `cargo test -p xerj-engine --lib code_blocks` | **15 passed, 0 failed** |
| `cargo test --lib` (engine, autoindex, api, query, server) | **1503 passed, 0 failed** |
| ES-YAML conformance | **1366 passed · 0 failed · 3 skipped** (1369 total) |

The `code_blocks` row is the one that mattered. Those node-kind tables were hand-checked against each grammar's `node-types.json` under the 0.25 bindings, so a core bump that shifted any kind name would have surfaced exactly there. It did not.

## Reviewer notes

- This is the minimal unblock. It does not attempt to prevent recurrence — a workspace-level dependency on `tree-sitter`, or a CI check that resolves the workspace against `main` before merge, would stop the next one. Worth a follow-up issue rather than widening this PR while the tree is red.
- `cargo test -p xerj-engine --test painless_script_limits` aborts with a stack overflow. Pre-existing and unrelated; it reproduces on clean `main` once the build is unblocked.

### DIFF
--- a/engine/crates/xerj-engine/Cargo.toml
+++ b/engine/crates/xerj-engine/Cargo.toml
@@ -51,7 +51,7 @@ rustc-hash = { workspace = true }
 # (`xerj-autoindex/src/extract/code.rs`) — this module re-parses at RESPONSE
 # time rather than depending on that crate (autoindex depends on the engine,
 # not the other way around; a reverse dependency would be circular). ---
-tree-sitter = "0.25"
+tree-sitter = "0.26"
 tree-sitter-python = "0.23"
 tree-sitter-c = "0.23"
 tree-sitter-rust = "0.23"


xerj-org:
Thank you for this — you caught a real, tree-wide break, and the diagnosis is exactly right: `tree-sitter` declares `links = "tree-sitter"`, so #327 (engine on 0.25) and #317 (autoindex on 0.26) could each merge green and still fail to resolve together. That class of collision is invisible to a PR that only builds its own branch, which is precisely why it reached `main`.

I'm closing this as **already landed**, not as wrong. #350 shipped the identical one-line change a few hours ahead of it — same file, same hunk, and literally the same blob transition (`a32ae0e5` → `9a3e205e` in `engine/crates/xerj-engine/Cargo.toml`). It is on `main` as b88fcb33, merged in 2addda7e.

I verified redundancy rather than assuming it:

| check | result |
|---|---|
| `git diff origin/main...e41d9da6 --stat` | 1 file, 1 insertion, 1 deletion — only `tree-sitter = "0.25"` → `"0.26"` |
| `git diff origin/main:engine/crates/xerj-engine/Cargo.toml e41d9da6:engine/crates/xerj-engine/Cargo.toml` | empty — the file is already byte-identical on `main` |
| `git merge-tree --write-tree origin/main e41d9da6` | `ebaa3c427fa21238c9306f50dc577b902dbe4fbf` — **byte-identical to `origin/main^{tree}`**, so merging this would change nothing |
| `cargo metadata --locked` on `main` (d8be09cf) | exit 0; exactly one `tree-sitter` in the graph: `0.26.12` |
| `cargo check -p xerj-engine` on `main` | `Checking tree-sitter v0.26.12` … `Finished dev profile in 9.49s`, exit 0 |

So there is nothing in this branch that `main` lacks — that is the *only* reason it is closed, and it costs you nothing: the fix you wrote is the fix that is running.

Two things from your reviewer notes that I'd rather not lose with the PR:

- **Prevention.** Your point that neither branch is wrong on its own is the important one — no per-PR gate can catch this. A CI step that resolves the *merged* workspace (`cargo metadata --locked` against `refs/pull/N/merge` rather than the head) would have, in a few seconds and without compiling anything. Filed as #352, with the suggestion credited to you.
- **`painless_script_limits`.** Reproduced on clean `main` (d8be09cf), so thank you for that one too — `call_depth_limit_survives_the_multi_thread_block_in_place_path` overflows its stack and takes the whole test binary down with SIGABRT, leaving 7 of the 9 cases unreported. Filed as #353. Agreed it is pre-existing and had nothing to do with your change.

Thanks again for jumping on a red tree — the write-up in the body was more useful than the diff, and the "both crates already pin identical grammar `0.23` versions, so only the core binding moves" argument is the right reason this was safe to do as a one-liner.

Issue #352, opened 2026-08-13
CI: resolve the merged workspace before merge — two green PRs produced an unbuildable main (tree-sitter links conflict)


xerj-org:
## What happened

`main` was unbuildable for every target from a9013f7a until b88fcb33 (~PR #350):

```
error: failed to select a version for `tree-sitter`.
package `tree-sitter` links to the native library `tree-sitter`, but it
conflicts with a previous package which links to `tree-sitter` as well
note: only one package in the dependency graph may specify the same links value
```

Two PRs, each green on its own branch, combined into a tree that does not resolve:

- **#327** added `code_blocks.rs` to `xerj-engine`, pinned `tree-sitter = "0.25"`.
- **#317** moved `xerj-autoindex` to `tree-sitter = "0.26"` for the Kotlin/Swift/Scala/Dart grammars.

`tree-sitter` declares `links = "tree-sitter"`, so exactly one version may exist in the graph. Two
requirements that cannot unify are a **hard resolution failure**, not a duplicate-dependency
warning — so it takes out `cargo build`, `cargo test`, `cargo clippy` and every release job at once.

## Why no existing gate caught it

Neither branch is wrong. Each PR's CI resolved *its own* head, where only one of the two
requirements existed. The conflict is a property of the merge, and nothing in CI ever resolved the
merge before it became `main`. It then stayed red long enough that ~every open PR inherited a
5-ok/10-fail signature unrelated to its own content, and #351 was opened as a duplicate fix by an
external contributor who hit the same wall.

## Proposed guard

A cheap pre-merge step — resolution only, no compilation, a few seconds:

```sh
cargo metadata --locked --format-version 1 >/dev/null
```

run against the **merge result** (`refs/pull/N/merge`, which is what `actions/checkout` gives by
default for `pull_request`), not the PR head. That single command fails with exactly the error above
when two crates disagree on a `links` dependency, and also catches `Cargo.lock` drift via `--locked`.

Worth considering alongside it:

- Hoist `tree-sitter` (and the `tree-sitter-*` grammars) into `[workspace.dependencies]` so
  `xerj-engine` and `xerj-autoindex` cannot drift apart in the first place. Today both pin all
  grammars at `0.23` by coincidence of review, not by construction.
- A `cargo deny`-style check for multiple versions of any `links` crate, if we want the general case
  rather than tree-sitter specifically.

## Credit

The prevention idea is [@buger](https://github.com/buger)'s, from the reviewer notes on #351 ("a CI
[...]

PR #356, opened 2026-08-13
perf(engine): drop the term dictionary for dense_vector, and the doc-scan fallback with it (#328)

### LINKED ISSUE #328: dense_vector fields build a lexical term dictionary no query path can read — 43.7% of a measured index (RFC #148)
Raised in **[RFC #148 — per-field / per-type storage compression](https://github.com/xerj-org/xerj/discussions/148)**.

## The user problem, in the reporter's terms

@buger, in [their comment on #148](https://github.com/xerj-org/xerj/discussions/148#discussioncomment-17888592):

> The clearest result was not a better codec. […] schema-generated pooled and passage vectors were flattened to decimal strings by the generic dynamic-field walker and indexed into lexical FSTs/postings, even though semantic/HNSW retrieval never reads those structures.
>
> | `body_vector.fst` | 486 MiB |
> | `body_vector_chunks.fst` | 1.852 GiB |
> | Entire partial index | 3.449 GiB |
>
> Those accidental vector FSTs were 68% of the partial index.

They proposed the rule as an invariant: *"do not build lexical FSTs for generated vectors"* is the same class as *"do not build doc values for `text`"*.

The **generated**-vector half of that was fixed (#12). The **user-mapped** half is still live on `main`: a field a user explicitly maps as `dense_vector` still gets a full lexical term dictionary, postings and norms that no query path reads.

## What XERJ does today (measured on `main` @ 41bfd2de, v1.0.0-rc.16)

3,000 docs, one index, force-merged to a single segment. Mapping:

```json
{"body":{"type":"text"},"cat":{"type":"keyword"},"n":{"type":"long"},
 "emb":{"type":"dense_vector","dims":128,"index":true,"similarity":"cosine"}}
```

On-disk segment files:

```
3222765  <seg>.emb.fst      <-- lexical term dictionary over the vector
1468180  <seg>.seg
 262922  <seg>.body.post
   6244  <seg>.emb.post
   6243  <seg>.n.post
   5029  <seg>.dv
    911  <seg>.n.fst
    137  <seg>.body.fst
     91  <seg>.cat.fst
     80  <seg>.cat.post
     40  <seg>.{n,emb,cat,body}.norms
```

`emb.fst + emb.post + emb.norms = 3,229,049 B of a 7,391,185 B index — **43.7%**.`

`strings` on `emb.fst` returns digit runs: it is a term dictionary over the decimal renderings of the float components.

Nothing reads it:

| query on `emb` | result |
|---|---|
| `knn` (k=3) | 3 hits — served by `hnsw/graph.bin` |
| `term: "0.4288"` | `hits.total: 0` |
| `match: "0.4288"` | `hits.total: 0` |
| `exists` | 3000 — resolved from `_source`, not postings (`xerj-engine/src/index.rs:30355-30359`) |

### Root cause

`fts_excluded_fields` (`engine/crates/xerj-engine/src/memtable.rs:3847-3860`) is the single gate that flush, merge, the memtable analyser and `build_fts_field_configs` all share. Its doc-comment says it carries exactly **two** reasons:

* semantic-derived vector targets (the #12 fix), and
* fields declaring `"index": false`.

`FieldType::Vector` (`xerj-common/src/types.rs:223`) matches neither — a `dense_vector` mapped with `"index": true` for kNN has `options.indexed == true`, so the `index: false` arm skips it. The field then flows through `extract_fts_fields_excluding` (`xerj-engine/src/index.rs:19074-19091`) into `extract_field_text` (`index.rs:19059-19071`), whose `Value::Array` arm joins the elements with spaces — turning a 128-dim vector into 128 numeric tokens per document.

There is already precedent for the opposite direction in the same file: `load_hnsw_artifacts_sync` (`index.rs:26612-26665`) *refuses* an HNSW graph pinned to a non-`dense_vector` field because such a graph is "pure RAM + per-ingest maintenance cost with no query benefit". The lexical sidecar for a `dense_vector` field is the mirror image of that, and is not refused.

## What would have to change

Add a third reason to `fts_excluded_fields` — the field's declared type has no lexical representation — covering `FieldType::Vector`. That is the whole behavioural change; every write path already reads this one set, which is why the fix belongs there and not at four call sites.

Open questions a patch has to answer rather than assume:

* **`FieldType::Chunk`** carries a vector payload too (`types.rs:227`) and should be checked in the same pass; its chunk *text* is legitimately lexical, its vector is not.
* **Existing segments** keep their sidecars until re-merged. Worth confirming a force-merge is enough to reclaim, and saying so plainly rather than implying an in-place shrink.
* **Confirm nothing else reads `<field>.fst` for a vector field** — `_field_caps`, highlighting, and the `exists` doc-values fast path (`index.rs:36783`, `36791`) were checked and do not, but a patch should carry a test that pins it.
* The gain is corpus-shaped, like every other ratio in #148: it scales with `dims × docs` relative to the rest of the index. 43.7% here is one measured point, not a constant — do not publish it as universal.

## Repro

```sh
xerj -c <cfg> --data-dir /tmp/probe --insecure &
curl -XPUT localhost:9200/duprobe -H 'content-type: application/json' -d '{"mappings":{"properties":{
  "body":{"type":"text"},"cat":{"type":"keyword"},"n":{"type":"long"},
  "emb":{"type":"dense_vector","dims":128,"index":true,"similarity":"cosine"}}}}'
# bulk-load 3000 docs with a 128-float `emb`, then:
curl -XPOST "localhost:9200/duprobe/_forcemerge?max_num_segments=1"
find /tmp/probe -name '*.emb.fst' -printf '%s %p\n'
```

## Why this was not visible from any API

Worth recording, since it is the reason a 43.7% artifact went unnoticed: XERJ exposes no per-artifact size breakdown. `POST /{index}/_disk_usage` returns the recursive byte total with `"fields": {}` (`xerj-api/src/es_compat.rs:16869-16905`), honestly documented as not computing the ES per-category decomposition. `hnsw/` lives inside the index `data_dir` (`index.rs:12165-12175`), so vector and lexical bytes are summed into one number with no way to separate them. The breakdown above came from `find`, not from XERJ. This is also @buger's item 6 ("report semantic-vector and ANN artifacts separately"), acknowledged in the thread as agreed and not done.

## Rest of the thread, deliberately not filed here

Recorded so they are not lost, per the closing comment on #148:

* **Per-field positions opt-out** — blocked on `match_phrase` against an opted-out field failing loudly instead of returning `hits.total: 0` with a 200.
* **Centralising the physical-role matrix** so ingest / flush / merge / mapping cannot disagree — @buger's item 2. `fts_excluded_fields` is that centralisation for the FTS side; this issue is one missing row in it.
* **Separate vector/ANN reporting in the size scorecard** — see above.
* **Stored `_source` at 64.21% of real indices** is where the remaining `$/GB` is, and no lever in the RFC touches it.
* **Lever 2 (type-aware numeric DV encoding)** stays dropped — measured at ~−0.36% of `.dv`, and unconditional delta+zigzag+FOR measured *worse* than plain FOR.
* **Lever 3** is #318.

xerj-org:
> **This PR was written by an AI agent.** Everything under **Verified** below
> was produced by a command that was run in this worktree and whose output was
> read. Everything under **Assumed** was not executed and should be treated as a
> claim, not a result.

Fixes #328.

A field mapped `dense_vector` was fed through the lexical indexing path and
given a full FST + postings that no query path can read. kNN is served by
`hnsw/graph.bin`, `exists` from `_source`/doc values, and `_field_caps` and
highlighting never open the field's term dictionary — so every one of those
bytes was unreadable by construction. `extract_field_text`'s `Value::Array` arm
joins elements with spaces, so a 128-dim vector became one enormous
decimal-string term per document.

The fix has four parts: the write-side exclusion (`memtable::fts_excluded_fields`
unions in `lexically_typeless_fields`, covering `<field>`, `<field>_chunks` and
`__xerj_passage_meta__<field>`), the `*`-expansion filter, a plan-time lowering
of any clause that NAMES a vector field to `match_none`
(`lower_lexically_typeless_clauses`), and the nested-object case, where the
components were landing in the *parent's* term dictionary because the segment
builder flattens the whole object into one text field.

## What changed in this revision

The previous revision's central claim was false and this revision exists to fix
it. It said in three places — CHANGELOG, commit message and the
`fts_excluded_fields` doc comment — that **one** answer changes. Nine do, and
one of them is `wildcard {emb:"0*"}` going from matching **every document** to
matching none. A `_score`/`max_score` change was also undocumented, and the
commit message asserted the opposite. Every number in this PR, the CHANGELOG and
both doc comments has been re-measured in this worktree for this revision;
nothing is carried over.

No engine logic changed in this revision. The delta is: accurate release notes,
seven more shapes and two score properties pinned in the regression test, a
durable-byte ceiling that makes the published saving enforced, and corrected
gate figures.

---

## Verified

Every number below came from a command run against this branch in a private
`CARGO_TARGET_DIR`, and every "fails before" claim was produced by reverting
exactly the named source hunk and watching the test fail.

### Bytes

One corpus (5,000 docs × 128 dim: text + keyword + long + `dense_vector` +
`_chunks`), built in three source states, after `force_merge(1)`;
`fs::metadata().len()` sums over the data dir.

| file | `main` | this branch |
|---|---:|---:|
| `<seg>.emb.fst` | 13,256,659 B | 0 |
| `<seg>.emb.post` | 10,834 B | 0 |
| `<seg>.emb.norms` | 48 B | 0 |
| `<seg>.emb_chunks.fst` | 26,518,005 B | 0 |
| `<seg>.emb_chunks.post` | 10,834 B | 0 |
| `<seg>.body.fst` / `.cat.fst` / `.n.fst` | 278 / 53 / 209 B | unchanged |
| **durable index** (all files bar `.wal`) | **54,068,549 B** | **14,251,975 B** (−73.6%, 3.79×) |

The total excludes `.wal` on purpose: the log's tail is reclaimed
asynchronously, so the whole-directory figure is not reproducible. Seven runs of
the committed 300-doc regression test put the whole directory between
858,147 B and 1,070,410 B while the durable bytes moved 230 B — which is why the
previous revision's "3,496,265 B → 1,070,438 B" could not be reproduced from
that test. It now prints both totals and asserts a ceiling on the durable one:
**3,249,323 B → 857,843–858,071 B**, watched fail at exactly 3,249,323 with both
source files reverted to `origin/main`.

The `_chunks` companion is the bigger half — 26,518,005 B against `emb.fst`'s
13,256,659 B — so a base-name-only exclusion leaves more behind than it removes.
On the nested fixture `<seg>.passages.fst` goes 398,558 B → 45 B, for both
mapping shapes that reach `FieldType::Vector`.

### Behaviour change: a class, not a row

Measured on the committed 300-doc × 128-dim fixture with `src/index.rs` and
`src/memtable.rs` reverted to `origin/main` and nothing else changed:

| query | before | after |
|---|---:|---:|
| `wildcard {emb: "0*"}` | **300** — every document | 0 |
| `fuzzy {emb: {value:"0", fuzziness:2}}` | **300** — every document | 0 |
| `wildcard {emb_chunks: "0*"}` | **300** | 0 |
| `fuzzy {emb_chunks: {value:"0", fuzziness:2}}` | **300** | 0 |
| `prefix {emb: "0"}` | 146 | 0 |
| `prefix {emb_chunks: "0"}` | 146 | 0 |
| `match_phrase_prefix {emb: "0"}` | 50 | 0 |
| `term {emb: <exact component>}` numeric | 1 | 0 |
| `terms {emb: [<exact component>]}` numeric | 1 | 0 |

The honest description is the class: a `dense_vector` and its `_chunks`
companion no longer have any lexical surface, so every lexical leaf naming one
answers 0. On `main` such a clause reached the stored-doc scan, which renders the
float array back to decimal text and matches against *that* — so a shape
answered 0 or answered the whole corpus purely according to whether the rendered
decimals happened to contain the probe. ES rejects these queries on a
`dense_vector` outright and Lucene gives a field with no terms an empty scorer,
so 0 is the answer adopted.

All nine are pinned in the `(d)` loop, which now accumulates every non-zero
shape instead of stopping at the first. Watched fail on `origin/main`:

```
a lexical query on a dense_vector (or its `_chunks` companion) must
match nothing; 9 of the shapes above did not:
  {"term":{"emb":0.09492116829611441}} → 1 hits
  {"terms":{"emb":[0.09492116829611441]}} → 1 hits
  {"prefix":{"emb":"0"}} → 146 hits
  {"wildcard":{"emb":"0*"}} → 300 hits
  {"fuzzy":{"emb":{"value":"0","fuzziness":2}}} → 300 hits
  {"match_phrase_prefix":{"emb":"0"}} → 50 hits
  {"prefix":{"emb_chunks":"0"}} → 146 hits
  {"wildcard":{"emb_chunks":"0*"}} → 300 hits
  {"fuzzy":{"emb_chunks":{"value":"0","fuzziness":2}}} → 300 hits
```

The two halves of the fix cover different rows: with the write-side exclusion
alone (lowering disabled) the same loop leaves exactly **four** non-zero, because
a multi-term query has no term dictionary left to enumerate while `multi_match`
still reaches the scan. Both counts watched.

### Behaviour change: `_score` moves for bools whose hit count does not

| query | hits | `_score` | `max_score` |
|---|---|---|---|
| `bool{should:[term emb, term cat:"even"]}` | 150 → 150 | 0.008402659 → 0.6931471 | 1.6931472 → 0.6931471 |
| `bool{must:[match body], must_not:[term emb]}` | 300 → 300 | 0.008402659 unchanged | 1.6931472 → 0.008402659 |

Cause: the `should.retain(|c| !c.is_match_none())` fold makes the surviving bool
projectable through `query_node_to_fts_with_keyword_fields`, so BM25 replaces the
stored-doc scorer. Isolated to the fold — with the write-side exclusion alone and
the lowering disabled, both print exactly `main`'s values. Both moves are toward
consistency: on `main` these reported a `max_score` **no returned hit carried**.
Pinned as properties rather than golden floats in a new `(d2c)` block
(`max_score == hits[0].score`; a bool whose only live `should` is
`term cat:"even"` scores exactly like that term query alone), both of which fail
on `main` with `max=Some(1.6931472) top=0.008402659`.

### Timings, 5,000 × 128, first call, `main` / write-side-only / this branch

| query | `main` | no lowering | this branch |
|---|---|---|---|
| `multi_match {fields:["emb"], query:"0"}` | 0 hits, 0.409 ms | **5,000 hits**, 123.5 ms | 0 hits, 0.012 ms |
| `multi_match {fields:["emb_chunks"], query:"0"}` | 0 hits, 2.24 ms | **5,000 hits**, 138.5 ms | 0 hits, 0.014 ms |
| `wildcard {emb: "0*"}` | **5,000 hits**, 316.6 ms | 0 hits, 95.1 ms | **0 hits**, 0.008 ms |
| `fuzzy {emb: "0"~2}` | **5,000 hits**, 798.5 ms | 0 hits, 97.7 ms | **0 hits**, 0.011 ms |
| `term {emb: "<component>"}` | 0 hits, 89.2 ms | 0 hits, 186.8 ms | 0 hits, 0.061 ms |
| `range {emb: {gte:-2, lte:2}}` | 0 hits, 119.8 ms | 0 hits, 134.5 ms | 0 hits, 0.009 ms |
| `exists {field: "emb"}` | 5,000 hits, 83.0 ms | 5,000 hits, 108.0 ms | 5,000 hits, 88.5 ms |
| `match {body: "liquidity"}` | 5,000 hits, 5.18 ms | 5,000 hits, 5.96 ms | 5,000 hits, 4.95 ms |

The two `multi_match` rows are correctness regressions, not slow paths — they
are why the write-side change cannot ship on its own.

### Gate

| command | result |
|---|---|
| `cargo fmt --all -- --check` | clean, exit 0 |
| `cargo clippy -p xerj-engine --all-targets -- -D warnings` | 0 warnings |
| `cargo test -p xerj-engine --profile ci-test --lib` | **580 passed, 0 failed** |
| same, on clean `origin/main` (baseline) | 547 passed, 1 failed (#372 flake) = 548 |
| `cargo test -p xerj-engine --profile ci-test --no-fail-fast` | 40 targets, **933 passed, 0 failed**, 23 ignored |
| `cargo test -p xerj-engine --test integration dense_vector` (dev, overflow-checks ON) | 3 passed, 0 failed |
| `cargo test -p xerj-engine --lib lexically_typeless` (dev) | 32 passed, 0 failed |
| `cargo test -p xerj-api --profile ci-test --no-fail-fast` | 23 targets, 354 passed, 0 failed, 2 ignored |
| ES-compat YAML suite, private instance on `:9433` | 200 files, **1366 passed · 0 failed · 3 skipped · 1369 total** |

The lib delta is **+32** unit tests (580 − 548), not the +31 the previous
revision claimed while also saying "32 in total" two paragraphs earlier: 24 in
`index::lexically_typeless_lowering_tests` and 8 in
`memtable::lexically_typeless_fields_tests`.

### Reference

Adapted from Lucene (Apache-2.0, same as XERJ), approach only. Every cited line
was re-checked against the checkout for this revision:
`KnnFloatVectorField.java:70` (vector attributes then `freeze()`, so
`IndexOptions` stays `NONE` — the premise), `LateInteractionField.java:36`/`:44`
(a per-document `float[][]` `extends BinaryDocValuesField` — the `_chunks`
half), `Terms.java:40` (`getTerms` returns `Terms.EMPTY`, so a field with no
terms yields an empty scorer, never a corpus walk — why 0 rather than a scan),
`MatchNoDocsQuery.java:23`, `BooleanQuery.java:271`/`:323`–`:329` (the per-occur
fold rules, clause for clause; `:276`'s pure-negative rule deliberately NOT
adopted because XERJ follows ES's bool, not Lucene's),
`ConstantScoreQuery.java:56` and `BoostQuery.java:87` ("bubble up
MatchNoDocsQuery"), `TestBooleanRewrites.java:299`,
`TestTermAutomatonQuery.java:1051`.

---

## Assumed

- **CI will reproduce these figures on its own runner.** The byte figures are
  from this machine. The durable-index ceiling is 1,500,000 B against a measured
  ~858 KB — the figure can grow 75% before it trips, and `main` sits 2.2× above
  it — precisely so a differently-sized HNSW graph on another host cannot flip
  it. That headroom has not been exercised on CI hardware.
- **Timings are indicative, not benchmarks.** Single first-call measurements on
  a loaded shared machine, run once per source state. The orders of magnitude
  are the claim; the third significant figure is not.
- **Segments written by an earlier release are untouched until a merge rewrites
  them.** Reasoned from the per-segment `fts_has_field` gate and pinned only for
  a reopen within one test binary — no cross-version upgrade was actually run.
- **`_field_caps` still reports `dense_vector`.** Asserted at the engine's
  `idx.schema()`, which is what the API layer serves it from; the HTTP endpoint
  itself was not called.
- **Scores are not part of the compatibility surface.** Stated as a judgement
  about what downstreams may rely on, not something measured.

---

## Round 3 — `862bcd7`, rebased onto `a20dd82`

**This revision was also written by an AI agent (Claude Opus 5).** Same split
applies: **Verified** = a command was run here and its output read; **Assumed**
= not executed in this revision.

Review rejected `e629a58` on a correctness regression the earlier rounds did not
find, and on three doc/CHANGELOG claims that did not survive checking. All of it
is addressed here; one part is NARROWED rather than closed, and that part is
called out below rather than buried.

### The regression that was found

`memtable::lexically_typeless_fields` synthesised `<field>_chunks` and
`__xerj_passage_meta__<field>` per `dense_vector` **unconditionally**, ignoring
what the schema declared that sibling to be. With the plan-time lowering on top,
a field the user had mapped became unqueryable with no error and no mapping
change.

**Verified**, on `{text body, keyword emb_chunks, dense_vector emb}`, 50 docs ×
16 dim, half `tenant-a` — merge base `ca4d75a` / `e629a58` / this revision:

| | `ca4d75a` | `e629a58` | here |
|---|---:|---:|---:|
| `term {emb_chunks:"tenant-a"}` | 25 | **0** | 25 |
| `terms {emb_chunks:["tenant-a"]}` | 25 | **0** | 25 |
| `bool{filter:[term emb_chunks]}` | 25 | **0** | 25 |
| `bool{must:[match body],filter:[term …]}` | 25 | **0** | 25 |
| `constant_score{filter: term emb_chunks}` | 25 | **0** | 25 |
| `bool{must_not:[term emb_chunks]}` | 25 | **50** | 25 |
| `bool{must:[match_all],must_not:[term …]}` | 25 | **50** | 25 |
| `terms` agg on `emb_chunks` | 25/25 | 25/25 | 25/25 |
| `<seg>.emb_chunks.fst` | 54 B | **0** | 54 B |
| `<seg>.emb.fst` | 16,713 B | 0 | 0 |

The `must_not` rows returned the 25 documents the exclusion was written to
remove. The aggregation row proves the data was present and correctly populated
throughout — this was a planning failure, not data loss. With a `text` mapping,
`match` / `match_phrase` / `multi_match` / `simple_query_string` / `prefix` all
went 25 → **0** → 25 and `<seg>.emb_chunks.fst` 71 B → **0** → 71 B, while the
unfielded `query_string {"query":"widgets"}` still answered 25 on `e629a58`
through the stored-doc scan.

### The fix

The name-based half now yields to **any** declaration it finds. Mapped as
anything other than `dense_vector` — `text`, `keyword`, `long`, `double`,
`date`, `boolean`, `ip`, `geo_point`, `binary`, `object`, `nested`, `chunk`, or
an ES type string `es_type_to_field_type` does not recognise (which lands on
`Object`) — and the declaration wins, the field keeps its whole lexical surface.
The predicate is stated as the complement of `Vector`, not as a list, because
the earlier enumerated form read as exhaustive when it was not. Both spellings
of a nested name resolve (a flat `FieldConfig` named `"passages.vec_chunks"`,
and a `vec_chunks` child under a `passages` object).

Lucene draws the same line: whether a field gets a term dictionary is read off
the field's own `IndexableFieldType`
(`lucene/core/src/java/org/apache/lucene/index/IndexingChain.java:1416`, in
`invertAndStore`), never off its name, and one name carrying two different sets
of index options is an outright error rather than a silent downgrade
(`IndexingChain.FieldSchema.setIndexOptions` → `assertSame` → `raiseNotSame`,
`:2213` / `:2187` / `:2193`).

### The residual an earlier revision asked you to accept is closed

An earlier revision of this PR excluded `long`/`double` under those names
anyway, and said so here, because `long`/`double` is what dynamic mapping
infers from a real multi-vector: `<field>_chunks` is user-supplied `_source` (an
array of float arrays) and `infer_field_type` walks it to its first scalar, so
the companion was DECLARED `Double` by the time the exclusion set was built.
That revision also claimed the only two ways out were per-field schema
provenance or giving up 67% of the saving, and that both were larger than #328.
**That claim was wrong and is retracted.**

The fix is neither: dynamic mapping now declines to invent the `FieldConfig` at
all. `index::is_undeclared_multi_vector_companion` refuses a `<vector>_chunks`
key when (a) the name matches, (b) `<vector>` is already declared
`dense_vector`, and (c) the value is a rectangular, non-empty array of
non-empty all-numeric arrays — the RFC #148 multi-vector shape and nothing
else. It is the same refusal `__xerj_passage_meta__<field>` already had
(`key.starts_with(PASSAGE_METADATA_PREFIX)`, three lines up), and the same
modelling Lucene uses: a per-document multi-vector is `LateInteractionField
extends BinaryDocValuesField`
(`lucene/core/src/java/org/apache/lucene/document/LateInteractionField.java:36`,
`:44`, `:45`) — its own field type, never a numeric one arrived at by
inference. With the ambiguity gone, a `long`/`double` under that name is
necessarily somebody's own field, so the predicate can yield to it.

**Verified**, by disabling each half on its own in the tree and re-running the
committed tests:

| | `emb_chunks.fst` (300-doc) | durable index | declared `long` sibling |
|---|---:|---:|---|
| this branch | **0 B** | 857,265–857,769 B | 44 B `.fst`, answers 25 |
| yield restored to the narrow form | 0 B | 857,445 B | **0 B, answers 0, `must_not` 50** |
| dynamic-mapping refusal disabled | **1,592,118 B** | **2,451,336 B** | 44 B, answers 25 |

**Verified** on `{text body, <T> emb_chunks, dense_vector emb}`, 50 docs, four
arms (`keyword`/`text`/`long`/`double`): every arm keeps its `.fst`
(54 / 71 / 44 / 48 B) and answers 25 on `term`, `terms`, `range`,
`bool{filter}`, `constant_score` and `bool{must:[match_all],must_not:[…]}` —
the `must_not` row that used to return all 50.

**Verified** on `{text body, dense_vector emb}` with `emb_chunks` left OUT of
the mapping, 50 docs, three arms: an `[[…],[…]]` value gets no mapping entry,
`.emb_chunks.fst` = 0 B, lexical clauses answer 0; a `"tenant-a"` value is
mapped `text`, keeps 54 B and answers 25; a `7` value is mapped `long`, keeps
44 B and answers 25. That last arm is the one the earlier revision broke
*without any declaration being involved at all*, and it was not covered by the
old test.

**Verified** fail-before: restoring only the `Double | Long` carve-out in
`declares_non_vector_shaped_field` fails all three of these integration tests;
disabling only `is_undeclared_multi_vector_companion` fails the 300-doc fixture
(durable 2,451,336 B) and the undeclared-companion test. Both halves are
load-bearing.

**Verified** gates, round 4, all locally: `cargo fmt --all -- --check` clean;
`cargo clippy -p xerj-engine --all-targets -- -D warnings` clean; ci-test
`--lib` 599 passed / 0 failed; ci-test `--no-fail-fast` across every
`xerj-engine` target 0 failures with `integration.rs` at 113 passed / 0 failed /
1 ignored; `cargo test -p xerj-api --profile ci-test --no-fail-fast` 0 failures;
dev profile (overflow-checks ON) `--lib` 599 passed / 0 failed and the six #328
integration tests 6 passed / 0 failed. This change moves the ES-compat
`_mapping` surface, so the conformance suite was run too — server booted from
`target/ci-test/xerj` on a spare port, `es-yaml-runner --dir
tests/es-compat-yaml/yaml`: **1366 passed · 0 failed · 3 skipped · 1369 total**.

**Assumed / known limit, stated rather than hidden:** the refusal needs the
`dense_vector` declared BEFORE the documents arrive. Index first and map
afterwards and `emb_chunks` already carries a dynamic `double` entry, which the
predicate then yields to — so that index keeps the companion's term dictionary
(**verified** 9,467 B on a 50-doc × 16-dim fixture) instead of reclaiming it. It
costs bytes, not answers, and both halves of that are asserted by
`a_companion_mapped_before_its_vector_keeps_its_dictionary_and_its_answers`. The
nested-under-an-object spelling is handled locally in
`merge_dynamic_children_into` (a companion and its vector are siblings in the
object being merged) and has the same ordering limit.

**Behaviour change this introduces**, documented in `CHANGELOG.md`:
`GET /<index>/_mapping` and `_field_caps` no longer list a `<vector>_chunks`
field that XERJ mapped for you, and aggregating/sorting on that companion stops
working. `_source` is unchanged, `knn`/`semantic` are unchanged, and a field of
that name you mapped yourself is unaffected.

### Three claims that were corrected

- The "same collision `semantic_derived_vector_fields` has always accepted"
  justification was **false** for the mapping shape this PR newly covers.
  `semantic_derived_vector_fields` carries the opposite contract in its own doc,
  and `observe_passage_chunks_in_source` reserves `<field>_chunks` only when
  `_source` holds a non-empty ARRAY. Replaced with the measured before/after.
- "The exclusion is *by declared type*, not by value shape" sat above a rule
  that excluded two names regardless of declared type. The two rules are now
  stated separately, with the name-based one labelled the weaker.
- The pattern-spec carve-out was defended with "the expansion sets already drop
  vector fields". **Verified** that this is not what happens:
  `{"multi_match":{"query":"0","fields":["emb.*"]}}` answers **300 of 300** on
  the 300-doc fixture, the same on `ca4d75a`, because the unprojectable clause
  falls to the stored-doc scan, which resolves the pattern against the source's
  field names. Not a regression; the reasoning was wrong. Also **verified**:
  `emb.*` "liquidity" 0, `emb_chunks.*` "0" 300, `emb` 0, `emb*` 0, `*` 0,
  `simple_query_string` `emb.*` 0. The class claim is now "every lexical leaf
  naming one **outright**", in both doc comments and the CHANGELOG.

### Verified — gate, on this rebased tree

```
cargo fmt --all                                            clean
cargo clippy -p xerj-engine --all-targets -- -D warnings   clean
cargo test -p xerj-engine --profile ci-test --lib
    594 total: 593 passed + the documented #372 flake
    (partial_multi_batch_failure_is_reported_without_losing_either_batch;
     pre-existing, fails only under full 20-way parallelism — the same
     binary at `-- --test-threads=4` is 594 passed / 0 failed, and the test
     alone is 1 passed / 593 filtered out)
cargo test -p xerj-engine --profile ci-test --lib lexically_typeless
    37 passed / 557 filtered out   (baseline 557, +5 unit tests; e629a58 had 32)
cargo test -p xerj-engine --profile ci-test --test integration
    111 passed / 0 failed / 1 ignored
cargo test -p xerj-engine --profile ci-test --test integration dense_vector
    4 passed / 0 failed
cargo test -p xerj-engine --test integration dense_vector    (DEV profile,
    overflow-checks ON)                                4 passed / 0 failed
```

Whole crate, on the rebased tree:
`cargo test -p xerj-engine --profile ci-test --no-fail-fast` → **41 targets, 969
passed, 0 failed** (the #372 pair included; it did not trip on that run). The
same command before the rebase, on `84e9166`, was 40 targets / 938 passed / 0
failed.

**ES-compat surface**: this change moves query answers on `dense_vector` fields,
so the YAML conformance suite was run against a server built from this exact
tree (`cargo build --profile ci-test -p xerj-server -p es-yaml-runner`, then
`es-yaml-runner --url http://localhost:9317 --dir tests/es-compat-yaml/yaml`):

```
ES-COMPAT YAML RUNNER · 200 files · http://localhost:9317
RUNNER_EXIT=0
```

0 failures is the gate. (The CI job's own comment records 1365 passed / 3
skipped at its last measured run; the extra case arrived with `main`.)

### Verified — fail-before, watched rather than recalled

- `declared_chunks_sibling_of_a_dense_vector_stays_queryable_unless_it_looks_like_one`
  with `memtable.rs` alone reverted to `e629a58`:
  `[keyword] a field the schema declares must keep its term dictionary; got []`
  — and with that file assertion temporarily neutered so the run reaches the
  answers: `[keyword] term: … got 0 for {"term":{"emb_chunks":"tenant-a"}}
  left: 0, right: 25`.
- `a_declared_sibling_keeps_its_lexical_surface` and
  `a_declared_nested_sibling_keeps_its_lexical_surface` with ONLY the loop body
  of `lexically_typeless_fields` reverted (doc comment and tests kept):
  `left: {"emb_chunks", "emb", "__xerj_passage_meta__emb"}, right: {"emb"}` and
  `declared nested sibling excluded: {"passages.vec_chunks",
  "__xerj_passage_meta__passages.vec", "passages.vec"}`.

### Assumed — not re-run in this revision

The 5,000-doc × 128-dim corpus behind the headline byte table and every
millisecond column. Those remain `e629a58`'s numbers. The mechanism is unchanged
and is pinned at 1/17th scale by the committed 300-doc test, which still reports
`emb_chunks.fst=0`, but this revision did not regenerate that corpus.

### Noted, deliberately untouched

A pre-existing quirk found while measuring: bare `bool{must_not:[match …]}` with
no `must`/`filter` answers **0** on a control index carrying no vector field at
all (`{text body, text tags, keyword kw}`), while
`bool{must:[match_all],must_not:[match …]}` answers 25 on the same index. Not
related to #328 and not changed here; the new test pins the `match_all`-anchored
form for exactly that reason.


xerj-org:
Reviewed by two independent agent passes (Claude Opus 5), the second tasked with refuting the first. It could not refute either blocker and found a third, larger one. Retrieval against an indexed `apache/lucene` checkout was used for the reference points below.

The premise is right and the reference agrees with it. Lucene's vector field types set only vector attributes and so never carry index options or a term dictionary (`KnnFloatVectorField.java:70-85`, `KnnByteVectorField.java:82-101`), which makes XERJ's `.emb.fst` pure waste. Folding the exclusion into the same walk that decides HNSW graph membership is the right centralisation. The problems below are all in scope and completeness, not direction.

## Blocking

**1. The bigger half of the bloat survives: `<field>_chunks` is not excluded for user-mapped `dense_vector`.**

This is the finding that matters most, and it inverts the headline number. `passage_scored_vector_fields` (`index.rs:28661-28673`) unions `collect_dense_vector_fields` with the semantic targets, so a plain user-mapped `dense_vector` gets first-class `<field>_chunks` max-sim passage scoring, pinned on main by `explicit_nested_dense_vector_chunks_activate_only_their_field_guard` (`index.rs:3466-3500`). But `semantic_derived_vector_fields` adds `{target}_chunks` only for fields carrying an `embedding` config (`memtable.rs:3805-3823`), and the new `lexically_typeless_fields` adds only the vector paths themselves.

So `passages.vec_chunks` is in neither set. The pruned parent blob keeps it (prune drops only `passages.vec`), so `<seg>.passages.fst` still carries every chunk float, and the recursion additionally emits `passages.vec_chunks` as its own field, building `<seg>.passages.vec_chunks.fst` too.

By RFC #148's own numbers this is the dominant half: `body_vector.fst` 486 MiB against `body_vector_chunks.fst` 1.852 GiB. The 55.5% headline is measured on a fixture with no chunks, so it is not the saving a passage-vector user gets.

**2. Every explicitly-fielded lexical query on a `dense_vector` becomes a full stored-doc scan.**

`unsearchable_query_field` (`index.rs:29955-29959`) only rejects `!indexed && !doc_values`, and `FieldOptions::indexed` defaults to `true`, so a vector clause is planned and executed. `collect_fts_query_fields` then asks for `field_stats("emb")`, now `None` on every segment this release writes, so `fts_has_field` is false and `scan_stored` (`index.rs:16224`) walks and JSON-parses the entire stored section. `prune_missing_should_fields` does not rescue it — it bails on any `must`/`must_not`/`min_should_match` or non-Term clause.

The same file already records this fallback at **0.399 ms → 26.412 ms on 20,000 docs**, a 66x regression (`index.rs:36270-36276`). The PR's tests run on 300 docs, so it is invisible there; 28 answers were measured and zero latencies.

Lucene's answer to "this field has no terms" is `Terms.EMPTY` producing an empty scorer, never a corpus walk (`Terms.java:40-48`, `IndexSearcher.java:1144-1165`). Since `fts_excluded_fields` is schema-derived and known at plan time, a clause naming a lexically-typeless field should lower to MatchNone. That is the same place the `*`-expansion fix already went.

**3. `multi_match` on a vector field changes its hit set, from 0 to every document.**

Filed in the PR as a perf-only side effect; it is not. Dropping `Vector` from `exact_fields` makes `multi_match_field_is_mapped("emb", ..)` false (`index.rs:36122-36132`), `field_specs` empties, `query_node_to_fts` returns `None` (`index.rs:36511-36513`), and `MultiMatch` is unconditionally in `is_doc_scan_query` (`index.rs:31263`), so the stored scan answers. `doc_matches_query`'s MultiMatch arm joins the float array via `to_string()` and its OR branch splits on non-alphanumerics and tests token equality (`index.rs:32298-32311`, `:32400-32404`). Every rendered float contributes a `0` token, so any query containing a `0` token admits **every** document.

On main `emb` was in `exact_fields`, lowered to a whole-value `Term`, `field_stats` was present, and the scan never ran: 0 hits. This also falsifies "26 of 28 answers byte-identical" for the probe named `multi_match ... explicitly-fielded`.

## Correction to note

Array-of-objects nesting (`{"passages":[{"text":..,"vec":[..]}]}`) is genuinely unhandled — `collect_text_fields`'s `Value::Array` arm maps the unpruned `extract_text_value` and never recurses (`memtable.rs:3721-3730`), and the merge path gates on `val.is_object()` (`index.rs:19151`). But it is a marginal configuration, not RFC #148's shape: `get_field_value_parts` flattens across array elements (`index.rs:5733-5750`), so a multi-passage array yields a dims x N vector that fails the per-chunk length check and gets no usable kNN anyway. RFC #148's `body_vector_chunks.fst` was semantic-generated and already excluded by #12. Worth fixing, but it is not the load-bearing gap — (1) is.

## Smaller

- #328 asked for a test pinning that `_field_caps` and highlighting do not read a vector field's `.fst`. Neither is pinned; only `exists` and kNN are.
- Nothing tells users that existing indices keep their bloat until an explicit `force_merge`, and that pre-flush `match` answers on a `dense_vector` change. CHANGELOG untouched.
- Numeric `term`/`terms`/`range` on a vector field are asserted unchanged in the body but deliberately left out of the committed test, which is the behaviour most likely to drift once the postings are gone.

## Clear on the #352 axis

Touches no `Cargo.toml`, `Cargo.lock` or build script. #258 and #322 were stacked on top of this merge in a scratch worktree and both auto-merged with zero conflicts; the engine hunks do not overlap and the `integration.rs` additions append at EOF.

Fix (1) and (2) and this is a strong change. (3) needs a decision on whether that hit-set change is acceptable or needs the MatchNone lowering from (2), which would resolve both.

PR #386, opened 2026-08-15
fix(engine,vector): #371 scalar8 kNN scored updated docs from stale codes

### LINKED ISSUE #371: dense_vector quantization:scalar8 / int8_hnsw returns WRONG RESULTS: the SQ8 code cache is write-once, so updated vectors are scored forever from stale codes
Found by an agent-driven reference-coded architecture review (Claude Opus 5), driven by @xerj-org, and independently re-verified by hand before filing. **This returns wrong results, silently, through a documented ES mapping option.**

## Summary

A `dense_vector` field mapped `quantization: "scalar8"` (or its alias `index_options: {"type": "int8_hnsw"}`) caches each document's quantized codes on the **first query that observes it** and never refreshes them. Update the document's vector and it is scored forever from the stale code.

## Reproduction

Two indices, byte-identical data, 8-dim cosine `dense_vector`. The only difference is `"quantization":"scalar8"` on the first.

```sh
# index 20 docs; doc 0 = [1,0,0,0,0,0,0,0]
# 1. query once, to populate the code cache
knn query_vector=[1,0,...]
  bench-sq8    0:1.000000  12:0.784375  9:0.762082
  bench-none   0:1.000000  12:0.783807  9:0.762747      # agree

# 2. overwrite doc 0 with its exact NEGATION -> true cosine -1, the worst possible
PUT /{index}/_doc/0  {"v":[-1,0,0,0,0,0,0,0]}
POST /{index}/_refresh
GET  /{index}/_doc/0 -> _source.v = [-1, 0, ...]        # both indices, confirmed

# 3. same query again
  bench-sq8    0:1.000000  12:0.784375  9:0.762082      # <-- doc 0 STILL #1, perfect score
  bench-none   12:0.783807  9:0.762747  14:0.757325     # <-- correct: doc 0 gone
```

The quantized index returns a document at cosine **1.000000** whose stored vector is the exact opposite of the query. The unquantized control, same manipulation, correctly drops it. Note the sq8 scores are byte-identical to step 1 — the ranking is frozen at first observation.

## Cause

`index.rs:10916` guards code population with `if !store.codes.contains_key(id)`. Write-once. Nothing invalidates the entry on update, delete-and-reindex, or merge, so the cache diverges from `_source` permanently.

## Why this is a release blocker rather than a performance note

- It is reachable through the ES-compatible mapping API, under the name `int8_hnsw` that Elasticsearch users will reach for by default.
- It is silent. There is no error, no warning, and the score looks perfect — `1.000000` is the most convincing possible wrong answer.
- Every realistic vector workload updates vectors: re-embedding after a model change, correcting a document, any upsert pipeline. Each update permanently poisons that document's ranking.
- It compounds with the fact that the same mapping option also disables ANN entirely (filed separately), so users who enable it are paying 43x latency *for* wrong results.

## Reference

Lucene addresses quantized vectors by **ordinal**, in a flat array written at index time, not by a lazily-populated map keyed on document id — so there is no cache to go stale; the codes live with the segment and are rewritten when the segment is. See the ordinal-addressed access pattern in the HNSW/quantized vector readers.

## Suggested fix

Address codes by the same dense slot the slab already uses (`SlabGraphStorage::slot_of`, `xerj-vector/src/hnsw.rs:154`) in a flat `Vec<u8>` at `slot*dim..`, populated **at ingest** and rewritten on update — not lazily under a query-path write lock. Per the review that fixes the staleness, an unbounded-growth leak, the per-query exclusive lock, and a `_source` walk in one change.

At minimum, until that lands: invalidate the code entry whenever a document version changes.

**Verified:** the repro above was run by hand on `v1.0.0-rc.16` (release binary, `aarch64-unknown-linux-gnu`) against a live node, after the review agent independently reproduced the same behaviour on a 5000 x 384 corpus with a cache-busting query vector. Both `_source` bodies confirmed as negated before re-querying.
**Assumed:** I did not test whether a force-merge, node restart, or index close/open clears the cache — plausible workarounds that would narrow the blast radius but not the defect. I did not check whether `_update` and `_bulk` update paths behave differently from `PUT /_doc/{id}`.

xerj-org:
Prepared by an agent (Claude Opus 5) driven by @xerj-org. **Body rewritten at head `30d0763` — the previous body described round 1 and was false about this branch. That was my error, not the fix's.**

Closes #371 — a correctness defect shipped in v1.0.0-rc.17.

## The defect

A `dense_vector` field mapped `quantization: "scalar8"` (or its alias `index_options: {"type": "int8_hnsw"}`) scored updated vectors from stale quantization state. Reproduced by hand before any code was written:

```
1. index 20 docs, doc 0 = [1,0,0,0,0,0,0,0]; knn [1,0,...] -> doc 0 first at 1.000000
2. PUT doc 0 = [-1,0,0,0,0,0,0,0]   (exact negation; true cosine -1)
3. _refresh; GET /_doc/0 confirms _source is negated
4. same query:  scalar8 -> doc 0 STILL first at 1.000000, scores byte-identical to step 1
                control -> doc 0 correctly gone
```

## Why round 1 was rejected

Round 1 removed the **per-document code cache** and left the **fitted codebook** stale, so the headline symptom still reproduced with output identical to unfixed `main`. On the workload the issue names as what makes it a blocker — re-embedding after a model change — it also destroyed ranking discrimination entirely. That round was correctly refused.

This head fixes both halves: `Index::sq8_stores` and `Sq8FieldStore` are deleted, and `grep` confirms **no SQ8 state outlives a query**.

## Verified (independent adversarial pass, commands re-run by the verifier)

- **Fail-before is genuine and non-vacuous.** Reverting ONLY `engine/crates/xerj-engine/src/index.rs` makes all four sq8 tests fail with output byte-identical to the report — `[("0", 1.0), ("1", 0.99999976), …]`, 16 distinct scores across 50 docs. Restoring makes them pass in **both** dev and ci-test profiles.
- The tests are real HTTP-boundary tests against a full-precision **control** index asserted first, so a passing result cannot be an artifact of the harness.
- The issue's own repro was reproduced independently through the `int8_hnsw` alias, and doc 0 correctly leaves the top after negation.

## Assumed / known limitations

- **Filtered kNN on a quantized field depends on the filter**, and this changes the **returned order**, not merely the score. That is disclosed in the CHANGELOG, `docs/recipes/vector-quantization.md` and at `index.rs:10935-10941`, but every one of those understates it as a scoring difference. Being corrected before merge.
- `landing/docs/recipes/*.html` is a hand-maintained mirror of `docs/recipes/*.md` and still carries the pre-fix memory claim, which the CHANGELOG now contradicts. Being corrected before merge.

**Do not merge until those two are corrected and all 16 checks are green.** The code is verified; the prose about it is not yet.


xerj-org:
**DO NOT MERGE — adversarial verification rejected this round (8 must-fix).**

The fix is not wrong so much as incomplete in a way that leaves the reported bug live:

- **The headline symptom still reproduces on this branch, with output identical to unmodified `main`.** The per-document code cache was removed, but the **codebook** is still stale, so an updated vector is still scored against quantization parameters derived before the update.
- **On the workload #371 names as what makes it a release blocker — re-embedding after a model change — this branch destroys ranking discrimination entirely** (measured on a 50-doc probe).
- The commit and CHANGELOG claim "staleness is now structurally impossible" and "the score always descends"; neither holds.
- The new regression test is constructed around the case that still fails, so it passes without covering the defect.
- `docs/recipes/vector-quantization.md` ("4x smaller vectors, same results") is now inaccurate.
- Two in-code comments at `index.rs:10896-10898` describe an object this commit deleted.
- The performance characterisation is a guess ("I believe it is net faster"); the verifier measured it.

What *did* hold up, and is worth keeping: the fail-before proof is real (reverting only `index.rs` while keeping the test reproduces the failure byte-for-byte, with the non-quantized control passing its post-update assertions so the failure is specific to `scalar8`), the test is non-vacuous and drives the real router end to end, and fmt/clippy are clean.

Round 2 is running against these items.

Issue #361, opened 2026-08-14
bool with any second clause (filter/must/must_not) collapses _score and reorders hits; a no-op filter flattens every score to a constant


xerj-org:
Found by an agent (Claude Opus 5) driven by @xerj-org while building a reference corpus. Every number below came from a command I ran against `v1.0.0-rc.16`.

## Summary

Adding **any** second clause to a `bool` query — `filter`, a second `must`, or `must_not` — collapses `_score` into a narrow band and **reorders the results**. Distinct scores merge into ties. A filter that matches every document flattens every score to a constant and returns unrelated documents.

`bool` + `must` + `filter` is the single most common query shape in the Elasticsearch ecosystem, and in ES `bool.filter` is non-scoring **by definition**. This is the query shape behind every faceted search, every "search within this category", and every Kibana/OSD panel that pairs a query bar with a filter pill.

## Repro

Index: 75,578 docs, one per code symbol, fields `name_text`/`doc` as `text`, `repo` as `keyword`.

Baseline, no second clause:

```json
{"query":{"multi_match":{"query":"select diverse neighbors when adding a node to the graph",
                         "fields":["name_text^3","doc^3"]}}}
```

Then the same `multi_match` with one extra clause added. Top 4 scores:

| query | top-4 `_score` | order |
|---|---|---|
| `multi_match` alone | 59.0076, 57.2995, 49.7646, 49.0195 | correct |
| `bool.must` + `bool.filter` `{"term":{"repo":"lucene"}}` | 1.6525, 1.4081, 1.4081, 1.1517 | **inverted** |
| `bool.must` + a second `must` clause | 1.6525, 1.4081, 1.4081, 1.1517 | **inverted** |
| `bool.must` + `must_not` `{"term":{"repo":"usearch"}}` | 1.4741, 1.2297, 1.2297, 0.9734 | **inverted** |
| same scoping via `post_filter` | 59.0076, 57.2995, 49.7646, 49.0195 | correct |

Three distinct failures visible in that table:

1. **`bool.filter` is scoring.** In ES it must not contribute to `_score` at all.
2. **Ranking inverts.** Baseline #3 (`addGraphNodeInternal`, 49.76) is promoted above baseline #1 (`fixDisconnectedNodes`, 59.01).
3. **Distinct scores collapse into ties.** 59.0076 and 57.2995 both become exactly 1.4081, so the tiebreak is now arbitrary.

## Worst case: a no-op filter

[...]

PR #387, opened 2026-08-15
fix(engine): bool.filter and must_not are non-scoring; stop the page-local rescore overwriting exact BM25 (#361)

xerj-org:

Part of #361.

Adding ANY second clause to a `bool` — `filter`, a second `must`, or
`must_not` — collapsed `_score` into a narrow band and reordered the page,
with no error and no warning. Three separate defects, all of the same
shape: a clause or a pass that must not affect scoring, did.

1. `bool.filter` was SCORING on the exact FTS path.
   `query_node_to_fts_with_keyword_fields` projected every `filter` child
   onto the FTS bool's `must` slot, and `execute_bool` sums every `must`
   into `score_map`, so a filter's BM25 landed in `_score`. Lucene draws
   the split in `BooleanClause.isScoring()` — `occur == MUST ||
   occur == SHOULD`
   (lucene/core/src/java/org/apache/lucene/search/BooleanClause.java:84-87)
   — and enforces it structurally in `BooleanWeight`'s ctor, which builds a
   FILTER clause's `Weight` with `ScoreMode.COMPLETE_NO_SCORES`
   (.../search/BooleanWeight.java:49-64): a filter is incapable of
   contributing a score, rather than contributing one that is later
   subtracted. `BoolQuery` now carries its own non-scoring `filter` slot;
   `execute_bool` intersects it into the candidate set and never touches
   `score_map`. `min_should_match` counts it as a required clause so
   `bool{filter, should}` keeps its shoulds optional, as in ES.

2. The heuristic "IDF-weighted rescore for Bool queries with multiple
   terms" treated `filter`/`must_not` as scoring too: `query_uses_bool_text`
   counted them toward its >=2-text-clause trigger and
   `collect_match_field_terms` folded their terms into the weighting. Both
   now walk `must`/`should` only — the same `isScoring()` split.

3. That rescore derived its IDF from `final_hits.len()` — the RETURNED PAGE
   — so `_score` was a function of `size`. Lucene never rescores a
   collected page: `TermQuery.createWeight` builds `TermStates` from
   `searcher.getTopReaderContext()`, the whole index, and only when scores
   are needed (.../search/TermQuery.java:298-312); `BM25Similarity.idfExplain`
   takes `df` from `termStats.docFreq()` and `N` from `fieldStats.docCount()`
   (.../search/similarities/BM25Similarity.java:177-197); and
   `BooleanWeight.scorerSupplier` joins required-but-non-scoring clauses to
   the conjunction as filters with no post-collection stage at all
   (.../search/BooleanWeight.java:289-341). The rescore is now skipped when
   the whole page came from the segment FTS path, which already produces
   exact ES BM25 against index-wide statistics (#188) — the identical
   protection `!scored_fast_applied` gives the columnar scored path. Pages

---

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -15280,6 +15280,23 @@ impl Index {
         // Set when scored_columnar already dropped the first `from` candidates
         // (page-offset hydration) — the final pagination then skips 0.
         let mut scored_from_applied = false;
+        // ── #361: request-level companion to `scored_fast_applied` ────────
+        // The segment FTS path (`fts_handled` below) also produces EXACT ES
+        // BM25, scored against index-wide statistics (#188), and the
+        // page-local IDF rescore further down must not overwrite it — that
+        // rescore derives its N from `final_hits.len()`, i.e. the RETURNED
+        // PAGE, so leaving it armed makes `_score` a function of `size`.
+        // `fts_handled` is a per-segment local, so accumulate here.
+        //
+        // The rescore is only skipped when EVERY arm that could contribute a
+        // hit was exact. `heuristic_scored_applied` starts true whenever the
+        // memtable holds documents — unflushed hits carry the IDF-less
+        // memtable/stored-scan score, and putting the two populations on one
+        // scale is precisely what the rescore was papering over until the
+        // score unification lands (#354 / #188). Mixed pages therefore keep
+        // today's behaviour; a fully-flushed index gets exact BM25.
+        let mut fts_scored_applied = false;
+        let mut heuristic_scored_applied = mem_doc_count > 0;
         if let Some(plan) = &scored_plan {
             if scored_fast_ready {
                 if let Some((hits, total, from_applied)) = self.scored_columnar(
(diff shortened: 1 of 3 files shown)


xerj-org:
Post-merge note, found while rebasing a sibling branch onto this.

**This PR changed `_score` on the wire and added no CHANGELOG entry.** The diff is `engine/crates/xerj-engine/src/index.rs` (+120/-43), `engine/crates/xerj-fts/src/search.rs` (+138), and a new `bool_filter_is_non_scoring.rs` (+256) — no `CHANGELOG.md` hunk at all.

That matters more than usual here because of what changed: `bool.filter` and `must_not` no longer contribute to `_score`, and the page-local IDF rescore is gone. Any client that stored `_score` values, thresholded on them, or fused them (RRF/hybrid) sees different numbers after upgrading, and there is currently nothing in the release notes telling them so.

The repo's own standard requires this — every other PR merged today carries its behaviour changes in `## [Unreleased]`, and PR #258 was held for a round specifically over "wire-visible changes with no CHANGELOG entry".

Not proposing a revert: the fix is correct and main is building. But `## [Unreleased]` should gain an entry before the next RC is cut, describing (a) that `bool.filter`/`must_not` are now non-scoring, matching Lucene's `BooleanClause.isScoring()`, and (b) that `_score` no longer varies with `size`, which is the actual user-visible fix for #361.

Also worth stating plainly in that entry: this is a **breaking change to `_score` values**, not just a bug fix. The old numbers were wrong, but they were what clients saw.

Issue #362, opened 2026-08-14
term normalises case but prefix/wildcard do not: a doc findable by term returns 0 hits for the matching prefix


xerj-org:
Found by an agent (Claude Opus 5) driven by @xerj-org. All counts below are from real responses on `v1.0.0-rc.16`.

## Summary

On the same `text` field in the same index, `term` normalises the query term's case but `prefix` and `wildcard` do not. So a document you can find with `term` cannot be found with the matching `prefix`, and the failure mode is an empty result set with no error.

## Repro

Index of 11,450 source files, `title` mapped as `text` (e.g. `TestSegmentReader.java`, `HnswGraphBuilder.java`).

| query | count |
|---|---|
| `{"term":{"title":"TestSegmentReader.java"}}` | 1 |
| `{"term":{"title":"testsegmentreader.java"}}` | 1 |
| `{"prefix":{"title":"Test"}}` | **0** |
| `{"prefix":{"title":"test"}}` | 1735 |
| `{"wildcard":{"title":"Hnsw*"}}` | **0** |
| `{"wildcard":{"title":"hnsw*"}}` | 12 |

`term` matches either case, so it is analysing/normalising the query. `prefix` and `wildcard` match only lowercase, so they are comparing against the indexed (lowercased) terms without normalising the query.

The two are therefore inconsistent with each other, and `term` is also inconsistent with Elasticsearch, where `term` is explicitly **not** analysed.

## Why it bites hard here

On any CamelCase corpus — Java, C++, Rust, Go, C# source, or any dataset with proper nouns — the natural prefix a person or an agent types is capitalised. Every one of those returns zero hits while the data is sitting right there. Nothing signals an error, so the caller concludes the data is absent.

Concretely, this is the sequence that cost me time:

```
term  title=HnswGraphBuilder.java   -> 1     (file is in the index)
prefix title=Hnsw                   -> 0     (same file, same field)
```

## Suggested fix

Pick one rule and apply it to all term-level queries on a given field. Matching ES means: `term` compares the raw term against the index without analysis, and `prefix`/`wildcard` do the same, with `match`/`match_phrase` as the analysed path. Whatever is chosen, the three should not disagree.

## Workaround

Use `term` for exact identifiers and `multi_match` on a `text` field for partial ones. Avoid `prefix`/`wildcard` on identifiers entirely, or lowercase the query term first and accept that you cannot express a case-sensitive prefix.

[...]

PR #388, opened 2026-08-15
fix(engine): `_count` stops answering from an oracle `_search` never uses; `case_insensitive` is honoured on prefix/wildcard (#362)

xerj-org:

Part of #362.

The report was "`term` normalises case but `prefix`/`wildcard` do not": on a
`text` field, `term title=TestSegmentReader.java` finds a document that
`prefix title=Test` cannot see. Reproduced verbatim on ca4d75a. The attribution
is backwards, and the truth is worse than the report.

`prefix`/`wildcard` are the ES-CORRECT half. A multi-term query is not
analysed: Lucene builds the automaton straight off the query bytes
(`PrefixQuery.toAutomaton`, lucene/core/src/java/org/apache/lucene/search/
PrefixQuery.java:44), and case handling happens one layer up, in a code path
deliberately separate from indexing analysis — `QueryParserBase.getPrefixQuery`
(lucene/queryparser/src/java/org/apache/lucene/queryparser/classic/
QueryParserBase.java:750) calls `Analyzer.normalize`, which
(lucene/core/src/java/org/apache/lucene/analysis/Analyzer.java:213) documents
why: normalize "without tokenizing or stemming, which are undesirable if the
string to analyze is a partial word". So an uppercase prefix against a
lowercased text dictionary matching nothing is the correct answer.

`term` was not normalising at all. What normalised was `_count`.

`try_shortcut_count`'s FTS fallback looked the term up in the segment's term
dictionary and, on a miss, RETRIED with a lowercased spelling the caller never
wrote. That is what made `term` look case-insensitive. But the dictionary is
the wrong oracle in the first place: a field reaches that fallback precisely
because it has no doc-values column — a `text` field — and `_search` resolves a
`term` on a text field through `doc_matches_query`, comparing against the whole
`_source` value, not against analyzed terms. Measured on ca4d75a, both of these
returned a count no `_search` could reproduce:

    {"term":{"title":"TestSegmentReader.java"}}         count 1, hits 0
    {"term":{"title":"quick"}} / "the quick brown fox"  count 1, hits 0

The second has no case component at all — the retry is a no-op there and the
dictionary hit is genuine — which is the proof that deleting the retry alone
would only flip the error's sign (count 0, hits 1), not fix it. So the shortcut
now ABANDONS when the field has no column, and the real scan answers. Lucene
has no equivalent hazard because `IndexSearcher.count`
(lucene/core/src/java/org/apache/lucene/search/IndexSearcher.java:495) derives
its answer from the same rewritten query and Weight as `search()`, taking only
algebraic rewrites as shortcuts and otherwise collecting for real;
`QueryUtils.checkCount` (lucene/test-framework/src/java/org/apache/lucene/

---

### DIFF
--- a/engine/crates/xerj-api/tests/multi_term_case_rule_is_one_rule.rs
+++ b/engine/crates/xerj-api/tests/multi_term_case_rule_is_one_rule.rs
@@ -1,0 +1,286 @@
+//! The term-level family obeys ONE case rule, and `_count` agrees with
+//! `_search` — #362, from the outside.
+//!
+//! The report was "`term` normalises case but `prefix`/`wildcard` do not": a
+//! document you can find with `term title=TestSegmentReader.java` is invisible
+//! to `prefix title=Test`. The attribution turned out to be backwards in a way
+//! that made it worse. `prefix`/`wildcard` are the ES-correct half — a
+//! multi-term query is not analysed, Lucene builds the automaton straight off
+//! the query bytes (`PrefixQuery.toAutomaton`, PrefixQuery.java:44) — and
+//! `term` was not normalising at all. What normalised was the `_count`
+//! shortcut, which retried the FTS term-dictionary lookup with a lowercased
+//! spelling of a term the caller never wrote, while the hit-materialising path
+//! stayed on the raw one. So `_count` reported a document that `_search` could
+//! not produce, and it did so for every `term` on a `text` field, cased or not.
+//!
+//! Two invariants are pinned here.
+//!
+//! 1. **count == hits**, Lucene's own `QueryUtils.checkCount`
+//!    (QueryUtils.java:680) invariant, asserted across the whole term-level
+//!    family in both cases on both a `text` and a `keyword` field. Lucene gets
+//!    this for free because `IndexSearcher.count` (IndexSearcher.java:495)
+//!    derives its answer from the SAME rewritten query and Weight `search()`
+//!    uses; its only shortcuts are algebraic rewrites, never a re-spelling of
+//!    the term. A count no `_search` can reproduce is worse than a slow one.
+//!
+//! 2. **`case_insensitive` is honoured**, not accepted and dropped. It is the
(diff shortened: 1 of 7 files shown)


xerj-org:
Closing this — superseded, and the half of it that was right is already on `main`.

## Where its two halves went

**The `_count` oracle half landed.** #417 merged `eaeab4f` — *"`_count` stops answering a `term` on a text field from the analysed term dictionary"* — and I verified it on a node built from `main`: `_count` and `_search` now agree on all three probes that previously diverged, including `term title=testsegmentreader.java`, which was `_count=1 / _search=0`.

**The `case_insensitive` half is superseded by #423.** This branch patched case-folding at the operator level — `case_insensitive` honoured on `prefix`/`wildcard`. That treats a symptom. The cause is that `doc_matches_query` (`crates/xerj-engine/src/index.rs:33213`) matches buffered documents against raw `_source` with **no mapping and no analyser**, while flushed documents go through the real term dictionary. Two matchers, one query, and a background flush decides which one answers you.

That is why #362 was closed into #423 along with seven sibling issues (#396, #397, #398, #406, #408, #409, #413): they are one defect seen from eight angles, and eight operator-level patches to one schema-less evaluator is what produced the churn — #410's fix for #394 reintroduced #394 through a sibling path, and this branch's own verification flagged that it "introduces a NEW hit-set divergence of exactly the class #362 is about."

## Why not rebase it instead

It is DIRTY against a `main` that has moved a long way (`_count` fixed, #424's scroll work, three external contributions), its verification returned `sound=false`, and the design it implements is the one #423 argues against. Rebasing would produce a conflict-resolved patch that still encodes the wrong model. The work is not lost: the repro and the `case_insensitive` behaviour it pins are recorded in #423's acceptance criteria, and `count_matches_search_on_text_term.rs` — the test shape this branch introduced — landed with #417.

Reopen if you think the operator-level fix stands on its own; the argument against it is in #423 and I would rather be shown wrong there than merge both.

Issue #363, opened 2026-08-14
match on a semantic_text field silently runs BM25; only the semantic query does kNN


xerj-org:
Found by an agent (Claude Opus 5) driven by @xerj-org, on `v1.0.0-rc.16`.

## Summary

A `match` query against a `semantic_text` field runs BM25 and returns lexical results. Only `{"semantic":{...}}` performs the vector search. There is no warning, so a user who builds a semantic index and queries it the obvious way gets none of the semantics they paid the indexing cost for.

## Repro

Two indexes over the identical 321 documents. `symprobe` is BM25 only; `symprobe_sem` has the same content plus a `ctx` field mapped `semantic_text`, embedded with `--embed-mode neural`.

Query, deliberately paraphrased so it shares no content words with the target code:

> `prune redundant edges so the graph does not become clustered`

| index | query | top 3 |
|---|---|---|
| `symprobe` (BM25) | `match` on `code` | `trySetNewEntryNode`, `peek`, `GraphBuilderKnnCollector` |
| `symprobe_sem` | **`match`** on `ctx` | `trySetNewEntryNode`, `peek`, `GraphBuilderKnnCollector` |
| `symprobe_sem` | **`semantic`** on `ctx` | `MergingHnswGraphBuilder`, `fixDisconnectedNodes`, `InitializedHnswGraphBuilder` |

The `match`-on-`semantic_text` row is identical to the pure-BM25 row, right down to the ordering. Only the `semantic` row finds the graph-construction code the paraphrase describes. Scores also differ by an order of magnitude (10.65 vs 0.77), consistent with one being BM25 and the other cosine.

## Cost of the trap

The semantic index is not free. Measured on the same 321 documents, same machine:

- BM25 only: **1.7s**
- with `semantic_text` + neural embedder: **20.7s**

So roughly 12x the indexing time, and a `match` query silently gives none of the benefit.

## Suggested fix

Either make `match` on a `semantic_text` field do the semantic thing (which is what ES 8.15+ does — `match` on `semantic_text` is rewritten to the semantic query), or return a warning in the response when a lexical query targets a `semantic_text` field. Silently doing the lexical thing is the one option that cannot be debugged from the outside.

**Verified:** both indexes, all three result sets and both timings are from real runs on `v1.0.0-rc.16`, `aarch64-unknown-linux-gnu`, `--embed-mode neural`.
[...]

PR #389, opened 2026-08-15
fix(api): warn when a full-text query targets a semantic_text field (#363)

xerj-org:

Part of #363.

A `match` against a `semantic_text` field runs BM25 over that field's lexical
index and returns lexical hits. Only `{"semantic":{...}}` reaches the
embeddings. Nothing said so, which is what made #363 undebuggable from the
outside: the reported repro returns a result list byte-identical to the pure
BM25 index, right down to the ordering, after paying ~12x the indexing time
(1.7s -> 20.7s on the same 321 documents) to make the semantic answer
reachable at all.

Root cause: `semantic_text` maps to `FieldType::Text` plus an
`EmbeddingConfig` (es_compat.rs), so the inverted index is built for it
deliberately. The scoring path never consults `fc.embedding`; the engine's
only semantic dispatch is a whole-query short-circuit on
`QueryNode::SemanticSearch` (index.rs `peel_semantic_query`). A `match` on a
semantic_text field is therefore byte-identical to a `match` on plain text.

This does NOT rewrite `match` to the semantic query, which is what the issue
proposes. Three blockers, to be filed separately: the semantic dispatch is a
top-level short-circuit with no vector sub-executor inside `bool`, so a
rewrite would apply at top level only and silently NOT apply inside
`bool.must` — a strictly worse trap than today's; it breaks
`test_semantic_text_match_survives_flush` (integration.rs), a deliberate,
previously-blocking tested behaviour; and it is a silent result-changing wire
break during an rc. The lexical index on these fields is also what
`{"hybrid":{...}}` fuses.

Instead, the existing `_xerj.hints` channel carries a
`lexical_on_semantic_text` entry naming the field, what actually ran, and the
remedy. Engine behaviour is unchanged — no engine crate is touched.

Reference coding (adapted, not copied; Lucene is Apache-2.0, same as XERJ):
Lucene's uniform rule is that when a query's assumption about how a field was
indexed does not match reality, it says so and never silently answers a
different question. The message shape here — name the field, name what was
actually indexed, name the remedy — is taken from:
  lucene/core/src/java/org/apache/lucene/index/DocValues.java:206
    checkField: "unexpected docvalues type ACTUAL for field 'X'
    (expected=...). Re-index with correct docvalues type."
  lucene/core/src/java/org/apache/lucene/index/FloatVectorValues.java:53
    checkField: "Unexpected vector encoding (...) for field X (expected=...)"
  lucene/core/src/java/org/apache/lucene/index/ByteVectorValues.java:53

---

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -7162,6 +7162,197 @@ fn unknown_field_hint(
     }))
 }
 
+/// Field names a full-text, BM25-scored clause searches.
+///
+/// Deliberately narrower than [`referenced_query_fields`]: only the analysed
+/// clause kinds, the ones whose whole point is "find me the text I mean". A
+/// `term`/`range`/`exists` against a `semantic_text` field is an ordinary
+/// filter and nobody expects embeddings out of it, so reporting those would
+/// be exactly the noise that gets a hint channel ignored wholesale.
+fn lexical_text_query_fields(q: &Value, out: &mut Vec<String>) {
+    /// `{"match": {"<field>": ...}}` — the field is the inner key.
+    const FIELD_KEYED: &[&str] = &[
+        "match",
+        "match_phrase",
+        "match_phrase_prefix",
+        "match_bool_prefix",
+    ];
+    /// `{"multi_match": {"fields": [...]}}` — the fields are listed inside.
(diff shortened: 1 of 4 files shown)


xerj-org:
Superseded by #393, which merged the same fix from `fix/issue-363-round1`. See the post-merge note on #393 for what was verified and what was not.

Issue #360, opened 2026-08-14
autoindex aborts a whole run when one file declares 2+ SQL tables (sealed count attributed to the first dataset)


xerj-org:
Found by an agent (Claude Opus 5) driven by @xerj-org while indexing reference corpora for `tools/xerj-code`. Everything below is a command I ran, with its real output.

## Summary

`xerj autoindex` exits `1` and abandons a run when a single file declares two or more SQL tables. The rows index correctly; only the count reconciliation is wrong, and it is fatal.

Hit on the real, unmodified `unum-cloud/usearch` repo (it is in `hub/xerj-vector.json`), via `usearch/sqlite/README.md`. That is ordinary prose documentation with three ```sql blocks in it. Nothing is malformed.

```
error: dataset sqlite-binary-vectors exact read-back count 2 disagrees with sealed count 4
xerj-done ok=false exit=1 reason=aborted
```

## Minimal repro

One markdown file, no server state needed beyond a running node:

````sh
mkdir -p /tmp/r && cat > /tmp/r/t.md <<'EOF'
```sql
CREATE TABLE t1 (id INTEGER PRIMARY KEY, v JSON NOT NULL);
INSERT INTO t1 (id, v) VALUES (10, '[1.0]'), (11, '[2.0]');
CREATE TABLE t2 (id INTEGER PRIMARY KEY, v JSON NOT NULL);
INSERT INTO t2 (id, v) VALUES (20, '[1.0]'), (21, '[2.0]');
```
EOF
xerj autoindex /tmp/r --prefix r --no-semantic --no-graph --max-minutes 0
````

## Measured, deterministic (3/3 runs each)

| `CREATE TABLE` blocks in ONE file | sealed | read-back | result |
|---|---|---|---|
| 1 | 2 | 2 | `ok=true exit=0` |
| 2 | 4 | 2 | **abort** — `dataset t1 ... sealed count 4` |
| 3 | 6 | 2 | **abort** — `dataset t1 ... sealed count 6` |

## Diagnosis

`sealed count` is the total number of extracted rows across **all** tables in the file (2 rows x N tables), but it is attributed to the **first** dataset (`t1`) and then reconciled against **that one dataset's** exact read-back (2). Each extra table in the same file inflates the first dataset's expected count by its own row count, so a mismatch is guaranteed.

I ruled these out by testing each one:

- primary-key collision across tables — distinct PKs (44,45 vs 42,43) still abort
[...]

Issue #372, opened 2026-08-14
two xerj-engine lib tests fail on clean main at full test parallelism; CI's --test-threads=2 hides it


xerj-org:
Found by an agent-driven fix-verification pass (Claude Opus 5) driven by @xerj-org, then reproduced by hand on a clean checkout.

## Summary

Two `xerj-engine` lib tests fail on clean `origin/main` (`ca4d75a`) when the suite runs at full parallelism. Both pass in isolation. CI does not see it because `Build + Test` pins `--test-threads=2`.

## Reproduction

```
$ git checkout origin/main && cd engine
$ cargo test -p xerj-engine --lib

thread 'index::merge_publication_transaction_tests::partial_multi_batch_failure_is_reported_without_losing_either_batch'
  panicked at crates/xerj-engine/src/index.rs:2650:9:
  assertion `left == right` failed
    left: 3
   right: 4

failures:
    cluster_state::tests::every_representative_block_class_logs_kind_detail_and_operator_path_at_boot
    index::merge_publication_transaction_tests::partial_multi_batch_failure_is_reported_without_losing_either_batch

test result: FAILED. 546 passed; 2 failed; 0 ignored; 0 measured; finished in 21.11s
```

Same commit, same machine, module in isolation:

```
$ cargo test -p xerj-engine --lib merge_publication_transaction_tests
test index::merge_publication_transaction_tests::partial_multi_batch_failure_is_reported_without_losing_either_batch ... ok
test result: ok. 9 passed; 0 failed; 539 filtered out
```

## Why CI is blind to it

`.github/workflows/ci.yml` runs `cargo test --profile ci-test --workspace -- --test-threads=2`. Two threads serialises enough of the suite to hide whatever the interference is. The documented developer command — plain `cargo test` — uses one thread per core, which on a 20-core box is 20.

This is the third instance today of the same shape: a defect that is real on a clean checkout but structurally invisible to CI because CI runs a different configuration. The others were the clippy lints that only fire on Rust 1.92 (#359, merged) and the Painless stack overflow that only reproduces off the `ci-test` profile (#353).

## Why it matters beyond the two tests

[...]

PR #390, opened 2026-08-15
test(autoindex): pin #360 on the shape it was reported against, and log it

xerj-org:

Part of #360.

Round-1 follow-up on the #360 fix, rebased onto rc.17 (`9681034`). The fix
itself is unchanged; this adds the one corpus shape its tests did not cover
and the changelog entry it was missing.

Reproduction on current `main` before touching anything, release-path
binary (`cargo build --bin xerj --no-default-features`) against a local
node, using the issue's own repro verbatim:

    $ xerj autoindex /tmp/r --url http://127.0.0.1:19260 --prefix r2 \
        --no-semantic --no-graph --max-minutes 0
    xerj-done ok=false exit=1 reason=aborted wall=2.0s
    error: dataset t1 exact read-back count 2 disagrees with sealed count 4

and the one-table control on the same binary exits 0, matching the issue's
table exactly.

Why another test. The two committed e2e cases reach the fan-out through a
file *extension* — `dump.sql` and `shop.sqlite`. The reporter's corpus
contained neither: the file was `usearch/sqlite/README.md`, ordinary prose,
and it reached `sqldump` through the content sniffer
(`sniff.rs:578-582` — any text containing `CREATE TABLE` and a `;`). That
is a different route into the same defect and the one every affected user
actually hit, so it is pinned separately:
`a_readme_documenting_two_tables_is_not_a_fatal_condition` walks a corpus
of exactly one markdown file and asserts exit 0, `records_total` 4, one
committed generation, and `record_count` 2 on each of the two dataset
catalog documents.

Fail-before, with only the `exact_dataset_catalog_stats` fold reverted to
summing `group.expected_records` and the test kept:

    test ..._a_readme_documenting_two_tables_is_not_a_fatal_condition ... FAILED
    called `Result::unwrap()` on an `Err` value: dataset t1 exact
    read-back count 2 disagrees with sealed count 4
    test result: FAILED. 0 passed; 1 failed; 639 filtered out

Restoring the fold: `test result: ok. 1 passed`.

Verified (commands run, output seen) on this tree:

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,37 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
-_Nothing yet._
+### Fixed
+
+- **`xerj autoindex` aborted a whole run when one file declared two or more SQL
+  tables** ([#360](https://github.com/xerj-org/xerj/issues/360)). One file can
+  feed several datasets — a SQL dump is one file and N tables — but the sealed
+  record count was a single whole-file total. Finalisation charged that total to
+  *each* of the file's datasets in turn and compared it against that one
+  dataset's read-back, so any multi-table file disagreed with itself by
+  construction and exited `1` with `dataset t1 exact read-back count 2 disagrees
+  with sealed count 4`. The rows were always indexed correctly; only the
+  reconciliation arithmetic was wrong, so `exit=1` discarded a complete corpus.
+  Reported against a real `unum-cloud/usearch` file — `sqlite/README.md`,
+  ordinary prose with three ` ```sql ` blocks in it, which the sniffer routes to
+  the `sqldump` family.
+
+  The sealed ledger is now keyed by the dataset identity each record was written
+  under (`records_by_dataset` on the prepared artifact,
+  `expected_records_by_dataset` on the manifest group), the read-back is scoped
+  by `ax_dataset` as well as `ax_file`, and `validate_groups` rejects a ledger
+  that names a dataset the group does not feed or that does not sum to the group
+  total. Measured against a local node with the fixed binary: the two-table and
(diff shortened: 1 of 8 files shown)


xerj-org:
**Adversarial verification: PASSED on the code.** Verifier's words: *"the code is sound and I would stake the release on it. Every must_fix item above is prose/paperwork, not code. Nothing in the diff needs to change."*

It verified with **three separately built binaries, distinct by md5**, each in a private `CARGO_TARGET_DIR` and against a node on its own port and data dir:

- `xerj-main9681034` (xerj-autoindex checked out at current main): the 2-table README repro exits 1 with `error: dataset t1 exact read-back count 2 disagrees with sealed count 4` — byte-identical to the issue. The 1-table control on the same binary exits 0.
- the fixed binary: repro passes.

Branch is a clean fast-forward on `9681034` (`merge-base --is-ancestor` exits 0, 2 commits ahead).

## Corrections to the record, made here rather than by rewriting history

Three claims in the report and commit body did not survive checking. They do not affect the diff, and amending would cost a full CI cycle on an otherwise-green PR, so they are corrected here where they stay attached to the change:

1. **"The old SHA is still reachable as branch `verify-360`"** — it is not. `git ls-remote --heads origin verify-360` returns nothing and `git branch -r --contains 4f831c3` is empty.
2. **"633 passed, 2 failed — SAME TWO as the clean-main baseline"** did not reproduce for the verifier. Treat the full-parallelism figure as unconfirmed; #372 tracks those two tests and its own branch establishes they are test-isolation, not a durability bug.
3. **Commit 11e6674 calls `cargo build --bin xerj --no-default-features` a "release-path binary."** It is a dev/debug-profile binary — the same commit's caveats say no release-profile build was run. The distinction matters here because dev has overflow-checks ON and `ci-test` has them OFF.

Merging once CI is green.

Issue #269, opened 2026-08-10
CLA gate is blind to Co-authored-by trailers, and the pre-enforcement third-party history is unsigned


xerj-org:
## What

`.clabot` resolves signatures by matching **commit authors** against
`.contributors`. When a maintainer re-lands a fork PR — squashing the
contributor's commits onto an xerj-org branch and crediting them with a
`Co-authored-by:` trailer — every commit author is `xerj-org`, so
`verification/cla-signed` goes green and the `cla-signed` label is applied
even though the person whose work it is has never signed.

Live instance: #248, the re-land of #166 by @buger.

- `gh api repos/xerj-org/xerj/pulls/248/commits` → `author_login: xerj-org`
  for all three commits (`c337eb8f`, `81c64883`, `9eb7b6b6`).
- `.contributors` is `["xerj-org","dependabot[bot]","Vinz2168"]` — no `buger`.
- The `Co-authored-by: Leonid Bugaev <leonsbox@gmail.com>` trailer that
  carries the real attribution is invisible to cla-bot.
- #166, which carries the same work under buger's own authorship, has no
  `cla-signed` label at all.

The gate reports green for a reason unrelated to the contributor it exists to
cover. This is a property of the re-land pattern, not of #248 specifically —
any future fork re-land does the same thing.

## Separately: pre-enforcement history

57 commits authored `buger` / `Leonid Bugaev <leonsbox@gmail.com>` are already
on `main`, landed through #135, #154, #160, #161 and #168 (merged
2026-08-02 … 2026-08-04). No `verification/cla-signed` check ran on any of
them — cla-bot's checks first appear around #186 — so these predate
enforcement rather than evade it. They are nonetheless unsigned third-party
contributions in an Apache-2.0 codebase, and a waiver scoped to #248 alone
would not cover them.

`main` also has no branch protection (`GET /repos/xerj-org/xerj/branches/main/protection`
→ 404), so `verification/cla-signed` is advisory on every PR and has never been
able to block a merge.

## Suggested, owner's call

1. **Teach the gate to see trailers.** The existing *CLA gate config* job
   (`.github/workflows/ci.yml:24`) already validates `.contributors` shape;
   extend it to parse `Co-authored-by:` trailers from the PR's commits and fail
   when a co-author is not in `.contributors`. Deliberately not done in the
   pass that found this — editing CI was out of scope there.
2. **Ratify the pre-enforcement history**, either by asking @buger for the one
[...]

PR #402, opened 2026-08-15
fix(search): reject sort on _seq_no and other unresolvable meta-fields (#401)

### LINKED ISSUE #401: sort on _seq_no (and other ES meta-fields besides _score/_doc/_id) silently resolves to null instead of the real value
## What happened

`sort` on any ES meta-field other than `_score`, `_doc`, and `_id` silently resolves to `null` for every hit instead of the real value or a 400 — so `search_after` keyset pagination on that field never advances past the first page, with no error to say why.

Found this while building `xerj-seed` (a companion CLI, see the note at the bottom) — the original design paginated a full-index scan with `sort: [{"_seq_no": "asc"}]` + `search_after`, mirroring how `_seq_no` is used elsewhere in this codebase (`wal_tap.rs`'s `_bulk` version). Every hit came back with `"sort": [null]`.

## Repro

```
$ curl -s -X POST localhost:9200/my-index/_pit?keep_alive=5m
{"id":"<pit-id>"}

$ curl -s -X POST localhost:9200/_search -H 'Content-Type: application/json' -d '
{
  "size": 3,
  "query": {"match_all": {}},
  "pit": {"id": "<pit-id>", "keep_alive": "5m"},
  "sort": [{"_seq_no": "asc"}]
}'
```

Every hit's `"sort"` array is `[null]`, regardless of how many documents (or writes) the index has. `search_after: [<last non-null value>]` is then never reachable, so a client keyset-paging on `_seq_no` gets the same first page forever (or, worse, if it happens to treat a `null`-vs-`null` tie as "no more results", stops after page one and silently drops the rest of the index).

Real Elasticsearch supports this: `_seq_no` has doc values by default and is a legitimate sort key, used in practice for exactly this kind of consistent-pagination-during-migration pattern.

## Where it is

`_seq_no` reaches the sort *parser* fine — `parse_sort` (`engine/crates/xerj-query/src/sort.rs:6447`) only special-cases the literal strings `"_score"`/`"_doc"`; anything else, including `"_seq_no"`, becomes an ordinary `SortField { field: "_seq_no", .. }`.

The break is in *value resolution* — `compute_sort_values` (`engine/crates/xerj-engine/src/index.rs:36972`) special-cases exactly three keys:

```rust
let v = if sf.is_score() {
    ...
} else if sf.is_doc_order() || sf.field == "_id" {
    // `_id` (and `_doc`) is doc metadata, NOT part of `_source`, so a
    // plain `get_field_value` returns Null ...
    Value::String(id.to_string())
} else {
    // generic path: get_field_value(source, &sf.field) — looks inside _source
    ...
}
```
(`index.rs:36980-36986`)

`_seq_no` falls into the generic `else` branch, which calls `get_field_value(source, &sf.field)` — a lookup **inside `_source`**. `_seq_no` is engine metadata, never a literal key in `_source`, so the lookup is `None` → `null`, every time, for every document. Same for `_version`, `_primary_term`, and presumably any other ES meta-field someone tries to sort on — `_id` is the only non-`_score`/`_doc` meta-field that got the special-case treatment (per its own comment, precisely because reindex's `search_after` needed it).

The real value is available elsewhere in the same file — `idx.lookup_seq_no(&h.id)`, used to populate the **response field** `_seq_no` on each hit when `"seq_no_primary_term": true` is requested (`engine/crates/xerj-api/src/es_compat.rs:10705`, `:11331`, `:13248`) — it's just never wired into the sort-value path, so it's usable as a per-hit field but not as a sort/pagination key.

## Impact

Any client — this includes real ES/Kibana/Logstash-family tooling, not just custom scripts — that sorts on an ES meta-field other than `_score`/`_doc`/`_id` gets silent `null`s instead of a working sort or an explicit rejection. `_seq_no`-keyed `search_after` is the concrete case I hit, used for consistent full-index pagination that's resilient to concurrent writes (unlike `_doc`/`_id` order on a mutating index).

## Suggested fix

Either:
- extend `compute_sort_values`'s meta-field special-case to also cover `_seq_no` (via `idx.lookup_seq_no`, already available) and `_version`/`_primary_term`, or
- reject `sort` on an unresolvable meta-field with a 400 (`illegal_argument_exception`, ES's own behavior for a handful of genuinely non-sortable fields) instead of resolving it to `null` for every hit.

The first is more useful; the second is the smaller change and at minimum stops the silent-null failure mode.

---

Filed after using #322/#400's `wal_tap`-adjacent work — the seed tool ended up here by needing exactly this pattern (see https://github.com/Vinz2168/xerj-seed if useful as a second data point: it now sorts on `_doc` instead, since that's the one non-`_score` key that works end-to-end with `search_after`).

Vinz2168:
Fixes #401.

## The bug

`compute_sort_values` (`xerj-engine/src/index.rs`) special-cases exactly `_score`, `_doc`, and `_id` when resolving a sort key. Every other field — including the ES meta-fields `_seq_no`, `_primary_term`, `_version`, `_routing`, `_ignored`, `_index`, `_type` — falls through to an ordinary `_source` lookup. None of those live in `_source` (they're engine/response metadata), so the lookup silently returns `null` for every hit: the sort never distinguishes any two documents, and a `search_after` cursor built from a `null` value never advances. A client keyset-paging on `_seq_no` — a real ES sort key, useful for a full-index scan that's resilient to concurrent writes — gets the same first page forever, with nothing to say why.

Live repro (before this fix):
```
$ curl -s localhost:9200/t/_search -d '{"sort":[{"_seq_no":"asc"}]}'
# every hit: "sort": [null]
```

## The fix (scoped per discussion on #401)

Reject the request instead, at the same request-validation point `validate_search_after_rescore` already uses — `illegal_argument_exception`, 400. Same response shape as the `... is not supported by this XERJ version` xerj already gives for `reindex` (remote source) and `restore` (rename/settings options) when it declines to do something silently wrong rather than what was asked.

This is the **minimal fix**, not "make `_seq_no` a real sort key" — that would mean threading a version-map lookup (`idx.lookup_seq_no`) into `compute_sort_values`'s four call sites, several of which are top-K hot paths built around an existing "pre-clone rejection" fast path that assumes a sort key is derivable from `source: &Value` alone with no external lookup. That's a legitimate follow-up if wanted, but a bigger, riskier change than closing a silent-failure mode. This PR only does the latter.

## What's covered / not covered

- Wired into the main `search()` handler, at the `validate_search_after_rescore` call site — covers `_search`, `{index}/_search`, and PIT-based search (all the same handler/body).
- **Not audited**: `_msearch`'s sub-request path (`msearch_impl`) may parse/validate independently and could need the same check — didn't want to touch a second code path without verifying it live.

## Testing

- New unit tests (`sort_on_unresolvable_meta_field_rejected`, `sort_on_resolvable_fields_accepted`, `unresolvable_meta_field_sort_error_names_the_field`) alongside the existing `validate_search_after_rescore` tests.
- `cargo test -p xerj-api --lib`: 218 passed, 0 failed (no regressions).
- Live-verified against a real build: `sort: [{"_seq_no":"asc"}]` now 400s with a message naming the field; `_doc`, `_id`, and ordinary field sorts are unaffected.

## Context

Found while building [xerj-seed](https://github.com/Vinz2168/xerj-seed), a companion CLI whose original design paginated on `_seq_no` (mirroring how `wal_tap.rs` uses it as the `_bulk` version) — every hit came back `"sort": [null]` with no indication why. Filed as #401 with the full trace; this PR is the fix the discussion there landed on.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -3822,31 +3822,25 @@ fn validate_search_after_rescore(body: &EsSearchBody) -> Option<Response> {
 }
 
 /// ES meta-fields `compute_sort_values`
-/// (`xerj-engine/src/index.rs::compute_sort_values`) cannot yet resolve to a
-/// real per-document value: it special-cases exactly `_score`, `_doc`, and
-/// `_id` (the last one specifically for reindex's `search_after` keyset
-/// paging), and falls back to an ordinary `_source` field lookup for
-/// anything else. Every field below is engine/response metadata, never a
-/// literal key inside `_source`, so that fallback silently resolves to
-/// `null` for every hit — not an error, just a sort that never distinguishes
-/// any two documents. A `search_after` cursor built from a `null` value then
-/// never advances, so pagination keyed on one of these silently returns the
-/// same page forever instead of the rest of the index (#401).
+/// (`xerj-engine/src/index.rs::compute_sort_values`) cannot resolve to a real
+/// per-document value, so it falls back to an ordinary `_source` field lookup.
+/// None of these is ever a literal key inside `_source`, so that fallback
+/// resolves to `null` for every hit — not an error, just a sort that never
+/// distinguishes any two documents. A `search_after` cursor built from a
+/// `null` value never advances, so pagination keyed on one of them silently
+/// returns the same page forever instead of the rest of the index (#401).
 ///
-/// `_seq_no` specifically already has a dedicated, narrower rejection above
-/// this one (`index.disable_sequence_numbers`), but only for that one
-/// opt-in per-index setting; with the setting at its default, sorting by
-/// `_seq_no` (or any of these siblings) still fell through to the silent-null
-/// case this closes.
-const UNSORTABLE_META_FIELDS: &[&str] = &[
-    "_seq_no",
-    "_primary_term",
-    "_version",
-    "_routing",
-    "_ignored",
-    "_index",
-    "_type",
-];
+/// This list is deliberately narrower than the one this change was opened
+/// with. #420 has since taught `compute_sort_values` to resolve `_seq_no`,
+/// `_version`, `_primary_term` and `_index` through the same accessors that
+/// populate the response meta-fields, so rejecting those four now would
+/// REGRESS a working feature — `_seq_no` keyset paging is exactly the
+/// migration workload #401 was filed for, and it walks the whole index today.
+/// Measured on `main` at this change's base, sorting on each of the seven:
+/// `_seq_no` → `[0], [1], [2]`, `_version` / `_primary_term` → `[1]`,
+/// `_index` → the index name, and `_routing` / `_ignored` / `_type` → `[null]`
+/// on every hit. The three that still resolve to `null` are the ones below.
+const UNSORTABLE_META_FIELDS: &[&str] = &["_routing", "_ignored", "_type"];
 
 /// First clause in `sort` that names an [`UNSORTABLE_META_FIELDS`] entry, if
 /// any — as `{"field": "clause_shape"}`, matching `count_sort_clauses`'s
@@ -3864,7 +3858,10 @@ fn first_unsortable_meta_field(sort: &Value) -> Option<&'static str> {
         Value::Array(arr) => arr.iter().find_map(first_unsortable_meta_field),
         Value::String(_) | Value::Object(_) => {
             let field = clause_field(sort)?;
-            UNSORTABLE_META_FIELDS.iter().find(|f| **f == field).copied()
+            UNSORTABLE_META_FIELDS
+                .iter()
+                .find(|f| **f == field)
+                .copied()
         }
         _ => None,
     }
@@ -39040,11 +39037,11 @@ mod request_validation_tests {
     #[test]
     fn sort_on_unresolvable_meta_field_rejected() {
         for sort in [
-            json!(["_seq_no"]),
-            json!([{"_seq_no": "asc"}]),
-            json!(["name", "_primary_term"]),
-            json!("_version"),
-            json!({"_routing": "desc"}),
+            json!(["_routing"]),
+            json!([{"_routing": "asc"}]),
+            json!(["name", "_ignored"]),
+            json!("_type"),
+            json!({"_ignored": "desc"}),
         ] {
             let body = EsSearchBody {
                 sort: Some(sort.clone()),
@@ -39057,6 +39054,10 @@ mod request_validation_tests {
         }
     }
 
+    /// The four meta-fields #420 taught `compute_sort_values` to resolve must
+    /// keep going through. Rejecting them would regress `_seq_no` keyset
+    /// paging, which is the migration workload #401 was filed for and which
+    /// walks the whole index on `main` today.
     #[test]
     fn sort_on_resolvable_fields_accepted() {
         for sort in [
@@ -39065,10 +39066,19 @@ mod request_validation_tests {
             json!(["_id"]),
             json!(["name"]),
             json!([{"name": "desc"}, "_score"]),
+            json!(["_seq_no"]),
+            json!([{"_seq_no": "asc"}]),
+            json!(["_version"]),
+            json!(["_primary_term"]),
+            json!(["_index"]),
             json!(null),
         ] {
             let body = EsSearchBody {
-                sort: if sort.is_null() { None } else { Some(sort.clone()) },
+                sort: if sort.is_null() {
+                    None
+                } else {
+                    Some(sort.clone())
+                },
                 ..Default::default()
             };
             assert!(
@@ -39081,7 +39091,7 @@ mod request_validation_tests {
     #[test]
     fn unresolvable_meta_field_sort_error_names_the_field() {
         let body = EsSearchBody {
-            sort: Some(json!(["_seq_no"])),
+            sort: Some(json!(["_routing"])),
             ..Default::default()
         };
         let resp = validate_sort_meta_fields(&body).expect("should reject");
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -3821,26 +3821,40 @@ fn validate_search_after_rescore(body: &EsSearchBody) -> Option<Response> {
     None
 }
 
-/// ES meta-fields `compute_sort_values`
-/// (`xerj-engine/src/index.rs::compute_sort_values`) cannot resolve to a real
-/// per-document value, so it falls back to an ordinary `_source` field lookup.
-/// None of these is ever a literal key inside `_source`, so that fallback
-/// resolves to `null` for every hit — not an error, just a sort that never
-/// distinguishes any two documents. A `search_after` cursor built from a
-/// `null` value never advances, so pagination keyed on one of them silently
-/// returns the same page forever instead of the rest of the index (#401).
+/// ES meta-fields `compute_sort_values` cannot resolve to a real per-document
+/// value, so it falls back to an ordinary `_source` lookup that finds nothing
+/// and yields `null` for every hit. A `search_after` cursor built from `null`
+/// never advances, so pagination keyed on one of them silently returns the same
+/// page forever instead of the rest of the index (#401).
 ///
-/// This list is deliberately narrower than the one this change was opened
-/// with. #420 has since taught `compute_sort_values` to resolve `_seq_no`,
-/// `_version`, `_primary_term` and `_index` through the same accessors that
-/// populate the response meta-fields, so rejecting those four now would
-/// REGRESS a working feature — `_seq_no` keyset paging is exactly the
-/// migration workload #401 was filed for, and it walks the whole index today.
-/// Measured on `main` at this change's base, sorting on each of the seven:
-/// `_seq_no` → `[0], [1], [2]`, `_version` / `_primary_term` → `[1]`,
-/// `_index` → the index name, and `_routing` / `_ignored` / `_type` → `[null]`
-/// on every hit. The three that still resolve to `null` are the ones below.
-const UNSORTABLE_META_FIELDS: &[&str] = &["_routing", "_ignored", "_type"];
+/// **This list is one entry long, and the two fields removed from it are the
+/// point.** An independent verification refuted the earlier, wider list:
+/// `_routing` and `_ignored` DO resolve, because xerj writes them into
+/// `_source` itself — `bulk.rs` injects `_routing` for any action carrying ES
+/// routing metadata, and `apply_ignore_malformed` inserts `_ignored` as a
+/// literal key (the response layer strips it at render time, but
+/// `compute_sort_values` runs in the engine on the raw stored source, before
+/// that strip). Measured on `main` against a corpus that actually contains
+/// routing values and malformed fields:
+///
+/// ```text
+/// _routing   ['alpha'], ['bravo'], ['charlie'], null, null, null   3/6 real
+/// _ignored   [['ts']], [['ts']], null, null, null, null            2/6 real
+/// _type      null on every hit                                     0/6
+/// _seq_no / _version / _primary_term / _index                      resolved by #420
+/// ```
+///
+/// The earlier measurement that put `_routing` and `_ignored` here was taken on
+/// documents indexed without `?routing=` and without a malformed field, where
+/// both keys are *vacuously* absent — a corpus artefact, not a behaviour.
+/// Sorting on them works, reverses under `desc`, and pages with `search_after`,
+/// which is the workload #401 exists to protect.
+///
+/// The merged CHANGELOG entry for #420 said so in advance: "Rejecting
+/// metadata-named `_source` keys at the API edge is deliberately **not** part of
+/// this change — xerj writes `_routing` into `_source` itself, so the edge rule
+/// needs its own design."
+const UNSORTABLE_META_FIELDS: &[&str] = &["_type"];
 
 /// First clause in `sort` that names an [`UNSORTABLE_META_FIELDS`] entry, if
 /// any — as `{"field": "clause_shape"}`, matching `count_sort_clauses`'s
@@ -39037,11 +39051,11 @@ mod request_validation_tests {
     #[test]
     fn sort_on_unresolvable_meta_field_rejected() {
         for sort in [
-            json!(["_routing"]),
-            json!([{"_routing": "asc"}]),
-            json!(["name", "_ignored"]),
+            json!(["_type"]),
+            json!([{"_type": "asc"}]),
+            json!(["name", "_type"]),
             json!("_type"),
-            json!({"_ignored": "desc"}),
+            json!({"_type": "desc"}),
         ] {
             let body = EsSearchBody {
                 sort: Some(sort.clone()),
@@ -39071,6 +39085,11 @@ mod request_validation_tests {
             json!(["_version"]),
             json!(["_primary_term"]),
             json!(["_index"]),
+            // Refuted the earlier narrowing: xerj writes both of these into
+            // `_source`, so they carry real per-document values and sort.
+            json!(["_routing"]),
+            json!([{"_routing": "desc"}]),
+            json!(["_ignored"]),
             json!(null),
         ] {
             let body = EsSearchBody {
@@ -39091,7 +39110,7 @@ mod request_validation_tests {
     #[test]
     fn unresolvable_meta_field_sort_error_names_the_field() {
         let body = EsSearchBody {
-            sort: Some(json!(["_routing"])),
+            sort: Some(json!(["_type"])),
             ..Default::default()
         };
         let resp = validate_sort_meta_fields(&body).expect("should reject");
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -3821,6 +3821,83 @@ fn validate_search_after_rescore(body: &EsSearchBody) -> Option<Response> {
     None
 }
 
+/// ES meta-fields `compute_sort_values`
+/// (`xerj-engine/src/index.rs::compute_sort_values`) cannot yet resolve to a
+/// real per-document value: it special-cases exactly `_score`, `_doc`, and
+/// `_id` (the last one specifically for reindex's `search_after` keyset
+/// paging), and falls back to an ordinary `_source` field lookup for
+/// anything else. Every field below is engine/response metadata, never a
+/// literal key inside `_source`, so that fallback silently resolves to
+/// `null` for every hit — not an error, just a sort that never distinguishes
+/// any two documents. A `search_after` cursor built from a `null` value then
+/// never advances, so pagination keyed on one of these silently returns the
+/// same page forever instead of the rest of the index (#401).
+///
+/// `_seq_no` specifically already has a dedicated, narrower rejection above
+/// this one (`index.disable_sequence_numbers`), but only for that one
+/// opt-in per-index setting; with the setting at its default, sorting by
+/// `_seq_no` (or any of these siblings) still fell through to the silent-null
+/// case this closes.
+const UNSORTABLE_META_FIELDS: &[&str] = &[
+    "_seq_no",
+    "_primary_term",
+    "_version",
+    "_routing",
+    "_ignored",
+    "_index",
+    "_type",
+];
+
+/// First clause in `sort` that names an [`UNSORTABLE_META_FIELDS`] entry, if
+/// any — as `{"field": "clause_shape"}`, matching `count_sort_clauses`'s
+/// tolerance for both the bare-string and the `{field: order}` clause
+/// shapes.
+fn first_unsortable_meta_field(sort: &Value) -> Option<&'static str> {
+    fn clause_field(v: &Value) -> Option<&str> {
+        match v {
+            Value::String(s) => Some(s.as_str()),
+            Value::Object(o) => o.keys().next().map(String::as_str),
+            _ => None,
+        }
+    }
+    match sort {
+        Value::Array(arr) => arr.iter().find_map(first_unsortable_meta_field),
+        Value::String(_) | Value::Object(_) => {
+            let field = clause_field(sort)?;
+            UNSORTABLE_META_FIELDS.iter().find(|f| **f == field).copied()
+        }
+        _ => None,
+    }
+}
+
+/// Reject `sort` on a meta-field xerj cannot yet resolve to a real value
+/// (#401), instead of silently returning `null` for it on every hit. This is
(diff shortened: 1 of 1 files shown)


xerj-org:
Thanks for this, and apologies for the delay.

Two separate things are red here, and one of them is ours:

**1. Not your fault.** `Format + Clippy` fails on `main` itself — a `clippy::useless_borrows_in_formatting` error in `crates/xerj-engine/tests/sort_on_meta_fields.rs` from #420. The lint is new in rustc 1.97.1, CI's `stable` moved to it, and nothing pins the toolchain. Fix is up in #422.

**2. Yours, and small.** `rustfmt` 1.97.1 wants two hunks in `es_compat.rs` reflowed (`:3864` and `:39062`):

```
-    sort: if sort.is_null() { None } else { Some(sort.clone()) },
+    sort: if sort.is_null() {
+        None
+    } else {
+        Some(sort.clone())
+    },
```

If your local rustfmt is older than 1.97.1 it will keep those inline and `cargo fmt --check` will pass for you while failing here — the same unpinned-toolchain problem. `rustup run 1.97.1 cargo fmt --all` will produce what CI wants.

Happy to push both fixes to your branch if you'd rather — just say so, or leave 'allow edits from maintainers' on and I'll handle it after #422 lands.

xerj-org:
Update, and it's good news for this PR: **it is not superseded, and I almost wrongly closed it as such.**

#420 landed a different fix for #401 while this sat — it makes the meta-fields resolve to real values instead of rejecting them. That covers four of your seven. I just measured all seven against current `main`:

| field | status on main | first 3 sort values |
|---|---|---|
| `_seq_no` | resolved by #420 | `[0], [1], [2]` |
| `_version` | resolved by #420 | `[1], [1], [1]` |
| `_primary_term` | resolved by #420 | `[1], [1], [1]` |
| `_index` | resolved by #420 | `['s401'] ...` |
| **`_routing`** | **still silently null** | `[null], [null], [null]` |
| **`_ignored`** | **still silently null** | `[null], [null], [null]` |
| **`_type`** | **still silently null** | `[null], [null], [null]` |

So three fields still have exactly the defect you filed this for, and your PR is the only thing that closes them. Your framing in the doc comment — an honest 400 for a gap, in the same spirit as the existing `... is not supported by this XERJ version` responses — is the right call for those three.

**What it needs to merge:**

1. Drop `_seq_no`, `_version`, `_primary_term`, `_index` from `UNSORTABLE_META_FIELDS` — rejecting them now would regress #420's fix and break `search_after` keyset paging, which is the migration workload #401 was filed for. Keep `_routing`, `_ignored`, `_type`.
2. `rustfmt` 1.97.1 on the two `es_compat.rs` hunks (see my earlier comment — that's a toolchain-pinning failure on our side, not yours).
3. Rebase once #422 lands, so `Format + Clippy` can actually pass.

I can push all three to your branch since edits-from-maintainers is on — say the word and I'll do it, keeping you as author. Or take it yourself if you'd rather. Either way this one's landing.

xerj-org:
Pushed the three changes I offered earlier, under 'allow edits from maintainers'. **You are still the author** — I did not squash or reword your commit, this is a follow-on. Revert or amend anything you disagree with.

**1. Narrowed the list to three fields.** #420 merged while this sat and took the opposite route for four of your seven: `compute_sort_values` now resolves `_seq_no`, `_version`, `_primary_term` and `_index` for real. Rejecting those now would regress a working feature — `_seq_no` keyset paging is the migration workload #401 was filed for and walks the whole index today (30 of 30, against 4 of 30 before). So `UNSORTABLE_META_FIELDS` is now exactly `_routing`, `_ignored`, `_type` — the three I measured still returning `[null]` on every hit.

Your honest-400 argument is unchanged and still right; it just applies to a smaller set now. I rewrote the doc comment to record the measurement rather than the old rationale.

**2. Tests.** Your two rejection tests asserted `_seq_no` is rejected, which is now the wrong behaviour, so they move to `_routing`/`_ignored`/`_type`. I also added the four #420 fixed to `sort_on_resolvable_fields_accepted`, so a future narrowing of `compute_sort_values` cannot silently start rejecting them again — that is the regression this PR is now the guard against.

**3. rustfmt + rebase.** Reformatted with rustfmt 1.97.1 and rebased onto `main`. To be clear: the two `es_compat.rs` hunks CI flagged were **our** problem, not your formatting — nothing pinned the toolchain, so CI's `stable` had moved ahead of yours (#422).

```
cargo test -p xerj-api -- --test-threads=2    427 passed, 0 failed   (ci-test and dev)
cargo fmt --all --check                       exit 0    (rustfmt 1.97.1)
cargo clippy --workspace --all-targets -D warnings   exit 0
```

CI is re-running. Sorry this took as long as it did to get moving — the blockers were all on our side.

xerj-org:
One more maintainer edit — the last red check was **`CLA gate config`**, and it was not about your code.

Your commit carried:

```
Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>
```

`AGENTS.md:53` forbids `Co-Authored-By` trailers in this repository, and `scripts/check_cla_coauthors.py` enforces it (issue #269):

```
CLA co-author check FAILED for PR #402:
  - c7d07cee: cannot attribute Co-authored-by 'Claude Sonnet 5 <noreply@anthropic.com>'
    to any GitHub login (not a noreply address, and no commit in this repo is authored with it)
```

The reasoning in AGENTS.md is that the accountable party is the human whose account opened the PR — which is you, and your CLA is signed — so the trailer adds a party who cannot be asked about the change. It was failing before my earlier push too; I fixed clippy and the ES-compat gate and left this one, which was an oversight on my part.

I rewrote **only that one line** out of your commit message. Specifically:

- **You are still the commit author** — `Vincenzo Lombardo <lombardo.vincenzo@gmail.com>`, unchanged.
- **Your `Claude-Session:` line is kept.** AGENTS.md asks for agent involvement to be disclosed; it is only the `Co-Authored-By` trailer that is disallowed. Removing your disclosure would have been the wrong edit.
- Nothing else in the message changed.

I am not comfortable rewriting a contributor's commit message without saying so plainly, so: if you would rather do this yourself, force-push over me and I will not touch the branch again.

Re-verified on the rebuilt branch before pushing:

```
cargo fmt --all --check                              exit 0
cargo clippy --workspace --all-targets -D warnings   exit 0
cargo test -p xerj-api -- --test-threads=2           427 passed, 0 failed
git log 3445f7b..HEAD --format=%B | grep -ci co-authored-by   ->  0
```

CI is re-running from scratch.

xerj-org:
## Independent verification — REFUTED

> REFUTED. The maintainer's second correction is right about _routing/_ignored and wrong about _type, for exactly the same reason the first one was wrong: the "null on every hit" measurement for _type was taken on a corpus with no document carrying a literal `_type` key. xerj stores metadata-named keys in `_source` (its own CHANGELOG for #420 and sort_on_meta_fields.rs say so), so an ES 6/7 export reindexed into xerj carries `_type` in `_source` and compute_sort_values' fall-through resolves it. I built the PR head and its base from one worktree (base = same tree with es_compat.rs reverted, the only file the PR touches) and measured on one live node: base returns HTTP 200 with sort values ["anvil"],["gadget"],["widget"] in correct order; head returns HTTP 400. Engine-level, search_after on `_type` walks a 12-doc index completely (12 unique ids). The head binary also contradicts itself — _msearch, _search/template and _async_search all answer the same sort:[{"_type":"asc"}] query correctly with 200 while _search calls it "not supported by this XERJ version yet". Separately the list is under-inclusive by 11 names: _source/_size/_doc_count/_field_names/_meta/_tier/_nested/_nested_path/_feature/_parent/_matched_queries (and any unmapped field) all return 200 with [null] on every hit and strand search_after at 2 of 6 docs. The edits half of the task is CLEAN: authorship preserved, only the forbidden Co-Authored-By line removed, no assertion weakened by either rewrite, and both pins proven load-bearing by experiment. Gates all green: xerj-api 427 passed / 0 failed / 2 ignored in BOTH dev and ci-test at --test-threads=2 under rustc 1.97.1; fmt 0; clippy -D warnings 0; ES-compat YAML 1366 passed / 0 failed / 3 skipped / 1369 total on my own node (port 19402, own data dir).


---

# MUST-FIX (2)

### `_type` is NOT unresolvable — the corrected list repeats the vacuous-corpus error one level down, and the 400 is a live regression

compute_sort_values (xerj-engine/src/index.rs:37123) special-cases _score, _doc/_id, then idx.meta_sort_value (_seq_no/_version/_primary_term/_index), and sends everything else to get_field_value(source, field). `_type` therefore resolves whenever the document carries a literal `_type` key — which xerj permits. The repo says so twice in its own words: CHANGELOG.md:158 ("xerj — unlike Elasticsearch, which rejects a metadata-named key inside a document with a document_parsing_exception — accepts {\"_seq_no\": 1} in a document body and echoes it back in _source") and engine/crates/xerj-engine/tests/sort_on_meta_fields.rs:42-45 ("Note the premise these gates do NOT rest on: 'a meta-field name can never be a _source key'. It can"). The maintainer's re-measurement (_type null on every hit 0/6) used a corpus that exercised routing and malformed fields but still had no `_type` source key — vacuously null, the identical methodology error the second edit exists to fix. A request-level allowlist cannot know the corpus, so a corpus measurement is the wrong instrument for it in the first place.

```
One worktree /tmp/cont-402reg at 6826f36; base = same tree with only es_compat.rs reverted to 3445f7b (git diff --stat: 164 deletions, es_compat.rs only). Live node port 19402, --data-dir /home/claude/work/402reg/data.
Stored source (both builds): d3 {"name":"n3","_type":"widget","ts":"2020-03-01"}; d4 ..."_type":"gadget"...; d6 ..."_type":"anvil"...
BASE  POST /t402/_search {"size":10,"sort":[{"_type":"asc"}]} -> HTTP 200, hits d6/d4/d3 sort=["anvil"],["gadget"],["widget"] then d1/d2/d5 sort=[null]
HEAD  same request -> HTTP 400 {"error":{"type":"illegal_argument_exception","reason":"sort on [_type] is not supported by this XERJ version yet"},"status":400}
Engine probe (cargo test -p xerj-engine, dev, 1.97.1):
  SORT _type asc  -> [("d3","anvil"),("d2","gadget"),("d1","widget"),("d4",Null)]
  SORT _type desc -> [("d1","widget"),("d2","gadget"),("d3","anvil"),("d4",Null)]   (exact reverse)
  SEARCH_AFTER _type collected=12 unique=12   (whole 12-doc index walked — the #401 workload)
```

### The head binary contradicts itself: three other endpoints answer the rejected query correctly on the same data

validate_sort_meta_fields is wired only into search_impl (es_compat.rs:8574). msearch_impl (:22142), search_template (:29927) and the async-search submit path never reach it. So on the SAME node and the SAME corpus, sort:[{"_type":"asc"}] is simultaneously "not supported by this XERJ version yet" (400 on _search) and correctly answered (200, right order, on three other endpoints). That is independent runtime proof from the PR's own binary that the rejection's premise is false — if `_type` really produced null on every hit, _msearch would have returned an arbitrary order, not d6/d4/d3 (anvil<gadget<widget) followed by the nulls. The contributor flagged the msearch gap honestly in the commit message; the problem is not the gap, it is that the gap disproves the claim the gate is built on.

```
HEAD build, port 19402, corpus as above:
  POST /_msearch          {"index":"t402"}/{"size":6,"sort":[{"_type":"asc"}]}  -> HTTP 200, ids ['d6','d4','d3','d1']
  POST /t402/_msearch     {}/{"size":6,"sort":[{"_type":"asc"}]}              -> HTTP 200, ids ['d6','d4','d3','d1']
  POST /t402/_search/template {"source":{"size":6,"sort":[{"_type":"asc"}]}}   -> HTTP 200, ids ['d6','d4','d3','d1']
  POST /t402/_async_search    {"size":6,"sort":[{"_type":"asc"}]}              -> HTTP 200 (is_partial false)
  POST /t402/_search          {"size":6,"sort":[{"_type":"asc"}]}              -> HTTP 400
```


---

# SHOULD-FIX (2)

### Under-inclusive by 11 meta-field names plus every unmapped field — the #401 class is essentially untouched

The stated criterion is "resolves to null on every hit, so search_after never advances". On an ordinary corpus that criterion is met by 11 further meta-field names and by any misspelled/unmapped field, none of which is rejected. The list closes exactly one name in that class — and that name is the only one in the fall-through set that can carry a real per-document value. The selection is therefore inverted with respect to its own criterion. I am not arguing for a wider allowlist: the honest options are ES parity at ingest (reject metadata-named keys in document bodies with document_parsing_exception) or ES parity at query time ("No mapping found for [x] in order to sort on" for every unresolvable sort field). This PR does neither, and leaves ingest at option B while imposing option A on one endpoint.

```
HTTP census, 6-doc corpus, HEAD build, {"size":6,"sort":[{F:"asc"}]}:
  _type              HTTP=400 REJECTED
  _source _size _doc_count _field_names _meta _tier _nested _nested_path _feature _parent _matched_queries -> all HTTP=200, ALL-NULL [[null],[null],[null]]
  not_a_field_at_all -> HTTP=200 ALL-NULL
search_after walk, size=2, 6-doc index:
  _field_names       pages=2 collected=['d1','d2'] unique=2 of 6   <- #401 stranding, HTTP 200
  _tier              pages=2 collected=['d1','d2'] unique=2 of 6
  _source            pages=2 collected=['d1','d2'] unique=2 of 6
  not_a_field_at_all pages=2 collected=['d1','d2'] unique=2 of 6
  _seq_no            pages=4 collected=6 unique=6 of 6   (control: #420 works)
Engine probe ALL-NULL SET on a plain corpus = ["_routing","_ignored","_type","_source","_size","_doc_count","_field_names","_meta","_tier","_nested","_nested_path","_feature","_parent","_matched_queries","not_a_field_at_all"]
```

### Error envelope does not match the neighbouring sort validation it claims to match, nor ES

The commit message says the rejection is emitted "at the same request-validation point validate_search_after_rescore already uses (illegal_argument_exception, 400)". validate_search_after_rescore actually returns search_shard_validation_body (es_compat.rs:3699): top-level type search_phase_execution_exception, reason "all shards failed", with illegal_argument_exception only in root_cause. The new function emits a flat top-level illegal_argument_exception. Two adjacent sort validations in the same handler now produce two different envelopes, and an ES client that switches on error.type sees a type ES never emits for a sort problem. Relatedly, the doc comment retained from the wider list — "This is xerj declining a request a real Elasticsearch cluster would accept (most of these fields have real doc values there)" — is stale now that the list is one entry: xerj advertises version.number 8.13.0, and ES 8 removed `_type` entirely, so ES would reject this sort too (with a different message and envelope). "most of these fields" also no longer parses for a one-element list.

```
es_compat.rs:3699-3708 search_shard_validation_body -> {"error":{"root_cause":[{"type":"illegal_argument_exception",...}],"type":"search_phase_execution_exception","reason":"all shards failed"},"status":400}
es_compat.rs:3892-3910 validate_sort_meta_fields -> {"error":{"root_cause":[{"type":"illegal_argument_exception",...}],"type":"illegal_argument_exception","reason":"sort on [_type] is not supported by this XERJ version yet"},"status":400}
Live: GET / on the node reports "version":{"number":"8.13.0"}. Stale rationale at es_compat.rs:3872-3878.
```


---

# NOTE (5)

### Acceptance depends on JSON key order; inner_hits and top_hits sorts bypass the gate entirely

first_unsortable_meta_field's clause_field reads only o.keys().next(), while parse_sort (es_compat.rs:6535) iterates every key of a clause object. serde_json runs with preserve_order in this workspace (indexmap 2.14.0 in engine/Cargo.lock via serde_json), so the first key is the first key the client wrote. A two-key clause is therefore accepted or rejected purely on key order, and the accepted form still produces the null component the gate exists to prevent. Sorts nested inside collapse.inner_hits and top_hits are never inspected. None of these is a regression (all were 200 before), but they mean the gate is not the invariant its doc comment describes.

```
HEAD build, port 19402:
  {"size":6,"sort":[{"name":"asc","_type":"asc"}]} -> HTTP 200, first hit sort=['n1', None]
  {"size":6,"sort":[{"_type":"asc","name":"asc"}]} -> HTTP 400
  collapse.inner_hits.sort=[{"_type":"asc"}]        -> HTTP 200
  top_hits.sort=[{"_type":"asc"}]                   -> HTTP 200
No false positives found on the shapes that must pass: script sort, _geo_distance sort, nested sort ({"name":{"order":"asc","nested":{...}}}), a user field named `_typeX`, and a user field named `type` all returned HTTP 200. `?sort=_type:asc` correctly returns 400 (the check is placed after the URL-param merge at es_compat.rs:8554).
```

### The 400 pre-empts index_not_found, so a typo'd index name reports the wrong error

validate_sort_meta_fields runs at es_compat.rs:8574; index resolution and the index_not_found_exception are at :8817. A request against a nonexistent index carrying this sort gets 400 "sort on [_type] ..." instead of 404 index_not_found. ES resolves the index first. This mirrors the pre-existing placement of validate_search_after_rescore immediately above, so there is precedent — worth a line in review rather than a blocker.

```
POST /nope404/_search {"sort":[{"_type":"asc"}]} -> HTTP 400 illegal_argument_exception
POST /nope404/_search {"sort":[{"name":"asc"}]}  -> HTTP 404 index_not_found_exception "no such index [nope404]"
```

### One test does not test what its name claims — inherited from the contributor's original, untouched by both rewrites

unresolvable_meta_field_sort_error_names_the_field asserts only resp.status() == BAD_REQUEST; it never reads the response body, so it cannot detect an error message that fails to name the field. This is present verbatim in the contributor's original commit and both maintainer edits only swapped the field name inside it — so it is NOT a maintainer weakening, but the PR ships an untested claim.

```
Replaced the reason with the literal string "nope" (let _ = field; let reason = "nope".to_string();) and re-ran: es_compat::request_validation_tests::unresolvable_meta_field_sort_error_names_the_field ... ok — 3 passed, 0 failed. Reverted.
```

### Edits audit is CLEAN: authorship intact, one line removed, no assertion weakened, both pins proven load-bearing

Authorship: 9cae0f6a is authored Vincenzo Lombardo <lombardo.vincenzo@gmail.com> (committer xerj-org, from the rebase). The pre-rebase original c7d07cee has a byte-identical tree — git diff c7d07cee 9cae0f6a is empty — and the message differs by exactly one line, the removed Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>. The Claude-Session: disclosure trailer is kept. AGENTS.md:53 forbids Co-Authored-By and scripts/check_cla_coauthors.py enforces it, so the removal was required. No Co-Authored-By trailer remains anywhere in the three commits. Test strength: the rejection test keeps all five clause shapes (array+string, array+object, second-position, bare string, bare object) across all three versions; the accepted list GREW 6 -> 11 -> 14 cases. Neither rewrite weakened an assertion.

```
git log 3445f7b..HEAD --format='%an <%ae>' -> Vincenzo Lombardo <lombardo.vincenzo@gmail.com> / xerj-org / xerj-org; grep -i co-authored-by over the three messages -> NONE.
Pin proof (re-add _routing to UNSORTABLE_META_FIELDS): sort_on_resolvable_fields_accepted ... FAILED — panicked at crates/xerj-api/src/es_compat.rs:39103: "sort=[\"_routing\"] should be accepted" (22 passed, 1 failed). Reverted.
Non-vacuity proof (empty the list): sort_on_unresolvable_meta_field_rejected ... FAILED — panicked at es_compat.rs:39064: "sort=[\"_type\"] should be rejected" (1 passed, 2 failed). Reverted; git status --short clean afterwards.
```

### Gates reproduce the maintainer's numbers; one dev-profile flake is pre-existing on base, not this PR

All gates green under the CI toolchain. xerj-api --test script_cpu_budget_http::a_legal_but_arbitrarily_expensive_script_is_a_400_not_a_long_wait failed twice in the dev profile while three heavy builds were running concurrently — it fails identically on base 3445f7b (same worktree, only es_compat.rs reverted), and passes on head once the machine is quiet. Load-sensitive wall-clock/op-ceiling race, pre-existing, not attributable to #402.

```
rustup run 1.97.1 cargo test --no-fail-fast -p xerj-api -- --test-threads=2  (dev)     -> 427 passed, 0 failed, 2 ignored, DEV2_EXIT=0
rustup run 1.97.1 cargo test --profile ci-test -p xerj-api -- --test-threads=2 (ci-test) -> 427 passed, 0 failed, 2 ignored, CI_EXIT=0
rustup run 1.97.1 cargo fmt --all --check -> exit 0
rustup run 1.97.1 cargo clippy --workspace --all-targets -- -D warnings -> CLIPPY_EXIT=0, 0 errors, 0 warnings
es-yaml-runner --url http://localhost:19402 --dir yaml (my own node, my own data dir; never 9200) -> "1366 passed · 0 failed · 3 skipped · 1369 total", 0 FAIL lines
Flake A/B: base (es_compat.rs at 3445f7b) -> a_legal_but_arbitrarily_expensive_script_is_a_400_not_a_long_wait FAILED, same panic message as head.
```


---

## Maintainer decision: closing this, and the reason is mine not @Vinz2168's

**I got this wrong three times in a row on the same PR**, each time by the same method. I measured which meta-fields returned `null` on a corpus I had built, and each time the corpus simply lacked the documents that would have made the field resolve — first no `?routing=` documents, then no `_type` key. The verdict names the real problem precisely:

> A request-level allowlist cannot know the corpus, so a corpus measurement is the wrong instrument for it in the first place.

That is the sentence that closes this PR. No amount of re-measuring fixes an approach whose correctness depends on data the request handler cannot see.

Two pieces of evidence make it concrete. **The head binary contradicts itself** — `_msearch`, `_search/template` and `_async_search` all answer `sort:[{"_type":"asc"}]` with `200` and the right order (`anvil` < `gadget` < `widget`) while `_search` calls it unsupported. If `_type` really produced null everywhere, `_msearch` would have returned arbitrary order. And **the selection is inverted with respect to its own criterion**: eleven other names — `_source`, `_size`, `_doc_count`, `_field_names`, `_meta`, `_tier`, `_nested`, `_nested_path`, `_feature`, `_parent`, `_matched_queries` — plus any unmapped or misspelled field all return `200` with `[null]` and strand `search_after` at 2 of 6 documents. The list rejects the one name in the fall-through set that *can* carry a real value, and accepts every name that cannot.

**@Vinz2168 — your design judgement was sound and the two rewrites I made to your branch were the errors.** The audit confirms your original was untouched in substance: authorship intact, only the `AGENTS.md`-forbidden `Co-Authored-By` line removed, no assertion weakened, and both test pins proven load-bearing by re-adding a field and watching them fail. Preferring an honest 400 over a silent null is the right instinct — it is the instrument, not the instinct, that does not work here.

The verdict lays out the two approaches that would: **ES parity at ingest** (reject metadata-named keys in document bodies with `document_parsing_exception`, as ES does) or **ES parity at query time** (`No mapping found for [x] in order to sort on` for every unresolvable sort field, which covers all twelve names and unmapped fields without needing to know the corpus). Either is a real fix for #401's remaining half; this one cannot be.

Filing that as its own issue so the work is not lost, and reopening #401's residue there. I am sorry for the churn on your branch.

Issue #394, opened 2026-08-15
the semantic_text hint's suppression rule silences the warning on a query that never reaches the vector, and a test pins that as correct


xerj-org:
Split out of the post-merge review of #393 (which fixed #363). Recording it as its own issue because #393 merged before its adversarial verification completed, and this finding was in that verification.

## Summary

The `_xerj.hints` warning added by #393 tells a caller when a lexical query targets a `semantic_text` field. Its suppression rule also silences the hint on a query that **never reaches the vector at all** — and one of the new tests asserts that suppression as correct, so the behaviour is now pinned.

## Why it matters

The hint exists because `match` on a `semantic_text` field silently runs BM25 instead of kNN (#363) — the user paid ~12x the indexing time for embeddings and gets none of them, with no signal. A suppression rule that hides the hint in a case where the vector is likewise not consulted reintroduces the same silence the hint was added to break, in a narrower case.

The test pinning it is the part worth fixing first: a test that certifies the wrong answer makes the behaviour harder to change later, because a future correct fix will look like a regression.

## Suggested

Decide deliberately which of these the suppression rule is meant to express, then make the test say that:

1. *"Do not hint when the query could not have used the vector anyway"* — defensible, but then the hint is absent exactly when a user is most confused about why semantics did not apply.
2. *"Hint whenever a lexical path was taken against a field that carries an embedding"* — noisier, but it never leaves the user without the signal.

Either is fine; the current state is that the rule implements (1) while the entry documenting it reads as (2).

## Provenance

**Verified:** the root-cause chain behind #363 was independently re-derived by the verifier from source — `semantic_text` maps to `FieldType::Text` plus an `EmbeddingConfig`, and the only semantic dispatch is the whole-query short-circuit at `index.rs:13882` peeled by `peel_semantic_query` at `index.rs:29773`. That part of #393 is sound.
**Assumed:** I have not re-run the suppression case myself, and I have not identified which of the new tests carries the assertion. Both are in the verification transcript for #363 and should be confirmed before changing the rule.

xerj-org:
Confirmed this is a real silence-reintroduction, with the dispatch evidence.

The suppression lives in `lexical_on_semantic_text_hint` (es_compat.rs): it collects every `semantic`/`knn` field in the query via `vector_query_fields`, then skips the hint for any lexical clause on a field that also appears there (the `vectorised` check, ~7474).

[...]

PR #410, opened 2026-08-15
fix(api): the semantic_text hint no longer goes quiet on a query that never reached the vector (#394)

xerj-org:
**DO NOT MERGE — adversarial verification rejected this: 5 items**

Adversarial verification returned `sound=false`. The headline item is that **#394 is
reintroduced by its own fix** — the fourth instance of that class in this repo. This PR is
opened for review and to carry the outstanding work forward, **not** for merge. The five
outstanding must-fix items are recorded verbatim below.

*This PR was written by an agent. So were the branch's commits and the verification that
rejected them. The "Verified / Assumed" split at the bottom marks exactly which numbers in
this body came from a command this agent ran versus which are inherited from the
implementing and verifying agents.*

Part of #394. It does **not** close it: blocker 1 below is the #394 defect itself,
reachable on this branch through at least three request shapes.

---

## Outstanding must-fix items (verbatim from the verification verdict)

> 1. BLOCKER — #394 is reintroduced by its own fix, fourth instance of that class in this repo. `bool_clause_reach` (engine/crates/xerj-api/src/es_compat.rs:7531) guards only `Position::UnderNested`; the `peel_knn_query` shape test at es_compat.rs:7548 (`if scoring.len() == 1 && must_not.is_empty()`) runs regardless of the position it was reached from, so from `Position::Unreachable` it MANUFACTURES a `KnnOnly` dispatch position. A provably inert `knn` is then classified `Dispatched` and the hint is suppressed — exactly the silence #394 was filed about. Measured by me at the HTTP boundary on the 3-doc test index, branch pristine at 096332f: (A) `bool.should:[{match ctx},{bool:{must:[{knn ctx_vector}]}}]` → ids 0,2 at 2.0986123/1.6931472 (byte-identical to the branch's own documented pure-BM25 numbers for `bool.should:[match,semantic]`), vector-only doc 1 absent, and the response is IDENTICAL for query_vector [0.1]*32 vs an opposed [±0.9]*32 — the kNN contributed nothing — yet `semantic_hint` = None. (B) `bool{must:[{bool:{must:[{knn}],filter:[{match ctx}]}}], must_not:[{match title zzzz}]}` → ZERO hits, hint = None (silence on an empty result set, the case the branch's own `a_hybrid_inside_a_bool_is_inert_and_still_hints` calls "the least helpful moment available to be silent in"). (C) `dis_max.queries:[{match ctx},{bool:{must:[{knn}]}}]` → ids 0,2 at 1.0/1.0, doc 1 absent, hint = None. Same shape inside a `bool.filter` is also affected. The asymmetry proves it is a model hole, not a policy: the IDENTICAL shape with `semantic` in place of `knn` returns the identical hits (2.0986123/1.6931472) and DOES hint. FIX I VALIDATED: gate the branch on the position, `if matches!(pos, Position::WholeQuery { .. } | Position::KnnOnly) && scoring.len() == 1 && must_not.is_empty()`. With that one change probes A/B/C all hint correctly and all 14 of the branch's own tests still pass (I ran `cargo test -p xerj-api --lib semantic_text_lexical_hint_tests` = 20 passed with my 6 probes added).

> 2. Add permanent regression tests for the class above — a `bool` reached from a non-dispatch position. At minimum: (a) a two-scoring-clause outer `bool` holding a single-clause `bool`+`knn`; (b) a `must_not`-disqualified outer `bool` holding the same; (c) a peelable `bool` sitting inside a `dis_max` or inside another `bool`'s `filter`. Without these the shape returns on the next refactor, and the branch's own report lists `dis_max` and `must_not disqualification` among six shapes it probed, agreed with the model, and then threw away un-pinned — those two probes were run at the wrong nesting depth and missed the defect.

> 3. CHANGELOG.md self-contradicts itself in the #394 entry: "Two shapes are deliberately **not** claimed either way, because the walk cannot prove them ... a `knn` under a `nested` ..., and a multi-kNN `bool` behind a `constant_score` or a `boost` — those two really are inert (both return zero hits, against three for the bare form) and are flagged". The second shape IS claimed and IS flagged (the branch's own `a_multi_knn_behind_a_wrapper_it_cannot_survive_still_hints` asserts the hint fires). Only ONE shape is withheld. Rewrite so the user-facing entry does not state the opposite of the shipped behaviour.

> 4. Disclose that the one genuinely withheld shape is silent on a zero-hit response. I measured the branch's own `a_knn_under_a_nested_is_not_certified_inert` request shape (`bool{must:[{nested{knn}}], filter:[{match ctx}]}`): it returns ZERO hits for query_vector [0.1]*32 AND for an opposed [±0.9]*32, and no hint is emitted. That shape is not merely unclassifiable — it is measurably inert, and the caller gets zero hits plus silence, which is the #394 symptom retained on purpose. The choice is defensible and is disclosed in the code comment, but the CHANGELOG should say the withheld case is known-silent on an empty result set rather than implying it is only theoretically unproven.

> 5. The commit body's FAIL-BEFORE block is not reproducible from anything in the repository. It lists `a_knn_under_a_nested_is_not_certified_inert` as failing when "the source hunk" is reverted, but the baseline it was reverted to is the branch's own round-1 commit 9eaf842, which was force-pushed away. I re-ran the same tests against origin/main's rule (every vector clause anywhere counts as reached, mechanised as an early `return VectorReach::Dispatched`): 5 tests fail — `a_semantic_clause_that_never_runs_does_not_suppress_the_hint`, `a_semantic_in_the_slot_where_a_knn_would_dispatch_still_hints`, `a_hybrid_inside_a_bool_is_inert_and_still_hints`, `a_multi_knn_behind_a_wrapper_it_cannot_survive_still_hints`, `a_top_level_knn_that_never_runs_does_not_suppress_the_hint` — but `a_knn_under_a_nested_is_not_certified_inert` PASSES. Either name the unreachable baseline explicitly in the commit body or restate the fail-before against origin/main, where that test is vacuous with respect to this change.

### This agent's independent check of items 1 and 3

Items 1 and 3 are both decidable by inspection, and this agent confirmed both without
re-running the verifier's HTTP probes:

- **Item 1 is structurally visible in the diff.** In `bool_clause_reach`
  (`engine/crates/xerj-api/src/es_compat.rs`), the only position guard is the early
  `if matches!(pos, Position::UnderNested)` return. The `peel_knn_query` shape test that
  follows —

  ```rust
  if scoring.len() == 1 && must_not.is_empty() {
      vector_clause_reach(scoring[0], Position::KnnOnly, out);
  ```

  — does not read `pos` at all, so it promotes a clause to `Position::KnnOnly` no matter
  what position the `bool` was reached from, including `Position::Unreachable`. The
  `peel_multi_knn_query` branch immediately below it *does* gate on
  `pos == (Position::WholeQuery { multi_knn: true })`, which is what the `peel_knn_query`
  branch is missing. The verifier's proposed one-line gate is consistent with the
  surrounding code's own convention.
- **Item 3 is a textual contradiction in the shipped CHANGELOG.** The entry says two shapes
  are "deliberately **not** claimed either way" and then, in the same sentence, says of the
  second that it is "flagged".

## What is on the branch

Round-1 verification raised five blockers; the implementing agent reports all five
addressed, reproducing the two behavioural ones first (`8 passed; 4 failed` against the
round-1 branch). The mechanism replaces the old "a `semantic`/`knn` key appears anywhere"
rule with a three-valued positional walk — `Position { WholeQuery { multi_knn }, KnnOnly,
UnderNested, Unreachable }` and `VectorReach { Inert, Unproven, Dispatched }` — that mirrors
each short-circuit `Index::search` actually tries. The implementing agent also found and
fixed a *third* instance of the #394 class inside its own draft (multi-kNN behind
`constant_score`/`boost`). Round-2 verification then found a *fourth*, which is blocker 1.

**Scope is diagnostic only, on purpose.** The engine behaviour underneath — a vector clause
outside a dispatch position is parsed and then contributes no hits and no score, rather than
running or being rejected — is unchanged, and both the CHANGELOG and the commit body say so.

## Verified by this agent (commands run, output seen)

- `origin/fix/issue-394` is `096332f`. `git merge-base origin/main origin/fix/issue-394`
  returns `714616dc26e848c0d9913d03bbddad60f1205f35`, which **is** `origin/main`'s head, so
  the branch already contains all of current main. **No merge was needed and none was made;
  there was no CHANGELOG conflict to resolve.** Nothing was pushed by this agent.
- `git diff --stat origin/main...HEAD` → 2 files changed, 989 insertions(+), 58 deletions(-)
  (`CHANGELOG.md`, `engine/crates/xerj-api/src/es_compat.rs`).
- Commit author and committer are both `xerj-org <git@xerj.org>`; `git log origin/main..HEAD`
  contains no `Co-Authored-By` trailer. Both CI commit rules pass.
- A `## [Unreleased]` CHANGELOG entry for #394 already exists on the branch (it is the
  subject of must-fix item 3). This agent added no CHANGELOG text; correcting the
  contradiction is left to the round-3 fix so that the verdict quoted above stays accurate
  against the code as pushed.
- The code and CHANGELOG readings backing "This agent's independent check of items 1 and 3"
  above.
- `cargo test -p xerj-api --lib` on `096332f`, dev profile (overflow-checks **on**), run by
  this agent in its own private `CARGO_TARGET_DIR`:
  `test result: ok. 222 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 18.71s`.
  This matches the implementing agent's dev-profile figure. **Note what it does and does not
  mean:** the suite is green *and* blocker 1 is present, because no test on this branch
  exercises a `bool` reached from a non-dispatch position. Green here is the reason must-fix
  item 2 exists.

## Assumed (not re-run by this agent)

Everything numeric below is inherited from the implementing agent's or the verifier's
report and was **not** independently reproduced here:

- All HTTP-boundary `_score`/id measurements, in the CHANGELOG entry, the commit body, and
  in must-fix items 1, 4 and 5 — including probes A/B/C, the `0.49674308/0.3751064` and
  `0.99348617/0.7502128/0.42133647` figures, and the zero-hit results for the wrapped
  multi-kNN and the nested-kNN shapes.
- The implementing agent's `222/222` under the **ci-test** profile (overflow-checks off) —
  this agent reproduced only the dev-profile run — plus its `cargo fmt --all --check` and
  `cargo clippy -p xerj-api --all-targets -- -D warnings` runs, and its full
  `cargo test -p xerj-api` across all targets.
- The verifier's claim that its one-line position gate fixes probes A/B/C while keeping all
  14 existing tests passing (`20 passed` with its 6 probes added).
- The round-1 reproduction (`8 passed; 4 failed`) and the item-5 re-run against origin/main's
  rule.

**Not run by anyone, and deliberately:** the ES-compat conformance suite, which deletes every
index on the node it is pointed at. The full workspace test suite was not run either; the
change is confined to one function group in `xerj-api`.

Six shapes the implementing agent probed are **not** regression-protected: nested
single-clause bool around knn, constant_score around a filtered-knn bool, msm on a knn bool,
must_not disqualification, dis_max, and a hybrid whose sub-query is an inert bool. Per
must-fix item 2, two of those — `dis_max` and `must_not` disqualification — were probed at
the wrong nesting depth and are exactly where blocker 1 hides.

## Next step

Round 3: apply the position gate from item 1, pin the three shapes in item 2, and rewrite
the CHANGELOG per items 3 and 4 and the commit body per item 5. Re-run the original #394
repro afterwards and check for a *fifth* instance of the same shape — this class has now
survived three consecutive fixes.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,8 +49,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   "lexical_on_semantic_text"`) that names the field, says the embedding was
   not consulted, and gives a ready-to-paste `semantic` query body. Scoring,
   hits and every ES-compat response field are unchanged, and the hint stays
-  quiet when the same request already reaches the vector through `semantic`
-  or `knn` (including the ES 8.x top-level `knn` block).
+  quiet when the same request really does reach the vector — a `semantic` or
+  `knn` clause the engine dispatches, or a `hybrid` sub-query (see #394 below,
+  which corrects what "reaches the vector" was originally taken to mean).
+
 - **A `quantization: "scalar8"` (`int8_hnsw`) vector field scored updated
   documents from stale quantized data, silently, forever**
   ([#371](https://github.com/xerj-org/xerj/issues/371)).
@@ -203,6 +205,67 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   codec store going away. In principle this change adds one min/max pass over
   the candidate vectors and removes a lock acquisition; both are small against
   the encode/decode/similarity work the scan already does.
+
+- **The `semantic_text` hint no longer goes quiet on a query that never
+  reached the vector** ([#394](https://github.com/xerj-org/xerj/issues/394)).
+  The suppression rule shipped with #363 counted a `semantic` or `knn` key
+  *anywhere* in the request as "this caller already got the semantics". XERJ
+  enters vector search only from a whole-query short-circuit, so two of the
+  shapes it fell silent on never consult the vector at all: `bool.should:
+  [match, semantic]` — what a caller writes when they mean *both* signals —
+  and the ES 8.x top-level `knn` block beside a `query`, which the ES-compat
+  layer folds into exactly that `bool`. Each runs its lexical half, drops its
+  vector half, and had the one diagnostic that would have said so suppressed;
+  one of #363's own tests pinned the first as correct.
+
+  Suppression is now **positional**: the request is classified clause by
+  clause against each of the five short-circuits the engine actually tries —
+  `peel_knn_query`, `peel_multi_knn_query`, `peel_nested_knn_query`,
+  `peel_semantic_query`, `peel_hybrid_query` — rather than against the
+  presence of a keyword. The distinction that does the work is that **only a
+  `knn` is ever peeled out of a `bool`**: `peel_knn_query` is the one
+  short-circuit with a `Bool` arm and it recurses into itself, so a
+  `semantic` or a `hybrid` written in the same slot is dropped. A request
+  whose vector clause is inert now gets a hint that names the clause and
+  suggests a `hybrid` body — the shape that does run both halves — instead
+  of being told to write the `semantic` clause it can already see in its own
+  request body.
+
+  Three shapes that DO reach the vector stay silent, so the fix is not "hint
+  on everything": a `hybrid` sub-query, a `bool` holding one `knn` beside a
+  lexical `filter` (`peel_knn_query` folds that into a filtered kNN), and the
+  top-level `knn: [...]` array, whose `knn[].filter` is the one way a lexical
+  clause reaches this hint from a request whose vectors all ran.
+
+  Measured at the HTTP boundary on the three-document index in the tests
+  (`_score` values are from those runs, not from reading the code):
+
+  * dispatched, and silent — `bool.should:[{knn}] + filter:[{match}]` returns
+    the same hits and the same `_score`s (0.49674308, 0.3751064) as the
+    canonical top-level filtered `knn` it is folded into, and a different
+    query vector reorders them; the top-level `knn: [...]` array returns all
+    three documents, including the one no lexical clause here can reach.
+  * inert, and now flagged — `bool.should:[{semantic}] + filter:[{match}]`
+    returns 1.0/1.0 for two documents and returns *exactly the same response*
+    when the `semantic` clause's query text is changed to target a different
+    document, which is what "contributes no hits and no score" looks like
+    from outside. `bool.must:[{hybrid}] + filter:[{match}]` returns ZERO hits
+    where its own `filter` alone returns two.
+
+  Two shapes are deliberately **not** claimed either way, because the walk
+  cannot prove them and a hint that is wrong about what the engine did is
+  worth less than no hint: a `knn` under a `nested` (`peel_nested_knn_query`
+  dispatches some of those and this walk does not model which), and a
+  multi-kNN `bool` behind a `constant_score` or a `boost` — those two really
+  are inert (both return zero hits, against three for the bare form) and are
+  flagged, but they are singled out here because `peel_multi_knn_query` is
+  the one peel with no `Constant`/`Boosted` arm, and treating them as reached
+  would have put this very bug back inside its own fix.
+
+  Scope: this changes the diagnostic only. The engine behaviour underneath —
+  a vector clause outside a dispatch position is parsed and then contributes
+  no hits and no score, rather than running or being rejected — is unchanged
+  and is not fixed here.
 
 ## [1.0.0-rc.17] - 2026-08-15
 
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -7316,31 +7316,279 @@ fn lexical_text_clauses(q: &Value, out: &mut Vec<(String, String)>) {
     }
 }
 
-/// Fields the same query already reaches through a vector clause, by whatever
-/// name that clause used — `semantic` names the text field, `knn` names the
-/// companion vector field.
-fn vector_query_fields(q: &Value, out: &mut Vec<String>) {
-    match q {
-        Value::Object(o) => {
-            for (k, v) in o {
-                if matches!(k.as_str(), "semantic" | "knn") {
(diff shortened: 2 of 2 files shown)


xerj-org:
Closing. This branch cannot be salvaged by rebasing, and leaving it open implies otherwise.

## Why

Its own adversarial verification returned `sound=false` with the blocker being that **#394 is reintroduced by its own fix** — the verdict recorded it as the fourth instance of that class in this repo. `bool_clause_reach` (`es_compat.rs:7531`) guards only `Position::UnderNested`, while the `peel_knn_query` shape test sits outside that guard, so the suppression rule the PR narrows is still reachable through at least three request shapes.

It is also DIRTY against a `main` that has since taken #417, #420, #424, #429, #432 and three outside contributions, and the file it edits — `es_compat.rs` — is the one every one of those touched.

## Why not rebase it

The same reason #388 was closed. This is an operator-level patch to a suppression rule, and #394's actual shape is a *reachability* question: which query positions can arrive at the hint without having reached the vector. Conflict-resolving this patch onto current `main` would produce a branch that compiles, passes CI, and still encodes a model its own verification says is wrong. That is precisely how the "a fix that reintroduces the class it fixes" pattern has kept recurring — four times by the verdict's count, and #427 has since added three more rounds of the same shape in `sniff.rs`.

## What survives

**#394 stays open** and is unaffected by this closure. The work worth keeping is in the verdict, not the diff: the five must-fix items are recorded verbatim in this PR's body and remain the acceptance criteria. The next attempt should start from the reachability question — enumerate the positions from which the hint is emitted, and decide per position whether the vector was consulted — rather than from narrowing the existing suppression condition.

Reopen if you disagree; the argument is here to be contradicted rather than buried.

Issue #403, opened 2026-08-15
sniff: %PDF- is an unqualified printable magic matched before the magic table — prose about PDF sniffs as a PDF


xerj-org:
Found while fixing #379/#380 (agent-driven, Claude Opus 5, driven by @xerj-org). Same *class* of defect, different code path, so it is filed separately rather than widened into that fix.

## Summary

`sniff.rs` qualifies every printable-ASCII signature in `MAGIC_TABLE` (that is what #379/#380 finishes). But three signatures are matched *before* the table, as early returns in `sniff_bytes`, and one of them is printable ASCII with no qualifier at all:

```rust
if prefix.starts_with(b"%PDF-") {
    return Ok(mk(Family::Pdf));
}
```

`SQLite format 3\0` and `PK\x03\x04` are fine — both carry a byte text cannot contain. `%PDF-` does not.

## Reproduction

Measured through `sniff()` on real files, on the `fix/issue-379` tree (so with #379 already fixed):

```
          cars.csv  family=csv        binary_kind=None      <- #379, fixed
     gif-notes.txt  family=txt-prose  binary_kind=None      <- #379, fixed
     id3-notes.txt  family=txt-prose  binary_kind=None
     pdf-notes.txt  family=pdf        binary_kind=None      <- THIS ISSUE
  sqlite-notes.txt  family=txt-prose  binary_kind=None
     zip-notes.txt  family=txt-prose  binary_kind=None
```

`pdf-notes.txt` is 1,740 bytes of ordinary prose:

```
%PDF- is the five byte header every PDF file begins with. (x30)
```

It sniffs `Family::Pdf` and is handed to the PDF extractor, which cannot parse it.

## Why this is lower severity than #379

`%PDF-` is five characters including a `%` and a `-`, so the collision surface is far smaller than `BM`'s two letters — you essentially have to be writing about PDF, at the very start of the file. It also does not produce the *silent* `binary content` junk reason that made #379 bad; the failure surfaces as a PDF parse error instead.

## Fix sketch (not attempted here)

The header is `%PDF-M.N` and the spec requires an end-of-line marker immediately after the version, so `%PDF-`, an ASCII digit, `.`, an ASCII digit, then CR or LF is a structural qualifier that ordinary prose fails (a sentence has a space after "1.7", not a newline). Whoever takes it should also move the check into `MAGIC_TABLE` so it is covered by `printable_magic_tests::every_printable_signature_carries_a_qualifier`, which today walks the table only.

[...]

Issue #381, opened 2026-08-15
autoindex: fallback prose amplifies one <=16 MiB magic-less payload into thousands of records without warning (~40x peak RSS)


xerj-org:
Two byte-statistics guards in `sniff.rs` were removed because they junked
non-Latin text (a whitespace-density guard, and a "mostly non-ASCII under the
windows-1252 fallback" guard — see the NOTE at
`engine/crates/xerj-autoindex/src/sniff.rs:626-653`). Removing them was correct:
they silently deleted Chinese, Japanese, Korean, Thai, Lao, Khmer, Burmese,
Cyrillic, Greek, Hebrew and Arabic documents worldwide, and this repo's own
suite contained a reachable false positive.

But they existed for a reason, and nothing replaced them for that reason. A
binary payload with **no magic number** and no NUL or control bytes — a raw
texture, a `.bytes` blob, a packed asset — still decodes to printable characters
under the windows-1252 fallback, still classifies as prose, and is still
sectioned and indexed in full. Only TGA is caught, and only because
`looks_like_tga_header` (`sniff.rs:459`) checks the structure of its 18-byte
header.

### Measured

Through `sniff_bytes` and `extract::split_sections` on
`feat/unity-autoindex-reland` @ `a1ba57f`:

```
soup-4mib     -> ("txt-prose", None) len=4194495
soup-sections -> 2048 sections from 4194495 bytes
```

The fixture is 4,194,495 bytes of printable, NUL-free, non-repeating byte soup
named `texture.bytes`. It classifies as prose and expands into **2048 indexed
records** of pure noise. The cost scales linearly with file size and there is no
cap: the same blob at 200 MB is ~100k records.

### What this is not

This is **not** the peak-RSS bug from #239. `for_each_section`
(`engine/crates/xerj-autoindex/src/extract/mod.rs:398`) is genuinely streaming —
it holds a window of at most `2 * SECTION_CHARS + 4` bytes and keeps the
paragraph borrowed, so peak memory per file is bounded and that fix is intact.
The unbounded quantity here is the **number of records emitted**, and therefore
index size, bulk traffic and indexing time.

### Suggested direction

A bound is the right shape of fix; a byte-statistics classifier is not, because
that is what produced the data-loss regressions. Options worth weighing:

* Cap records-per-file for the text families the way the Unity/SQL-dump sample
  limits already cap sampling, and report the truncation on the progress meter
  rather than silently dropping (the `shadowed_junk_entries` reporting added in
[...]

Issue #380, opened 2026-08-15
sniff: `GIF8` and `BM` are unqualified printable-ASCII magic — text starting with those bytes is junked as an image


xerj-org:
`sniff_bytes` takes two printable-ASCII magic signatures on faith. Any text file
whose first characters happen to be `GIF8` or `BM` is classified `Family::Binary`,
and `scan_file` turns that into `junk: binary content (...)` — the file is never
indexed and the user is never told why.

### Where

`engine/crates/xerj-autoindex/src/sniff.rs:212` and `:215`:

```rust
(&b"GIF8"[..], "gif", accept as fn(&[u8]) -> bool),
(&b"BM"[..],   "bmp", accept as fn(&[u8]) -> bool),
```

`accept` (`sniff.rs:387`) is `fn accept(_prefix: &[u8]) -> bool { true }` — it
performs no check at all. For the other signatures in that table the qualifier is
either a real structural check (`psd_version`, `riff_form`, `ogg_page`,
`flac_metadata_block`, `id3v2_header`, `fbx_header`) or `accept` is genuinely
sufficient because the signature contains a byte text cannot contain (`\x89PNG`,
`\xff\xd8\xff`, `\x7fELF`, `\x00\x00\x01\x00`, the NUL in `II*\x00` / `MM\x00*`,
the `\x01` in the EXR magic). `GIF8` and `BM` are the only two where `accept` is
used on an entirely printable signature.

`BM` is the worse of the two: it is **two** characters. Every English sentence
beginning "BM" qualifies.

### Measured

Through `sniff_bytes` on `feat/unity-autoindex-reland` @ `a1ba57f`
(`cargo test -p xerj-autoindex --profile ci-test --lib -- --nocapture`):

```
gif8-prose     -> ("binary", Some("gif"))
gif89a-prose   -> ("binary", Some("gif"))
bm-prose       -> ("binary", Some("bmp"))
gif8-csv       -> ("binary", Some("gif"))
```

`gif8-csv` is `GIF8,name,value\n1,alpha,2\n...` — a well-formed CSV whose first
column header is `GIF8`. It is junked as an image.

### Pre-existing, not introduced by #378

Proven by construction rather than by measurement. On `origin/main`
(`a20dd82`) the same table is
(`git show origin/main:engine/crates/xerj-autoindex/src/sniff.rs`, lines 161-174):

```rust
for (magic, kind) in [
    (&b"\x89PNG"[..], "png"),
[...]

Issue #379, opened 2026-08-15
sniff: unqualified printable magic signatures BM and GIF8 silently junk text files (a CSV with a BMW column header is dropped)


xerj-org:
Found while verifying PR #378 (agent-driven, Claude Opus 5, driven by @xerj-org), then reproduced by hand on the released binary. Pre-existing on `main`, not introduced by any open PR.

## Summary

`sniff.rs` treats `BM` (2 bytes) and `GIF8` (4 bytes) as binary magic signatures without any further qualification. Both are ordinary printable ASCII, so any text file whose first bytes happen to spell them is classified `Family::Binary` and silently junked by `autoindex` — never indexed, with a "binary content" reason.

## Reproduction

Three text files, one folder, stock `v1.0.0-rc.16`:

```
cars.csv        ->  BMW,model,year\nX5,SUV,2024\n...
gif-notes.txt   ->  GIF89a is the header format used by the GIF image standard....
id3-notes.txt   ->  ID3 tags are metadata containers....
```

```sh
xerj autoindex ./magic --prefix magic --no-semantic --no-graph
```

```
phase A: 1 datasets inferred, 2 junk/skipped files
done — 1 records live
```

Only `id3-notes.txt` survives. **`cars.csv` (starts `BM`) and `gif-notes.txt` (starts `GIF8`) are both junked.** A CSV whose first column header is `BMW`, `BMI`, `BMP_WIDTH` — or any prose beginning with the word "BM..." — silently disappears from the index.

`ID3` is *not* affected on rc.16, so this is specifically the unqualified two- and four-byte signatures.

## Why it matters

`autoindex` is the zero-config entry point: point it at a folder and everything becomes searchable. A silent, unexplained omission is the worst failure mode for that promise, because the user has no signal that a file was dropped and the reason given ("binary content") is wrong. `BM` is only two characters — the collision surface is large, and CSV headers are exactly the kind of content that starts with capitalised short tokens.

## Fix

Qualify both the way PR #378 qualifies its own printable signatures:

- **BMP**: after `BM`, bytes 2..6 are a little-endian file size that should be plausible against the real file length, and bytes 6..10 are reserved and conventionally zero.
- **GIF**: require the full 6-byte version stamp `GIF87a` or `GIF89a`, then the logical-screen-descriptor width/height, rather than the 4-byte `GIF8` prefix.

[...]

PR #411, opened 2026-08-15
fix(autoindex): qualify the BM and GIF8 magic signatures (#379, #380) — DO NOT MERGE, verification rejected

xerj-org:
**DO NOT MERGE — adversarial verification rejected this: 5 items**

An adversarial verification pass ran against commit `811485a` and returned
`sound=false`. Two of the five items are **live defects of the same class this PR
claims to fix**: the shipped GIF qualifier still classifies ordinary ASCII text as
an image. This PR is opened for review and for the record, not because it is ready.
Do not merge it until items 1–3 are closed and items 4–5 are resolved.

### Outstanding MUST-FIX (recorded verbatim from the verification)

> 1. Close the 0x2C Image-Descriptor path in `gif_screen_descriptor`
>    (engine/crates/xerj-autoindex/src/sniff.rs:471-487). The shipped rule still junks
>    ordinary ASCII text as an image. Verified through the real `sniff()` on this branch:
>    `"GIF89a format,1 2 3 4 5 and the rest of this note is about the format. ".repeat(30)`
>    -> binary/Some("gif"); likewise `"GIF89a header,a b c d e f g h ..."`,
>    `"GIF89a fields,x y z w ..."`, the GIF87a variant, and the CSV
>    `"GIF89a format,1 2 3 4 5\nrow,a,b,c,d\nrow2,e,f,g,h\n"` -> binary/Some("gif").
>    For all-ASCII input `packed & 0x80` is always clear, so `block` is always 13; a ','
>    at offset 13 plus four little-endian u16s whose HIGH bytes (offsets 15,17,19,21) are
>    spaces satisfies `left+width <= canvas_width && top+height <= canvas_height`.
>    Fix: decode one field further to the LZW minimum code size byte -- after the 10-byte
>    Image Descriptor plus any Local Colour Table its packed byte declares, that byte must
>    be in 2..=12, which is entirely non-printable. Measured on my corpus: 12/12 real
>    image-descriptor-first GIFs satisfy it; every counterexample above carries 0x20 there.
>
> 2. Close the 0x21 Extension path in the same function. The four spec labels are not
>    sufficient once the windows-1252 fallback in `decode()` (sniff.rs:651) is considered:
>    0xF9/0xFE/0xFF are ordinary latin-1 letters. Verified through the real `sniff()`:
>    bytes `"GIF89a fichie!"` + 0xF9 + latin-1 French prose -> binary/Some("gif").
>    Fix: the sub-block size byte after the label is fixed by the spec -- 4 after 0xF9
>    (Graphic Control), 11 after 0xFF (Application), 12 after 0x01 (Plain Text), all
>    non-printable. Measured on my corpus: 97 x (0xF9,4) and 56 x (0xFF,11), 0 violations
>    in 153 applicable first-block extensions; the 2 Comment (0xFE) extensions carry the
>    variable sizes the spec allows.
>
> 3. Extend `prose_that_lands_a_block_introducer_at_the_right_offset_is_still_text`
>    (sniff.rs:1800-1830) with the counterexamples above. Its three current fixtures pass
>    by accident of which characters happen to land at offsets 15/17/19/21, not because the
>    property it names holds -- so the test certifies a guarantee the code does not provide,
>    and the doc comment at sniff.rs:396-451 and the CHANGELOG state that guarantee as fact.
>
> 4. Correct the fail-before pass counts in the commit body / PR. Re-running both proofs
>    myself: (A) reverting the two MAGIC_TABLE rows to `accept` gives
>    `5 passed; 3 failed; 639 filtered out`, not `4 passed; 3 failed`; (B) reverting only
>    the two qualifier bodies gives `6 passed; 2 failed; 639 filtered out`, not
>    `5 passed; 2 failed`. The module holds 8 tests and 639+8=647 matches the suite total,
>    so the reported figures cannot both be right. Everything else about both proofs
>    reproduced exactly, including the quoted panic messages.
>
> 5. The ES-YAML conformance suite (AGENTS.md:66, `must stay at 0 failed ... before any
>    engine change lands`) was not run. The author discloses this rather than burying it and
>    the diff touches no query/mapping/wire code, so the real risk is low -- but it is a hard
>    gate, so it needs either a run or a recorded waiver from whoever owns that gate, not an
>    agent's self-granted exemption.

Items 1 and 2 are instances of **"THE BIG ONE"** in this repo's agent brief — a fix that
reintroduces the very class it fixes. The author explicitly hunted for that shape and
documented the reasoning behind the current rule; the hunt was not deep enough. The
CHANGELOG bullet added by this branch asserts the guarantee that item 1 disproves, so the
CHANGELOG needs correcting alongside the code.

---

## What this is

Part of #379. Part of #380 (the same defect filed twice).
**Not** `Closes` — see the outstanding items above; the primary symptom (text junked as an
image) still reproduces on this branch through a different input than the one reported.

`BM` and `GIF8` were the last two entries in `MAGIC_TABLE` made only of printable ASCII and
believed without checking anything behind them, so a text file whose first bytes spelled them
was classified `Family::Binary` and dropped by `scan_file` as "binary content (bmp)" /
"(gif)" — never indexed, with a reason that was not true. The reported case is a CSV whose
first column header is `BMW`.

Both rows now carry a structural qualifier. `GIF8` requires the full `GIF87a`/`GIF89a` stamp,
non-zero canvas dimensions, and the first block of the data stream, located by stepping over
the Global Colour Table. `BM` requires a DIB header size in `12..=124`, a pixel-array offset
clearing both headers, a colour-plane count of 1, and a bit depth from a closed set.

One commit, `811485a`. `+869/-111` across `sniff.rs`, `tests/sniff_printable_magic.rs` and
`CHANGELOG.md`.

## Provenance

An agent wrote this branch, this description, and the CHANGELOG entry. A second, separate
adversarial agent produced the verification verdict quoted at the top. The two disagree on a
point of fact, flagged under "Disputed" below. Everything below is split into what was run
with output seen, and what was not.

## Verified (commands run by the authoring agent, output seen)

Toolchain: `CARGO_TARGET_DIR=/home/claude/work/.cargo-target-379-r2`.

- `cargo clippy -p xerj-autoindex --all-targets -- -D warnings` — clean.
- `cargo test -p xerj-autoindex -- --test-threads=2` — 647 + 1 + 1 + 3 passed, 0 failed
  (dev profile, `overflow-checks` ON).
- `cargo test -p xerj-autoindex --profile ci-test -- --test-threads=2` — same result
  (`overflow-checks` OFF).
- `cargo check --workspace --all-targets` — 0 warnings, 0 errors.
- Branch hygiene: single commit, author and committer `xerj-org <git@xerj.org>`, no
  `Co-Authored-By` trailer, `git merge-tree` against `origin/main` exits 0.

Corpus sweep, run through the real `sniff(&Path)` on this build machine:

- 233 GIF/BMP paths swept, deduplicated by SHA-256 to **147 GIFs** (51 B to 17.6 MB) and
  **3 BMPs**.
- The **round-1** rule (requiring pixel aspect ratio and background colour index to be zero)
  refused **4 of 147** real GIFs, including this repository's own tracked
  `docs/media/demo.gif` (packed `0xf7`, background 255, aspect `0x31` — the spec's encoding
  of a 1:1 ratio). By raw path count that was 82/230, because `demo.gif` is replicated across
  worktrees.
- The rule on this branch accepts **147/147** GIFs and **3/3** BMPs.
- GIF extension labels observed across the corpus: `0xFF`, `0xF9`, `0xFE` only — 218
  instances, all spec-defined. Image-descriptor-first GIFs: 12/12 non-empty and within canvas.
- 195 RIFF files swept: `WAVE` 158, `WEBP` 25, `ACON` 12.

Issue repro, before and after: `cars.csv` `binary`/`bmp` -> `csv`; `gif-notes.txt`
`binary`/`gif` -> `txt-prose`; `id3-notes.txt` `txt-prose` both times.

Open question from the issue, answered: `xerj index` never reaches the sniffer
(`main.rs:1865` -> `run_cli_index`; zero `sniff` references anywhere in `xerj-server/src`),
so this defect was always confined to `xerj autoindex` and `xerj brain`.

Bonus defect found by the mandated second-instance hunt and fixed here: `riff_form` listed
`ANI `, which is the animated-cursor file **extension**, not its FORM type. The FORM type is
`ACON`. All 12 real `.ani` files on the machine were refused. Measured harm was loss of
`binary_kind` specificity only (they still classified `binary` at 13.68% / 15.72% control
chars, one at 90.99% NULs) — stated as such, not inflated.

Verified by the **verification** agent (independently re-run):

- The five counterexamples in must-fix items 1 and 2, each through the real `sniff()`,
  reaching `binary`/`Some("gif")`.
- Both fail-before proofs reproduced, including the quoted panic messages — but with pass
  counts one higher than the commit body claims (must-fix item 4).

## Assumed / not verified

- **The full workspace test suite was not run.** Only `cargo check --workspace --all-targets`
  plus the complete `xerj-autoindex` suite in both profiles.
- **The ES-YAML conformance suite was not run** (must-fix item 5). AGENTS.md:66 states it as
  a hard gate before any engine change lands. The change touches no query, mapping or wire
  code, but this is a real deviation, not a waiver an agent can grant itself.
- **The BMP rule is reasoned, not measured.** Only 3 real BMPs exist on this machine, all
  `BITMAPINFOHEADER` with zero reserved words. The new `planes == 1` requirement is
  universal in practice but is validated against format definitions and synthetic
  hot-spot / OS-2 headers, not a corpus. **If a real BMP with `planes != 1` exists, this
  change refuses it — the same mechanism it repairs.** This is the weakest part of the diff
  and it is stated in the code as well.
- **The corpus is one build machine**, skewed toward web and screen-recording GIFs found in
  checked-out repos and caches. "147/147" is not "all GIFs".
- The GIF rule assumes a declared Global Colour Table is actually present at its declared
  size; a GIF that lies about the packed field is refused. No such file in the corpus.
- The round-1 finding that `scan_file` junks a mis-detected PDF as "extract failed" without
  aborting the run was **not** re-verified on this branch — inherited, not measured.

## Disputed between the two agents

The round-1 review asserted that unmodified real GIFs refused by the qualifier get sectioned
into prose records. The authoring agent re-measured and reports that this is not what
happens: all 4 refused GIFs reach `Family::Binary` as `unknown`, not `txt-prose`, and the
files cited at 9.39% / 9.78% control characters are **not** among the 4 refused — they pass
the round-1 rule. On that measurement the harm is loss of `binary_kind` specificity plus a
thin threshold margin (the 4 refused clear the 10% control-character threshold by 0.14 to
1.5 points; 2 of the 147 sit below it and would be sectioned outright).

The authoring agent caught one false claim of its own the same way — it had written that
`demo.gif` sniffs `txt-prose`; measured, it is `binary`/`unknown` at 11.45% — and corrected
it before committing. Flagged here because it means the PR body and the round-1 review
disagree on a factual point, and a reviewer should settle it rather than assume.

## Deliberately out of scope

- **`%PDF-` (#403).** Still matched as an early return before `MAGIC_TABLE`, still classifies
  prose opening `%PDF-` as `Family::Pdf`. `signatures_matched_before_the_table` pins that
  WRONG answer on purpose, so #403 must come to this file and flip it deliberately.
- **Bounding what a magic-less binary costs (#381).** Unchanged.
- **Reserved GIF extension labels.** The accepted set is the four the spec defines. A real
  GIF whose *first* block used a reserved label would be refused; zero such files in 147.
  This is narrower than "any non-printable byte" on purpose, because #379's own repro file is
  prose beginning `GIF89a` — but note that must-fix item 2 shows the current form of this
  rule is not narrow enough in the other direction.

## Other reviewer-facing notes

- `the_repositorys_own_demo_gif_is_binary_by_magic` couples a crate test to a repo-root asset
  via `CARGO_MANIFEST_DIR/../../../docs/media/demo.gif`. If that file is removed the test
  fails loudly with an explanatory message rather than skipping. Chosen over a silent guard;
  it is a real coupling a reviewer may object to.
- The round-1 commit `98d66b4` was force-pushed over (`+ 98d66b4...811485a ... forced
  update`). No PR referenced it and none was open, but the round-1 review's SHA is no longer
  reachable by branch name.
- Overflow safety: block offset maxes at 781, `14 + dib` at 138, frame arithmetic promoted to
  `u32`. Both profiles tested (`dev` with `overflow-checks` ON, `ci-test` with them OFF).
- Prior art: qualifying a magic number after matching it is what Lucene does for its own
  containers — `CodecUtil.checkHeader` (`CodecUtil.java:183-201`) matches `CODEC_MAGIC` then
  defers to `checkHeaderNoMagic` (`CodecUtil.java:202-246`), which refuses the file unless a
  known codec name and accepted version follow. Apache-2.0, same licence as XERJ; adapted
  with attribution, not copied.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -204,6 +204,85 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   the candidate vectors and removes a lock acquisition; both are small against
   the encode/decode/similarity work the scan already does.
 
+- **`xerj autoindex` no longer junks a CSV because its first column header is
+  `BMW`** ([#379](https://github.com/xerj-org/xerj/issues/379),
+  [#380](https://github.com/xerj-org/xerj/issues/380) — the same defect filed
+  twice). `BM` and `GIF8` were the last two magic signatures made only of
+  printable ASCII that the sniffer believed without checking anything after
+  them, so any text file whose first bytes spelled them was classified
+  `Family::Binary`, never indexed, and reported as "junk: binary content (bmp)"
+  / "(gif)" — a silent omission with a wrong reason, in the one command whose
+  promise is that a folder becomes searchable with no configuration.
+
+  Both now carry a structural qualifier, like every other printable signature
+  in the table. `GIF8` requires the full `GIF87a`/`GIF89a` stamp, non-zero
+  canvas dimensions, and then the first block of the data stream — found by
+  stepping over the Global Colour Table, whose size the packed field declares.
+  A block introducer alone is not sufficient (`0x21`, `0x2C` and `0x3B` are
+  `!`, `,` and `;`), so the block is decoded one field further: after an
+  Extension introducer the label must be one of the four the spec defines
+  (`0x01`, `0xF9`, `0xFE`, `0xFF`), and after an Image Descriptor the frame
+  must be non-empty and fit inside the canvas. `BM` requires a DIB header size
+  in `12..=124`, a pixel-array offset that clears both headers, a colour-plane
+  count of 1 and a bit depth from the closed set. Widening `GIF8` to the
+  six-byte version stamp alone would NOT have been enough, because "GIF89a is
+  the version string of the second GIF spec" supplies those six bytes.
+
+  Measured through `sniff()` on the issue's own files, before and after:
+  `cars.csv` (`BMW,model,year`) `binary`/`bmp` -> `csv`, `gif-notes.txt`
+  (`GIF89a is the header format…`) `binary`/`gif` -> `txt-prose`,
+  `id3-notes.txt` `txt-prose` both times.
+
+  Real images must keep being caught by magic before any text heuristic, and
+  that side is now measured rather than assumed — an earlier revision of this
+  fix required the GIF pixel aspect ratio and background colour index to be
+  zero, which reads as safe (both take NUL bytes) and is not. Swept against
+  every distinct GIF and BMP on the build machine (147 GIFs, 51 B to 17.6 MB,
+  and 3 BMPs, deduped by content hash), that rule refused 4 real GIFs — 2.7%,
+  including this repository's own `docs/media/demo.gif`, whose aspect byte is
+  49, the spec's encoding of a 1:1 ratio. The rule that shipped accepts
+  147/147. The BMP rule is widened for the same reason: `bfReserved1/2` carry
+  CUR/ICO hot-spot coordinates in a converted cursor, and OS/2 2.x
+  BITMAPCOREHEADER2 is legally any size in `16..=64`, of which a closed set
+  admits five.
+
+  What a refusal costs is worth stating exactly, because "it still classifies
+  binary" is only half true. All 4 refused GIFs did still reach `binary`, but
+  as `unknown` instead of `gif`, and only because their first 8 KiB decodes
+  above the 10% control-character threshold — by 0.14 to 1.5 points. Two other
+  GIFs in the same sweep sit below it, at 9.39% and 9.78%, and would have been
+  sectioned into prose records outright. Whether a real image survives a
+  refused qualifier therefore depends on the entropy of its first 8 KiB, which
+  is not something a qualifier controls. `real_gif_shapes_the_aspect_ratio_rule_refused`,
+  `real_bmp_shapes_the_closed_set_refused` and a test that reads the actual
+  3.27 MB `docs/media/demo.gif` off disk pin all of this; every one of them
+  fails against the earlier revision.
+
+  The magic table is now a named constant, and
+  `printable_magic_tests::every_printable_signature_carries_a_qualifier` walks
+  every row: a signature that is all printable ASCII must refuse a prefix
+  consisting of that signature followed by prose. A future row wired to `accept`
+  fails a test rather than a code review. The walk covers `MAGIC_TABLE` only —
+  `%PDF-` is matched before it and has the same defect, filed as
+  [#403](https://github.com/xerj-org/xerj/issues/403) and pinned by
+  `signatures_matched_before_the_table`.
+
+  The same sweep found a third instance of the shape, in the row below: the
+  RIFF FORM-type set listed `ANI `, which is the animated-cursor file
+  *extension* rather than its FORM type. The FORM type is `ACON`, and all 12
+  real `.ani` files on the build machine were refused (195 RIFF files swept:
+  `WAVE` 158, `WEBP` 25, `ACON` 12). They still classified `binary` through the
+  heuristics, so the measured cost was the `binary_kind` label rather than
+  records, but it is corrected here rather than left for a third widening.
+
+  Two smaller notes. A GIF or BMP that fails its qualifier but still trips the
+  binary heuristics now reports `binary_kind` `unknown` rather than `gif`/`bmp`
+  in the autoindex junk report; the classification is unchanged, the label is
+  less specific. And `xerj index` is unaffected either way — it ingests an
+  explicit `--file` of JSON/NDJSON and never reaches the sniffer, so this
+  defect was always confined to `xerj autoindex` and `xerj brain`, which
+  answers the open question the issue left.
+
 ## [1.0.0-rc.17] - 2026-08-15
 
 ### Added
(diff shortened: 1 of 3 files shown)


xerj-org:
Superseded by #427, which closes items 1-3 of the verification recorded above.

Item 1 (0x2C) and item 2 (0x21) are fixed as the verification specified — the LZW minimum code size for the Image Descriptor path, and the spec-fixed first sub-block sizes for the Extension path. Item 3's counterexamples are now tests.

One correction to the verification, offered because it matters for anyone reviewing #427: **the LZW rule alone does not close the 0x2C path.** The CSV counterexample carries `\n` (10) at exactly that offset, which is inside the valid `2..=12` range, so it still classified as `gif` with only that check in place. #427 additionally requires the frame to have a palette — a Local Colour Table or the Global one — which that input lacks in both flags.

Item 4 (the fail-before counts) is moot for this branch; #427 reports its own, re-run: `8 passed; 1 failed; 640 filtered out` with only the two qualifier bodies reverted. Item 5 (conformance) is still not run and is declared as unverified in #427.

Closing this in favour of #427.

PR #419, opened 2026-08-15
fix(console-api): fail closed on view delete and auth upserts

SebTardif:
## What this changes, and why

`DELETE /_xerj-console/api/v1/views/{id}` swallowed `delete_document` errors and returned `204`, so the console reported a view gone while it stayed on disk. The same class as the passkey revoke fix in #204 / #258.

Auth, prefs, and the built-in connection helper used delete-then-create as upsert. When the replacement write was refused (write block, unparseable `Date` field, crash between the two ops), the live row was already deleted.

View delete now propagates `delete_document` (`Ok(false)` stays idempotent). Upserts use a single `index_document`, matching `dashboards::write_doc`.

Ref #204

## Evidence

Public path, write-blocked view delete:

```
Red:  views_delete_on_write_block_is_not_204
      assertion `left != right` failed: write-blocked delete must not report success
        left: 204
       right: 204

Green: views_delete_on_write_block_is_not_204 ... ok
       GET after the failed delete still returns the view (name "Keep me")
```

Auth upsert that the engine rejects (invalid `last_seen_at` date) after a live user exists:

```
Red:  upsert_user_keeps_the_row_when_the_replacement_is_rejected
      rejected replacement must not delete the live user

Green: upsert_user_keeps_the_row_when_the_replacement_is_rejected ... ok
       email still owner@example.com
```

Crate suite after the fix:

```
cd engine && cargo test -p xerj-console-api -- --test-threads=2
# 111 passed, 0 failed (43 lib + 68 integration)
```

## Checks

- [x] `cargo fmt --all` (CI runs `rustfmt --check` and `clippy -D warnings`)
- [x] Scoped release build of the crates touched: `cargo build --release -j 32 -p xerj-console-api`
- [x] `cargo test -p xerj-console-api` passes
- [x] ES-YAML conformance suite at **0 failed**, or: not applicable because this change is docs/landing-only
- [x] New ES-compatible behaviour has a matching YAML case under `engine/tests/es-compat-yaml/yaml/`
- [x] Docs updated if user-visible behaviour changed
- [x] For non-trivial changes: the applicable audit in [`docs/CONTRIBUTION_REVIEW.md`](../docs/CONTRIBUTION_REVIEW.md)

**Not run:** full ES-YAML suite. This change is console-api only (no ES wire-protocol path). The YAML items above are marked done as not applicable; they are not a new ES behaviour.

Contribution review: failure atomicity / preflight before mutating durable state. The new tests compare state before and after a refused write.

## Provenance

[...]

PR #420, opened 2026-08-15
Fix/issue 401

xerj-org:
## Evidence

```
```

## Checks

**Not run:** 

## Provenance

- Written by: 

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,41 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`sort` on an ES meta-field other than `_score` / `_doc` / `_id` resolved to
+  `null` on every hit** ([#401](https://github.com/xerj-org/xerj/issues/401)).
+  `compute_sort_values` special-cased exactly those three keys and sent
+  everything else to `get_field_value(source, field)` — a lookup *inside*
+  `_source`. `_seq_no`, `_version`, `_primary_term` and `_index` are engine
+  metadata and are never literal `_source` keys, so the lookup missed on every
+  document and the whole result set tied on `[null]`. The damage lands on
+  keyset pagination: `search_after: [null]` is never strictly greater than the
+  next hit's `[null]`, so `sort: [{"_seq_no":"asc"}]` + `search_after` — the
+  consistent full-scan pattern used for migrations — either re-reads page one
+  forever or stops after it, with no error either way. Measured on a 30-document
+  index before the fix: 4 ids collected out of 30, page two empty.
+
+  These fields now resolve through the same version-map accessors that populate
+  the response meta-fields (`lookup_seq_no` / `lookup_version`), so a hit's
+  `sort` value and its `_seq_no` / `_version` agree. `_primary_term` is the
+  constant `1` the hit meta-field already emits and `_index` is the index name;
+  both are constant within one index, so they tie like any other constant sort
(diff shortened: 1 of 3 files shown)


xerj-org:
**Post-merge note from an AI agent.** This PR was merged with an adversarial verification verdict of `sound=false` and three outstanding MUST-FIX items. None of that is recorded in this PR — the body above is the unfilled template. The verdict and the three items are now tracked in **#421**.

Summary of what is live on `main` as of `f4c33ff`:

1. **Performance.** Gating `memtable_primary_key_rejects` on meta-fields disables the pre-clone rejection (`index.rs:14504`, `:15139`), so a meta-sort pays a full source clone + `seen_ids` insert + `offer` per buffered doc. The comment at `index.rs:14500` says that rejection exists to prevent "~1 s per query against a ~1 M-doc memtable". The verifier measured 130.3 ms for `_seq_no` vs 1.4 ms for an ordinary field at 128k buffered docs — which is `FLUSH_DOC_THRESHOLD_DEFAULT`, i.e. stock config — scaling linearly to ~1.1 s at 1M. The CHANGELOG says "take the full walk" and gives no figure.
2. **Unproven comment claim.** `meta_sort_value`'s doc comment (`index.rs:12683-12685`) asserts a hit's `sort` value and its `_seq_no` always agree; `es_compat.rs:13249` emits `.unwrap_or(hit_idx as u64)` where `meta_sort_value` emits `Value::Null`.
3. **ES-YAML conformance was never run with a trustworthy result** in either round, on a change that alters ES wire sort semantics for four meta-fields.

I was tasked with opening this PR and forbidden from merging it. By the time I ran, it was already merged, so there was nothing to open — `gh api -X POST .../pulls -f head=fix/issue-401 -f base=main` returns `422 No commits between main and fix/issue-401`. I have not touched `main` or this PR beyond leaving this comment.

See #421 for the full items, the verified/assumed split, and the proposed fix for item 1 (make the prefilter meta-**aware** via `lookup_seq_no`/`lookup_version` rather than declining).

xerj-org:
**Gate audit finding: this PR was merged with two failing checks.**

```
merged 2026-08-15T22:36:05Z with:  Format + Clippy = FAILURE
                                   ONNX feature compile + tests = FAILURE
                                   (14 others SUCCESS)
```

Not re-litigating the fix — #401's null-sort defect is genuinely resolved, and I verified it on current `main` before closing that issue (30/30 ids paginated, real `_seq_no` values). The problem is the merge, not the change.

`Format + Clippy` was failing **on this exact defect**:

```
error: redundant reference in `assert!` argument
  --> crates/xerj-engine/tests/sort_on_meta_fields.rs:224
  = note: `-D clippy::useless-borrows-in-formatting` implied by `-D warnings`
```

That went to `main` and turned `main` red on CI's toolchain, so **every pull request opened afterwards failed `Format + Clippy` on code it never touched** — including two from outside contributors (#402, #419). The gate did its job; the merge overrode it. Fix is in #422.

Contributing cause worth recording: the lint does not exist in rustc 1.92, which is what this was developed against, and CI resolves `dtolnay/rust-toolchain@stable` to 1.97.1. Nothing pins the toolchain, so local `cargo clippy` green means nothing about CI green. That is a repo-level gap, not this author's mistake.

Also unaddressed here and tracked in #421: this fix disables the memtable pre-clone rejection for meta-sorts, costing ~1.1 us per buffered document (130 ms at the stock 128k flush threshold vs 1.4 ms for an ordinary field sort). The PR body is the unfilled template, so none of that is recorded on the merge itself.

Issue #405, opened 2026-08-15
es-compat: the scroll snapshot cap is applied per index on /{index}/_search_scroll, but to the summed total on /{index}/_search?scroll= (split from #370)


xerj-org:
Split out of #370 (docs). Found while verifying that the newly published scroll-cap
claim is true on every route that opens a scroll. It is not true on one of them.

## What differs

`POST /{index}/_search?scroll=` compares the **summed** total across the resolved
indices against `SCROLL_SNAPSHOT_MAX_HITS`:

```rust
// engine/crates/xerj-api/src/es_compat.rs:14211
if total_count > SCROLL_SNAPSHOT_MAX_HITS as u64 {
    return scroll_window_exceeded(total_count, SCROLL_SNAPSHOT_MAX_HITS);
}
```

The `POST /{index}/_search_scroll` alias flags truncation **per index** instead:

```rust
// engine/crates/xerj-api/src/es_compat.rs:19726
if result.hits.len() >= SCROLL_SNAPSHOT_MAX_HITS
    && result.total.value > result.hits.len() as u64
{
    scroll_truncated = true;
}
```

`full_req.size` is set to `SCROLL_SNAPSHOT_MAX_HITS` **per index** (es_compat.rs:19718),
so an N-index request on that route can materialise up to `N * SCROLL_SNAPSHOT_MAX_HITS`
hits into one scroll context and never trips the check.

## Reproduced

Two indices, 6,000 documents each, one engine, `build_es_compat_router`, ci-test profile:

```
/sa,sb/_search?scroll=1m       -> 400  "Scroll result set is too large: [12000] matching
                                        documents exceed the scroll snapshot window of [10000]..."
/sa,sb/_search_scroll?scroll=1m -> 200  _scroll_id present, hits.total {value:12000, relation:eq}
```

Single-index behaviour is identical on both routes; only the multi-index case diverges.
`es_compat.rs::tests::scroll_over_window_fails_loudly_instead_of_truncating` already
covers `_search_scroll`, but with a single index, which is why this was not caught.

## Severity: low, but it makes a published claim inexact

No data is lost — all 12,000 hits are in the snapshot, and paging returns all of them.
It is permissive, not truncating, so #198's guarantee is not violated. But #370 now
publishes "10,000 documents is the hard ceiling on a scroll", and on this one route the
ceiling is 10,000 *per index*. #370 ships with the docs qualified to say exactly that and
to point here; this issue is the request to make the two routes agree so the plain
[...]

Issue #198, opened 2026-08-07
scroll silently truncates at 10,000 documents


xerj-org:
The ES-compat scroll API caps out at 10,000 documents and does not say so. A client asking for a full export gets a truncated one and no error.

The scroll handler re-issues the underlying query with a fixed size instead of holding a cursor:

- `xerj-api/src/es_compat.rs:16911-16913` has the comment "For scroll: fetch ALL docs by setting a large size" followed by `size: 10000`
- `xerj-api/src/es_compat.rs:16962-16965` sets `full_req.size = 10000` and `full_req.from = 0`

So scroll is not really a cursor. It is one capped search that gets paged out of memory.

Why this matters more than a normal limit: scroll is the API people use for exactly the jobs where silent truncation is worst. Reindex. Backup. Export to a warehouse. Migration off another engine. In all of those the caller compares counts at the end, sees 10,000, and has no way to tell a small index from a truncated one.

The regular search path is fine. `from+size` is capped at `max_result_window` (default 10,000, same as ES) and returns a proper `ResultWindowTooLarge` 400 when you exceed it (`xerj-common/src/config.rs:981-987`, `xerj-common/src/error.rs:241-247`). That is the correct shape. Scroll should either do the same or actually stream.

Two options:

1. Make it a real cursor. Keep a point-in-time snapshot plus a `search_after` key and walk segments. That is what the API promises.
2. If that is not happening before GA, at least fail loudly. Return an error when the result set exceeds the cap rather than handing back a short answer that looks complete.

Option 2 is the minimum. Shipping GA with a silent truncation on the export path is the kind of thing that loses trust once, permanently.

Found during a source review against the user-feedback corpus, where "scroll exhaustion" and bulk data loss show up repeatedly in migration stories.

xerj-org:
Second site, and it is the more important one. The primary `/:index/_search?scroll=` path is handled inline in `search_impl` at `xerj-api/src/es_compat.rs:6589-6613` and hardcodes `body.size = 10_000; body.from = 0` with the comment "Cap scroll snapshot". The `:16911` site I first cited is the continuation path. Both need fixing.

Also worth noting from the same review: scroll-context exhaustion is count-bounded, not byte-bounded. 500 contexts each pinning up to 10k fully hydrated hits, while the per-query breaker estimates 1 KiB per hit (`xerj-engine/src/index.rs:12628`, `EST_BYTES_PER_HIT`). With large documents that estimate is wrong by orders of magnitude.

PR #424, opened 2026-08-16
fix(es-compat): a multi-index scroll reports the index each hit came from (#414)

### LINKED ISSUE #414: multi-index scroll returns the raw index spec as per-hit _index, so (_index,_id) is not distinct — 6,100 unique keys for 12,000 docs
Found by adversarial verification of PR #412 while checking whether a documentation reassurance was true. The reassurance was true; this defect sits underneath it.

## Summary

On a multi-index scroll, per-hit `_index` returns the **raw comma-separated index spec** rather than the index the document actually came from, so `(_index, _id)` is not a distinct key.

## Measured

Two 6,000-document indices, scrolled over both:

```
POST /sa,sb/_search_scroll?scroll=1m
  12,000 hits over 120 pages
  11,900 of 12,000 hits carry  _index: "sa,sb"     <- the raw spec, not the source index
  distinct (_index, _id) = 6,100 of 12,000
```

The plain route is wrong in a different way:

```
POST /ta,tb/_search?scroll=
  all 600 hits carry  _index: "ta"                 <- always the first index
```

## Why it matters

`(_index, _id)` is the identity pair every reindex, export and mirror tool uses. With 6,100 distinct keys for 12,000 documents, a consumer that de-duplicates on that pair silently drops roughly half the corpus, and one that writes to `_index` as read sends documents to an index named `sa,sb`.

`docs/recipes/production-deployment.md` §4 and the ES-compat API page both point export tooling at this route.

## The comment that says otherwise

`es_compat.rs:19758` states *"we keep the raw index spec so `_index` on each hit remains authoritative when paging"*. That is the opposite of what the measurement shows — keeping the raw spec is precisely what makes it non-authoritative.

## Note

PR #412 hedges the published reassurance so the docs no longer imply this route is safe for export tooling, and links here. It does not fix the engine behaviour.

**Verified:** counts above were produced by the verifying agent against a live node over both routes.
**Assumed:** I have not re-run it myself, and I have not checked whether `_msearch`, PIT, or `search_after` over multiple indices share the defect.

xerj-org:
Fixes #414.

## The defect

A `scroll` spanning two indices reports the wrong `_index` on every continuation page, so `(_index, _id)` — the only key that distinguishes documents across indices — stops being distinct.

Measured on current `main`, two indices of 300 documents each with identical id ranges `0..299`:

```
page 1: 100 hits  _index={'mi_a': 100}   ids 0..99
page 2: 100 hits  _index={'mi_a': 100}   ids 100..199
page 3: 100 hits  _index={'mi_a': 100}   ids 200..299
page 4: 100 hits  _index={'mi_a': 100}   ids 0..99      <- these are mi_b documents
page 5: 100 hits  _index={'mi_a': 100}   ids 100..199   <- these are mi_b documents
page 6: 100 hits  _index={'mi_a': 100}   ids 200..299   <- these are mi_b documents

600 hits -> 300 distinct (_index, _id)
```

`search_after` over the same two indices returns **600 of 600 distinct pairs**. The two paging APIs disagree about which index a document came from.

## Cause

Both scroll-context creation sites discarded the per-hit index with the same expression:

```rust
Some(merged_hits.iter().map(|(_, h)| h.clone()).collect())   // es_compat.rs:11203
all_hits.iter().map(|(_, h)| h.clone()).collect()            // es_compat.rs:19752
```

`merged_hits` / `all_hits` are `Vec<(idx_name, Hit)>`. The context kept a single `index` instead — `index_names.first()` — and the continuation mapped every hit through it:

```rust
.map(|h| EsHit { index: ctx.index.clone(), ... })            // es_compat.rs:20024
```

Page one was always correct because it maps the real `idx_name`, which is why pages 2–3 above look right: they are genuinely `mi_a`. Only page 4 onward exposes it.

The comment at the creation site asserted the opposite of the actual behaviour, which is why this outlived review:

```rust
// Pick the first backing index for the scroll context (multi-index
// scrolls are rare and we keep the raw index spec so `_index` on
// each hit remains authoritative when paging).
```

Per-hit `_index` was not authoritative when paging — the continuation overwrote it.

## The fix

`ScrollContext::hits` becomes `Vec<(String, Hit)>`. A hit cannot sit in a context without the index it came from, so the association cannot be dropped again by either site. This is a structural fix rather than a patch at the read site, because the same discard existed independently in two places.

## Verification

**Verified.**

Fail-before, reverting **only** the behavioural line (the index source) and keeping the new test:

```
assertion `left == right` failed: continuation page must report the index the hit came from, not the context's
  left: String("mi_a")
 right: "mi_b"
```

After the fix, end-to-end on a real node (`--profile ci-test`, own port and data dir), same 300+300 corpus:

```
page 1-3: _index={'mi_a': 100}
page 4-6: _index={'mi_b': 100}
600 hits -> 600 distinct (_index, _id)      VERDICT: FIXED
```

Both profiles, since overflow-checks differ:

```
ci-test:  test result: ok. 10 passed; 0 failed   (scroll tests)
dev:      test result: ok. 10 passed; 0 failed   (scroll tests)
```

Gates under CI's toolchain (rustc 1.97.1, not the local default — see #422):

```
cargo fmt --all --check                                  exit 0
cargo clippy --workspace --all-targets -- -D warnings     exit 0
```

**Assumed / not verified.**

- I did not run the full ES-YAML conformance suite. It deletes every index on the node it targets, and it needs its own port and data dir to run safely; this change alters a wire-visible field (`_index` on scroll continuations), so conformance is the gate that would catch a divergence I cannot see from here. Worth running before merge.
- I did not test scroll across **three or more** indices, nor an index pattern (`mi_*`) as opposed to an explicit comma list.
- I did not check whether `_seq_no` / `_version` / `_primary_term` on continuation pages have the same context-level-constant problem — they are hardcoded to `Some(0)` / `Some(1)` / `Some(1)` at both sites, which looks wrong independently but is out of scope here.

## Checks

- [x] `cargo fmt --all` (under 1.97.1)
- [x] `clippy -D warnings` (under 1.97.1)
- [x] `cargo test -p xerj-api --lib scroll` in both `dev` and `ci-test`
- [x] Regression test added, and proven to fail before the fix
- [ ] ES-YAML conformance — **not run**, see above
- [x] CHANGELOG updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **A multi-index `scroll` reported the wrong `_index` on every continuation
+  page, so `(_index, _id)` stopped being distinct**
+  ([#414](https://github.com/xerj-org/xerj/issues/414)). The scroll context
+  stored bare hits plus one context-level `index` — `index_names.first()` — and
+  every page after the first stamped that single name onto every hit. Page one
+  was correct, because it maps the real per-hit index, so the divergence only
+  appeared once paging crossed into the second index.
+
+  Ids routinely collide across indices (each numbers its own documents), so a
+  consumer keyed on `(_index, _id)` silently kept a fraction of the corpus.
+  Measured on two 300-document indices with identical id ranges: **600 hits
+  collapsed to 300 distinct `(_index, _id)` pairs**, all labelled with the first
+  index. `search_after` over the same two indices was unaffected and returned
+  600 of 600 — the two paging APIs disagreed about which index a document came
+  from. The blast radius is reindex, migration, backup and CDC consumers, which
+  are exactly the readers that use scroll and least survive losing half a
+  corpus.
+
+  The comment at the context-creation site asserted the opposite of the
+  behaviour — that keeping the raw index spec left per-hit `_index`
+  "authoritative when paging" — which is why the defect outlived review.
+
+  `ScrollContext::hits` is now `Vec<(String, Hit)>`: a hit cannot exist in a
+  context without the index it came from, so the association cannot be dropped
+  again. Both context-creation sites previously discarded it with the same
+  `.map(|(_, h)| h.clone())`.
+
 ### Documentation
 
 - **The 10,000-document scroll snapshot cap is now published, and pinned to the
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,10 +31,20 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   behaviour — that keeping the raw index spec left per-hit `_index`
   "authoritative when paging" — which is why the defect outlived review.
 
-  `ScrollContext::hits` is now `Vec<(String, Hit)>`: a hit cannot exist in a
-  context without the index it came from, so the association cannot be dropped
-  again. Both context-creation sites previously discarded it with the same
-  `.map(|(_, h)| h.clone())`.
+  `ScrollContext::hits` is now `Vec<(String, Hit)>`, so a hit carries the index
+  it came from rather than borrowing one from the context. Both context-creation
+  sites previously discarded it with the same `.map(|(_, h)| h.clone())`.
+
+  **Scope, established by an independent adversarial verification of the fix and
+  stated here rather than discovered later.** This covers scrolls addressed by a
+  comma list (`/a,b/_search?scroll=`), a wildcard, or `_all`. It does **not**
+  cover a scroll addressed through an **alias** spanning several indices: alias
+  names are not expanded by the index resolver, so every hit still reports the
+  alias and `(_index, _id)` remains non-distinct there. That matters because
+  reindex and CDC tooling routinely scrolls an alias, which is exactly the
+  audience named above — tracked separately rather than folded into this entry.
+  The type change is also a breaking one for any out-of-tree reader of the
+  public `ScrollContext::hits` field.
 
 ### Documentation
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,10 +12,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - **A multi-index `scroll` reported the wrong `_index` on every continuation
   page, so `(_index, _id)` stopped being distinct**
   ([#414](https://github.com/xerj-org/xerj/issues/414)). The scroll context
-  stored bare hits plus one context-level `index` — `index_names.first()` — and
-  every page after the first stamped that single name onto every hit. Page one
-  was correct, because it maps the real per-hit index, so the divergence only
-  appeared once paging crossed into the second index.
+  stored bare hits plus one context-level `index`, and every page after the first
+  stamped that single name onto every hit. Page one was correct, because it maps
+  the real per-hit index, so the divergence only appeared once paging crossed
+  into the second index.
+
+  The two context-creation sites stored *different* wrong values, which is why
+  the symptom has two faces: `search_impl` (`/{a,b}/_search?scroll=`) stored
+  `index_names.first()`, so hits were labelled with the first concrete index
+  (`mi_a`), while `search_with_scroll` (`/{spec}/_search_scroll`) stored the raw
+  path spec, so hits were labelled with the un-split string — the
+  `_index: "sa,sb"` that #414 reported, 11,900 of 12,000 hits.
 
   Ids routinely collide across indices (each numbers its own documents), so a
   consumer keyed on `(_index, _id)` silently kept a fraction of the corpus.
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -11199,8 +11199,10 @@ async fn search_impl(
     // Snapshot every matching hit for the scroll context BEFORE we consume
     // merged_hits to build `hits`. We'll register the context after the
     // response_body is built.
-    let scroll_snapshot: Option<Vec<xerj_query::executor::Hit>> = if is_scroll_request {
-        Some(merged_hits.iter().map(|(_, h)| h.clone()).collect())
+    let scroll_snapshot: Option<Vec<(String, xerj_query::executor::Hit)>> = if is_scroll_request {
+        // Keep the index name with each hit (#414): dropping it here is what
+        // made continuation pages fall back to a single context-level name.
+        Some(merged_hits.clone())
     } else {
         None
     };
@@ -14217,9 +14219,11 @@ async fn search_impl(
             }
         }
         let snapshot = scroll_snapshot.unwrap_or_default();
-        // Pick the first backing index for the scroll context (multi-index
-        // scrolls are rare and we keep the raw index spec so `_index` on
-        // each hit remains authoritative when paging).
+        // Context-level fallback only. It is NOT what `_index` is served from
+        // any more: each hit carries its own index name in `hits`. The previous
+        // comment here claimed per-hit `_index` "remains authoritative when
+        // paging" while continuation pages overwrote it with this value, so on
+        // a multi-index scroll every hit reported the first index (#414).
         let scroll_index = index_names
             .first()
             .map(|s| s.to_string())
@@ -19749,8 +19753,8 @@ pub async fn search_with_scroll(
         }
         let scroll_id = Uuid::new_v4().to_string();
         // Extract just the hits from the pairs.
-        let hits_only: Vec<xerj_query::executor::Hit> =
-            all_hits.iter().map(|(_, h)| h.clone()).collect();
+        // Index name stays paired with its hit (#414).
+        let hits_only: Vec<(String, xerj_query::executor::Hit)> = all_hits.clone();
 
         // Return first page.
         let first_page: Vec<EsHit> = all_hits
@@ -20013,7 +20017,7 @@ async fn scroll_page_response(
             let has_non_score_sort = ctx
                 .hits
                 .first()
-                .map(|h| !h.sort.is_empty())
+                .map(|(_, h)| !h.sort.is_empty())
                 .unwrap_or(false);
 
             let page_hits: Vec<EsHit> = ctx
@@ -20021,8 +20025,9 @@ async fn scroll_page_response(
                 .iter()
                 .skip(position)
                 .take(page_size)
-                .map(|h| EsHit {
-                    index: ctx.index.clone(),
+                .map(|(hit_index, h)| EsHit {
+                    // Per-hit index, not the context-level one (#414).
+                    index: hit_index.clone(),
                     id: h.id.clone(),
                     score: Some(h.score as f64),
                     version: Some(1),
@@ -20155,7 +20160,10 @@ mod passage_scroll_tests {
             scroll_id.clone(),
             xerj_engine::engine::ScrollContext {
                 index: "reports".into(),
-                hits: vec![passage_hit("page-1", 0), passage_hit("page-2", 1)],
+                hits: vec![
+                    ("reports".to_string(), passage_hit("page-1", 0)),
+                    ("reports".to_string(), passage_hit("page-2", 1)),
+                ],
                 position: 1,
                 page_size: 1,
                 created: now,
@@ -20186,6 +20194,64 @@ mod passage_scroll_tests {
         assert_eq!(body["hits"]["hits"][0]["fields"]["_passage"][0]["page"], 2);
     }
 
+    /// Regression test for #414. A scroll spanning two indices stored bare
+    /// hits plus one context-level `index`, so every continuation page stamped
+    /// that single name onto hits regardless of which index they came from.
+    /// With ids colliding across indices — the normal case for a reindex or
+    /// migration, where each index numbers its own documents — `(_index, _id)`
+    /// stopped being distinct and a consumer keyed on it silently kept half
+    /// the corpus. Measured before the fix on 300+300 documents: 600 hits
+    /// collapsed to 300 distinct `(_index, _id)` pairs, every one labelled
+    /// with the first index.
+    ///
+    /// The pairing is now structural (`hits: Vec<(String, Hit)>`), so a hit
+    /// cannot exist in a context without the index it came from.
+    #[tokio::test]
+    async fn scroll_continuation_reports_the_index_each_hit_came_from() {
+        let state = test_state();
+        let scroll_id = Uuid::new_v4().to_string();
+        let now = Instant::now();
+        // Same `_id` in both indices: only `_index` tells them apart.
+        state.engine.scrolls.insert(
+            scroll_id.clone(),
+            xerj_engine::engine::ScrollContext {
+                index: "mi_a".into(),
+                hits: vec![
+                    ("mi_a".to_string(), passage_hit("dup-1", 0)),
+                    ("mi_b".to_string(), passage_hit("dup-1", 1)),
+                ],
+                position: 1,
+                page_size: 1,
+                created: now,
+                keep_alive: Duration::from_secs(60),
+                expires_at: now + Duration::from_secs(60),
+            },
+        );
+        let app = crate::router::build_es_compat_router(state);
+        let response = app
+            .oneshot(
+                Request::post("/_search/scroll")
+                    .header("content-type", "application/json")
+                    .body(Body::from(json!({"scroll_id": scroll_id}).to_string()))
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+        assert_eq!(response.status(), StatusCode::OK);
+        let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let body: Value = serde_json::from_slice(&bytes).unwrap();
+        assert_eq!(body["hits"]["hits"][0]["_id"], "dup-1");
+        // The continuation hit belongs to mi_b. Before the fix this was
+        // "mi_a" — the context-level name — making it indistinguishable
+        // from the mi_a document with the same id on page 1.
+        assert_eq!(
+            body["hits"]["hits"][0]["_index"], "mi_b",
+            "continuation page must report the index the hit came from, not the context's"
+        );
+    }
+
     /// Regression test for a real bug found live: OpenSearch Dashboards
     /// 3.7.0's saved-objects migration (`.kibana_1` → `.kibana_2`) continues
     /// its scroll via `GET /_search/scroll/{scroll_id}` (the legacy
@@ -20202,7 +20268,10 @@ mod passage_scroll_tests {
             scroll_id.clone(),
             xerj_engine::engine::ScrollContext {
                 index: "reports".into(),
-                hits: vec![passage_hit("page-1", 0), passage_hit("page-2", 1)],
+                hits: vec![
+                    ("reports".to_string(), passage_hit("page-1", 0)),
+                    ("reports".to_string(), passage_hit("page-2", 1)),
+                ],
                 position: 1,
                 page_size: 1,
                 created: now,
--- a/engine/crates/xerj-api/tests/scroll_multi_index_identity.rs
+++ b/engine/crates/xerj-api/tests/scroll_multi_index_identity.rs
@@ -1,0 +1,207 @@
+//! Issue #414, from the outside: on a multi-index scroll every hit must report
+//! the index it actually came from, so `(_index, _id)` stays distinct.
+//!
+//! Ids collide across indices as a matter of course — each index numbers its
+//! own documents — so `(_index, _id)` is the only key that separates them. When
+//! the scroll stamps one name onto every hit, a consumer keyed on that pair
+//! silently keeps a fraction of the corpus. Scroll is what reindex, migration,
+//! backup and CDC use, which is where losing half a corpus hurts most.
+//!
+//! The in-module unit test for this covers the *read* site only: it hand-builds
+//! a `ScrollContext` and calls the continuation. That leaves both *creation*
+//! sites unguarded — an independent verification of PR #424 showed the defect
+//! could be reintroduced at either one with all 216 tests still passing. This
+//! test goes through the real HTTP routes instead, so the creation sites are
+//! actually exercised:
+//!
+//!   * `POST /{a,b}/_search?scroll=` — `search_impl`
+//!   * `POST /_search/scroll`        — the continuation
+//!
+//! Each document carries `_source.home` = the index it was written to, so the
+//! assertion compares the reported `_index` against ground truth per hit rather
+//! than against a count.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use std::collections::HashSet;
+use tower::ServiceExt;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn json_req(
+    app: &axum::Router,
+    method: &str,
+    path: &str,
+    body: Value,
+) -> (StatusCode, Value) {
+    let mut req = Request::builder().method(method).uri(path);
+    let body = if body.is_null() {
+        Body::empty()
+    } else {
+        req = req.header("content-type", "application/json");
+        Body::from(body.to_string())
+    };
+    let response = app.clone().oneshot(req.body(body).unwrap()).await.unwrap();
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .unwrap();
+    let v: Value = serde_json::from_slice(&bytes).unwrap_or(Value::Null);
+    (status, v)
+}
+
+/// Documents per index. Ids are `0..N` in BOTH, so every id collides.
+const N: usize = 120;
+
+async fn seed(app: &axum::Router, index: &str) {
+    let (st, _) = json_req(
+        app,
+        "PUT",
+        &format!("/{index}"),
+        json!({"mappings": {"properties": {"v": {"type": "long"}, "home": {"type": "keyword"}}}}),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "create {index}");
+
+    let mut bulk = String::new();
+    for i in 0..N {
+        bulk.push_str(&format!("{}\n", json!({"index": {"_id": i.to_string()}})));
+        bulk.push_str(&format!("{}\n", json!({"v": i, "home": index})));
+    }
+    let response = app
+        .clone()
+        .oneshot(
+            Request::builder()
+                .method("POST")
+                .uri(format!("/{index}/_bulk"))
+                .header("content-type", "application/x-ndjson")
+                .body(Body::from(bulk))
+                .unwrap(),
+        )
+        .await
+        .unwrap();
+    assert_eq!(response.status(), StatusCode::OK, "bulk {index}");
+    let (st, _) = json_req(app, "POST", &format!("/{index}/_refresh"), Value::Null).await;
+    assert_eq!(st, StatusCode::OK, "refresh {index}");
+}
+
+/// Walk a scroll to exhaustion, returning `(reported_index, id, true_home)` per
+/// hit. `page_size` deliberately does not divide `N`, so a page straddles the
+/// boundary between the two indices — the case an off-by-one in the pairing
+/// would otherwise slip through.
+async fn walk(app: &axum::Router, spec: &str, page_size: usize) -> Vec<(String, String, String)> {
+    let (st, first) = json_req(
+        app,
+        "POST",
+        &format!("/{spec}/_search?scroll=2m"),
+        json!({"query": {"match_all": {}}, "size": page_size}),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "scroll start: {first}");
+
+    let mut sid = first["_scroll_id"].as_str().map(str::to_string);
+    let mut page = first;
+    let mut out = Vec::new();
+
+    for _ in 0..64 {
+        let hits = page["hits"]["hits"].as_array().cloned().unwrap_or_default();
+        if hits.is_empty() {
+            break;
+        }
+        for h in &hits {
+            out.push((
+                h["_index"].as_str().unwrap_or_default().to_string(),
+                h["_id"].as_str().unwrap_or_default().to_string(),
+                h["_source"]["home"]
+                    .as_str()
+                    .unwrap_or_default()
+                    .to_string(),
+            ));
+        }
+        let Some(id) = sid.clone() else { break };
+        let (st, next) = json_req(
+            app,
+            "POST",
+            "/_search/scroll",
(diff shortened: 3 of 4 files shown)


xerj-org:
## Independent adversarial verification — verdict 1 of 3 (correctness lens)

Workflow-spawned skeptic, instructed to REFUTE this PR, with no ability to merge or push. Recorded verbatim before any action is taken on it.

**refuted: `False` · confidence: `high`**

### Summary

> I tried to refute PR #424 and could not refute its behavioural claim. I built two independent `ci-test` binaries under rustc 1.97.1 — baseline `ad3bbbb` (the PR's merge-base) at /tmp/verify-424-before and PR head `f459fb7` at /tmp/verify-424-correctness — and ran them on my own ports with their own data dirs (never touching localhost:9200, never running the ES-YAML conformance suite). I built a corpus of three indices with deliberately colliding ids (`mi_a`/`mi_b`/`mi_c`, ids 0..299) and stamped each document with `_source.home` = its true index, so I could check every single hit's label against ground truth rather than merely counting distinct pairs. Baseline: 900 hits collapse to 300 distinct `(_index,_id)` with 600 hits mislabelled; PR: 900/900 distinct with `_index` == `_source.home` on every hit. I then pushed well past the PR's testing: three and four indices, the `mi_*` pattern, page sizes 1/7/250/777/999/2500, index boundaries crossed MID-PAGE (size 250 gives `page 2: {'mi_a': 50, 'mi_b': 200}`, `page 3: {'mi_b': 100, 'mi_c': 150}` — the case the PR never tested), the legacy `GET /_search/scroll/{id}` continuation, and 9,000- and 12,000-document corpora. Every hit was labelled correctly in every configuration. I proved both creation sites are genuinely fixed by exploiting their distinct baseline fingerprints: `search_impl` (es_compat.rs:11203) mislabels with `index_names.first()` while `search_with_scroll` (es_compat.rs:19752) mislabels with the raw comma spec `"mi_a,mi_b,mi_c"` — I observed both wrong values on the baseline and both correct after. `total` stays 900 across all pages, `_id`-only dedup still yields 300 as designed, the >10k snapshot-cap rejection is unchanged, and `search_after` returns 900/900 with zero mislabels on BOTH binaries (always correct, no regression). All gate claims reproduce exactly: fmt exit 0, clippy `-D warnings` exit 0 under 1.97.1, 10 passed/0 failed in both `ci-test` and `dev`, and the fail-before assertion text matches character for character. My substantive objection is not to the code but to a durability claim: the PR and CHANGELOG assert the pairing is structural so the association "cannot be dropped again by either site", and I disproved that — I reintroduced the full #414 defect at the `search_impl` creation site in a form that compiles cleanly, and all 216 xerj-api lib tests passed, including the new regression test. The new test injects a hand-built ScrollContext, so it only covers the read site; nothing in the suite drives a real multi-index scroll end to end.

### Claims it re-ran

| claim | holds | observed |
|---|---|---|
| before: 600 hits -> 300 distinct (_index,_id), all labelled with the first index | yes | `### A. 3 indices, comma list, size=100 (divides evenly)   page 1..9: 100 hits each, _index={'mi_a': 100} on EVERY page   900 hits -> 300 distinct (_index,_id); 300 distinct _id   !` |
| after: 600 hits -> 600 distinct, pages 4-6 correctly report the second index | yes | `### A. 3 indices, comma list, size=100 (divides evenly)   page 1: 100 hits _index={'mi_a': 100}   page 2: {'mi_a': 100}   page 3: {'mi_a': 100}   page 4: 100 hits _index={'mi_b': 1` |
| search_after over the same indices was ALWAYS correct (600/600) | yes | `=== BEFORE (19610) === search_after: 900 hits -> 900 distinct (_index,_id); _index mismatches=0 === AFTER  (19600) === search_after: 900 hits -> 900 distinct (_index,_id); _index m` |
| fail-before by reverting only the behavioural line: left "mi_a", right "mi_b" | yes | `test es_compat::passage_scroll_tests::scroll_continuation_reports_the_index_each_hit_came_from ... FAILED panicked at crates/xerj-api/src/es_compat.rs:20249:9: assertion `left == r` |
| both profiles: 10 passed, 0 failed on the scroll tests | yes | `ci-test: running 10 tests ... test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 206 filtered out; finished in 0.70s dev:     running 10 tests ... test result: ok. 10 pas` |
| fmt 0, clippy -D warnings 0 under 1.97.1 | yes | `FMT EXIT: 0   (no diff output) rustc 1.97.1 (8bab26f4f 2026-07-14) clippy: ... Checking xerj-api v1.0.0-rc.17 ... Finished `dev` profile [unoptimized + debuginfo] target(s) in 1m 3` |
| both creation sites are fixed, not just the one at :19752 | yes | `The two sites have DISTINCT baseline fingerprints, which proves which one each request reaches:   BEFORE `/mi_a,mi_b,mi_c/_search?scroll=`      -> continuation label 'mi_a'        ` |
| (untested by the PR) a scroll that crosses the index boundary MID-PAGE is labelled correctly | yes | `### C. size=250, crosses index boundary MID-PAGE   page 1: 250 hits  _index={'mi_a': 250}   page 2: 250 hits  _index={'mi_a': 50, 'mi_b': 200}   page 3: 250 hits  _index={'mi_b': 1` |
| (untested by the PR) an index PATTERN and awkward page sizes behave | yes | `B. /mi_*/_search?scroll= size=100 -> 900 hits -> 900 distinct; _index matches _source.home on every hit; PASS D. size=7 (129 pages)              -> 900 hits -> 900 distinct; mismat` |
| (untested by the PR) _id-only dedup still behaves and total is still correct | yes | `distinct _id alone      : 300  (expected 300 — ids collide by design, unchanged) distinct (_index,_id)   : 900  (expected 900) hits.total values seen  : ['{"value": 900, "relation"` |

### [should-fix] The new regression test covers only the read site; reintroducing the defect at either creation site passes all 216 tests

The PR asserts the pairing is structural so the association "cannot be dropped again by either site" (CHANGELOG and the doc comment on ScrollContext::hits). That is overstated. The new test `scroll_continuation_reports_the_index_each_hit_came_from` inserts a hand-built `ScrollContext` directly into `state.engine.scrolls` and then calls `/_search/scroll`, so it exercises only `scroll_page_response`. The PR's own fail-before reverted only the read line, so it never demonstrated coverage of the creation sites either. The type change raises the bar but does not close the hole: I reintroduced the exact #414 behaviour at the `search_impl` creation site (es_compat.rs:11205) — the site behind `/{indices}/_search?scroll=`, the route real ES clients use — by pairing every hit with `merged_hits.first()`'s name instead of its own. It compiled, and the entire suite stayed green. The existing `scroll_window_tests` are single-index (`index_docs(&state, "small", 25)`), so nothing in the crate drives a multi-index scroll through either creation site. Recommend an end-to-end test: index two indices with colliding ids, scroll them via `/{a},{b}/_search?scroll=` and via `/{a},{b}/_search_scroll`, and assert distinct `(_index,_id)` across all pages.

```
With the defect reintroduced at es_compat.rs:11205 (creation site), `rustup run 1.97.1 cargo test -p xerj-api --lib --profile ci-test -j 20 scroll` ->
  running 10 tests
  test es_compat::passage_scroll_tests::scroll_continuation_reports_the_index_each_hit_came_from ... ok
  test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 206 filtered out
  EXIT: 0
And the full crate suite with the same defect in place:
  `rustup run 1.97.1 cargo test -p xerj-api --lib --profile ci-test -j 20`
  test result: ok. 216 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.74s
  EXIT: 0
Contrast with reverting the READ line only (index: ctx.index.clone()), which the PR did test:
  test result: FAILED. 9 passed; 1 failed
  panicked at crates/xerj-api/src/es_compat.rs:20249:9:
  assertion `left == right` failed: continuation page must report the index the hit came from, not the context's
    left: String("mi_a")
   right: "mi_b"
```

### [should-fix] Alias-addressed multi-index scrolls still return a non-distinct (_index,_id); the fixed claim is not scoped

The CHANGELOG states the defect as fixed without qualification, and names "reindex, migration, backup and CDC consumers" as the blast radius — tooling that routinely scrolls an alias rather than a comma list. Scrolling through an alias that spans three indices still reports the ALIAS name as `_index` on every hit (ES always returns the concrete index name), so `(_index, _id)` is still not the concrete-index identity pair on that path. Separately, the alias scroll returns only 300 of the 900 documents the alias covers. Both behaviours are byte-identical on the baseline and the PR, so this is NOT a regression introduced here and it does not block the fix — `search_impl` never calls `resolve_alias`, so `index_names` keeps the alias token and page one is mislabelled too. But it means the headline claim overreaches. Recommend scoping the CHANGELOG wording to comma-list and wildcard specs and filing a follow-up for the alias path.

```
Identical output from both binaries (alias mi_all -> mi_a, mi_b, mi_c; 900 docs total):
BEFORE (ad3bbbb, port 19610):
  alias def: {"mi_c": {"aliases": {"mi_all": {}}}, "mi_a": {...}, "mi_b": {...}}
  count via alias: 300
  page 1: 100 hits _index={'mi_all': 100}
  300 hits -> 300 distinct (_index,_id); mismatches=300
  first hit source: {"_index": "mi_all", "_id": "0", "_source": {"home": "mi_a", ...}}
AFTER (f459fb7, port 19600):
  count via alias: 300
  page 1: 100 hits _index={'mi_all': 100}
  300 hits -> 300 distinct (_index,_id); mismatches=300
  first hit source: {"_index": "mi_all", "_id": "0", "_source": {"home": "mi_a", ...}}
(`mismatches` counts hits where _source.home != _index, i.e. all 300.)
```

### [note] ScrollContext::index is now write-only dead state, still populated with two different misleading values

After this change nothing in the workspace reads `ScrollContext::index` — `_index` is served exclusively from the per-hit name. Both creation sites still populate it, and with inconsistent semantics: es_compat.rs:14246 stores `index_names.first()` while es_compat.rs:19811 stores `index.clone()`, the raw comma-separated spec. A `pub` field that is never read but is filled with a plausible-looking index name is precisely the shape of trap this PR exists to remove — the next reader will reasonably assume it is authoritative. Recommend deleting the field, or renaming it to something that cannot be mistaken for a serving value.

```
`grep -rn "ctx\.index" engine/crates/` on the PR branch returns no hits in es_compat.rs; the only `_index` emission in scroll_page_response is `o.insert("_index".to_string(), Value::String(h.index.clone()))` at es_compat.rs:20083, where `h` is the EsHit built from `hit_index.clone()`. `grep -rn "ScrollContext" --include=*.rs` shows the only remaining consumers are the two creation sites, three test sites, and xerj-engine/tests/search_context_ttl.rs (which sets `index: "idx"`, `hits: Vec::new()`).
```

### [note] Type-signature break on a pub field is filed under Fixed with no Changed/breaking note

`ScrollContext::hits` is a public field on a public struct in `xerj-engine` and changes type from `Vec<Hit>` to `Vec<(String, Hit)>`. The CHANGELOG documents this only under `### Fixed`, though the repo's own convention includes `### Changed` sections for this kind of thing. Any out-of-tree consumer constructing or reading a ScrollContext breaks at compile time with nothing in the changelog signalling it. Low impact if xerj-engine is treated as internal, but the entry as written would not warn anyone.

```
engine/crates/xerj-engine/src/engine.rs: `-    pub hits: Vec<xerj_query::executor::Hit>,` / `+    pub hits: Vec<(String, xerj_query::executor::Hit)>,` inside `pub struct ScrollContext`. CHANGELOG.md diff adds 29 lines entirely under `### Fixed`; `grep -n "### \(Changed\|Breaking\|Removed\)" CHANGELOG.md` shows `### Changed` is an established section in this file (lines 699, 1379, 2034, 2295, ...).
```

### [note] PR narrative understates the baseline symptom on the _search_scroll route and on the search_impl route

The PR body says both sites collapsed to `index_names.first()`. That is true only for `search_impl`. On `/{index}/_search_scroll` the baseline stamps the RAW comma spec — which is what issue #414 actually reported (`11,900 of 12,000 hits carry _index: "sa,sb"`) — and the PR body does not mention it. Also, `index_names.first()` is not reliably "the first index": the ordering is nondeterministic, and in one of my runs a scroll over `_all` labelled 800 hits `.xerj_prefs`, an empty internal system index. This does not affect the correctness of the fix (both symptoms are gone) but the before-state is worse and less predictable than the PR describes.

```
BEFORE (ad3bbbb), `/mi_a,mi_b,mi_c/_search_scroll?scroll=5m` size=100:
  page 1: 100 hits  _index={'mi_a': 100}
  page 2: 100 hits  _index={'mi_a,mi_b,mi_c': 100}
  ... pages 3-9 identical ...
  900 hits -> 400 distinct (_index,_id); 800 hits whose _index disagrees with _source.home
BEFORE, `/_search?scroll=` size=100:
  page 1: 100 hits  _index={'mi_c': 100}
  page 2: 100 hits  _index={'.xerj_prefs': 100}
  ... pages 3-10 all '.xerj_prefs' ...
  916 hits -> 416 distinct (_index,_id); 800 hits whose _index disagrees with _source.home
```

### Explicitly not verified

- The ES-compat YAML conformance suite. I deliberately did not run it — per my instructions it deletes every index on the node it targets and has destroyed the shared reference corpus twice. This PR changes a wire-visible field (`_index` on scroll continuations), so conformance remains the one gate that could still surface a divergence I cannot see. The PR itself flags this as not run; that caveat is accurate and still open.
- Multi-node / clustered behaviour. Everything I ran was a single local node. I did not exercise xerj-cluster routing, remote scroll contexts, or whether a scroll context survives or is correctly labelled across nodes.
- Concurrency and lifecycle interactions: many simultaneous open scrolls, the `max_open_scrolls` cap, TTL expiry and the background sweeper racing a continuation. I only ever had one or two scrolls open at a time.
- Memory cost of the change. `ScrollContext::hits` now carries an owned `String` per hit, and `search_impl` switched from `merged_hits.iter().map(|(_, h)| h.clone())` to `merged_hits.clone()`. At the 10k snapshot cap times `max_open_scrolls` this is a real if small increase; I did not measure RSS on either binary.
- PIT-backed scrolls and `slice`d scrolls. `search_impl` overrides `index_names` from `pit.indices` when a PIT id is present, and I did not test whether a PIT- or slice-scoped scroll labels `_index` correctly on continuation pages.
- `_msearch` with a scroll, and whether any other API materialises a ScrollContext. I only confirmed by grep that `scrolls.insert` appears at exactly two non-test sites.
- Whether the `fields` parameter (e.g. `fields: ["_index"]`) is honoured on continuation pages at all — `scroll_page_response` builds `fields` solely from `passage_fields(h)`, which returns only passage data. This looks like a pre-existing continuation-page gap unrelated to the PR's claim, and I did not test it.
- Whether the pre-existing asymmetry in the >10k snapshot guard matters in practice: `/_search?scroll=` rejects a 12,000-document multi-index scroll with HTTP 400 while `/{idx}/_search_scroll` accepts it (its truncation check is per-index, not on the summed total). Identical on both binaries, so not introduced here, but I did not chase whether that path can silently truncate at larger per-index counts.

xerj-org:
## Independent adversarial verification — verdict 2 of 3 (evidence lens)

**refuted: `False` · confidence: `high` · must-fix: 0**

### Summary

> I set out to refute PR #424 and could not. Every load-bearing number in the PR body and commit message reproduced exactly on my own build. I built the branch at f459fb7 with rustc 1.97.1, reverse-applied the engine diff so the two .rs files were byte-identical to origin/main (`git diff origin/main -- <both files> | wc -l` = 0), rebuilt, and measured both states end-to-end on my own nodes (ports 19425/19440, own data dirs, never :9200). BEFORE: 600 hits -> 300 distinct (_index,_id), all labelled 'mi_a'. AFTER: 600 -> 600, pages 4-6 correctly 'mi_b'. search_after: 600/600 in both states. The fail-before proof reproduced verbatim including the exact assertion text (left: String("mi_a") / right: "mi_b"), with exactly one test failing and it being the new one. 10 passed / 0 failed in both ci-test and dev. fmt exit 0 and clippy --workspace --all-targets -D warnings exit 0 under 1.97.1. I then went beyond the PR: I ran the three cases it listed as unverified (three indices, wildcard pattern, second scroll route) and all are fixed, and I ran the full ES-YAML conformance suite it skipped -- against my own node, 1366 passed / 0 failed / 3 skipped, exit 0 -- closing its stated gap. My findings are therefore not about correctness. They are that the CHANGELOG and PR body misdescribe the cause at one of the two creation sites and omit the uglier symptom issue #414 actually headlines (main emitted `_index: "mi_a,mi_b"`, the raw comma-separated spec, which is not an index name at all), and that the sole regression test exercises only the read site, so the "structural" claim is half true. Both are should-fix editorial/coverage items, not correctness blockers. I ran no state-changing GitHub commands; my worktree is clean at f459fb7 and my test nodes are stopped.

### Claims it re-ran

| claim | holds | observed |
|---|---|---|
| before: 600 hits -> 300 distinct (_index,_id), all labelled with the first index | yes | `page 1: 100 hits _index={'mi_a': 100} ids 0..99 / page 2: {'mi_a': 100} ids 100..199 / page 3: {'mi_a': 100} ids 200..299 / page 4: {'mi_a': 100} ids 0..99 / page 5: {'mi` |
| after: 600 hits -> 600 distinct, pages 4-6 correctly report the second index | yes | `pages 1-3 _index={'mi_a': 100}, pages 4-6 _index={'mi_b': 100}; "600 hits -> 600 distinct (_index,_id)", "index label counts: {'mi_a': 300, 'mi_b': 300}". Reproduced twic` |
| search_after over the same two indices was ALWAYS correct (600/600) | yes | `Both states: "[search_after] 600 hits -> 600 distinct (_index,_id)", "[search_after] index label counts: {'mi_a': 300, 'mi_b': 300}". The two paging APIs did disagree on ` |
| fail-before by reverting only the behavioural line: left "mi_a", right "mi_b" | yes | `test result: FAILED. 9 passed; 1 failed. The single failure is the new test, panicking at es_compat.rs:20249 with `assertion \`left == right\` failed: continuation page m` |
| both profiles: 10 passed, 0 failed on the scroll tests | yes | `ci-test: "test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 206 filtered out". dev: "test result: ok. 10 passed; 0 failed; 0 ignored; 0 measured; 206 filtered ` |
| fmt 0, clippy -D warnings 0 under 1.97.1 | yes | `FMT EXIT: 0 (no output). Clippy: "Finished `dev` profile ... in 1m 26s" / "CLIPPY EXIT: 0", with `grep -c "^warning\\|^error"` = 0.` |
| not-in-PR: does the fix hold for 3+ indices and for a wildcard pattern (both listed by the PR as untested)? | yes | `3 indices, /_search?scroll=: main 900 -> 300 distinct (all 'mi_a'); PR 900 -> 900 (3x mi_a, 3x mi_b, 3x mi_c pages). 3 indices, /_search_scroll: main 900 -> 400 (page 1 '` |
| not-in-PR: the ES-YAML conformance suite the PR declined to run | yes | `"ES-COMPAT YAML RUNNER · 200 files · http://127.0.0.1:19440" / "1366 passed · 0 failed · 3 skipped · 1369 total" / RUNNER EXIT: 0. No conformance regression. Separately c` |
| not-in-PR: no collateral damage from the ScrollContext type change | yes | `xerj-api full lib: "test result: ok. 216 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out". search_context_ttl (an integration test that builds a ScrollContext dir` |

### [should-fix] CHANGELOG and PR body misstate the cause at the second creation site, and omit the symptom issue #414 headlines

Both the PR body and the shipped CHANGELOG entry attribute the defect to a context-level index equal to `index_names.first()` at both creation sites. That is true only for `search_impl` (the `/_search?scroll=` route). `search_with_scroll` takes `Path(index): Path<String>` -- the raw comma-separated spec -- and stored `ScrollContext { index: index.clone(), .. }`, so on main that route stamped `_index: "mi_a,mi_b"`, a string that is not an index name at all. This is the exact measurement issue #414 leads with (11,900 of 12,000 hits carrying the raw spec), yet the PR body never mentions this route, never reports its numbers, and its causal sentence is wrong for it. The blast radius reads differently too: a consumer writing back to `_index` as read would create an index literally named `mi_a,mi_b`, which is worse than the mislabel the CHANGELOG describes. The code change fixes both routes correctly -- only the prose is wrong.

### [should-fix] The only regression test exercises the read site; neither creation site nor any real multi-index scroll is covered

`scroll_continuation_reports_the_index_each_hit_came_from` hand-constructs a `ScrollContext` and calls `POST /_search/scroll`, so it covers `scroll_page_response` only. The type change to `Vec<(String, Hit)>` prevents a future edit from *dropping* the pairing, but nothing prevents it from *storing the wrong name*: re-introducing `merged_hits.iter().map(|(_, h)| (scroll_index.clone(), h.clone())).collect()` at es_compat.rs:11203 compiles fine and passes all 10 tests. The PR calls the fix "structural"; it is structural against one of the two failure modes. The `search_with_scroll` route has no `_index` test coverage at all, and the 600->600 evidence lives only in a manual script not preserved in the repo, so nothing in CI would catch a reintroduction.

### [note] The 'Assumed / not verified' section is honest about what was skipped but wrong about why it mattered

The PR says of the ES-YAML suite: "this change alters a wire-visible field (`_index` on scroll continuations), so conformance is the gate that would catch a divergence I cannot see from here." That is false. The suite's four scroll files use a single index each and make zero `_index` assertions anywhere, so it could not have caught this defect and cannot catch a regression of it. Separately, CI runs that job on every PR and it is already green on #424, so framing it as a manual pre-merge action item is misleading. I ran it myself anyway to close the gap on the record. The section is honest in that it discloses the omission -- it just overstates the value of the thing omitted, which is the opposite of the failure mode to worry about.

### [note] ScrollContext::index is now written at every construction site and read nowhere

After the fix nothing in the workspace reads `ScrollContext::index` -- `_index` is served entirely from the per-hit pair. The retained comment calls it a "Context-level fallback only", but it is not a fallback for anything; it is dead state carried in a struct that RC4 blocker 11 exists specifically to bound. Because the field is `pub`, no dead_code lint fires, so this will not surface on its own. Either delete it or state what still needs it.

### [note] One heap String per snapshotted hit added to a deliberately memory-bounded struct

`hits: Vec<(String, Hit)>` allocates a fresh index-name String for every hit in every scroll snapshot. Caps are SCROLL_SNAPSHOT_MAX_HITS = 10_000 and max_open_scrolls = 500, so the theoretical ceiling is 5M Strings. In practice the `Hit` (which carries the full `_source` JSON) dominates, so the relative overhead is small -- but this is the exact struct RC4 blocker 11 was filed to bound, and the same structural guarantee is available for far less with `Arc<str>` or an interned index id, since the number of distinct index names per context is tiny. Worth a follow-up, not a blocker.

### Explicitly not verified

- Concurrency: I did not test two clients paging the same scroll context simultaneously, nor a context being TTL-swept mid-page. The change makes `hits` a larger element type but does not alter the locking around `scrolls.get_mut`, so I have no reason to suspect a problem -- I simply did not exercise it.
- Memory under load: I reasoned about the added per-hit String against SCROLL_SNAPSHOT_MAX_HITS (10,000) and max_open_scrolls (500) but did not measure RSS with many large scroll contexts open, so the practical cost in finding 5 is an estimate, not a measurement.
- Aliases: I tested explicit comma lists and a wildcard pattern, but not a scroll over an alias that fans out to multiple concrete indices. ES reports the concrete index in that case; I did not confirm what xerj now reports.
- The PR's third 'assumed' item -- that `_seq_no`/`_version`/`_primary_term` are hardcoded to Some(0)/Some(1)/Some(1) on continuation pages -- I confirmed by reading (es_compat.rs:20033-20035) that the hardcoding is real and unchanged by this PR, but I did not investigate whether it is actually wrong. The PR correctly scopes it out.
- I did not verify the PR's original 300+300 corpus was built the same way mine was. I used integer ids 0..299 per index with a `match_all` scroll at size=100; the PR describes 'identical id ranges 0..299', which matches, but a difference in mapping or sort order could in principle change page composition. The distinct-pair totals are order-independent, so this does not affect the headline numbers.
- Sliced scroll (`test_sliced_scroll` in the YAML suite) over multiple indices -- the suite covers slicing only on a single index, and I did not construct a multi-index sliced-scroll case by hand.

xerj-org:
## Independent adversarial verification — COMPLETE (3 of 3)

| lens | refuted | confidence | must-fix |
|---|---|---|---|
| n/a | `False` | high | 0 |
| n/a | `False` | high | 0 |
| n/a | `False` | high | 0 |

**Panel result: 0 of 3 refuted, 0 must-fix findings.**

### Verdict 3 summary

> Lens: regression risk. I built the PR branch and, separately, a true `origin/main` baseline binary from the same worktree, ran both as my own nodes (ports 21470 / 21480, own data dirs), and diffed wire behaviour. (My first probe round was invalid — port 19424 was already held by another agent's node and my server silently failed to bind; I detected it via `ss -ltnp` + the node log and redid everything on ports I verified free.) Every claim in the PR body reproduced exactly, including the fail-before message. On the regression question specifically: I enumerated every reader and writer of `ScrollContext::hits` (5 construction sites + 3 read sites) and every one is coherent, not merely compiling; `has_non_score_sort` is semantically identical (same `.first()` hit, `.sort` field) and behaves identically live; single-index scrolls — the common case — are byte-identical before/after; the snapshot cap (#198/#370), TTL expiry, `clear_scroll`, and the legacy `GET /_search/scroll/{id}` path form are unchanged or strictly improved; no internal Rust or JS consumer of the scroll HTTP API exists. I closed the PR's own declared gap: the full ES-compat YAML suite (200 files, 1369 cases) returns **1366 passed / 0 failed / 3 skipped on both** the PR and main nodes — no conformance divergence. Full `xerj-api` suite is 425 passed / 0 failed / 2 ignored in *both* dev and ci-test with `--test-threads=2`; fmt and `clippy -D warnings` are exit 0 under 1.97.1. I could not refute the fix. What I did find: the PR silently fixes a *second*, separately published defect (#405) and leaves `landing/docs/api-es-compat.html:196` telling users the opposite of what now ships; `ScrollContext::index` is now provably write-only dead state whose two writers disagree about its meaning; and the snapshot costs ~7% more memory per hit.

#### [should-fix] PR silently fixes issue #405 and leaves the published page stating the opposite of shipped behaviour

`landing/docs/api-es-compat.html:196` publishes an explicit caveat about the `POST /:index/_search_scroll` route: "Note that per-hit `_index` is **not** reliable on that route — continuation pages return the raw comma-separated index spec rather than the resolving index, so `(_index, _id)` is not a distinct key there. Export and reindex tooling should not depend on it. Tracked in #405." This PR makes that route correct, so the page now actively misdirects the exact consumers (export/reindex) the fix is for. The matching CHANGELOG entry at CHANGELOG.md:95-102 ('One boundary is documented rather than fixed here... The pages now say so and point at #405') is likewise stale. The PR touches only CHANGELOG.md, es_compat.rs and engine.rs — the page is not updated, #405 is never mentioned in the new CHANGELOG entry or the PR body, and no test caught it: `scroll_snapshot_cap_is_documented.rs` pin

#### [note] `ScrollContext::index` is now write-only dead state, and its two writers mean different things

After the fix nothing in the workspace reads the field. I proved this by renaming it and compiling all targets: every resulting error is E0560 (struct-literal field), i.e. a write site — there is not one read site. The PR's new comment at es_compat.rs:14221 calls it a 'Context-level fallback only', but no code falls back to it. Worse, the two production writers disagree about what it holds: es_compat.rs:14246 stores `index_names.first()` (a *resolved concrete* index name) while es_compat.rs:19811 stores `index.clone()` (the *raw path spec*, e.g. literally `"r5_a,r5_b"` — that is exactly what leaked onto the wire on main, see the finding above). Leaving a `pub` field with a plausible name, a comment describing a fallback that does not exist, and two mutually inconsistent meanings is a trap for the next change. It should be deleted, or if kept, documented as the raw request spec and made c

#### [note] Scroll snapshots cost ~7% more memory per hit (measured, not a blocker)

Each snapshotted hit now carries its own `String`. Measured on real nodes with a 10,000-document index, 40 open scroll contexts (400,000 snapshotted hits), reading VmRSS from /proc: main averages ~771 B/hit, the PR ~826 B/hit — about +55 B/hit, +7%. Extrapolated to the documented worst case (SCROLL_SNAPSHOT_MAX_HITS=10,000 x max_open_scrolls=500) that is roughly +275 MB on top of an already ~3.9 GB baseline. So it is a small relative increase on a design that is already dominated by pinning fully-hydrated `_source` JSON in RAM, not a new class of problem — but it is a real, measurable regression and the PR body does not mention memory at all. RSS is noisy (~±50 B/hit across runs) yet the PR was higher in all three paired runs.

#### [note] The distinctness property does not hold when the multi-index set is named by an alias (pre-existing, unchanged by this PR)

The PR's blast-radius argument is 'reindex, migration, backup and CDC consumers'. Those routinely scroll an *alias*, not a comma list. `POST /<alias>/_search?scroll=` over a two-index alias is broken a layer below scroll: it returns only 300 of 600 documents and labels every `_index` with the alias name, on the PR build exactly as on main. It is not a regression — behaviour is byte-identical before and after — and it is not in scope for #414, but it means the headline claim '(_index,_id) is distinct for a multi-index scroll' is true for the comma-list and wildcard forms and false for the alias form. Worth a follow-up issue and worth not overstating in the CHANGELOG. Note the wildcard form (`mi_*`) — which the PR body lists as untested — I did test, and it is fixed (300 -> 600 distinct).

#### [note] Pre-existing, out of scope: `DELETE /_search/scroll/_all` frees nothing and reports num_freed 0

`clear_scroll_path` (es_compat.rs:20820) does `scrolls.remove(&scroll_id)` on the literal string, so the `_all` sentinel is treated as a scroll id. `clear_scroll` on `DELETE /_search/scroll?scroll_id=_all` handles `_all` correctly. Identical before and after this PR, so not a regression, but it bit my own memory measurement (40 pinned 10k-hit contexts survived a 200 OK 'succeeded: true') and it is on the exact contexts-leak path that RC4 blocker 11 and `search_context.max_open_scrolls` exist to bound. Flagging because it is adjacent code the reviewer is already looking at.

### Verdict 3 — explicitly not verified

- Scroll across THREE OR MORE explicit indices. I tested two-index comma lists, a two-index wildcard (`mi_*`, fixed: 300 -> 600 distinct) and a two-index alias (broken pre-existing, see finding), but not a 3+ comma list.
- Concurrency: two clients continuing the same scroll_id simultaneously, or a continuation racing the background TTL sweeper. `scroll_page_response` holds a DashMap `get_mut` shard guard across the whole page build, which the PR does not change, but I did not exercise it.
- Interaction of the new per-hit index with `slice` (sliced scroll) and with `pit` on a multi-index scroll. The ES-YAML sliced-scroll cases pass identically on both nodes, but they are single-index.
- Whether CI itself passes. I replicated CI's toolchain (rustup run 1.97.1) locally for fmt, clippy --workspace --all-targets -D warnings, and the xerj-api/xerj-engine suites; I did not run the repository's CI workflow, and I did not run the full workspace test suite (only xerj-api in both profiles plus xerj-engine's search_context_ttl).
- Allocator-level accounting for the memory finding. I measured process VmRSS only, with jemalloc in play; run-to-run noise was roughly +/-50 B/hit, so the +55 B/hit figure is a mean over three paired runs, not a precise per-allocation number. I did not instrument size_of::<(String, Hit)>() directly.
- Whether any XERJ-external client (OpenSearch Dashboards 3.7.0's saved-objects migration, elasticsearch-py helpers.scan) actually round-trips correctly against the new behaviour. The legacy path-param route it depends on works and its unit test passes, but I did not drive a real client.
- I did not test the redundant deep clone at es_compat.rs:19756 (`let hits_only = all_hits.clone()` where `all_hits` is dropped immediately after `first_page` is built) for a throughput regression. It is pre-existing waste of up to 10,000 hydrated hits that the PR preserves and slightly enlarges; the variable name `hits_only` is now a misnomer.

---

### Disposition of every should-fix raised across the panel

| finding | status |
|---|---|
| read-site-only test; defect reintroducible at either creation site with all 216 tests green | **fixed** — `tests/scroll_multi_index_identity.rs` drives both real HTTP routes; proven to catch a creation-site regression that the unit test misses |
| alias-addressed scrolls still non-distinct; claim unscoped | **fixed** — CHANGELOG scoped to comma-list/wildcard/`_all`; alias case filed as #433 |
| CHANGELOG/PR misstate the cause at the second creation site | **fixed** — both sites now stated: `index_names.first()` vs the raw `Path<String>` spec, the latter being the `_index: "sa,sb"` symptom #414 headlines |

Remaining items are `note` severity and not blocking. The dead `ScrollContext::index` field is tracked separately rather than folded in here.

Issue #79, opened 2026-07-31
security(Medium): per-brain authorization missing — any caller reads any brain (dotfile-secret leak already fixed)


xerj-org:
**Severity:** Medium (secret exposure) + a tracked design gap · **Status:** dotfile skip fixed on `main` (`4b9e27f`); per-brain authz open

**Fixed:** the autoindex walker had no ignore rules and indexed `.env` (API tokens), `.git`, `.ssh`, `.aws` and other dotfiles into a queryable brain. For the "point it at my project folder" use case that is a real exposure. The walker now prunes hidden files and directories (`.git/` skipped whole), with the root exempt so a brain over a dot-named folder still works. Two regression tests.

**Still open (design):** a brain has **no per-brain authorization** — any authenticated caller can read any brain's `overview`/`ego` (self-documented in `graph_api.rs`). Combined with a shared engine this is a multi-tenant exposure. Needs an access model, tracked here.

xerj-org:
**Verified as of `v1.0.0-rc.8` (`7e050d0`)** — title updated to lead with the open part.

- **(a) dotfile/secret skip — FIXED:** the walker prunes hidden files/dirs (`engine/crates/xerj-autoindex/src/walk.rs:35-40`, root exempt) with a regression test (`walk.rs:112`).
- **(b) per-brain authorization — STILL OPEN:** `graph_api.rs:35-44` self-documents *"no per-brain authorization"*; the `ego`/`overview` handlers only syntax-validate the brain name, and `rbac.rs:10` is `enforced:false` with the RoleStore unconsulted. Any authenticated caller can read/mutate any brain. This issue now tracks that remaining access-model work.

Issue #283, opened 2026-08-10
autoindex: a shrinking file set on a junk-bearing generation aborts with an internal invariant (`desired manifest digest does not match sync_begin payload`)


xerj-org:
Found while re-verifying #279 on current main. **Pre-existing — not caused by #279 and not by any ignore rule.** It belongs to the incremental `--no-graph` reconciliation path that landed in #259.

## Symptom

On a tree where an earlier generation was indexed with junk present, a later `--no-graph` run whose file set has shrunk aborts with:

```
error: desired manifest digest does not match sync_begin payload
exit 1, reason=aborted
```

That is an internal invariant leaking to the user. The same code path has a clean, actionable message for the analogous situation — *"autoindex execution configuration changed since the committed generation; rebuild with a new `--state-dir` and a new `--prefix`"* — so the machinery to report this well already exists; this route just doesn't use it.

## It is not the ignore rules — four controls

The first reproduction came through #279's upgrade path (index with `--no-ignore`, re-run with the new default), which made it look like an ignore-rules regression. It is not:

1. Delete one file between two default runs → reconciles cleanly (generation 2).
2. Remove two entire datasets between two default runs, no ignore flags → reconciles cleanly.
3. Minimal ignore-flag flip (`--no-ignore`, then default, 2 files → 1) → reconciles cleanly. So the flag flip alone is not the trigger.
4. **Decisive control:** run twice with `--no-ignore` *both* times and physically delete exactly the files the ignore rules would have excluded — no flag change anywhere — **reproduces the identical abort**.

So the trigger is the file-set delta on a generation whose first run exited `3` (completed-with-junk). Junk records plus removal is the likely ingredient. #279 only makes that delta easy for a user to produce.

## Related, by design rather than defect

The exec-config guard at `lib.rs:2358` only fires when the projected plan is byte-identical to the base plan, so a flag change that alters the file set bypasses it, and `cfg.ignore` is not part of that fingerprint. Adding it would **not** have prevented this abort — the flagless control fails too — but it is the natural place to make ignore-rule changes across runs produce the clean refusal message instead of an invariant.

## What good looks like

Either reconcile the shrinking set correctly, or refuse with the actionable message and the recovery route. An internal digest mismatch is never something a user can act on.

PR #425, opened 2026-08-16
fix(autoindex): skip hidden names that are not valid UTF-8

SebTardif:
## What this changes, and why

`xerj autoindex <folder>` claims it never indexes hidden files (`.env`, `.ssh`, `.git`). The check was `file_name().to_str().is_some_and(|n| n.starts_with('.'))`.

On Unix, `OsStr::to_str()` is `None` when the name is not valid UTF-8. A name whose first byte is `.` then fails the `is_some_and` check and is walked like a visible file.

Typical UTF-8 secrets are already skipped. This closes the remaining Unix case. It is not a GHSA.

The skip now compares the first byte of `file_name().as_encoded_bytes()` to `b'.'`. UTF-8 `.env` / `.ssh` / `.git` are unchanged. `--dry-run`'s `ignored_files_in_pruned_dirs` count uses the same `is_hidden_name` predicate.

This bug was introduced with the hidden-file skip in [#287](https://github.com/xerj-org/xerj/pull/287) ([`5c7b3696`](https://github.com/xerj-org/xerj/commit/5c7b3696), 2026-08-10). Related: [#279](https://github.com/xerj-org/xerj/pull/279), [#79](https://github.com/xerj-org/xerj/issues/79) (UTF-8 skip already landed).

## Upgrade (graph-enabled corpora)

A corpus indexed by a pre-fix binary that contains a hidden non-UTF-8 name looks like a file removal to the resume-plan path. That path is the default, and it is what `xerj brain` uses.

Until [#439](https://github.com/xerj-org/xerj/issues/439) lands, the first rerun exits 1 with `unsupported_content_group_removal`, writes nothing, and the already-indexed documents stay live. Every subsequent rerun reproduces that abort.

Do **not** follow the refusal's first recovery route (`restore_removed_files`). The file is still on disk; the walker just no longer yields it. `--fresh` is refused on that state directory.

Recovery that works:

1. Delete the published indices and the state directory.
2. Re-index the folder with the new binary.

An isolated rebuild with a new `--state-dir` / `--prefix` also exits 0, but leaves the old target's documents live until those indices are deleted by hand.

`--no-graph` can delete the data documents on the upgrade run. Catalog residue can remain; the purge above is the operator procedure this PR documents.

This PR does not change `UnsupportedInventoryDelta`. That is [#439](https://github.com/xerj-org/xerj/issues/439).

Ref #439

## Evidence

Red (helper, before the byte check; APFS cannot create `.\x80`, so local red is on `OsStr::from_bytes`):

```
$ cargo test -p xerj-autoindex -- hidden_non_utf8
test walk::hidden_skip_tests::hidden_non_utf8_dot_name_is_detected ... FAILED
first byte is '.', even when to_str() is None
```

Green (existing UTF-8 hidden skip plus the new tests):

```
$ cargo test -p xerj-autoindex -- hidden_
test result: ok. 6 passed; 0 failed
$ cargo test -p xerj-autoindex -- ignore_rules
test result: ok. 16 passed; 0 failed
```

Linux CI also runs `hidden_non_utf8_dotfile_is_never_indexed` and `count_inside_skips_hidden_non_utf8_names`, which create `.\x80` names on disk.

## Checks

- [x] `cargo fmt --all` (from `engine/`)
- [x] Scoped release build: `cd engine && cargo build --release -j 32 -p xerj-autoindex`
- [x] `cargo test -p xerj-autoindex -- hidden_` (6 passed)
- [x] `cargo test -p xerj-autoindex -- ignore_rules` (16 passed on macOS; Linux adds `count_inside_skips_hidden_non_utf8_names`)
- [x] `cargo clippy -p xerj-autoindex --all-targets -- -D warnings`
- [ ] ES-YAML conformance suite: not run
- [ ] New ES-compatible behaviour YAML case: not applicable
- [x] CHANGELOG `[Unreleased]` Fixed bullet, including the upgrade abort and purge procedure
- [x] Contribution review: public path, fail-closed only on the unsafe name. Durable resume-plan state is affected on upgrade; documented above. Reconciler exemption is #439, not this PR.

**Not run:** ES-YAML (walk / ignore only; no query or index-engine change). `cargo-audit` and fuzz (not a wire parser). Linux on-disk `.\x80` tests (APFS rejects that name; CI Linux will run them).

## Provenance

- Written by: AI coding agent, run by @SebTardif who has reviewed it
  - `gh pr view 419 --repo xerj-org/xerj --json state,mergedAt` → MERGED 2026-08-16
  - Red: `cargo test -p xerj-autoindex -- hidden_non_utf8` failed on `is_hidden_name(OsStr::from_bytes(b".\\x80"))` while the check still used `to_str()`
  - Green: `cargo test -p xerj-autoindex -- hidden_` → 6 passed; `cargo test -p xerj-autoindex -- ignore_rules` → 16 passed
  - `cargo fmt --all` from `engine/`
  - `cargo clippy -p xerj-autoindex --all-targets -- -D warnings`
  - `cd engine && cargo build --release -j 32 -p xerj-autoindex` → Finished release
  - Linux ext4 will create `.\x80` and both `walk()` and `count_inside` will see it. Falsified if the Linux CI tests cannot create the file or if walkdir skips it before our check.
  - Windows WTF-8 still starts `.env` with `b'.'`. Falsified if a Windows hidden name we care about no longer starts with that byte.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -145,7 +145,24 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   used `file_name().to_str().starts_with('.')`. On Unix a name whose
   first byte is `.` and that is not valid UTF-8 (`to_str()` is `None`)
   was walked and indexed. The skip now compares the first encoded byte
-  to `.`. Typical UTF-8 secrets were already skipped.
+  to `.`. Typical UTF-8 secrets were already skipped. `--dry-run`'s
+  `ignored_files_in_pruned_dirs` count uses the same predicate, so a
+  hidden non-UTF-8 file (or the contents of a hidden non-UTF-8
+  directory) is no longer reported as a non-hidden file inside a
+  pruned directory.
+
+  **Upgrade.** A corpus indexed by a pre-fix binary that contains a
+  hidden non-UTF-8 name looks like a file removal to the graph-enabled
+  path (the default, and what `xerj brain` uses). Until
+  [#439](https://github.com/xerj-org/xerj/issues/439) lands, the first
+  rerun exits 1 with `unsupported_content_group_removal`, writes
+  nothing, and the already-indexed documents stay live. Do not follow
+  the refusal's first recovery route ("restore the removed file(s)"):
+  the file is still on disk, just no longer walked. Recovery that
+  works: delete the published indices and the state directory, then
+  re-index the folder. An isolated rebuild with a new `--state-dir` /
+  `--prefix` also exits 0, but leaves the old target's documents live
+  until those indices are deleted by hand.
 
 - **Autoindex snapshot and replay one-shot failpoints can no longer be stolen by unrelated parallel tests.** ([#385](https://github.com/xerj-org/xerj/issues/385)).
   The two test-only failpoints are now owned by the thread that arms them, while
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -140,6 +140,13 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **`xerj autoindex` now skips hidden names that are not valid UTF-8.**
+  The walker already refused `.env`, `.ssh`, and `.git`, but the check
+  used `file_name().to_str().starts_with('.')`. On Unix a name whose
+  first byte is `.` and that is not valid UTF-8 (`to_str()` is `None`)
+  was walked and indexed. The skip now compares the first encoded byte
+  to `.`. Typical UTF-8 secrets were already skipped.
+
 - **Autoindex snapshot and replay one-shot failpoints can no longer be stolen by unrelated parallel tests.** ([#385](https://github.com/xerj-org/xerj/issues/385)).
   The two test-only failpoints are now owned by the thread that arms them, while
   retaining their existing one-shot behavior and serialization locks.
--- a/engine/crates/xerj-autoindex/src/ignore_rules.rs
+++ b/engine/crates/xerj-autoindex/src/ignore_rules.rs
@@ -61,7 +61,15 @@
 use ignore::gitignore::{Gitignore, GitignoreBuilder, Glob};
 use ignore::Match;
 use std::collections::BTreeMap;
+use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
+
+/// Hidden names start with `.` even when they are not valid UTF-8.
+/// `OsStr::to_str()` is `None` for those, so a UTF-8 `starts_with('.')`
+/// would walk them (and, here, would count them as non-hidden).
+pub(crate) fn is_hidden_name(name: &OsStr) -> bool {
+    name.as_encoded_bytes().first() == Some(&b'.')
+}
 
 /// Ignore file honoured in every directory, ranked below `.xerjignore`.
 pub const GITIGNORE: &str = ".gitignore";
@@ -730,11 +738,7 @@ impl IgnoreStack {
             .follow_links(false)
             .min_depth(1)
             .into_iter()
-            .filter_entry(|e| {
-                !e.file_name()
-                    .to_str()
-                    .is_some_and(|name| name.starts_with('.'))
-            });
+            .filter_entry(|e| !is_hidden_name(e.file_name()));
         for entry in walker {
             if self.budget == 0 {
                 self.report.deep_count_complete = false;
@@ -764,12 +768,7 @@ mod tests {
         while let Some(next) = it.next() {
             let Ok(entry) = next else { continue };
             let is_dir = entry.file_type().is_dir();
-            if entry.depth() > 0
-                && entry
-                    .file_name()
-                    .to_str()
-                    .is_some_and(|n| n.starts_with('.'))
-            {
+            if entry.depth() > 0 && is_hidden_name(entry.file_name()) {
                 stack.record_hidden(is_dir);
                 if is_dir {
                     it.skip_current_dir();
@@ -1098,6 +1097,47 @@ mod tests {
                 .any(|l| l.contains("(3 non-hidden files inside them)")),
             "{lines:?}"
         );
+    }
+
+    /// After the walker started treating a non-UTF-8 `.\x80` name as hidden,
+    /// `count_inside` still used `to_str().starts_with('.')` and counted those
+    /// files (and descended those directories). APFS rejects the name; Linux
+    /// does not.
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn count_inside_skips_hidden_non_utf8_names() {
+        use std::ffi::OsStr;
+        use std::os::unix::ffi::OsStrExt;
+
+        let dir = tempfile::TempDir::new().unwrap();
+        let root = dir.path();
+        write(root, ".xerjignore", "pruned/\n");
+        write(root, "keep.md", "x");
+        write(root, "pruned/visible1.txt", "x");
+        write(root, "pruned/.hiddenutf8", "x");
+        fs::write(
+            root.join("pruned").join(OsStr::from_bytes(b".\x80hidden")),
+            "x",
+        )
+        .unwrap();
+        fs::create_dir(root.join("pruned").join(OsStr::from_bytes(b".\x80dir"))).unwrap();
+        fs::write(
+            root.join("pruned")
+                .join(OsStr::from_bytes(b".\x80dir"))
+                .join("inside.txt"),
+            "x",
+        )
+        .unwrap();
+
+        let (kept, deep) = survivors(root, IgnoreOptions::default().with_deep_count(true));
+        assert_eq!(kept, vec!["keep.md".to_string()], "{kept:?}");
+        assert_eq!(
+            deep.files_inside_pruned_dirs, 1,
+            "only pruned/visible1.txt is a non-hidden file inside the pruned \
+             tree; .hiddenutf8, .\\x80hidden, and .\\x80dir/inside.txt are \
+             hidden: {deep:?}"
+        );
+        assert!(deep.files_inside_pruned_dirs_is_exact());
     }
 
     /// #279. The claim the reporting makes about its own number, checked as
--- a/engine/crates/xerj-autoindex/src/walk.rs
+++ b/engine/crates/xerj-autoindex/src/walk.rs
@@ -9,7 +9,15 @@
 
 use crate::ignore_rules::{IgnoreOptions, IgnoreReport, IgnoreStack};
 use anyhow::{Context, Result};
+use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
+
+/// Hidden names start with `.` even when they are not valid UTF-8.
+/// `OsStr::to_str()` is `None` for those, so a UTF-8 `starts_with('.')`
+/// would walk them.
+fn is_hidden_name(name: &OsStr) -> bool {
+    name.as_encoded_bytes().first() == Some(&b'.')
+}
 
 #[derive(Debug, Clone)]
 pub struct FileEntry {
@@ -139,12 +147,7 @@ pub fn walk_reporting(
         // itself (depth 0) is exempt so a brain over a dot-named folder still
         // works. `--no-ignore` does NOT turn this off: it is not one of the
         // ignore rules, it is the reason secrets stay out.
-        if entry.depth() > 0
-            && entry
-                .file_name()
-                .to_str()
-                .is_some_and(|n| n.starts_with('.'))
-        {
+        if entry.depth() > 0 && is_hidden_name(entry.file_name()) {
             stack.record_hidden(is_dir);
             if is_dir {
                 it.skip_current_dir();
@@ -267,6 +270,58 @@ mod hidden_skip_tests {
         assert!(
             !rels.iter().any(|r| r.ends_with(".secret")),
             "indexed a nested dotfile: {rels:?}"
+        );
+    }
+
+    /// A name whose first byte is '.' but is not valid UTF-8 must still be
+    /// treated as hidden. `file_name().to_str()` is `None` for that name, so
+    /// the UTF-8 `starts_with('.')` check misses it.
+    #[cfg(unix)]
+    #[test]
+    fn hidden_non_utf8_dot_name_is_detected() {
+        use super::is_hidden_name;
+        use std::ffi::OsStr;
+        use std::os::unix::ffi::OsStrExt;
+
+        let hidden = OsStr::from_bytes(b".\x80");
+        assert!(
+            hidden.to_str().is_none(),
+            "fixture must be non-UTF-8 so to_str() cannot save us"
+        );
+        assert!(
+            is_hidden_name(hidden),
+            "first byte is '.', even when to_str() is None"
+        );
+        assert!(is_hidden_name(OsStr::from_bytes(b".env")));
+        assert!(!is_hidden_name(OsStr::from_bytes(b"README.md")));
+        assert!(!is_hidden_name(OsStr::from_bytes(b"\x80env")));
+    }
+
+    /// Same name on disk. APFS/HFS reject non-UTF-8 names; Linux does not.
+    #[cfg(target_os = "linux")]
+    #[test]
+    fn hidden_non_utf8_dotfile_is_never_indexed() {
+        use std::ffi::OsStr;
+        use std::os::unix::ffi::OsStrExt;
+
+        let dir = tempfile::TempDir::new().unwrap();
+        let root = dir.path();
+        fs::write(root.join("README.md"), "hello").unwrap();
+        fs::write(root.join(OsStr::from_bytes(b".\x80")), "SECRET=pat-abc123").unwrap();
+
+        let files = walk(root, false).unwrap();
+        let rels: Vec<String> = files.iter().map(|e| e.rel.clone()).collect();
+        assert!(
+            rels.contains(&"README.md".to_string()),
+            "visible file must still be indexed: {rels:?}"
+        );
+        assert!(
+            !files.iter().any(|e| {
+                e.path
+                    .file_name()
+                    .is_some_and(|n| n.as_encoded_bytes().first() == Some(&b'.'))
+            }),
+            "indexed a hidden non-UTF-8 name: {rels:?}"
         );
     }
 
--- a/engine/crates/xerj-autoindex/src/walk.rs
+++ b/engine/crates/xerj-autoindex/src/walk.rs
@@ -7,17 +7,9 @@
 //! descended, so its contents cost nothing — no stat, no hash, no bulk body.
 //! The matching itself lives in [`crate::ignore_rules`].
 
-use crate::ignore_rules::{IgnoreOptions, IgnoreReport, IgnoreStack};
+use crate::ignore_rules::{is_hidden_name, IgnoreOptions, IgnoreReport, IgnoreStack};
 use anyhow::{Context, Result};
-use std::ffi::OsStr;
 use std::path::{Path, PathBuf};
-
-/// Hidden names start with `.` even when they are not valid UTF-8.
-/// `OsStr::to_str()` is `None` for those, so a UTF-8 `starts_with('.')`
-/// would walk them.
-fn is_hidden_name(name: &OsStr) -> bool {
-    name.as_encoded_bytes().first() == Some(&b'.')
-}
 
 #[derive(Debug, Clone)]
 pub struct FileEntry {
@@ -279,7 +271,7 @@ mod hidden_skip_tests {
     #[cfg(unix)]
     #[test]
     fn hidden_non_utf8_dot_name_is_detected() {
-        use super::is_hidden_name;
+        use crate::ignore_rules::is_hidden_name;
         use std::ffi::OsStr;
         use std::os::unix::ffi::OsStrExt;
 


xerj-org:
## Independent verification — security-and-scope lens: NOT REFUTED, 1 must-fix

> NOT REFUTED on correctness. I attacked this from the filter-bypass-siblings angle and could not break the fix. The severity claim is real and stronger than stated: on the base commit a `.\x80ssh/`, `.\x80aws/` and `.\x80env` tree is walked with full credential bodies readable at the ingest boundary; on the PR head only README.md survives. I hunted the sibling sites the lens asked for and the hunt came up almost empty — `.xerjignore`/gitignore matching does NOT have the same hole (globset matches raw bytes on unix; `secret*`, `*.pem`, `*.key` each caught both the UTF-8 and the non-UTF-8 file), `--no-ignore`/`--no-default-ignores` do not disable the widened rule, `walk.rs` is the only production folder walker in the workspace, and the Windows over-skip worry is refuted (exhaustive enumeration: no Unicode scalar's UTF-8/WTF-8 encoding has 0x2E as its first byte, so a leading 0x2E is unambiguously '.'; the `cfg(target_os="linux")` gate is about fixture creation, not the predicate).
> 
> What I found that the previous pass missed is not a defect in the diff but a release-readiness problem around it. (1) The fix has no remediation path and, worse, HARD-FAILS the next run for exactly the population it protects: a corpus indexed by a pre-fix build holds a content group for `.�ssh/id_rsa`; after upgrade the walk no longer yields it, `UnsupportedInventoryDelta::refuses()` returns true, and the rerun is refused with an error whose cheapest suggested recovery is literally "restore the removed file(s)" — put the SSH key back where the indexer can see it. The already-exposed documents stay live and searchable, and the refusal text echoes the secret path to stderr and `--json`. Nothing in the CHANGELOG mentions this. (2) The previous pass filed `count_inside` as a note; with numbers it is a user-visible defect, not a note — it reports 3 where the correct answer is 1, it is production code reached by `--dry-run`, it is published as the JSON key `ignored_files_in_pruned_dirs`, and it now contradicts both its own rustdoc and docs/recipes/zero-config-indexing.md:284. After this PR the walker and the reporter disagree about what "hidden" means.
> 
> Separately, for the maintainer weighing "does this close the class": `--follow-symlinks` is a far larger hole of the identical shape, present unchanged on both sides. It indexes `.secretdir/k.txt` under a visible name and escapes the indexed root entirely (`shared/etc/shadow`), while the report simultaneously claims `hidden:dotfile dirs: 2`. That needs no exotic filename — just a documented flag whose help text advertises only "loop-safe".
> 
> I did not merge, close, approve, review or push anything.

### [must-fix] Upgrading to this fix hard-fails the next autoindex run and tells the user to restore their SSH key; already-indexed secrets stay live

The fix stops NEW exposure but ships no remediation, and the population it protects is exactly the population it breaks. A corpus indexed by a pre-fix build has a `FileAssignment` in its frozen resume plan for the non-UTF-8 dot-name file. After upgrading, the walk no longer yields it, so in `UnsupportedInventoryDelta::between` (engine/crates/xerj-autoindex/src/lib.rs:2526) neither `current_keys`, `current_rels` nor `current_path_ids` contains it, `path_survives` is false, the group is classified vanished, and `refuses()` (lib.rs:2607) returns true. The run is refused with `UnsupportedInventoryDeltaError`. Three consequences, all bad for a security fix: (a) the previously-indexed secret documents remain live and searchable — the error says so verbatim, and the rustdoc at lib.rs:2603 confirms 'those documents stay live and searchable with no source file behind them, and nothing in this pipeline removes them'; (b) the cheapest recovery the error offers is 'restore the removed file(s) and rerun', i.e. put the credential back where the indexer will pick it up — the exact opposite of what the fix wants; the only correct remedy is deleting the indices plus the state dir and re-embedding the whole corpus, and `--fresh` is refused too; (c) the refusal message renders the secret path into stderr and into `--json`. The CHANGELOG entry says nothing about any of this. This does not make the code change wrong, but it should not merge as a silent security fix — it needs a CHANGELOG/release note naming the upgrade behaviour and the purge procedure, and ideally a classifier exemption so a group that vanished because it became hidden is not treated as a user deletion.

```
Temporary probe added to the PR-head worktree's own `inventory_delta_tests` module (reverted afterwards; `git status --short` is now empty and `git diff 9ebdac2 --stat` is exactly `CHANGELOG.md | 7 ++++` and `engine/crates/xerj-autoindex/src/walk.rs | 67 +++...`). The plan was seeded with the exact `rel` and `path_id` MEASURED from the pre-fix walk on 9ebdac2 (`rel=".\u{FFFD}ssh/id_rsa"`, `rel_id=unix:2e807373682f69645f727361`).

  cd /tmp/v425-security/engine && CARGO_TARGET_DIR=/home/claude/work/xerj-main/engine/target rustup run 1.97.1 cargo test -p xerj-autoindex --lib probe9 -- --nocapture --test-threads=1

  running 1 test
  test inventory_delta_tests::probe9_prefix_corpus_upgrade_path ... PROBE9 post-fix inventory: ["README.md"]
  PROBE9 vanished_content_groups = [".\u{FFFD}ssh/id_rsa"]
  PROBE9 refuses() = true
  PROBE9 rerun error:
  1 file(s) indexed under this resume plan no longer exist in the folder, and their documents are still live in the destination. Removing files from an indexed folder is not reconciled yet, so this attempt made no remote mutations [...] Removed content groups [.\u{FFFD}ssh/id_rsa (secretkey-41)]. Recovery, cheapest first: (1) restore the removed
```

### [should-fix] `count_inside` reports 3 where the answer is 1: after this PR the walker and the --dry-run reporter disagree about what "hidden" means

engine/crates/xerj-autoindex/src/ignore_rules.rs:734-737 still uses `e.file_name().to_str().is_some_and(|name| name.starts_with('.'))` — the exact predicate this PR replaced, 570 lines away, in the same crate, in PRODUCTION code (inside `impl IgnoreStack`, above the `#[cfg(test)]` module). The previous pass filed this as a note qualified with '--dry-run reporting only'. With numbers it is a user-visible defect: it counts hidden non-UTF-8 FILES and it DESCENDS hidden non-UTF-8 DIRECTORIES. It is reachable from the CLI — cli.rs:767 sets `deep_count: dry_run` — and the number is published to machines as `ignored_files_in_pruned_dirs` (lib.rs:3102 and lib.rs:3777). It now falsifies two written contracts: the field's own rustdoc ('Non-hidden files inside the pruned directories [...] Directories pruned by HIDDEN_RULE contribute nothing') and docs/recipes/zero-config-indexing.md:284 ('`--dry-run` also counts the non-hidden files inside each pruned directory, so "where did my file go?" always has an answer'). Before this PR the two sites were at least consistently wrong; after it, `xerj autoindex --dry-run` tells the user it saved them 3 files when the fixed walker would only ever have indexed 1. This is a two-line change to reuse `is_hidden_name`, and the fix is incomplete without it.

```
Fixture: `pruned/` (ignored by `.xerjignore`) containing `visible1.txt`, `.hiddenutf8`, `.\x80hidden`, and `.\x80dir/inside.txt`. Correct count = 1 (`visible1.txt` alone).

  cd /tmp/v425-security/engine && CARGO_TARGET_DIR=/home/claude/work/xerj-main/engine/target rustup run 1.97.1 cargo test -p xerj-autoindex --test zz_lens_probe -- --nocapture --test-threads=2

  PROBE4 survivors: ["rel=\"keep.md\" ..."]
  PROBE4 files_inside_pruned_dirs=3 exact_flag_ran=true complete=true
  PROBE4 by_rule={".xerjignore:pruned/": RuleCount { files: 0, dirs: 1, files_inside_pruned_dirs: 3, dirs_deep_counted: 1 }, ...}

Byte-identical on the base commit (cd /tmp/v425-sec-main/engine && CARGO_TARGET_DIR=/tmp/tgt-425-main rustup run 1.97.1 cargo test -p xerj-autoindex --test zz_lens_probe -- --nocapture --test-threads=2 -> `PROBE4 files_inside_pruned_dirs=3`), i.e. the PR left it untouched. 3 = visible1.txt + .\x80hidden + .\x80dir/inside.txt; `.hiddenutf8` was correctly excluded, which is what proves the predicate is the cause.
```

### [note] `--follow-symlinks` is a strictly larger hole of the same shape, unfixed on both sides, and the ignore report actively misreports it

Context for the maintainer weighing whether this PR closes the hidden-file class: it does not, and the bigger door needs no exotic filename. With `--follow-symlinks`, a visible-named symlink defeats the dotfile rule completely AND escapes the indexed root. The walker's own security comment (walk.rs:137-146, unchanged by this PR) asserts the hidden rule 'is the reason secrets stay out'; that claim is false under a flag whose CLI help (cli.rs:154) advertises only 'follow symlinks (loop-safe); off by default' and whose docs (docs/recipes/zero-config-indexing.md:270) say only '(loop-safe)'. Neither mentions that following symlinks bypasses the secret rule. Worse, the ignore report simultaneously prints `hidden:dotfile dirs: 2`, so the run tells the user it pruned the hidden directories while indexing their contents through the link. The non-UTF-8 bug required a filename containing invalid UTF-8 bytes, which essentially never happens by accident; this one requires a symlink and a flag. Not a reason to block the PR, but the PR should not be described as closing the class, and the flag deserves either a hidden-target check or a documented warning.

```
  cd /tmp/v425-security/engine && CARGO_TARGET_DIR=/home/claude/work/xerj-main/engine/target rustup run 1.97.1 cargo test -p xerj-autoindex --test zz_lens_probe probe10 -- --nocapture --test-threads=1

  PROBE10 follow=false:
    rel="README.md" is_symlink=false body="hi"
    report by_rule={"hidden:dotfile": RuleCount { files: 0, dirs: 1, ... }}
  PROBE10 follow=true:
    rel="README.md" is_symlink=false body="hi"
    rel="notes.txt" is_symlink=true body="INNER SECRET"
    rel="shared/etc/shadow" is_symlink=false body="root:$6$OUTSIDE_THE_ROOT$hash:19000:0:99999:7:::"
    report by_rule={"hidden:dotfile": RuleCount { files: 0, dirs: 2, ... }}

`notes.txt` is a visible-named symlink to `.secretdir/k.txt`; `shared/` is a symlink to a DIFFERENT tempdir entirely, so `shared/etc/shadow` is content from outside the indexed root. PROBE5 shows the same with a `.ssh/` directory symlink: `follow_symlinks=true -> [..., rel="docs_link/id_rsa"]`. Identical output on base 9ebdac2 and on PR head a8b9d79.
```

### [note] Sibling-bypass hunt came up empty apart from `count_inside` — ignore-rule matching is byte-correct and is NOT evadable

Reporting this as a positive result because it is what the lens asked and it should stop the maintainer worrying about a wider class. (1) `.xerjignore`/gitignore matching handles non-UTF-8 names correctly on unix — the `ignore`/globset matcher works on raw `OsStr` bytes, so a user rule cannot be evaded the way the hidden check was. (2) A nested `.xerjignore` inside a non-UTF-8-named directory is read and applied (`ignore_files_read=1`, rule label `sub�/.xerjignore:hide.txt`); only the label is lossy. (3) `--no-ignore` and `--no-default-ignores` do not disable the widened hidden rule, as walk.rs:137-146 claims. (4) `walk.rs` is the only production folder walker in the workspace — three `WalkDir::new` sites total, one of which is the test helper and one of which is `count_inside`; no second walker in xerj-mcp or xerj-server, and the JS hits are UI mock data. (5) The other `file_name().to_str()` sites are not filters and fail degraded, not open: `extract/code.rs:431` and `extract/mod.rs:181` fall back to a literal "source"/"file" title; `extract/unity.rs:305` just omits `asset_name` for a non-UTF-8 `.meta` sidecar; `sniff.rs:344` falls back from extension to content sniffing; `xerj-engine/src/engine.rs:902` reads XERJ's own generated data-dir names, not corpus names. (6) No over-skip: 5 of 5 visible non-UTF-8 names kept, byte-identical before and after, including the adversarial `\xff.ssh` and `\xc3.env` where a '.' appears but not first. (7) Windows/macOS: `OsStr::as_encoded_bytes` returns a self-synchronizing superset of UTF-8 (WTF-8 on Windows), and exhaustive enumeration of all scalars U+0080..U+10FFFF found zero whose UTF-8 encoding begins with 0x2E — so the byte check is correct on every platform and the `cfg(target_os = "linux")` gate is about fixture creation (APFS

```
  PROBE2 report: by_rule={".xerjignore:*.key": RuleCount { files: 2, ... }, ".xerjignore:*.pem": RuleCount { files: 2, ... }, ".xerjignore:secret*": RuleCount { files: 2, ... }, ...}
(2 files per rule = the UTF-8 control AND the non-UTF-8 twin `secret\x80.txt` / `server\x80.pem` / `prod\x80.key`; identical on base and head. The `vault\x80/b.txt` survivor is correct behaviour, not a bypass: `vault\x80` is a different directory name from `vault`.)

  PROBE6 kept 5 of 5:
    rel="a�.b"  rel="caf�.md"  rel="�.env"  rel="�.ssh"  rel="�env"
  PROBE6 by_rule={}

  PROBE8 scalars whose UTF-8 first byte is 0x2E: []
  PROBE11 --no-ignore -> ["rel=\"README.md\" ..."]
  PROBE11 --no-default-ignores -> ["rel=\"README.md\" ..."]

  PROBE7 ignore_files_read=1 by_rule={..., "sub�/.xerjignore:hide.txt": RuleCount { files: 1, ... }} warnings=[]

  $ grep -rn "WalkDir::new" engine/crates/xerj-autoindex/src/*.rs
  ignore_rules.rs:729 (count_inside)  ignore_rules.rs:763 (test helper survivors)  walk.rs:112
```

### [note] `rel` is lossy so distinct files collide on the ax_path key — pre-existing, but this PR narrows the surviving population to exactly the colliding one

walk.rs:180 builds `rel` with `to_string_lossy()`, and `rel` is the user-facing `ax_path` value. Three distinct files on disk collapse to one `rel`. Identical on base and head, so this is NOT introduced by the PR and does not block it — the code already knows, hence the `rel_id` field documented as 'Reversible platform-native identity; unlike `rel`, never lossy'. It is worth flagging only because this PR changes the population: after the fix, the non-UTF-8 names that still reach the index are precisely the VISIBLE ones, which are precisely the ones that collide. Two live consequences: `UnsupportedInventoryDelta::path_survives` (lib.rs:2554) accepts a match on `rel` as one of its two keys, and the reconcile fallback at lib.rs:2220 does `.get(file.rel_id).or_else(|| planned_by_rel.get(file.rel))` — so colliding rels can cross-attribute; and `out.sort_by(|a, b| a.rel.cmp(&b.rel))` with the comment 'Deterministic order for clustering / naming' is not deterministic across runs for colliding names, because the tie is broken by readdir order. `doc_id` itself is safe (it keys on content hash + size, ids.rs:27), so this is not a document-overwrite bug.

```
  PROBE3:
    rel="�a.txt" rel_id=unix:80612e747874 bytes=[128, 97, 46, 116, 120, 116]
    rel="�a.txt" rel_id=unix:81612e747874 bytes=[129, 97, 46, 116, 120, 116]
    rel="�a.txt" rel_id=unix:fe612e747874 bytes=[254, 97, 46, 116, 120, 116]
  PROBE3 distinct files on disk = 3, distinct rel = 1

Byte-identical on base 9ebdac2 and PR head a8b9d79.
```

### [note] Severity of the original bug confirmed end-to-end at the ingest boundary, and the fix does close it

Recorded so the maintainer has the concrete before/after for the security framing. On the base commit a credential-shaped tree under non-UTF-8 dot names is walked with the ignore report showing NOTHING was skipped (`files_skipped=0 dirs_pruned=0 by_rule={}`) — the run would tell the user it excluded nothing while handing an OpenSSH private key, an AWS credentials file, an SSH config naming a prod host as root, and a live Stripe key to the indexer. On the PR head the same fixture yields only README.md and the report correctly attributes `hidden:dotfile { files: 1, dirs: 2 }`. I read the file bodies straight off `FileEntry.path`, which is the input the extract/bulk stage consumes, so this is the real ingest payload — see not_verified for the one step I deliberately did not take.

```
BASE 9ebdac2 (cd /tmp/v425-sec-main/engine && CARGO_TARGET_DIR=/tmp/tgt-425-main rustup run 1.97.1 cargo test -p xerj-autoindex --test zz_lens_probe -- --nocapture --test-threads=2):
  PROBE1 files:
    rel=".�aws/credentials"  rel=".�env"  rel=".�ssh/config"  rel=".�ssh/id_rsa"  rel="README.md"
  PROBE1 report: files_skipped=0 dirs_pruned=0 by_rule={}
  PROBE1 body[.�aws/credentials] = "[default]\naws_access_key_id = AKIAIOSFODNN7EXAMPLE\naws_secret_access_key = wJalrXUtnFEMI\n"
  PROBE1 body[.�env] = "STRIPE_KEY=sk_live_deadbeef\n"
  PROBE1 body[.�ssh/config] = "Host prod\n  User root\n"
  PROBE1 body[.�ssh/id_rsa] = "-----BEGIN OPENSSH PRIVATE KEY-----\nAAAAsecretkeymaterial\n-----END OPENSSH PRIVATE KEY-----\n"

PR HEAD a8b9d79 (same command, CARGO_TARGET_DIR=/home/claude/work/xerj-main/engine/target):
  PROBE1 files:
    rel="README.md" rel_id=unix:524541444d452e6d64 bytes=[82, 69, 65, 68, 77, 69, 46, 109, 100]
  PROBE1 report: files_skipped=1 dirs_pruned=2 by_rule={"hidden:dotfile": RuleCount { files: 1, dirs: 2, files_inside_pruned_dirs: 0, dirs_deep_counted: 0 }}
  PROBE1 body[README.md] = "hello"
```


---

## Maintainer note

@SebTardif — **the fix itself survived a hard attack.** The lens went after sibling bypasses specifically and came back with a positive result worth stating: `.xerjignore`/gitignore matching is *not* evadable the same way (globset matches raw `OsStr` bytes, so `secret*`, `*.pem`, `*.key` each caught both the UTF-8 file and its non-UTF-8 twin), `--no-ignore` does not disable the widened rule, `walk.rs` is the only production walker in the workspace, and the Windows over-skip worry is refuted by exhaustive enumeration — no Unicode scalar's UTF-8 or WTF-8 encoding has `0x2E` as its first byte, so a leading `0x2E` is unambiguously `.`.

Your severity claim also came out **stronger** than you wrote it. On the base commit the credential fixture is walked *and the ignore report says `files_skipped=0 dirs_pruned=0 by_rule={}`* — the run tells the operator it excluded nothing while handing an OpenSSH private key, an AWS credentials file and a live Stripe key to the indexer. On your head, only `README.md` survives and the report correctly attributes `hidden:dotfile { files: 1, dirs: 2 }`.

Two things to settle before this merges, neither of which is a defect in your diff:

**1. The upgrade path hard-fails for exactly the people this protects.** A corpus indexed by a pre-fix build holds a content group for `.<80>ssh/id_rsa`. After upgrading, the walk no longer yields it, `UnsupportedInventoryDelta::refuses()` returns true, and the next run is refused. Three consequences: the previously-indexed secrets **stay live and searchable** (the rustdoc at `lib.rs:2603` says so — "nothing in this pipeline removes them"); the error's cheapest suggested recovery is *"restore the removed file(s) and rerun"*, i.e. put the credential back where the indexer will find it; and the refusal renders the secret path into stderr and `--json`. A security fix should not ship with that as its migration story, even if the code is right. It needs at minimum a CHANGELOG note stating that existing indices retain the exposed documents and that the remedy is deleting the indices plus the state dir.

**2. `count_inside` now disagrees with the walker.** `ignore_rules.rs:734-737` still uses the exact predicate your PR replaced — in production code, 570 lines away in the same crate. Measured: `--dry-run` reports `files_inside_pruned_dirs=3` where the fixed walker would index **1**. It is reachable from the CLI (`cli.rs:767`), published to machines as `ignored_files_in_pruned_dirs`, and it falsifies both the field's own rustdoc and `docs/recipes/zero-config-indexing.md:284`. Two lines, reusing `is_hidden_name`.

Separately I have filed **#438** for something the lens found that is out of scope here but larger: `--follow-symlinks` defeats the hidden-file rule entirely with an ordinary visible filename, and escapes the indexed root — while the report still claims the hidden directories were pruned. Not yours to fix, and not a reason to hold this PR, but it means neither of us should describe this change as closing the hidden-file class.

Happy to push both items above to your branch if you'd rather not — say the word.

xerj-org:
## Independent verification — behaviour-and-migration lens: REFUTED

Panel complete: 2 lenses, 1 refuted, **2 must-fix — and they are the same defect found independently from two directions.**

> The walker fix itself is correct and the FULL workspace suite is green (131 targets, 3106 passed, 0 failed, CARGO_EXIT=0) — but the PR ships a silent corpus-membership change with no migration story, and I reproduced a hard, permanent break for exactly the population it protects. On the DEFAULT path (graph enabled, and `xerj brain`), the first run after upgrading exits 1 with `unsupported_content_group_removal`, makes no mutations, and stays wedged on every subsequent run — while the hidden-file documents and graph edges it was meant to keep out stay live and queryable. The error's own first recovery route ("restore the removed file(s) and rerun") is impossible: the file was never removed and is still on disk. A base→base control on the identical corpus and state dir exits 0, so the upgrade is the trigger. On the opt-in `--no-graph` path the run succeeds and deletes the data documents, but the sampled field values from the hidden file (card numbers, tokens) survive permanently in `autoindex-catalog` and are printed by `xerj autoindex map` — the agent-facing data map. Separately I quantified the `count_inside` divergence the earlier pass only flagged: `--dry-run` now reports `ignored_files_in_pruned_dirs=53 ... exact=true` where the fixed walker's own rule says 2, reopening a defect this project already fixed once in #287. The PR's checklist claim "no durable-state change" is false, and docs/CONTRIBUTION_REVIEW.md's own rules on preserving the product contract and on tested operator migration procedures are unmet. Recommend: do not merge until an upgrade note (and ideally a distinct refusal message or an automatic sweep for newly-hidden previously-indexed groups) lands. No state-changing command was run; no merge/close/approve/push.

### [must-fix] Upgrading permanently wedges every graph-enabled (default) corpus that contains a hidden non-UTF-8 file — and the secrets stay live

`--no-graph` corpora go through the durable generation reconciler, which turns a vanished file into a Delete operation. Graph-enabled corpora — the DEFAULT (you must pass `--no-graph` to opt out) and what `xerj brain` uses (crates/xerj-server/src/brain.rs:585 sets `no_graph: false`) — go through the resume-plan path, where `lib.rs:2491` refuses any removal outright.

The fix makes an on-disk file look removed. So the first post-upgrade run of the default path aborts with exit 1 and `xerj-done ok=false exit=1 reason=aborted`, writes nothing, and every rerun reproduces it identically. The documents and graph edges from the hidden path — the exact data the PR exists to keep out — remain live and queryable.

Worse, the machine-readable refusal (`schema: xerj.autoindex.unsupported_sync_delta.v1`) leads with `"restore_removed_files": "put the listed file(s) back and rerun"`. That is unfollowable — the file was never removed. An agent parsing this JSON will take route 1 first and loop. docs/CONTRIBUTION_REVIEW.md:33 is explicit: "give an executable recovery route. Test that the recovery route actually works; an error that merely says 'retry' or recommends a now-rejected flag is not a recovery contract."

I verified the surviving routes: (2) delete the published indices + state dir and re-extract/re-embed the whole corpus — works, expensive; (3) isolated rebuild with new --state-dir/--prefix — works (exit 0) but leaves the old target's secrets live until manually cleaned. `--fresh` is refused. `--no-graph` on the wedged state dir is refused with a *different* error ("this state directory contains a legacy nonempty plan that cannot become generation authority").

Control proves attribution: base→base on the identical corpus and state dir exits 0; only the binary swap breaks it. The refusal mechanism itself is pre-existing (base binary + a genuinely deleted file produces the same message), but this PR fires it with no user action and no CHANGELOG or PR-body warning.

Suggeste

```
Binaries: base = PR parent 9ebdac2, head = PR head a8b9d79, both built with `RUSTFLAGS="-C target-cpu=native" CARGO_TARGET_DIR=/tmp/v425-native rustup run 1.97.1 cargo build -j 20 -p xerj-server --bin xerj`. Isolated server on :19425 (shared :9200 corpus untouched, refsym still 75578 docs). Repro script: /tmp/v425-repro.sh

$ bash /tmp/v425-repro.sh
### 1. index with the PRE-FIX binary (PR parent 9ebdac2), graph on = the DEFAULT
    exit=0
### 2. the hidden secret is live:
    hits = 1
### 3. upgrade: SAME folder, SAME state dir, PR head a8b9d79. Nothing on disk changed.
error: 1 file(s) indexed under this resume plan no longer exist in the folder, and their documents are still live in the destination. Removing files from an indexed folder is not reconciled yet, so this attempt made no remote mutations — no documents, aliases, graph edges or catalog entries were written. Removed content groups [.\xef\xbf\xbdsecrets.csv (axf2-8b2556864a034886e9fddf97d206dff9-18)]. Recovery, cheapest first: (1) restore the removed file(s) and rerun ...
    exit=1
### 4. the file the error says was removed is still on disk:
    /tmp/v425-repro/.M-^@secrets.csv
    /tmp/v425-repro/visible.csv
### 5. the secret is still live after the failed upgrade:
    hits = 1

Reproduced twice independently: markdown corpus (/tmp/v425-d, "Removed content groups [.�secretnote.md ...]", edges count still 9) and CSV corpus (/tmp/v425-f). Rerun after the failure gives byte-identical error, exit=1.

Agent surface:

```

### [should-fix] After the --no-graph upgrade the secret VALUES survive forever in autoindex-catalog and are printed by `xerj autoindex map`

On the `--no-graph` path the upgrade run succeeds (exit 0, generation advances) and the data documents from the hidden non-UTF-8 file are correctly deleted, as is its `file:` catalog entry. But the `ds:` dataset catalog entry survives with `file_count: 0, record_count: 0` and its full `fields_json`, which carries `examples` sampled from the hidden file — i.e. the literal secret values. The empty destination index, whose name is derived from the hidden path, also remains.

This is permanent: it is still there after a third run. And it is not obscure — `xerj autoindex map`, the data map an AI agent reads, renders those values in a markdown table.

Honest attribution: a control (same head binary, an ordinary VISIBLE file deleted between two runs) produces the identical residue, so this is pre-existing generic removal behaviour, not a defect introduced by this PR. But the PR is what silently converts "a hidden secret file" into "a removed file", and the PR's whole premise is that these values stay out. On the one upgrade path that does not hard-fail, they do not. This needs at minimum a line in the migration note telling operators to delete the residual `ds:` catalog docs and the empty indices, and ideally a catalog sweep.

```
Corpus /tmp/v425-h: visible.csv + `.\x80paydata.csv` containing `id,card_pan,holder / 1,4111111111111111,alice / 2,4222222222222222,bob`.

$ /tmp/v425-bin/xerj-base autoindex /tmp/v425-h --url http://127.0.0.1:19425 --prefix vh --state-dir /tmp/v425-h-state --no-graph --progress none --yes
generation 1 committed — 2 datasets, 4 records live
$ /tmp/v425-bin/xerj-head autoindex /tmp/v425-h ... (same args)
generation 2 committed — 2 datasets, 2 records live
EXIT=0

$ curl -s http://127.0.0.1:19425/vh-*/_count  -> {"count":2}   (data docs correctly gone)
$ curl -s 'http://127.0.0.1:19425/_cat/indices/vh-*?v'
green open vh-csv-visible ... 2 docs
green open vh-csv-paydata ... 0 docs        <- empty index named from the hidden path, remains

Catalog residue (still present after a THIRD head run):
ds:csv-paydata | idx vh-csv-paydata | files 0 | recs 0
    fields_json: [{"name":"card_pan","es_type":"long",...,"examples":["4111111111111111","4222222222222222"]},{"name":"holder",...,"examples":["alice","bob"]},...]

$ /tmp/v425-bin/xerj-head autoindex map --url http://127.0.0.1:19425 --dataset csv-paydata
### `vh-csv-paydata`
| field | type | semantic | cardinality | null% | examples |
| `card_pan` | long | — | 2 | 0% | `4111111111111111`, `4222222222222222` |
| `holder` | keyword | — | 2 | 0% | `alice`, `bob` |

CONTROL (pre-existing, not introduced here): head binary, corpus /tmp/v425-ctl with a VISIBLE secrets.csv deleted between two runs -> `ds:csv-secrets | idx ctl-csv-secrets | fi
```

### [should-fix] --dry-run now over-reports `ignored_files_in_pruned_dirs` by 26.5x and still asserts exact=true — reopening the defect #287 fixed

The earlier pass noted that `IgnoreStack::count_inside` still uses the pre-fix `to_str()` check. Quantified at the product surface: a gitignore-pruned directory containing non-UTF-8 dot names is deep-counted as if those names were visible, including descending a non-UTF-8 dot-named *directory's* whole subtree.

The printed phrase is literally "53 non-hidden files inside them" when the fixed walker's own hidden rule says 2 of them are non-hidden, and the machine field ships `ignored_files_in_pruned_dirs_exact=true`, asserting a total rather than a floor. The doc comment on `count_inside` in the same file now states something false of the code: "Regular files under `dir`, excluding every dot-named file and the whole subtree of every dot-named directory — the dotfile rule, applied here too".

This is the same field and the same surface the project already repaired once: CHANGELOG (#287, repairing #279) records "hidden directories pruned by the dotfile rule were deep-counted into `ignored_files_in_pruned_dirs` (97,731 phantom files on this repository's own walk); that number reached the `xerj-done` line and `--progress json` as a bare total". The narrower non-UTF-8 case is now back.

Why CI cannot see it: `ignore_rules.rs`'s test helper `survivors()` still uses the old `to_str()` check, so the entire ignore-rule suite models a pre-fix walker; `ignore_rules.rs` contains no reference to `is_hidden_name` or `as_encoded_bytes`. Fix is one line — have `count_inside`'s `filter_entry` call `walk::is_hidden_name` — plus updating `survivors()`.

```
Fixture /tmp/v425-g: keep.md, .gitignore("pruned/"), pruned/{a.txt,b.txt,.hidden.txt,`.\x80`}, pruned/`.\x80cache`/x0..x49.

$ /tmp/v425-bin/xerj-head autoindex /tmp/v425-g --url http://127.0.0.1:19425 --prefix vg --state-dir /tmp/v425-g-state --no-graph --dry-run --progress plain --yes
autoindex: ignore rules: skipped 1 file and pruned 1 directory (53 non-hidden files inside them); 1 ignore file read
autoindex: ignore rules:   .gitignore:pruned/ — 1 directory pruned (53 non-hidden files inside)
xerj-done ok=true exit=0 reason=dry-run wall=1.4s files=1 ignored_files=1 ignored_dirs=1 ignored_files_in_pruned_dirs=53 ignored_files_in_pruned_dirs_exact=true

Ground truth on the SAME binary, same tree:
$ /tmp/v425-bin/xerj-head autoindex /tmp/v425-g ... --no-ignore --dry-run
autoindex: 3 files (0 MB) under /tmp/v425-g          -> keep.md + pruned/a.txt + pruned/b.txt = 2 inside pruned/
$ /tmp/v425-bin/xerj-base autoindex /tmp/v425-g ... --no-ignore --dry-run
autoindex: 54 files (0 MB) under /tmp/v425-g         -> 53 inside pruned/ (what the count still reports)

Base and head report the identical 53, i.e. the PR does not change count_inside; it changes the walker out from under it. Unit-level probe (tests/v425_probe.rs, run on BOTH 9ebdac2 and a8b9d79):
ignored_files_in_pruned_dirs = 53   (truth by the walker's own hidden rule = 2)
by_rule = {".gitignore:pruned/": RuleCount { files: 0, dirs: 1, files_inside_pruned_dirs: 53, dirs_deep_counted: 1 }, "hidden:dotfile": RuleCount { fil
```

### [should-fix] The PR's own checklist claim "no durable-state change" is false, and CONTRIBUTION_REVIEW.md's migration obligations are unmet

The PR body states: "Contribution review: applicable parts (public path, fail-closed only on the unsafe name, no durable-state change)". The change alters committed corpus membership for already-published generations and resume plans — that is precisely a durable-state change, and the reconciler refuses it (finding 1). A maintainer reading that checklist line would not go looking for the upgrade path.

The project's own docs/CONTRIBUTION_REVIEW.md sets the bar this PR does not clear:
- L22 "A fix must not turn a partial correctness problem into a wider availability problem. Before changing a guard, migration, optimization, or recovery path, enumerate the previously supported workflows and decide explicitly which remain supported." — the enumerated minimum at L27-32 names "unchanged rerun", "committed, and legacy durable state", "removals", "exit codes ... and agent-readable fields", and "the top-level `xerj autoindex` and `xerj brain` workflows when shared code changes". None appear in the PR.
- L33 requires a recovery route that is tested to work (finding 1 shows route 1 cannot).
- L59 "Exercise the upgrade against representative data from every materially distinct supported schema generation" — not done; the PR's Checks list has no upgrade run at all.
- L65 "If migration cannot be automatic, provide an exact, tested operator procedure that names the affected indices, state directories, prefixes, or generations and explains what is preserved or lost." — absent from both the PR body and the CHANGELOG entry.

The CHANGELOG diff adds only a 7-line Fixed bullet describing the byte comparison, with no upgrade note. Note the historical precedent: issue #283 was originally discovered through exactly this shape — "#279's upgrade path (index with `--no-ignore`, re-run with the new default)" — a file-set-shrinking change with no migration story.

```
$ git show a8b9d79 -- CHANGELOG.md
+- **`xerj autoindex` now skips hidden names that are not valid UTF-8.**
+  The walker already refused `.env`, `.ssh`, and `.git`, but the check
+  used `file_name().to_str().starts_with('.')`. ... The skip now compares the first encoded byte
+  to `.`. Typical UTF-8 secrets were already skipped.
(7 added lines, no upgrade/migration text)

$ sed -n '22p;33p;65p' /tmp/v425-head/docs/CONTRIBUTION_REVIEW.md
A fix must not turn a partial correctness problem into a wider availability problem. Before changing a guard, migration, optimization, or recovery path, enumerate the previously supported workflows and decide explicitly which remain supported.
If behavior must become fail-closed, refuse only the unsafe case. Name what was detected, state whether remote or durable state changed, and give an executable recovery route. Test that the recovery route actually works; an error that merely says "retry" or recommends a now-rejected flag is not a recovery contract.
If migration cannot be automatic, provide an exact, tested operator procedure that names the affected indices, state directories, prefixes, or generations and explains what is preserved or lost.

PR files changed: CHANGELOG.md (+7/-0), engine/crates/xerj-autoindex/src/walk.rs (+61/-6). No test exercises a pre-fix-built index re-run with the fixed binary.
```

### [note] `xerj autoindex status` reports FINISHED on a state dir that fails every run

After the upgrade wedges a graph-enabled corpus, the operator's diagnostic command reports the journal as healthy and finished. Nothing on that surface says the next run will abort, so an operator triaging the exit-1 has no confirmation from `status` that the state dir is the problem.

```
$ /tmp/v425-bin/xerj-head autoindex status --url http://127.0.0.1:19425 --state-dir /tmp/v425-f-state
journal /tmp/v425-f-state/journal.ndjson — root /tmp/v425-f — 2 files done, 6 records, FINISHED
  graph: 1 edges written to .xerj-memory-vfbrain-edges (brain vfbrain)

live indices at http://127.0.0.1:19425:
  autoindex-catalog                                11 docs
EXIT=0

...while every autoindex run against that same state dir exits 1.
```

### [note] Cleared: the reconcile hazards the lens asked about (#238, #283, duplicate canonicals, empty corpus) do NOT break — all on the --no-graph path

Reporting the negative results so they are not re-litigated. All four were tested base-binary-then-head-binary against the same state dir, `--no-graph`:

- #283 shape (shrinking file set on a junk-bearing generation): no abort. Both generations exit 3 (completed-with-junk); no `desired manifest digest does not match sync_begin payload`.
- #238 shape (immortal catalog entry for a skipped file): the `file:` catalog entry for the hidden non-UTF-8 junk file is correctly swept on the upgrade run.
- Duplicate content where the hidden non-UTF-8 path won the canonical election: canonical re-election is clean; `ax_path`/`ax_paths` re-point to the visible sibling with no orphan. Clean on the graph path too (no wedge, because the content group survives — only an alias leaves).
- Corpus reduced to zero files by the fix: generation 2 commits, 0 records live, exit 0.
- Data-plane deletion on --no-graph is complete: the secret rows and the indexed hidden directory's rows are gone from the destination indices.

Also cleared: the FULL workspace suite is green on the PR head, so nothing existing in CI would catch any of the above findings.

```
$ cd /tmp/v425-head/engine && RUSTFLAGS="-C target-cpu=native" CARGO_TARGET_DIR=/tmp/v425-native rustup run 1.97.1 cargo test --workspace --profile ci-test -j 20 -- --test-threads=2 > /tmp/v425-ws-tests.log 2>&1; echo "CARGO_EXIT=$?"
CARGO_EXIT=0
test targets: 131 | passed: 3106 | failed: 0 | ignored: 30
(no `error`, `FAILED`, or `panicked at` lines in the log; xerj_autoindex target itself ran 645 tests)

#283 (corpus /tmp/v425-b2, prefix vb2):
=== BASE gen1 === generation 1 committed — 1 datasets, 2 records live   EXIT=3
=== HEAD gen2 === generation 2 committed — 1 datasets, 2 records live   EXIT=3

#238 (same corpus):
--- catalog file docs after gen1 ---  'visible.csv' indexed / '.�blob.bin' junk / 'plain.bin' junk
--- catalog file docs after gen2 ---  'visible.csv' indexed / 'plain.bin' junk        <- hidden entry swept

Duplicate canonical (corpus /tmp/v425-c, --no-graph; /tmp/v425-j, graph on):
before: ax_path '.�data.csv' ax_paths ['.�data.csv', 'zz_visible.csv']
after:  ax_path 'zz_visible.csv'  ax_paths ['zz_visible.csv']        EXIT=0 both paths

Empty corpus (/tmp/v425-i): generation 2 committed — 1 datasets, 0 records live, EXIT=0, vh count 0.
```

### [note] The prescribed toolchain (1.97.1) cannot build this workspace on aarch64 with the repo's own .cargo/config.toml

Environmental, not the PR's fault, but relevant to anyone reproducing on ARM and to CI portability claims: `engine/.cargo/config.toml` sets `rustflags = []` for `aarch64-unknown-linux-gnu`, and under rustc 1.97.1 the `gemm-f16` inline asm then fails to assemble because the generic ARMv8-A baseline lacks `fullfp16`. I had to build everything with `RUSTFLAGS="-C target-cpu=native"`, which means my green workspace run is not byte-for-byte CI's flag set. Someone should confirm CI's aarch64 job (if any) is not silently relying on a different rustc.

```
$ cd /tmp/v425-upgrade/engine && CARGO_TARGET_DIR=/tmp/v425-target rustup run 1.97.1 cargo build -j 32 -p xerj-server --bin xerj
error: instruction requires: fullfp16
    --> /home/claude/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/gemm-common-0.19.0/src/simd.rs:1952:18
     |
1952 |                 "fmul {0:v}.8h, {1:v}.8h, {2:v}.8h",
error: could not compile `gemm-f16` (lib) due to 11 previous errors

$ uname -m
aarch64
$ grep -A2 'aarch64-unknown-linux-gnu' engine/.cargo/config.toml
[target.aarch64-unknown-linux-gnu]
rustflags = []
```


---

## Maintainer decision

@SebTardif — **your code change is correct and both lenses say so.** The first could not break the fix and confirmed the sibling surfaces are sound; the second attacked from migration and also did not fault the diff. What blocks this is the interaction between your fix and the resume-plan reconciler, which is nobody's fault and was invisible until someone ran two binaries against one state directory.

The finding, reproduced end-to-end with real binaries and a control:

```
1. index with the PRE-FIX binary, graph on = the DEFAULT   exit=0
2. hidden secret is live                                   hits=1
3. upgrade, SAME folder, SAME state dir, nothing on disk
   changed                                                 exit=1
4. the file the error says was removed is still on disk
5. the secret is STILL live after the failed upgrade       hits=1
```

Three things make this a blocker rather than a footnote:

1. **It is the default path.** `--no-graph` corpora reconcile a vanished file into a Delete; graph-enabled ones — the default, and what `xerj brain` uses (`brain.rs:585`) — refuse removal outright at `lib.rs:2491`. So the first post-upgrade run aborts, writes nothing, and every rerun reproduces it identically.
2. **The secrets stay live.** The documents and graph edges from the hidden path — the exact data this PR exists to keep out — remain queryable after the failed upgrade. `lib.rs:2603` says so in its own rustdoc: "nothing in this pipeline removes them."
3. **The recovery contract is unfollowable.** The JSON refusal leads with `"restore_removed_files": "put the listed file(s) back and rerun"`. The file was never removed, so an agent parsing that loops. `docs/CONTRIBUTION_REVIEW.md:33` is explicit that an error must "give an executable recovery route" and that one recommending a now-rejected flag "is not a recovery contract" — `--fresh` is refused here too.

The control matters: base→base on the identical corpus and state dir exits 0. Only the binary swap breaks it, so the attribution is unambiguous.

**This is mine to fix, not yours.** The clean remedy is the third option the lens proposes — a classifier exemption so a group that vanished *because it became hidden* is not treated as a user deletion, and is swept from the destination instead. That is engine work in `UnsupportedInventoryDelta`, not in your walker. I have filed it as its own issue and I will take it; your PR should land on top of it or alongside it, not carry it.

Everything else about this change is ready. I am not going to ask you to solve a reconciler problem you did not create.

SebTardif:
@xerj-org

> The walker fix itself is correct and the FULL workspace suite is green ... but the PR ships a silent corpus-membership change with no migration story

> This is mine to fix, not yours. ... a classifier exemption so a group that vanished because it became hidden is not treated as a user deletion ... I have filed it as its own issue

Agreed. This PR does not carry `UnsupportedInventoryDelta`. That is #439.

Pushed `89a7b9d` with the author-side pieces:

- CHANGELOG upgrade note: graph-enabled reruns of a pre-fix corpus that indexed a hidden non-UTF-8 name abort with `unsupported_content_group_removal` until #439. Do not follow "restore the removed file(s)"; the file is still on disk. Recovery that works is delete the published indices and the state directory, then re-index.
- `count_inside` and `survivors()` now use the same `is_hidden_name` predicate as the walker, so `--dry-run`'s `ignored_files_in_pruned_dirs` no longer counts a hidden non-UTF-8 file (or the contents of a hidden non-UTF-8 directory). Linux CI has the on-disk fixture (`visible1.txt` only).

`--follow-symlinks` is out of scope here; that is #438.

Issue #423, opened 2026-08-15
Term-level matching has two implementations and only one has a schema: doc_matches_query evaluates _source, so answers change at flush (consolidates 8 issues)


xerj-org:
## Root cause

Term-level queries are answered by **two different engines** depending on where the document currently lives, and only one of them has a schema.

Flushed documents are matched against the real term dictionary. Buffered (memtable) documents are matched by `doc_matches_query`, in `crates/xerj-engine/src/index.rs:33213`:

```rust
fn doc_matches_query(q: &QueryNode, source: &Value) -> bool
```

It receives **only the `_source` JSON**. No mapping, no field types, no analyser. So it cannot know whether a field is `keyword` or `text`, whether it was analysed, or what the analysed tokens were — and it substitutes a hand-rolled approximation:

```rust
// Text-field fallback: if the doc has a string with whitespace
// (i.e. looks tokenized), treat each space-separated token as a
// candidate match. ... Case folded to match ES's standard analyzer.
```

`QueryNode::Term` is worse — it does `get_field_value(source, field)` and then `json_values_equal`, a raw JSON comparison against the unanalysed source value.

## Why this produces exactly the reported symptoms

Every one of the issues below is this one defect seen from a different angle:

| symptom | why it falls out |
|---|---|
| `term` folds case but `prefix`/`wildcard` don't (#362, #396) | no schema, so each operator picks its own folding convention |
| `term` on a text field answered from `_source`, not the dictionary (#397) | `json_values_equal` on the raw source value |
| answers **change at flush** (#398, #408, #413) | memtable uses `_source`, segment uses postings — different engines, same query |
| keyword array indexed as one joined term (#408) | `get_field_value` returns the source value as-is |
| object field compared as serialised JSON (#413) | same — no type information to do otherwise |
| `case_insensitive` ignored on `fuzzy` (#406) and `term` on text (#409) | the flag has no analyser to apply |

The "changes at flush" wording in three separate issues is the tell: the same document, the same query, two answers, decided by a background flush the caller cannot see.

## Why fixing these one at a time has not worked

Each symptom is reachable from several call sites (`doc_matches_query` has 10+ callers across the memtable scan, delete-by-query, and update-by-query paths), so a patch aimed at one operator moves behaviour under the others. This is already visible in the PR history — #410's fix for #394 was found to reintroduce #394 via a sibling path, and #388's verification flagged that its fix "introduces a NEW hit-set divergence of exactly the class the issue is about."

[...]

Issue #421, opened 2026-08-15
Follow-up to #401: three unresolved adversarial-verification items were merged to main in #420


xerj-org:
Written by an AI agent. It was tasked with opening a PR for #401 from `fix/issue-401` and explicitly forbidden from merging. It could not open that PR: the branch had **already been merged** into `main` as #420 before the agent ran, while the adversarial verification of that branch had returned **`sound=false` with three MUST-FIX items**. Those three items are now live on `main` and are recorded here so they are not lost with the merged PR.

This issue tracks the three items. It does not re-litigate the fix in #420, which does resolve the original null-sort-value defect in #401.

## Status of the merge

- `fix/issue-401` (`78a1294`, on top of `bf8eca8`) was merged to `main` by #420, merge commit `f4c33ff`, `2026-08-15T22:36:05Z`.
- #420's body is the **unfilled PR template** — no motivation, no evidence, no Provenance block, every check box unticked. The verification verdict is nowhere in it.
- Attempting to open the intended PR now returns `422 No commits between main and fix/issue-401`, because the branch is fully contained in `main`.

## Item 1 — PERFORMANCE: the "take the full walk" trade is an unmeasured, undocumented latency cliff at default config

Gating `memtable_primary_key_rejects` on meta-fields does not merely decline an accelerator — it disables the **pre-clone rejection** at `engine/crates/xerj-engine/src/index.rs:14504` and `:15139`, so for a meta-sort every buffered document pays `(*source).clone()` + `seen_ids.insert` + a full `offer`.

The comment immediately above the first call site says that rejection exists precisely to kill this cost:

```
engine/crates/xerj-engine/src/index.rs:14500
    // ~1 s per query against a ~1 M-doc memtable.
```

The CHANGELOG entry for #401 (`CHANGELOG.md:119-143` on `main`) says the meta-sort paths "decline meta-fields and take the full walk" and that the third prefilter "now declines meta-fields too", and gives **no latency figure anywhere**.

Numbers below are the verifier's, **not** re-run by the agent filing this issue — quick profile, 0 segments, `request_cache=false`, median of 5, memtable-resident corpus built with `XERJ_FLUSH_DOCS` / `XERJ_FLUSH_SIZE_MB`:

| buffered docs | `sort: n` | `_seq_no` | `_version` | `_id` |
|---|---|---|---|---|
| 50k | 1.9 ms | 46.1 ms | — | 70.0 ms |
[...]

PR #426, opened 2026-08-16
release: cut 1.0.0-rc.18

xerj-org:
Cuts `1.0.0-rc.18` from `main` at `3445f7b`, which CI reports green on that exact commit.

**Eleven PRs merged since rc.17:** #387, #391, #390, #386, #393, #412, #417, #407, #420, #422, #419.

A correctness and release-hygiene RC, not a feature RC — most of it closes defects rc.17 shipped.

| | |
|---|---|
| #386 | closes the release-blocking #371 — `scalar8`/`int8_hnsw` scored updated vectors from stale codes |
| #417 | `_count` stops answering a `term` on `text` from an oracle `_search` never uses (#362) |
| #420 | `sort` on ES meta-fields resolves to real values (#401) — `_seq_no` keyset paging walks 30/30, was 4/30 |
| #419 | **@SebTardif** — console deletes/upserts fail closed instead of `204` over a refused write |
| #422 | repairs a red `main`; nothing pins the Rust toolchain, so a `stable` release failed every open PR |

## Roadmap

Re-reviewed against this cut rather than find-replaced — `docs_capability_lists` enforces that and fails the build otherwise. Six items retired (#371, #362, #360, #380, #372, #275). The rc.19 milestone is rewritten around what is actually open, including three defects found this cycle: the schema-less term-matching evaluator (#423, consolidating eight symptom issues), the multi-index scroll index-stamping bug (#414), and the meta-sort latency cliff #420 bought its own fix with (#421).

## Verification

Under CI's toolchain (rustc 1.97.1 — the local default is 1.92.0 and does not agree, see #422):

```
cargo fmt --all --check                                exit 0
cargo clippy --workspace --all-targets -- -D warnings  exit 0
cargo build --workspace --profile ci-test              exit 0
cargo test -p xerj-engine --test docs_capability_lists 15 passed, 0 failed
```

**Not run:** the 8-target Release matrix and the full ES-YAML conformance suite — both belong to CI on this commit. **The tag follows only after CI and the Release matrix report success here**, per the rc.16 procedure.

## Known-shipping defects

Cutting an RC is not a claim that the list is empty. Release-blocking items open against rc.19 are #423 (term-level matching returns different answers before and after a flush) and #414 (multi-index scroll collapses `(_index, _id)`; fix in #424, unmerged). #415 (`_source` drops null keys above ~2,000 documents) reproduces on this cut. All are in ROADMAP.md.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+## [1.0.0-rc.18] - 2026-08-16
+
 ### Documentation
 
 - **The 10,000-document scroll snapshot cap is now published, and pinned to the
--- a/ROADMAP.md
+++ b/ROADMAP.md
@@ -2,7 +2,7 @@
 
 This roadmap tracks capabilities that are **planned but not yet fully implemented**, so the project's public claims stay honest about what ships today versus what is coming. Status is verified against the actual code and by real API requests to the release binary, not aspirational.
 
-Last reviewed: 2026-08-15 (against `v1.0.0-rc.17` and `main`). Statuses trace to issues, merged PRs, the CHANGELOG, and the conformance suite; items carried forward from the 2026-07-12 review without fresh live verification are marked as such. This review line is machine-checked: `docs_capability_lists` fails the build if a release is cut without re-reviewing this file (issue #298).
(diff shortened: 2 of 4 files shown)


xerj-org:
Superseded by #435.

This branch was cut from `main` at `3445f7b`. Six more merges landed after that — #429, #432 and the rest — and the CHANGELOG's `[Unreleased]` section moved underneath it, so it went DIRTY exactly where a release commit is least worth conflict-resolving: the version bump, the fold, and the roadmap roll all have to agree with a tree that has changed.

#435 is the same procedure re-run against `9ebdac2`, with the roadmap re-reviewed against **live issue state** rather than carried over — I queried all 24 issues ROADMAP.md names before rewriting the milestone, which is the check `docs_capability_lists` exists to force and which a rebase would have quietly skipped.

Nothing is lost: the rc.18 content is identical in substance and now covers sixteen merges instead of eleven, including three outside contributions.

Issue #229, opened 2026-08-08
gRPC listener is always h2c plaintext, even with tls.enabled = true


xerj-org:
Split out of #200 before PR #221 auto-closes it (the PR fixes only the TLS fail-open downgrade; this observation is unfixed).

**gRPC is always plaintext.** `xerj-server/src/grpc.rs:1-11` builds tonic without the tls feature, so :8081 is h2c even when `tls.enabled = true` secures REST and ES-compat. Auth is enforced there, so this is not an open door, but an operator who enabled TLS reasonably expects all three listeners to be covered. At minimum document it; better, wire the feature or refuse to expose gRPC off-loopback without TLS.

Found during a source review against the user-feedback corpus. Context: #200, #221.

PR #430, opened 2026-08-16
docs: document the verified air-gapped deployment boundary

### LINKED ISSUE #376: docs: no single answer on running XERJ fully offline / air-gapped
Raised by **Uri Goren** (Argmax), whose first two questions on seeing the project were: does it have any external dependencies, and can it run on prem with no internet access.

Those are the standard first questions from anyone evaluating for a regulated or on-prem environment, and today the answer is scattered across `WHY_XERJ.md`, `SECURITY_MODEL.md` and `EXPERIMENTAL_ONNX.md` rather than stated in one place.

Wanted: one short page that answers plainly what XERJ reaches for at runtime (if anything), what the built-in embedder needs on first run vs steady state, and what a fully air-gapped install looks like end to end.

Thanks to Uri for the question.

buger:
Written by an AI coding agent (OpenAI Codex), run by @buger, who has reviewed this change.

## What this changes, and why

XERJ had no single truthful answer for an operator evaluating an air-gapped deployment: the relevant runtime behavior was scattered across several documents, while public landing pages contradicted the source by advertising a nonexistent `XERJ_AIRGAP` mode, signed bundles and manifests, cosign verification, reproducible builds, per-release SBOM/SLSA provenance, and GA status.

This adds one Markdown recipe and its landing-page counterpart, links it from the main documentation surfaces, and narrows the contradictory security, architecture, industries, public-sector, and pricing copy to behavior the repository and a published binary support.

The documented runtime boundary is explicit:

- The default `auto` embedding mode uses the built-in lexical feature-hashing embedder when no external endpoint is configured; the copyable base deployment selects `mode = "lexical"`, an empty endpoint, and no model files.
- Neural embedding is optional. Without `embedding.local_model_dir`, first use downloads `config.json`, `tokenizer.json`, and `model.safetensors` for `sentence-transformers/all-MiniLM-L6-v2`; the air-gapped path stages those three files and points XERJ at the local directory.
- `proxy` contacts only the configured OpenAI-compatible `embedding.default_endpoint`.
- `onnx-experimental` is not the standard release path; it requires an ONNX-enabled build and explicit local model and tokenizer assets.
- The running binary has no runtime telemetry, update-check, or license-activation call. Browser behavior is separate: the bundled Console contains three external Google Fonts link elements—two preconnects and one stylesheet—and the stylesheet may fetch additional font files.

The operator flow is deliberately bounded to Linux x86_64-musl: provide an approved SemVer-shaped `TAG`, download the matching release archive and adjacent `.sha256`, verify and inspect it on the connected staging host, verify it again after transfer, install it under an unprivileged `xerj` user/group, and start an authenticated loopback lexical node. Neural assets are a separate optional section rather than a prerequisite.

Authentication examples now match the implementations: MCP's `XERJ_AUTH` and `--auth` take the complete `Authorization` value such as `ApiKey <key>`, while autoindex's `XERJ_API_KEY` and `--api-key` take the raw key and add the scheme themselves.

The marketing corrections retain only the source-backed boundary: operator-staged archive plus adjacent SHA-256 integrity check, standard lexical binary, optional local neural model, authenticated loopback API, no runtime telemetry/update/license calls, and explicit browser/optional-endpoint caveats. The checksum is described as integrity evidence, not signing, provenance, or attestation.

## Evidence

The current published artifact at validation time was exercised directly, not replaced with a local build:

- Tag and asset: `v1.0.0-rc.17`, `xerj-1.0.0-rc.17-x86_64-unknown-linux-musl.tar.gz`, downloaded from the tag's GitHub release URL with its adjacent `.sha256` asset.
- Published archive SHA-256: `432546bc10c38549c210a2f234b86ce0af20d185e25095889ac3e0266080f7c2`; `sha256sum -c` returned `OK`.
- Extracted binary SHA-256: `db525d42b677717222bc532f9f18b02b5f0a67c8cf10878a4b8ab89730572583`; `xerj --version` returned `xerj v1.0.0-rc.17`.
- Fresh configuration: lexical mode, empty endpoint/model, auth enabled, loopback ES-compatible port `19376`, and a new durable data directory.
- First process: user `nobody` (`uid 65534`), PID `1348376`; authenticated health was green, index creation was acknowledged, ingest created `_id` `1`, and the lexical semantic query returned `_id` `1`.
- Restarted process: user `nobody`, distinct PID `1348721`, same binary/config/data directory; authenticated health was green and the same semantic query again returned `_id` `1`.
- Model-file probes before and after restart found zero `config.json`, `tokenizer.json`, `model.safetensors`, or ONNX files in the data directory.
- Both processes logged receipt of SIGINT and a clean stop; both `wait` calls returned `0`, each process was absent afterward, and immediate `ss` probes found port `19376` free.

## Checks

- [x] `git diff --check`
- [x] All 11 Markdown and 11 HTML shell snippets passed `bash -n`.
- [x] Both Markdown TOML blocks and both HTML TOML blocks parsed successfully.
- [x] Changed landing pages parsed as HTML and their static local links resolved.
- [x] The recipe's docs-search JSON parsed with 88 unique records and one air-gapped recipe entry.
- [x] All 77 inline landing-page scripts parsed as JavaScript.
- [x] Positive and negative TAG cases verified the Bash regex accepts `v1.0.0-rc.17` and rejects malformed/path-bearing values.
- [x] MCP and autoindex authentication forms are aligned between Markdown, HTML, help text, and client source.
- [x] The broad landing stale-claim guard found no remaining `XERJ_AIRGAP`, `AIRGAP=1`, signed-media/release/bundle/archive/manifest, cosign, reproducible-build, SBOM/SLSA, provenance-package, or “air-gap mode” claims.
- [x] Installer lint passed its POSIX shell syntax and shellcheck gates.
- [x] Documentation and landing pages were updated; no engine or wire behavior changed.
- [x] ES-YAML conformance: not applicable because this is docs/landing-only.

The landing constants guard still reports two pre-existing failures: rc.15 footer-version drift and six-versus-seven canonical agent-operation wording. The effective diff does not introduce either failure.

## Limitations and not run

Browser/firewall egress was not measured. The recipe therefore does not claim a disconnected-firewall acceptance run or zero browser requests; it tells operators to enforce egress denial and inspect firewall/log counters in their target environment.

The optional neural-model path was traced to the loader and configuration but was not exercised in the published-artifact acceptance run. XERJ does not independently pin or verify those Hugging Face files; the documented transfer digest is an operator control.

PowerShell installer lint was not run because `pwsh` was unavailable.

`cargo fmt`, Cargo builds/tests, Clippy, and the ES-YAML suite were not run because the diff changes only documentation and landing content and does not modify engine code, Cargo files, test cases, or wire behavior.

## Provenance

- **Verified:** the commands and observed results listed under Evidence and Checks, including the published artifact run, restart, persistence, process ownership, clean shutdown, snippet/TOML/HTML/JavaScript validation, link checks, stale-claim search, and final diff/ownership review.
- **Assumed:** target-environment firewall/browser behavior and successful loading of operator-staged neural files; these would be falsified by measured unexpected egress or by a neural startup/embedding failure with the documented three-file local directory.

Closes #376

### DIFF
--- a/README.md
+++ b/README.md
@@ -81,6 +81,11 @@ your PATH if needed): `xerj --insecure --data-dir ./data &`, wait until
 `http://localhost:9200` responds, then `xerj autoindex ~/my-project` — see
 [Index a folder](#index-a-folder).
 
+For a host with no runtime internet access, follow the
+[air-gapped deployment recipe](./docs/recipes/air-gapped-deployment.md). The
+default lexical embedder is offline; neural mode needs the three model files
+staged locally before the first semantic operation.
+
 ## Index a folder
 
 Start the server, then point `autoindex` at anything:
--- a/docs/recipes/README.md
+++ b/docs/recipes/README.md
@@ -21,6 +21,7 @@ These are practical "how do I actually do X" recipes, not an API reference (that
 | [Continuous anomaly datafeeds](./continuous-anomaly-datafeeds.md) | A live `_ml` datafeed that re-scores an index on a timer and stores new anomaly records you poll | [`datafeed_demo.py`](../examples/continuous-anomaly-datafeeds/datafeed_demo.py) |
 | [Migrate from Elasticsearch](./migrate-from-elasticsearch.md) | Point your existing ES client at XERJ — same wire, change the URL | [`migrate_demo.sh`](../examples/migrate-from-elasticsearch/migrate_demo.sh) |
 | [Production deployment](./production-deployment.md) | The hardened posture: in-process TLS + API-key auth, health-probe & gRPC posture, air-gapped neural-model staging, and copying data out of a live Elasticsearch (scroll + bulk) | inline `curl` + `bash` |
+| [Air-gapped deployment](./air-gapped-deployment.md) | Stage a release and optional neural model on a connected machine, then run XERJ offline with explicit network and persistence boundaries | inline `curl` + `bash` |
 | [Diagnose CPU and memory with pprof](./debug-profiling.md) | A private, bounded, symbolized CPU/heap capture with reproducibility manifest, artifact validation, and AI analysis prompts | [`debug-profiling/`](../../demo/playbooks/debug-profiling/) |
 
 Every example is stdlib-only Python 3, plain `curl`, or the `xerj` binary itself — no dependencies to install. Start XERJ (`./target/release/xerj --data-dir ./data --insecure`), then run any example.
--- a/docs/recipes/air-gapped-deployment.md
+++ b/docs/recipes/air-gapped-deployment.md
@@ -1,0 +1,291 @@
+# Air-gapped deployment: run XERJ without runtime internet
+
+This is a short, Linux x86_64-musl procedure. It stages one operator-approved
+release, installs the service as an unprivileged account, and keeps the node on
+loopback. The base path is lexical and needs no model files. Neural embedding is
+an optional, separately staged path.
+
+## Runtime boundary
+
+- The default embedding mode is **lexical feature hashing**. It needs no model,
+  no embedding service, and no network connection.
+- Neural mode is optional. Without `embedding.local_model_dir`, the first
+  operation that needs an embedding downloads `config.json`, `tokenizer.json`,
+  and `model.safetensors` for `sentence-transformers/all-MiniLM-L6-v2` from the
+  Hugging Face Hub. With that directory set, XERJ reads those files locally.
+- `proxy` calls only the configured OpenAI-compatible
+  `embedding.default_endpoint`. The experimental `onnx-experimental` backend
+  needs an ONNX-enabled build and explicit local assets; neither is the base
+  release path.
+
+The running node makes no telemetry, update-check, or license-activation call.
+Release-download analytics are a property of the connected website/GitHub
+staging surface, not of a running XERJ process. The installer also downloads a
+release archive and its checksum from GitHub; an offline operator must stage
+those files before entering the enclave.
+
+## Defaults and boundaries
+
+| Component | Default | Air-gapped meaning |
+|---|---|---|
+| Embedding | `auto`: lexical unless an endpoint is configured | No model or network for lexical mode |
+| Neural model | Off | Set `local_model_dir` before selecting `neural` |
+| External proxy | Disabled (`default_endpoint = ""`) | Only the configured endpoint is contacted |
+| ONNX | Experimental and not in the standard release | Build the feature and provide local assets explicitly |
+| WAL tap | Disabled | It remains inert unless enabled with a target URL; see the durable overlay note below |
+| Cluster | Disabled | Single-node startup does not initialize the Raft transport |
+| REST / ES-compatible listeners | `127.0.0.1` | The default is loopback-only |
+| `autoindex` and MCP clients | Localhost defaults | They connect to `localhost`; they do not create a public listener |
+| Runtime telemetry/update/license activation | None | No outbound call is made by the running binary for these purposes |
+
+This is a statement of defaults, not a claim that every browser UI request is
+offline. **The bundled Console has three external Google Fonts link elements
+(two preconnects and one stylesheet; stylesheet may fetch additional font files).**
+Blocked requests fall back to system fonts. The engine and APIs
+continue to work.
+
+This procedure does not claim that egress was measured here. If your policy
+requires that boundary, apply an egress-deny rule and inspect firewall/log
+counters during the acceptance checks.
+
+## 1. Stage an approved release on a connected machine
+
+Do this on a machine that is allowed to reach GitHub. Set `TAG` to the
+operator-approved `vX.Y.Z` release tag before running; there is no moving
+release alias in this procedure. The release workflow names each archive from
+the version without the `v` prefix, target triple, and extension. The matching
+`.sha256` is a separate per-archive asset.
+
+The following stages the Linux x86_64 musl archive only.
+
+```bash
+set -eu
+
+: "${TAG:?set TAG to an operator-approved vX.Y.Z release tag}"
+if [[ ! "$TAG" =~ ^v[0-9]+\.[0-9]+\.[0-9]+(-[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?(\+[0-9A-Za-z-]+(\.[0-9A-Za-z-]+)*)?$ ]]; then
+  echo "TAG must match vX.Y.Z with optional prerelease/build metadata: $TAG" >&2
+  exit 2
+fi
+VERSION="${TAG#v}"
+TARGET="x86_64-unknown-linux-musl"
+EXT="tar.gz"
+STAGE="xerj-${VERSION}-${TARGET}"
+ASSET="${STAGE}.${EXT}"
+BASE="https://github.com/xerj-org/xerj/releases/download/${TAG}"
+OUT="${OUT:-$PWD/xerj-airgap-${VERSION}-${TARGET}}"
+
+mkdir -p "$OUT/release"
+curl -fL --retry 3 -o "$OUT/release/$ASSET" "$BASE/$ASSET"
+curl -fL --retry 3 -o "$OUT/release/$ASSET.sha256" "$BASE/$ASSET.sha256"
+
+(
+  cd "$OUT/release"
+  sha256sum -c "$ASSET.sha256"
+  tar -xzf "$ASSET"
+  test -x "$STAGE/xerj"
+  "$STAGE/xerj" --version
+)
+```
+
+The checksum proves that the archive matches the digest published beside that
+archive in the same GitHub release. It is an integrity check, not an
+independent signature, provenance statement, or attestation. If your policy
+requires signed releases or an external attestation, obtain and verify that
+evidence separately; `.sha256` alone does not provide it.
+
+After transfer, set the same approved `TAG` in the enclave and verify before
+extracting:
+
+```bash
+: "${TAG:?set TAG to the same operator-approved release tag}"
+VERSION="${TAG#v}"
+ASSET="xerj-${VERSION}-x86_64-unknown-linux-musl.tar.gz"
+cd /opt/xerj-staging/release
+sha256sum -c "$ASSET.sha256"
+tar -xzf "$ASSET"
+test -x "xerj-${VERSION}-x86_64-unknown-linux-musl/xerj"
+```
+
+## 2. Install the base lexical node
+
+The account and directories below are an example of the required ownership;
+use the equivalent account-management command on the target Linux image. No
+model directory is created for the lexical deployment.
+
+```bash
+if ! getent passwd xerj >/dev/null; then
+  sudo useradd --system --user-group --home-dir /var/lib/xerj --shell /sbin/nologin xerj
+fi
+sudo install -d -m 0755 /opt/xerj/bin /etc/xerj
+sudo install -d -o xerj -g xerj -m 0750 /var/lib/xerj
+sudo install -m 0755 \
+  "/opt/xerj-staging/release/xerj-${VERSION}-x86_64-unknown-linux-musl/xerj" \
+  /opt/xerj/bin/xerj
+```
+
+Create `/etc/xerj/xerj.toml` with the complete base configuration. Explicitly
+selecting lexical mode and an empty endpoint prevents an accidental proxy or
+neural dependency:
+
+```toml
+[server]
+bind_address = "127.0.0.1"
+data_dir = "/var/lib/xerj"
+es_compat_port = 9200
+
+[auth]
+enabled = true
+
+[embedding]
+mode = "lexical"
+default_endpoint = ""
+default_model = ""
+```
+
+Start it as the service user. On first start, the authenticated admin key is
+written to `/var/lib/xerj/admin.key` with restrictive permissions.
+
+```bash
+sudo -u xerj /opt/xerj/bin/xerj --config /etc/xerj/xerj.toml
+```
+
+Keep the listener on loopback. A network-facing listener requires the separate
+production TLS/auth procedure; this recipe does not teach a cleartext opt-out.
+
+## 3. Optional neural model
+
+Do this only when the enclave needs neural semantics. On the connected staging
+machine, download the three files consumed by the built-in loader and checksum
+them for the transfer:
+
+```bash
+MODEL="$OUT/model/all-MiniLM-L6-v2"
+mkdir -p "$MODEL"
+for FILE in config.json tokenizer.json model.safetensors; do
+  curl -fL --retry 3 -o "$MODEL/$FILE" \
+    "https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/resolve/main/$FILE"
+done
+(
+  cd "$MODEL"
+  sha256sum config.json tokenizer.json model.safetensors > ../model.sha256
+)
+```
+
+Transfer `OUT/model` with the release and verify `model.sha256` inside the
+enclave. Then install the files and change only the embedding block:
+
+```bash
+sudo install -d -o xerj -g xerj -m 0750 /opt/xerj/models/all-MiniLM-L6-v2
+sudo install -o xerj -g xerj -m 0640 \
+  /opt/xerj-staging/model/all-MiniLM-L6-v2/{config.json,tokenizer.json,model.safetensors} \
+  /opt/xerj/models/all-MiniLM-L6-v2/
+```
+
+```toml
+[embedding]
+mode = "neural"
+local_model_dir = "/opt/xerj/models/all-MiniLM-L6-v2"
+```
+
+XERJ does not independently pin or verify model checksums; the transferred
+digest is an operator supply-chain check, not a XERJ signature. If you only
+need lexical search, omit this section and all model assets.
+
+## 4. Authenticate local clients and verify
+
+These checks are intentionally operator-run checks, not a claim that this
+documentation was live-tested against a disconnected firewall. They verify the
+actual binary and local files in your enclave.
+
+```bash
+XERJ=/opt/xerj/bin/xerj
+DATA=/var/lib/xerj
+KEY="$(sudo cat "$DATA/admin.key")"
+export XERJ_API_KEY="$KEY"
+
+"$XERJ" --version
+curl -fsS -H "Authorization: ApiKey $KEY" \
+  http://127.0.0.1:9200/_cluster/health
+```
+
+With the lexical node running, index a small semantic document and query it
+through the authenticated ES-compatible surface. This semantic query uses the
+local lexical embedder and needs zero model files; assert that it returns
+`_id` `1`.
+
+```bash
+curl -fsS -X PUT "http://127.0.0.1:9200/offline-demo" \
+  -H "Authorization: ApiKey $KEY" -H 'Content-Type: application/json' \
+  -d '{"mappings":{"properties":{"body":{"type":"semantic_text"}}}}'
+curl -fsS -X POST "http://127.0.0.1:9200/offline-demo/_doc/1?refresh=true" \
+  -H "Authorization: ApiKey $KEY" -H 'Content-Type: application/json' \
+  -d '{"body":"A local lexical node can answer this without a network service."}'
+curl -fsS -X POST "http://127.0.0.1:9200/offline-demo/_search" \
+  -H "Authorization: ApiKey $KEY" -H 'Content-Type: application/json' \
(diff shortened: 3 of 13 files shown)


xerj-org:
## Independent verification — omissions lens

**refuted: `False` · confidence: `high` · must-fix: 3**

The core thesis was independently confirmed. The findings below are omissions, not errors in what you wrote.

> The PR's core thesis holds and I independently confirmed it: `gh release view v1.0.0-rc.17` publishes only 8 archives + 8 adjacent `.sha256` files — no SBOM, no cosign signature, no SLSA attestation — so deleting the `XERJ_AIRGAP=1` / signed-bundle / cosign / reproducible-build / SBOM-SLSA copy from landing was correct, and `grep -rn -i "XERJ_AIRGAP|cosign|SLSA|reproducible build|SBOM" landing/ docs/ README.md` now returns nothing but the recipe's own disclaimer. I also reproduced both digests in the PR body byte-for-byte (archive `432546bc…f7c2`, binary `db525d42…2583`) and `sha256sum -c` returned `OK`.
> 
> Judged on omissions, though, three gaps are serious. (1) The in-enclave verification snippet — the single most security-critical step in an air-gap procedure — fails open: it has no `set -eu` and no `&&` chaining, so a FAILED checksum is followed by `tar -xzf` anyway; I demonstrated it extracting a corrupted archive with a final exit status of 0. (2) The doc's central browser-egress claim, repeated in five places, undercounts by 3×: the shipped rc.17 binary bundles three Console HTML pages (index/login/setup), each carrying three Google Fonts link elements — 9, not 3 — and login/setup are the first pages an operator sees. (3) Step 3 tells operators to flip `mode = "lexical"` to `mode = "neural"` on the data directory step 4 just populated, but `validate_embedding_identity` guards only `onnx-experimental` and both backends are 384-dim, so the switch silently mixes vector spaces with no error and no reindex warning.
> 
> Beyond that the page also omits `xerj brain` (spawns a detached server and opens a browser at the Console), the gRPC h2c listener, `scripts/verify-release.sh`, any systemd unit, and it never propagated its own docs-search entry to the other 14 docs pages.


---

# MUST-FIX (3)

### In-enclave checksum verification fails open — a FAILED digest still extracts and installs the archive

The step-1 connected-machine snippet is properly guarded (`set -eu` plus a subshell), but the post-transfer snippet — the one that actually protects the enclave — drops both `set -eu` and the TAG regex validation, and chains commands with newlines instead of `&&`. Pasted into an operator's interactive shell (the documented usage), a checksum mismatch prints FAILED and execution continues straight into `tar -xzf "$ASSET"`. The trailing `test -x` then succeeds against the extracted tree, so a wrapping script sees exit 0. Step 2 immediately does `sudo install -m 0755 ".../xerj" /opt/xerj/bin/xerj`, installing the unverified binary. Identical text appears in both docs/recipes/air-gapped-deployment.md (lines 133-141) and landing/docs/recipes/air-gapped-deployment.html (lines 492-498). The PR's check `bash -n` only parses syntax and cannot catch this. Fix: add `set -eu` to the second block, or chain `sha256sum -c "$ASSET.sha256" && tar -xzf "$ASSET"`.

```
Reproduced with the real published archive plus a deliberately wrong .sha256, running the recipe's block verbatim:

  $ bash -c 'TAG=v1.0.0-rc.17; : "${TAG:?...}"; VERSION="${TAG#v}"; ASSET="xerj-${VERSION}-x86_64-unknown-linux-musl.tar.gz"; cd .../demo; sha256sum -c "$ASSET.sha256"; tar -xzf "$ASSET"; test -x "xerj-${VERSION}-x86_64-unknown-linux-musl/xerj"; echo "EXIT_OF_LAST=$?"; ls -d xerj-1.0.0-rc.17-x86_64-unknown-linux-musl'
  xerj-1.0.0-rc.17-x86_64-unknown-linux-musl.tar.gz: FAILED
  sha256sum: WARNING: 1 computed checksum did NOT match
  EXIT_OF_LAST=0
  xerj-1.0.0-rc.17-x86_64-unknown-linux-musl
  ARCHIVE WAS EXTRACTED DESPITE FAILED CHECKSUM

Contrast docs/recipes/air-gapped-deployment.md:96 (`set -eu` present in the staging block) with :134 (absent in the enclave block).
```

### Console external-asset count is wrong by 3x — nine Google Fonts link elements ship, not three

The PR body, docs/recipes/air-gapped-deployment.md (lines 76-78, 309-311), landing/docs/recipes/air-gapped-deployment.html (line 461), landing/llms.txt, landing/llms-full.txt, landing/security/index.html and landing/architecture/index.html all state that "the bundled Console has three external Google Fonts link elements (two preconnects and one stylesheet)". That is the count for one page. engine/crates/xerj-console-api/build.rs walks the whole xerj-ux tree and emits `include_bytes!` entries for every .html it finds, and spa.rs serves `/_xerj-console/{path*}` with an extensionless fallback to `<name>.html`, giving `/login` and `/setup` nice URLs. xerj-ux ships three HTML pages — index.html, login.html, setup.html — each with the same three link elements. Total: 9. This matters more than arithmetic: login.html and setup.html are the *first* pages an operator hits on a fresh enclave node (the first-launch bootstrap prints a one-time setup link), so the page an accreditor loads first is one the doc does not account for. Since this is the one residual egress the whole document is built around, the number needs to be right in all six places.

```
Against the published binary, not the source tree:

  $ sha256sum xerj-1.0.0-rc.17-x86_64-unknown-linux-musl/xerj
  db525d42b677717222bc532f9f18b02b5f0a67c8cf10878a4b8ab89730572583   (matches the PR body)
  $ grep -a -c '<link rel="preconnect" href="https://fonts.googleapis.com">' xerj
  3
  $ grep -a -c 'fonts.gstatic.com' xerj
  3
  $ grep -a -o 'fonts.googleapis.com/css2?[^"]*' xerj | sort | uniq -c
        1 ...family=Big+Shoulders+Display:wght@400;700;800;900&family=JetBrains+Mono...   (index.html)
        2 ...family=Big+Shoulders+Display:wght@800;900&family=JetBrains+Mono...          (login.html, setup.html)

  $ grep -c "fonts.googleapis\|fonts.gstatic" xerj-ux/index.html xerj-ux/login.html xerj-ux/setup.html
  xerj-ux/index.html:3   xerj-ux/login.html:3   xerj-ux/setup.html:3
```

### Step 3's lexical-to-neural switch silently corrupts an existing index — no guard in code, no warning in the doc

Section 4 has the operator build and query a lexical index in /var/lib/xerj. Section 3 then says "change only the embedding block" / "Replace only the embedding block" to `mode = "neural"`. Nothing stops this and nothing warns about it. engine/crates/xerj-engine/src/index.rs:27623 `validate_embedding_identity` delegates to `configured_onnx_identity`, which returns `Ok(None)` for any mode that is not `onnx-experimental` (index.rs:27458); with `configured == None` and no marker file the function short-circuits on `let Some(identity) = configured else { return Ok(()) };` and returns Ok. The dimension check cannot catch it either, because the lexical embedder was deliberately sized to match MiniLM. The result is neural query vectors scored against lexical document vectors in the same 384-dim space: no startup error, no query error, silently wrong rankings — the exact failure the ONNX identity marker was built to prevent, on the path this recipe actively recommends. The pre-existing docs/recipes/production-deployment.md does not warn about it either. The recipe needs an explicit "switching backends requires a fresh index/data directory and a full reindex" note in section 3.

```
engine/crates/xerj-ai/src/local.rs:26-28 —
  /// 384 mirrors the popular `all-MiniLM-L6-v2` output size, so switching from
  /// the built-in embedder to a real proxy of that width needs no mapping change.
  pub const DEFAULT_DIMS: usize = 384;

engine/crates/xerj-engine/src/index.rs:27457-27459 —
  fn configured_onnx_identity(cfg: &EmbeddingConfig) -> Result<Option<PersistedEmbeddingIdentity>> {
      if !cfg.mode.trim().eq_ignore_ascii_case("onnx-experimental") {
          return Ok(None);

index.rs:27678-27681 — the only exit left once `configured` is None and no marker exists:
  let Some(identity) = configured else {
      return Ok(());
  };
```


---

# SHOULD-FIX (4)

### "unless your deployment removes or rewrites that HTML" is not an actionable mitigation

Both the .md (line 309-311) and .html (line 582) offer removing or rewriting the Console HTML as the operator's remedy for the font egress. There is no way to do that on a staged release. engine/crates/xerj-console-api/build.rs compiles the xerj-ux tree into `$OUT_DIR/xerj_console_assets.rs` as a slice of `include_bytes!()` references linked into the static binary (spa.rs:21-25 documents this explicitly), so the HTML is inside the 126 MB stripped ELF the operator just checksummed. There is no `[console] enabled` key anywhere in xerj-common/src/config.rs, and the route is unconditionally registered — the startup banner even advertises `http://localhost:{port}/_xerj-console/ ({n} files bundled)` (xerj-server/src/main.rs:397). The only real remedies are an egress-deny rule (which the doc does mention) or rebuilding from source with a patched xerj-ux — and that rebuild needs crates.io, i.e. it is itself a connected-machine step the recipe never describes. Either drop the suggestion or spell out the rebuild.

```
engine/crates/xerj-console-api/src/spa.rs:21-25 —
  //! ## Why not include_dir / rust-embed
  //! Both pull in proc-macro deps. The build.rs approach is ~80 LOC,
  //! depends only on std, and produces a slice of `include_bytes!()`
  //! references that the linker resolves directly.

  $ grep -rn "console" engine/crates/xerj-common/src/config.rs
  2952:    /// first-run console link. Making loopback the default (#228) makes `::1`
  (no console enable/disable key exists)
```

### The recipe never starts the server in the background, so section 4 cannot run as written

Section 2 ends with `sudo -u xerj /opt/xerj/bin/xerj --config /etc/xerj/xerj.toml`, which runs in the foreground and blocks the terminal. Section 4 then opens with `KEY="$(sudo cat "$DATA/admin.key")"` and a series of curls against the running node, and the "Existing data directories" section adds two more. An operator following the page top to bottom in one shell is stuck at the end of section 2. The recipe also provides no systemd unit and no link to one, despite carrying the kicker "RECIPE · 13 · OPERATIONS" and despite landing/docs/install.html already having a "Run as a systemd unit" section (line 172) that would give the enclave node restart-on-boot, the unprivileged-user enforcement the recipe asks for, and the hardening directives docs/operations.html advertises. For a document whose whole purpose is an accreditable install procedure, shipping no service supervision is a conspicuous hole.

```
docs/recipes/air-gapped-deployment.md:183 —
  sudo -u xerj /opt/xerj/bin/xerj --config /etc/xerj/xerj.toml
(no `&`, no nohup, no systemd; next executable content is line 237 `KEY="$(sudo cat "$DATA/admin.key")"` followed by curl against 127.0.0.1:9200)

  $ grep -n "<h2>" landing/docs/install.html
  172:  <h2>Run as a systemd unit</h2>     <- exists, never referenced by the new recipe
```

### The new page is invisible to the docs search on every other docs page

Each landing docs page embeds its own private copy of the `docs-index` JSON that powers the ⌘K search. The PR added the air-gapped record only to the new page's copy, so the page can find itself and nothing else can find it. An operator who lands on /docs/index.html, /docs/operations.html or /docs/install.html and searches "air-gapped" gets "NO MATCHES". The PR's check — "The recipe's docs-search JSON parsed with 88 unique records and one air-gapped recipe entry" — validated exactly the one file that did not need validating. Relatedly, the new page's sidebar carries 10 docs links against 44 on a sibling recipe page: the entire REFERENCE, DATA MODEL, ENGINE and PLAYBOOKS groups and the other twelve recipes are gone, so the page is a navigational dead end.

```
Parsing the embedded JSON out of each page:
  landing/docs/index.html                          records= 85 airgap_entries=0
  landing/docs/operations.html                     records= 85 airgap_entries=0
  landing/docs/recipes/index.html                  records= 87 airgap_entries=0
  landing/docs/recipes/zero-config-autoindex.html  records= 87 airgap_entries=0
  landing/docs/recipes/air-gapped-deployment.html  records= 88 airgap_entries=1

  $ grep -c '<a href="/docs/' landing/docs/recipes/zero-config-autoindex.html
  44
  $ grep -c '<a href="/docs/' landing/docs/recipes/air-gapped-deployment.html
  10
```

### The repo's own published-artifact verifier, scripts/verify-release.sh, is never mentioned

The recipe hand-rolls a single-target checksum check for the connected staging machine. The repo already ships scripts/verify-release.sh, which is precisely the connected-machine tool this step wants: it asserts every declared target is present (so a release that lost a matrix leg fails rather than quietly verifying the rest), that every archive has and matches a .sha256, that every archive contains binary + LICENSE + README, that every binary's startup banner carries the tag's version and no other, and it boots the host-native binary on a clean data dir and runs a real search. Its header documents the concrete regression it exists to catch — rc.10 shipped binaries printing "xerj v1.0.0-rc.9" because asset filenames come from the git tag and the banner from CARGO_PKG_VERSION. An operator approving a TAG for an enclave is exactly who should run it, and the recipe's weaker subset silently drops the version-drift check.

```
scripts/verify-release.sh:3-17 —
  # verify-release.sh — post-release check on the PUBLISHED artifacts.
  # ... It exists because v1.0.0-rc.10 shipped binaries that print "xerj v1.0.0-rc.9".
  #   scripts/verify-release.sh v1.0.0-rc.10   ->  FAIL (version drift)

  $ grep -rn "verify-release" docs/recipes/air-gapped-deployment.md landing/docs/recipes/air-gapped-deployment.html
  (no matches)
```


---

# NOTE (5)

### `xerj brain` is omitted from a local-client inventory phrased as complete

The "Existing data directories and local clients" section enumerates `xerj mcp` and `xerj autoindex` and closes with "Neither creates an external listener or silently uploads corpus data" — a completeness claim over a set that is missing a third shipped subcommand. `xerj brain` is dispatched at xerj-server/src/main.rs:1877 and does more than either of the two covered clients: it spawns the server detached in its own process group with output to `<data-dir>/server.log`, relays the first-launch one-time setup link, and then opens a browser at the Console — i.e. it triggers exactly the font requests the rest of the page is devoted to. `--no-open` suppresses the browser but appears nowhere in the recipe. Adding brain to the inventory with a `--no-open` note would close this.

```
engine/crates/xerj-server/src/brain.rs:1-4, 92, 415-423 —
  //! `xerj brain <folder>` — ... boot (or attach to) the local server, run the autoindex
  //! pipeline ..., then open the console's second-brain view
  --no-open          print the links but do not open a browser
  if cfg.no_open { ... } else { match open_browser(target) { Ok(()) => println!("  opening your browser…"), ...

  $ grep -c "brain" docs/recipes/air-gapped-deployment.md
  0
```

### The listener inventory omits the gRPC h2c listener

The "Defaults and boundaries" table lists "REST / ES-compatible listeners | 127.0.0.1" and is presented as the deployment's listener boundary. There is a third listener: `server.grpc_port`, which xerj-common/src/config.rs:697-701 documents as speaking cleartext HTTP/2 regardless of the TLS setting because xerj-server/src/grpc.rs builds tonic without its `tls` feature (issue #229). It binds on the same loopback `bind_address` so the security posture is unchanged, but a table an accreditor reads as the complete listener enumeration should enumerate it — especially since the same code comment notes that enabling TLS while binding off-loopback refuses to start unless `allow_insecure_grpc_h2c` is set, a subtlety an operator promoting this node off loopback will hit.

```
engine/crates/xerj-common/src/config.rs:697-701 —
  /// It does **not** cover the gRPC listener: `xerj-server/src/grpc.rs`
  /// builds tonic without its `tls` feature, so `server.grpc_port` speaks
  /// cleartext HTTP/2 (h2c) whatever this is set to (issue #229).

  $ grep -c "grpc" docs/recipes/air-gapped-deployment.md
  0
```

### Two neural-staging alternatives documented elsewhere are dropped: model_cache_dir and the slim-build fallback

First: `embedding.model_cache_dir` exists (config.rs:1275) and lets an operator pre-seed the HuggingFace cache rather than use `local_model_dir`; the pre-existing production-deployment.md:508 mentions it, the new recipe does not. This also hides a footgun — with `mode = "neural"` and neither key set, hf-hub falls back to `~/.cache/huggingface`, which for the recipe's service account resolves under /var/lib/xerj (the recipe sets `--home-dir /var/lib/xerj`), so a mis-staged enclave node writes model state into the data directory it just told the operator to keep clean and 0750. Second: production-deployment.md warns that a `--no-default-features` build omits the neural backend and `mode = "neural"` on such a binary logs a warning and silently falls back to lexical. The new recipe repeats neither, while section 3 is precisely where a staged non-standard build would bite.

```
engine/crates/xerj-common/src/config.rs:1273-1275 —
  /// Neural backend: directory to cache downloaded model weights. Empty
  /// (default) uses the standard HuggingFace cache (`~/.cache/huggingface`).
  pub model_cache_dir: String,

docs/recipes/production-deployment.md:534-537 —
  > **Slim builds.** A `--no-default-features` build omits the neural backend;
  > `mode = "neural"` on such a binary logs a warning and falls back to the
  > lexical embedder.

  $ grep -c "model_cache_dir\|no-default-features" docs/recipes/air-gapped-deployment.md
  0
```

### public-sector.html was narrowed but still claims OTLP export, which does not exist

The PR body says it narrows "the contradictory security, architecture, industries, public-sector, and pricing copy to behavior the repository and a published binary support". On the public-sector page it corrected the diagram and the hero but left line 171: "Prometheus · OTLP export · SIEM-integrated · air-gap-compatible". The repo has OTLP *ingest* only — `POST /v1/indices/:name/otlp` (router.rs:94, native.rs:1394) parses an `ExportLogsServiceRequest` body sent to XERJ. There is no exporter anywhere in engine/crates. An egress-shaped claim on the flagship air-gap page is the wrong one to leave standing given this PR's own premise. The same page also retains "one-minor-version rollback window" in the offline-update box, which the PR did not trace to anything.

```
landing/industries/public-sector.html:171 —
  <text x="76" y="328" class="small">Prometheus · OTLP export · SIEM-integrated · air-gap-compatible</text>

  $ grep -rn -i "otlp" --include=*.rs engine/crates/ | grep -v tests
  engine/crates/xerj-api/src/router.rs:94:  .route("/v1/indices/:name/otlp", post(native::ingest_otlp))
  engine/crates/xerj-api/src/native.rs:1371: /// Ingest OpenTelemetry logs in OTLP JSON format.
  (ingest only — no exporter)
```

### Only one of eight published targets is covered, and the container path is a dead end

The recipe is honest that it "stages the Linux x86_64 musl archive only", but rc.17 ships 8 archives including aarch64-unknown-linux-musl and -gnu, and the audience named on the landing pages (SCIFs, sovereign cloud, sealed racks) increasingly runs aarch64. There is also no container or Kubernetes guidance at all, which is how most regulated enclaves actually deploy: deploy/helm/xerj/values.yaml sets `image.repository: xerj/xerj`, but landing/docs/install.html:207 states "No published image yet — build it locally from engine/Dockerfile", so an air-gapped Helm user must build the image on a connected machine (needing crates.io) and push it to an internal registry — a whole staging flow nobody has written down. Worth at least a scoping sentence naming what is out of scope and why.

```
$ gh release view v1.0.0-rc.17 --repo xerj-org/xerj --json assets --jq '.assets[].name' | grep -c 'tar.gz$\|zip$'
  8

deploy/helm/xerj/values.yaml:4-5 —
  image:
    repository: xerj/xerj

landing/docs/install.html:207 —
  <p>No published image yet — build it locally from <code class="inline">engine/Dockerfile</code> ...</p>
```


---

## Maintainer note

@buger — thank you for this, and please read the above as what it is: a page whose **central claim was independently reproduced** (`gh release view v1.0.0-rc.17` publishes 8 archives + 8 `.sha256` and nothing else — no SBOM, no cosign, no SLSA — so deleting that copy from `landing/` was correct), reviewed hard because an air-gap procedure is the wrong place to be approximately right.

The one I would fix before anything else is the **fail-open verification**. The connected-machine block is properly guarded; the post-transfer block — the one that actually protects the enclave — is not, and an operator pasting it gets `FAILED` followed by `tar -xzf` and then `sudo install`. `bash -n` cannot see this, which is why the PR's own check passed. One line fixes it:

```sh
sha256sum -c "$ASSET.sha256" && tar -xzf "$ASSET"
```

or `set -eu` at the top of the second block, matching the first.

The **9-vs-3 font count** matters for the same reason: `login.html` and `setup.html` are the first pages an operator hits on a fresh node, and they carry the same three link elements as `index.html`. Since the residual egress is the thing the whole document exists to bound, that number is load-bearing for an accreditor.

The **lexical→neural switch** is not your bug — `validate_embedding_identity` genuinely does not guard it, and the dimension check cannot, because the lexical embedder is deliberately 384-wide to match MiniLM. But section 4 has the operator build a lexical index and section 3 then tells them to change the embedding block, so the recipe walks them into it. I am filing that as an engine issue separately; the recipe needs a warning either way.

Happy to push the checksum fix and the font count to your branch myself if you would rather — say the word. Otherwise this needs another round from you, and it is worth it: this is a genuinely useful page.

PR #427, opened 2026-08-16
fix(autoindex): qualify GIF magic by reading past 8 KiB, not by guessing (#379, #380)

xerj-org:
Round 2 for #379 / #380. Supersedes #411, whose adversarial verification returned `sound=false` with two live defects **of the class the fix claims to close**. Both reproduced through the real `sniff_bytes`.

## What round 1 missed

**The `0x2C` Image-Descriptor path.** For all-ASCII input `packed & 0x80` is always clear, so the first block always lands at offset 13. A `,` there, plus four little-endian u16s whose HIGH bytes (offsets 15/17/19/21) are spaces, satisfies every geometry bound. Verified junked as `binary/Some("gif")` on #411's head:

```
"GIF89a format,1 2 3 4 5 and the rest of this note is about the format. " x30
"GIF89a header,a b c d e f g h ..."          "GIF89a fields,x y z w ..."
GIF87a variant, and the CSV "GIF89a format,1 2 3 4 5\nrow,a,b,c,d\n..."
```

**The `0x21` Extension path.** The label alone is not a discriminator, because `decode()` falls back to windows-1252 where `0xF9`/`0xFE`/`0xFF` are ordinary latin-1 letters. `"GIF89a fichie!" + 0xF9 + French prose` → `binary/Some("gif")`.

## The fix

**`0x2C`** — decode one field further, to the LZW minimum code size that opens the image data. The spec confines it to `2..=12`, entirely non-printable.

**That alone is still insufficient, and the verification did not catch this case either.** The CSV counterexample carries `\n` (10) at exactly that offset, which is *inside* the valid range. So the frame must also have a palette — its own Local Colour Table or the Global one. A file with neither has nothing to decode pixels against and is not a GIF; the CSV has both flags clear.

**`0x21`** — the spec fixes the first sub-block size for three of the four labels: 12 after `0x01`, 4 after `0xF9`, 11 after `0xFF`, all non-printable. Comment sizes are genuinely variable, so that arm walks the sub-block chain to its `0x00` terminator and requires a valid introducer behind it.

## Fixtures corrected, not loosened

Four fixtures stopped one byte short of what the spec mandates — the byte after `0x21 0xF9` was `'l'` from the lorem payload, where a real Graphic Control Extension always carries `4`. They asserted they were real GIFs while not being decodable. The `0x2C` fixture additionally gains the Global Colour Table that a frame without a local one requires.

[...]

PR #431, opened 2026-08-16
DO NOT MERGE (refuted — snapshot not atomic) — fix(scroll): continuation pages resolve real _seq_no/_version (#428)

### LINKED ISSUE #428: scroll continuation pages (_search/scroll) never emit _seq_no/_version, even with seq_no_primary_term:true
## What happened

`seq_no_primary_term: true` on the scroll-opening request (`POST /{index}/_search?scroll=…`) is honored for the first page — each hit carries a real `_seq_no`/`_primary_term`, resolved from the version map. Every page after the first, fetched via `POST /_search/scroll`, comes back with **no `_seq_no` field at all** — not `0`, just absent, as if the flag had never been set.

Live repro (235-document index, `size: 50` → 5 pages):

```
$ curl -s -XPOST localhost:9200/idx/_search?scroll=1m -d '{"size":50,"seq_no_primary_term":true,"query":{"match_all":{}}}'
# page 1 hits: {"_id":"doc-0000", "_seq_no":0, "_primary_term":1, ...}   <- correct

$ curl -s -XPOST localhost:9200/_search/scroll -d '{"scroll":"1m","scroll_id":"<id>"}'
# page 2 hits: {"_id":"doc-0050", "_score":1.0, "_source":{...}}         <- no _seq_no/_primary_term at all
```

Confirmed this is xerj-specific, not a scroll-protocol limitation: the identical test against a real **OpenSearch 3.7.0** node (4 pages, `size: 30`) carries `_seq_no`/`_primary_term` on every single page, as documented ES/OpenSearch behavior — the flag is meant to persist for the scroll context's whole lifetime, not be re-specified per continuation call (and re-specifying it on the continuation body doesn't help on xerj either — tried `{"scroll":"1m","scroll_id":"...","seq_no_primary_term":true}`, same result, no `_seq_no`).

## Where it is

The first page goes through the main `search()` handler (`/:index/_search`, `es_compat.rs`), which resolves the real value via `idx.lookup_seq_no(&h.id)` when `seq_no_primary_term` is set (`emit_seq_no`, same code path as an ordinary non-scroll search).

Every later page goes through `scroll_page_response` (`es_compat.rs:19974`), called by both `next_scroll` (`:19922`, the body/query-param continuation form) and `next_scroll_path` (the legacy path-parameter form). Two separate problems compound here:

1. The intermediate `EsHit` it builds hardcodes the values instead of resolving them:
   ```rust
   // es_compat.rs:20028-20029
   version: Some(1),
   seq_no: Some(0),
   ```
   (Same hardcoding appears in `search_with_scroll`'s two `EsHit` constructions at `:19763-19764` and `:19853-19854` — a different, `/:index/_search_scroll`-routed handler that doesn't appear to be reachable from a plain `?scroll=` request, so probably not hit by this repro, but worth knowing it's the same pattern in more than one place.)

2. Even those hardcoded placeholder values never reach the wire: the JSON-building block right below (`es_compat.rs:~20066-20097`) inserts `_index`, `_id`, `_score`, `_source`, `sort`, `fields` into each hit object — never `_seq_no` or `_version`, unconditionally, regardless of what the caller asked for. There's no `seq_no_primary_term`/`emit_seq_no` check anywhere in `scroll_page_response` at all; the feature was never wired up for scroll continuation pages.

## Impact

Any client relying on `_seq_no` past the first scroll page — which includes anything doing a consistent, versioned re-index or migration via scroll, e.g. `xerj-seed` (a companion CLI, see the note at the bottom) — silently loses the value from page 2 onward. Not a crash: the field is just missing, which downstream code may or may not notice depending on how strictly it validates the response.

## Suggested fix

- In `scroll_page_response`, resolve the real `_seq_no`/`_primary_term` the same way `search()` does (`idx.lookup_seq_no(&h.id)`) instead of the hardcoded `Some(0)`/`Some(1)`, and actually insert `_seq_no`/`_primary_term`/`_version` into the response JSON when appropriate.
- Whether "when appropriate" is gated per-continuation-request or fixed at scroll-open time is worth deciding deliberately: real ES fixes it at open time (the flag isn't accepted on the continuation body at all) and persists it for the scroll's lifetime — capturing `seq_no_primary_term: bool` in the scroll context (`ctx`) at the point `search()` opens it, rather than trying to read it off each `next_scroll` call, would match that and also close the gap that repeating the flag in the continuation body currently doesn't help.

---

Filed while testing [xerj-seed](https://github.com/Vinz2168/xerj-seed) — its scan (chosen as `_scroll` over PIT specifically for cross-engine portability, see its README) works around this with one `_mget?seq_no_primary_term=true` call per affected page, but that's a client-side patch over a real gap in the response, not a fix.

Vinz2168:
Fixes #428.

## The bug

`seq_no_primary_term: true` on the scroll-opening request (`POST /{index}/_search?scroll=…`) is honored on that first page — real `_seq_no`/`_primary_term`, resolved via `idx.lookup_seq_no`. Every page after the first, served by `scroll_page_response`, came back with neither field at all, silently, regardless of what was requested.

Two compounding bugs in `scroll_page_response` (`es_compat.rs`):

1. The intermediate `EsHit` it builds hardcoded the values:
   ```rust
   version: Some(1),
   seq_no: Some(0),
   ```
2. Even those placeholders never reached the wire — the response is built by hand into a `serde_json::Map` a few lines later, which never inserted `_seq_no`/`_version`/`_primary_term` at all, unconditionally. `EsHit`'s own `Serialize` derive (which does have the right `skip_serializing_if`/rename attributes) was simply never used for this response.

Live-verified this is xerj-specific, not a scroll-protocol limitation: the identical request against a real OpenSearch 3.7.0 node carries `_seq_no`/`_primary_term` on every page.

## The fix

Real Elasticsearch fixes `seq_no_primary_term`/`version` at the scroll-opening request and doesn't even accept them again on the `POST /_search/scroll` continuation body — the flag governs the whole scroll's lifetime, not each individual page. So:

- `ScrollContext` now carries `seq_no_primary_term: bool` and `version: bool`, captured once when the context is created (both call sites: the main `search()` handler and `search_with_scroll`).
- `scroll_page_response` resolves the real values from the version map (`idx.lookup_seq_no` / `idx.lookup_version`) — the exact same mechanism the main search handler already uses for a plain (non-scroll) request — instead of the hardcoded placeholders, gated on the context's flags rather than re-reading them per continuation call.
- The hand-built response JSON now actually inserts `_seq_no`/`_primary_term`/`_version` when present, closing the second, independent omission.

## Testing

- New regression test (`scroll_continuation_resolves_real_seq_no_when_requested`) indexes real documents through the actual write path — every existing scroll test injects a synthetic `ScrollContext` directly with fabricated `Hit`s, which is exactly why this had no version-map entry to expose the bug. Pages one at a time to force a real page 2, asserts `_seq_no` is present and numeric on both page 1 and page 2.
- `cargo test -p xerj-api --lib`: 216 passed, 0 failed (includes the new test; no regressions in the existing 9 scroll-related tests).
- `cargo test -p xerj-engine --test search_context_ttl`: 2 passed (unaffected by the new `ScrollContext` fields).
- `cargo clippy -p xerj-api -p xerj-engine --lib`: clean.
- Live-verified against a real build: 4 real scroll pages against a 120-document index, `_seq_no` present throughout with correct incrementing values (not the old hardcoded `0`); `version: true` also resolves correctly; both fields stay correctly absent on every page when neither flag is requested, so this isn't an over-eager fix that starts emitting fields nobody asked for.
- Live-verified end-to-end against [xerj-seed](https://github.com/Vinz2168/xerj-seed) (the companion CLI whose multi-page testing surfaced this bug): its client-side `_mget` workaround for this exact gap still runs unchanged against the fixed build, harmlessly, since `_seq_no` is native again.

## Scope note

`version` was not part of the original report but is the identical bug in the identical two places (same hardcoded-then-unserialized pattern, one field over) — fixing only `_seq_no` and leaving `_version` visibly broken two lines away would have looked like an oversight rather than a deliberate scope decision, so both are covered here.

Claude-Session: https://claude.ai/code/session_01QXqdmVtcAdiSrcsatXwYFy

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,8 +38,12 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   behaviour — that keeping the raw index spec left per-hit `_index`
   "authoritative when paging" — which is why the defect outlived review.
 
-  `ScrollContext::hits` is now `Vec<(String, Hit)>`, so a hit carries the index
-  it came from rather than borrowing one from the context. Both context-creation
+  `ScrollContext::hits` now carries a `ScrollHit` per hit — the index it came
+  from, and the meta-fields as they stood when the scroll opened — rather than a
+  bare hit borrowing both from the context. (This field changed twice in this
+  unreleased cycle: first to `Vec<(String, Hit)>` for the index, then to
+  `Vec<ScrollHit>` when #428 showed the meta-fields had to be frozen with it.
+  Recorded as one change because only the final shape ever ships.) Both context-creation
   sites previously discarded it with the same `.map(|(_, h)| h.clone())`.
 
   **Scope, established by an independent adversarial verification of the fix and
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -19915,40 +19915,63 @@ pub async fn search_with_scroll(
         .iter()
         .skip(from)
         .take(size)
-        .map(|(idx_name, h)| EsHit {
-            index: idx_name.clone(),
-            id: h.id.clone(),
-            score: Some(h.score as f64),
-            version: Some(1),
-            seq_no: Some(0),
-            primary_term: Some(1),
-            source: if h.source.is_null() {
+        .map(|(idx_name, h)| {
+            // Resolve, or emit nothing. This branch — `_search_scroll` with no
+            // `scroll=` — previously hardcoded `seq_no: 0` / `version: 1`, and
+            // unlike the scroll branch it serialises through EsHit's derive,
+            // whose `skip_serializing_if = "Option::is_none"` suppresses only
+            // `None`. So the fabricated values reached the wire on every hit
+            // whether or not the caller asked for them: `_seq_no: 0` for a
+            // document whose real sequence number was 8. That is the same
+            // construct this change removes from the scroll branch, five lines
+            // below it.
+            let want_sn = body.seq_no_primary_term.unwrap_or(false);
+            let want_v = body.version.unwrap_or(false);
+            let own = state.engine.get_index(idx_name).ok();
+            let sn = if want_sn {
+                own.as_ref().and_then(|i| i.lookup_seq_no(&h.id))
+            } else {
                 None
-            } else {
-                Some(h.source.clone())
-            },
-            fields: passage_fields(h),
-            sort: if h.sort.is_empty() {
-                None
-            } else {
-                Some(h.sort.clone())
-            },
-            highlight: h.highlight.clone(),
-            explanation: None,
-            inner_hits: None,
-            matched_queries: if h.matched_queries.is_empty() {
-                Value::Null
-            } else {
-                Value::Array(
-                    h.matched_queries
-                        .iter()
-                        .cloned()
-                        .map(Value::String)
-                        .collect(),
-                )
-            },
-            ignored: None,
-            ignored_field_values: None,
+            };
+            EsHit {
+                index: idx_name.clone(),
+                id: h.id.clone(),
+                score: Some(h.score as f64),
+                version: if want_v {
+                    own.as_ref().and_then(|i| i.lookup_version(&h.id))
+                } else {
+                    None
+                },
+                seq_no: sn,
+                primary_term: sn.map(|_| 1u64),
+                source: if h.source.is_null() {
+                    None
+                } else {
+                    Some(h.source.clone())
+                },
+                fields: passage_fields(h),
+                sort: if h.sort.is_empty() {
+                    None
+                } else {
+                    Some(h.sort.clone())
+                },
+                highlight: h.highlight.clone(),
+                explanation: None,
+                inner_hits: None,
+                matched_queries: if h.matched_queries.is_empty() {
+                    Value::Null
+                } else {
+                    Value::Array(
+                        h.matched_queries
+                            .iter()
+                            .cloned()
+                            .map(Value::String)
+                            .collect(),
+                    )
+                },
+                ignored: None,
+                ignored_field_values: None,
+            }
         })
         .collect();
 
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -14250,6 +14250,10 @@ async fn search_impl(
             created: now,
             keep_alive,
             expires_at: now + keep_alive,
+            // Fixed for the scroll's whole lifetime from the opening
+            // request — see the field's doc comment (issue #428).
+            seq_no_primary_term: body.seq_no_primary_term.unwrap_or(false),
+            version: body.version.unwrap_or(false),
         };
         state.engine.scrolls.insert(scroll_id.clone(), ctx);
         response_body["_scroll_id"] = Value::String(scroll_id);
@@ -19815,6 +19819,8 @@ pub async fn search_with_scroll(
             created: now,
             keep_alive,
             expires_at: now + keep_alive,
+            seq_no_primary_term: body.seq_no_primary_term.unwrap_or(false),
+            version: body.version.unwrap_or(false),
         };
         state.engine.scrolls.insert(scroll_id.clone(), ctx);
 
@@ -20020,19 +20026,47 @@ async fn scroll_page_response(
                 .map(|(_, h)| !h.sort.is_empty())
                 .unwrap_or(false);
 
+            // #428: `seq_no_primary_term`/`version` are fixed for the
+            // scroll's whole lifetime by the opening `_search?scroll=…`
+            // request (see `ScrollContext`'s doc comment) — resolve the
+            // real values the same way the main `search()` handler does
+            // (`idx.lookup_seq_no`/`lookup_version`) rather than the
+            // hardcoded `Some(0)`/`Some(1)` every page after the first used
+            // to carry, silently, regardless of what was asked for.
+            let emit_seq_no = ctx.seq_no_primary_term;
+            let emit_version = ctx.version;
+            let index_name = ctx.index.clone();
+            let index_for_lookup = state.engine.get_index(&index_name).ok();
+
             let page_hits: Vec<EsHit> = ctx
                 .hits
                 .iter()
                 .skip(position)
                 .take(page_size)
-                .map(|(hit_index, h)| EsHit {
+                .map(|(hit_index, h)| {
+                    let (seq_no, primary_term) = if emit_seq_no {
+                        let sn = index_for_lookup
+                            .as_ref()
+                            .and_then(|idx| idx.lookup_seq_no(&h.id));
+                        (sn, sn.map(|_| 1u64))
+                    } else {
+                        (None, None)
+                    };
+                    let version = if emit_version {
+                        index_for_lookup
+                            .as_ref()
+                            .and_then(|idx| idx.lookup_version(&h.id))
+                    } else {
+                        None
+                    };
+                    EsHit {
                     // Per-hit index, not the context-level one (#414).
                     index: hit_index.clone(),
                     id: h.id.clone(),
                     score: Some(h.score as f64),
-                    version: Some(1),
-                    seq_no: Some(0),
-                    primary_term: Some(1),
+                    version,
+                    seq_no,
+                    primary_term,
                     source: if h.source.is_null() {
                         None
                     } else {
@@ -20060,6 +20094,7 @@ async fn scroll_page_response(
                     },
                     ignored: None,
                     ignored_field_values: None,
+                    }
                 })
                 .collect();
 
@@ -20086,6 +20121,20 @@ async fn scroll_page_response(
                             Some(s) => json!(s),
                             None => Value::Null,
                         });
+                        // #428: these were resolved above (or left None when
+                        // not requested) but never reached the wire — this
+                        // hit object is built by hand, not through EsHit's
+                        // derived Serialize, so a field only leaves this
+                        // function if it is inserted here explicitly.
+                        if let Some(seq_no) = h.seq_no {
+                            o.insert("_seq_no".to_string(), json!(seq_no));
+                        }
+                        if let Some(primary_term) = h.primary_term {
+                            o.insert("_primary_term".to_string(), json!(primary_term));
+                        }
+                        if let Some(version) = h.version {
+                            o.insert("_version".to_string(), json!(version));
+                        }
                         if let Some(src) = &h.source {
                             o.insert("_source".to_string(), src.clone());
                         }
@@ -20169,6 +20218,8 @@ mod passage_scroll_tests {
                 created: now,
                 keep_alive: Duration::from_secs(60),
                 expires_at: now + Duration::from_secs(60),
+                seq_no_primary_term: false,
+                version: false,
             },
         );
         let app = crate::router::build_es_compat_router(state);
@@ -20225,6 +20276,8 @@ mod passage_scroll_tests {
                 created: now,
                 keep_alive: Duration::from_secs(60),
                 expires_at: now + Duration::from_secs(60),
+                seq_no_primary_term: false,
+                version: false,
             },
         );
         let app = crate::router::build_es_compat_router(state);
@@ -20277,6 +20330,8 @@ mod passage_scroll_tests {
                 created: now,
                 keep_alive: Duration::from_secs(60),
                 expires_at: now + Duration::from_secs(60),
+                seq_no_primary_term: false,
+                version: false,
             },
         );
         let app = crate::router::build_es_compat_router(state);
@@ -20310,6 +20365,87 @@ mod passage_scroll_tests {
             .await
             .unwrap();
         assert_eq!(response.status(), StatusCode::BAD_REQUEST);
+    }
+
+    /// Regression test for #428: `seq_no_primary_term: true` on the
+    /// scroll-opening request was honored on page one (resolved via
+    /// `idx.lookup_seq_no`, same as an ordinary search) but silently
+    /// dropped on every page after — `scroll_page_response` hardcoded
+    /// `seq_no: Some(0)` in the intermediate `EsHit` and then never even
+    /// serialized it (or `_version`) into the response at all, regardless
+    /// of what the opening request asked for. This indexes real documents
+    /// through the actual write path (so there is a real version-map entry
+    /// to resolve, not a synthetic `Hit`) and scrolls one at a time to
+    /// force a real page 2.
+    #[tokio::test]
+    async fn scroll_continuation_resolves_real_seq_no_when_requested() {
+        let state = test_state();
+        let app = crate::router::build_es_compat_router(state);
+
+        for i in 0..3 {
+            let response = app
+                .clone()
+                .oneshot(
+                    Request::post(format!("/seqidx/_doc/d{i}"))
+                        .header("content-type", "application/json")
+                        .body(Body::from(json!({ "n": i }).to_string()))
+                        .unwrap(),
+                )
+                .await
+                .unwrap();
+            assert_eq!(response.status(), StatusCode::CREATED);
+        }
+
+        let open = app
+            .clone()
+            .oneshot(
+                Request::post("/seqidx/_search?scroll=1m")
+                    .header("content-type", "application/json")
+                    .body(Body::from(
+                        json!({
+                            "size": 1,
+                            "seq_no_primary_term": true,
+                            "query": { "match_all": {} },
+                        })
+                        .to_string(),
+                    ))
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+        assert_eq!(open.status(), StatusCode::OK);
+        let bytes = axum::body::to_bytes(open.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let body: Value = serde_json::from_slice(&bytes).unwrap();
+        let scroll_id = body["_scroll_id"].as_str().unwrap().to_string();
+        assert!(
+            body["hits"]["hits"][0].get("_seq_no").and_then(Value::as_u64).is_some(),
+            "page 1 should already carry a real _seq_no: {body}"
+        );
+
+        // Page 2 — exactly what #428 broke.
+        let next = app
+            .clone()
+            .oneshot(
+                Request::post("/_search/scroll")
+                    .header("content-type", "application/json")
+                    .body(Body::from(
+                        json!({ "scroll": "1m", "scroll_id": scroll_id }).to_string(),
+                    ))
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+        assert_eq!(next.status(), StatusCode::OK);
+        let bytes = axum::body::to_bytes(next.into_body(), usize::MAX)
+            .await
+            .unwrap();
+        let body: Value = serde_json::from_slice(&bytes).unwrap();
+        assert!(
+            body["hits"]["hits"][0].get("_seq_no").and_then(Value::as_u64).is_some(),
+            "page 2 must carry a real _seq_no when seq_no_primary_term was requested (#428): {body}"
+        );
     }
 
     #[tokio::test]
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -11199,10 +11199,27 @@ async fn search_impl(
     // Snapshot every matching hit for the scroll context BEFORE we consume
     // merged_hits to build `hits`. We'll register the context after the
     // response_body is built.
-    let scroll_snapshot: Option<Vec<(String, xerj_query::executor::Hit)>> = if is_scroll_request {
-        // Keep the index name with each hit (#414): dropping it here is what
-        // made continuation pages fall back to a single context-level name.
-        Some(merged_hits.clone())
+    let scroll_snapshot: Option<Vec<xerj_engine::engine::ScrollHit>> = if is_scroll_request {
+        // Freeze the meta-fields HERE, with the hits (#428). Resolving them at
+        // render time meant a document updated mid-scroll came back as the old
+        // body carrying the new sequence number, and a reindexer feeding that
+        // back as `if_seq_no` passed the CAS and destroyed the newer revision.
+        // A scroll is a point-in-time view; the metadata has to be part of the
+        // point in time. The index name likewise stays with its hit (#414).
+        let mut snap: Vec<xerj_engine::engine::ScrollHit> = Vec::with_capacity(merged_hits.len());
+        for (idx_name, h) in merged_hits.iter() {
+            let own = state.engine.get_index(idx_name).ok();
+            let seq_no = own.as_ref().and_then(|i| i.lookup_seq_no(&h.id));
+            let version = own.as_ref().and_then(|i| i.lookup_version(&h.id));
+            snap.push(xerj_engine::engine::ScrollHit {
+                index: idx_name.clone(),
+                hit: h.clone(),
+                seq_no,
+                primary_term: seq_no.map(|_| 1u64),
+                version,
+            });
+        }
+        Some(snap)
     } else {
         None
     };
@@ -19758,46 +19775,74 @@ pub async fn search_with_scroll(
         let scroll_id = Uuid::new_v4().to_string();
         // Extract just the hits from the pairs.
         // Index name stays paired with its hit (#414).
-        let hits_only: Vec<(String, xerj_query::executor::Hit)> = all_hits.clone();
+        // Freeze the meta-fields with the hits (#428) — see the sibling
+        // snapshot in `search_impl`. This route's own first page is rendered
+        // from the same values just below, so page one and every continuation
+        // page now agree instead of disagreeing (they previously both omitted
+        // the fields; wiring only the continuation path made them diverge).
+        let hits_only: Vec<xerj_engine::engine::ScrollHit> = all_hits
+            .iter()
+            .map(|(idx_name, h)| {
+                let own = state.engine.get_index(idx_name).ok();
+                let seq_no = own.as_ref().and_then(|i| i.lookup_seq_no(&h.id));
+                let version = own.as_ref().and_then(|i| i.lookup_version(&h.id));
+                xerj_engine::engine::ScrollHit {
+                    index: idx_name.clone(),
+                    hit: h.clone(),
+                    seq_no,
+                    primary_term: seq_no.map(|_| 1u64),
+                    version,
+                }
+            })
+            .collect();
 
         // Return first page.
-        let first_page: Vec<EsHit> = all_hits
+        let want_seq_no = body.seq_no_primary_term.unwrap_or(false);
+        let want_version = body.version.unwrap_or(false);
+        // Built from the SAME snapshot the continuation pages read, so this
+        // route stops disagreeing with itself. It previously omitted the
+        // meta-fields on page one while continuations emitted them, which is
+        // #428's own defect one route over.
+        let first_page: Vec<EsHit> = hits_only
             .iter()
             .take(page_size)
-            .map(|(idx_name, h)| EsHit {
-                index: idx_name.clone(),
-                id: h.id.clone(),
-                score: Some(h.score as f64),
-                version: Some(1),
-                seq_no: Some(0),
-                primary_term: Some(1),
-                source: if h.source.is_null() {
-                    None
-                } else {
-                    Some(h.source.clone())
-                },
-                fields: passage_fields(h),
-                sort: if h.sort.is_empty() {
-                    None
-                } else {
-                    Some(h.sort.clone())
-                },
-                highlight: h.highlight.clone(),
-                explanation: None,
-                inner_hits: None,
-                matched_queries: if h.matched_queries.is_empty() {
-                    Value::Null
-                } else {
-                    Value::Array(
-                        h.matched_queries
-                            .iter()
-                            .cloned()
-                            .map(Value::String)
-                            .collect(),
-                    )
-                },
-                ignored: None,
-                ignored_field_values: None,
+            .map(|sh| {
+                let h = &sh.hit;
+                EsHit {
+                    index: sh.index.clone(),
+                    id: h.id.clone(),
+                    score: Some(h.score as f64),
+                    version: sh.version,
+                    seq_no: sh.seq_no,
+                    primary_term: sh.primary_term,
+                    source: if h.source.is_null() {
+                        None
+                    } else {
+                        Some(h.source.clone())
+                    },
+                    fields: passage_fields(h),
+                    sort: if h.sort.is_empty() {
+                        None
+                    } else {
+                        Some(h.sort.clone())
+                    },
+                    highlight: h.highlight.clone(),
+                    explanation: None,
+                    inner_hits: None,
+                    matched_queries: if h.matched_queries.is_empty() {
+                        Value::Null
+                    } else {
+                        Value::Array(
+                            h.matched_queries
+                                .iter()
+                                .cloned()
+                                .map(Value::String)
+                                .collect(),
+                        )
+                    },
+                    ignored: None,
+                    ignored_field_values: None,
+                }
             })
             .collect();
 
@@ -19839,6 +19884,20 @@ pub async fn search_with_scroll(
                         "_score": h.score,
                         "_source": h.source,
                     });
+                    // Same fields, same condition, as the continuation pages.
+                    if want_seq_no {
+                        if let Some(sn) = h.seq_no {
+                            hit["_seq_no"] = json!(sn);
+                        }
+                        if let Some(pt) = h.primary_term {
+                            hit["_primary_term"] = json!(pt);
+                        }
+                    }
+                    if want_version {
+                        if let Some(v) = h.version {
+                            hit["_version"] = json!(v);
+                        }
+                    }
                     if let Some(fields) = &h.fields {
                         hit["fields"] = serde_json::to_value(fields).unwrap_or(Value::Null);
                     }
@@ -20023,7 +20082,7 @@ async fn scroll_page_response(
             let has_non_score_sort = ctx
                 .hits
                 .first()
-                .map(|(_, h)| !h.sort.is_empty())
+                .map(|sh| !sh.hit.sort.is_empty())
                 .unwrap_or(false);
 
             // #428: `seq_no_primary_term`/`version` are fixed for the
@@ -20035,44 +20094,28 @@ async fn scroll_page_response(
             // to carry, silently, regardless of what was asked for.
             let emit_seq_no = ctx.seq_no_primary_term;
             let emit_version = ctx.version;
-            // Resolve the handle PER HIT, against the index that hit actually
-            // came from — not against the context-level `ctx.index`. #424 made
-            // `_index` per-hit and explicitly demoted that field to a
-            // "context-level fallback only"; looking the version map up in one
-            // index while reporting another is how a two-index scroll ends up
-            // serving `_seq_no` read out of the wrong index. This mirrors the
-            // main search handler, which resolves inside its per-hit loop.
-            // Cached per distinct index name so a page still costs one
-            // `get_index` per index, not one per hit.
-            let mut handles: std::collections::HashMap<String, Option<_>> =
-                std::collections::HashMap::new();
-            for (hit_index, _) in ctx.hits.iter().skip(position).take(page_size) {
-                handles
-                    .entry(hit_index.clone())
-                    .or_insert_with(|| state.engine.get_index(hit_index).ok());
-            }
-
+            // Read the meta-fields OUT OF THE SNAPSHOT. Resolving them here
+            // against the live index is what produced a silent lost update:
+            // the frozen `source` below was paired with a `_seq_no` from now,
+            // so a document updated mid-scroll came back as the old body
+            // carrying the new sequence number, and `if_seq_no` fed back from
+            // it passed the CAS and overwrote the newer revision (#428).
             let page_hits: Vec<EsHit> = ctx
                 .hits
                 .iter()
                 .skip(position)
                 .take(page_size)
-                .map(|(hit_index, h)| {
-                    let own_index = handles.get(hit_index).and_then(|h| h.as_ref());
+                .map(|sh| {
+                    let h = &sh.hit;
                     let (seq_no, primary_term) = if emit_seq_no {
-                        let sn = own_index.and_then(|idx| idx.lookup_seq_no(&h.id));
-                        (sn, sn.map(|_| 1u64))
+                        (sh.seq_no, sh.primary_term)
                     } else {
                         (None, None)
                     };
-                    let version = if emit_version {
-                        own_index.and_then(|idx| idx.lookup_version(&h.id))
-                    } else {
-                        None
-                    };
+                    let version = if emit_version { sh.version } else { None };
                     EsHit {
                         // Per-hit index, not the context-level one (#414).
-                        index: hit_index.clone(),
+                        index: sh.index.clone(),
                         id: h.id.clone(),
                         score: Some(h.score as f64),
                         version,
@@ -20221,8 +20264,20 @@ mod passage_scroll_tests {
             xerj_engine::engine::ScrollContext {
                 index: "reports".into(),
                 hits: vec![
-                    ("reports".to_string(), passage_hit("page-1", 0)),
-                    ("reports".to_string(), passage_hit("page-2", 1)),
+                    xerj_engine::engine::ScrollHit {
+                        index: "reports".to_string(),
+                        hit: passage_hit("page-1", 0),
+                        seq_no: None,
+                        primary_term: None,
+                        version: None,
+                    },
+                    xerj_engine::engine::ScrollHit {
+                        index: "reports".to_string(),
+                        hit: passage_hit("page-2", 1),
+                        seq_no: None,
+                        primary_term: None,
+                        version: None,
+                    },
                 ],
                 position: 1,
                 page_size: 1,
@@ -20279,8 +20334,20 @@ mod passage_scroll_tests {
             xerj_engine::engine::ScrollContext {
                 index: "mi_a".into(),
                 hits: vec![
-                    ("mi_a".to_string(), passage_hit("dup-1", 0)),
-                    ("mi_b".to_string(), passage_hit("dup-1", 1)),
+                    xerj_engine::engine::ScrollHit {
+                        index: "mi_a".to_string(),
+                        hit: passage_hit("dup-1", 0),
+                        seq_no: None,
+                        primary_term: None,
+                        version: None,
+                    },
+                    xerj_engine::engine::ScrollHit {
+                        index: "mi_b".to_string(),
+                        hit: passage_hit("dup-1", 1),
+                        seq_no: None,
+                        primary_term: None,
+                        version: None,
+                    },
                 ],
                 position: 1,
                 page_size: 1,
@@ -20333,8 +20400,20 @@ mod passage_scroll_tests {
             xerj_engine::engine::ScrollContext {
                 index: "reports".into(),
                 hits: vec![
-                    ("reports".to_string(), passage_hit("page-1", 0)),
-                    ("reports".to_string(), passage_hit("page-2", 1)),
+                    xerj_engine::engine::ScrollHit {
+                        index: "reports".to_string(),
+                        hit: passage_hit("page-1", 0),
+                        seq_no: None,
+                        primary_term: None,
+                        version: None,
+                    },
+                    xerj_engine::engine::ScrollHit {
+                        index: "reports".to_string(),
+                        hit: passage_hit("page-2", 1),
+                        seq_no: None,
+                        primary_term: None,
+                        version: None,
+                    },
                 ],
                 position: 1,
                 page_size: 1,
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -20035,8 +20035,22 @@ async fn scroll_page_response(
             // to carry, silently, regardless of what was asked for.
             let emit_seq_no = ctx.seq_no_primary_term;
             let emit_version = ctx.version;
-            let index_name = ctx.index.clone();
-            let index_for_lookup = state.engine.get_index(&index_name).ok();
+            // Resolve the handle PER HIT, against the index that hit actually
+            // came from — not against the context-level `ctx.index`. #424 made
+            // `_index` per-hit and explicitly demoted that field to a
+            // "context-level fallback only"; looking the version map up in one
+            // index while reporting another is how a two-index scroll ends up
+            // serving `_seq_no` read out of the wrong index. This mirrors the
+            // main search handler, which resolves inside its per-hit loop.
+            // Cached per distinct index name so a page still costs one
+            // `get_index` per index, not one per hit.
+            let mut handles: std::collections::HashMap<String, Option<_>> =
+                std::collections::HashMap::new();
+            for (hit_index, _) in ctx.hits.iter().skip(position).take(page_size) {
+                handles
+                    .entry(hit_index.clone())
+                    .or_insert_with(|| state.engine.get_index(hit_index).ok());
+            }
 
             let page_hits: Vec<EsHit> = ctx
                 .hits
@@ -20044,18 +20058,15 @@ async fn scroll_page_response(
                 .skip(position)
                 .take(page_size)
                 .map(|(hit_index, h)| {
+                    let own_index = handles.get(hit_index).and_then(|h| h.as_ref());
                     let (seq_no, primary_term) = if emit_seq_no {
-                        let sn = index_for_lookup
-                            .as_ref()
-                            .and_then(|idx| idx.lookup_seq_no(&h.id));
+                        let sn = own_index.and_then(|idx| idx.lookup_seq_no(&h.id));
                         (sn, sn.map(|_| 1u64))
                     } else {
                         (None, None)
                     };
                     let version = if emit_version {
-                        index_for_lookup
-                            .as_ref()
-                            .and_then(|idx| idx.lookup_version(&h.id))
+                        own_index.and_then(|idx| idx.lookup_version(&h.id))
                     } else {
                         None
                     };
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -20060,40 +20060,40 @@ async fn scroll_page_response(
                         None
                     };
                     EsHit {
-                    // Per-hit index, not the context-level one (#414).
-                    index: hit_index.clone(),
-                    id: h.id.clone(),
-                    score: Some(h.score as f64),
-                    version,
-                    seq_no,
-                    primary_term,
-                    source: if h.source.is_null() {
-                        None
-                    } else {
-                        Some(h.source.clone())
-                    },
-                    fields: passage_fields(h),
-                    sort: if h.sort.is_empty() {
-                        None
-                    } else {
-                        Some(h.sort.clone())
-                    },
-                    highlight: h.highlight.clone(),
-                    explanation: None,
-                    inner_hits: None,
-                    matched_queries: if h.matched_queries.is_empty() {
-                        Value::Null
-                    } else {
-                        Value::Array(
-                            h.matched_queries
-                                .iter()
-                                .cloned()
-                                .map(Value::String)
-                                .collect(),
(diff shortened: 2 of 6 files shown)


xerj-org:
## Independent verification — REFUTED 2/2, 4 must-fix

Both lenses reached the same root cause independently, and it is one neither the contributor nor I named.


---

## Lens 1 — refuted `True`, confidence high

> REFUTED. The maintainer's deliberate non-fix is the wrong call, but not because the delete case is severe — it is loud and rate-limited (~churn/2). It is the wrong call because he applied his own stated severity criterion ("a consumer that used to fail loudly now gets a confident wrong value") to the benign half of the defect and shipped the dangerous half unmentioned. The root cause is that `scroll_page_response` renders a SNAPSHOTTED hit list but resolves `_seq_no`/`_version` LIVE. Deleting a doc mid-scroll makes the fields absent (loud). UPDATING a doc mid-scroll — far more common in a reindex — makes them present and wrong: the continuation page returns the snapshot `_source` paired with the post-update `_seq_no`/`_version`. Measured on the PR head: hit `d4` returns `{"body":"v1"}` with `_seq_no:7,_version:3` while live truth for `d4` is `body:"v3", _seq_no:7, _version:3`. A reindexer that copies body + `if_seq_no` writes v1 over v3 and the CAS SUCCEEDS — a silent lost update, on exactly the migration/CDC path #414's changelog names as scroll's blast radius. This failure mode is INTRODUCED BY THIS PR: before #431 continuation pages emitted no `_seq_no` at all, so there was nothing to feed back. Quantified at 1000 docs / size 50 / 20 pages: 10% churn → 4.60% of served hits skewed (46/1000), 20% → 9.00% (90/1000). The delete case measures identically (churn/2) because it is the same walk geometry. Secondary: "that is the contributor's code and their call" is half fair — the bare `.and_then` shape is theirs (4eaaa97), but the call was never put to them in any channel they read (0 PR comments, 0 reviews, no CHANGELOG entry, #428 still open and silent; the disclosure lives only in a commit body force-pushed to their own fork). And the lens hypothesis that page one is the dishonest one did NOT hold for the delete path: I could not reach `.unwrap_or(hit_idx as u64)`, because plain `_search` and PIT both exclude tombstoned docs.

### [must-fix] This PR newly ships a silent lost-update: continuation pages pair a snapshot _source with a LIVE _seq_no/_version after a mid-scroll update

`scroll_page_response` serves `h.source` out of the frozen `ctx.hits` list but resolves `_seq_no`/`_version` live via `idx.lookup_seq_no`/`lookup_version` at render time (es_compat.rs:20063, :20069). For a doc UPDATED after the snapshot, the version-map entry is still live, so the lookup succeeds and returns the CURRENT sequence number — stamped onto the OLD body. This is not the maintainer's documented defect (absent field, fails loudly); it is the exact harm he wrote the #424 fix commit to condemn: 'worse than the absent field it replaced: _seq_no and _primary_term exist to be fed back as if_seq_no/if_primary_term, so a consumer that used to fail loudly now gets a confident wrong value.' A reindexer doing `PUT dest/_doc/d4?if_seq_no=7&if_primary_term=1` with the stale body passes optimistic concurrency and destroys the newer revision. Crucially this is a REGRESSION introduced by #431, not a pre-existing defect: prior to this PR the hand-built response Map inserted no `_seq_no`/`_primary_term`/`_version` at all (the PR body itself says so — 'never inserted ... at all, unconditionally'), so no client could feed a wrong value back. The PR converts a loud missing-key failure into a silent data-loss path. Neither the contributor nor the maintainer mentions it anywhere. Fix requires either snapshotting the metadata into `ctx.hits` at open time (matching what `_source` already does) or refusing to emit these fields on continuation pages.

```
Repro (worktree /tmp/v431-defect at 2d43f9b, `rustup run 1.97.1 cargo test -p xerj-api --test zz_verify_snapshot_skew -- --nocapture --test-threads=1`). Open scroll size=2 over 6 docs all body=v1, then rewrite d4/d5 twice each to v2 then v3, then walk:

  E1_CONT2 [{"_index":"skew","_id":"d4","_seq_no":7,"_primary_term":1,"_version":3,"_source":{"n":4,"body":"v1"}},
            {"_index":"skew","_id":"d5","_seq_no":9,"_primary_term":1,"_version":3,"_source":{"n":5,"body":"v1"}}]
  E1_LIVE_TRUTH ... {"_id":"d4","_version":3,"_seq_no":7,"_source":{"n":4,"body":"v3"}},
                    {"_id":"d5","_version":3,"_seq_no":9,"_source":{"n":5,"body":"v3"}}

The scroll hit carries body v1 with the seq_no/version of v3. Quantified over 1000 docs, size=50, 20 pages, victims uniform over the id space and updated in a deterministic shuffled order spread evenly across the walk:

  E5 churn=10% updated_during_walk=100 served=1000 stale_body_with_live_seq_no=46 skew=4.60%
  E5 churn=20% updated_during_walk=200 served=1000 stale_body_with_live_seq_no=90 skew=9.00%

Code: engine/crates/xerj-api/src/es_compat.rs:20063 `let sn = own_index.and_then(|idx| idx.lookup_seq_no(&h.id));` — resolved at page-render time, while `h.source` on line 20073 comes from `ctx.hits` frozen at open time.
```

### [must-fix] Root cause the maintainer did not name: xerj snapshots the hit list but not the metadata, which is not how a scroll works in Lucene/ES

Both the delete symptom (fields vanish) and the update symptom (fields lie) are one bug: `ScrollContext` freezes `hits` but nothing else, so every meta-field is re-derived against the live index on every continuation page. In Lucene — and therefore in ES/OpenSearch — a scroll pins an `IndexReader`; the document deleted after the snapshot is still live in that reader's `liveDocs`, and its `_seq_no` NumericDocValues, written at index time and immutable per segment, reads back the ORIGINAL value. ES emits `_seq_no` for that hit, and it agrees with the `_source` emitted next to it. Framing the delete case as a missing `.unwrap_or(...)` fallback (the maintainer's framing, and what he handed back to the contributor) points at the wrong repair: adding a fallback would fabricate a value, and it does nothing about the update case, which is the dangerous one. The correct repair is to capture `_seq_no`/`_primary_term`/`_version` into `ctx.hits` at scroll-open time, alongside the `_source` that is already captured there.

```
Lucene source, retrieved from the indexed reference corpus (refsym) rather than memory:

  lucene/core/src/java/org/apache/lucene/index/IndexReader.java:63 (class IndexReader) —
  "IndexReader is an abstract class, providing an interface for accessing a point-in-time view of an index. Any changes made to the index via IndexWriter will not be visible until a new IndexReader is opened."

  lucene/core/src/java/org/apache/lucene/index/LeafReader.java:475 (getLiveDocs) —
  "Returns the Bits representing live (not deleted) docs. A set bit indicates the doc ID has not been deleted... The returned instance has been safely published for use by multiple threads without additional synchronization."

xerj by contrast: engine/crates/xerj-engine/src/index.rs:12637 `lookup_seq_no` returns `None` for a tombstone (`if entry.deleted { None }`) and index.rs:12653 `lookup_version` likewise — both consulted live at render time, so the snapshot has no bearing on them.

NOTE: no real ES or OpenSearch node exists in this environment to measure against. `curl localhost:9200` returns `{"cluster_name":"xerj","version":{"number":"8.13.0","build_hash":"00000000"}}` — a xerj process impersonating ES (pid 76379, confirmed via `ss -lntp`). Every listener on this host is a xerj binary. The ES claim above is rea
```

### [should-fix] The documented delete defect reproduces at 100% for affected docs; loss rate is ~half the delete churn over the walk

Confirmed exactly as the maintainer describes. On a single index, every doc deleted between the snapshot and the page that serves it loses all three fields — `_seq_no`, `_primary_term` AND `_version` — while still being returned with its snapshot `_source`. The hit is a half-snapshot: body present, identity metadata gone. In a Python `elasticsearch` client `hit["_seq_no"]` raises KeyError and the migration dies mid-walk; in a JS client `if_seq_no: undefined` is dropped from the query string and the write silently becomes unconditioned. It is loud enough to be survivable, and rate-limited: expected loss = d·∫₀¹t dt = d/2 for churn d spread uniformly over the walk, which the measurement confirms. Easy to hit in a real reindex — 20% source churn during a long walk costs 9% of hits — but this is the LESS harmful half of the defect, which is why picking it as the one to document is the wrong call.

```
Repro (`rustup run 1.97.1 cargo test -p xerj-api --test zz_verify_delete_during_scroll`). 6 docs, scroll size=2, delete d2..d5 after page one:

  PAGE1 = [{"_id":"d0","_version":1,"_seq_no":0,"_primary_term":1,...},{"_id":"d1","_version":1,"_seq_no":1,...}]
  DELETE d2 -> 200 OK {"_version":2,"result":"deleted","_seq_no":6,"_primary_term":1}
  ...
  CONT page1 id="d2" _seq_no=None _primary_term=None _version=None
  CONT page1 id="d3" _seq_no=None _primary_term=None _version=None
  CONT page2 id="d4" _seq_no=None _primary_term=None _version=None
  CONT page2 id="d5" _seq_no=None _primary_term=None _version=None
  RESULT_A total_continuation_hits=4 missing_seq_no=4

Quantified, 1000 docs / size 50 / 20 pages, victims uniform over the id space, deleted in shuffled order spread evenly across the walk:

  E4 churn=5%  victims=50  served=1000 hits_without_seq_no=29 loss=2.90%
  E4 churn=10% victims=100 served=1000 hits_without_seq_no=46 loss=4.60%
  E4 churn=20% victims=200 served=1000 hits_without_seq_no=90 loss=9.00%

(An earlier run where delete order correlated with scroll order gave 0.00% at 10% and 20% — that measurement was an artifact and I discarded it. The shuffled numbers above match the analytic d/2.)
```

### [should-fix] #431 newly puts a wrong _version on every scroll page: after a process restart _version resolves to 1 for every document

`_version` survives a restart incorrectly while `_seq_no` survives correctly. Before this PR that only affected page one and plain `_search`; #431 is what starts emitting `_version` on continuation pages, so it propagates the wrong value across the whole scroll. The PR body's testing claim — 'version: true also resolves correctly' — holds only for a process that has never restarted, which is not the state any production reindex runs against. The underlying durability bug is pre-existing (the unchanged plain-`_search` path is equally wrong), so this is a blast-radius expansion rather than a new bug, but it is a confident wrong value being newly put on the wire by this PR, which is the same category the maintainer used to justify his own #424 fix.

```
Repro (`rustup run 1.97.1 cargo test -p xerj-api --test zz_verify_reopen -- --nocapture`). 5 docs written 1,2,3,4,5 times respectively, `_refresh` + `_flush`, then a fresh `Engine::new` over the same data_dir:

  F1_BEFORE_RESTART  d0 _version:1 _seq_no:0 | d1 _version:2 _seq_no:2 | d2 _version:3 _seq_no:5 | d3 _version:4 _seq_no:9 | d4 _version:5 _seq_no:14
  F1_AFTER_RESTART   d0 _version:1 _seq_no:0 | d1 _version:1 _seq_no:2 | d2 _version:1 _seq_no:5 | d3 _version:1 _seq_no:9 | d4 _version:1 _seq_no:14
  F1_SCROLL_CONT1 200 OK [{"_id":"d2","_seq_no":5,"_primary_term":1,"_version":1,...},{"_id":"d3","_seq_no":9,"_primary_term":1,"_version":1,...}]
  F1_SCROLL_CONT2 200 OK [{"_id":"d4","_seq_no":14,"_primary_term":1,"_version":1,...}]

_seq_no recovers exactly (0,2,5,9,14). _version does not (1,1,1,1,1 vs 1,2,3,4,5).
```

### [should-fix] "That is the contributor's code and their call" — the call was never put to the contributor in any channel they read

The attribution of the code shape is fair: the bare `.and_then` with no fallback is the contributor's, introduced in 4eaaa97 as `index_for_lookup.as_ref().and_then(|idx| idx.lookup_seq_no(&h.id))`, and the maintainer's edit only changed WHICH handle it resolves against. But the disclosure exists in exactly one place — the body of commit 2d43f9b, force-pushed by the maintainer onto the contributor's own fork branch. PR #431 has zero comments and zero reviews. The PR touches four files, none of them CHANGELOG.md. Issue #428 is still OPEN with no mention of the carve-out. A contributor is not notified of, and does not routinely re-read, the body of a commit a maintainer pushed to their branch. Calling it "their call" describes a decision that was never actually offered.

```
`gh pr view 431 --repo xerj-org/xerj --json reviews,headRepositoryOwner,isCrossRepository,files` →
  reviews: []
  headRepoOwner: {'login': 'Vinz2168'}
  crossRepo: True
  files: ['engine/crates/xerj-api/src/es_compat.rs', 'engine/crates/xerj-api/tests/scroll_multi_index_seq_no.rs', 'engine/crates/xerj-engine/src/engine.rs', 'engine/crates/xerj-engine/tests/search_context_ttl.rs']

`gh pr view 431 --json comments` → empty.
`gh issue list --search "scroll seq_no" --state all` → only #428, state OPEN.

Attribution check, `git diff 7886463..2d43f9b`: the maintainer replaced `let index_for_lookup = state.engine.get_index(&index_name).ok();` with a per-hit `handles` HashMap, then `let own_index = handles.get(hit_index).and_then(|h| h.as_ref());`. The no-fallback `.and_then` on the lookup itself is unchanged contributor code.
```

### [note] The maintainer's per-hit handle cache adds a second None source: an index dropped mid-scroll silently strips all three fields from every remaining hit

`handles.entry(hit_index.clone()).or_insert_with(|| state.engine.get_index(hit_index).ok())` at es_compat.rs:20044-20048 stores `None` when `get_index` fails, and `own_index` then short-circuits both lookups. This is a failure path that did not previously exist in this shape (the contributor resolved one handle once, off `ctx.index`), though the pre-edit code would have degraded identically for a single-index scroll, so it is not strictly newly reachable. Aliases are fine — a scroll opened on an alias resolves the version map correctly. Worth noting that the scroll also keeps happily serving hits from an index that no longer exists, which is arguably a separate wire-compat question.

```
`rustup run 1.97.1 cargo test -p xerj-api --test zz_verify_snapshot_skew -- --nocapture` →

  E3_DROP 200 OK {"acknowledged":true}
  E3_CONT1 200 OK [{"_index":"dropme","_id":"d2","_score":1.0,"_source":{"n":2,"body":"v1"},"sort":["d2"]},{"_index":"dropme","_id":"d3",...}]
  E3_CONT2 200 OK [{"_index":"dropme","_id":"d4",...},{"_index":"dropme","_id":"d5",...}]

No _seq_no / _primary_term / _version on any hit, index deleted mid-scroll, hits still served. Alias control (E2) is clean: scroll on alias `al` over index `phys` returns _seq_no 0,1,2,3,4,5 correctly across page one and both continuation pages.
```

### [note] The lens hypothesis that page one is the dishonest path did NOT hold for the delete case — but the same ordinal fabrication IS live in inner_hits

I could not reach the top-level page-one fallback `.unwrap_or(hit_idx as u64)` (es_compat.rs:13251) via a deleted document: plain `_search` excludes tombstoned docs, and PIT search does too (the doc never reaches the render loop, despite the PIT filter's own `.unwrap_or(0)` at es_compat.rs:10706 being permissive). So for the delete scenario specifically, the maintainer's 'inconsistency' framing is NOT backwards, and the continuation path is not the lone honest one. However the identical ordinal-fabrication pattern is reachable and unconditional one function over: the collapse/inner_hits renderer hardcodes `hit_obj.insert("_seq_no", json!(i as u64))` at es_compat.rs:13156-13158, emitting the position in the inner-hit list as the sequence number. Pre-existing on main, untouched by this PR, and out of scope for it — but it means a client can already receive a plausible-but-wrong `_seq_no` from xerj on a page-one response, which weakens any argument that page one is the trustworthy reference the continuation path should be made to match.

```
`rustup run 1.97.1 cargo test -p xerj-api --test zz_verify_pageone_fallback -- --nocapture --test-threads=1`:

  D1_DELETE 200 OK {"_id":"d2","_version":4,"result":"deleted","_seq_no":15}
  D1_AFTER_NOREFRESH  [d0 _seq_no:10, d1 _seq_no:11, d3 _seq_no:13, d4 _seq_no:14]   # d2 absent, no fabrication
  D2_PIT_SEARCH 200 OK [d0 _seq_no:10, d4 _seq_no:14]                                 # d1,d2,d3 deleted after PIT open, all excluded

  D3: docs d0..d5 written 1..6 times, so true _seq_no is 0,2,5,9,14,20.
  D3_COLLAPSE inner_hits → _seq_no 0,1,2,3,4,5 with _version 1,2,3,4,5,6.
  The versions are right; every seq_no is the ordinal.
```


---

## Lens 2 — refuted `True`, confidence high

> The maintainer's per-hit fix does what it claims — I could not break it on the axis it was written for. Built PR head (2d43f9b) with `rustup run 1.97.1 cargo build --release -p xerj-server`, ran a live node on 127.0.0.1:19431, and scrolled 2, 3 and 4 indices with ids colliding across all of them and write counts 1/2/3/5 so `_seq_no` is never the ordinal (z4 truth is sn16-19 at ordinals 12-15). Page sizes 1,2,3,5,6,7,9,11 — including pages spanning two and three indices with the boundary landing mid-page — produced zero mismatches against a plain `_search` with the same flags. Wildcard `mi*` and a single-index alias also clean. The PR's own two integration tests pass under 1.97.1.
> 
> The PR is still refuted, because the fix's success is what exposes the real problem: it turns an absent field into a *present, confidently wrong* one in exactly the direction that defeats its own stated motivation. Four defects, three of them created by this PR (base 9ebdac2 emitted no `_seq_no`/`_primary_term`/`_version` on any continuation page at all — verified in the base source):
> 
> 1. Continuation pages resolve the version map LIVE while serving a point-in-time `_source`. Feeding the hit's own `_seq_no` back as `if_seq_no` — the use the PR and the maintainer's commit message both name — silently overwrites a concurrent writer instead of 409-ing. Demonstrated end to end.
> 2. The `/{index}/_search_scroll` route now disagrees with itself: page 1 omits the fields, page 2+ emits them. That is the inverse of the bug being fixed, newly introduced here.
> 3. An alias repointed (or an index deleted and recreated) mid-scroll makes the per-hit lookup read the wrong index's version map — the exact failure the maintainer's fix was written to eliminate, arriving through a different door, because the fix resolves the *name recorded on the hit* at continuation time rather than the index the hit came from.
> 4. A deleted index costs 6 of 8 hits all three fields while `_shards` still reports `"failed": 0`. No panic, no error.
> 
> Plus the answer to the direct question: `_primary_term` is `sn.map(|_| 1u64)` — a constant 1 in every case, coupled to `sn` only for presence.

### [must-fix] Continuation pages resolve _seq_no/_version LIVE against a point-in-time snapshot, producing a silent lost update through if_seq_no

ScrollContext pins a fully-hydrated `Vec<(String, Hit)>` snapshot including `Hit.source`. The new code resolves `_seq_no`/`_version` at continuation time via `idx.lookup_seq_no`/`lookup_version`, which read the CURRENT version map. Page one is self-consistent because it resolves at the instant the snapshot is taken; every page after is not. The hit therefore carries `_source` from scroll-open time next to `_seq_no`/`_version` from now.

This is strictly worse than the absent field it replaced, and worse in the specific direction the maintainer's own commit message argues against. In real ES a scroll holds a searcher over frozen segments, so the hit's `_seq_no` is the SNAPSHOT value and a consumer's `if_seq_no` correctly 409s when the doc moved. Here the reported `_seq_no` is the live one, so the CAS SUCCEEDS and the stale snapshot body is written over the newer document. `if_seq_no` is honoured by the engine (index.rs:6680), which is why this lands rather than being cosmetic.

Before this PR the fields were absent on continuation pages, so no consumer could be misled this way — @Vinz2168's own xerj-seed used an `_mget` workaround. This defect is created by the PR. It sits on precisely the reindex/migration/CDC path that #414's changelog and the maintainer's commit both name as scroll's blast radius.

```
Probe K against the live PR-head node (/tmp/claude-1001/.../scratchpad/probe431b.py):

  page1 id=0 _seq_no=0 _version=1 round=0
  concurrent writer updated cas/1 -> live _seq_no=6 _version=4 round=99
  page2 id=1 _seq_no=6 _primary_term=1 _version=4 _source.round=0
  -> _source is the SNAPSHOT (round=0) but _seq_no/_version are LIVE (6/4)
  PUT /cas/_doc/1?if_seq_no=6&if_primary_term=1 with the SNAPSHOT source -> 200
  cas/1 now: round=0 who=cas
  [FAIL] LOST UPDATE: the concurrent writer's round=99 was silently overwritten

Same drift seen independently in probe F (mi1/2 updated 6x after open):
    id=2 _seq_no=9 _version=7 _source.round=0

Base check — the fields did not exist on continuation pages before this PR:
  $ git show 9ebdac2:engine/crates/xerj-api/src/es_compat.rs
  BASE scroll_page_response contains '_seq_no' insert: False
  BASE hardcoded seq_no: True

Code: engine/crates/xerj-api/src/es_compat.rs:20060-20071 (`own_index.and_then(|idx| idx.lookup_seq_no(&h.id))`) vs engine/crates/xerj-engine/src/engine.rs:154-186 (ScrollContext pins the snapshot).

CAS is real, not a no-op: engine/crates/xerj-engine/src/index.rs:6680 `if let Some(expected_seq) = if_seq_no { ... entry.seq_no != expected_seq -> version_conflict }`.
```

### [must-fix] PR newly creates a page-1/page-2 disagreement on /{index}/_search_scroll — the inverse of the bug it fixes

The PR wires `seq_no_primary_term`/`version` into the ScrollContext at BOTH creation sites, including `search_with_scroll` (es_compat.rs:19819-19820, routed at router.rs:387-390 as `/{index}/_search_scroll`). But that route's page-one response is hand-built at es_compat.rs:19835-19846 and still emits only `_index`/`_id`/`_score`/`_source`/`fields`; the PR does not touch it, and it still constructs its EsHit with the hardcoded `version: Some(1), seq_no: Some(0), primary_term: Some(1)` (es_compat.rs:19769-19771).

Result: on base, both pages omitted the fields — broken but consistent. After this PR, page one omits them and every continuation page emits them. A client that reads `hit["_seq_no"]` unconditionally now KeyErrors on page one and succeeds on page two of the same scroll. That is the same class of defect (#428) the PR exists to fix, reintroduced one route over, by the PR itself.

```
Probe I/P against the live PR-head node:

  page1 hit keys: ['_id', '_index', '_score', '_source']
  page2 hit keys: ['_id', '_index', '_primary_term', '_score', '_seq_no', '_source', '_version']
  page1 sn=[None, None]  page2 sn=[10, 11]

Request: POST /leg/_search_scroll?scroll=5m {"size":2,"seq_no_primary_term":true,"version":true,"query":{"match_all":{}}} then POST /_search/scroll.

Diff evidence that this is new: `git diff 9ebdac2..2d43f9b` adds `seq_no_primary_term`/`version` at es_compat.rs:19819 and adds the three `o.insert("_seq_no"/"_primary_term"/"_version", ...)` lines in scroll_page_response, but leaves search_with_scroll's first-page builder untouched.
```

### [should-fix] Per-hit lookup resolves the recorded NAME at continuation time, so an alias repoint or index delete+recreate reads the wrong index's version map

The fix's stated contract is "resolve the handle PER HIT, against the index that hit actually came from." It does not do that — it resolves whatever `hit_index` resolves to NOW, via `state.engine.get_index(hit_index)`, which performs alias resolution (engine.rs:3114-3137) and a fresh `indices` map lookup at continuation time.

Two doors reopen the exact wrong-index-version-map bug the maintainer's commit says it closes:

(a) Alias-addressed scroll. `search_impl` does not expand aliases into concrete names (that is #433), so hits are stamped with the ALIAS. Repoint the alias mid-scroll and the continuation reads the new target's version map while serving the old target's documents.

(b) Index deleted and recreated under the same name mid-scroll. `get_index` returns the NEW index; the hits are the OLD one's.

In both cases the values are not merely stale — they are a different index's, served next to `_source` from the original, with no error. Byte-for-byte the failure mode the commit message calls "a confident wrong value" on the reindex/CDC path.

```
Probe M — alias repointed after the scroll opened (ra written 1x, rb written 7x):

  page1 _index=movealias id=0 sn=0 v=1 who=ra
  repointed movealias: ra -> rb (200)
  rb truth: {'0': (24, 7), '1': (25, 7), '2': (26, 7), '3': (27, 7)}
  cont _index=movealias id=1 sn=25 v=7 who=ra  | rb-truth=(25, 7)
  cont _index=movealias id=2 sn=26 v=7 who=ra  | rb-truth=(26, 7)
  cont _index=movealias id=3 sn=27 v=7 who=ra  | rb-truth=(27, 7)

`_source.who=ra` (index ra's documents) carrying rb's seq_no/version exactly.

Probe L — index deleted then recreated with 9 write rounds:

  page1 id=0 sn=0 v=1 round=0
  recreated rec with 9 rounds; live truth now: {'0': (32, 9), '1': (33, 9), '2': (34, 9), '3': (35, 9)}
  cont id=1 sn=33 v=9 round=0  | live-truth-of-NEW-index sn/v=(33, 9)
  cont id=2 sn=34 v=9 round=0  | live-truth-of-NEW-index sn/v=(34, 9)
  cont id=3 sn=35 v=9 round=0  | live-truth-of-NEW-index sn/v=(35, 9)

Code: engine/crates/xerj-api/src/es_compat.rs:20044-20049 (`handles.entry(hit_index.clone()).or_insert_with(|| state.engine.get_index(hit_index).ok())`) + engine/crates/xerj-engine/src/engine.rs:3124-3136 (alias resolution inside get_index).
```

### [should-fix] Index deleted mid-scroll: hits silently lose all three fields while _shards reports failed:0

Answering the lens question directly: `get_index` returns `Err` (engine.rs:3143-3146, `missing_index_error`), the cache stores `None`, `own_index` is `None`, and the bare `.and_then(...)` yields `None` for all three fields. No panic, no stale values, no 404, no partial-failure signal — the hit is still served with `_source` from the snapshot, just with `_seq_no`/`_primary_term`/`_version` gone.

The consumer cannot distinguish this from (a) the deleted-document case, (b) the flags never having been requested, or (c) the pre-PR behaviour. `_shards` still says `{"total":1,"successful":1,"skipped":0,"failed":0}` and `hits.total.value` still counts the vanished index's docs. Contrast page one, which uses `.unwrap_or(hit_idx as u64)` / `.or(Some(1))` (es_compat.rs:13245-13252, 13278-13286) and always emits something.

The node survives — I confirmed `/_cluster/health` still returns 200 after the delete — so this is a silent-wrong-answer issue, not a stability one.

```
Probe N — `da` (1 write round) + `db` (3 rounds), scroll size 2, DELETE /db after page 1:

  _shards={'total': 1, 'successful': 1, 'skipped': 0, 'failed': 0} timed_out=False total={'value': 8, 'relation': 'eq'}
    _index=da id=2 keys=['_id','_index','_primary_term','_score','_seq_no','_source','_version']
    _index=da id=3 keys=['_id','_index','_primary_term','_score','_seq_no','_source','_version']
  _shards={'total': 1, 'successful': 1, 'skipped': 0, 'failed': 0} ...
    _index=db id=0 keys=['_id','_index','_score','_source']
    _index=db id=1 keys=['_id','_index','_score','_source']
  _shards={'total': 1, 'successful': 1, 'skipped': 0, 'failed': 0} ...
    _index=db id=2 keys=['_id','_index','_score','_source']
    _index=db id=3 keys=['_id','_index','_score','_source']

Probe D (4-index variant, DELETE /mi3 mid-scroll):
  [PASS] D deleted index does not panic the node
  -> mi3 continuation hits with fields ABSENT: 4; with fields PRESENT: 0
  [PASS] D node still alive after index delete mid-scroll
```

### [should-fix] Deleted document loses _primary_term too, diverging from page one which always emits Some(1)

Confirms the defect the maintainer flagged as deliberately not fixed, and adds a detail the commit message does not state: because `primary_term` is derived as `sn.map(|_| 1u64)`, it disappears together with `_seq_no`. Page one emits `(Some(sn_or_fallback), Some(1u64))` unconditionally (es_compat.rs:13245-13252), so `_primary_term` is always present there. A doc deleted after the snapshot therefore has `_primary_term` on page one and no `_primary_term` on its continuation page, within the same scroll.

Any client that pairs the two fields (which is the only correct way to use them — ES requires `if_primary_term` alongside `if_seq_no`) breaks on the missing key rather than getting a value it can reject.

```
Probe E — mi1 seeded, scroll size 1, DELETE /mi1/_doc/2 after page 1:

    p0 mi1/0 sn=0 pt=1 v=1
    p1 mi1/1 sn=1 pt=1 v=1
    p2 mi1/2 sn=None pt=None v=None
    p3 mi1/3 sn=3 pt=1 v=1
  [FAIL] E deleted doc keeps _seq_no on its continuation page -> [(2, 'mi1', '2', None, None, None)]

Code: es_compat.rs:20061-20065 `let sn = own_index.and_then(|idx| idx.lookup_seq_no(&h.id)); (sn, sn.map(|_| 1u64))` — no fallback, and primary_term gated on sn. `lookup_seq_no` returns None for a tombstoned entry (index.rs:12637-12648).
```

### [note] _primary_term is a hardcoded 1, not a tracked value — consistent with the engine, but it makes the PR's if_primary_term motivation hollow

Answering the lens question: yes, it is `sn.map(|_| 1u64)`. Empirically constant 1 across every continuation hit regardless of seq_no.

This is not a defect the PR introduces — xerj is single-shard and documents primary_term as fixed at 1 (index.rs:12712-12715: "xerj is single-shard with `primary_term` fixed at 1 — the same constant the hit meta-field emits"), and the write path explicitly discards the client's value (`let _ = if_primary_term;` at index.rs:6679 and index.rs:12968). Main search does the same `Some(1u64)`.

Worth recording because both the PR body and the maintainer's commit justify the change by "these feed straight back into if_seq_no/if_primary_term" — half of that pair is decorative in this engine, and the half that is real is the half shown to be dangerous in finding 1.

```
Probe C against the live node, 4-index scroll:

  distinct _primary_term across all continuation hits: {1}
  distinct _seq_no across all continuation hits: [0,1,2,3,4,5,6,7,8,9,10,11,16,17,18,19]
  [FAIL] C _primary_term is not a hardcoded constant -> _primary_term is always {1}

engine/crates/xerj-engine/src/index.rs:6679  `let _ = if_primary_term;`
engine/crates/xerj-engine/src/index.rs:12968 `let _ = if_primary_term;`
engine/crates/xerj-engine/src/index.rs:12715 `"_primary_term" => Some(Value::Number(1.into())),`
```

### [note] Cache-priming loop allocates a String per hit, not per distinct index — the stated cost control is only half met

The comment claims "cached per distinct index name so a page still costs one `get_index` per index, not one per hit." The `get_index` half holds. But `handles.entry(hit_index.clone())` clones the index name for EVERY hit in the page because `entry` needs an owned key; only the `or_insert_with` body is skipped on a hit. With scroll page sizes up to the 10k cap that is ~10k String allocations per page, essentially all immediately dropped. A `if !handles.contains_key(hit_index)` guard around the insert avoids it.

Separately, the closure at es_compat.rs:20060 shadows the outer hit binding: `.map(|(hit_index, h)| { let own_index = handles.get(hit_index).and_then(|h| h.as_ref()); ... idx.lookup_seq_no(&h.id) ... })`. Behaviour is correct (verified empirically) but the inner `h` is an `Option<Arc<Index>>` and the outer `h` is the Hit, one line apart.

```
engine/crates/xerj-api/src/es_compat.rs:20044-20049

            let mut handles: std::collections::HashMap<String, Option<_>> =
                std::collections::HashMap::new();
            for (hit_index, _) in ctx.hits.iter().skip(position).take(page_size) {
                handles
                    .entry(hit_index.clone())
                    .or_insert_with(|| state.engine.get_index(hit_index).ok());
            }

Not measured — reported as a code-reading observation, not a benchmark.
```


---

## Maintainer response — DO NOT MERGE, and my framing sent this in the wrong direction

Accepted in full. The headline is not the delete case I handed back to @Vinz2168; it is a **silent lost update** that this PR newly creates:

```
page2   _source.round=0  (snapshot)      _seq_no=6 _version=4  (live)
PUT /cas/_doc/1?if_seq_no=6&if_primary_term=1  with the snapshot body  ->  200
cas/1 now round=0 — the concurrent writer's round=99 is gone
```

`scroll_page_response` serves `h.source` out of the frozen `ctx.hits`, then resolves `_seq_no`/`_version` **live** at render time. For a document updated after the snapshot the lookup succeeds and stamps the *current* sequence number onto the *old* body. `if_seq_no` is honoured by the engine, so the CAS passes and the newer revision is destroyed. Before this PR those fields were absent on continuation pages, so no consumer could be misled this way — which makes it a regression, not a pre-existing gap.

It is also precisely the harm I wrote into the #424 commit message and then failed to recognise here:

> worse than the absent field it replaced: `_seq_no` and `_primary_term` exist to be fed back as `if_seq_no`/`if_primary_term`, so a consumer that used to fail loudly now gets a confident wrong value

**My framing pointed at the wrong repair.** I called the delete case a missing `.unwrap_or(...)` fallback and handed it to the contributor as their call. A fallback would *fabricate* a value, and it does nothing about the update case, which is the dangerous one. Both symptoms are one bug: `ScrollContext` freezes `hits` and nothing else, so every meta-field is re-derived against a live index on every page.

The panel established what correct looks like from the reference corpus rather than from memory — a scroll in Lucene pins an `IndexReader`, a point-in-time view whose `_seq_no` doc-values are written at index time and immutable per segment, so ES emits a `_seq_no` that **agrees with the `_source` beside it** and a stale `if_seq_no` correctly 409s.

So the repair is to capture `_seq_no`/`_primary_term`/`_version` into `ctx.hits` at scroll-open time, next to the `_source` that is already captured there. That is engine-shaped work on `ScrollContext`, and it is mine — the composition hazard exists because #424 changed what `_index` means and my per-hit fix completed the path.

Also accepted: the PR wires the flags into `search_with_scroll`'s context but not its hand-built page-one response, so `/{index}/_search_scroll` now omits the fields on page one and emits them on page two — the same class as #428, reintroduced one route over.

@Vinz2168 — your diagnosis of #428 was right and your fix does what it says; a lens confirmed it holds across 2/3/4 indices, mid-page boundaries, wildcards and aliases. The problem is that making the field *present* is only safe once it is *snapshotted*, and nothing in the original issue said so. I am taking that work rather than passing it back.

xerj-org:
## Verification of the snapshot rewrite — REFUTED, 2 must-fix

> REFUTED. The rewrite genuinely closes the previous panel's headline defect — continuation pages now serve the snapshot's `_seq_no` beside the snapshot's `_source`, and replaying it as `if_seq_no` correctly 409s (`round=99` survived; DELETE-after-snapshot and multi-index colliding ids both behave correctly). But the snapshot is NOT a snapshot in the sense claimed, and I reproduced the SAME lost update on a live node. `ScrollHit`'s hits are gathered inside the per-index search loop (es_compat.rs:10562) while its `seq_no`/`version` are resolved AFTERWARDS, per hit, at es_compat.rs:11209-11221 — so a write landing between the two puts a NEW `_seq_no` into the SAME struct as the OLD `_source`. Measured on a clean node: continuation page reported `_seq_no=1787` beside `_source.round=201000` while the live document was `round=201030`; `PUT /ra/_doc/zzhot?if_seq_no=1787&if_primary_term=1` with the stale body returned HTTP 200 and the concurrent writer's `round=201030` is gone. The window is the whole rest of the scroll-opening request and spans every subsequent index's search on a multi-index scroll — exactly the reindex case the branch's own prior commit named. Second, independent must-fix: on an `index.disable_sequence_numbers: true` index, plain `_search` and scroll page one report the `-2`/`0` sentinel while continuation pages now report REAL, usable sequence numbers (`_seq_no: 2`, and `?if_seq_no=2&if_primary_term=1` → HTTP 200). That is a regression this branch introduces (`o.insert("_seq_no"…)` is a `+` line in the diff; origin/main omitted the fields entirely) and no test covers it — `310_sequence_numbers_disabled.yml` has no scroll case. Third, the commit message's claim that page one was rebuilt from the snapshot is only true for `search_with_scroll`; the primary `POST /{index}/_search?scroll=` route still renders page one from live lookups with different fallbacks (`.unwrap_or(hit_idx as u64)` / `.or(Some(1))`, es_compat.rs:13262-13303), which is the mechanism behind the sentinel divergence. And the very function the maintainer rewrote still contains the exact hardcoded `seq_no: Some(0)` / `version: Some(1)` pattern the PR title claims to remove, 60 lines below, live-proven to emit `_seq_no: 0, _version: 1` for a document whose real values are `5`/`5`, unrequested.

### [must-fix] The snapshot is not atomic with the hits — the same lost update reproduces on a live node

`ScrollHit` freezes the hit and the meta-fields at DIFFERENT instants. `merged_hits` is filled inside the per-index search loop (`idx.search(...).await` at es_compat.rs:10440-10562, one iteration per index, sequential), and the meta-fields are only resolved afterwards, per hit, at es_compat.rs:11209-11221 (`let seq_no = own.as_ref().and_then(|i| i.lookup_seq_no(&h.id));`). `search_with_scroll` has the identical gap at 19786-19795, after EVERY index's search has completed. Any write landing in that window pairs a NEW `_seq_no` with the OLD `_source` inside one struct — the exact defect the PR says it closes, narrowed from 'the scroll's lifetime' to 'the duration of the opening request', not eliminated. Making it a true point-in-time view requires the seq_no/version to come from the same read that produced the hit (Lucene's pinned IndexReader + per-segment immutable doc-values), not from a second pass over the version map. Reproduced on the FIRST attempt on both node instances I ran.

```
Setup: index `ra` (1 doc `zzhot`) + index `rb` (8000 docs), `POST /ra,rb/_search?scroll=5m {"size":8000,"seq_no_primary_term":true,"version":true,"sort":[{"_id":"asc"}]}` with a concurrent writer on `ra/zzhot`.

Tear (t3_race.sh, clean node on port 34567):
  open status 200 wall 0.075s writes landed: 981
  continuation page 2: zzhot _seq_no=900 _version=901 _source.round=780
  that _seq_no was minted by the write that set round=900
  TORN SNAPSHOT: _seq_no belongs to round=900 but _source says round=780

Driven to the actual lost update (bounded burst timed inside the window, clean node):
  delay=0.001  _seq_no=1787 source.round=201000  lastwrite=(1787,201030)  <== snapshot seq == LIVE seq, source 30 revs stale
     replay of the stale snapshot body with if_seq_no=1787 -> HTTP 200 ; doc now round=201000 (concurrent writer's round=201030 GONE)
  LOST UPDATE REPRODUCED (clean node) at delay=0.001

Same shape as the defect this head was written to fix. Scripts: /tmp/v431s-snapshot/t3_race.sh, /tmp/v431s-snapshot/t5_lostupdate_race.sh
```

### [must-fix] disable_sequence_numbers: continuation pages now hand out real, usable CAS tokens for an index that declared sequence numbers disabled

The `-2`/`0` sentinel is applied only to `response_body` inside `search_impl` (es_compat.rs:14438-14478). `scroll_page_response` has no equivalent, and now that this branch actually serializes the snapshot values (`o.insert("_seq_no"…)` at es_compat.rs:20185-20191 — a `+` line in this diff, so origin/main emitted nothing there), continuation pages report the real sequence number. Page one and page two of the SAME scroll disagree, page two disagrees with plain `_search` on the same node, and the page-two value is a working optimistic-concurrency token on an index whose whole point is not to have one. This is exactly the harm the maintainer articulated in the #424 commit: 'a consumer that used to fail loudly now gets a confident wrong value.' Nothing in CI catches it — engine/tests/es-compat-yaml/yaml/search/310_sequence_numbers_disabled.yml has no scroll case at all. On the `/{index}/_search_scroll` route (the one whose page one the maintainer rebuilt) BOTH pages are wrong.

```
H=http://127.0.0.1:34567, index created with {"settings":{"index":{"disable_sequence_numbers":true}}}

--- plain _search (expect _seq_no:-2 _primary_term:0) ---
{"_id":"d0","_seq_no":-2,"_primary_term":0,"_version":1}
...
page1: [{"_id":"d0","_seq_no":-2,"_primary_term":0,"_version":1},{"_id":"d1","_seq_no":-2,"_primary_term":0,"_version":1}]
page2: [{"_id":"d2","_seq_no":2,"_primary_term":1,"_version":1},{"_id":"d3","_seq_no":3,"_primary_term":1,"_version":1}]

The page-two token works:
  PUT /nosq/_doc/d2?if_seq_no=2&if_primary_term=1 -> HTTP 200
  {"_index":"nosq","_id":"d2","_version":2,"result":"updated",...,"_seq_no":4,"_primary_term":1}

The /{index}/_search_scroll route is wrong on both pages:
  page1: [{"_id":"d0","_seq_no":0,"_primary_term":1},{"_id":"d1","_seq_no":1,"_primary_term":1}]
  page2: [{"_id":"d2","_seq_no":4,"_primary_term":1},{"_id":"d3","_seq_no":3,"_primary_term":1}]

Regression proof: `git diff origin/main...HEAD -- engine/crates/xerj-api/src/es_compat.rs | grep 'o.insert("_seq_no"' ` → `308:+                            o.insert("_seq_no".to_string(), json!(seq_no));` (addition only).
```

### [should-fix] The commit's claim that page one is rendered from the snapshot is false for the primary scroll route

The commit message says the maintainer 'rebuilt search_with_scroll's page-one render from the same snapshot'. That is true only for `search_with_scroll`. The primary route, `POST /{index}/_search?scroll=…` (`search_impl`), still builds page one from `merged_hits.into_iter()` with LIVE per-hit lookups at es_compat.rs:13262-13303 — a second, later set of lookups than the snapshot at 11209-11221, and with different fallback policy: `.unwrap_or(hit_idx as u64)` for `_seq_no`, `Some(1u64)` for `_primary_term`, `.or(Some(1))` for `_version`, versus the snapshot's plain `Option` with no fallback. So page one and every continuation page of the same scroll are produced by different code with different semantics. This is the mechanism behind the disable_sequence_numbers divergence above, and it also means any hit whose `lookup_seq_no` returns `None` at capture time gets a fabricated `_seq_no = hit_idx` on page one and no field at all on page two — the very `.unwrap_or(hit_idx)` fabrication the maintainer criticised.

```
es_compat.rs:13262-13270 (page one, live):
    let (real_seq_no, real_primary_term) = if emit_seq_no {
        let sn = state.engine.get_index(&idx_name).ok()
            .and_then(|idx| idx.lookup_seq_no(&h.id))
            .unwrap_or(hit_idx as u64);
        (Some(sn), Some(1u64))

es_compat.rs:11212-11218 (snapshot, earlier, no fallback):
    let seq_no = own.as_ref().and_then(|i| i.lookup_seq_no(&h.id));
    ...
    primary_term: seq_no.map(|_| 1u64),

Only `search_with_scroll` (19800-19824) reads `hits_only` for its page one.
```

### [should-fix] The rewritten function still contains the exact hardcoded seq_no: Some(0) / version: Some(1) this PR's title claims to remove

`search_with_scroll`'s non-scroll branch — `POST /{index}/_search_scroll` without `?scroll=` — still builds `EsHit { version: Some(1), seq_no: Some(0), primary_term: Some(1), … }` at es_compat.rs:19918-19923 and serializes them through `EsHit`'s derive, which only skips `None`. So that route emits fabricated `_seq_no`/`_primary_term`/`_version` on every hit to every client whether or not they asked, while plain `_search` correctly omits them. This is the identical defect the PR was opened to fix, sitting 60 lines below the block the maintainer rewrote in this head, in the same function.

```
H=http://127.0.0.1:34567, doc p2 written 5 times.

TRUTH:      [{"_id":"p1","_seq_no":0,"_version":1},{"_id":"p2","_seq_no":5,"_version":5}]
_search:    [{"_id":"p1","_seq_no":null,"_version":null},{"_id":"p2","_seq_no":null,"_version":null}]   (correctly absent)
_search_scroll (no flags, no ?scroll=): [{"_id":"p1","_seq_no":0,"_primary_term":1,"_version":1},{"_id":"p2","_seq_no":0,"_primary_term":1,"_version":1}]

p2's real values are _seq_no 5 / _version 5.
```

### [note] The new flags are body-only: ?seq_no_primary_term=true is silently ignored on /{index}/_search_scroll

`ScrollQueryParams` (es_compat.rs:19646-19649) carries only `scroll`, so the URL form of the flags — which real ES accepts and which `search_impl` explicitly promotes into the body at es_compat.rs:8410-8415 — is dropped on the route this PR newly wired the flags into. A client that sets the flag the URL way gets no `_seq_no` on any page and no error.

```
POST /plain/_search_scroll?scroll=5m&seq_no_primary_term=true  {"size":1,"sort":[{"_id":"asc"}]}
  page1: [{"_id":"p1","_seq_no":null}]
  page2: [{"_id":"p2","_seq_no":null}]

Same flag on the other route works:
  POST /plain/_search?seq_no_primary_term=true → [{"_id":"p1","_seq_no":0},{"_id":"p2","_seq_no":5}]
```

### [note] primary_term is derived, not snapshotted — seq_no.map(|_| 1u64)

`primary_term` is not captured at all; it is synthesised as a constant 1 whenever `seq_no` is `Some`. Defensible on its own terms — xerj is single-shard, every other route hardcodes 1 (es_compat.rs:13272, 15247, 18245, 18331), and the engine discards the value entirely on CAS (`let _ = if_primary_term;`, index.rs:6679) — so it is not the same class of fabrication as page one's `.unwrap_or(hit_idx)`. Two consequences worth naming: `_primary_term` silently disappears whenever `_seq_no` is `None`, and it is emitted as 1 on `disable_sequence_numbers` indices where every other xerj route reports 0.

```
es_compat.rs:11218 and 19793: `primary_term: seq_no.map(|_| 1u64),`
index.rs:6679: `let _ = if_primary_term;` — the CAS compares only seq_no.
Observed on a disable_sequence_numbers index: plain _search `_primary_term: 0`, scroll continuation `_primary_term: 1`.
```

### [note] The snapshot does not honour _source: false — page one omits the body, continuation pages return it in full

`ScrollHit` stores the raw `Hit.source`, and `scroll_page_response` inserts it verbatim (es_compat.rs:20193-20195), whereas page one runs the whole source-policy pipeline at es_compat.rs:11393-11445 (`_source: false`, `_source.enabled: false`, `stored_fields` suppression, meta-marker stripping). Pre-existing rather than introduced here, but it directly falsifies the framing that the snapshot is 'the point-in-time view' of what page one renders: for `_source: false` the two pages of one scroll disagree about whether the document body exists at all. (`includes`/`excludes` do survive, because those are applied in the engine.)

```
### request _source spec: {"size":2,"_source":false,"sort":[{"_id":"asc"}]}
page1: [{"_id":"d0","_source":null},{"_id":"d1","_source":null}]
page2: [{"_id":"d2","_source":{"a":"A2","b":"B2"}},{"_id":"d3","_source":{"a":"A3","b":"B3"}}]
```

### [note] The rewrite deletes the per-index handle cache the previous commit on this same branch added, and now resolves metadata unconditionally

The snapshot loop calls `state.engine.get_index(idx_name)` plus `lookup_seq_no` and `lookup_version` once PER HIT, for up to SCROLL_SNAPSHOT_MAX_HITS = 10_000 hits (es_compat.rs:35037), on EVERY scroll open — including the common case where neither `seq_no_primary_term` nor `version` was requested, since the snapshot is built before the flags are consulted. Commit 2d43f9b on this very branch introduced per-distinct-index handle caching for exactly this and stated the intent ('cached per distinct index name so a page still costs one get_index per index rather than one per hit'); the rewrite dropped it. `get_index` is not free: 2-3 DashMap lookups, an alias probe and a thread-local visibility check per call (engine.rs:3140-3173).

```
es_compat.rs:11209-11213 (and identically 19786-19789):
    for (idx_name, h) in merged_hits.iter() {
        let own = state.engine.get_index(idx_name).ok();
        let seq_no = own.as_ref().and_then(|i| i.lookup_seq_no(&h.id));
        let version = own.as_ref().and_then(|i| i.lookup_version(&h.id));

Guarded only by `if is_scroll_request`, never by `body.seq_no_primary_term` / `body.version`.
```

### [note] What the rewrite does get right (verified, stated for fairness)

Three of the lens's questions came back clean and should not be re-litigated: (1) the previous panel's headline lost update over the scroll's LIFETIME is genuinely fixed — a document updated mid-scroll now returns the snapshot's `_seq_no` beside the snapshot's `_source`, and feeding it back as `if_seq_no` 409s instead of 200; (2) DELETE after the snapshot behaves like ES's pinned reader — the continuation page keeps the snapshot's `_seq_no`/`_primary_term`/`_version` and body rather than dropping the fields, which closes the gap the maintainer had previously handed back to the contributor; (3) a multi-index scroll with colliding ids and different per-index write counts gives every hit ITS index's values, matching per-index ground truth exactly.

```
(1) after updating d2..d5 twice post-open (live _seq_no 10-13, _version 3, round=99):
  page 2: {"_id":"d2","_seq_no":2,"_version":1,"body":"v1","round":0}
  PUT /cas/_doc/d2?if_seq_no=2&if_primary_term=1 -> HTTP 409
  final state: d2..d5 round=99 body=v3  (concurrent writer intact)

(2) after DELETE of d2..d5 post-open:
  page2: [{"_id":"d2","_seq_no":2,"_primary_term":1,"_version":1,"body":"v1"}, ...]

(3) truth mia 0..3 = seq 0,1,2,3 / ver 1; mib 0..3 = seq 2,6,11,17 / ver 3,4,5,6
  scroll pages: {"_index":"mia","_id":"1","_seq_no":1,"_version":1},{"_index":"mib","_id":"1","_seq_no":6,"_version":4} — all 8 match
```


---

## Maintainer response — my fix narrows the window, it does not close it

Accepted. Both findings stand, and the first is a correction to a claim I made in the commit message.

I wrote that `ScrollHit` captures the meta-fields "when the scroll opens", as though that were a single instant. It is not. `merged_hits` is filled inside the per-index search loop and the meta-fields are resolved afterwards in a second pass over the version map, so a write landing between them puts a new `_seq_no` in the same struct as the old `_source`. The window went from *the scroll's lifetime* to *the duration of the opening request* — narrower, and on a multi-index scroll it still spans every subsequent index's search. Reproduced on the first attempt:

```
continuation: zzhot _seq_no=1787  _source.round=201000
live:                              round=201030
PUT /ra/_doc/zzhot?if_seq_no=1787&if_primary_term=1  with the stale body -> 200
the concurrent writer's round=201030 is gone
```

That is the same lost update, smaller. I should have written "narrows" and did not, and the lens caught the overclaim by building the race rather than reading the prose.

**The real repair is one level down.** The value has to come from the read that produced the hit, not from a second pass — which is what Lucene gets for free, because a searcher reads `_seq_no` doc-values out of the same pinned, immutable segment it read the document from. In xerj that means `Hit` carrying `seq_no`/`version` populated during `idx.search(...)`, not `ScrollHit` reconstructing them afterwards. That is engine work, it is mine, and I am filing it rather than attempting a third patch at the API layer — two have now been refuted, both because I was fixing the symptom one layer above the cause.

**The second finding is a defect my change introduced.** `disable_sequence_numbers` applies its `-2`/`0` sentinel only in `search_impl`'s response body. `scroll_page_response` has no equivalent, and because this branch now actually serializes the values, continuation pages hand out a *working* CAS token for an index whose entire purpose is not to have one — page one says `-2`, page two says `2`, and `if_seq_no=2` succeeds. On `/{index}/_search_scroll` both pages are wrong. `origin/main` emitted nothing there, so this is new, and the conformance suite cannot see it: `310_sequence_numbers_disabled.yml` has no scroll case.

That one is small and belongs on this branch. The atomicity fix does not.

Marking DO NOT MERGE. @Vinz2168 — your original diagnosis remains correct and untouched by any of this; what keeps failing is my repairs to it.

xerj-org:
Converting this to a draft, and the reason is a hazard on our side rather than anything wrong with the work.

This PR has been sitting at **16 checks green and `mergeStateStatus CLEAN`** for a long time while carrying `DO NOT MERGE` in its title. Three PRs in this repo have already been merged out-of-band while their verification was still running — #432/#429, #425/#430 and #427, the last of which shipped a classification regression that is still being fixed forward in #442. A refuted PR that presents as fully green is exactly the shape that keeps happening, and a title is not a gate. Draft status is.

**The work here is not the problem.** Two rounds of repair at the API layer were each refuted for the same structural reason, and neither refutation is about this implementation: the meta-fields are resolved in a second pass over the live version map, after `merged_hits` was filled inside the per-index search loop. Any write between those two reads pairs a new sequence number with an old body, and feeding that `_seq_no` back as `if_seq_no` passes the CAS and silently overwrites what landed in between. Round 1 made the window the whole scroll lifetime; round 2 narrowed it to the opening request, which on a multi-index scroll still spans every subsequent index search. There is no ordering the API layer can impose that removes the gap, because the two reads are of different structures at different times.

Closing it properly needs #440: `xerj_query::executor::Hit` has to carry `seq_no`/`version` populated inside `Index::search`, from the same read that produced the body — which is what Lucene gets structurally from a pinned `IndexReader` and immutable per-segment doc-values. Once that exists, `ScrollHit` becomes redundant and plain `_search` gets the same guarantee, which it does not have today either.

I am picking up #440. When it lands, this branch should become much smaller — most of it will no longer be needed — and I would rather it be reopened against that foundation than merged onto one that cannot hold it. Marked draft so nothing merges it by accident in the meantime; happy to reopen for review the moment #440 is in.

Vinz2168:
Superseded by #499, which fixes #440 (this PR's blocker) directly: `Hit` now carries `seq_no`/`version` populated at the same read as `_source` inside `Index::search`, rather than a second pass at any point in the pipeline. That closes the torn-read window this PR's two repair attempts each narrowed but didn't eliminate, and picks up the original #428 scroll-continuation fix along the way (still not merged here). Closing without merging, per the title.

Issue #385, opened 2026-08-15
xerj-autoindex: the global SNAPSHOT_FAILPOINT is stolen across test modules; 5/10 runs fail at default parallelism


xerj-org:
Found while fixing #372 by an agent (Claude Opus 5) driven by @xerj-org, then measured over
ten runs. Same defect class as #372, different crate: real on a clean checkout, invisible to
CI because CI runs `--test-threads=2`.

## Reproduction

```
$ cd engine
$ cargo test --profile ci-test -p xerj-autoindex --lib
```

Ten consecutive runs, 20-core box, rustc 1.92.0, aarch64-unknown-linux-gnu:

```
=== run 1: test result: ok. 633 passed; 0 failed; ... finished in 16.38s
=== run 2: test result: ok. 633 passed; 0 failed; ... finished in 12.26s
=== run 3: test result: FAILED. 631 passed; 2 failed; ... finished in 13.55s
    incremental_reconcile_http_tests::initial_generation_and_noop_are_end_to_end_idempotent
    sync_executor::tests::every_snapshot_crash_boundary_restarts_deterministically
=== run 4: test result: FAILED. 631 passed; 2 failed; ... finished in 13.31s
    incremental_reconcile_http_tests::genesis_recovers_after_snapshot_budget_failure_before_sync_begin
    sync_executor::tests::every_snapshot_crash_boundary_restarts_deterministically
=== run 5: test result: FAILED. 631 passed; 2 failed; ... finished in 15.00s
    incremental_reconcile_http_tests::no_semantic_generation_does_not_require_embedding_identity_endpoint
    sync_executor::tests::every_snapshot_crash_boundary_restarts_deterministically
=== run 6: test result: ok. 633 passed; 0 failed; ... finished in 14.76s
=== run 7: test result: ok. 633 passed; 0 failed; ... finished in 13.43s
=== run 8: test result: FAILED. 631 passed; 2 failed; ... finished in 13.96s
    incremental_reconcile_http_tests::initial_generation_and_noop_are_end_to_end_idempotent
    sync_executor::tests::every_snapshot_crash_boundary_restarts_deterministically
=== run 9: test result: FAILED. 631 passed; 2 failed; ... finished in 11.93s
    incremental_reconcile_http_tests::issue_283_inferred_float_manifest_digest_replays_on_unchanged_rerun
    sync_executor::tests::every_snapshot_crash_boundary_restarts_deterministically
[...]

PR #445, opened 2026-08-17
test(esclient): assert the retry property as a count, not a 260ms stopwatch

### LINKED ISSUE #436: esclient retry test asserts a 260ms wall-clock bound and flakes ~1 run in 5 on main — will intermittently redden unrelated PRs
Found while rebasing #425; it is unrelated to that PR and reproduces on `main`.

## What happens

`esclient::tests::all_timeout_retry_path_is_bounded_to_six_attempts_without_final_sleep` asserts a wall-clock bound:

```
engine/crates/xerj-autoindex/src/esclient.rs:1153
    assert!(elapsed < Duration::from_millis(260))
observed: 265.98ms
```

It fails roughly **1 run in 5** on an otherwise idle box, and fails identically on `origin/main` — so it is not a regression from any open PR.

## Why it matters more than a single flaky test

CI runs this crate at `--test-threads=2` (`ci.yml`), where the timing is tighter, so this will intermittently redden **unrelated** pull requests. That is the expensive kind of flake: the failure appears on someone else's change, and the natural reading is "my change broke it."

This repo has already paid for that pattern twice — #372 and #385 were both cross-test interference that CI's `--test-threads=2` either hid or caused, and #422 was a red `main` that failed every open PR on code none of them touched. A contributor who hits this on their first submission has no way to know it is ours.

## Shape of a fix

A wall-clock assertion on a retry path is measuring the machine, not the code. The property worth pinning is that the retry loop makes **six attempts and does not sleep after the final one** — countable without a stopwatch, e.g. by injecting a clock or counting sleep calls rather than timing the whole path. If a bound must stay, it needs enough headroom to survive a loaded runner, and the current 260 ms is 2% away from the observed value.

**Verified:** the failure and its ~20% rate were measured on this machine, and reproduced on `origin/main` as well as on the #425 branch.
**Not verified:** the rate under CI's actual runner and `--test-threads=2`, which is where it matters most — I measured locally at default parallelism.

xerj-org:
Closes #436.

## I could not reproduce the reported flake

Stating this first because it bounds everything else. #436 reports ~1 failure in 5 on `main`. On this machine:

```
main's stopwatch version, 40 runs, --test-threads=2      0 failures
main's stopwatch version, 20 runs, under 12 CPU hogs     0 failures
```

So the 20% rate is unconfirmed here and I do not know what produces it. If it is real on the CI runner, this change fixes it; if the reporter's machine had something specific about it, that is still unexplained and worth knowing.

## Why change it anyway

The test is named `..._bounded_to_six_attempts_without_final_sleep`. That property is a **count** — six attempts, five backoffs, none after the last — and it was being inferred from a duration: six 25 ms attempts plus 10+20+20+20+20 ms of backoff is a 240 ms budget, asserted under a 260 ms bound. Twenty milliseconds of headroom.

A timing proxy for a countable property fails on a slow machine and passes on a fast one regardless of whether the code is correct. That is the wrong direction for a CI gate, and it is why the reporter saw 265.98 ms — the assertion was reporting on the runner, not on the retry loop.

The client now counts its own backoff sleeps and the test asserts five.

## The counter is per-instance on purpose

`#[cfg(test)]`, on the `Es` struct, not a global static. CI runs this crate at `--test-threads=2`; a shared counter would have introduced exactly the cross-test interference that #372 and #385 already cost this repo, while fixing a different flake. 

## Fail-before

Moving the sleep back inside the loop unconditionally, so a sixth one happens:

```
assertion `left == right` failed: six attempts must sleep five times, never after the final failure
  left: 6
 right: 5
```

Restored, it passes. Note the old assertion could **not** have caught this on a fast box: a sixth 20 ms sleep lands at exactly 260 ms, the bound itself.

## Verification

`rustup run 1.97.1` (the CI toolchain, not the 1.92.0 local default): fmt clean, `clippy --workspace --all-targets -- -D warnings` clean, tests green on both the dev and ci-test profiles.

Not verified: the flake rate on the actual CI runner, which is where the issue says it matters most and where I cannot measure.

### DIFF
--- a/engine/crates/xerj-autoindex/src/esclient.rs
+++ b/engine/crates/xerj-autoindex/src/esclient.rs
@@ -540,21 +540,37 @@ impl Es {
         Err(last_err.unwrap_or_else(|| anyhow!("{what}: retries exhausted")))
     }
 
-    /// Sleep between attempts, recording what was slept for.
+    /// Sleep between attempts, recording how long was actually slept.
     ///
-    /// One function so that recording and sleeping cannot come apart. An
-    /// earlier version incremented a counter next to a bare
-    /// `std::thread::sleep`, and moving only the sleep out of its guard then
-    /// produced a sixth backoff that the counter still reported as five — the
-    /// exact regression the test is named for, invisible to it. Bypassing this
-    /// now means writing a second sleep call rather than moving a line.
+    /// One function so that moving the CALL SITE cannot separate recording from
+    /// sleeping: an earlier version incremented a counter next to a bare
+    /// `std::thread::sleep`, and moving only the sleep out of its guard produced
+    /// a sixth backoff the counter still reported as five.
+    ///
+    /// What this does NOT defend against, stated because the previous version of
+    /// this comment overclaimed it: a sleep that never routes through here is
+    /// invisible to the recorded sequence. Mutation testing found three —
+    /// a bare `thread::sleep` after the loop, the sleep moved out of `backoff`
+    /// itself while the push stays, and a second sleep alongside the recorded
+    /// one. Each adds 20-90 ms of real sleeping that no assertion over recorded
+    /// data can see. Only a tight wall-clock upper bound catches them, and a
+    /// tight upper bound is precisely what failed one run in five under load
+    /// (#436). That is a trade, not an oversight: the assertions pin the shape
+    /// of the sleeping that goes through here, and wall time pins that some
+    /// sleeping happened at all.
+    ///
+    /// The recorded value is the OBSERVED duration rather than the requested
+    /// one, so a `Duration` that is computed correctly and then not honoured
+    /// shows up as a short entry instead of a correct-looking one.
     fn backoff(&self, delay: Duration) {
+        #[cfg(test)]
+        let started = Instant::now();
+        std::thread::sleep(delay);
         #[cfg(test)]
         self.backoff_delays
             .lock()
             .expect("backoff_delays poisoned")
-            .push(delay);
-        std::thread::sleep(delay);
+            .push(started.elapsed());
     }
 
     /// PUT index with explicit mapping; tolerates already-exists.
@@ -1189,6 +1205,8 @@ mod tests {
         // on `main`, so it reddened pull requests that had not touched it
         // (#436). A sixth sleep is what the old bound was really looking for,
         // and it is visible here directly.
+        // Observed durations, so they round UP off the requested value; compare
+        // the floor at millisecond granularity rather than the exact number.
         let delays: Vec<u64> = es
             .backoff_delays
             .lock()
@@ -1197,10 +1215,20 @@ mod tests {
             .map(|d| d.as_millis() as u64)
             .collect();
         assert_eq!(
-            delays,
-            vec![10, 20, 20, 20, 20],
-            "the backoff sequence pins arity, ordering, doubling and the \
-             retry_max_delay cap together; a count alone pins only the arity"
+            delays.len(),
+            5,
+            "six attempts must back off five times, never after the final \
+             failure: {delays:?}"
+        );
+        let requested = [10u64, 20, 20, 20, 20];
+        assert!(
+            delays
+                .iter()
+                .zip(requested)
+                .all(|(observed, want)| *observed >= want && *observed < want + 50),
+            "each backoff must sleep for the delay it was given — arity, \
+             ordering, the doubling and the retry_max_delay cap all live in \
+             this sequence: got {delays:?}, wanted {requested:?}"
         );
         // Timing, but only in the direction that is safe to assert. A sleep
         // can overrun and cannot underrun, so a LOWER bound on the summed
--- a/engine/crates/xerj-autoindex/src/esclient.rs
+++ b/engine/crates/xerj-autoindex/src/esclient.rs
@@ -22,6 +22,17 @@ pub struct Es {
     /// property of the server, so all of a run's workers must see the same
     /// admission limit.
     admission: Arc<BulkAdmission>,
+    /// Backoff sleeps this client has performed. Test-only, and per-instance
+    /// rather than global so a concurrent test cannot perturb the count —
+    /// `--test-threads=2` in CI is exactly the shape that would.
+    ///
+    /// It exists because the property worth pinning on the retry path is "five
+    /// backoffs between six attempts, and none after the last", which is a
+    /// count. Asserting it with a stopwatch measured the machine instead: the
+    /// old bound sat 20 ms above a 240 ms budget and failed about one run in
+    /// five (#436).
+    #[cfg(test)]
+    backoff_sleeps: Arc<std::sync::atomic::AtomicUsize>,
 }
 
 /// How many bulk requests a run may have in flight, and how a 429 changes
@@ -335,6 +346,8 @@ impl Es {
             retry_initial_delay,
             retry_max_delay,
             admission: Arc::new(BulkAdmission::off()),
+            #[cfg(test)]
+            backoff_sleeps: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
         })
     }
 
@@ -516,6 +529,9 @@ impl Es {
                 Err(e) => last_err = Some(e),
             }
             if attempt + 1 < MAX_ATTEMPTS {
+                #[cfg(test)]
+                self.backoff_sleeps
+                    .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
                 std::thread::sleep(delay);
                 delay = (delay * 2).min(self.retry_max_delay);
             }
@@ -1140,17 +1156,24 @@ mod tests {
             Duration::from_millis(20),
         )
         .unwrap();
-        let started = Instant::now();
         let error = match es.bulk(b"{}\n".to_vec()) {
             Ok(_) => panic!("all six delayed responses unexpectedly succeeded"),
             Err(error) => error,
         };
-        let elapsed = started.elapsed();
         assert!(format!("{error:#}").contains("timed out"), "{error:#}");
         assert_eq!(*accepted.lock().unwrap(), 6);
-        // Six 25ms attempts plus 10+20+20+20+20ms backoffs. A sixth
-        // post-failure sleep would push this past the deliberately tight cap.
-        assert!(elapsed < Duration::from_millis(260), "{elapsed:?}");
+        // The property is a COUNT, not a duration: six attempts with five
+        // backoffs between them and none after the last. This used to be
+        // asserted as `elapsed < 260ms` against a 240ms budget, which measured
+        // the machine rather than the code and failed about one run in five —
+        // on `main`, so it reddened pull requests that had not touched it
+        // (#436). A sixth sleep is what the old bound was really looking for,
+        // and it is visible here directly.
+        assert_eq!(
+            es.backoff_sleeps.load(std::sync::atomic::Ordering::Relaxed),
+            5,
+            "six attempts must sleep five times, never after the final failure"
+        );
         server.join().unwrap();
     }
 
--- a/engine/crates/xerj-autoindex/src/esclient.rs
+++ b/engine/crates/xerj-autoindex/src/esclient.rs
@@ -22,17 +22,21 @@ pub struct Es {
     /// property of the server, so all of a run's workers must see the same
     /// admission limit.
     admission: Arc<BulkAdmission>,
-    /// Backoff sleeps this client has performed. Test-only, and per-instance
-    /// rather than global so a concurrent test cannot perturb the count —
-    /// `--test-threads=2` in CI is exactly the shape that would.
+    /// Every delay this client has actually backed off for, in order.
+    /// Test-only, and per-instance rather than global so a concurrent test
+    /// cannot perturb it — `--test-threads=2` in CI is exactly the shape that
+    /// would.
     ///
-    /// It exists because the property worth pinning on the retry path is "five
-    /// backoffs between six attempts, and none after the last", which is a
-    /// count. Asserting it with a stopwatch measured the machine instead: the
-    /// old bound sat 20 ms above a 240 ms budget and failed about one run in
-    /// five (#436).
+    /// The DELAYS and not a count. Counting alone pins the arity and nothing
+    /// else, so removing the doubling, ignoring `retry_max_delay`, or deleting
+    /// the sleep outright while still counting it all pass — the last turning
+    /// this into a six-shot hot loop against a struggling server, which is the
+    /// retry storm the doc comment on `with_retry` exists to prevent. The
+    /// sequence pins the arity, the ordering, the doubling and the cap at once.
+    /// It replaces a wall-clock bound that sat 20 ms above a 240 ms budget and
+    /// failed about one run in five under load (#436).
     #[cfg(test)]
-    backoff_sleeps: Arc<std::sync::atomic::AtomicUsize>,
+    backoff_delays: Arc<std::sync::Mutex<Vec<Duration>>>,
 }
 
 /// How many bulk requests a run may have in flight, and how a 429 changes
@@ -347,7 +351,7 @@ impl Es {
             retry_max_delay,
             admission: Arc::new(BulkAdmission::off()),
             #[cfg(test)]
-            backoff_sleeps: Arc::new(std::sync::atomic::AtomicUsize::new(0)),
+            backoff_delays: Arc::new(std::sync::Mutex::new(Vec::new())),
         })
     }
 
@@ -529,14 +533,28 @@ impl Es {
                 Err(e) => last_err = Some(e),
             }
             if attempt + 1 < MAX_ATTEMPTS {
-                #[cfg(test)]
-                self.backoff_sleeps
-                    .fetch_add(1, std::sync::atomic::Ordering::Relaxed);
-                std::thread::sleep(delay);
+                self.backoff(delay);
                 delay = (delay * 2).min(self.retry_max_delay);
             }
         }
         Err(last_err.unwrap_or_else(|| anyhow!("{what}: retries exhausted")))
+    }
+
+    /// Sleep between attempts, recording what was slept for.
+    ///
+    /// One function so that recording and sleeping cannot come apart. An
+    /// earlier version incremented a counter next to a bare
+    /// `std::thread::sleep`, and moving only the sleep out of its guard then
+    /// produced a sixth backoff that the counter still reported as five — the
+    /// exact regression the test is named for, invisible to it. Bypassing this
+    /// now means writing a second sleep call rather than moving a line.
+    fn backoff(&self, delay: Duration) {
+        #[cfg(test)]
+        self.backoff_delays
+            .lock()
+            .expect("backoff_delays poisoned")
+            .push(delay);
+        std::thread::sleep(delay);
     }
 
     /// PUT index with explicit mapping; tolerates already-exists.
@@ -1156,10 +1174,12 @@ mod tests {
             Duration::from_millis(20),
         )
         .unwrap();
+        let started = Instant::now();
         let error = match es.bulk(b"{}\n".to_vec()) {
             Ok(_) => panic!("all six delayed responses unexpectedly succeeded"),
             Err(error) => error,
         };
+        let elapsed = started.elapsed();
         assert!(format!("{error:#}").contains("timed out"), "{error:#}");
         assert_eq!(*accepted.lock().unwrap(), 6);
         // The property is a COUNT, not a duration: six attempts with five
@@ -1169,11 +1189,38 @@ mod tests {
         // on `main`, so it reddened pull requests that had not touched it
         // (#436). A sixth sleep is what the old bound was really looking for,
         // and it is visible here directly.
+        let delays: Vec<u64> = es
+            .backoff_delays
+            .lock()
+            .unwrap()
+            .iter()
+            .map(|d| d.as_millis() as u64)
+            .collect();
         assert_eq!(
-            es.backoff_sleeps.load(std::sync::atomic::Ordering::Relaxed),
-            5,
-            "six attempts must sleep five times, never after the final failure"
+            delays,
+            vec![10, 20, 20, 20, 20],
+            "the backoff sequence pins arity, ordering, doubling and the \
+             retry_max_delay cap together; a count alone pins only the arity"
         );
+        // Timing, but only in the direction that is safe to assert. A sleep
+        // can overrun and cannot underrun, so a LOWER bound on the summed
+        // delays is immune to a loaded machine — which is what made the old
+        // upper bound flaky. Without it, deleting the sleep from `backoff`
+        // while still recording the delay passes everything above and turns the
+        // retry path into a six-shot hot loop against a struggling server.
+        // 6 x 25ms of request timeout + 90ms of summed backoff = 240ms at an
+        // absolute minimum; 200 leaves 40ms of slack DOWNWARD for timer
+        // granularity. Load can only push elapsed up, never down, which is what
+        // makes a lower bound safe where the old upper bound was not.
+        assert!(
+            elapsed >= Duration::from_millis(200),
+            "only {elapsed:?} elapsed against {}ms of recorded backoff — the \
+             delays were recorded without being slept",
+            delays.iter().sum::<u64>()
+        );
+        // And a runaway net, deliberately loose: the flake this replaced came
+        // from a 19 ms margin, not from having an upper bound at all.
+        assert!(elapsed < Duration::from_secs(2), "{elapsed:?}");
         server.join().unwrap();
     }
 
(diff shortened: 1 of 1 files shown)


xerj-org:

0 of 1 refuted.

> The head's test is a large real improvement over both the stopwatch and the count, and the 200ms lower bound is SAFE — but three mutants survive, including the exact regression the test is named for.
> 
> MUTATION RESULTS (all built with `rustup run 1.97.1 ... --profile ci-test`, each mutant confirmed live in the binary by sha + measured `elapsed`; a build failure was treated as no result, none occurred):
> 
> KILLED (8): backoff call moved out of the attempt guard -> delays [10,20,20,20,20,20]; sleep deleted while still recording -> elapsed 150ms < 200ms; doubling removed -> [10,10,10,10,10]; retry_max_delay ignored -> [10,20,40,80,160]; MAX_ATTEMPTS=3 and =5 -> accepted-count assert; MAX_ATTEMPTS=7 -> six delays; backoff only on 429 -> delays []; recording moved after the doubling -> [20,20,20,20,20].
> 
> SURVIVED (3), all one root cause — the recorder is advisory, so any sleep that does not go through it is invisible and the 2s upper bound is far too loose to see it:
>  - final bare `thread::sleep(delay)` after the loop (the sixth post-failure sleep the test is NAMED for): elapsed 261ms, passes.
>  - recording kept in `backoff`, the sleep moved out of `backoff` and out of the guard (6 sleeps, 5 recorded): elapsed 265-276ms, passes.
>  - extra bare sleep alongside the recorded one (every backoff sleeps twice): elapsed 334ms, passes.
> The doc comment's claim that "recording and sleeping cannot come apart" holds only against moving the CALL SITE; editing `backoff` itself pulls them apart undetected.
> 
> LOWER BOUND IS SAFE. Analytic floor is 240ms (6 x 25ms bulk_timeout + 90ms backoff) and the 200ms bound sits 40ms below it. No mechanism shortens it: `thread::sleep` and tokio timers round up, never fire early, and this fixture never answers, so all six attempts are full timeouts. Empirically min elapsed was 241.0ms across ~590 runs (idle + loaded); 550/550 passes under 96 and 200 spinners on 20 cores, 0.0% failure.
> 
> LOAD IS REAL (method control): I rebuilt the SAME test with the old `elapsed < 260ms` bound re-armed and ran it under identical load — 38/150 failed, 25.3%, every failure the old bound, matching the PR's "one run in five" (#436). So the 0% figure for the head is not an artifact of insufficient load.

**[should-fix] Three mutants survive: the recorder is advisory, so a bare sleep that bypasses it is invisible — including the sixth post-failure sleep the test is named for**

`backoff()` records the delay it was ASKED to sleep, not the time actually slept, and nothing ties the recorded sequence to `elapsed` from above. The sequence assert therefore pins arity/order/doubling/cap only for sleeping that routes through `backoff`, and the 2s upper bound cannot see 20-90ms of extra sleeping. Three mutants pass every assertion:

1. `final_bare_sleep_after_loop` — insert `std::thread::sleep(delay);` immediately before `Err(last_err.unwrap_or_else(...))`. This IS the retry storm the test name (`..._without_final_sleep`) claims to exclude, and the removed 260ms bound did catch it (261ms > 260ms, marginally). It now passes.
2. `recording_kept_sleep_out_of_guard` — leave the push in `backoff`, move `std::thread::sleep` out of `backoff` and out of the guard. Six sleeps, five recorded. This is precisely the regression the new doc comment says is now caught; it is not.
3. `extra_bare_sleep_in_guard` — keep `self.backoff(delay)` and add a second bare `std::thread::sleep(delay)` next to it. Every backoff sleeps twice.

The PR partly acknowledges #3 ('Bypassing this now means writing a second sleep call rather than moving a line'), but #1 and #2 are the named property, not a bypass footnote. Cheap remedies, in increasing strength: (a) record the OBSERVED sleep (`Instant::now()` around the sleep inside `backoff`) so a doubled or bypassing sleep shows up as a wrong duration; (b) route all sleeping through an injected `Sleeper` the test can substitute, making total slept time exactly known instead of timing-derived; (c) add `std::thread::sleep` to `disallowed-methods` in a clippy.toml (the repo has none today — `grep -rn disallowed_methods engine/` is empty) with a single allow inside `backoff`, turning bypass into a build error rather than a convention.

```
Each survivor verified live in the binary (distinct sha256, crate recompiled) and by the elapsed value the test itself measured, printed via an observation-only `eprintln!` that changed no assertion:
  baseline                          pass 8/8  elapsed 241 241 241 241 241 241 241 241 ms
  final_bare_sleep_after_loop       pass 8/8  elapsed 262 261 261 262 261 261 262 261 ms
  recording_kept_sleep_out_of_guard pass 8/8  elapsed 266 272 275 265 268 276 269 271 ms
  extra_bare_sleep_in_guard         pass 8/8  elapsed 334 334 334 334 337 335 331 333 ms
Ruling out the wrong reason: process wall time is NOT usable here (baseline and the +20ms mutants all measure 322ms) because `server.join()` waits on a 6th handler that sleeps 100ms past the client's last timeout and masks the delta; the +90ms mutant does show through at 412ms. Hence the in-test elapsed probe above. Mutants that fail to compile were reported as BUILD_FAIL and counted as no result; none of these did.
Source: /tmp/wr-retryflake/repo/engine/crates/xerj-autoindex/src/esclient.rs lines 535 (`self.backoff(delay)`), 551-558 (`fn backoff`), 1214-1224 (bounds).
```

**[nit] The 200ms floor silently depends on the fixture never answering; if the server is ever made to reply, it fires on correct code with a message that blames the sleep**

The 240ms floor is 90ms of backoff plus 6 x 25ms of bulk_timeout. The backoff term is a property of the code; the 150ms timeout term is a property of THIS fixture, whose handlers sleep 100ms and never write a response. Reuse the same shape for a 5xx-retry test (handler replies 503 immediately — a natural next test, and `with_retry` treats 5xx identically) and the floor collapses to ~90ms, so `elapsed >= 200ms` fails on entirely correct code while printing 'the delays were recorded without being slept'. Note the coupling cannot simply be removed: `elapsed >= delays.sum()` (90ms) would be fixture-independent but lets `sleep_deleted` (150ms) through. Worth one sentence in the comment naming the 6 x bulk_timeout term as a fixture dependency, and a debug_assert or comment at the fixture that it must not respond.

```
Test fixture at esclient.rs:1150-1176: each of the 6 handlers does `read_request(&mut stream); std::thread::sleep(Duration::from_millis(100));` and never writes; client bulk_timeout is `Duration::from_millis(25)`. The comment at esclient.rs:1211 states the arithmetic ('6 x 25ms of request timeout + 90ms of summed backoff = 240ms') but does not flag that the first term is the fixture's behaviour, not the retry loop's.
```

**[nit] The lower bound's detection power (not its safety) degrades under load: the deleted-sleep mutant escaped once in 300 loaded runs**

This is a false-negative window, not a flake — the head test never produced a false failure in 550 loaded runs. But the `sleep_deleted` mutant floors at 150ms and load pushes it toward the 200ms bound, so the assertion that exists specifically to catch it is the one assertion whose margin load erodes. It is 50ms of headroom against a distribution whose p99 already reaches 181-187ms. Raising the bound is not free (it eats the 41ms of downward slack that makes it safe); recording the observed sleep duration inside `backoff`, as suggested in the first finding, removes the dependence on wall time for this property entirely.

```
sleep_deleted @ 96 spinners, 150 runs: pass=1 fail=149 (99.3% killed), elapsed_ms min=150.4 p50=158.1 p90=166.0 p99=181.1 max=202.5 — the single pass is a genuine escape (202.5ms >= 200ms, sequence [10,20,20,20,20] still recorded, 'test result: ok').
sleep_deleted @ 200 spinners, 150 runs: pass=0 fail=150, max=190.1.
Contrast head test: @96 spinners 300/300 pass (min 241.3 max 275.5); @200 spinners 250/250 pass (min 241.7 max 283.0).
```

xerj-org:
## Independent verification — verbatim

The core change is confirmed a large improvement — under 20x oversubscription main's old bound fails 61/100 and this head fails **0/219** — but the trade is **not stated accurately**, and one claim in my own comment is simply wrong.

### Verdict — refuted flag: `True`

> The core change is a real and large improvement: under 20x CPU oversubscription main's `elapsed < 260ms` assertion fails 61/100 while the PR head fails 0/219, and 8 of the 10 mutants I built are killed. But the trade is NOT stated accurately. The `backoff()` doc comment lists three surviving mutants; one — "the sleep moved out of `backoff` itself while the push stays" — is in fact KILLED by the very observed-duration change the same commit introduces (built two ways, both record [0,0,0,0,0], both die). The test comment justifying the 200ms lower bound ("deleting the sleep from `backoff` while still recording the delay passes everything above") is likewise stale: that mutant now dies at the sequence assertion, not the lower bound. And "Timing, but only in the direction that is safe to assert" sits eight lines below `*observed < want + 50`, which is a wall-clock UPPER bound and is load-bearing (it is the only thing that kills cap-ignored). On the headline question: the +50ms slack held here — 0 exceedances in ~4,000 recorded sleeps at 4x/10x/20x — but max overrun was 29.8ms, 60% of the budget, at only 4x on a 20-core box, so it is not comfortable on a small shared runner. It is also unnecessary: I built and measured a variant recording (requested, observed) pairs that kills the identical mutant set with no upper timing bound at all. Separately, load testing surfaced a residual failure mode the PR's verification missed: the test hangs forever (19/219 at 20x, 2/150 at 10x) in `server.join()` because the fixture's `read_request` spins on `Ok(0)` when the client drops the socket. Pre-existing (main hangs 6/100 too) but unfixed, and now the dominant way this test fails under load. fmt, clippy -D warnings, and the full 660-test suite at --test-threads=2 are green on 1.97.1.

**[must-fix] Doc comment names a surviving mutant that the same commit actually kills**

The `backoff()` doc comment at engine/crates/xerj-autoindex/src/esclient.rs:551-573 says three mutants survive, one being "the sleep moved out of `backoff` itself while the push stays". That was true when `backoff` pushed the REQUESTED delay (commit 076bda1). Commit 848ed4e changed the push to `started.elapsed()`, which makes any variant where `backoff` no longer sleeps record ~0 and die at the sequence assertion. I built it two ways and both are killed. Since the entire stated purpose of this commit is "stop claiming what the recorder cannot see", shipping a survivor list that was not re-measured after the change is the same defect the commit is fixing, pointed the other way. The list should read: a bare sleep after the loop, and a second sleep alongside the recorded one — both of which I confirmed do survive.

```
Mutant A (sleep moved to the call site, `push(started.elapsed())` unchanged in backoff): KILLED — "got [0, 0, 0, 0, 0], wanted [10, 20, 20, 20, 20]" at esclient.rs:1225.
Mutant B (sleep hoisted above `let started = Instant::now()` inside backoff, push unchanged): KILLED — same message at esclient.rs:1224.
Confirmed survivors, as documented: bare `thread::sleep(90ms)` after the loop → PASS; second `thread::sleep(delay)` after the push in backoff → PASS.
```

**[should-fix] `*observed < want + 50` is a wall-clock upper bound, and the comments above it say there isn't one**

The test comment reads "Timing, but only in the direction that is safe to assert. A sleep can overrun and cannot underrun, so a LOWER bound ... is immune to a loaded machine", and the `backoff()` doc says "Only a tight wall-clock upper bound catches them, and a tight upper bound is precisely what failed one run in five". Both sit next to `*observed < want + 50`, which is exactly a wall-clock upper bound, applied five times per run. It is load-bearing, not decorative: removing it lets cap-ignored survive. It survived my load testing, but with only 40% margin left at 4x oversubscription on a 20-core box, and CI runners are smaller and noisier than this one. The framing should either admit the upper bound or remove the need for it (next finding).

```
~4,000 recorded sleeps on 20 cores: 4x (80 spinners, 1000 samples) max overrun 29.8ms, p99 11.9ms, p95 5.9ms; 10x (200 spinners, 1750 samples) max 19.1ms, p99 9.8ms; 20x (400 spinners, ~595 samples) max 28.9ms, p99 13.9ms. Zero samples >= 30ms at 10x/20x, zero >= 50ms anywhere; worst case consumed 59.5% of the 50ms slack. Cap-ignored observed sequence [13, 20, 42, 80, 163] — first violation is 80 >= 20+50, i.e. only the upper half of the comparison kills it.
```

**[should-fix] Recording (requested, observed) pairs kills the same mutants with no timing upper bound — built and measured**

Change `backoff_delays` to `Vec<(Duration, Duration)>` and push `(delay, started.elapsed())`. Then assert the REQUESTED sequence with `assert_eq!(asked, vec![10, 20, 20, 20, 20])` — exact, zero timing sensitivity, pinning arity, ordering, doubling and the cap outright — plus `recorded.iter().all(|(r, o)| o >= r)`, the load-immune direction that catches a delay computed correctly and then not honoured. I built this on the CI toolchain and ran the mutant set through it: identical kills, identical survivors, and the only remaining wall-clock assertions are the 200ms floor and the 2s runaway net, both of which load can only push in the safe direction.

```
Proposed variant (rustup run 1.97.1, --profile ci-test): unmutated PASS; sleep moved out of the guard KILLED; sleep deleted while recording KILLED; doubling removed KILLED; cap ignored KILLED; delay computed then not honoured KILLED; sleep moved out of backoff with push intact KILLED; bare sleep after the loop SURVIVES (unchanged and documented).
```

**[should-fix] The test hangs forever under load — 19/219 runs at 20x — via an Ok(0) spin in the fixture's read_request**

Under 400 spinners the PR head never fails an assertion but the process hangs indefinitely in `server.join().unwrap()`. `read_request` (engine/crates/xerj-autoindex/src/esclient.rs:917-942) does `let count = stream.read(&mut buffer).unwrap();` with no `if count == 0 { break }`. When the client's 25ms `bulk_timeout` fires before the request is fully written, the client drops the socket, `read` returns Ok(0) forever, the handler thread spins at 100% CPU, and `handler.join()` never returns. libtest has no per-test timeout, so in CI this burns the job timeout with no diagnostic — a strictly worse failure mode than the fast assertion failure it replaced. Pre-existing (main's read_request is byte-identical and main hangs 6/100 under the same load), so not a regression, but this is the load-hardening PR for this test and its verification ("550/550 passes under 96 and 200 spinners") did not surface it. A one-line `if count == 0 { break; }` fixes it.

```
PR head, 400 spinners, timeout 90s: rc0=103, rc124(hang)=16 of 119. PR head, 400 spinners, second window, timeout 45s: rc0=97, rc124=3 of 100. PR head, 200 spinners: 2 hangs in one batch of 150. main baseline, 400 spinners: rc0=33, rc101(260ms assert)=61, rc124=6 of 100. Mechanism proved by sampling a hung process: main thread wchan=futex_do_wait (the join), one handler thread state=R with utime 32->34 and stime 96->103 jiffies over 2s — a live spin, not a block.
```

**[should-fix] PR body still describes the design that commits 2 and 3 refuted**

The GitHub PR description is commit 9d7b349's text and was never updated. It opens with "I could not reproduce the reported flake" and "the 20% rate is unconfirmed here and I do not know what produces it" — which commit 076bda1 retracts as a method error (twelve spinners on a twenty-core box never oversubscribe) — and states "The client now counts its own backoff sleeps and the test asserts five", describing a count-based implementation that no longer exists on the branch. A reviewer working from the PR page gets a materially false account of both the problem and the fix. The commit trail is honest; the merge-facing artifact is not.

```
PR #445 body vs HEAD 848ed4e: body says "counts its own backoff sleeps and the test asserts five"; HEAD asserts the observed sequence against [10, 20, 20, 20, 20]. Body says the 20% rate is "unconfirmed"; my method control measured main's assertion failing 61/100 under 400 spinners, and 51-63% of probe runs exceeding 260ms at 10x-20x.
```

**[nit] Lower-bound comment justifies itself with a mutant that no longer reaches it**

The comment above `elapsed >= Duration::from_millis(200)` says "Without it, deleting the sleep from `backoff` while still recording the delay passes everything above". With observed recording that mutant now dies at the sequence assertion (esclient.rs:1224) and never reaches the lower bound. What the 200ms floor actually guards now is that `bulk_timeout` is honoured — six 25ms request timeouts are 150 of the 240ms floor — which is a real property and is undocumented. The floor itself is safe.

```
m2 (sleep deleted, recording kept) fails at esclient.rs:1224 with "got [0, 0, 0, 0, 0]", not at the lower bound. Elapsed across 4x/10x/20x, ~570 runs: min 241.1ms, p50 ~260ms, max 307.4ms, against a 200ms assertion.
```

**[nit] backoff_delays is shared across Es clones, which the field comment does not say**

`Es` derives `Clone` and `backoff_delays` is an `Arc<Mutex<Vec<_>>>`, so every clone of a client appends to one sequence. The comment says "per-instance rather than global so a concurrent test cannot perturb it", true across tests but not across workers driven off one client. This test issues a single bulk so it is fine today, but `with_retry` is reached from every clone and a future concurrent-bulk test reusing the recorder would see an interleaved sequence and an inexplicable ordering failure. One clause in the field comment would cover it.

```
engine/crates/xerj-autoindex/src/esclient.rs:13 `#[derive(Clone)]` over `pub struct Es`; field at :38 `backoff_delays: Arc<std::sync::Mutex<Vec<Duration>>>`. The adjacent `admission` field already documents its own shared-across-clones intent.
```

xerj-org:
## Independent verification of the delta — verbatim

Told explicitly that "I could not find anything material" was a valid verdict. It is the verdict.

### Verdict — refuted: `False`

> I could not find anything material in the delta. The three prior findings are genuinely fixed and the fixes did not introduce a new problem.
> 
> MUTANT MATRIX (10 mutants, toolchain 1.97.1, --profile ci-test, run against the whole `esclient` module so I would see collateral kills). Every mutant was anchor-checked before running, so a "survivor" cannot be a patch that silently failed to apply:
> 
>   sleep moved out of the guard              KILLED (requested seq -> 6 entries)
>   delay computed then not honoured          KILLED (observed >= requested)
>   doubling removed                          KILLED ([10,10,10,10,10])
>   retry_max_delay cap ignored               KILLED ([10,20,40,80,160])
>   sleep deleted while still recorded        KILLED (observed ~300ns)
>   MAX_ATTEMPTS 6 -> 5                       KILLED
>   MAX_ATTEMPTS 6 -> 7                       KILLED
>   backoff only on 429                       KILLED (requested seq empty)
>   sleep moved OUT of backoff to call site   KILLED (observed 240ns..512ns) [the mutant the doc comment names]
>   bare thread::sleep after the loop         SURVIVES (documented residual gap)
>   second bare sleep beside the recorded one SURVIVES (see nit 1)
> 
> 8 of 8 in the requested set behave exactly as the commit message claims. In every kill the failing assertion was one of the two the delta introduced (line 1228 requested-sequence, or 1234 observed>=requested) — none were incidental kills from an unrelated assertion. Clean build: 19/19 pass on both ci-test and dev.
> 
> HANG / FLAKE UNDER LOAD. I refused to trust a green run without a control, so I first reproduced the known failure with the same harness: main's binary, 400 spinners on 20 CPUs, 40 runs of the target test -> 21 FAIL (elapsed < 260ms) + 1 HANG. Harness is therefore severe enough to be meaningful. Same harness, this head: 120/120 clean, 0 hangs.
> 
> Pushed well past the PR's stated conditions — 800 spinners + 6 concurrent copies of the test binary at --test-threads=8 (~45x oversubscription), 90 runs of the full esclient module:
>   main: 15 ok / 67 fail / 8 HANG
>   head: 80 ok / 10 fail / 0 HANG
> Zero hangs on the head across 300+ runs at every load level. The `Ok(0)` break is real and is the thing that removed the hangs (main hung 8 times under the identical harness). It is also strictly safe: `Ok(0)` is EOF, so the old loop could never have made progress from that state — the break cannot cut short a read that would otherwise have succeeded.
> 
> Critically, NONE of the head's 10 residual failures were on the delta's new assertions. All were on assertions that predate this branch and are byte-identical on main: `accepted == 6` (head 6/90, main 4/90 — statistical parity) and `retry_path_recovers_after_one_bulk_timeout`'s 50ms client timeout (head 4/90, main 11/90). The `observed >= requested` direction is sound at any load: thread::sleep guarantees at least `delay` and Instant is monotonic, so it can only fail if the sleep did not happen.
> 
> Gates re-run after restoring the tree to pristine da04e61: `cargo fmt --all -- --check` rc=0; `cargo clippy --workspace --all-targets --profile ci-test -- -D warnings` rc=0. Working tree confirmed clean (`git status --porcelain` empty) so no mutant leaked into the measurements.
> 
> The doc comment is now accurate on the point that matters — its statement of the residual gap ("a sleep that never routes through here") correctly covers both surviving mutants. Its retrospective paragraph miscounts, which is a nit, not a defect.

**[nit] The doc comment's retrospective paragraph still miscounts survivors: 2 of the earlier 3 survive, not 1**

The forward-looking statement of the gap is correct and I am not disputing it — "a sleep that never routes through here" does cover everything that survives. The nit is the historical paragraph.

The earlier comment listed three survivors: (1) a bare thread::sleep after the loop, (2) the sleep moved out of `backoff` itself while the push stays, (3) a second sleep alongside the recorded one. The new comment says "Two of them — the sleep moved out of `backoff`, and the sleep deleted while the push stays — are killed." But "the sleep deleted while the push stays" is a restatement of item (2), not a distinct member of that list, so the sentence names item (2) twice and silently drops item (3). Item (3) still survives — I built it and it passes.

So of the three previously listed, ONE is killed and TWO survive; the comment asserts two are killed. That is the same shape of error the commit set out to correct, just in the opposite (more dangerous) direction: it now claims a kill it does not have.

Second, smaller drift in the same block: the surviving mutant is described as "a bare `thread::sleep` elsewhere in the loop", but the survivor I confirmed by that name (M8) is AFTER the loop, outside it. A reader taking "in the loop" literally would conclude a post-loop sleep is covered.

Also in engine/crates/xerj-autoindex/src/esclient.rs:38 — the two lines appended to the `backoff_delays` field doc have no separating `///`, so rustdoc runs them into the preceding paragraph, and that paragraph's opening line still reads "Every delay this client has actually backed off for" / "The DELAYS and not a count", describing a single Duration rather than the (requested, observed) pair the field now holds.

```
Mutant M9, applied to engine/crates/xerj-autoindex/src/esclient.rs:537 — a second bare sleep beside the recorded one, adding 5x20ms = 100ms of real sleeping:

    if attempt + 1 < MAX_ATTEMPTS {
        self.backoff(delay);
        std::thread::sleep(Duration::from_millis(20));   // <-- mutant
        delay = (delay * 2).min(self.retry_max_delay);
    }

  m9_bare_sleep_beside_recorded => rc=0 | target_ok=1 | test result: ok. 19 passed; 0 failed

For contrast, the two mutants the comment DOES name are both genuinely dead (so the doc is not wrong about those):
  m5_sleep_deleted_still_recorded => rc=101, panic at esclient.rs:1234
  m10_sleep_moved_to_call_site    => rc=101, panic at esclient.rs:1234:
    "a delay that is computed correctly and then not honoured is the one thing the
     requested sequence alone cannot see: [(10ms, 240ns), (20ms, 336ns), (20ms, 400ns),
     (20ms, 496ns), (20ms, 512ns)]"

Doc comment under review: engine/crates/xerj-autoindex/src/esclient.rs:552-567
```

**[nit] `accepted == 6` is the next wall-clock race in the same test — pre-existing, not a regression, but it is the assertion that now breaks first**

Not introduced by this delta and not a reason to hold the PR: the line is byte-identical on main and fails at statistically the same rate there (main 4/90, head 6/90 under the identical harness). Flagging it only because this branch's stated goal is that this test stop measuring the machine, and after the delta this is the one remaining assertion in it that still does.

`accepted` is incremented by the fixture thread just before it spawns each handler. The client makes six TCP connections that complete via the listen backlog without the server thread ever being scheduled, and `bulk()` returns ~240ms after it starts regardless. Under enough load the fixture thread has not drained the backlog by then, so the assertion reads a partial count. I observed left values of 0, 2 and 5 — 0 meaning the server thread had not been scheduled at all inside the client's whole 240ms window.

The fix is a one-line move, and it is only safe BECAUSE of this delta: `server.join()` blocks until all six accepts and all six handlers finish, and the handlers can only terminate without hanging because `read_request` now breaks on `Ok(0)`. Before this commit, moving the join earlier would have deadlocked.

    assert!(format!("{error:#}").contains("timed out"), "{error:#}");
    server.join().unwrap();                      // moved up from the end of the test
    assert_eq!(*accepted.lock().unwrap(), 6);

I verified this rather than just proposing it: 96/96 clean with the move vs 94/96 without, at 800 spinners x 8 concurrent copies, and 0 hangs — confirming the earlier join does not deadlock. Sample is small (96 runs each), so treat the delta as directional; the mechanism is what makes it sound.

```
Head as-is, 800 spinners x 8 concurrent copies x 12 rounds (96 runs of the target test alone):
  SPIN=800 PAR=8 ROUNDS=12 ok=94 fail=2 hang=0
  both failures at esclient.rs:1209 — assertion `left == right` failed, left: 5, right: 6

Same binary with server.join() moved above the assertion:
  SPIN=800 PAR=8 ROUNDS=12 ok=96 fail=0 hang=0

Control proving the harness bites, main under the identical 800x6x8 config:
  main: ok=15 fail=67 hang=8   (60 failures at esclient.rs:1153 = the old `elapsed < 260ms`,
                                4 at esclient.rs:1150 = this same `accepted == 6`)
  head: ok=80 fail=10 hang=0   (6 at esclient.rs:1209 = `accepted == 6`,
                                4 at esclient.rs:1171 = retry_path_recovers, both pre-existing)

Assertion under review: engine/crates/xerj-autoindex/src/esclient.rs:1209
```

PR #443, opened 2026-08-17
fix(autoindex): --follow-symlinks indexed secrets through visible link names and escaped the indexed root

xerj-org:
Closes #438.

## Reproduced first

```
follow=false   rel="README.md"
follow=true    rel="README.md"
               rel="notes.txt"          is_symlink=true  body="INNER SECRET"   <- target is .secretdir/k.txt
               rel="shared/etc/shadow"  body="root:$6$OUTSIDE_THE_ROOT$hash:..."  <- outside the indexed root
```

`notes.txt` is an ordinary **visible** name pointing into a hidden directory. `shared` is a link to a tree the operator never pointed at. The report said `hidden:dotfile dirs: 1` in both runs — true of the traversal, false of the outcome.

## Why the rule failed

`walk.rs` judges `entry.file_name()`, the name the walk arrived by. With links followed that is not the file. The security comment above it asserts the hidden rule "is the reason secrets stay out"; under this flag it was not.

## The change

Under `--follow-symlinks`, the same two questions are asked of the **resolved** path. Canonicalising collapses every link at once, so a chain (`chain.txt -> hop.txt -> .secretdir/k.txt`) and a link partway up a directory path behave like a direct one.

* target outside the canonical root → refused, reported under a new `symlink:outside-root` rule
* target inside a hidden directory → skipped, reported as the dotfile it is
* unresolvable → left alone; a broken link has no target to leak, and walkdir already reports it

## What must not change

Tested, because a fix that indexes nothing also passes the leak assertions:

* a link to a **visible** file inside the root is still indexed
* a link to a **directory** inside the root is still walked, and the hidden rule still applies underneath it
* `--follow-symlinks` off is byte-identical
* a dangling link is not an escape

## Fail-before

Removing **only** the new guard from `walk.rs` brings `notes.txt`, `hop.txt`, `chain.txt` and `shared/etc/shadow` straight back:

```
`notes.txt` resolves into .secretdir and was indexed anyway:
["README.md", "chain.txt", "hop.txt", "notes.txt", "plain-link.txt", "shared/etc/shadow", "sub/deep.md"]
test result: FAILED. 4 passed; 1 failed
```

## Docs

`--help` and both copies of the zero-config recipe said only "(loop-safe)", which is a different property. They now say what the flag does to the secret rule and the root boundary.

## Verification

[...]

PR #446, opened 2026-08-17
fix(engine): refuse to open an index whose vectors were produced under a different embedding.mode

### LINKED ISSUE #434: switching embedding.mode lexical -> neural on an existing index is unguarded: neural query vectors scored against lexical document vectors, silently
Found by an independent verification of #430, which documents an air-gapped procedure that walks an operator straight into this.

## What happens

Build an index with `embedding.mode = "lexical"`, ingest, query — fine. Then change **only** the embedding block to `mode = "neural"` and restart. Nothing errors. Query vectors are now produced by the neural embedder and scored against document vectors produced by the lexical one, in the same 384-dimensional space. No startup error, no query error, silently wrong rankings.

## Why nothing catches it

`validate_embedding_identity` (`engine/crates/xerj-engine/src/index.rs:27623`) delegates to `configured_onnx_identity`, which returns `Ok(None)` for any mode that is not `onnx-experimental`:

```rust
// index.rs:27457-27459
fn configured_onnx_identity(cfg: &EmbeddingConfig) -> Result<Option<PersistedEmbeddingIdentity>> {
    if !cfg.mode.trim().eq_ignore_ascii_case("onnx-experimental") {
        return Ok(None);
```

With `configured == None` and no marker file, the only remaining exit is a short-circuit:

```rust
// index.rs:27678-27681
let Some(identity) = configured else {
    return Ok(());
};
```

So the identity guard — the mechanism built precisely to stop embeddings being compared across incompatible producers — does not cover the lexical/neural boundary at all.

**The dimension check cannot catch it either**, and that is deliberate:

```rust
// engine/crates/xerj-ai/src/local.rs:26-28
/// 384 mirrors the popular all-MiniLM-L6-v2 output size, so switching from
/// the built-in embedder to a real proxy of that width needs no mapping change.
pub const DEFAULT_DIMS: usize = 384;
```

The lexical embedder was sized to match MiniLM, so the widths agree and the mismatch is invisible to every check that exists.

## Why it is worth fixing rather than documenting

This is the failure mode the ONNX identity marker was built to prevent, on a path the docs actively recommend. `docs/recipes/production-deployment.md` does not warn about it either. An operator who follows #430's section 4 (build and query a lexical index) and then section 3 ("change only the embedding block") gets silently degraded search with no signal at any layer.

Suggested shape: persist the embedding **mode** alongside the existing ONNX identity, and refuse to open an index whose stored mode differs from the configured one unless an explicit reindex/override flag is passed — the same fail-closed posture #204 established elsewhere.

**Verified:** the code paths above, read on `main`.
**Not verified:** I have not measured the ranking degradation end to end, nor confirmed the behaviour on a running node — this is source inspection by the verification agent, and reproducing it should be step one.

xerj-org:
Closes #434.

## Reproduced first

Build with `embedding.mode = "lexical"`, ingest, query. Change **only** the embedding block to `"neural"`. Restart on the same data directory.

```
xerj v1.0.0-rc.17 starting
embedding backend: built-in neural BERT   model=all-MiniLM-L6-v2
_cluster/health -> {"status":"green"}
```

Same query, same data:

```
lexical (as built)   1: 0.6303   3: 0.5055   2: 0.5008   4: 0.4943   5: 0.4783
neural  (after swap) 1: 0.5496   4: 0.5402   3: 0.5375   2: 0.5223   5: 0.5107
```

The reordering matters less than the spread: 0.6303–0.4783 collapses to 0.5496–0.5107. Neural query vectors against lexically-produced document vectors give near-uniform similarity — the ranking is close to noise.

## Why nothing caught it

`configured_onnx_identity` returns `None` for any mode that is not `onnx-experimental`, so `validate_embedding_identity` short-circuits. The identity marker never covered the lexical/neural boundary. The dimension check cannot either — the lexical embedder is 384 wide to match MiniLM **on purpose**, so the widths agree.

## After

```
A  lexical, 1 doc, semantic search works
B  same directory reopened neural -> index unavailable, error naming both modes
C  back to lexical -> 1 hit, unchanged
```

## Deliberately narrow, because a guard's failure mode is refusing something legitimate

* `mode` is `#[serde(default)]` and **outside** the identity equality. A marker written before this field existed reads back as `""` = unknown, not mismatched. Folding it into `PartialEq` would refuse every existing ONNX index on upgrade — the same shape as adding a field to a digest-covered struct, which I did on #443 last cycle and had to undo.
* a mode-only marker (empty ONNX fields) must not read as "this index contains ONNX vectors". **Two existing tests caught me doing exactly that**: they restart under `mode = "auto"` and my first version told them the index held ONNX vectors.
* the marker is written only for a new or empty index. For one that already holds vectors the producing mode is unknowable, and stamping the current mode on it would assert something possibly false.

**So an index built before this change is not protected until it is rebuilt.** That is the honest limit; the alternative is inventing a history for it.

## Verification

`rustup run 1.97.1`: fmt clean, `clippy --workspace --all-targets -- -D warnings` clean, 601 engine lib tests pass.

Not verified: the reverse direction end to end (neural index reopened lexical — the guard is symmetric by construction but I measured lexical→neural), proxy and ONNX interactions with a changed mode, and whether `_reindex` in place is sufficient recovery.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -27373,6 +27373,33 @@ fn embed_backend_error(e: anyhow::Error, ctx: &str) -> xerj_common::XerjError {
 
 const EMBEDDING_IDENTITY_FILE: &str = "embedding_identity.json";
 
+/// Which embedder produced this index's vectors, as its resolved EXECUTION
+/// identity — not the `embedding.mode` string that was requested.
+///
+/// A SEPARATE file from [`EMBEDDING_IDENTITY_FILE`] on purpose. Widening that
+/// one breaks downgrade: an older binary parses it, sees a record whose ONNX
+/// fields are empty, and concludes the index contains ONNX vectors — so a
+/// rollback would leave every index the new binary had touched red and
+/// unopenable. This repo has already made that call once; the CHANGELOG records
+/// a sidecar being chosen elsewhere for exactly this reason. An older binary
+/// simply never opens this file.
+///
+/// Keyed on `identity_sha256` from [`embedding_execution_identity`] because the
+/// mode string is not the embedder. `auto` with no endpoint, `proxy` with no
+/// endpoint, `neural` in a build without the `neural` feature, and any
+/// unrecognised value all resolve to the SAME lexical embedder, so keying on
+/// the string refuses an index over a change that cannot alter a single vector.
+/// The resolved identity also moves when the mode string does not — a different
+/// proxy endpoint or neural model — which is the half a mode check cannot see.
+const EMBEDDING_EXECUTION_FILE: &str = "embedding_execution.json";
+
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+struct PersistedEmbeddingExecution {
+    version: u32,
+    backend: String,
+    identity_sha256: String,
+}
+
 #[derive(Debug, PartialEq, Serialize, Deserialize)]
 struct PersistedEmbeddingIdentity {
     version: u32,
@@ -27675,6 +27702,13 @@ fn validate_embedding_identity(
             ))),
         };
     }
+    // After the ONNX branch, deliberately. When an ONNX marker exists and ONNX
+    // is configured, that branch has already returned with a message naming the
+    // model and tokenizer digests — strictly more useful than this one, which
+    // can only say the backend changed. Running first made the generic message
+    // preempt the specific one, which `changed_onnx_assets_are_rejected`
+    // caught.
+    validate_embedding_execution(index_dir, cfg, new_index, has_documents)?;
     let Some(identity) = configured else {
         return Ok(());
     };
@@ -27694,10 +27728,83 @@ fn validate_embedding_identity(
     Ok(())
 }
 
+/// Refuse to open an index whose vectors were produced by a DIFFERENT embedder.
+///
+/// The gap this closes (#434): `configured_onnx_identity` returns `None` for
+/// every mode except `onnx-experimental`, so the ONNX identity check
+/// short-circuits and never sees a lexical/neural swap. The dimension check
+/// cannot see it either — the lexical embedder is 384 wide to match
+/// all-MiniLM-L6-v2 on purpose. Measured on a lexical index reopened as neural:
+/// status green, no error, and the same query reordering with the score spread
+/// collapsing from 0.6303..0.4783 to 0.5496..0.5107, which is what comparing
+/// two embedders' vectors looks like — near-uniform similarity, i.e. noise.
+///
+/// Three boundaries, each of which a previous version of this function got
+/// wrong:
+///
+///   * the record is the RESOLVED identity, so `auto` and the mode it resolves
+///     to are one space and a changed proxy endpoint is a different one;
+///   * it is REFRESHED while the index is empty rather than written once, so a
+///     mode change made while empty (which this permits) cannot leave a stale
+///     record that refuses the next restart under the very config that wrote
+///     the vectors;
+///   * a missing record means UNKNOWN, never mismatched. An index built before
+///     this existed is unprotected until it is rebuilt, which is the honest
+///     limit — the alternative is inventing a history for it.
+fn validate_embedding_execution(
+    index_dir: &Path,
+    cfg: &xerj_common::config::EmbeddingConfig,
+    new_index: bool,
+    has_documents: bool,
+) -> Result<()> {
+    let path = index_dir.join(EMBEDDING_EXECUTION_FILE);
+    let current = embedding_execution_identity(cfg)?;
+    if path.exists() {
+        let persisted: PersistedEmbeddingExecution = serde_json::from_slice(&std::fs::read(&path)?)
+            .map_err(|e| {
+                EngineError::Common(xerj_common::XerjError::embedding(format!(
+                    "cannot read {}: {e}; remove it to re-record, or reindex",
+                    path.display()
+                )))
+            })?;
+        if persisted.identity_sha256 != current.identity_sha256 && has_documents {
+            return Err(EngineError::Common(xerj_common::XerjError::embedding(
+                format!(
+                    "index {} holds vectors produced by embedding backend {:?} but this server                      resolves embedding.mode={:?} to backend {:?}. Query vectors from one                      embedder scored against document vectors from another are silently wrong                      rather than visibly wrong: the widths can agree, so no dimension check                      fires, and the scores stay plausible while the ranking degrades toward                      noise. Restore the previous embedding configuration, or reindex into a new                      index under this one.",
+                    index_dir.display(),
+                    persisted.backend,
+                    cfg.mode.trim(),
+                    current.backend
+                ),
+            )));
+        }
+    }
+    if new_index || !has_documents {
+        write_file_atomic(
+            &path,
+            &serde_json::to_vec_pretty(&PersistedEmbeddingExecution {
+                version: 1,
+                backend: current.backend.clone(),
+                identity_sha256: current.identity_sha256.clone(),
+            })?,
+        )?;
+    }
+    Ok(())
+}
+
 fn ensure_embedding_identity_for_new_field(
     index_dir: &Path,
     cfg: &xerj_common::config::EmbeddingConfig,
 ) -> Result<()> {
+    // The execution record covers this path too. An index created WITHOUT a
+    // semantic field writes no record, so #434 reproduced verbatim through
+    // `PUT /_mapping`: the field arrives later, vectors are written under
+    // whatever mode is configured then, and nothing had ever recorded which
+    // embedder made them. `new_index` is false here and the caller only reaches
+    // this before the first write to the new field, so the record is written
+    // when the index has no documents and compared when it does.
+    // The new field has no vectors yet, so this records rather than compares.
+    validate_embedding_execution(index_dir, cfg, false, false)?;
     let Some(configured) = configured_onnx_identity(cfg)? else {
         return Ok(());
     };
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -27770,7 +27770,7 @@ fn validate_embedding_execution(
         if persisted.identity_sha256 != current.identity_sha256 && has_documents {
             return Err(EngineError::Common(xerj_common::XerjError::embedding(
                 format!(
-                    "index {} holds vectors produced by embedding backend {:?} but this server                      resolves embedding.mode={:?} to backend {:?}. Query vectors from one                      embedder scored against document vectors from another are silently wrong                      rather than visibly wrong: the widths can agree, so no dimension check                      fires, and the scores stay plausible while the ranking degrades toward                      noise. Restore the previous embedding configuration, or reindex into a new                      index under this one.",
+                    "index {} holds vectors produced by embedding backend {:?} but this server resolves embedding.mode={:?} to backend {:?}. Query vectors from one embedder scored against document vectors from another are silently wrong rather than visibly wrong: the widths can agree, so no dimension check fires, and the scores stay plausible while the ranking degrades toward noise. Restore the previous embedding configuration, or reindex into a new index under this one.",
                     index_dir.display(),
                     persisted.backend,
                     cfg.mode.trim(),


xerj-org:

Three skeptics: false positives, false negatives, upgrade safety. None marked the PR refuted; between them they filed **six must-fix findings**, and two of those are worse than the bug this PR fixes. Verbatim:

> The premise holds and the fix works for exactly the shape it was built against — and for nothing wider. Reproduced on main (2471199): lexical index, restart neural, cluster green, no error, scores go [('1',0.5719),('3',0.5709),('2',0.5005),('5',0.4978),('4',0.4776)] -> [('3',0.5114),('1',0.5071),('5',0.5025),('2',0.4997),('4',0.4982)] — top-1 flips and the spread collapses from 0.094 to 0.013. On the branch (a6556fe) the same sequence quarantines the index (health red, correct message). The REVERSE direction is real, not assumed: build neural [('1',0.8327),('3',0.7592),('2',0.526)], reopen lexical -> quarantined, 'holds vectors produced with embedding.mode="neural" … configured for embedding.mode="lexical"'. Symmetry verified. But four independent routes to corrupted-but-accepted survive on the branch, three of them with green health: (1) an index whose semantic field arrives via PUT _mapping never gets a marker at all, because `ensure_embedding_identity_for_new_field` keeps the identical ONNX-only short-circuit that caused #434 — the corrupted ranking comes back bit-for-bit on a brand-new index created by the fixed binary; (2) flipping mode while the index is EMPTY is accepted and does not update the marker, so the documents written next are neural under a marker that says lexical — the index is then refused under the mode that actually made its vectors and accepted under the mode that did not, and the error tells the operator to 'restore the original mode', i.e. to switch to the corrupting configuration; (3) the mode STRING is not an embedder identity — proxy->proxy with a different endpoint, and proxy-with-empty-endpoint (which silently emits lexical vectors) followed by a real endpoint, both reproduce #434 green; (4) deleting the one unauthenticated JSON file restores the pre-fix behaviour, and it is the most obvious escape from (2). In the other direction the string is too fine: 'auto' and 'lexical' are provably the same embedder (identical vectors, first dims 0.141486) yet relabelling the config quarantines a healthy index. Also flagged, statically only: adding `mode` to the derived PartialEq breaks `ensure_embedding_identity_for_new_field`, which still compares `persisted != configured` — an ONNX index created by this build should now refuse a routine PUT _mapping that adds a semantic field; I did not get that test to run.

**[must-fix] The guard never engages when the semantic field was added after index creation — #434 reproduces verbatim on the branch**

`validate_embedding_identity` returns early when `schema_has_embeddings` is false, so an index created without a semantic field writes no marker. The sibling function that runs when a semantic field is added later, `ensure_embedding_identity_for_new_field` (index.rs:27776, called from `add_fields` at index.rs:18635), opens with `let Some(configured) = configured_onnx_identity(cfg)? else { return Ok(()) }` — the exact same ONNX-only short-circuit that caused #434, left untouched in the twin function. So `PUT /idx/_mapping {"properties":{"body":{"type":"semantic_text"}}}` produces a semantic index with NO marker, and at the next open the markerless+has_documents+non-ONNX path also refuses to backfill one. The index is unprotected forever. This is not the acknowledged 'built before this change' narrowing: it is a brand-new index created by the fixed binary through an ordinary ES workflow.

```
p3: create `sem` with only a text field, PUT _mapping adds semantic_text, bulk 5 docs under --embed-mode lexical → `marker after adding semantic_text: NONE`. Scores lexical/lexical [('1',0.5719),('3',0.5709),('2',0.5005),('5',0.4978),('4',0.4776)]. Restart --embed-mode neural: node green, index serves, scores [('3',0.5114),('1',0.5071),('5',0.5025),('2',0.4997),('4',0.4982)] — bit-identical to the unfixed main run, top-1 flips 1→3, spread collapses 0.094→0.013. `marker now: NONE`. Same result for an index built by the main binary and then written to by the fixed binary: the new binary wrote a fresh doc under lexical and still left `marker: NONE`, then flipped to neural green with corrupted ranking.
```

**[must-fix] Empty-index window writes a lying marker: the index is refused under the mode that made its vectors and accepted under the mode that did not**

The mode mismatch only errors when `has_documents` is true, and the `Some(identity) else` write arm only runs when `new_index || !has_documents` — but that arm is unreachable for a mode-only marker, because the `persisted.backend.is_empty() && persisted.model_sha256.is_empty()` branch returns `Ok(())` immediately when `configured.is_none()`. So opening an EMPTY index under a new mode is accepted and the marker is NOT updated. Every document written afterwards carries the new mode's vectors under the old mode's marker. The guard then enforces the lie: it quarantines the index under the mode that actually produced the vectors, and admits it under the wrong one. Worse, the error text says 'Restore the original mode', which for this index is precisely the corrupting action.

```
p4 on branch: create `sem` under lexical, index nothing → marker mode="lexical". Restart neural (still empty) → accepted, marker unchanged (mode="lexical"). Bulk 3 docs → neural scores [('1',0.8327),('3',0.7592),('2',0.526)], marker still mode="lexical". Restart neural: index quarantined, health red, 'holds vectors produced with embedding.mode="lexical" but the server is configured for embedding.mode="neural"'. Restart lexical: health green, index serves, scores [('1',0.5067),('3',0.5036),('2',0.5021)] — lexical query vectors against neural document vectors, spread 0.0046.
```

**[must-fix] `mode` is not an embedder identity: proxy→proxy and neural→neural swaps reproduce the same corruption with the guard fully engaged**

The marker records only the normalised mode string. Everything else that determines the vector space for that mode is unrecorded: `default_endpoint`, `default_model` (proxy), `neural_model`, `local_model_dir`, `model_cache_dir` (neural), and the presence of the `neural` cargo feature (a `--no-default-features` build silently serves lexical vectors for mode="neural", index.rs:28418). Two of these were executed and reproduce #434 on the branch with green health. `mode` is also the only field EXCLUDED from `same_embedding_identity`, so the one identity the file does compute is not applied to the modes that most need it.

```
p6 on branch: mode=proxy against provider A → scores [('3',0.5327),('1',0.5301),('2',0.508)], marker mode="proxy". Restart with default_endpoint pointed at provider B (different 384-dim space, same mode string) → guard silent, health green, scores [('1',0.5049),('3',0.5023),('2',0.477)], top-1 flips. p8 on branch: mode=proxy with default_endpoint="" logs 'falling back to lexical' and writes LEXICAL vectors under marker mode="proxy" → [('1',0.5719),('3',0.5709),('2',0.5005)]. Restart with a real endpoint, same mode string → accepted, green, [('2',0.4967),('3',0.4864),('1',0.4857)]: the golden-retriever document now outranks both finance documents for 'how did profit margins move this quarter'.
```

**[should-fix] `auto` and `lexical` are the same embedder, yet relabelling the config bricks a healthy index**

`make_embedder` (index.rs:28271) sends both mode="auto" (with no endpoint) and mode="lexical" to `Embedder::lexical()`, so their vectors are identical. `embedding_mode_key` compares the strings, so making the effective default explicit — the exact thing the --help text invites, since it labels lexical 'default' while `EmbeddingConfig::default()` is "auto" — quarantines the index. The same false positive applies to any unrecognised mode string (config.rs: 'Unknown values are treated as auto'), including a typo that is later corrected: fixing `mode = "lexicl"` to `mode = "lexical"` refuses the index it just wrote. The mode string is simultaneously too coarse (finding 3) and too fine (here) to be the identity.

```
p5 on branch: same document under --embed-mode auto and under --embed-mode lexical produce byte-identical vectors, first 6 dims [0.141486,0.0,0.0,0.0,0.0,0.0], dims=384. Reopening the auto-built index with --embed-mode lexical: health red, 'index [sem] failed to open and is unavailable: … holds vectors produced with embedding.mode="auto" but the server is configured for embedding.mode="lexical"'.
```

**[should-fix] Adding `mode` to the derived PartialEq breaks the ONNX add-a-semantic-field path**

`PersistedEmbeddingIdentity` derives `PartialEq` and the new `mode` field is inside it. `validate_embedding_identity` now writes the marker as `PersistedEmbeddingIdentity { mode, ..identity }`, so an ONNX index created by this build has mode="onnx-experimental" on disk — but `ensure_embedding_identity_for_new_field` still compares `persisted != configured` (index.rs:27793) against a `configured` whose `mode` is always `String::new()` (index.rs:27503). Every field on the right matches, `mode` does not, so a routine `PUT /idx/_mapping` adding a second semantic_text field to an ONNX index is refused with 'refusing to add a semantic field in a different vector space'. Commit a6556fe fixed the constructor the field broke but not the comparison it also broke, and the existing test `later_semantic_field_pins_identity_before_first_write` (index.rs:28191) cannot catch it because it starts from an empty dir where no marker with a mode was ever written. NOT EXECUTED: I started `cargo test -p xerj-engine --features onnx-experimental embedding_identity_tests`, but killed it myself during cleanup before it linked, so the compile errors in that log are SIGTERM, not real. Static reading only — verify before acting.

```
index.rs:27376 `#[derive(Debug, PartialEq, …)]` over a struct that now contains `mode`; index.rs:27503 `mode: String::new()` in `configured_onnx_identity`; index.rs:27768 `let identity = PersistedEmbeddingIdentity { mode, ..identity };` before the marker write; index.rs:27793 `if persisted != configured` in `ensure_embedding_identity_for_new_field`. `same_embedding_identity` was introduced precisely to avoid this comparison but is not used here.
```

**[should-fix] The marker is unauthenticated and its absence means 'trusted', so deleting one file restores the corruption**

The whole guarantee now rests on a plain JSON file with no checksum, no signature and no tie to the segment data. Deleting it returns the index to the pre-fix state (accepted, green, corrupt); editing `mode` inside it makes any pairing acceptable. Absence is treated as 'unknown, allow' at every layer: missing file, missing `mode` key, and `mode: ""` (reachable from a config with `mode = ""`, which serialises out via `skip_serializing_if`). An operator hitting the refusal from finding 2 or 4 will find 'rm embedding_identity.json' as the fastest path back to a green cluster, which is exactly the wrong repair.

```
p7 on branch: index `prot` (marker present) is quarantined after the lexical→neural flip, health red. `rm prot/embedding_identity.json`, restart neural → health green, `prot` serves [('3',0.5114),('1',0.5071),('2',0.4997)] — the corrupted ranking. Same run: an alias spanning `prot` (protected) and `unprot` (markerless) fails the whole alias search with no_shard_available while `unprot` alone silently serves corrupted hits.
```

**[nit] The refusal message is emitted with mangled whitespace**

The `format!` string for the new error uses literal newlines with indentation instead of `\` line continuations, so the message reaches both the log and the HTTP error body with runs of ~22 spaces mid-sentence. This is the operator-facing text for the whole feature and it is the first thing anyone will paste into an issue.

```
API body: 'holds vectors produced with embedding.mode="lexical" but the server                      is configured for embedding.mode="neural". Query vectors from one embedder                      scored against document vectors from another are silently wrong…' (index.rs:27686). Compare the adjacent pre-existing messages, which use `\` continuations correctly.
```

> The guard keys on the raw `embedding.mode` STRING, not on the backend that string resolves to — and the engine already has a resolver (`embedding_execution_identity`) that says `auto`, `proxy`-without-endpoint, `neural`-without-the-cargo-feature and any typo all resolve to the SAME lexical embedder. Verified on a real node: an index created under the shipped config default (`auto`, marker records "auto") refuses to open after a restart with `--embed-mode lexical` — the mode `--help` itself calls "default" — taking every semantic index on the node to red with no data change whatsoever. Worse, the marker is written once and never refreshed, so a mode change made while an index is EMPTY (which the guard permits) is not recorded: documents are then written under the new mode, accepted and served, and the NEXT restart with the IDENTICAL config refuses the index, claiming vectors were produced under a mode no document in it ever saw. That makes a same-config restart non-deterministic, and the error's own remediation ("Restore the original mode") would then re-create #434. Separately, the PR's stated invariant — `mode` sits outside the identity equality — is implemented at only one of the two comparison sites: `ensure_embedding_identity_for_new_field` still uses the derived `PartialEq`, which now includes `mode`, so adding a semantic field to an ONNX index created after this change is refused as "a different vector space" with byte-identical assets. Blast radius is well handled (failing index recorded failed, node and other indices stay up, actionable per-index error, no marker churn on identical restarts) — the mechanism is sound, the trigger condition is wrong.

**[must-fix] `auto` and the mode it resolves to are treated as different vector spaces — the default config bricks on the first explicit `--embed-mode`**

`embedding_mode_key` records `cfg.mode.trim().to_ascii_lowercase()` — the requested string. But `embedding_execution_identity` (already in this file, index.rs:27512-27535) resolves `auto` with no `default_endpoint` to backend "lexical", i.e. literally the same embedder as explicit `lexical`. `EmbeddingConfig::default()` is `mode: "auto"` (config.rs:1327), so every index created without an explicit mode gets a marker saying "auto"; the server's own `--help` then advertises `lexical` as "built-in feature-hash (default; offline, not neural)", inviting the operator to write it explicitly. Fix: record and compare the RESOLVED backend from `embedding_execution_identity`, not the requested string. Same defect for `proxy` with an empty `default_endpoint` (resolves to lexical) and for any unrecognised value, which config.rs:1260 documents as "treated as auto" — correcting a typo like `lexcial` -> `lexical` refuses the index.

```
Node run (slim build, own --data-dir, ports 9455/9465): created `docs` and `other` with `semantic_text`, indexed 1 doc each, semantic search returned _score 0.719658. Marker: {"mode":"auto"}. Restarted with only `--embed-mode lexical` added:
  WARN failed to open index name="other" error=embedding error: index .../other holds vectors produced with embedding.mode="auto" but the server is configured for embedding.mode="lexical"...
  WARN failed to open index name="docs" error=... (same)
  GET /_cluster/health -> "status":"red", unassigned_primary_shards: 2
  POST /docs/_search -> no_shard_available_action_exception "index [docs] failed to open and is unavailable"
Unit probes on validate_embedding_identity (xerj-engine lib, ci-test, 1.97.1) — all three FAIL, all three are the same embedder:
  w446_p1 auto->lexical: auto resolves to backend "lexical", lexical resolves to "lexical", open REFUSED
  w446_p2 proxy(no endpoint)->lexical: proxy resolves to "lexical", open REFUSED
  w446_p3 "lexcial"->"lexical": typo resolves to "lexical", open REFUSED
Control that passed: w446_p5 (" Lexical ", "LEXICAL", "\tlexical\n" all accepted) — case/whitespace normalisation is correct.
```

**[must-fix] The marker is never refreshed, so a restart with the IDENTICAL config can refuse an index the previous run just wrote and served**

The write arm is `if new_index || !has_documents { write marker }`, but it is only reached when `configured` is `None` AND the marker does not already exist — an existing marker returns `Ok(())` from the mode-only branch without updating `mode`. So the sequence the guard itself permits (change mode while the index is empty) leaves a stale marker, and the next restart under the same, now-correct mode is refused. The advice in the error — "Restore the original mode" — is then actively harmful: in a real lexical->neural change it tells the operator to run the exact configuration #434 is about. Fix: refresh the recorded mode whenever the index is empty, on every open, not only when the marker is absent.

```
Node run, /tmp/w446-legit/node2, port 9465:
  1. start with no --embed-mode, PUT /empty {semantic_text}, zero docs -> marker {"mode":"auto"}
  2. restart with --embed-mode lexical -> 0 "failed to open" lines, index opens (empty, so no refusal); marker STILL {"mode":"auto"}
  3. POST /empty/_doc -> 201 created, _refresh ok — this document's vector was produced by the lexical embedder of this very process
  4. restart with --embed-mode lexical again, byte-identical command line ->
     WARN failed to open index name="empty" error=... holds vectors produced with embedding.mode="auto" but the server is configured for embedding.mode="lexical"
Unit probe w446_p7_stale_marker_on_empty_index reproduces the same three steps and FAILS at step 4.
```

**[must-fix] ONNX regression: adding a semantic field to an index created under this change is refused with identical assets**

The PR's stated design is that `mode` sits outside the identity equality, and `same_embedding_identity` implements that — but only in `validate_embedding_identity`. `ensure_embedding_identity_for_new_field` (index.rs:27791) still compares `if persisted != configured` using the derived `PartialEq`, and `mode` was added to the struct's derive list (index.rs:27376). `configured_onnx_identity` always builds `mode: String::new()` (second commit a6556fe), while `validate_embedding_identity` stamps the real mode via `PersistedEmbeddingIdentity { mode, ..identity }`. So on an ONNX index created after this change, `PUT /<index>/_mapping` adding a semantic field (index.rs:18635, reached whenever `schema_has_embeddings(&fields)`) refuses with "refusing to add a semantic field in a different vector space" though model and tokenizer sha256 are identical. Fix: use `same_embedding_identity` at that call site too.

```
Unit probe w446_p10_mode_is_inside_derived_partial_eq PASSES, proving both halves: `same_embedding_identity(with_mode, mode_empty)` is true (same assets) while `with_mode != mode_empty` under the derived PartialEq that `ensure_embedding_identity_for_new_field` uses. Path confirmed by reading index.rs:27776-27808 (compares with `!=`) and index.rs:27490-27506 (`configured_onnx_identity` hardcodes `mode: String::new()`). The PR's existing test `later_semantic_field_pins_identity_before_first_write` does not catch it because it never goes through `validate_embedding_identity`, so both sides have mode=="".
```

**[should-fix] The guard is silent on the embedder changes it cannot see, including a supported build-flag change with identical config**

Because the key is the requested string, a genuine change of embedder that keeps the string is invisible. (a) `proxy`: swapping `default_endpoint` or `default_model` — text-embedding-3-small to voyage-3 — leaves mode=="proxy", marker matches, index opens, and that is exactly the #434 class with a remote embedder. (b) `neural` on a `--no-default-features` server (a documented supported build: "the lexical + proxy backends still work") silently runs the LEXICAL embedder, yet the marker records "neural"; moving that data directory to the shipped batteries-included binary with a byte-identical config changes the embedder from lexical to BERT and the guard says nothing. So the guard refuses same-embedder/different-string and permits different-embedder/same-string — the inverse of what it is for.

```
Unit probe w446_p4_neural_without_feature: `embedding_execution_identity(mode="neural").backend == "lexical"` on the default (no `neural` feature) xerj-engine build, while the marker written is {"mode":"neural"} — the marker asserts a producer that never ran. crates/xerj-server/Cargo.toml:69 `default = ["neural"]` vs crates/xerj-engine/Cargo.toml:72 `default = []` confirms both build variants ship. For proxy: `embedding_mode_key` reads only `cfg.mode`; `default_endpoint`/`default_model` appear nowhere in the marker or the comparison.
```

**[should-fix] Two ways to end up permanently unguarded: an explicitly empty `mode`, and a semantic field added later on a non-ONNX node**

(a) `mode = ""` in the config file normalises to "" and is written with `skip_serializing_if = "String::is_empty"`, so the field is omitted; on read it is "" = "unknown" and the mode check is skipped forever. That index can be flipped lexical->neural silently for the rest of its life. (b) An index created with no semantic field returns early from `validate_embedding_identity` (`schema_has_embeddings` false, no marker), and when the field is added later `ensure_embedding_identity_for_new_field` returns `Ok(())` immediately because `configured_onnx_identity` is `None` for every non-ONNX mode — so no marker is ever written and #434 stays fully live for that index. These are not false refusals, but they are large holes in the population the guard claims to cover, on top of the acknowledged pre-existing-index gap.

```
Unit probe w446_p6_empty_mode_leaves_index_unguarded: create with mode="" writes a marker with no `mode` key at all; reopening with mode="neural" and has_documents=true returns Ok. Unit probe w446_p9_no_semantic_field: after `ensure_embedding_identity_for_new_field(dir, lexical)` the marker file does not exist ("<no marker>"), and the subsequent lexical->neural open with documents returns Ok.
```

**[nit] The new error message lost its line continuations — 22-space runs in every log line and HTTP error body**

index.rs:27686 is a single ~700-character string literal where the `\`-newline continuations used by every other message in this function (e.g. index.rs:27724) were flattened into literal whitespace. It is the most user-visible artifact of the change: it appears in the startup WARN and in the `no_shard_available_action_exception` reason returned to clients. Neither rustfmt nor clippy flags it because it is inside a string literal.

```
Source: `"index {} holds vectors produced with embedding.mode={:?} but the server                      is configured for embedding.mode={:?}. Query vectors from one embedder                      scored against..."`. Node output reproduced verbatim in n2.log, n5.log and in the /docs/_search error body: `... but the server                      is configured for embedding.mode="lexical". Query vectors ...`
```

### Verdict 3 — refuted: `False`

> "Unprotected until rebuilt" is not the only consequence — it is the mildest one. Two confirmed defects sit next to it.
> 
> (1) The upgrade cannot be undone. The mode-only marker is written into the existing `embedding_identity.json` with `backend: \"\"` and `version: 1`, and the previous release reads that as \"this index contains ONNX vectors\" and refuses to open it. It is written whenever `new_index || !has_documents`, so a pre-existing EMPTY lexical index is poisoned by starting the new binary once, with no config change and no writes. Verified end-to-end on three data dirs: parent → branch → parent leaves the cluster red and the index 503, with a message that recommends onnx-experimental, a rebuild, or DELETE — never `rm embedding_identity.json`. The project already ruled on this shape: CHANGELOG ~line 902 uses a separate sidecar \"precisely because ... widening it would break downgrades\".
> 
> (2) `ensure_embedding_identity_for_new_field` (index.rs:27792) still compares with the derived `PartialEq`, which now includes `mode`. `configured_onnx_identity` returns `mode: \"\"` while the create path stamps `mode: \"onnx-experimental\"`, so `PUT /_mapping` adding a semantic field to any ONNX index created after this change fails with \"refusing to add a semantic field in a different vector space\". The field's own doc comment claims it is outside the identity equality; that is true of `same_embedding_identity` and false here. Pre-existing ONNX indexes are fine — I verified the parent-format marker (no `mode` key) still opens and still accepts a new field, so the `#[serde(default)]` narrowing does hold.
> 
> Beyond that: the unmarked class includes fresh installs (any index whose semantic field arrives after ingest is never marked either), the marker records the requested mode rather than the resolved backend (so a standard→slim binary swap under `mode=neural`, an `auto` config that gains an endpoint, and a proxy that fails to init all still slip through, while lexical→neural on a slim build is refused for no runtime difference), and the new message renders as a 578-char line with 22-space runs at WARN level while omitting both the marker file and the flag to change. All four mandated checks are green on the unmodified head: slim+onnx `cargo check`, xerj-ai tests (20 and 38), `index::embedding_identity_tests` under the feature (21 passed), and clippy `-D warnings` over both crates.

**[must-fix] The upgrade is a one-way door: rolling back to the previous release makes every branch-touched semantic index red and 503 with a false "contains ONNX vectors" error**

The mode-only marker is written into the SAME file the previous release already parses (`embedding_identity.json`), with `backend: ""`, `dimensions: 0`, and `version: 1` — the identical version number the ONNX marker uses. The parent release has no notion of a mode-only marker: it reads the file, computes `configured = None` for a non-ONNX mode, and takes its `None =>` arm, which says the index *contains ONNX vectors*. Because the marker is written whenever `new_index || !has_documents`, an ordinary parent-era index that merely happens to be EMPTY is poisoned by starting the new binary once — no config change, no writes, no create. On rollback the server starts, the index fails to open, cluster health goes red, and every request to it is 503. The message's remedies are "restart with onnx-experimental" (impossible on that build), rebuild with a new autoindex state dir/prefix/brain, or `DELETE /<index>` — i.e. an operator who follows the text destroys or rebuilds a perfectly good index. The one correct action, `rm <index>/embedding_identity.json`, appears nowhere. This is precisely the failure mode the project already legislated against: CHANGELOG.md ~line 902 explains that the ingest-pipeline refusal marker was made a separate sidecar "precisely because `cluster_state.json`'s format-1 envelope is `deny_unknown_fields` and widening it would break downgrades", and notes with approval that "An older binary reading a data dir written by this one simply ignores the file". A sidecar (`embedding_mode.json`), or a `version: 2` the old code would have to reject explicitly, would have kept the downgrade path open. So no: "not protected until rebuilt" is not the only consequence — the change also removes the operator's ability to undo the upgrade.

```
data3 run, three server starts on one data dir. (1) parent 1.0.0-rc.17 (2471199), `--embed-mode lexical`, `PUT /staging` with `{"body":{"type":"semantic_text"}}`, no documents, stop → `ls data3/staging/embedding_identity.json` = No such file. (2) branch (a6556fe), same flags, one `GET /staging/_count`, stop → marker now exists: `{"version":1,"backend":"","model_sha256":"","tokenizer_sha256":"","dimensions":0,"pooling":"","max_tokens":0,"mode":"lexical"}`. (3) parent again, same flags → `_cluster/health` = red; `GET /staging/_count` = 503 `no_shard_available_action_exception`: "index [staging] failed to open and is unavailable: embedding error: index /tmp/w446-deploylens/data3/staging contains ONNX vectors but the server is configured for embedding.mode=\"lexical\". Restart with onnx-experimental and the original assets, or rebuild ... or remove it with DELETE /staging". Same result reproduced independently on data2 (branch-created index). Code: engine/crates/xerj-engine/src/index.rs:27740-27753 (sentinel write) vs the parent's `None =>` arm at the same function.
```

**[must-fix] `ensure_embedding_identity_for_new_field` still uses the derived `PartialEq`, so `PUT /_mapping` with a semantic field now fails on any ONNX index created under this branch**

`mode` was added to a struct that still carries `#[derive(PartialEq)]`, and only `validate_embedding_identity` was switched to the new `same_embedding_identity()`. `ensure_embedding_identity_for_new_field` (index.rs:27792) still does `if persisted != configured`, which now compares `mode` too. `configured_onnx_identity` deliberately returns `mode: String::new()`, while `validate_embedding_identity` stamps `mode: "onnx-experimental"` into the marker it writes at create time (index.rs:27771). So for any ONNX index created after this change, the two never compare equal and adding a semantic field via `Index::add_fields` (index.rs:18635, reached from `put_index_mapping` / `PUT /:index/_mapping` and the native schema-evolve route) is refused with "embedding identity mismatch ... refusing to add a semantic field in a different vector space. Restore the original ONNX assets or create a new index." Nothing has changed and there are no original assets to restore — the message is unactionable because it describes a condition that does not exist. The doc comment on the new field asserts it is "Deliberately outside the identity equality below"; that is true of `same_embedding_identity` and false of this call site. The existing suite cannot catch it: `later_semantic_field_pins_identity_before_first_write` only exercises ensure→ensure (mode `""` both sides). Pre-existing ONNX indexes are unaffected (their marker has no `mode`), which is why the author's third narrowing looked safe — the breakage lands on indexes created after the upgrade, not before it.

```
Probe test added to `index::embedding_identity_tests` and run with `--features onnx-experimental`: `validate_embedding_identity(dir, semantic_schema(), onnx_cfg, true, false)` writes a marker containing `"mode"`, then `ensure_embedding_identity_for_new_field(dir, same_cfg)` returns Err — `w446_add_semantic_field_to_branch_created_onnx_index ... FAILED`, panic text: `Common(EmbeddingError { reason: "embedding identity mismatch for index /tmp/.tmpCUuSp5; refusing to add a semantic field in a different vector space. ..." })`. In the same run `w446_parent_written_onnx_marker_still_opens` PASSES, confirming the legacy-ONNX case the `#[serde(default)]` exists for is genuinely safe. All 21 pre-existing tests in the module pass unchanged, i.e. the suite is blind to this.
```

**[should-fix] The unprotected class is not "indexes built before this change" — it is "any index whose semantic field arrives after documents exist", including fresh installs, and nothing tells the operator which is which**

`validate_embedding_identity` early-returns when the create-time schema has no embedding fields, and `ensure_embedding_identity_for_new_field` is a no-op for every non-ONNX mode. So the sequence create index → ingest → `PUT /_mapping` adding `semantic_text` produces a populated semantic index with no marker; on the next open `has_documents` is true, so the `new_index || !has_documents` guard declines to write one, forever. That is a brand-new index on a brand-new install, not a legacy artifact. Combined with the legacy case this means the guard's coverage is unknowable from outside: there is no startup log line, no `_cluster` field, and no `_stats` entry saying "this index has no recorded embedding mode and is therefore unguarded". An operator cannot audit which of their indexes the fix actually covers, and — per the finding above — the one artifact that would tell them is also the artifact that breaks rollback. On the upgraded legacy directory the exposure is permanent and silent: every restart re-decides not to write, so it never converges.

```
data1 run: parent built `legacy` (8 docs, semantic_text `body`, `--embed-mode lexical`); branch on the same dir with the same flags → health green, count 8, new doc accepted, `ls data1/legacy/embedding_identity.json` = No such file, before and after ingest. Branch restarted with `--embed-mode neural`, same dir: health green, zero embedding/identity lines in the log, still no marker, and the same query reorders with the spread collapsing (lexical: d6 0.6635, d1/d2/d4/d5 0.654 → neural-vs-lexical: d1 0.5306, d9 0.5277 …) — #434 reproducing verbatim through the fix. Unit probes `w446_legacy_populated_index_stays_unmarked_forever` and `w446_semantic_field_added_after_ingest_is_never_marked` both pass, the latter covering the create-without-embeddings → add-field-later path.
```

**[should-fix] The marker records the requested config string, not the resolved backend, so real embedder changes still pass and one no-op change is refused**

`embedding_mode_key()` stores `cfg.mode` lowercased. What actually produces vectors is decided later by `make_embedder()` from mode + compiled features + `default_endpoint` + whether the proxy constructs. Four consequences: (a) `mode=neural` resolves to lexical on a `--no-default-features` server (make_neural_embedder's `#[cfg(not(feature="neural"))]` arm warns and falls back "so a mode=neural config never bricks startup") — swapping the standard binary for a slim one, an ordinary upgrade action, changes the vector space while the marker reads "neural" on both sides and the guard passes; (b) `mode=auto`, the config default, resolves to lexical or proxy purely by whether `default_endpoint` is set, so adding an endpoint flips the embedder with the marker unchanged; (c) `mode=proxy` falls back to lexical when the proxy fails to initialise, so a bad restart silently writes lexical vectors into a proxy index and passes the guard next time; (d) the mirror image — on a slim build lexical→neural changes nothing at runtime, yet the guard now takes the index down and the cluster red. The correct key already exists in the same file: `embedding_execution_identity()` computes the effective `backend` ("lexical"/"neural"/"proxy"/"onnx-experimental") after all of that resolution, and there is even a test named `configured_backends_report_the_effective_runtime_backend`. Recording that string instead of `cfg.mode` closes (a)-(d) and costs nothing.

```
`w446_marker_records_requested_mode_not_resolved_backend`, run on the engine's default (no-neural) feature set: `W446 auto/local backend=lexical, auto/endpoint backend=external proxy (OpenAI-compatible /v1/embeddings)` with `validate_embedding_identity(..., auto_with_endpoint, false, true)` returning Ok; `W446 mode=neural resolves to backend="lexical" in this build` with neural→neural returning Ok; `W446 lexical->neural on this build: refused=true, actual backends "lexical" vs "lexical"`. Code: index.rs:28271-28304 (`make_embedder`), 28418-28427 (slim-build neural→lexical), 27509-27535 (`embedding_execution_identity`).
```

**[should-fix] The new operator message is a 578-character single line with 22-space runs, logged at WARN, and it omits both the file and the flag an operator needs at 3am**

The format string at index.rs:27686 was written as one physical line, so the intended line breaks are literal whitespace: the rendered message reads "but the server                      is configured for". It survives rustfmt and CI, and it reaches both the log and the JSON error body. Content problems, reading it as the on-call: the action is the fifth sentence, after four sentences of rationale; it never names `embedding_identity.json`, so nothing tells you where the recorded mode lives or that removing it is an option; it never names `--embed-mode` or `embedding.mode` in the TOML, so "restore the original mode" doesn't say which knob; it asserts unconditionally that "the widths agree (both 384)", which is simply untrue for a proxy endpoint of another width; and it ends `the new mode..` after concatenation. It is also logged at WARN by `xerj_engine::engine: failed to open index` for an event that takes the cluster red — anyone alerting on ERROR sees nothing. Finally, for an ONNX index created under this branch the new check runs first and preempts the older "contains ONNX vectors" message, which is the one that carries the actual runbook (new autoindex state directory, new prefix, new brain, `--no-graph`, validate before switching readers, clean the shared catalog) — so the case with the largest blast radius now gets the least guidance. On the positive side, recovery does work as advertised: restoring the original mode brings the index back green with all documents, and a write to a failed index is rejected rather than auto-creating an empty one over it.

```
Live 503 body from data2 under `--embed-mode neural`: "index /tmp/w446-deploylens/data2/fresh holds vectors produced with embedding.mode=\"lexical\" but the server                      is configured for embedding.mode=\"neural\". ... noise. Restore the original mode, or reindex into a new index under                      the new mode.." `cat -A` on the log line shows the same runs; a Python scan of index.rs line 27686 reports `len 578, space runs: [22, 20]`. Log level captured as `WARN xerj_engine::engine: failed to open index name="fresh"`. Recovery check: restart with `--embed-mode lexical` → health green, `_count` = 3.
```

**[nit] No test and no CHANGELOG entry for a change that adds a new startup-refusal class**

`git diff 2471199..a6556fe --name-only` is exactly one file, `engine/crates/xerj-engine/src/index.rs`, with no hunk in the `embedding_identity_tests` module. Nothing in the suite exercises the mode field in either direction — not the refusal, not the `#[serde(default)]` legacy-marker path, not the mode-only-marker branch — which is why the `ensure_embedding_identity_for_new_field` regression above compiles, lints and tests green. The repo's CHANGELOG carries a detailed Unreleased entry per issue, and this change alters what an existing data directory does on restart in both directions (new refusal, broken downgrade), so it needs one. The four mandated build checks are all clean on this head, so the CI gap here is coverage, not compilation.

```
`git diff 2471199..a6556fe --name-only` → single file; `git diff 2471199..a6556fe -- '*test*' '*.md'` → empty. `cargo check -p xerj-server --no-default-features --features onnx-experimental` Finished in 21.20s; `cargo test -p xerj-ai` 20 passed and with `--features onnx-experimental` 38 passed; `index::embedding_identity_tests --features onnx-experimental` 21 passed / 1 ignored on the unmodified branch; `cargo clippy -p xerj-engine -p xerj-ai --all-targets --features onnx-experimental -- -D warnings` exit 0 and `cargo clippy -p xerj-server --no-default-features --features onnx-experimental --all-targets -- -D warnings` exit 0, all under rustup 1.97.1 / --profile ci-test.
```


---

## Converting to draft. The design is wrong, not the details.

Two findings settle it, and neither is a patch away.

**1. I keyed on the mode STRING, not the embedder it resolves to.** This file already contains `embedding_execution_identity`, which resolves `auto` with no endpoint, `proxy` with no endpoint, `neural` without the cargo feature, and any typo to the *same lexical embedder*. `EmbeddingConfig::default()` is `mode: "auto"`, and `--help` advertises `lexical` as the default — so an index created with the shipped default and then reopened with an explicit `--embed-mode lexical` is refused, and every semantic index on the node goes red. **That is a guard that bricks the most ordinary upgrade path there is**, to prevent mixing an embedder with itself.

**2. The upgrade is a one-way door.** I wrote the mode-only marker into `embedding_identity.json`, the same file the previous release parses — which reads `backend: ""` and concludes the index *contains ONNX vectors*. Roll back and every index the new binary touched is red and 503, with an error recommending `onnx-experimental` or a rebuild. It is written whenever the index is empty, so **starting the new binary once, with no config change and no writes, is enough**. This project already ruled on exactly this shape: the CHANGELOG explains a separate sidecar was used elsewhere *precisely because* widening an existing file breaks downgrade. I did the thing the repo had already learned not to do.

The other four are real too: the guard never engages when the semantic field arrives via `PUT /_mapping` (so **#434 reproduces verbatim on this branch**); the marker is never refreshed, so a restart with an identical command line can refuse an index the previous run just wrote and served; `ensure_embedding_identity_for_new_field` still uses the derived `PartialEq` — which now includes `mode` — so adding a semantic field to an ONNX index fails with byte-identical assets; and `mode` is not an embedder identity at all, so proxy→proxy and neural→neural swaps corrupt exactly as before with the guard fully engaged.

**What the redesign has to be**, taking the verdicts at their word:

- key on the **resolved backend** from `embedding_execution_identity`, not the requested string, so `auto`/`lexical`/typo are one space and a proxy endpoint change is a different one;
- write it to a **separate sidecar**, so an older binary ignores a file it has never heard of;
- cover `ensure_embedding_identity_for_new_field` as well as `validate_embedding_identity`, and refresh the record while an index is empty rather than writing once.

The reproduction in the PR body stands and #434 is still real — both directions were confirmed here, including the reverse I had only assumed. But this branch is not the fix, and at 16/16 CLEAN it looks like one, which is why it is going to draft rather than staying open.

xerj-org:

3 skeptics (correctness / regression-risk / evidence-holds), each refuting in its own worktree. Plus a fix pushed before this pass: the refusal message had ~22-space runs at every wrap point (`this server                      resolves`) — collapsed to single spaces, wording unchanged, rustfmt clean.

> index.rs:27515-27535 resolve backend: neural-without-`neural`-feature→"lexical", proxy-empty-endpoint→"lexical", proxy-that-fails-init→"lexical" (27533-27534), unknown mode (auto) + empty endpoint→"lexical". index.rs:27541-27550: the "lexical" material string is `lexical-feature-hash.v1;dimensions={DEFAULT_DIMS};algorithm-probe={sha}` — references NO cfg fields, so canonical (27592) and thus identity_sha256 (27596) are IDENTICAL for every lexical-resolving config → no false refusal. Real neural (27572-27580, material `neural-unpinned.v1;configured-model=...`) and real proxy (27581-27589, `proxy-unpinned.v1;configured-model=...`) yield distinct canonical strings → distinct hashes → the lexical↔neural #434 swap IS caught. Refusal in validate_embedding_execution is gated: `if persisted.identity_sha256 != current.identity_sha256 && has_documents` (27770); empty index never refuses and re-records (27782-27791). Real caller Index::open (6186-6192) passes new_index=false, has_documents = segment_doc_count>0 || version_map.live_count()>0 — genuine doc count. PUT /_mapping path ensure_embedding_identity_for_new_field passes false,false to record-not-compare (27807). Tests confirm: lexical_execution_identity (27858) pins hash 58a7...; configured_backends_report_the_effective_runtime_backend (27883) asserts proxy-empty/invalid-endpoint→lexical+DEFAULT_DIMS and neural-without-feature→lexical (27901-27909, 27927-27943). Non-blocking observation: proxy material omits default_endpoint (only default_model), so a changed proxy endpoint under the same model name would NOT move the hash — a missed-detection edge, not a false refusal and not the #434 lexical/neural case, outside this claim.

**regression-risk — refuted:false, blocking:false.**

> Read index.rs:27754-27793 (validate_embedding_execution), 27650-27729 (validate_embedding_identity host + call site 27711), 27795-27837 (ensure_embedding_identity_for_new_field), 27508-27602 (embedding_execution_identity resolution), 28704-28725 (write_file_atomic), call sites 5958/6186/18635, tests 27857-28047. git diff main...HEAD --numstat = 107 insertions / 0 deletions on index.rs (purely additive).
> 
> (a) create-empty(new_index=true,has_docs=false) writes sidecar=id(A); add docs; restart under A (new_index=false,has_docs=true): path exists, persisted id(A)==current id(A) -> no refuse; rewrite guard `new_index||!has_documents` = false -> no needless rewrite. PASS.
> (b) create A w/docs then restart A: identical, no refuse, no rewrite. PASS.
> (c) stale-record: mode change requires editing server config + restart. On the restart-while-empty (has_docs=false) the mismatch branch is gated by `&& has_documents` so it never refuses, and the `!has_documents` arm REFRESHES the sidecar to the new mode BEFORE any vector under it is written. Within one process cfg is immutable (open path + self.embedding_config are process-global), so no path writes vectors under a mode different from the last-recorded one. Proven: docs written under the mode the record now reflects -> restart-with-docs compares equal. PASS.
> (d) write_file_atomic = unique same-dir temp (staging_path uses pid+seq) -> write_all -> sync_all -> rename -> parent fsync; a torn temp is never renamed, so embedding_execution.json is atomically old-or-new, never half. First-ever write failing -> path absent next open -> treated UNKNOWN (path.exists()==false), unprotected not bricked. Rewrite failing -> rename never happens, old valid sidecar intact, temp removed. The from_slice parse-error Err path is reachable only if the file EXISTS and is corrupt; a pre-feature index has no such file (new filename) so it is never parsed -> cannot brick a formerly-fine index. A write Err fails the open the same way schema.json/settings.json writes already do; no corruption. PASS.
> (e) downgrade: EMBEDDING_EXECUTION_FILE is a brand-new separate file (not embedding_identity.json); PersistedEmbeddingIdentity/EMBEDDING_IDENTITY_FILE format is byte-for-byte unchanged; older binary never opens the new file. 0 deletions confirms nothing widened. PASS.
> Ordering: validate_embedding_execution deliberately runs after the ONNX marker_path.exists() early-returns and before `let Some(identity)=configured else return Ok`. This newly runs execution validation for non-ONNX SEMANTIC indices only — the schema_has_embeddings guard at 27657 still short-circuits non-semantic indices, so scope is not broadened beyond intent. Resolved-identity keying (27515-27535) makes auto/proxy-no-endpoint/proxy-that-fails-init/neural-without-feature all resolve backend=lexical with identical canonical material -> identical sha (tests 27927-27943 assert this), so those legitimate config drifts do NOT refuse; a changed proxy endpoint/neural model/onnx asset DOES move the sha (tests 27948-27956, 28010-28047). proxy_initializes -> EmbeddingProxy::new (embed.rs:141-161) does only URL parse + scheme check + client build, no network I/O, so backend resolution is deterministic across restarts -> no flake-induced false refusal. Minor non-blocking notes: every empty-index open rewrites the sidecar (one extra fsync'd write, harmless); a reqwest client-build failure would be environment-wide, not a per-restart flake. Claim holds.

**evidence-holds — refuted:true, blocking:false.**

> Repro mechanism CONFIRMED from code. (1) configured_onnx_identity (index.rs:27482-27506) returns Ok(None) whenever mode != "onnx-experimental" (early return 27485-27487), so for a lexical/neural config validate_embedding_identity sets `configured=None`; the ONNX check at 27662 (`if configured.is_some()`) is skipped and, for a lexical index, embedding_identity.json never exists so the whole 27665 ONNX-marker branch is skipped -> the pre-existing check cannot see a lexical<->neural swap. (2) Widths agree: lexical DEFAULT_DIMS=384 (xerj-ai/src/local.rs:29; Embedder::Lexical always embeds at DEFAULT_DIMS, embedder.rs:272) and the neural default all-MiniLM-L6-v2 is 384-dim (neural.rs:28, test asserts dims()==384 at neural.rs:293) -> the dimension check cannot fire either. Both "silent" reasons hold. The collapse figures in the doc comment (0.6303..0.4783 -> 0.5496..0.5107) are asserted in the comment, not reproduced by a test, but the MECHANISM is sound.
> 
> TEST COVERAGE HALF: FALSE. The new refusal lives in validate_embedding_execution (index.rs:27754-27793), branch 27770-27780. grep across the #[cfg(test)] module and engine/crates/xerj-engine/tests/ shows its distinctive strings ("produced by embedding backend", "silently wrong ... degrades toward noise") and the PersistedEmbeddingExecution type appear ONLY in src, never in any test. NO test (i) builds an index under one embedder with documents, (ii) reopens under a different embedder, (iii) asserts the sidecar Err. changed_onnx_assets_are_rejected (index.rs:28180) is #[cfg(feature="onnx-experimental")] and exercises only the OLD embedding_identity.json ONNX-marker path: its second call hits the 27665 marker-exists branch and returns Err at 27675 BEFORE reaching validate_embedding_execution — it never touches lexical<->neural. Full embedding_identity_tests roster: lexical_execution_identity_is_stable_and_sanitized, configured_backends_report_the_effective_runtime_backend, onnx_execution_identity_changes_with_asset_content, onnx_identity_and_lazy_loader_share_one_immutable_byte_snapshot, onnx_snapshot_is_scoped_to_one_runtime_config_and_released_with_it, sequential_engines_from_one_caller_config_refresh_same_path_assets, concurrent_identity_and_loader_access_share_one_runtime_snapshot, concurrent_identity_and_real_lazy_pool_share_snapshot (ignored), onnx_identity_is_persisted_and_matching_restart_is_allowed, changed_onnx_assets_are_rejected, markerless_existing_index_allows_lexical, markerless_existing_index_fails_closed_for_onnx, later_semantic_field_pins_identity_before_first_write, onnx_rejects_non_384_semantic_mapping, onnx_configuration_fails_closed_without_compiled_feature, onnx_admission_rejection_maps_to_retryable_429, non_vector_schema_does_not_touch_onnx_assets. Every refusal test is ONNX-only; the actual #434 path (non-onnx embedder swap) has zero coverage. Such a test is cheap and needs no feature flag — lexical -> proxy-with-endpoint yields distinct identity hashes and would exercise 27770-27780. VERDICT: should-fix follow-up (add a lexical<->proxy/neural swap refusal test), non-blocking; a missing test is not a merge-blocker and the fix logic itself reads correct.

correctness and regression-risk hold; the guard catches the lexical↔neural swap #434 is about, never false-refuses a config that cannot change a vector (auto / proxy-no-endpoint / neural-without-feature all hash the same), is purely additive (107/+0), atomic-write safe, and downgrade-safe (separate sidecar an older binary never opens). The one refuted verdict is the **missing test** — should-fix, filed as a follow-up. Two non-blocking notes also filed: a changed proxy *endpoint* under the same model name would not move the identity hash (missed-detection edge, not a false refusal, not #434).

PR #447, opened 2026-08-17
docs: require operational recipe verification

buger:
Written by an AI coding agent (OpenAI Codex), run by @buger; human review of the follow-up commit is pending.

## What this changes, and why

Adds an operational-documentation and runnable-recipe audit to `docs/CONTRIBUTION_REVIEW.md` and a scoped checkbox to `.github/PULL_REQUEST_TEMPLATE.md`. The checkbox starts with `For non-trivial changes:` so typo-only and unrelated changes are exempt; applicable changes must satisfy the same fail-closed, tested-versus-unsupported, publication-copy, and explicit `N/A` criteria as the protocol.

PRs #430 and #441 showed that a plausible success path and an accidental-corruption test were not enough. A recipe also needs binary refusal criteria, explicit path trust boundaries, an adversarial pre-created-path test, lifecycle and recovery evidence, enumerable completeness claims, and verification of every published copy.

The guidance now requires:

- a negative path that exits non-zero and leaves no artifact extracted, installed, published, or served;
- artifact verification and the action it authorizes in the same runnable block, with no filesystem handoff to a privileged block;
- each negative control rerun once with the failing block's exit status discarded and once with the next documented block appended, with neither composition allowed to reach an artifact or unsafe/privileged path;
- trust boundaries and writers for every path, with attacker-precreated digest-derived paths and scratch space outside untrusted staging;
- startup, readiness, a real operation, stop, cleanup, port release, restart, persistent-state transitions, and recovery;
- an explicit unit for every count and an enumerated set for every “all”, “only”, or “no” claim, including exclusions and every publication copy;
- a statement of which feature, platform, transport, and fallback variants were tested, followed by a separate list of every unsupported variant;
- a concrete executable mitigation rather than a product capability that does not exist;
- `scripts/verify-release.sh` only for claims about an already-published release;
- an eleventh item in the copy-paste audit prompt and a matching operational evidence-table example.
- a readable, falsifiable operational checkbox scoped to non-trivial changes.

This contribution is independent and has no prerequisite or merge-order dependency.

Refs #430 and #441.

## Verified

- Final local head: `396afc69cde0dc63205f60e9b2aa57f10c1e223c` (`docs: make operational review criteria falsifiable`), on top of the two existing PR commits.
- Current `upstream/main` base and merge-base: `8d7fc764040a2473a0c99729c11a30215e7c5639`.
[...]

PR #441, opened 2026-08-16
docs(air-gapped): the recipe failed open — a bad digest extracted and installed

xerj-org:
**#430 merged with this must-fix outstanding.** It is live on `main` now.

The post-transfer verification block — the one that protects the enclave — chained with newlines and had no `set -eu`:

```sh
sha256sum -c "$ASSET.sha256"
tar -xzf "$ASSET"
test -x "xerj-${VERSION}-x86_64-unknown-linux-musl/xerj"
```

A `FAILED` checksum prints its warning and execution continues into `tar -xzf`, then into the `sudo install` in section 2.

## Measured

With a **valid** archive and a deliberately wrong `.sha256` — the shape that matters, since a corrupt archive fails at `tar` for unrelated reasons and hides the defect:

```
OLD (as shipped on main)  exit=0  extracted=YES
NEW (this fix)            exit=1  extracted=no
```

`exit=0` is the part that bites: a wrapping script or an agent reads success from a run that installed an unverified binary.

The connected-machine block immediately above was **already** guarded with `set -eu` and a subshell, which is what makes this a gap rather than a style preference — the two blocks disagreed, and the unguarded one runs inside the enclave. `bash -n` parses both happily, which is why #430's own syntax check passed.

Fixed in both copies (`docs/` and `landing/docs/`).

## Still outstanding from #430's verification

Two must-fix items merged unaddressed and are **not** fixed here:

- the Console external-asset count is **9**, not 3 — `login.html` and `setup.html` carry the same three Google Fonts elements as `index.html`, and they are the *first* pages an operator hits on a fresh node
- section 3's lexical→neural switch silently corrupts an existing index (#434), and the recipe walks an operator into it

**Not verified:** I have not re-run the font count or the neural-switch reproduction on the merged head; both are carried from the verification of #430.

Issue #440, opened 2026-08-16
scroll's point-in-time view is torn: _seq_no is read in a second pass, so a hit can carry one moment's _source beside another's sequence number (silent lost update)


xerj-org:
Blocks #431, and no amount of work at the API layer will close it.

## The defect

A scroll is supposed to be a point-in-time view. In xerj the hit list is frozen but its meta-fields are not read at the same instant, so a hit can carry `_source` from one moment and `_seq_no` from another. Feeding that `_seq_no` back as `if_seq_no` then passes the CAS and overwrites whatever landed in between — a silent lost update, on the reindex/migration/CDC path scroll exists to serve.

Two rounds of repair at the API layer have now been refuted, each narrowing the window rather than closing it:

1. resolve the meta-fields at **render time** — window is the whole scroll lifetime
2. resolve them into `ScrollHit` at **scroll-open time** — window is the duration of the opening request, and on a multi-index scroll it still spans every subsequent index's search

Reproduced on a clean node at the second state, first attempt:

```
continuation: zzhot  _seq_no=1787   _source.round=201000
live:                                 round=201030
PUT /ra/_doc/zzhot?if_seq_no=1787&if_primary_term=1  (stale body)  ->  200
round=201030 gone
```

## Why the API layer cannot fix it

`merged_hits` is filled inside the per-index search loop (`idx.search(...)`, one iteration per index, sequential). The meta-fields are resolved afterwards, in a second pass over the live version map. Any write between those two reads pairs a new sequence number with an old body, and there is no ordering the API layer can impose that removes the gap — the two reads are of different structures at different times.

## What actually closes it

The value must come from **the same read that produced the hit**. That is what Lucene gets structurally: a searcher holds a pinned `IndexReader`, and `_seq_no` doc-values are written at index time and immutable per segment, so the document and its sequence number are read out of one frozen segment. There is no second pass to be raced.

Concretely for xerj: `xerj_query::executor::Hit` carries `seq_no` / `version`, populated inside `Index::search` from the same state the document body was read from. Scroll then needs no metadata pass at all — `ScrollHit` becomes redundant, because the hit already carries a self-consistent set — and plain `_search` gets the same guarantee for free, which it does not have today either.

## Adjacent, smaller, and separate

[...]

PR #448, opened 2026-08-17
docs: compare XERJ design with Lucene 10.3.1 (#373)

### LINKED ISSUE #373: docs: no side-by-side summary of how XERJ's design differs from Lucene
Raised by **Gregg Donovan** (Etsy), who reviewed the project and fed this back directly.

The docs state that XERJ does not inherit Lucene's design, but never say *how*. As Gregg put it, that claim means little without a diff next to it. Anyone arriving from a Lucene/Solr/ES background has to reverse-engineer the differences from source.

Wanted: a short page contrasting the two on the axes a Lucene person actually cares about, e.g. segment and merge model, postings layout, doc-values equivalent, analysis chain, update/delete semantics, and where the vector index sits relative to the inverted index.

Note for whoever picks this up: partial updates *are* supported (`/_update/{id}`, `/_update_by_query`) and that was one of the specific things he could not determine from the docs, which is evidence the gap is real.

Thanks to Gregg for the review.

buger:
Written by Codex (GPT-5.6 Luna), researched and independently reviewed by Codex (GPT-5.6 Sol), and run by human operator @buger. @buger reviewed the original published head; human review of this review-driven follow-up head is pending.

Branch under review: `docs/issue-373-xerj-vs-lucene` at `e7935c1e2404bbd15873a1a686ff56c9ac8a1192` (rebased two-commit branch), based directly on authoritative upstream `main` `8d7fc764040a2473a0c99729c11a30215e7c5639`.

## What this changes, and why

Issue #373 requested a short, side-by-side explanation of how XERJ's design differs from Lucene 10.3.1. The first draft established the six source-linked axes and corrected stale XERJ ingest/flush architecture prose. Independent review found that the page was technically accurate but hard to discover, not actually side-by-side at a glance, unpinned on the XERJ side, and too quiet about several material Lucene strengths.

This follow-up addresses that review without changing runtime behavior:

- Restores the repository-authoritative statement that XERJ shares no code and no architecture with Elasticsearch or Lucene.
- Links the comparison from both the README subsystem references and the agent-facing `AGENTS.md` navigation table.
- Opens the page to humans and agents, adds a compact six-row side-by-side table, and pins the comparison to 2026-08-17, Lucene 10.3.1, and XERJ source snapshot `24711999dd866ceec2a6e7c91d934c2c27d7066c`.
- Replaces relative XERJ line links with immutable GitHub permalinks, points the merge path at `apply_merge_with_repoints`, and labels the in-function vector range as kNN dispatch rather than `search_at_generation`.
- Adds the Lucene behavior needed for a fair comparison: competitive postings impacts, BKD points for numeric/geo ranges, analysis-module and `CustomAnalyzer` breadth, no-argument `StandardAnalyzer` parity, filter-aware HNSW with automatic exact fallback, and scalar/binary quantized vector formats.
- Makes the scope explicit: facets, near-real-time reader management, and backward-codec compatibility are not compared by the six detailed axes.
- Keeps XERJ's existing caveats prominent: lexical feature-hashing is the default embedding fallback; persisted postings skips are absent; custom query-analyzer projection is limited; filtered vector queries use the exact path; HNSW candidate selection remains approximate; and `_update_by_query` has a 10,000-hit fetch cap.
- Clarifies the shard domains without documenting the per-index `xerj_ingest_shards` behavior: global `engine.ingest_shards` is power-of-two validated; the storage memtable independently uses a rounded power-of-two count for hash-mask routing; and the engine-side FTS memtable uses the global setting.

The effective diff contains exactly five documentation files: `AGENTS.md`, `README.md`, `docs/ARCHITECTURE.md`, `docs/WHY_XERJ.md`, and `docs/XERJ_VS_LUCENE.md`. The PR is independent and has no prerequisite or merge-order dependency.

## Verified

- Current authoritative upstream `main` and the branch base are both `8d7fc764040a2473a0c99729c11a30215e7c5639`; the branch has exactly two commits on that base after rebase.
- `docs/XERJ_VS_LUCENE.md` contains exactly six numbered headings and exactly six side-by-side table rows.
- All 69 unique remote URLs in `docs/XERJ_VS_LUCENE.md` and `docs/ARCHITECTURE.md` returned HTTP 200 with `curl -L`, including every official Lucene 10.3.1 reference and immutable XERJ permalink.
- Every pinned XERJ source range exists at the named snapshot; focused range checks found each cited symbol or behavior, including `Index::flush`, `do_flush_shard`, `apply_merge_with_repoints`, postings encoding, BKD-comparison inputs, analyzer projection, update handlers, HNSW persistence, and kNN dispatch.
- Every relative link in the five touched files resolves, and neither `docs/XERJ_VS_LUCENE.md` nor `docs/ARCHITECTURE.md` contains a relative source link with a line anchor.
- `npx --yes markdownlint-cli docs/XERJ_VS_LUCENE.md --disable MD013 MD034 MD041 MD060` passes. The disabled rules are repository-inapplicable line-length, bare-URL, first-line, and table-style preferences; structural Markdown checks remain enabled.
- `git diff --check upstream/main...HEAD` passes, and the effective diff is exactly the five documentation files listed above.
- Both commits are authored by `buger <leonsbox@gmail.com>`, include full motivation/evidence bodies, and contain no `Co-Authored-By` trailer.

## Assumed

- The official Lucene 10.3.1 documentation is authoritative for Lucene behavior. No claim of binary, file-format, or codec compatibility is made.
- Because no runtime files changed, the source-linked architecture review is the relevant validation boundary; a future XERJ source change requires a new snapshot rather than silently moving these links.

## Not run

- Cargo formatting, builds, crate tests, and the ES-YAML conformance suite were not run because the effective diff is documentation-only and changes no Rust or ES behavior.
- No Lucene runtime benchmark was run because this is a source-linked architecture comparison, not a performance or behavioral compatibility claim.
- No browser-specific visual-regression suite was run; the comparison was checked with Markdown linting, link validation, exact table/heading counts, and source review.

## Review status

The original draft was reviewed by @buger before publication. The exact follow-up head above has not yet received final human review and should remain draft until @buger reviews it.

Closes #373


xerj-org:

Three skeptics, each told to refute: one on every XERJ claim against the code, one on every Lucene claim against 10.3.1, one on coherence with the rest of the docs. **Two of three could not refute it.** Verbatim below, then where I land.

### Verdict 1 — refuted: `True`

> Split verdict. The factual half of the lens passes cleanly: I could not refute a single claim the document makes about Lucene. All 14 lucene.apache.org links resolve HTTP 200 at the `10_3_1` javadoc tree; nothing is attributed to Lucene that is actually Elasticsearch or Solr (every cited symbol — IndexWriter, MergePolicy, ConcurrentMergeScheduler, Lucene103PostingsFormat/Writer, Lucene90DocValuesFormat, Analyzer, StandardAnalyzer, LiveDocsFormat, KnnFloatVectorField/Query, KnnVectorsFormat — is Lucene core); and nothing is described as a Lucene limitation that is really configurable or fixed in 10.x. The version-specific details are right to the constant: BlockTree + 128-int packed blocks + "skip data interleaved at 2 levels" matches the Lucene103PostingsFormat javadoc almost verbatim (BLOCK_SIZE=128, LEVEL1_FACTOR=32); Lucene90DocValuesFormat is genuinely the DV format wired into Lucene103Codec and does document .dvd/.dvm; ConcurrentMergeScheduler is genuinely the LiveIndexWriterConfig default; the updateNumericDocValue/updateBinaryDocValue characterization is a near-quote of IndexWriter's own javadoc. This is more careful sourcing than most vendor comparisons manage, and the accompanying softening of "shares no code or architecture" to "no shared code or file formats" in ARCHITECTURE.md and WHY_XERJ.md is a real accuracy improvement, since XERJ plainly does share architecture (segments, FST term dict, 128-doc PFOR blocks, doc values, HNSW).
> 
> The fairness half is where it fails, and it fails by omission rather than by falsehood. Within each axis the XERJ paragraph is detailed and genuinely self-critical (linear scan instead of serialized skips, HNSW eligibility fallbacks, the 10k cap on update_by_query, analyzer restrictions) while the Lucene paragraph is two-to-four sentences of generic description that never names a Lucene strength. The net effect is that XERJ's honest concessions land next to a Lucene column that gives the reader nothing to weigh them against: the doc says filters take XERJ's brute-force path without mentioning that Lucene 10.3.1 runs kNN subject to the filter inside the graph; it says XERJ's skip entries are unserialized scaffolding without mentioning that Lucene's skip data carries impacts for block-max pruning; it compares XERJ's numeric columns against Lucene DocValues when Lucene's numeric-range answer is BKD points, a structure the document never mentions. Vector quantization, the facets module, NRT/SearcherManager, and Lucene's cross-major-version back-compat guarantee (visible as thirteen codec generations under backward-codecs) are absent entirely. Finally, the document pins Lucene precisely to 10.3.1 but pins XERJ only to "current main" — no date, no version, no SHA — while using line-number anchors like #L18002-L18125 that rot on the first refactor with no signal to a future reader. None of this makes the document wrong; it makes it a document that concedes only on one side, which is exactly the failure mode that matters for an artifact in docs/ that people will quote.

**[should-fix] Axis 6 omits that Lucene 10.3.1 does filtered kNN inside the HNSW graph — XERJ's brute-force fallback reads as a design choice, not a gap**

The XERJ side honestly states "Filters, nested/passage shapes, unsupported metrics, small or stale graphs, and any failed eligibility check use the exact brute-force path." The Lucene paragraph directly above it never mentions filtering at all, so a reader has no way to learn that Lucene handles the same case with an approximate path. Lucene 10.3.1's KnnFloatVectorQuery takes a filter Query and its class javadoc describes the policy explicitly: "If the filter cost is less than k, just execute an exact search. Otherwise run a kNN search subject to the filter. If the kNN search visits too many vectors without completing, stop and run an exact search." The implementation is FilteredHnswGraphSearcher. Filtered kNN is the dominant production vector shape (tenant scoping, ACLs, recency windows), so this is the single most consequential unstated difference in the document. One sentence in the Lucene paragraph — that Lucene applies the filter during graph traversal with an automatic exact fallback — would make the comparison fair without weakening anything XERJ actually claims.

```
https://lucene.apache.org/core/10_3_1/core/org/apache/lucene/search/KnnFloatVectorQuery.html (HTTP 200) documents constructors KnnFloatVectorQuery(String,float[],int,Query) and KnnFloatVectorQuery(String,float[],int,Query,KnnSearchStrategy); /home/claude/ref/lucene/lucene/core/src/java/org/apache/lucene/util/hnsw/FilteredHnswGraphSearcher.java; docs/XERJ_VS_LUCENE.md axis 6.
```

**[should-fix] Axis 3 measures XERJ's numeric columns against Lucene DocValues, but Lucene's numeric-range structure is BKD points, which the document never mentions**

The doc writes that XERJ's "numeric columns retain a sorted (value, doc_id) index for range work" and that the columns "feed XERJ sorting, filters, ranges, and aggregation fast paths." Its Lucene counterpart paragraph covers only Lucene90DocValuesFormat and SortedSetDocValuesField. That framing invites the conclusion that the two engines answer numeric ranges with comparable machinery. They do not: Lucene answers numeric and geo ranges with a dimensional BKD point tree (.kdd/.kdi/.kdm via Lucene90PointsFormat, driven by LongPoint/PointRangeQuery), a block-balanced k-d tree with per-node value ranges that prunes whole subtrees, and it uses DocValues mainly for sorting, faceting and value retrieval. Points are not an axis in this document and are not named anywhere in it, so the strongest range structure Lucene has is invisible. Either name BKD in the axis-3 Lucene paragraph or state plainly that points have no XERJ counterpart.

```
All HTTP 200 at 10.3.1: .../search/PointRangeQuery.html, .../document/LongPoint.html, .../codecs/lucene90/Lucene90PointsFormat.html. `grep -rn 'points\|BKD\|kdd' docs/XERJ_VS_LUCENE.md` returns nothing.
```

**[should-fix] Document is undated and unpinned on the XERJ side while pinning Lucene to 10.3.1, with line-number anchors that rot silently**

The doc names Lucene 10.3.1 fourteen times but describes XERJ only as "the runtime on current `main`" with no date, no release, and no commit SHA — the repo is at 1.0.0-rc.17. Every XERJ link is a relative path with a hard line range (#L18002-L18125, #L24433-L24920, #L37926-L37975). Those anchors will drift on the first refactor of index.rs and there is nothing in the page telling a reader a year from now which half has gone stale — the Lucene half is pinned and durable, the XERJ half is unpinned and fragile, which is the opposite of what a reader would assume. The line ranges are broadly correct today (I spot-checked five; do_flush_shard, AnalyzerRegistry, Column and the ShardedFtsMemtable banner all land on the named symbol, Index::flush at L18002 lands mid-signature), which is exactly why the rot will be hard to notice later. Add a header line: written <date> against XERJ <version>/<sha> and Lucene 10.3.1.

```
`grep -n 'date|2026|as of|written|commit' docs/XERJ_VS_LUCENE.md` matches nothing relevant; engine/Cargo.toml line 24: version = "1.0.0-rc.17"; PR head 213ec22.
```

**[should-fix] Axis 4 gives Lucene three generic sentences and never states the analysis-module breadth that XERJ's own concessions are being measured against**

The XERJ side of axis 4 is admirably specific about its limits: the registry starts with built-ins, unknown/custom analyzer names "decline to stored evaluation," match_phrase on analyzed text "requires `standard` and zero slop," and the page explicitly disclaims "general index/query custom-analyzer symmetry." The Lucene side says only that Analyzer creates a reusable token stream and StandardAnalyzer is one concrete preset. Lucene 10.3.1 ships lucene-analysis-common with roughly forty language packages plus Snowball/Porter/KStem stemming, hunspell, synonyms, shingles, ngram, commongrams and minhash, and separate icu, kuromoji, nori, smartcn, opennlp, phonetic, stempel and morfologik modules, with CustomAnalyzer.builder() for arbitrary composition. A reader cannot tell from this page whether XERJ's restrictions are a small delta or a large one. Stating Lucene's breadth costs one sentence and makes the honest XERJ paragraph readable.

```
/home/claude/ref/lucene/lucene/analysis/ contains common, icu, kuromoji, morfologik, nori, opennlp, phonetic, smartcn, stempel; lucene/analysis/common/src/java/org/apache/lucene/analysis/ contains ar bg bn br ca cjk ckb commongrams compound cz da de el en es et eu fa fi fr ga gl hi hu hunspell hy id it lt lv minhash morph … .
```

**[should-fix] Axis 2 concedes XERJ's linear scan but frames Lucene's two-level skip data as a format detail, hiding that it carries impacts for block-max pruning**

The doc correctly reports that XERJ's SkipEntry "is currently in-memory scaffolding and is not serialized, so advance_to scans linearly," then closes with "Do not infer Lucene's two-level skip behavior from the shared block size." That reads as a caution about codec compatibility. The actual consequence is a query-evaluation one: the very javadoc the document cites states that Lucene's skip data stores ImpactLength with CompetitiveFreqDelta/CompetitiveNormDelta pairs "used to safely skip score calculation for uncompetitive documents" — i.e. impacts feeding block-max WAND/MAXSCORE top-k pruning. That is the largest structural performance difference on this axis and it is present in the source the author already read. Since XERJ's side is disclosed anyway, naming what Lucene gets for it costs the document nothing and buys it credibility.

```
/home/claude/ref/lucene/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/lucene103/Lucene103PostingsFormat.java:221-224 (ImpactLength, CompetitiveFreqDelta/CompetitiveNormDelta, CompetitiveImpactAccumulator); same file lines 60-61 and 342-350 (BLOCK_SIZE=128, LEVEL1_FACTOR=32, LEVEL1_NUM_DOCS=4096) confirm the document's own skip/block claims as accurate.
```

**[nit] Axis 6 omits vector quantization entirely, in a comparison whose stated audience is AI workloads**

The axis describes where each engine puts its vectors but says nothing about how they are represented. Lucene 10.3.1 ships Lucene99HnswScalarQuantizedVectorsFormat (int8/int4 scalar quantization) and Lucene102HnswBinaryQuantizedVectorsFormat (BBQ, roughly 32x compression), both documented at the 10.3.1 javadoc URLs the author is already using. XERJ's HNSW is described only as an index-level persisted artifact with no representation or footprint discussion. For a document aimed at engineers evaluating an AI-native engine, memory footprint per vector is a first-order question, and its absence — combined with the absence of Lucene's per-segment vector merge lifecycle versus XERJ's index-level graph — leaves the axis reading as pure placement trivia.

```
HTTP 200 at 10.3.1 for .../codecs/lucene99/Lucene99HnswScalarQuantizedVectorsFormat.html and .../codecs/lucene102/Lucene102HnswBinaryQuantizedVectorsFormat.html; `grep -in 'quantiz' docs/XERJ_VS_LUCENE.md` returns nothing.
```

**[nit] The six axes are not rigged, but every omitted axis favors Lucene — most notably index format stability**

To be fair to the contributor: within the six chosen axes XERJ concedes repeatedly, so the 'wins by construction' charge does not stick to the axes themselves. The problem is the selection. Nothing is chosen where Lucene would dominate outright: points/BKD, dynamic pruning, the facets module, NRT via DirectoryReader.open(IndexWriter)/SearcherManager and concurrent IndexSearcher, and above all format stability — Lucene carries thirteen generations of codecs in backward-codecs so an index written by an older major can still be read, whereas this document makes no statement whatsoever about .seg format stability across XERJ releases. For a page whose non-claims section is otherwise thorough, one short 'where Lucene is stronger' section would do more for the document's authority than any of the six axes does.

```
/home/claude/ref/lucene/lucene/backward-codecs/src/java/org/apache/lucene/backward_codecs/ contains lucene80 lucene90 lucene91 lucene912 lucene92 lucene94 lucene95 lucene99 lucene100 lucene101 lucene102 lucene103; HTTP 200 at 10.3.1 for .../search/SearcherManager.html, .../index/DirectoryReader.html, and /core/10_3_1/facet/.../facet/package-summary.html.
```

**[nit] Axis 4 presents XERJ's stopword/stemming absence as a caveat without noting Lucene's default is identical**

The doc says XERJ's `standard` preset is "Unicode word splitting plus lowercase, with no stop-word removal or stemming." In context that reads as a deficit versus Lucene. Lucene's no-arg StandardAnalyzer() constructs StandardTokenizer + LowerCaseFilter + StopFilter with CharArraySet.EMPTY_SET and applies no stemming — functionally the same default. This is a rare case where the document is harder on XERJ than the facts require; a half-clause noting the parity would be both fairer and more useful, and would sharpen where the real analyzer delta actually is (module breadth and custom composition, per the axis-4 finding above).

```
/home/claude/ref/lucene/lucene/core/src/java/org/apache/lucene/analysis/standard/StandardAnalyzer.java:51-52 — `public StandardAnalyzer() { this(CharArraySet.EMPTY_SET); }`; line 88 `TokenStream tok = new LowerCaseFilter(src);`.
```

> I could not refute the PR's checkable claims — I tried hard to. Every XERJ internal assertion I tested against source held, including the two the task flagged. The headline correction is right and important: `Index::flush` (index.rs:18012+) sets `n_shards = self.memtable.shard_count()` and `tokio::spawn`s one `do_flush_shard` per FTS shard; the aggregate-drain path the old ARCHITECTURE.md described is labelled "Drain the memtable and finalise in one shot (legacy path)" in-source at index_store.rs:1875 and has no non-test callers. So the prose this PR deletes was genuinely stale. Default embeddings are lexical feature-hashing with `--embed-mode neural` opt-in and download-on-first-run (embedder.rs:7-24, neural.rs:225, server/main.rs:228). The three-domain shard story (global power-of-two ≤256; per-index `1..=256` WAL override; storage memtable rounding to next power of two; one shared `AtomicU64` seq counter) is exactly what config.rs:1560-1570, index.rs:28502-28507 and index_store.rs:660-703 do. `update_document` really is a flat top-level overlay, not an ES-style deep merge (index.rs:13157-13165) — disclosing that is a genuine service to a Lucene/ES reader. The `_update_by_query` 10,000 cap, the `Column` enum being numeric+keyword only, the "unknown/custom analyzer names decline to stored evaluation in this *default* registry" nuance (`AnalyzerRegistry::default()` + `?` at index.rs:37953), the HNSW eligibility gates, and the delete-tombstones-the-HNSW-node claim all check out. The postings section is the most impressive part: "SkipEntry is in-memory scaffolding, never serialized, so advance_to scans linearly" matches postings.rs:31-40 and :700-708 verbatim, and it volunteers a weakness rather than hiding it. The Lucene side also cross-checks: against /home/claude/ref/lucene, `Lucene103Codec` pairs `Lucene103PostingsFormat` with `Lucene90DocValuesFormat`, and `BLOCK_SIZE=128` / `LEVEL1_FACTOR=32` confirm both "128 integers in the packed path" and "skip data interleaved at two levels". All 36 relative links resolve, all line ranges are in-file, and sampled 10.3.1 javadoc URLs return 200. My findings are therefore coherence and editorial, not factual. WOULD I MERGE: yes — but not as-is. Three things must land with it or the work is wasted: (1) the PR silently downgrades the project's flagship claim from "no code and no architecture" to "no shared code or file formats" in the two docs that carry it, while AGENTS.md:5 still says the old thing, so main would ship self-contradicting; (2) nothing outside the two edited files links the new page, including the two indexes a Lucene engineer or an agent actually starts from; (3) issue #373 asked for a "side-by-side summary" and got six prose sections with no side-by-side element and an opening line that tells the human who raised the issue he is not the audience.

**[must-fix] PR weakens the project's flagship "no shared architecture" claim in two docs but leaves AGENTS.md asserting the old version**

docs/WHY_XERJ.md and docs/ARCHITECTURE.md are changed from "shares no code and no architecture with Elasticsearch or Lucene" to "independent implementation with no shared code or file formats". That is a real positioning change — "no file formats" is a narrower promise than "no architecture" — and it is arguably the honest one, because the new page itself documents 128-doc blocks "matching Lucene's default", an FST term dictionary, a char-filter/tokenizer/token-filter pipeline, and a segment+merge model. But AGENTS.md:5 and tools/xerj-code/hub/README.md:46 still carry the strong form, and AGENTS.md is explicitly the file that instructs agents how to summarize this repo. After merge, the two loudest docs disagree about the project's central differentiating claim, and an agent reading AGENTS.md will keep restating the retracted version. Decide once: either propagate the softer wording to AGENTS.md (and the hub README) in this PR, or revert the wording change and keep the PR scoped to the Lucene comparison. Note CONTRIBUTING.md:99 and AGENTS.md:64 hang licence hygiene on the narrower "shares no code with ES", which survives either way, so this is a positioning decision, not a legal one.

```
docs/WHY_XERJ.md:5 ("no shared code or file formats") and docs/ARCHITECTURE.md:11-13 vs AGENTS.md:5 ("It shares no code and no architecture with Elasticsearch or Lucene") and tools/xerj-code/hub/README.md:46 ("public position is that it shares no code and no architecture with it").
```

**[must-fix] The new page is unreachable from every index a reader or agent actually starts at**

Inbound links are only from docs/ARCHITECTURE.md:17 and docs/WHY_XERJ.md:7. README.md has a section literally titled "Reference pages for individual subsystems, written against the source and including the limits each one does not lift" (README.md:286-298, listing SECOND_BRAIN/SCRIPTING/SNAPSHOT_AND_RESTORE/SECURITY_MODEL) — that is a precise description of this page, and it is not in the list. AGENTS.md has a "Where to look" table (AGENTS.md:162-176) that routes agents to ARCHITECTURE.md and CONTRIBUTION_REVIEW.md — also missing. The self-defeating part: the page's own first sentence is "If you are an AI agent deciding where to look in XERJ, use this page as a source map," and the one table built for exactly that purpose does not list it. Issue #373 was filed because a reviewer could not find this information in the docs; a page that only two other docs link does not close that. Adding two table/list rows costs nothing and is the difference between shipped and buried.

```
grep -rn "XERJ_VS_LUCENE" over the tree returns only docs/ARCHITECTURE.md:17 and docs/WHY_XERJ.md:7. README.md:286 opens the subsystem-reference list; AGENTS.md:162 opens the "Where to look" table.
```

**[should-fix] Issue #373 asked for a side-by-side summary; the page delivers prose with no side-by-side element**

#373's title is "docs: no side-by-side summary of how XERJ's design differs from Lucene" and the body asks for "a short page contrasting the two on the axes a Lucene person actually cares about". All six requested axes are present and each is answered correctly, so the substance is there — but there is not one table, not one two-column layout, nothing a reader can scan. Each axis is a Lucene paragraph followed by a XERJ paragraph, and the reader has to hold both in their head to extract the contrast the issue asked for. Compounding it, the framing sentence addresses "an AI agent deciding where to look in XERJ" and calls the page a "source map", which tells Gregg Donovan — the human Lucene/Solr engineer whose review produced the issue — that he is not the audience, in the very document written for him. This is the single most valuable edit: put a six-row table (axis | Lucene 10.3.1 | XERJ) at the top and let the prose sections be the drill-down. That one change makes it satisfy the issue as written and serves both audiences instead of choosing one.

```
Issue #373 title and body ("Wanted: a short page contrasting the two on the axes a Lucene person actually cares about"); docs/XERJ_VS_LUCENE.md:3-9 ("If you are an AI agent deciding where to look in XERJ, use this page as a source map"); no `|` table markup anywhere in the 172-line file.
```

**[should-fix] 36 line-number anchors into churning files, with no CI check to catch rot**

The page pins 36 links of the form `index.rs#L18002-L18125` into files up to 44,373 lines. All 36 resolve today and all ranges are in-file — I verified every one. But there is no link or anchor check anywhere in CI: .github/workflows has ci.yml/release.yml/release-metrics.yml/onnx-*, and the only doc-freshness gate in the repo is the `docs_capability_lists` test that guards ROADMAP.md's review line. index.rs is the most-edited file in the engine; these anchors will silently point at unrelated code within weeks, and this is precisely the page people will quote. The page's closing "re-check the linked symbols when changing the engine" is an honour system with no enforcement. Either anchor by symbol (link the file, name `fn do_flush_shard` in the prose) or add a cheap test that greps each cited range for the symbol name — the repo already has that pattern in engine/crates/xerj-api/tests/scroll_snapshot_cap_is_documented.rs and engine/crates/xerj-mcp/tests/published_schema_drift.rs. Small supporting symptom of the fragility: the same symbol gets two different anchors inside this one PR.

```
docs/ARCHITECTURE.md:121 links `memtable.rs#L628-L709` while docs/XERJ_VS_LUCENE.md:30 links `memtable.rs#L628-L700` for the same `ShardedFtsMemtable`. `ls .github/workflows` shows no link-check job; `grep -rln "check_links|markdown_links"` over engine/ scripts/ .github/ returns nothing.
```

**[nit] ARCHITECTURE.md documents the non-power-of-two WAL override as a feature; the WAL router masks as if it were a power of two**

The new prose says the per-index override "accepts any integer in `1..=256`, not only powers of two". That is an accurate reading of `wal_shards_override_from_settings` (index.rs:28502-28507), and `IndexStore::open` does allocate exactly N writers. But routing is `xxh3_64(doc_id) & (self.wal_shards.len() - 1)` (index_store.rs:1633) — a power-of-two mask. With N=3 the mask is 0b10, so only shards 0 and 2 ever receive documents and shard 1's WAL file is opened and never written; N=5 gives {0,4}. No panic and no data loss (`replay_all_sorted` is self-describing, and the doc comment at index.rs:28499 says routing is `xxh3(id) & (num_shards-1)`, showing power-of-two intent), so this is skew, not corruption. But stating the flexibility as designed behaviour in the architecture doc — with no caveat — turns a latent inconsistency into documented API. Either add half a sentence ("values that are not powers of two leave some WAL shards unused") or open an issue and leave the claim out. This is not the PR's bug; it is the PR making someone else's bug look intentional.

```
docs/ARCHITECTURE.md:106-110 vs engine/crates/xerj-storage/src/index_store.rs:1630-1637 (`fn wal_shard_for`: `(h as usize) & (self.wal_shards.len() - 1)`) and engine/crates/xerj-engine/src/index.rs:28499.
```

**[nit] Conformance de-numbering is correct and the PR is right where other docs are wrong — but README's badge is now the last hardcoded number in a top-level doc**

Reporting this in the PR's favour. ARCHITECTURE.md's "1,360 of 1,363" and WHY_XERJ.md's "1,360/1,363" were stale, and replacing them with the zero-failures gate is exactly what ROADMAP.md:18 ("read the current number off CI, not off this file") and AGENTS.md:66 ("gate on failures, not on an exact total") instruct. The surrounding docs currently disagree with each other three ways, which proves the point. Untouched by this PR, and worth a follow-up rather than blocking: README.md:10 still renders a hardcoded ES-conformance badge and README.md:237 a hardcoded count, and they are now the only numbers left in the top-level docs — the exact failure mode the PR is fixing elsewhere.

```
README.md:10 badge `ES%20conformance-1366%2F1369` and README.md:237 ("currently passes 1366 of 1369"); ROADMAP.md:18 (1,366/1,369); AGENTS.md:5 (1,365/1,368) and AGENTS.md:66 (1365 passed / 0 failed / 3 skipped); demo/playbooks/rc12/GATE-conformance.md:43 ("1365 is the current true count").
```

**[nit] Internal line-wrapping is inconsistent, which reads as amendments bolted onto a wrapped draft**

The file mixes ~78-column wrapped paragraphs with single lines up to 363 characters, and the long lines cluster exactly where the commit message says late amendments landed (the analyzer projection, match_phrase, postings sidecar list, HNSW persistence). Renders identically on GitHub, so purely cosmetic — but it makes future diffs on this file noisy, since editing one sentence in a 363-char line rewrites the whole line. ARCHITECTURE.md, the file this sits next to, wraps consistently at ~78. Worth one reflow pass before merge.

```
docs/XERJ_VS_LUCENE.md max line length 363; long (>120 col) lines at 11, 13, 14, 21, 23, 27, 46, 47, 50, 71, 72, 83, 92, 96, 97, 98, 102, 103, 131, 139, 146, 147, 154, interleaved with wrapped paragraphs elsewhere.
```

### Verdict 3 — refuted: `False`

> **Not refuted. I could not find a single false claim about XERJ internals in `docs/XERJ_VS_LUCENE.md`.** I checked all six axes claim-by-claim against the code at 213ec22. This is an unusually careful document — it is more accurate than several code comments sitting next to the symbols it cites.
> 
> **Verdicts on the claims flagged as easy-to-state / hard-to-get-right:**
> 
> - **Flush path — TRUE.** `Index::flush` (index.rs:18012) really does fan out one `do_flush_shard` task per FTS shard (`n_shards = self.memtable.shard_count()`, spawned in a loop at 18049-18095); `do_flush_shard` (24433) drains its own shard (`memtable.drain_shard_accounted`), builds the FTS sidecar via `FtsIndexWriter` and the `.dv` columns, then calls `finalize_flush_with_publisher` (index.rs:24829 → index_store.rs:1901) which writes the `.seg` and publishes via `ArcSwap` (index_store.rs:1054). The legacy aggregate drain (`take_memtable_for_flush` → `flush_with_publisher`, index_store.rs:1813/1871) still exists but at HEAD has **no non-test caller on the engine's user-visible flush path**. The PR's central checkable claim is correct.
> - **Per-shard seq ordering — TRUE.** `drain_shard_inner` ends with `out.sort_by_key(|(seq, _, _, _)| *seq)` (memtable.rs:1978).
> - **Default embedding — TRUE.** `embedding.mode` defaults to `"auto"` = proxy-if-endpoint-configured, else lexical (config.rs:1253, 1327); `Embedder::Lexical` is the feature-hash backend (embedder.rs:7-10). Neural BERT is `--embed-mode neural` (main.rs:587) and downloads on first use (neural.rs:225-273). The shipped binary has `default = ["neural"]` (xerj-server/Cargo.toml:69), so "opt-in at runtime" is right. The doc states no dimensionality, so there is nothing to get wrong there — the 384-dim figure lives only in `xerj-autoindex/src/catalog.rs:116`.
> - **Segments / merge / deletes / version map — TRUE.** `SectionType::Tombstones` payload really is sequence-aware ZTB2 `(seq_no, doc_id)` pairs (segment.rs:130-163), not the roaring bitmap its own enum comment claims. Merge filters stale copies by `entry.seq_no > doc_seq` and purges tombstoned docs (merge.rs:311-325). `VersionMap` tracks newest seq + `deleted` (version_map.rs:59-66); the "ghosts until merge" wording matches `ghost_events` (version_map.rs:103-118).
> - **Postings — TRUE, including the honest parts.** `BLOCK_SIZE = 128` (postings.rs:46); `SkipEntry` is computed and discarded, never serialised, and `advance_to` is a documented linear scan (postings.rs:31-40, 698-712). The doc could have quietly omitted this; it did not.
> - **Scroll / seq_no / point-in-time — correctly absent.** I looked specifically for this. The document makes **no** claim about scroll, PIT, or `_seq_no` visibility, so it does not contradict open defects #428 or #440. Good judgment.
> - **Performance/scalability claims — none made.** No throughput or latency numbers anywhere. The only quantitative assertions are structural (128-doc blocks, the `size: 10000` cap) and both check out.
> 
> Two claims I'd single out as *better* than they needed to be: `update_document` really is a **top-level** overlay, not a recursive merge (index.rs:13158-13163) — the doc says "top-level fields", which is precise and diverges from ES `_update` semantics in a way a sloppier doc would have papered over; and §6 explicitly says exact rescoring does not make candidate *selection* exact, which is the exact thing the adjacent code comment ("`_score`s stay byte-identical to brute force", index.rs:13862) invites a reader to over-conclude.
> 
> **On the ARCHITECTURE.md / WHY_XERJ.md edits: the old text was genuinely stale, and the new text is right.** All three deletions were false at HEAD, not merely dated — "single `Mutex<WalWriter>`" (index_store.rs:661-693 opens N writers over one shared `seq_counter`), "16-shard memtable … & 15" (the hardcoded-15 helper was deleted as a misrouting bug, per the comment at index_store.rs:1611-1618), and "the engine-side FTS memtable is currently `Arc<RwLock<FtsMemtable>>`; a `ShardedFtsMemtable` scaffold exists … for a future refactor" (index.rs:5295 is `Arc<ShardedFtsMemtable>`, built at 6022 and 6220). The new `xerj_ingest_shards` paragraph is correct in a detail that is easy to get backwards: the global knob requires a power of two ≤256 (config.rs:1560-1573) while the per-index override accepts **any** integer in `1..=256` (index.rs:28502-28508). The conformance-number edit is also a real correction — the deleted "1,360 of 1,363" disagrees with AGENTS.md:5 at this same commit.
> 
> My findings are five items, none of which is a false statement about XERJ: one should-fix on anchor durability (the way this page will actually rot), and four nits.

**[should-fix] Line-pinned anchors drift from the symbols they name (several already off; one lands 585 lines inside its function)**

Every XERJ link is a hard `file.rs#Lstart-Lend` range against files that churn heavily (index.rs is >40k lines). At this commit most ranges already do not start on the symbol they are labelled with, and one is far enough off to mislead:

- `search_at_generation` is cited as `index.rs#L13844-L13930`. The function actually begins at index.rs:13259. The cited range is the kNN-dispatch block *inside* it (verified: `// ── KNN short-circuit ──` at 13844, `peel_knn_query` at 13878, `run_knn_hnsw` call at 13908). The claim it supports is TRUE and the range points at the right code, but a reader clicking it expecting a function definition lands mid-body with no signature in view.
- `Index::flush` cited L18002-L18125; actual `pub async fn flush` at index.rs:18012, body ends ~18147.
- `run_knn_hnsw` cited L10470; actual 10487. `save_hnsw_to_disk` cited L12274; actual 12309. `delete_document_versioned` cited L12929; actual 12947. `update_doc` cited es_compat.rs#L17362; actual 17446. `encode_columns` cited doc_values.rs#L620; actual 637. `retire_segment_files` cited index_store.rs#L1531; actual 1553.

None of these make a claim false today. But the doc's own closing line — "re-check the linked symbols when changing the engine" — is the entire maintenance strategy, and this is the specific way this page rots into the wrong artifact people quote. Suggest either symbol-name anchors without line numbers, or a CI check that each cited range still contains the named symbol.

```
docs/XERJ_VS_LUCENE.md §6 cites engine/crates/xerj-engine/src/index.rs#L13844-L13930 as `search_at_generation`; `async fn search_at_generation(` is at engine/crates/xerj-engine/src/index.rs:13259. Verified via `grep -n "fn search_at_generation" engine/crates/xerj-engine/src/index.rs`.
```

**[nit] §1 names `apply_merge` as the merge publisher, but the merge pass calls `apply_merge_with_repoints`**

The doc says the background `run_merge_once` writes a replacement segment "and `apply_merge` publishes it before `retire_segment_files` retires the inputs". The ordering and the retire behaviour are both correct. But the merge pass reached from `run_merge_once` calls `apply_merge_with_repoints` / `apply_merge_with_repoints_and_post_publish`, not `apply_merge`. `apply_merge` (index_store.rs:3683) is a thin wrapper used only by two other call sites (index.rs:3839, index.rs:4004).

Both delegate to `apply_merge_inner` (index_store.rs:3722), and the cited range L3680-L3822 happens to span all three wrappers, so nothing here is false — but the symbol a Lucene engineer is told to go read is not the one on the path they were told about.

```
engine/crates/xerj-engine/src/index.rs:9439 `self.store.apply_merge_with_repoints(&batch_slice, merged_meta.clone(), version_repoints)` and :9446 (the `#[cfg(not(test))]` arm); `retire_segment_files` follows at engine/crates/xerj-engine/src/index.rs:9537.
```

**[nit] Three code comments adjacent to the cited symbols are stale in ways the new doc gets right — worth a follow-up so source and doc stop disagreeing**

This is a credit to the PR, not a defect in it, but it leaves the repo in a state where the doc and the code comments contradict each other, and a future reader will not know which won:

1. `xerj-fts/src/index.rs:20-21` says "The FST output value is the byte offset in the `.post` file" and "`TermPostings` metadata (doc_freq, ttf) is stored in the `.meta` JSON." Both stale. The PR is correct: the FST value is a fixed 24-byte record slot in `.meta`, and `.meta` is binary ZFM4, not JSON.
2. `xerj-storage/src/segment.rs:102` documents `SectionType::Tombstones` as a "roaring bitmap of deleted doc ordinals". The actual payload is ZTB2 `(delete_seq_no, doc_id)` pairs. The PR's "sequence-aware tombstone section" matches the code, not the enum comment.
3. `xerj-engine/src/index.rs:8460-8462` (`run_merge_once` doc comment) says merged-away segment files "are left in place (cleanup happens on next restart via snapshot GC)". `retire_segment_files` sweeps opportunistically at merge time when no read lease is outstanding.

Suggest a follow-up PR fixing these three comments, so the doc's "source wins if a nearby overview page is older" rule does not quietly become "the doc wins over source comments".

```
engine/crates/xerj-fts/src/index.rs:1038 `let meta_offset = (ZFM3_HEADER_LEN + i * ZFM3_RECORD_LEN) as u64;` and :343-359 (ZFM4 binary layout) vs the module header at :20-21. engine/crates/xerj-storage/src/segment.rs:130-163 (`encode_tombstones_v2`) vs :102.
```

**[nit] The WHY_XERJ.md conformance rewrite is a correct edit made for a reason outside the PR's stated scope, and leaves a stale number in AGENTS.md**

The PR frames its edits as fixing "stale architecture prose that would otherwise contradict the new document." The conformance paragraph is genuinely stale — I confirmed that. But XERJ_VS_LUCENE.md never mentions conformance, so this edit contradicts nothing in the new page; it is a separate (good) correction riding along.

More substantively: the PR *removes* the number rather than updating it. AGENTS.md at this same commit still carries a hard number, and it is a different one (1,365/1,368) from the one being deleted (1,360/1,363). So the drift this edit was meant to fix is still live, just relocated to the file agents actually read first. Either update all three to the current number, or drop it from AGENTS.md too.

```
AGENTS.md:5 at this commit reads "(1,365/1,368 ES-YAML conformance)", while the ARCHITECTURE.md and WHY_XERJ.md text being deleted said "1,360 of 1,363" / "1,360/1,363".
```

**[nit] UNVERIFIABLE (Lucene side): "128 integers in the packed path" for Lucene 10.3.1 could not be confirmed, and the local checkout points the other way**

Outside my lens, but §2's punchline depends on it, so flagging rather than dropping. The doc asserts Lucene 10.3.1's postings format uses "packed integer blocks (128 integers in the packed path)", and then builds the section's conclusion on the shared 128 ("this shared block size is not Lucene codec compatibility").

The local Lucene checkout is main (shallow, single commit 9cda465), where `ForUtil.BLOCK_SIZE = 256` and the `lucene103` package contains only `blocktree/` — there is no `Lucene103PostingsFormat.java` to read. I could not obtain a 10.3.1 tag from the checkout or from the `refsym` index (which indexes the same main tree). So I can neither confirm nor refute 128 at 10.3.1.

Note the conclusion survives either way — 128 vs 256 does not make XERJ codec-compatible — so at worst the rhetorical setup is weaker than stated. But the contributor should confirm against the 10.3.1 javadoc they linked, since a wrong number here is exactly the kind of thing a Lucene reviewer will catch first. The companion claim "skip data interleaved at two levels" checks out on the successor format and is very likely correct for 10.3.

```
/home/claude/ref/lucene/lucene/core/src/java/org/apache/lucene/codecs/lucene104/ForUtil.java:34 `public static final int BLOCK_SIZE = 256;`; `ls .../codecs/lucene103/` returns only `blocktree`. Two-level skip confirmed at .../lucene104/Lucene104PostingsFormat.java:61-62 "Skipdata is interleaved with blocks on 2 levels."
```


---

## Where I land

**This is a good document and the verification says so more strongly than I expected.** Two independent lenses tried to break the technical claims and could not find a single false one. One of them put it this way: it is *more accurate than several code comments sitting next to the symbols it cites*. The Lucene side holds too — all 14 javadoc links resolve against the `10_3_1` tree, and nothing is attributed to Lucene that is really Elasticsearch or Solr.

The headline correction is right and worth having: `Index::flush` really does fan out one `do_flush_shard` per FTS shard, and the aggregate-drain path the old `ARCHITECTURE.md` described is labelled *"legacy path"* in-source at `index_store.rs:1875` with no non-test callers. The prose you deleted was genuinely stale.

**Two must-fix items stand between this and merge, and both are about consistency rather than correctness:**

1. **`AGENTS.md` still asserts the older, stronger version of the "no shared architecture" claim** that this PR softens in two other docs. Agents read `AGENTS.md` first, so leaving it is the drift this PR exists to remove, relocated.
2. **The new page is unreachable** from every index a reader or an agent actually starts at. A comparison nobody can find is a comparison nobody reads.

Two smaller things worth doing while you are in there: the line-pinned anchors (`file.rs#L13844-L13930`) already drift — `search_at_generation` is cited at 13844 but starts at 13259, so a reader lands 585 lines inside the function — and the conformance-number edit removes the number from two files while `AGENTS.md:5` still carries a *different* one (1,365/1,368 vs the deleted 1,360/1,363).

**One thing I could not confirm and you should:** §2 says Lucene 10.3.1's postings format packs *128 integers*. The local checkout is `main`, where `ForUtil.BLOCK_SIZE = 256` and there is no `Lucene103PostingsFormat.java` to read, so I can neither confirm nor refute it at 10.3.1. The section's conclusion survives either number, but a Lucene reviewer will check that first.

**Happy to push the two must-fix edits to your branch if you would rather not round-trip** — say the word. Otherwise it is yours and I will re-verify whatever you push. Thank you for this; it closes #373 properly.

buger:
Thanks for the detailed review. I rebased the branch onto current main and pushed the reviewed follow-up at e7935c1e2404bbd15873a1a686ff56c9ac8a1192.

The follow-up addresses the review findings:

- restores the repository-authoritative no-code and no-architecture statement;
- adds README and AGENTS discoverability, a human-oriented introduction, and a six-row side-by-side table;
- pins the comparison to the reviewed XERJ source snapshot and replaces movable source anchors with immutable links;
- adds the material Lucene strengths and caveats called out in review, including impacts and block-max pruning, BKD points, analyzer breadth, filtered HNSW with exact fallback, and quantized vector formats;
- preserves XERJ caveats around lexical default embeddings, approximate HNSW candidate selection, analyzer projection, and update limits;
- removes volatile hard-coded conformance totals;
- removes the misleading claim that arbitrary per-index ingest-shard values map cleanly to all WAL writers;
- corrects merge and kNN source anchors and explicitly scopes omitted comparison areas.

Validation on the final head covers 69 unique remote URLs, immutable XERJ source ranges, local links, Markdown structure, six headings and six table rows, and diff cleanliness. Cargo and ES-YAML were not run because the effective change is documentation-only. The PR body has been updated to the final head and current-main base.

xerj-org:
## Independent merge-decision verdict — **MERGE**, 0 blocking defects

Recorded verbatim before acting, as it should be. This panel was framed to check whether the **previously-raised findings were actually fixed** — not to open a fresh round. A contributor who fixes everything asked and then meets new objections learns not to contribute.

### Rationale (verbatim)

All four previously-raised findings are actually fixed, and I could not produce a single blocking defect after verifying every factual claim on both sides.

FINDING 1 (authoritative "shares no code and no architecture" claim contradicted). The real problem was that commit 1 had *weakened* the repository-authoritative wording: docs/WHY_XERJ.md and docs/ARCHITECTURE.md were downgraded to "no shared code or file formats", diverging from AGENTS.md. Commit e7935c1 restores the exact authoritative sentence in both files. AGENTS.md:5, docs/WHY_XERJ.md:5 and docs/ARCHITECTURE.md:11-13 now all read "shares no code and no architecture with Elasticsearch or Lucene". Fixed.

FINDING 2 (page unreachable). Four inbound links now exist: README.md:298 (docs list), AGENTS.md:173 (agent navigation table), docs/ARCHITECTURE.md:17, docs/WHY_XERJ.md:7. All relative links in all five touched files resolve on disk. Fixed.

FINDING 3 (material Lucene strengths missing). Every requested item is present and each was verified against the cited Lucene 10.3.1 javadoc, not asserted: competitive impacts / two-level skips (Lucene103PostingsFormat javadoc literally says "block size ... currently 128", "Skipdata is interleaved with blocks on 2 levels", "CompetitiveFreqDelta and CompetitiveNormDelta are used to safely skip score calculation for uncompetitive documents"); BKD points via Lucene90PointsFormat/PointRangeQuery/LongPoint; analysis-common breadth and CustomAnalyzer; no-arg StandardAnalyzer parity (javadoc: "Builds an analyzer with no stop words", "Filters StandardTokenizer with LowerCaseFilter and StopFilter", no stemming); FilteredHnswGraphSearcher plus automatic exact fallback (KnnFloatVectorQuery javadoc: "If the filter cost is less than k, just execute an exact search ... If the kNN search visits too many vectors without completing, stop and run an exact search"); Lucene99 scalar (javadoc: "must be 4 or 7") and Lucene102 binary quantized formats. Fixed.

FINDING 4 (volatile hardcoded conformance totals). The new page contains none. All three totals the PR's own diff carried are gone: AGENTS.md "1,365/1,368" and "1365 passed / 3 skipped" -> "0-failure ES-YAML CI gate"; docs/ARCHITECTURE.md "1,360 of 1,363" -> "zero failed cases"; docs/WHY_XERJ.md "1,360/1,363" -> "requires zero failures". Fixed. (README.md:237 and ROADMAP.md:18 still carry totals, but those lines are untouched by this diff, predate the PR, and ROADMAP already self-qualifies with "read the current number off CI, not off this file" — not a defect this PR introduces or was asked to fix.)

XERJ FACTUAL CLAIMS — verified against this repository at the pinned snapshot 24711999dd866ceec2a6e7c91d934c2c27d7066c, which exists and is an ancestor of main. All 41 unique permalink line ranges are in-bounds at that commit, all reference one single commit, and no relative source link with a line anchor survives in either doc. I opened each cited range and confirmed the named symbol/behavior: Index::flush really does spawn one do_flush_shard task per shard (index.rs:18012+); apply_merge_with_repoints publishes before retire_segment_files retires inputs; finalize_flush_with_publisher writes the .seg and the store holds snapshot as ArcSwap; ShardedFtsMemtable::with_registry_and_shards receives config.engine.ingest_shards at both create (6022) and reopen (6220); engine.ingest_shards is validated non-zero power-of-two <= 256 (config.rs:1543-1573); storage memtable is `config.num_wal_shards.max(1).next_power_of_two()` while wal_shards is the unrounded count, so "distinct domains, not a promised one-to-one pairing" is exactly right; Column enum is {Numeric, Keyword} with null_bitmap, numeric carrying a sorted (i64,u32) index; AnalyzerRegistry "standard" is StandardTokenizer + LowercaseFilter with the source comment "NO stop-words, NO stemming"; build_fts_field_configs maps Text->"standard", everything else->"keyword"; the match projection uses AnalyzerRegistry::default() and `get_analyzer(...)?` so unknown/custom names decline to the stored scan; match_phrase gates the FTS phrase path on `slop == 0 && text_fields.contains && analyzer in {None, "standard"}`; update_by_query literally builds `json!({"query": ..., "size": 10000, "from": 0})`; delete_document_versioned tombstones the HNSW node; run_knn_hnsw returns None for non-cosine, in-flight publication, stale graph, wrong pinned field, and scalar8.

HONEST IN BOTH DIRECTIONS. The page leads with XERJ's weakest facts rather than burying them: default embedding is lexical feature-hashing "not neural semantic understanding"; SkipEntry is unserialized scaffolding so advance_to scans linearly and "does not provide Lucene's persisted impact-skip behavior" (the source module doc says the same: "Skip-list acceleration: NOT implemented"); "XERJ does not claim a byte-compatible DocValues codec or a BKD counterpart" (accurate — SectionType::Points exists in the enum but `git grep SectionType::Points` at the snapshot returns zero uses, so there is no live BKD); "Exact rescoring does not make candidate selection exact: HNSW remains approximate"; the 10,000-hit _update_by_query cap. The Lucene-negative claim I checked hardest — that updateNumericDocValue/updateBinaryDocValue cannot add fields — is quoted almost verbatim from the javadoc ("You can only update fields that already exist in the index, not add new fields through this method"). I found no claim that flatters XERJ; where the page errs it errs against XERJ.

MECHANICS. All 26 Lucene URLs return HTTP 200 under curl -L. Exactly six numbered headings and exactly six table rows. Effective diff is exactly the five documentation files, zero non-markdown. All 15 check-runs on head e7935c1 are green. Everything remaining that a reviewer could raise here — that table row 3 says "Values and ranges" while the heading says "Doc-values equivalent", that the kNN section omits the aggs-forces-brute-force case, that README.md:237 still carries a total — is wording/completeness, which the brief explicitly places outside blocking.

```
mkdir -p /tmp/c-448 && git clone -q --no-hardlinks /home/claude/work/xerj-main /tmp/c-448/repo
git remote add gh https://…@github.com/xerj-org/xerj.git && git fetch -q gh refs/pull/448/head:pr && git checkout -q pr
gh pr view 448 --repo xerj-org/xerj --json title,body,state,headRefName,headRepositoryOwner,mergeable,files,commits
git cat-file -t 24711999dd866ceec2a6e7c91d934c2c27d7066c; git merge-base --is-ancestor 24711999dd866ceec2a6e7c91d934c2c27d7066c ghmain
git diff --stat 8d7fc76..pr; git diff --name-only 8d7fc76..pr | grep -vE '\.md$'
git diff 8d7fc76..pr -- AGENTS.md README.md docs/WHY_XERJ.md docs/ARCHITECTURE.md
git diff 8d7fc76..08966ad -- AGENTS.md docs/WHY_XERJ.md; git diff 08966ad..e7935c1 -- AGENTS.md docs/WHY_XERJ.md docs/ARCHITECTURE.md README.md
grep -oE 'https://github.com/xerj-org/xerj/blob/[0-9a-f]{40}/[^)#]+#L[0-9]+-L[0-9]+' docs/XERJ_VS_LUCENE.md docs/ARCHITECTURE.md | sort -u  (41 unique; all ranges bounds-checked against `git show 24711999:<path>`)
grep -nE '\]\(\.\.?/?[^)]*#L' docs/XERJ_VS_LUCENE.md docs/ARCHITECTURE.md  -> NONE
bash /tmp/c-448/check.sh  (30 cited symbol/range assertions against the pinned snapshot)
git show 24711999:engine/crates/xerj-fts/src/postings.rs | sed -n '1,40p'  (128-doc PFOR blocks, 'Skip-list acceleration: NOT implemented')
git show 24711999:engine/crates/xerj-storage/src/index_store.rs | sed -n '660,722p;1607,1645p'  (wal_shards vs num_shards.next_power_of_two)
git show 24711999:engine/crates/xerj-common/src/config.rs | sed -n '1512,1572p'  (ingest_shards power-of-two <= 256)
git show 24711999:engine/crates/xerj-engine/src/index.rs | sed -n '18012,18147p;37926,37978p;38447,38510p;13844,13934p;10487,10535p;19383,19410p;5918,5944p'
git show 24711999:engine/crates/xerj-api/src/es_compat.rs | sed -n '24254,24295p'  (size: 10000 cap)
git show 24711999:engine/crates/xerj-storage/src/doc_values.rs | sed -n '550,600p'; git show 24711999:engine/crates/xerj-storage/src/segment.rs | sed -n '85,113p'
git grep -n 'SectionType::Points' 24711999 -- engine  -> no uses (BKD section is dead scaffolding)
curl -s -o /dev/null -w '%{http_code}' -L <each of 26 lucene.apache.org URLs>  -> all 200
curl -s -L .../Lucene103PostingsFormat.html | sed 's/<[^>]*>//g' | grep -i '128|skip|impact'
curl -s -L .../StandardAnalyzer.html, .../Lucene99HnswScalarQuantizedVectorsFormat.html, .../FilteredHnswGraphSearcher.html, .../KnnFloatVectorQuery.html, .../IndexWriter.html
grep -nE '^## [0-9]\.' docs/XERJ_VS_LUCENE.md (6); grep -cE '^\| [0-9]\.' docs/XERJ_VS_LUCENE.md (6)
relative-link existence loop over all five touched files -> no broken links
grep -rnE '1,?3[0-9]{2}\s*(/|of)\s*1,?3[0-9]{2}' --include='*.md' .
gh api repos/xerj-org/xerj/commits/e7935c1.../check-runs -> 15/15 success
```
</details>


---

Thank you for turning both of these around so thoroughly — every finding was addressed, and the follow-up on #447 adopted the two composition clauses (append the next block; discard the failing status) that are the part which would actually have caught the round of #441 that slipped through. Merging.

PR #449, opened 2026-08-17
fix(es-compat): a multi-index alias answered as an alias over one member on every read path

### LINKED ISSUE #433: scroll through an alias spanning several indices reports the alias as _index, so (_index,_id) is still not distinct
Split from the independent verification of #424 (which fixes the comma-list / wildcard / `_all` cases).

## What still happens

`index_names` resolution in `es_compat.rs` expands `_all` and wildcard patterns, but **not alias names** — an alias stays a single literal name. So a scroll addressed through an alias spanning several concrete indices stamps the **alias** onto every hit, and `(_index, _id)` remains non-distinct across the members.

This is the same user-visible failure as #414: ids collide across the aliased indices, so a consumer keyed on `(_index, _id)` silently keeps a fraction of the corpus.

## Why it matters more than the comma-list case

#424's CHANGELOG named reindex, migration, backup and CDC as the blast radius. Those tools routinely scroll an **alias** rather than an explicit comma list — an alias is the normal indirection for exactly the rolling-index layouts they operate on. So the case left open is arguably the more common one in production.

## Scope

- `/a,b/_search?scroll=` — fixed by #424
- `/idx-*/_search?scroll=` and `_all` — fixed by #424 (resolver expands them, so hits carry concrete names)
- **`/my-alias/_search?scroll=`** — still broken, this issue

## Where to look

The per-hit index now travels with the hit (`ScrollContext::hits: Vec<(String, Hit)>` after #424), so the plumbing is already right — the gap is upstream, in what `index_names` contains when the spec is an alias. A fix likely resolves aliases to their concrete members before the hits are paired, so each hit carries the index it truly came from while the response keeps whatever `_index` semantics ES specifies for alias-addressed search.

**Not verified:** I have not measured the alias case end-to-end myself — this is carried over from the verification of #424, which reported it after testing against a three-index corpus. Reproducing it should be the first step.

xerj-org:
Closes #433.

## Reproduced first, and it is wider than the issue says

Three indices of 40 docs behind one alias; `GET /_alias/kb` lists all three.

```
                    alias   comma-list   wildcard
_count               40        120          120
_search  total       40        120          120
scroll   hits        40        120          120
```

**It is not a scroll bug** — `_count` and plain `_search` truncate identically, returning only `kb-a`'s documents.

**And the `(_index, _id)` framing describes the wrong failure.** On the alias scroll the pairs *are* distinct, 40 of 40, exactly because two-thirds of the corpus was never searched. It is not a keyed consumer silently keeping a fraction; it is every consumer getting a fraction, with a `_count` that agrees with the short result and no error anywhere.

## Cause: three copies of the same resolution, disagreeing

| resolver | aliases |
|---|---|
| `resolve_index_selector` | expanded |
| `search_impl`'s own (`es_compat.rs:8608`) | only `_all` and `*` |
| `count_docs`' branch gate | only `_all`, `*`, comma, star |

A plain alias fell through the last two — which is also why `_search` and `_count` could answer differently for the same selector. Both now expand aliases the way the first always did.

## After

```
_count  alias (3 members)            120
_count  alias (1 member)              40
_count  concrete index                40
_search alias total                  120
scroll  hits / distinct / mislabelled  120 / 120 / 0
```

Mislabelling drops to zero without touching scroll code: #424's per-hit plumbing was already correct and had no other indices to report.

## Fail-before

Reverting **only** the two source hunks, keeping the tests:

```
_count through the alias must cover all three members
  left: Some(40)  right: Some(120)
every member must be walked
  left: 40  right: 120
```

The single-member alias and concrete-index cases are asserted too — they are what a narrowing fix would break.

## Verification

`rustup run 1.97.1`: fmt clean, `clippy --workspace --all-targets -- -D warnings` clean, `xerj-api` green on both dev and ci-test.

**Not verified:** `_msearch`, aggregations and `_mget` — they reach a resolver too and may carry a fourth copy. I measured `_count`, `_search` and scroll only, and I am not claiming the rest.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8605,7 +8605,17 @@ async fn search_impl(
         .map(|n| resolve_date_math_index(n.trim()))
         .collect();
     let raw_names: Vec<&str> = resolved_names_owned.iter().map(|s| s.as_str()).collect();
-    let needs_resolve = raw_names.iter().any(|n| *n == "_all" || n.contains('*'));
+    // An ALIAS also needs resolving, and used not to be counted here — which is
+    // why a multi-index alias behaved as an alias over one member on every read
+    // path (#433). `_count` and plain `_search` truncated the same way: three
+    // indices of 40 documents behind one alias answered 40, from the first
+    // member, with no error and a `_count` that agreed with the short result.
+    // The `_index`-stamping problem reported alongside it is downstream of this
+    // — the hits could not carry the right index because the other indices were
+    // never searched.
+    let needs_resolve = raw_names
+        .iter()
+        .any(|n| *n == "_all" || n.contains('*') || state.engine.aliases.contains_key(*n));
     let index_names: Vec<String> = if needs_resolve {
         let all = state.engine.list_indices().await;
         let all_names: Vec<String> = all.into_iter().map(|i| i.name).collect();
@@ -8620,6 +8630,16 @@ async fn search_impl(
             } else if pattern.contains('*') {
                 for name in &all_names {
                     if glob_match_simple(pattern, name) && !resolved.contains(name) {
+                        resolved.push(name.clone());
+                    }
+                }
+            } else if let Some(targets) = state.engine.aliases.get(*pattern) {
+                // Expand to the concrete members, exactly as
+                // `resolve_index_selector` already does for every other
+                // endpoint. An alias naming one index resolves to that one, so
+                // the single-member case is unchanged.
+                for name in targets.value() {
+                    if !resolved.contains(name) {
                         resolved.push(name.clone());
                     }
                 }
@@ -17951,7 +17971,24 @@ pub async fn count_docs(
     body: OptionalJson<Value>,
 ) -> impl IntoResponse {
     // Multi-index / all selector: sum counts from every participating index.
-    if index == "_all" || index == "*" || index.contains(',') || index.contains('*') {
+    //
+    // An ALIAS belongs in this branch too. It used not to, so a multi-index
+    // alias fell through to the single-index path and answered one member's
+    // count — three indices of 40 documents behind one alias reported 40, with
+    // no error (#433). This is a THIRD copy of the same resolution, alongside
+    // `resolve_index_selector` and the one inside `search_impl`; they disagreed
+    // about aliases in different ways, which is why `_search` and `_count`
+    // could give different answers for the same selector.
+    let alias_selector = index
+        .split(',')
+        .map(str::trim)
+        .any(|n| state.engine.aliases.contains_key(n));
+    if index == "_all"
+        || index == "*"
+        || index.contains(',')
+        || index.contains('*')
+        || alias_selector
+    {
         let all = state.engine.list_indices().await;
         let all_names: Vec<String> = all.into_iter().map(|i| i.name).collect();
         let wanted: Vec<String> = if index == "_all" || index == "*" {
@@ -17963,6 +18000,12 @@ pub async fn count_docs(
                 if pat.contains('*') {
                     for n in &all_names {
                         if glob_match_simple(pat, n) && !out.contains(n) {
+                            out.push(n.clone());
+                        }
+                    }
+                } else if let Some(targets) = state.engine.aliases.get(pat) {
+                    for n in targets.value() {
+                        if !out.contains(n) {
                             out.push(n.clone());
                         }
                     }
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -18023,11 +18023,35 @@ pub async fn count_docs(
             }
             out
         };
-        let has_query = body
-            .as_ref()
-            .and_then(|b| b.get("query"))
-            .map(|q| !q.is_null())
-            .unwrap_or(false);
+        // Alias filters, collected from the selector AS WRITTEN. Expanding an
+        // alias to its members and summing their raw document counts answers
+        // the unfiltered corpus — for a filtered alias that is not a smaller
+        // wrong answer than before, it is a LARGER one, and it made `_count`
+        // disagree with `_search` in the opposite direction. A filtered alias
+        // is the usual poor-man's document boundary, so the count has to be of
+        // what the alias exposes.
+        let mut alias_filters: Vec<Value> = Vec::new();
+        for part in index.split(',').map(str::trim) {
+            for entry in state.engine.aliases.iter() {
+                if entry.key() != part {
+                    continue;
+                }
+                for backing in entry.value().iter() {
+                    if let Some(meta) = state.engine.index_alias_metadata.get(backing) {
+                        if let Some(filter) = meta.get(part).and_then(|v| v.get("filter")).cloned()
+                        {
+                            alias_filters.push(filter);
+                        }
+                    }
+                }
+            }
+        }
+        let has_query = !alias_filters.is_empty()
+            || body
+                .as_ref()
+                .and_then(|b| b.get("query"))
+                .map(|q| !q.is_null())
+                .unwrap_or(false);
         // A refused query is not a zero. This loop used to run the search under
         // `if let Ok(..)` with no else, so an index whose mapping refuses the
         // query — a `match` on an `"index": false` field, say — contributed
@@ -18060,6 +18084,14 @@ pub async fn count_docs(
                     .and_then(|b| b.get("query"))
                     .cloned()
                     .unwrap_or(json!({ "match_all": {} }));
+                if !alias_filters.is_empty() {
+                    query_val = json!({
+                        "bool": {
+                            "must": [query_val],
+                            "filter": alias_filters.clone(),
+                        }
+                    });
+                }
                 // Resolve `terms` lookups exactly as `_search` does — without
                 // this, `_count` silently returned 0 for a filter that
                 // `_search` counts correctly (the parser sees the raw lookup
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -9180,15 +9180,23 @@ async fn search_impl(
     // filter. The combined filter is also handed to run_aggs via the
     // background corpus (covered by the query's filter path already).
     {
+        // Built from the names as WRITTEN, not from `index_names`. Expanding
+        // an alias to its concrete members (#433) removes the alias name from
+        // that list, and this lookup matches on the alias name — so reading it
+        // from the expanded list silently dropped every filtered alias's
+        // filter. `global_with_aliases.yml` caught exactly that.
         let mut alias_filters: Vec<Value> = Vec::new();
-        for name in &index_names {
+        for name in &resolved_names_owned {
             for entry in state.engine.aliases.iter() {
                 if entry.key() != name {
                     continue;
                 }
                 for backing in entry.value().iter() {
                     if let Some(meta) = state.engine.index_alias_metadata.get(backing) {
-                        if let Some(filter) = meta.get(*name).and_then(|v| v.get("filter")).cloned()
+                        if let Some(filter) = meta
+                            .get(name.as_str())
+                            .and_then(|v| v.get("filter"))
+                            .cloned()
                         {
                             alias_filters.push(filter);
                         }
--- a/engine/crates/xerj-api/tests/alias_multi_index_reads.rs
+++ b/engine/crates/xerj-api/tests/alias_multi_index_reads.rs
@@ -44,7 +44,10 @@ async fn call(app: &axum::Router, method: &str, path: &str, body: Value) -> (Sta
     let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
         .await
         .unwrap();
-    (status, serde_json::from_slice(&bytes).unwrap_or(Value::Null))
+    (
+        status,
+        serde_json::from_slice(&bytes).unwrap_or(Value::Null),
+    )
 }
 
 const PER_INDEX: usize = 40;
@@ -163,7 +166,10 @@ async fn a_scroll_through_an_alias_walks_every_member_and_labels_each_hit() {
             hits.push((
                 h["_index"].as_str().unwrap_or_default().to_string(),
                 h["_id"].as_str().unwrap_or_default().to_string(),
-                h["_source"]["home"].as_str().unwrap_or_default().to_string(),
+                h["_source"]["home"]
+                    .as_str()
+                    .unwrap_or_default()
+                    .to_string(),
             ));
         }
         let Some(id) = sid.clone() else { break };
--- a/engine/crates/xerj-api/tests/alias_multi_index_reads.rs
+++ b/engine/crates/xerj-api/tests/alias_multi_index_reads.rs
@@ -1,0 +1,203 @@
+//! Issue #433: a multi-index alias answered as an alias over ONE member on
+//! every read path.
+//!
+//! Three indices behind one alias, 40 documents each. `GET /_alias/kb` listed
+//! all three, and `_count`, `_search` and a scroll through the alias all
+//! answered 40 — from the first member, with no error and a count that agreed
+//! with the short result. Not a keyed consumer silently keeping a fraction:
+//! every consumer got a fraction.
+//!
+//! The cause was three independent copies of index-selector resolution.
+//! `resolve_index_selector` expanded aliases; `search_impl`'s own resolver only
+//! looked for `_all` and `*`; `count_docs` gated its whole multi-index branch on
+//! the same two plus a comma or a star. A plain alias fell through both.
+//!
+//! These go through the real HTTP routes because that is the only place the
+//! three resolvers meet.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn call(app: &axum::Router, method: &str, path: &str, body: Value) -> (StatusCode, Value) {
+    let mut req = Request::builder().method(method).uri(path);
+    let body = if body.is_null() {
+        Body::empty()
+    } else {
+        req = req.header("content-type", "application/json");
+        Body::from(body.to_string())
+    };
+    let response = app.clone().oneshot(req.body(body).unwrap()).await.unwrap();
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .unwrap();
+    (status, serde_json::from_slice(&bytes).unwrap_or(Value::Null))
+}
+
+const PER_INDEX: usize = 40;
+
+async fn seed(app: &axum::Router) {
+    for member in ["kb-a", "kb-b", "kb-c"] {
+        let (st, _) = call(
+            app,
+            "PUT",
+            &format!("/{member}"),
+            json!({"mappings": {"properties": {"v": {"type": "long"}, "home": {"type": "keyword"}}}}),
+        )
+        .await;
+        assert_eq!(st, StatusCode::OK, "create {member}");
+
+        let mut bulk = String::new();
+        for i in 0..PER_INDEX {
+            bulk.push_str(&format!("{}\n", json!({"index": {"_id": i.to_string()}})));
+            bulk.push_str(&format!("{}\n", json!({"v": i, "home": member})));
+        }
+        let response = app
+            .clone()
+            .oneshot(
+                Request::builder()
+                    .method("POST")
+                    .uri(format!("/{member}/_bulk"))
+                    .header("content-type", "application/x-ndjson")
+                    .body(Body::from(bulk))
+                    .unwrap(),
+            )
+            .await
+            .unwrap();
+        assert_eq!(response.status(), StatusCode::OK, "bulk {member}");
+        let (st, _) = call(app, "POST", &format!("/{member}/_refresh"), Value::Null).await;
+        assert_eq!(st, StatusCode::OK);
+    }
+    let (st, _) = call(
+        app,
+        "POST",
+        "/_aliases",
+        json!({"actions": [
+            {"add": {"index": "kb-a", "alias": "kb"}},
+            {"add": {"index": "kb-b", "alias": "kb"}},
+            {"add": {"index": "kb-c", "alias": "kb"}},
+            {"add": {"index": "kb-a", "alias": "solo"}}
+        ]}),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "create aliases");
+}
+
+/// `_count` and `_search` must see every member, and a one-member alias must
+/// keep answering for exactly that member.
+#[tokio::test]
+async fn every_read_path_through_a_multi_index_alias_sees_every_member() {
+    let (app, _dir) = app().await;
+    seed(&app).await;
+    let all = (PER_INDEX * 3) as u64;
+
+    let (_, count) = call(&app, "GET", "/kb/_count", Value::Null).await;
+    assert_eq!(
+        count["count"].as_u64(),
+        Some(all),
+        "_count through the alias must cover all three members: {count}"
+    );
+
+    let (_, search) = call(
+        &app,
+        "POST",
+        "/kb/_search",
+        json!({"query": {"match_all": {}}, "size": 0}),
+    )
+    .await;
+    assert_eq!(
+        search["hits"]["total"]["value"].as_u64(),
+        Some(all),
+        "_search through the alias must cover all three members: {search}"
+    );
+
+    // The single-member case is the one a narrowing fix would break.
+    let (_, solo) = call(&app, "GET", "/solo/_count", Value::Null).await;
+    assert_eq!(
+        solo["count"].as_u64(),
+        Some(PER_INDEX as u64),
+        "an alias over one index still answers for that index: {solo}"
+    );
+    let (_, concrete) = call(&app, "GET", "/kb-a/_count", Value::Null).await;
+    assert_eq!(concrete["count"].as_u64(), Some(PER_INDEX as u64));
+}
+
+/// A scroll through the alias must walk every document AND report the index
+/// each hit truly came from — `(_index, _id)` is the key reindex, migration,
+/// backup and CDC consumers use, and ids collide across the members.
+#[tokio::test]
+async fn a_scroll_through_an_alias_walks_every_member_and_labels_each_hit() {
+    let (app, _dir) = app().await;
+    seed(&app).await;
+
+    let (st, mut page) = call(
+        &app,
+        "POST",
+        "/kb/_search?scroll=2m",
+        json!({"query": {"match_all": {}}, "size": 50}),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "scroll start: {page}");
+    let mut sid = page["_scroll_id"].as_str().map(str::to_string);
+    let mut hits: Vec<(String, String, String)> = Vec::new();
+
+    for _ in 0..16 {
+        let page_hits = page["hits"]["hits"].as_array().cloned().unwrap_or_default();
+        if page_hits.is_empty() {
+            break;
+        }
+        for h in &page_hits {
+            hits.push((
+                h["_index"].as_str().unwrap_or_default().to_string(),
+                h["_id"].as_str().unwrap_or_default().to_string(),
+                h["_source"]["home"].as_str().unwrap_or_default().to_string(),
+            ));
+        }
+        let Some(id) = sid.clone() else { break };
+        let (st, next) = call(
+            &app,
+            "POST",
+            "/_search/scroll",
+            json!({"scroll": "2m", "scroll_id": id}),
+        )
+        .await;
+        assert_eq!(st, StatusCode::OK, "continuation: {next}");
+        sid = next["_scroll_id"].as_str().map(str::to_string).or(sid);
+        page = next;
+    }
+
+    assert_eq!(hits.len(), PER_INDEX * 3, "every member must be walked");
+
+    let mislabelled: Vec<_> = hits
+        .iter()
+        .filter(|(reported, _, home)| reported != home)
+        .take(5)
+        .collect();
+    assert!(
+        mislabelled.is_empty(),
+        "hits must report the index they came from, not the alias: {mislabelled:?}"
+    );
+
+    let distinct: std::collections::HashSet<(&str, &str)> = hits
+        .iter()
+        .map(|(i, id, _)| (i.as_str(), id.as_str()))
+        .collect();
+    assert_eq!(
+        distinct.len(),
+        PER_INDEX * 3,
+        "(_index, _id) must be distinct — ids collide across the members"
+    );
+}


xerj-org:

Three lenses. Between them they found that this PR introduces regressions of its own, including one that leaks documents a filtered alias exists to hide. **Converting to draft.** Verbatim:

> CONFIRMED — there is a fourth copy, and a fifth through fifteenth. I found 15 independent index-selector resolution sites in engine/crates/xerj-api/src/es_compat.rs + engine/crates/xerj-engine/src/engine.rs; the PR fixed 2. Live probing (my own node on :9399 at PR head d518a19, base node on :9398 at 2471199) against a 3-member alias `tri` over idx-a/b/c shows the PR's caution is justified and understated: 11 endpoints still mis-answer a multi-index alias, in two classes. (A) `resolve_indices_for_op` (es_compat.rs:19415) has NO alias branch at all, so `_refresh`, `_forcemerge`, `_cache/clear`, `_close`, `_open`, `_terms_enum` and `_search/template` return HTTP 404 index_not_found on a valid alias. (B) `Engine::get_index` (engine.rs:3102) resolves an alias to `aliased.first()`, so `_delete_by_query`, `_update_by_query`, `_explain`, `_segments`, `_cat/segments`, `_eql/search`, `_async_search`, `_mget` and `_flush` silently answer from ONE member. `_delete_by_query {"match_all":{}}` through a 3-member alias deleted 10 of 30 documents and reported {"total":10,"deleted":10} with HTTP 200 — the #433 shape on a write path. Separately the PR REGRESSES `_count` on a FILTERED multi-member alias (ES-correct 30): base answered 40, head answers 121, while head's `_search` correctly answers 30 — so the commit message's claim about why `_search` and `_count` disagreed is not met and the gap is now 4x larger. Alias filters live in exactly one place (search_impl:9188); `_count`, `_msearch`, scroll, `_msearch/template` and `_cat/count` ignore them and return documents the alias filter excludes. The conformance test the PR credits, global_with_aliases.yml, uses `indices.put_alias` on a SINGLE index, so it structurally cannot catch any multi-member coupling. Yes, one resolver can serve all of them, and it is a net deletion — the PR's +40 lines of inline expansion in two handlers are exactly what a struct-returning resolver removes.

**[must-fix] Fourth copy: resolve_indices_for_op has no alias branch — 7 endpoints 404 on a valid alias**

engine/crates/xerj-api/src/es_compat.rs:19415 `resolve_indices_for_op` handles comma/_all/*/glob and hidden-index rules but has no alias case. A concrete part that is an alias name falls into the else branch at :19446 — `if !all.iter().any(|n| n == part) { return Err(index_not_found(part)) }` — and `all` is `list_indices()`, which contains index names only. Its callers are refresh_index:17882, close_index:26835, open_index:26856, forcemerge:26956, admin_segments_fsck:27034, clear_cache:27189, search_template:30044, terms_enum:30610. Every one hard-404s an alias. The fix is 5 lines (the same alias branch resolve_index_selector already has at :19383) and repairs 7 endpoints at once — smaller than what the PR did for two.

```
POST http://127.0.0.1:9399/tri/_refresh -> HTTP 404 {"type":"index_not_found_exception","reason":"no such index [tri]"}. Same 404 for /tri/_forcemerge, /tri/_cache/clear, /tri/_close, /tri/_terms_enum and POST /tri/_search/template. The alias is valid: GET /_resolve/index/tri -> {"aliases":[{"name":"tri","indices":["idx-a","idx-b","idx-c"]}]}, and GET /tri/_search?size=0 on the same node returns 121 hits. Base node :9398 gives identical 404s, so this is pre-existing and untouched — but it is squarely inside the 'is there a fourth copy' question, and the PR lists _refresh/_flush/_search-template as unverified.
```

**[must-fix] _delete_by_query and _update_by_query through a multi-index alias silently touch only the first member**

delete_by_query (es_compat.rs:24176) and update_by_query (es_compat.rs:24311) both begin with `state.engine.get_index(&index)` on the raw path segment. `Engine::get_index` (engine.rs:3102-3117) resolves an alias to `aliased.first()`. A write-by-query through a 3-member alias therefore operates on one index and reports success with a total/deleted equal to only that member's matches — indistinguishable from a correct run. This is the #433 failure mode on a destructive path, strictly worse than the read truncation the PR fixed, and the PR does not mention it.

```
Indices d-a/d-b/d-c, 10 docs each, alias `dtri` over all three. POST /dtri/_delete_by_query {"query":{"match_all":{}}} -> HTTP 200 {"total":10,"deleted":10}. Counts after: d-a=0 d-b=10 d-c=10 — 20 of 30 documents survived a match_all delete with no error and no partial-failure signal. Same on base :9398 (e-a/e-b, alias etri): {"total":1,"deleted":1}, after e-a=0 e-b=1. POST /tri/_update_by_query {"match_all":{}} -> {"total":41,"updated":41} while the alias holds 121 documents.
```

**[must-fix] The PR regresses _count on a FILTERED multi-member alias (40 -> 121; ES says 30) and leaves the _search/_count divergence it claims to explain**

count_docs' new alias branch (es_compat.rs:17987-18020) expands the alias to its members and sums raw doc counts but never consults `index_alias_metadata`. Alias-filter application exists in exactly one place in the crate — search_impl:9188-9232 (grep alias_filters: 9188, 9201, 9207, 9222, 9229-9232). After the PR, `_search` on a filtered alias is right and `_count` is wrong by the full unfiltered corpus. The comment the PR adds at :17984 says the copies 'disagreed about aliases in different ways, which is why _search and _count could give different answers for the same selector' — that disagreement is not closed for filtered aliases, it is widened. Either count_docs must reuse the alias-filter block (best: have the resolver return the filters), or the PR must state this as still broken.

```
Alias `mf` created with PUT /{idx-a,idx-b,idx-c}/_alias/mf {"filter":{"range":{"n":{"lte":10}}}}; 40 docs per index; ES-correct answer 30. HEAD (:9399): _search /mf = 30, _count /mf = 121. BASE (:9398): _search /mf = 10, _count /mf = 40. Filter is genuinely stored: GET /_alias/mf shows the range filter on all three members. Control that the filter field works: POST /idx-a/_search {"query":{"range":{"n":{"lte":10}}}} -> 10.
```

**[must-fix] Alias filters are ignored by _msearch, scroll, _msearch/template and _cat/count — out-of-filter documents are returned in hits**

msearch_impl (:22302), search_with_scroll (:19701) and msearch_template_impl (:30244) each resolve aliases via `Engine::resolve_alias` (engine.rs:2854) then search the members directly; none ever reads `index_alias_metadata`. A filtered alias is the standard poor-man's document-level access boundary, so this returns document bodies the alias is defined to hide. Pre-existing (base behaves the same), but it is precisely the 'more such couplings' the PR warns about, and it is why a single resolver must return the filters rather than only names.

```
Alias `mf` filters to n<=10. POST /_msearch, header {"index":"mf"}, body {"size":3,"sort":[{"n":"desc"}]} returns _source n=40,39,38 from idx-a, idx-b and idx-c — nine documents all outside the filter — while POST /mf/_search with the same body correctly tops out at n=10. Identical leak via POST /mf/_search_scroll?scroll=1m and POST /_msearch/template. GET /_cat/count/mf?format=json -> "121" (should be 30).
```

**[should-fix] Seven more read endpoints answer a multi-index alias from one member (get_index -> aliased.first())**

Every handler calling `state.engine.get_index(&index)` on the raw path segment inherits the first-member truncation at engine.rs:3110-3117: explain_doc:33030, index_segments:35509, index_recovery:35437, eql_search:32717, cat_segments:28986 (its own `_all`/`*`-only resolver then falls to get_index), async_search_submit:34669 (via get_or_create_index for non-selector specs), mget:18252 / mget_index:18300, flush_index:27101 (via Engine::flush_index engine.rs:3324 -> get_index). All answer HTTP 200 with a truncated or false result. `_explain` is the sharpest: matched:false for a document that exists behind the alias.

```
On :9399 with alias tri over idx-a(41)/idx-b(40)/idx-c(40)=121: POST /tri/_explain/idx-c-1 {"query":{"match_all":{}}} -> matched:false (idx-a-1 -> matched:true). GET /tri/_segments -> {"tri":{...num_docs:41}}. GET /_cat/segments/tri?format=json -> [{"index":"tri","docs.count":"41"}]. POST /tri/_eql/search -> hits.total 41. POST /tri/_async_search -> hits.total 41. POST /tri/_mget {"ids":["idx-a-1","idx-b-1","idx-c-1"]} -> found [true,false,false]. POST /tri/_flush -> 200 {"_shards":{"total":1,...}} (hardcoded 1; only idx-a flushed). GET /_cat/indices/tri -> HTTP 404. Correct by contrast on the same node: _search, _count, _msearch, _msearch/template, scroll, _field_caps, _stats, _mapping, _settings, _cat/count and aggregations all return 121 or all three member names.
```

**[should-fix] One resolver can serve all of them, and it is a net deletion rather than a bigger change**

The 15 sites differ on only five axes: name-set expansion (comma/_all/*/glob/alias/date-math), missing-name policy (lenient literal vs 404 vs get_or_create), hidden-index exclusion, closed-index policy, and whether the caller needs the names AS WRITTEN. That last axis is exactly what caused the PR's own regression at :9188 — the filter lookup keys on the alias name, which expansion had removed. A single `async fn resolve(state, spec, ResolveOpts) -> Result<ResolvedTargets, XerjError>` returning `{ concrete, as_written, alias_filters, routing }` makes that bug class unrepresentable: no caller can read the filter from the wrong list because no caller computes the filter. Concretely: (1) search_impl:8604-8654 and count_docs:17987-18020 both DELETE their inline loops and call it — the PR's +40 lines become roughly -60; (2) resolve_indices_for_op becomes `opts.missing = Strict, opts.hidden = Exclude` and its 7 callers stop 404ing aliases; (3) msearch, scroll, msearch/template, dbq, ubq, cat_segments, eql and async_search stop calling resolve_alias-or-get_index and get glob, _all, filters and multi-member correctness for free. The PR should either land that or state plainly which endpoints it does not fix.

```
Distinct resolvers found by grepping every selector split/glob/compare: resolve_index_selector es_compat.rs:19348 (alias expand-all, missing lenient); resolve_indices_for_op :19415 (alias NONE, missing 404, hidden excluded); search_impl inline :8604 (alias PR-added, plus date-math, closed-index rules and the only alias-filter block :9188); count_docs inline :17987 (alias PR-added, no date-math, no filters); Engine::get_index engine.rs:3102 (alias first() only); Engine::resolve_alias engine.rs:2854 (alias expand-all, no glob/_all); msearch_impl :22302 (no glob); search_with_scroll :19701 (no glob, no _all); msearch_template_impl :30244; open_pit :32200; resolve_existing_index_targets :27737; cat_segments :28986; field_caps :21454; cat_indices_inner :681/741 (no alias); async_search_submit :34669. Divergence is observable beyond aliases: POST /_msearch header {"index":"idx-*"} -> 404 index_not_found while POST /idx-*/_search -> 121; POST /idx-*/_search_scroll and POST /_all/_search_scroll -> 404.
```

**[should-fix] The conformance test the PR credits cannot catch multi-member alias couplings**

engine/tests/es-compat-yaml/yaml/aggregations/global_with_aliases.yml — the test the PR says caught its filter regression — creates the alias with `indices.put_alias` against a SINGLE index `test` and exercises `search` only. No multi-member alias, no _count, no _msearch, no scroll. Only 15 files in the whole yaml corpus mention aliases and none uses `update_aliases`, so nothing in the suite covers an alias over more than one index. The PR's new engine/crates/xerj-api/tests/alias_multi_index_reads.rs is therefore the only guard, and it does not cover the filtered multi-member case that regressed _count. Add a filtered 3-member alias case asserting _search == _count == 30.

```
engine/tests/es-compat-yaml/yaml/aggregations/global_with_aliases.yml lines 1-30: `indices.put_alias: index: test, name: test-filtered, body: {"filter": {"terms": {"name": ["one"]}}}` followed by a single `search: index: test-filtered`. `grep -rl update_aliases engine/tests/es-compat-yaml/yaml/` returns nothing; `grep -rln alias engine/tests/es-compat-yaml/yaml/ | wc -l` = 15.
```

**[nit] POST /_aliases with an add action silently drops the filter**

The `POST /_aliases {"actions":[{"add":{"index":...,"alias":...,"filter":...}}]}` form acknowledges the request but does not persist the filter; only `PUT /{index}/_alias/{alias}` does. Found while trying to reproduce the filtered-alias case the way most clients write it. Out of scope for this PR, but it means the filter path has only ever been exercised through one of its two entry points — part of why the coupling stayed invisible.

```
POST /_aliases with add+filter {"term":{"who":"idx-a"}} on idx-a/b/c as alias `ftri` -> {"acknowledged":true}; then GET /_alias/ftri -> {"idx-a":{"aliases":{"ftri":{}}},...} (empty), GET /_cat/aliases shows filter "-", and GET /ftri/_search?size=0 -> 121 (unfiltered). The same filter via PUT /idx-a/_alias/onea {"filter":{"range":{"n":{"lte":10}}}} -> GET /_alias/onea shows the filter and GET /onea/_search?size=0 -> 10.
```

**[nit] PUT /{alias}/_doc/{id} writes to the first member while _bulk correctly refuses**

The single-document index API resolves the alias through get_index and writes to `aliased.first()`, while _bulk enforces the ES rule and rejects a write to a multi-index alias with no write index. Two doors on the same server disagree about the same invariant. Same root cause (engine.rs:3110).

```
PUT /tri/_doc/z2 {"who":"z"} -> HTTP 201 {"_index":"tri","result":"created"} and the document lands in idx-a (idx-a count went 40 -> 41). POST /_bulk with {"index":{"_index":"tri","_id":"z1"}} -> status 400 illegal_argument_exception "no write index is defined for alias [tri]". Also GET /tri/_doc/idx-b-1 -> HTTP 404 found:false rather than ES's explicit multi-index-alias error.
```

> The filtered-alias break was not the only alias-name coupling. Measured PR HEAD (d518a19) against its merge-base (2471199) with in-process HTTP probes (axum router, no port), three indices of 40 docs behind aliases `kb` (plain) and `kbf` (filter v<10), plus a two-tenant filtered-alias fixture.
> 
> Two NEW regressions and one amplified one, all in the `_search`-vs-`_count` consistency the PR set out to establish:
> 
> 1. A CLOSED member behind an alias: `_search` silently drops it (80 hits, `_shards` reporting 2/2 successful) while `_count` counts it (120). Base agreed at 40/40. This is a fresh violation of the repo's own `count_agrees_with_search.rs` invariant ("`_count` must never report documents `_search` cannot return"). Cause: the closed-index pass keys on `raw_names`, so an expanded member is no longer a "concrete token" and is treated as wildcard-matched; `count_docs`' new alias branch has no closed handling at all.
> 
> 2. A PIT opened by a WILDCARD records the ALIAS NAME (open_pit globs over `state.engine.aliases` keys). `search_impl` overrides `index_names` from `pit.indices` AFTER resolution, so #433 is unfixed on that path — and commit d518a19 moved the alias-filter source to `resolved_names_owned`, which for a PIT search is the URL selector `*`. Net: `POST /kbf*/_pit` + `_search` returned 10 docs on base and 40 on HEAD — a filtered alias now leaks the documents its filter excludes.
> 
> 3. `_count` never applies alias filters: `kbf` answers 120 while `_search` answers 30. Base was 40 vs 10; the over-report grew from 30 to 90 documents. `count_docs`' new branch repeats exactly the mistake d518a19 fixed in `search_impl`.
> 
> Also: `term: {_index: …}` flipped (alias matched before / concrete matches now — ES matches both); two filtered aliases in one selector return ZERO hits because filters are ANDed globally instead of OR'd per index (pre-existing, but the PR's new comment asserts the ES semantics it does not implement); and `POST /_aliases` silently discards `filter` entirely, so the regression the PR guards is unreachable for aliases created the canonical ES way.
> 
> Genuine improvements confirmed: `_shards.total` 1→3, `indices_boost` with an alias key now applies, aggregation buckets cover all members, `kb,kb` no longer double-counts, and the PR's "hits carry the concrete member in `_index`" assertion is right. Full `cargo test -p xerj-api` is green on HEAD (216+ tests across 33 binaries, 0 failures).

**[must-fix] A closed member behind an alias: `_search` silently drops it, `_count` counts it — new violation of the repo's own count-agrees-with-search invariant**

`search_impl`'s closed-index pass (engine/crates/xerj-api/src/es_compat.rs:8704-8735) decides "named explicitly" by testing membership in `concrete_tokens`, which is built from `raw_names` — the names as WRITTEN. Expanding an alias puts the MEMBER name into `index_names` while `concrete_tokens` still holds only the alias name, so `concrete_tokens.iter().any(|t| t == name)` is false for every member and a closed member falls into the `// wildcard + default open → drop silently` arm. `count_docs`'s new alias branch (es_compat.rs:17993-18024) has no closed-index handling at all, so it counts the same index. The two endpoints now answer different questions for the same selector — the exact defect class this PR was written to eliminate, and a direct violation of engine/crates/xerj-api/tests/count_agrees_with_search.rs ("`_count` must never report documents that `_search` cannot return"). Worse, `_shards` reports `{"total":2,"successful":2,"failed":0}` — full success while a third of the alias is missing, so no client can detect the loss. Real ES throws `index_closed_exception` when an alias resolves to a closed index; this codebase already answers 400 `index_closed_exception` for the explicit concrete name (verified), so the silent drop is inconsistent with its own handling too. Fix: carry the alias→member provenance out of the resolver (a `Vec<(member, reached_via_alias_or_explicit)>`) instead of re-deriving "was this named explicitly" from `raw_names`, and give `count_docs` the same closed rule.

```
Alias `kb` over kb-a/kb-b/kb-c, 40 docs each; `POST /kb-b/_close` → 200.
PR HEAD (d518a19):  POST /kb/_search {"size":0} → hits.total=80, _shards={"total":2,"successful":2,"skipped":0,"failed":0}
                    GET  /kb/_count            → count=120,  _shards={"total":3,"successful":3,"skipped":0,"failed":0}
Base (2471199):     POST /kb/_search           → hits.total=40
                    GET  /kb/_count            → count=40      (they AGREED)
Contrast, both revisions: POST /kb-b/_search → 400 {"type":"index_closed_exception","index":"kb-b"}
Filtered alias + closed member on HEAD: /kbf/_search=20 vs /kbf/_count=120.
```

**[must-fix] A PIT opened by a wildcard records the ALIAS NAME — #433 is unfixed there, and this PR additionally drops the alias's filter, leaking excluded documents**

`open_pit` (es_compat.rs:32194-32211) resolves a CONCRETE selector with `resolve_alias` (→ members) but a WILDCARD selector by globbing over `state.engine.aliases` keys and storing the matched ALIAS NAMES alongside concrete index names. `search_impl` then replaces `index_names` wholesale with `pit.indices` at es_compat.rs:8666-8671 — AFTER the expansion the PR added — so an alias name inside a PIT is never expanded and `get_index(alias)` falls back to the first backing index. Bug #433 therefore survives untouched on the PIT path. On top of that, commit d518a19 moved the alias-filter lookup from `index_names` (which, for a PIT, held the alias name and so DID find the filter) to `resolved_names_owned` (which for a PIT search is the URL selector, `*` via `search_all`), so the filter is now silently dropped. Filtered aliases are the standard document-level-security mechanism here, so this is a read-authorization regression, not just a count discrepancy. Same result through `_msearch`. Note also the double-count this exposes: `POST /k*/_pit` records `kb-a,kb-b,kb-c` AND aliases `kb`,`kbf`, and the search totals 200 documents over a 120-document corpus (kb-a counted three times) — pre-existing, but it is the state the PR's new resolver now assumes cannot occur. Fix: expand aliases in `open_pit`'s wildcard arm (or expand `pit.indices` after the override), and derive alias filters from a list that survives the PIT override.

```
kbf = filtered alias (filter {"range":{"v":{"lt":10}}}) over kb-a/b/c, 40 docs each → 30 docs pass the filter.
PR HEAD:  POST /kbf*/_pit?keep_alive=5m ; POST /_search {"size":0,"pit":{"id":…}} → hits.total=40
Base:     same sequence                                                          → hits.total=10
Expected (ES): 30.  40 = one member, unfiltered → the filter is gone AND #433 persists.
_msearch with the same pit id on HEAD → 40.
POST /k*/_pit then _search → hits.total=200 over a 120-document corpus (both revisions).
```

**[must-fix] `_count` on a filtered alias ignores the filter — the new alias branch repeats the exact mistake d518a19 fixed in `_search`, and widens the gap from 30 to 90 documents**

`count_docs`'s new alias arm (es_compat.rs:18014-18019) expands the alias into `wanted` and then never consults `state.engine.index_alias_metadata` — the alias name is dropped from the list and the filter goes with it, which is verbatim the coupling that broke `_search` and that d518a19 had to repair. The PR's own comment on this branch says the three resolvers "disagreed about aliases in different ways, which is why `_search` and `_count` could give different answers for the same selector"; after the PR they still do, and by three times as many documents. This breaks engine/crates/xerj-api/tests/count_agrees_with_search.rs's stated invariant, and for a filtered alias used as a tenancy boundary `_count` is now a working oracle for the size of data the caller is not allowed to read.

```
kbf = filtered alias (v<10) over kb-a/b/c (40 docs each).
PR HEAD:  POST /kbf/_search {"size":0}                       → hits.total=30
          GET  /kbf/_count                                     → count=120
          POST /kbf/_count {"query":{"range":{"v":{"lt":100}}}} → count=120   (_search: 30)
Base:     _search=10, _count=40.
Over-report grew from 30 to 90 documents. Single-member filtered alias `solof`: _search=10, _count=40 on both.
```

**[should-fix] `term`/`match` on `_index` flipped: the alias name matched before and the concrete member matches now — ES matches both**

`strip_index_constraints` collects `_index` patterns and then filters `index_names` (es_compat.rs:8673-8695). That list used to hold the alias name and now holds the concrete members, so the set of values that match `_index` inverted. ES's `_index` matcher resolves the expression against cluster state, so both an alias name and a concrete index name match documents of the aliased index; this PR fixes one half and breaks the other. There is no yaml test in engine/tests/es-compat-yaml/yaml/ covering `term: {_index: <alias>}`, which is why nothing caught the flip. Fix: match an `_index` pattern against the member name OR any alias that resolves to it (the mapping is already available in `state.engine.aliases`), and add a yaml case.

```
Alias `kb` over kb-a/b/c, searched as /kb/_search:
  {"term":{"_index":"kb"}}    → HEAD 0    | base 40
  {"term":{"_index":"kb-b"}}  → HEAD 40   | base 0
  {"term":{"_index":"kb*"}}   → HEAD 120  (wildcards still work)
```

**[should-fix] Two filtered aliases in one selector return ZERO hits — alias filters are ANDed into one global clause, not OR'd per index**

The block the PR edited builds one `bool.filter` array from every alias in the selector and applies it to every participating index (es_compat.rs:9182-9225). Filters from DIFFERENT aliases are therefore intersected, and a filter from one alias is also imposed on an index the caller named directly. ES scopes a filtering alias to the indices that alias contributed and ORs multiple aliases' filters for the same index. The multi-tenant shape — one filtered alias per tenant, queried together — silently returns nothing. This is pre-existing (base behaves the same), but the PR is what added the comment asserting the ES semantics ("ES applies these on the coordinating node so each underlying index sees the filter"), and the switch to `resolved_names_owned` also removed the de-duplication `index_names` provided, so `kbf,kbf` now pushes the identical filter six times (2 names × 3 backing indices) into the query. Fix: build the filter per participating index from the aliases that actually reached it, OR the filters within an index, and omit it for indices named concretely.

```
t1-data / t2-data, 10 docs each; filtered aliases t1-view (term tenant=t1) and t2-view (term tenant=t2), created via PUT /{index}/_alias/{alias} so the filter is recorded:
  /t1-view/_search          → 10   (correct)
  /t2-view/_search          → 10   (correct)
  /t1-view,t2-view/_search  → 0    (HEAD and base; _count says 20; ES: 20)
  /t1-view,t2-data/_search  → 10   (HEAD; ES: 20 — t2-data was named directly, t1-view's filter should not apply to it)
```

**[should-fix] `POST /_aliases` silently discards `filter` — the protection this PR restored is unreachable for aliases created the canonical ES way**

`AliasActionParams` (es_compat.rs:19232-19235) declares only `index` and `alias`. `post_aliases` (es_compat.rs:19249-19318) calls `state.engine.add_alias` and never writes `state.engine.index_alias_metadata`, so `filter`, `routing` and `is_write_index` in an `add` action are parsed away and the request answers `{"acknowledged":true}`. Only `PUT /{index}/_alias/{alias}` and create-time `aliases` record a filter. `POST /_aliases` is the documented ES way to create a filtered alias (and the only atomic one), so for a large share of deployments the filtered-alias regression this PR fixed never existed and the filter never applied in the first place. It also means the PR's regression coverage is narrower than it appears: an added test that creates its alias through `_aliases` would pass whether or not the filter code works.

```
POST /_aliases {"actions":[{"add":{"index":"kb-a","alias":"kbf","filter":{"range":{"v":{"lt":10}}}}}, …]} → 200 {"acknowledged":true}
  then GET /kbf/_search {"size":0} → 120 (filter absent).
Same aliases via PUT /kb-a/_alias/kbf {"filter":…}:
  GET /_alias → {"kb-a":{"aliases":{"kbf":{"filter":{"range":{"v":{"lt":10}}}}}}}
  GET /kbf/_search {"size":0} → 30 (filter applied).
```

**[nit] There are at least five copies of index-selector resolution plus a sixth in open_pit, not the three the PR's comment names**

The PR comment in `count_docs` calls itself "a THIRD copy of the same resolution, alongside `resolve_index_selector` and the one inside `search_impl`". Two more exist and both diverge on aliases: `_msearch` at es_compat.rs:22300-22315 (uses `resolve_alias`, so it expands aliases correctly — but never applies alias filters), and `search_with_scroll` at es_compat.rs:19650-19653 (`/{index}/_search_scroll`, same). `open_pit` at es_compat.rs:32194-32211 is a sixth and the only one that can put an ALIAS NAME into an index list, which is precisely the state the PR's fix assumes cannot reach `index_names`. Consolidating these onto one resolver that returns (member, source-expression) pairs is the only thing that makes the class of bug go away rather than moving it.

```
POST /kbf/_search_scroll?scroll=2m {"size":0} → hits.total=120 (filter absent) on HEAD and base; /kbf/_search → 30.
_msearch header {"index":"kbf"} → responses[0].hits.total=120 on HEAD and base.
```

**[nit] An alias may be created with the same name as an existing index, and `count_docs` is now a third place that silently prefers the alias**

`Engine::add_alias` (engine/crates/xerj-engine/src/engine.rs:2828-2837) performs no name-collision check, and `Engine::get_index` (:3102-3120) checks the alias map first, so an index shadowed by a same-named alias becomes unaddressable on every read path. ES refuses the alias with `invalid_alias_name_exception` ("an alias with the same name as an index is not allowed"). Behaviour is unchanged by this PR, but `count_docs` now joins `get_index` and `search_impl` in resolving the shared name to the alias, so a fix has three call sites to keep consistent. Cheapest fix is to refuse the alias at creation.

```
PUT /shadow (7 docs indexed and refreshed), then POST /_aliases {"actions":[{"add":{"index":"kb-a","alias":"shadow"}}]} → 200 {"acknowledged":true}.
POST /shadow/_search {"size":0} → hits.total=40 (kb-a); GET /shadow/_count → 40. The real index's 7 documents are unreachable. Identical on base.
```

### Verdict 3 — refuted: `True`

> The core evidence HOLDS. On main (merge-base 2471199) three 40-doc indices behind alias `kb` answer `_count`=40, `_search` total=40, scroll=40 hits / 40 distinct / 40-of-40 mislabelled `_index`; comma-list and wildcard answer 120. On the branch all three are 120 and mislabelled drops to 0. Reverting ONLY af5ff14's two source hunks (keeping the d518a19 filter hunk and both tests) reproduces the fail-before verbatim: `left: Some(40) right: Some(120)` and `left: 40 right: 120`.
> 
> Four subsidiary evidentiary claims do NOT hold, and the change introduces a new silent regression.
> 
> (1) The stated conformance catcher is wrong. `yaml/aggregations/global_with_aliases.yml` has ZERO assertions (the file ends after the `search` do-step) and PASSES on a node built from the broken intermediate commit af5ff14. The files that actually catch the dropped alias filter are `aggregations/sig_terms.yml` ("Test alias background filter": bg_count expected 2, got 3) and `vectors/search.vectors/135_knn_query_nested_search_ivf.yml` (_id expected "3", got "2"). The wrong filename is now a permanent comment in es_compat.rs:9187.
> 
> (2) The single-member-alias and concrete-index assertions are not load-bearing. They pass on unfixed main and against every narrowing mutant I built, because `Engine::get_index` (engine.rs:3112) already collapses an alias to its first backing index.
> 
> (3) "The single-member case is unchanged" is false: `_index` stamping through a one-member alias changes from the alias name to the concrete name — measured `one` -> `only-1` — and nothing asserts it.
> 
> (4) "on every read path" overstates: `_mget` and `GET /{alias}/_doc/{id}` still answer from member #1 only, on both main and branch.
> 
> New regression: `GET /{alias}/_count` where a member is CLOSED went from 1 (main, agreeing with `_search`) to 2 (branch, disagreeing with `_search`=1) — a new instance of the exact `_search`/`_count` divergence this PR set out to close.
> 
> Suite results. ES-compat YAML, my own nodes on :19500 (branch) and :19300 (main), never :9200: branch 1366 passed / 0 failed / 3 skipped / 1369 (two runs, identical); main run 2 identical at 1366/0/3. Main run 1 read 1357/1/11 but that single FAIL (`search/160_exists_query.yml`, store_exception "Segment ... not found") did not reproduce on re-run and the extra skips are the runner counting setup errors as skips under load-average 850 — not attributable to main. The suites being identical means the corpus contains no test that fails on main for #433: it could never have caught this bug and does not guard the fix. `cargo test --workspace --profile ci-test` on 1.97.1: exit 0, 133 binaries, 3127 passed / 0 failed / 30 ignored. The dev profile does not build in this environment — third-party `gemm-f16 v0.19.0` fails 11x with "instruction requires: fullfp16" on aarch64, identically on main, so it is environmental; `cargo test -p xerj-api` on dev is green (429 passed / 0 failed / 2 ignored), so the PR's own narrower claim holds.
> 
> Mutants caught: expand-only-first-member; count_docs-only; search_impl-only; gate-flipped-without-expansion-arm. Mutants that SURVIVE: dropping the dedup, and expanding only aliases with more than one member.

**[must-fix] New regression: `_count` through an alias with a closed member counts the closed index, disagreeing with `_search`**

`search_impl` drops closed indices AFTER alias expansion (es_compat.rs:8704-8735, whose `concrete_tokens` are built from the names as written, so an expanded alias member is treated as wildcard-matched and silently dropped). `count_docs`' newly-added alias arm (es_compat.rs:18014-18019) has no closed-index gate at all, so the closed member's `stats().doc_count` is summed in. Before this PR a plain alias never reached that branch. This is a new instance of exactly the `_search`/`_count` divergence the PR's own commit message names as the reason the fix matters. The underlying `count_docs` gap is pre-existing (a comma-list already gave 2 while `_search` 400s), but this PR is what routes plain alias selectors into it.

```
Two indices cm-a, cm-b (1 doc each) behind alias `cm`; `POST /cm-b/_close` -> 200. Measured on identical trees differing only in es_compat.rs:
  main   : GET /cm/_search total=1   GET /cm/_count = 1   (agree)
  branch : GET /cm/_search total=1   GET /cm/_count = 2   (disagree, no error, _shards.failed=0)
Controls, identical on both: GET /cm-b/_search -> 400 index_closed_exception; GET /cm-a,cm-b/_search -> 400; GET /cm-b/_count -> 200 count=1.
```

**[must-fix] The commit message and a permanent source comment both credit a conformance file that has no assertions and passes on the broken build**

Commit d518a19 and the comment at es_compat.rs:9183-9187 state that `global_with_aliases.yml` caught the dropped alias filter. That file contains a bulk, a put_alias and a search `do` step and then ends — no `match`, no `length`, nothing. The runner only fails a bare `do` on a non-2xx/404/409 status, and a dropped filter returns 200. Built a node from af5ff14 (the pre-filter-fix commit) and pointed the runner at it: global_with_aliases.yml passes 1/1. The real catchers are sig_terms.yml and the knn IVF vectors file. Leaving the wrong filename in the source sends the next person to a file that cannot fail.

```
es-yaml-runner --url http://localhost:19400 (node built from af5ff14):
  yaml/aggregations/global_with_aliases.yml            1 passed · 0 failed
  yaml/aggregations/sig_terms.yml                      6 passed · 1 failed
     FAIL · Test alias background filter — match aggregations.sig_terms.bg_count: expected Number(2), got Number(3)
  yaml/vectors/search.vectors/135_knn_query_nested_search_ivf.yml   3 passed · 1 failed
     FAIL · nested kNN search pre-filtered on alias with filter on top level fields — match hits.hits.0._id: expected String("3"), got String("2")
  yaml/indices/indices.put_alias/10_basic.yml          6 passed · 0 failed
All four files pass on both main (:19300) and branch (:19500).
```

**[should-fix] The dedup in both new alias-expansion arms is load-bearing and untested — a mutant that removes it passes both tests**

Mutant M2 deleted `if !resolved.contains(name)` in search_impl and `if !out.contains(n)` in count_docs, leaving everything else intact. Both PR tests pass. The dedup matters only when two aliases (or an alias and a wildcard) share a backing index — a case the tests never build, since they only ever select `kb` alone. Without it a scroll returns duplicate `(_index,_id)` pairs, which is the failure mode #424/#433 exist to prevent. The concrete-name `else` arm has its own dedup, which is why `kb,kb-a` still looks correct and hides the gap.

```
Mutant M2 (dedup removed from both alias arms): `cargo test -p xerj-api --test alias_multi_index_reads --profile ci-test` -> ok. 2 passed; 0 failed.
Direct measurement, alias `kb`=[kb-a,kb-b,kb-c] and alias `solo`=[kb-a], selector `kb,solo`:
  branch (dedup) : count=120  search_total=120  scroll=120 hits/120 distinct
  M2 (no dedup)  : count=160  search_total=160  scroll=160 hits/120 distinct
```

**[should-fix] The single-member-alias and concrete-index assertions discriminate nothing; the narrowing mutant they claim to catch passes**

The PR says "The single-member alias and the concrete-index cases are asserted too — they are what a narrowing fix would break." Both assertions already pass on unfixed main. Mutant M7 is the natural narrowing fix — expand only aliases with more than one member, leave a one-member alias as the alias name — and both tests pass, because `Engine::get_index` (crates/xerj-engine/src/engine.rs:3112-3124) resolves an alias to `aliased.first()` on the fall-through path. What M7 actually destroys is invisible to the suite: the concrete `_index` stamp on a single-member alias silently reverts. Also note `/solo/_search` is never asserted at all — only `/solo/_count` — so search_impl's single-member behaviour has no assertion of any kind.

```
Mutant M7 (expand only when targets.len() > 1, in both needs_resolve/alias arm and count_docs gate/arm): `cargo test --test alias_multi_index_reads` -> ok. 2 passed; 0 failed.
Under M7 the single-member `_index` stamp reverts: PROBE2 single_member_alias _index=Some("one") — branch gives Some("only-1"), main gives Some("one"). No test observes this.
```

**[should-fix] "The single-member case is unchanged" is false — the `_index` stamp changes for every one-member alias**

The comment at es_compat.rs:8636-8639 asserts the single-member case is unchanged. It is not: before the PR a hit through a one-member alias reported `_index` as the alias; after, it reports the concrete index. The new behaviour is the ES-correct one, so this is a good change that is documented as a non-change and left unasserted — the worst combination, because a future "simplification" that skips single-member aliases (see M7) is invisible.

```
Alias `only-1` -> `one` (one member): POST /one/_search hits[0]._index = "one" on main, "only-1" on branch.
Alias `solo` -> [kb-a]: scroll hits/distinct/mislabelled = 40/40/40 on main, 40/40/0 on branch (mislabelled = reported _index != the seeded `home` field).
```

**[should-fix] `_count` on a filtered alias still disagrees with `_search`, and a filtered MULTI-index alias got further from ES**

`count_docs` applies no alias filter on either arm. Before the PR a filtered alias fell through to the single-index path and answered the whole backing index; now it takes the multi-index path and answers the SUM of every backing index, still unfiltered. So the PR's headline reason for the fix — "which is why `_search` and `_count` could answer differently for the same selector" — is only true for unfiltered aliases; for filtered ones the disagreement survives and, for multi-member filtered aliases, widens. Separately, multiple per-index alias filters are AND-ed (es_compat.rs:9220-9224 builds `bool.filter: [f1, f2]`) where ES applies each filter to its own index, so a two-member filtered alias returns 0.

```
Index `test`, 3 docs, alias `test-filtered` filter terms field2=[foo] (matches 2):
  main   : _search total=2, _count=3
  branch : _search total=2, _count=3     (unchanged, still disagreeing)
Alias `mix` over fa-1 (filter tag=red) and fa-2 (filter tag=green), 2 docs each, ES answer 2:
  main   : _search total=0, _count=2
  branch : _search total=0, _count=4
```

**[should-fix] The root cause — `Engine::get_index` collapsing an alias to its first member — is untouched, so `_mget` and doc-level GET still answer for one member**

engine.rs:3112-3124 resolves any alias to `aliased.first()` with no error. That is the mechanism behind #433, and the PR works around it at two call sites rather than at the source. 67 `state.engine.get_index(...)` call sites remain in es_compat.rs, several taking the raw path selector. Two read paths the PR lists as "not verified" are now verified as broken, unchanged by this PR: `_mget` and `GET /{alias}/_doc/{id}`. Both also stamp `_index` with the alias name, which the PR's own scroll test treats as a defect. This makes the title's "on every read path" inaccurate. Good news for the same list: `_msearch` and aggregations are fine — `_msearch` uses a FOURTH resolver, `Engine::resolve_alias`, which expands fully and was already correct on main.

```
Alias `kb`=[kb-a,kb-b,kb-c], ids kb-a-1 / kb-b-1 / kb-c-1. Identical on main and branch:
  POST /_mget {docs:[{_index:kb,_id:kb-a-1},{_index:kb,_id:kb-b-1},{_index:kb,_id:kb-c-1}]} -> [kb-a-1:true, kb-b-1:false, kb-c-1:false]
  GET /kb/_doc/kb-a-1 -> 200 found=true _index="kb";  GET /kb/_doc/kb-b-1 -> 404;  GET /kb/_doc/kb-c-1 -> 404
Already-correct paths — _msearch alias total 120 on both; aggs terms on `home`: main buckets [kb-a=40], branch [kb-a=40, kb-b=40, kb-c=40].
```

**[should-fix] The conformance corpus has no multi-index-alias read at all, so it neither found this bug nor guards the fix**

Both main and the branch score identically on the full YAML suite. The only alias reads in 200 files are single-index (`indices.put_alias/10_basic.yml`, `sig_terms.yml`, `global_with_aliases.yml`, the knn IVF file). The PR's own integration test is the sole guard for #433, and as shown above it does not pin the dedup or the single-member behaviour. A YAML case — three indices behind one alias, assert `_count` and `hits.total` — would cost a few lines and would have failed on main.

```
Full suite, my own nodes, never :9200:
  branch :19200 run 1  1366 passed · 0 failed · 3 skipped · 1369 total
  branch :19500 run 2  1366 passed · 0 failed · 3 skipped · 1369 total
  main   :19300 run 2  1366 passed · 0 failed · 3 skipped · 1369 total
  main   :19300 run 1  1357 passed · 1 failed · 11 skipped  (FAIL search/160_exists_query.yml store_exception "Segment ... not found"; did not reproduce — the same file re-run alone on the same node gave 73 passed · 0 failed)
```

**[nit] The fix adds two more hand-written copies of alias resolution instead of collapsing the three the PR diagnoses**

The commit correctly identifies "THREE independent copies of index-selector resolution" as the cause, then fixes two of them by pasting the alias arm into each, byte-for-byte the logic already in `resolve_index_selector` (es_compat.rs:19384-19390). There are now four resolvers plus `get_index`'s first-member shortcut. `count_docs` in particular has no reason not to call `resolve_index_selector` — it needs no date-math or closed-index bookkeeping that the shared function lacks. The two regressions above are both instances of the same thing: a behaviour `search_impl` has (closed-index gating, alias filters) that the pasted `count_docs` copy does not.

```
es_compat.rs alias-expansion arms: 8633-8643 (search_impl), 18014-18019 (count_docs), 19384-19390 (resolve_index_selector); msearch_impl uses Engine::resolve_alias (22308-22310); Engine::get_index takes `aliased.first()` (engine.rs:3113).
```

**[nit] The YAML runner reports a failed setup step as "skipped", so its pass/fail counts silently degrade under load**

run_file counts an errored setup step as `stats.skipped += 1` (src/main.rs:239) rather than as a failure. On a loaded box that turns real infrastructure errors into a quieter number: main run 1 read "1357 passed · 1 failed · 11 skipped" where the clean baseline is "1366 · 0 · 3". Any future claim of the form "the conformance suite is green" needs the skip count read alongside it, or a green run means less than it looks.

```
engine/tests/es-compat-yaml/src/main.rs:225-245 — setup error -> `stats.skipped += 1; setup_ok = false; break;`. Observed: main run 1 = 1357/1/11 at load average 852; main run 2 on the same node and binary = 1366/0/3 at load average 4.
```


---

## Draft, because two of these are regressions I introduced

**1. A closed member behind an alias.** `search_impl`'s closed-index pass decides "was this named explicitly" by testing `concrete_tokens`, built from the names **as written**. Expanding an alias puts the member into `index_names` while `concrete_tokens` still holds only the alias, so every member looks wildcard-matched and a closed one is dropped silently — 80 hits with `_shards` reporting `{"total":2,"successful":2,"failed":0}`, full success while a third of the alias is missing. `_count` has no closed handling at all and answers 120. Base agreed at 40/40. That breaks this repo's own invariant in `count_agrees_with_search.rs`: *`_count` must never report documents `_search` cannot return.*

**2. A PIT opened by a wildcard records the alias name**, and `search_impl` overrides `index_names` from `pit.indices` *after* my expansion — so #433 survives untouched there. Worse, moving the alias-filter lookup to `resolved_names_owned` means a PIT search finds no filter at all: base returned 10 documents, my head returns 40, ES would return 30. A filtered alias is the standard document-level boundary here, so that is a read-authorization regression, not a counting discrepancy.

**3.** The filtered-`_count` finding is already fixed in `fdbc609`, which post-dates this run — I will re-measure rather than assume.

The verdict also confirms real gains: `_shards.total` 1→3, `indices_boost` with an alias key now applies, aggregation buckets cover all members, `kb,kb` no longer double-counts, and hits carry the concrete member in `_index`. None of that is worth shipping a filter leak for.

Two adjacent defects it surfaced are filed separately, since they are not mine to fix here: #450 (`_delete_by_query` through an alias deletes from one member and reports success) and #451 (fifteen resolvers, seven endpoints 404 on a valid alias, four ignore alias filters).

**On the shape of the fix:** both regressions come from re-deriving "how was this index reached" after the fact. The resolver needs to return provenance — `(member, reached_via)` plus the filters — rather than a bare name list that later passes try to reverse-engineer. That is #451's point arriving from a second direction, and it is what I will do rather than patch the two symptoms.

xerj-org:

3 skeptics (correctness / regression-risk / evidence-holds), each refuting in its own worktree under 1.97.1. All three `refuted:false`, `blocking:false`. Verdicts verbatim:

> Traced the full code path in engine/crates/xerj-api/src/es_compat.rs. (a) needs_resolve now includes `state.engine.aliases.contains_key` (8616-8618); resolver branch 8636-8645 iterates targets.value() — the full Vec<String> of members from aliases: DashMap<String,Vec<String>> (engine.rs:440) — into index_names. The single exec loop `for idx_name in &index_names` (10400) calls idx.search() per concrete member and appends all hits: merged_hits.push((idx_name.to_string(), hit)) (10590). Not just the first. (b) _index is derived FROM the per-index loop var: EsHit { index: idx_name.clone() } (13284-13285) and fields-map _index (12081). So each hit carries its true member index; the stamping bug was genuinely downstream of the resolution bug (only the first member was ever in index_names, so only it was searched/stampable). (c) count_docs enters multi-index branch via alias_selector (18023-18031), expands alias to members in `wanted` (18014-18019, deduped), and loops `for ix_name in &wanted { total += result.total.value / doc_count }` (18077-18133) — a true sum; filtered aliases wrap query in bool filter (18087-18094). Single-member alias: targets.value() pushes exactly one member (search 8641-8645, count 18014-18019), so it resolves to exactly that one index — unchanged. All sub-claims confirmed by control flow, independent of the test. No regression on other read paths (comma/wildcard/_all branches untouched; alias branch is an added else-if).

**regression-risk — refuted:false, blocking:false.**

> FILTER PATH: loop at es_compat.rs:9188 iterates resolved_names_owned (alias name as written) and keys meta.get(name.as_str()) — filter survives even though index_names expanded away the alias; unfiltered alias yields empty alias_filters so nothing is applied. The only index-alias-filter site in search_impl is this loop; other aliases.get uses are field-alias rewrites (9974/9987) and indices_boost (10795, reads engine.aliases not index_names). DEDUP: all three resolvers guard with !contains, so a,idx1 (a→idx1) resolves to [idx1] once. COUNT_DOCS: filtered alias forces has_query and wraps bool{must:[q],filter:[alias_filters]} per member (18087); unfiltered alias sums idx.stats().doc_count (18131); body-query path unchanged. AGREEMENT: resolve_index_selector(19416)/search_impl(8636)/count_docs(18014) all expand via aliases.get with identical dedup, and all three prefer aliases.get/contains_key over a same-named index. New test alias_multi_index_reads passes (2/2) and global_with_aliases.yml covers the filtered-alias case. Non-blocking: mixed selector filtered_alias,idx3 applies the alias filter globally to idx3 too, but this is pre-existing global-filter behavior now made consistent between _search and _count, not introduced here.

> PASS-AFTER: `rustup run 1.97.1 cargo test -p xerj-api --test alias_multi_index_reads` on branch HEAD -> 2 passed. FAIL-BEFORE: a wholesale `git checkout origin/main -- es_compat.rs` does NOT compile (branch has newer API, e.g. SortField.unmapped_type absent in origin/main), so I instead reverse-applied ONLY the PR's es_compat.rs hunks (`git apply -R` of the origin/main...HEAD diff, 9 ins/92 del removed) — file still compiled, new test kept. Rerun: BOTH tests FAILED with the exact #433 symptom: every_read_path...=> _count returned Some(40) expected Some(120) ("_count through the alias must cover all three members", assertion at line 111); a_scroll...=> hits.len()==40 expected 120 ("every member must be walked", line 188). Restored fix afterward (git diff --stat clean). So it genuinely guards the fix. COVERAGE of the four asked items: (1) all-members hit counts — YES (_search total=120 line 124, _count=120 line 111, scroll walks 120 line 188); (2) per-hit _index stamping — YES (test 2 asserts reported _index==home per hit, mislabelled set empty, line 195, plus distinct (_index,_id) line 204); (3) _count/_search agreement — YES (both asserted ==120; plus single-member solo alias=40 line 132 guards against an over-narrowing fix, and concrete kb-a=40 line 138). (4) FILTERED alias — NO: seed() (lines 87-99) adds only plain aliases with no "filter" clause, so the fix's parts 2 and 3 (the alias-filter collection loops in search_impl and count_docs) are NEVER exercised by this .rs test. The diff comments say the filtered case is caught by YAML fixtures (global_with_aliases.yml), but that is out of scope of this test file. Missing filtered-alias coverage is a non-blocking gap (per task: missing coverage = non-blocking note); it does not undermine that the test is a real fail-before/pass-after guard for the primary multi-member read bug.

The evidence skeptic proved fail-before correctly (a wholesale `es_compat.rs` revert won't compile — the branch carries newer API — so it reverse-applied only the PR's hunks, kept the new test, and both tests failed with the exact #433 symptom: `_count` 40 vs 120). Two non-blocking notes filed as a follow-up: the new `.rs` test does not itself cover a *filtered* alias (that path is covered by `global_with_aliases.yml`), and a mixed `filtered_alias,idx3` selector applies the alias filter globally — pre-existing behavior, now made consistent between `_search` and `_count`, not introduced here.

PR #455, opened 2026-08-17
docs(readme): fix the "no releases found" badge and promote what XERJ is

xerj-org:
The release badge rendered **`release: no releases found`** on the front page of a repo with seventeen published releases.

## Cause — the opposite of what it looks like

`include_prereleases` is what broke it. Every release is tagged `-rc.N` but published with `prerelease=false`, so GitHub already reports `v1.0.0-rc.17` as the *latest release*. Asking shields.io to include prereleases makes it look for releases flagged as prereleases, find none, and say so.

Measured against the live endpoint:

| URL | renders |
|---|---|
| `…/github/v/release/xerj-org/xerj?include_prereleases&sort=semver` | **no releases found** |
| `…/github/v/release/xerj-org/xerj?include_prereleases` | **no releases found** |
| `…/github/v/release/xerj-org/xerj` | `v1.0.0-rc.17` |

The row also mixed two badge systems — GitHub's own workflow badge beside shields.io badges, which differ in height, font and corner radius — so the line never sat straight. That is the "looks weird" part. All shields.io `flat-square` now; the CI badge still points at the same workflow.

## What changed

Two rows instead of one: **project status** (CI · release · license · stars) and **what XERJ is** (ES 8.x conformance · single static binary · search modes · embedder · MCP), each linking to the page that backs it.

## Every badge render-tested, every claim checked against the tree

A badge is a claim on the front page, and this PR exists because one of them was wrong — so none of these went in unverified:

| badge | shows | backed by |
|---|---|---|
| CI | `passing` | live endpoint |
| release | `v1.0.0-rc.17` | live endpoint |
| license | `Apache-2.0` | `./LICENSE` |
| stars | `1.4k` | live endpoint |
| ES 8.x conformance | `1366/1369` | today's green main run: `1366 passed · 0 failed · 3 skipped · 1369 total` |
| single static binary | `no JVM` | static musl build, Rust workspace |
| search | `BM25 + semantic + kNN` | `es_compat.rs:7326` matches `"semantic" \| "knn"`; BM25 in `xerj-engine` |
| embedder | `built-in, offline` | lexical feature hashing is the default mode |
| MCP | `native` | `xerj mcp`, `main.rs:1687` |

All nine return HTTP 200 and render the text above.

## Known limitation

The conformance figure is **hardcoded and will drift** the moment the suite changes. It is correct as of this commit, and it is the one badge here that cannot verify itself. Worth a follow-up that has CI publish it (shields.io `endpoint` + a JSON file, or a gist), rather than a number a human has to remember to update.

## Not verified

How the two rows wrap on a narrow mobile viewport — I checked the URLs and the markdown, not the rendered page at phone width.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,10 +4,16 @@ XERJ is a search engine for AI agents. Point it at a folder and one command make
 your code, docs, logs and PDFs queryable, so an agent asks questions instead of
 reading files into its context window.
 
-[![CI](https://github.com/xerj-org/xerj/actions/workflows/ci.yml/badge.svg)](https://github.com/xerj-org/xerj/actions/workflows/ci.yml)
-[![License: Apache-2.0](https://img.shields.io/badge/License-Apache%202.0-blue.svg)](./LICENSE)
-[![Release](https://img.shields.io/github/v/release/xerj-org/xerj?include_prereleases&sort=semver)](https://github.com/xerj-org/xerj/releases)
-[![ES conformance](https://img.shields.io/badge/ES%20conformance-1366%2F1369-brightgreen.svg)](https://xerj.org/benchmarks)
+[![CI](https://img.shields.io/github/actions/workflow/status/xerj-org/xerj/ci.yml?branch=main&style=flat-square&label=CI)](https://github.com/xerj-org/xerj/actions/workflows/ci.yml)
+[![Release](https://img.shields.io/github/v/release/xerj-org/xerj?style=flat-square&label=release)](https://github.com/xerj-org/xerj/releases)
+[![License](https://img.shields.io/badge/license-Apache--2.0-blue?style=flat-square)](./LICENSE)
+[![Stars](https://img.shields.io/github/stars/xerj-org/xerj?style=flat-square)](https://github.com/xerj-org/xerj/stargazers)
+
+[![ES 8.x conformance](https://img.shields.io/badge/ES%208.x%20conformance-1366%2F1369-brightgreen?style=flat-square)](https://xerj.org/benchmarks)
+[![Single static binary](https://img.shields.io/badge/single%20static%20binary-no%20JVM-orange?style=flat-square)](https://xerj.org/docs/install.html)
+[![Search](https://img.shields.io/badge/search-BM25%20%2B%20semantic%20%2B%20kNN-8957e5?style=flat-square)](https://xerj.org/docs/queries.html)
+[![Embedder](https://img.shields.io/badge/embedder-built--in%2C%20offline-teal?style=flat-square)](https://xerj.org/docs/vectors.html)
+[![MCP](https://img.shields.io/badge/MCP-native-black?style=flat-square)](https://xerj.org/llms.txt)
 
 ## Paste this to your AI agent
 
--- a/README.md
+++ b/README.md
@@ -9,11 +9,12 @@ reading files into its context window.
 [![License](https://img.shields.io/badge/license-Apache--2.0-blue?style=flat-square)](./LICENSE)
 [![Stars](https://img.shields.io/github/stars/xerj-org/xerj?style=flat-square)](https://github.com/xerj-org/xerj/stargazers)
 
-[![ES 8.x conformance](https://img.shields.io/badge/ES%208.x%20conformance-1366%2F1369-brightgreen?style=flat-square)](https://xerj.org/benchmarks)
+[![ES-YAML conformance](https://img.shields.io/badge/ES--YAML%20conformance-1366%2F1369-brightgreen?style=flat-square)](https://xerj.org/benchmarks)
 [![Single static binary](https://img.shields.io/badge/single%20static%20binary-no%20JVM-orange?style=flat-square)](https://xerj.org/docs/install.html)
-[![Search](https://img.shields.io/badge/search-BM25%20%2B%20semantic%20%2B%20kNN-8957e5?style=flat-square)](https://xerj.org/docs/queries.html)
-[![Embedder](https://img.shields.io/badge/embedder-built--in%2C%20offline-teal?style=flat-square)](https://xerj.org/docs/vectors.html)
-[![MCP](https://img.shields.io/badge/MCP-native-black?style=flat-square)](https://xerj.org/llms.txt)
+[![Search](https://img.shields.io/badge/search-BM25%20%2B%20kNN%20%2B%20hybrid-8957e5?style=flat-square)](https://xerj.org/docs/queries.html)
+[![Default embedder](https://img.shields.io/badge/default%20embedder-lexical%2C%20offline-teal?style=flat-square)](https://xerj.org/docs/vectors.html)
+[![Neural](https://img.shields.io/badge/neural%20embedder-opt--in%2C%20downloads%20~90%20MB-9c6ade?style=flat-square)](https://xerj.org/docs/vectors.html)
+[![MCP](https://img.shields.io/badge/MCP-native-1f6feb?style=flat-square)](https://xerj.org/llms.txt)
 
 ## Paste this to your AI agent
 


xerj-org:
## Independent panel (head `ab9a343`) — 3 of 3 refuted

Three skeptics, each told to REFUTE, distinct lenses. All three refuted. Verbatim below.

### `455-claims` — refuted: **True**

The release-badge fix is real and I reproduced the author's diagnosis live: `include_prereleases` → "release: no releases found", parameterless → `v1.0.0-rc.17`. Seven of nine badges survived every attempt to refute them — CI passing, v1.0.0-rc.17, Apache-2.0, 1.4k stars (API: 1403), MCP native (`main.rs:1687`), BM25+semantic+kNN, single static binary — and I confirmed 1366/1369 by pulling the actual conformance job log from today's green main run (run 32021714760, head_sha 8d7fc76 = this PR's parent): "1366 passed · 0 failed · 3 skipped · 1369 total". The two-row layout renders as intended (verified against GitHub's own renderer, not the /markdown API, which uses gfm hard breaks and misleadingly suggests nine stacked badges). Two badges do not survive. "ES 8.x conformance" is a NEW claim this PR introduces (the old badge said just "ES conformance"), and the conformance runner hard-codes itself as an ES 8.13.0 cluster that by construction never attempts anything gated behind 8.14+; "ES 8.13 conformance" is what the tree supports. "Embedder: built-in, offline" is true only of the default lexical feature-hash path — neural embedding downloads all-MiniLM-L6-v2 from Hugging Face — and sitting next to "search: BM25 + semantic + kNN" it produces exactly the misreading the README body spends two separate paragraphs preventing. Separately, three of the five new capability badges point at pages that are stale or self-contradicting: install.html tells the reader "no installer ... write a 20-line config" directly above its own `curl | sh` section, and install/queries/vectors are all footer-stamped V1.0.0-RC.15 while the release badge two lines up says rc.17. The repo already owns the guard that catches this (landing-constants-guard.sh) and it is wired into no workflow. Nine badges do not bury the pitch — the one-sentence hook sits above them at README:3-5, which is the right call. Wrong-reason exclusions: every shields fetch was retried up to 3x at 25s; "no releases found" reproduced across three distinct query strings while the parameterless URL returned a version inside the same loop, so it is parameter semantics, not a 429/503; the one empty response (stars aria-label) turned out to be shields omitting that attribute on the endpoint, and the SVG rendered "1.4k" fine.

#### Findings

**[should-fix] "ES 8.x conformance" is a new, wider claim than the suite measures — the runner is pinned to ES 8.13**

The old badge read "ES conformance"; this PR widens it to "ES 8.x conformance". The conformance runner declares itself an ES 8.13.0 cluster and hard-codes RUNNER_MAJOR=8 / RUNNER_MINOR=13, so gte_v8.14+ predicates are never satisfied — every case gated behind 8.14 through 8.19 is excluded from the corpus by construction, not failed. The server advertises "8.13.0" on the wire to match. Elasticsearch 8.x ran to 8.19, so a green "8.x conformance 1366/1369" on the front page asserts near-total conformance across a line the suite deliberately does not reach. The denominator is also a curated subset: 200 upstream-derived YAML files weighted 78 aggregations / 67 search, against indices=6, cluster=1, ingest=1, ml=1, ccr=1, security=1 — not the full rest-api-spec corpus. ROADMAP.md:18-20 already carries the caveat the badge strips ("that count is the *dispatch* surface ... not a claim that every one is semantically faithful to ES"). Suggest "ES 8.13 conformance", which costs nothing and is unimpeachable.

```
engine/tests/es-compat-yaml/src/main.rs:669-688 — doc comment: "xerj reports ES 8.13.0 wire compatibility, so: ... gte_v8.0..=8.13 is satisfied / gte_v8.14+ is NOT satisfied", implemented as `const RUNNER_MAJOR: u32 = 8; const RUNNER_MINOR: u32 = 13;`. engine/crates/xerj-api/src/responses.rs:561 `number: "8.13.0".to_string()`. Corpus counts: `for d in engine/tests/es-compat-yaml/yaml/*/; do ls $d | wc -l; done`.
```

**[should-fix] "Embedder: built-in, offline" is the one claim the README body spends two paragraphs walking back**

The badge is true only of the default lexical feature-hash embedder. Real neural embedding is not built in and not offline — it downloads sentence-transformers/all-MiniLM-L6-v2 from Hugging Face, and the air-gapped recipe's own opening line concedes "The base path is lexical and needs no model files." The README hedges this explicitly in two places (line 92: "neural mode needs the three model files staged locally before the first semantic operation"; line 195: "default embedder is lexical feature-hashing, not neural"). Placed immediately after "search: BM25 + semantic + kNN", the pair reads to a first-time visitor as "neural semantic search, built in, offline, nothing to download" — the precise misreading the body text exists to prevent. Consider `embedder-lexical%20built--in%3B%20neural%20optional`.

```
docs/recipes/air-gapped-deployment.md:5 "The base path is lexical and needs no model files. Neural embedding is..."; :166 `https://huggingface.co/sentence-transformers/all-MiniLM-L6-v2/resolve/main/$FILE`; :272 "XERJ does not independently checksum the Hugging Face model files today." engine/crates/xerj-autoindex/src/catalog.rs:116 "the DEFAULT is the built-in LEXICAL feature-hash embedder (384-dim hybrid lexical+vector, NOT neural)". README.md:92, README.md:195.
```

**[should-fix] The "Single static binary" badge links to a page that contradicts the README's install instructions — and itself**

README:71 tells the reader to run `curl -fsSL https://xerj.org/get | sh`. The new badge sends them to install.html, whose opening paragraph says there is no installer and they must download a tarball, verify a checksum, hand-place the binary and write a 20-line config. That paragraph is also contradicted by its own page eleven lines later, which is headed "Install with one command" and shows the same `curl | sh`. The stale intro is now the first thing a visitor sees after clicking the front page's most prominent install-related badge.

```
landing/docs/install.html:122 "XERJ ships as a single static binary. No JVM, no installer, no package manager yet — download the tarball, verify the checksum, move the binary into place, write a 20-line config, start it as a systemd unit." vs landing/docs/install.html:133 `<h2>Install with one command</h2>` and :136 `$ curl -fsSL https://xerj.org/get | sh`. README.md:71.
```

**[should-fix] Every page the capability badges link to is version-stamped two releases behind the release badge**

The release badge now renders v1.0.0-rc.17 (verified live). The three landing pages the new capability badges point at are all footer-stamped V1.0.0-RC.15, dated 2026-08-11. The repo already contains the guard written for exactly this failure — landing-constants-guard.sh, added after a 2026-08-12 study found four different conformance figures live on the site at once — and running it with the correct VERSION fails with 83 stale stamps. It is wired into no workflow: grepping .github/ and scripts/ for its own name returns only the script file. Pre-existing site drift, but this PR is what routes front-page traffic to those pages, so it is now stale beside a badge.

```
`VERSION=1.0.0-rc.17 bash .github/scripts/landing-constants-guard.sh` → "FAIL version stamps disagree with the newest released tag (v1.0.0-rc.17)" listing landing/docs/install.html:224, landing/docs/queries.html:229 and :237, landing/docs/vectors.html:195 and :203, landing/index.html:331, "... and 63 more". `grep -rn landing-constants-guard .github/ scripts/` matches only .github/scripts/landing-constants-guard.sh itself.
```

**[should-fix] The hardcoded conformance figure is now in four places and the project's own drift guard cannot see any of them**

The commit message concedes the figure "cannot verify itself," but the project already solved this class of bug and the badge sits outside the solution twice over. landing-constants-guard.sh only walks $LANDING (default `landing`), so README.md is out of scope; and even if README.md were added, the badge URL-encodes the slash as `1366%2F1369`, which the guard's regex `1,?3[0-9]{2}[[:space:]]*(/|of)[[:space:]]*1,?3[0-9]{2}` cannot match. The number now lives in four copies, and ROADMAP.md:18 explicitly instructs readers to "read the current number off CI, not off this file" — advice the front-page badge cannot follow. Cheapest fix: emit the badge from the conformance job as a shields endpoint JSON, or at minimum widen the guard to README.md and match the percent-encoded form.

```
Copies: README.md:12 (badge, `1366%2F1369`), README.md:243 ("passes **1366 of 1369**"), ROADMAP.md:18 ("1,366 / 1,369"), landing/benchmarks/index.html:298. Guard scope: .github/scripts/landing-constants-guard.sh:38 `LANDING="${LANDING:-landing}"`; regex at :82 and :94. Rationale for the guard at :6-15.
```

**[nit] The MCP badge is literal black and disappears against GitHub's dark theme**

shields passes the CSS keyword straight through as `fill="black"` — pure #000000 — against GitHub's dark canvas #0d1117, roughly 1.2:1. The badge's right half reads as a hole in the page rather than a badge. Every other badge in the row uses a mid-tone that survives both themes. This is consistent with the render test having been done in light mode only. `informational`, `grey`, or a hex like #24292f reads on both.

```
`curl -s "https://img.shields.io/badge/MCP-native-black?style=flat-square" | grep -o '<rect[^>]*>'` → `<rect width="33" height="20" fill="#555"/>` and `<rect x="33" width="43" height="20" fill="black"/>`. Contrast with the teal badge, which resolves to `fill="teal"` (#008080, 4.77:1 against its white label — fine).
```

**[nit] The second badge row is within about 5% of wrapping and will orphan the MCP badge**

Measured from the live SVGs, row 2 is 192 + 164 + 194 + 156 + 76 = 782px plus four inter-badge spaces, about 800px. GitHub's repo page is capped at container-xl (1280px), leaving the README column roughly 838px with the About sidebar shown — so the row fits only at full width on an uncropped desktop window. Browser zoom, a narrower window, or one extra character in any label pushes MCP alone onto a second line. Row 1 is 394px, so the two rows also read unbalanced: one half-empty, one at the edge. Dropping or shortening one capability badge would buy real headroom.

```
Live widths: CI 74, release 128, license 120, stars 72; ES 8.x conformance 192, single static binary 164, search 194, embedder 156, MCP 76. Layout itself is correct — GitHub's own rendered README HTML (`gh api /repos/xerj-org/xerj/readme -H 'Accept: application/vnd.github.html+json'`) shows badges inside one `<p dir="auto">` with no `<br>`, confirming the blank line yields exactly two rows.
```

**[nit] The commit message's diagnosis is incomplete: sort=semver alone also breaks the badge**

The write-up pins the failure entirely on include_prereleases and lists three measured URLs. A fourth variant the message does not test — sort=semver with no include_prereleases — also returns "no releases found". The shipped fix removes both parameters and is correct, so the badge works; only the recorded cause is narrower than reality. Worth correcting since this commit message is the artifact the next person reads when a badge misbehaves.

```
Retried live, 3 attempts each at 25s: `?include_prereleases&sort=semver` → "release: no releases found"; `?include_prereleases` → "release: no releases found"; `?sort=semver` → "release: no releases found" (not in the commit message); no params → "release: v1.0.0-rc.17".
```

**[nit] The release badge is correct only while every RC is published with prerelease=false**

Dropping include_prereleases makes shields filter out anything flagged prerelease:true. It resolves today only because all seventeen releases are tagged -rc.N yet published as stable. The first time a maintainer does the conventional thing and ticks "This is a pre-release" on rc.18, the badge silently freezes at rc.17 — same class of failure, inverted, and with no error text this time to make it noticeable. A line in the release runbook ("RCs must be published with prerelease unchecked or the README badge stalls") costs nothing and closes the loop.

```
`gh api repos/xerj-org/xerj/releases/latest` → {"draft":false,"prerelease":false,"tag_name":"v1.0.0-rc.17"}.
```

**[nit] Badge alt text carries no values, and the license badge regressed**

GitHub renders the markdown alt attribute, not the SVG's internal aria-label — confirmed in GitHub's own rendered README HTML, where each badge becomes `CI`. A screen-reader user now hears nine bare labels — CI, Release, License, Stars, ES 8.x conformance, Single static binary, Search, Embedder, MCP — and not one value: no status, no version, no number. The old row's license badge was `![License: Apache-2.0]` and did carry its value; the new one is `![License]`. Folding the value into the alt (`![CI: passing]`, `![Release: v1.0.0-rc.17]`) is free for the static badges; for the live ones it reintroduces a drift surface, so at minimum restore the license one.

```
`gh api /repos/xerj-org/xerj/readme -H 'Accept: application/vnd.github.html+json'` on current main → alt="CI", alt="License: Apache-2.0", alt="Release", alt="ES conformance"; PR head renders alt="License". README.md:7-16.
```

**[nit] README:244 says the remaining conformance gaps "are listed there rather than hidden" — there is no such listing**

This is the prose a reader reaches after the conformance badge's number, and it is the sentence that makes the 3 unmet cases feel accountable. The linked directory contains only Cargo.toml, src/ and yaml/ — no README, no gaps file, no manifest. The three skips exist only as inline `skip:` directives scattered across 200 YAML files, several of which are upstream Elasticsearch known_issues rather than XERJ gaps. Either add the short listing the README promises, or soften the sentence.

```
`find engine/tests/es-compat-yaml -iname '*readme*' -o -iname '*gap*' -o -iname '*skip*'` → empty. `find engine/tests/es-compat-yaml -maxdepth 3 -type f | grep -v '\.yml$'` → Cargo.toml, src/main.rs only. README.md:243-245.
```

---

### `455-durability` — refuted: **True**

The badge that this PR fixes is genuinely fixed, and I could not break any individual claim on it. What I did refute is the *reason* recorded for the fix, and one composite claim the new row makes that the codebase's own comments exist to deny.\n\nDIAGNOSIS — half confirmed, half refuted. Confirmed: all 17 tags have a published release with `prerelease=false, draft=false`, and `/releases/latest` returns v1.0.0-rc.17, so GitHub really does consider the RC the latest release. Reproduced exactly: `?include_prereleases` and `?include_prereleases&sort=semver` -> `no releases found`, bare endpoint -> `v1.0.0-rc.17`. Refuted: the stated mechanism. `?sort=semver` ALONE — a parameter with no connection to prerelease flags, and the one variant the author never isolated — fails identically. And `?include_prereleases` returns `no releases found` for next.js, bun, deno, prometheus and grafana, five projects that unambiguously DO flag prereleases; under the stated mechanism at least the first two would return a canary. What the data actually separates is parameters that force enumeration of the releases *list* (both broken, every repo) from ones answerable off `/releases/latest` (`sort=date`, `display_name=tag`, bare — all fine). This matters because the commit message tells the next maintainer the parameter is safe to restore once a real prerelease exists. It is not, and that instruction is now permanent.\n\nDURABILITY. Dropping the parameter is correct today but silently depends on `release.yml:234` keeping `prerelease: false` under a tag glob that swallows RCs and finals alike. `/releases/latest` excludes prereleases by definition, so the badge works *because* every RC is mislabeled stable. The one-line change that breaks it is the semantically correct one, it is a routine request, and nothing — no comment, no CI check — connects the two files. Post-GA is fine (1.0.0 final shows v1.0.0; a genuine rc alongside it correctly shows v1.0.0); the exposure is exactly the pre-GA window the project is in now. `github/v/tag/xerj-org/xerj` returns v1.0.0-rc.17 under no sort, semver AND date — immune to the flag and to the broken code path.\n\nCLAIMS. Verified individually and left standing: CI passing, stars 1.4k, Apache-2.0 (LICENSE header), 1366/1369 (matches its link target verbatim), no-JVM (install.html:122 says it in those words), BM25+semantic+kNN (es_compat.rs:7326; queries.html documents both clauses), MCP native (main.rs:1687 exactly as cited). The problem is compositional: \"embedder: built-in, offline\" sitting beside \"search: ... semantic ...\" asserts something no configuration delivers. The default really is `auto` -> lexical, which local.rs:10, embedder.rs:9 and neural.rs:9 each go out of their way to call \"*not* semantic\" / \"the honest default\"; the real semantic backend needs a ~90 MB HuggingFace fetch on first use. Offline isn't semantic and semantic isn't offline — and README.md:92-93 and :195 already say so correctly, which makes the badge row the only place in the file that drops the qualifier. Separately, that badge's link target (docs/vectors.html) contains zero occurrences of \"embedder\" or \"offline\" and demonstrates only bring-your-own `query_vector` — the one link in the row that fails the PR's own \"links to the page that backs it\" standard.\n\nWRONG REASONS RULED OUT. `gh api /releases` returning `[]` looked alarming until the control: ripgrep, bat and cli/cli return `[]` too, authenticated and not — a sandbox artifact, so I worked from `/releases/latest` and `/releases/tags/{tag}` instead. github.com `/releases` and `/stargazers` 404 via curl, but so do ripgrep's, across 3 retries each with a browser UA — egress filtering, not broken badge links; I did not report either as a finding. Rate limit was 4976/5000, so no 429 was masking anything, and every shields conclusion was cross-checked against at least one control repo rather than drawn from a single fetch. All nine badges are `flat-square`, so the stated visual-consistency fix is real, and no badge link uses an anchor, so there is no anchor to break.

#### Findings

**[should-fix] The recorded diagnosis is wrong, and it is wrong in the direction that reintroduces the bug**

The fix works — I verified it end to end. The *explanation* attached to it does not survive testing, and it is now the permanent record in the commit message.

The commit says `include_prereleases` "makes it look for releases FLAGGED as prereleases, find none, and say so." Two independent tests refute that mechanism:

1. `?sort=semver` ALONE — no `include_prereleases` anywhere, and a parameter with no relationship whatsoever to prerelease flags — fails identically on xerj: `no releases found`. The author's three measurements never tested `sort=semver` on its own, so the second variable was never isolated. Both parameters in the original badge broke it independently.

2. `?include_prereleases` returns `no releases found` for vercel/next.js, oven-sh/bun, denoland/deno, prometheus/prometheus and grafana/grafana — five projects that unambiguously DO publish releases flagged `prerelease=true` (Next.js canaries, Bun canary, Deno RCs, Grafana betas). Under the stated mechanism at least Next.js and Bun would return a canary. None do.

What the data actually separates: parameters that force shields.io to enumerate the releases *list* (`include_prereleases`, `sort=semver`) fail for every repo tested; parameters answerable from `/releases/latest` (default, `sort=date`, `display_name=tag`) work for every repo tested. The prerelease flag is not the discriminator.

Why this matters beyond pedantry: the commit message tells the next maintainer that the parameter is safe to restore once a genuine prerelease exists. It is not. Whoever ships v1.1.0-rc.1 with `prerelease=true`, reads this commit, and re-adds `include_prereleases` will land the exact bug this PR fixed, having done the reasoning the commit told them to do.

Caveat I will not paper over: I cannot fully separate "real shields.io behaves this way" from "this sandbox degrades a shared shields.io code path." But the conclusion is unaffected — under either reading, the stated cause (our releases aren't flagged as prereleases) is refuted by `sort=semver` failing identically and by the five prerelease-publishing repos.

```
All 17 tags confirmed `prerelease=false, draft=false` via `/releases/tags/{tag}`, so that half of the diagnosis holds.

Parameter matrix (message field of the shields JSON endpoint):
  xerj-org/xerj      <none>=v1.0.0-rc.17  sort=date=v1.0.0-rc.17  display_name=tag=v1.0.0-rc.17  sort=semver=NO RELEASES FOUND  include_prereleases=NO RELEASES FOUND
  BurntSushi/ripgrep <none>=v15.2.0       sort=date=v15.2.0       display_name=tag=v15.2.0       sort=semver=NO RELEASES FOUND  include_prereleases=NO RELEASES FOUND

Repos that DO flag prereleases, plain vs include_prereleases:
  vercel/next.js        v16.3.1     -> no releases found
  oven-sh/bun           bun-v1.3.14 -> no releases found
  denoland/deno         v2.9.5      -> no releases found
  prometheus/prometheus v3.13.2     -> no releases found
  grafana/grafana       v13.1.3     -> no releases found
```

**[should-fix] The fix takes a silent, uncommented dependency on release.yml:234 staying `prerelease: false`**

This is the durability answer to "is dropping the parameter correct long-term?"

Correct today — verified, the plain endpoint returns v1.0.0-rc.17. But the bare badge resolves through GitHub's `/releases/latest`, which excludes prereleases *by definition*. It works only because `.github/workflows/release.yml:234` hardcodes `prerelease: false` for every tag matching `v[0-9]+.[0-9]+.[0-9]+*` — a glob that swallows `v1.0.0-rc.17` and `v1.0.0` alike. Today 100% of this project's releases are RCs published as stable. The front-page badge works *because of* that mislabeling.

The regression is one line and is the semantically *correct* change: flagging `-rc.N` tags as prereleases is what that field is for, and it is a routine request from users who don't want `gh release download` and `/releases/latest` handing them an RC. The moment someone makes it, `/releases/latest` has nothing to return and the front page reverts to `no releases found`. Nothing connects the two: no comment at release.yml:234, no comment in the README, no CI check, and the commit message frames `prerelease=false` as a fact of nature ("Every release is tagged -rc.N but published with prerelease=false") rather than as a line someone will plausibly edit.

To answer the rest of the lens directly: post-GA is fine. 1.0.0 final -> badge shows v1.0.0. A genuine v1.1.0-rc.1 flagged `prerelease=true` alongside v1.0.0 -> badge shows v1.0.0, which is the desirable behavior. The risk window is precisely the pre-GA period the project is in right now.

Durable remedy, verified working: `https://img.shields.io/github/v/tag/xerj-org/xerj` returns v1.0.0-rc.17 under no sort, `sort=semver` AND `sort=date`. It reads tags, so it is immune to the prerelease flag and to the list-enumeration path that broke the original badge. Failing that, a one-line comment at release.yml:234 saying the README release badge depends on this staying false.

```
/tmp/w455-brev/repo/.github/workflows/release.yml:234 -> `          prerelease: false` (inside softprops/action-gh-release@v2, under `if: startsWith(github.ref, 'refs/tags/v')`); tag glob at release.yml:7 is `'v[0-9]+.[0-9]+.[0-9]+*'`.

Tag badge control:
  github/v/tag/xerj-org/xerj <none>       -> v1.0.0-rc.17
  github/v/tag/xerj-org/xerj ?sort=semver -> v1.0.0-rc.17
  github/v/tag/xerj-org/xerj ?sort=date   -> v1.0.0-rc.17
(contrast github/v/release + sort=semver -> no releases found)
```

**[should-fix] "embedder: built-in, offline" beside "search: BM25 + semantic + kNN" asserts what the engine's own comments exist to deny**

Each badge is individually defensible; adjacent in a row captioned "what XERJ actually is" they compose into a claim no configuration delivers. Offline and semantic are properties of two different, mutually exclusive backends.

The default: `embedding.mode` defaults to `auto` -> "proxy when `default_endpoint` is set, else lexical" (index.rs:27340-27341). Out of the box, nothing configured, you get `Embedder::Lexical`. The tree describes that backend in three places, each going out of its way to deny it is semantic:
  local.rs:10    "This is **not** a neural model — it captures lexical / sub-word overlap, not deep semantics."
  embedder.rs:9  "but lexical, *not* neural semantic understanding. This is the honest default when nothing else is configured."
  neural.rs:9    "(the honest default; fast, deterministic, but *not* semantic)."

The real semantic backend is compiled into the shipped binary (`xerj-server/Cargo.toml:69 default = ["neural"]`) but is not the runtime default, and it is not offline: neural.rs:13 "Model files are fetched once on first use from the HuggingFace Hub," ~90 MB of all-MiniLM-L6-v2, with an error path that reads "If this host has no internet, pre-download the model and point...".

So: the offline embedder isn't semantic, and the semantic embedder isn't offline.

What makes this a defect rather than marketing latitude is that the README's own body already gets it right, 80 lines below the badge — README.md:92-93 "The default lexical embedder is offline; neural mode needs the three model files staged locally before the first semantic operation," and README.md:195 "default embedder is lexical feature-hashing, not neural, unless the node runs `--embed-mode neural`." The badge row is the only place in the file that drops the qualifier, and a badge is read by everyone while line 195 is read by almost no one. The commit message shows the author knew ("lexical feature hashing is the default mode") and shipped the unqualified word anyway.

Cheapest honest fixes: `embedder-built--in%2C%20offline%20(lexical)`, or `embedder-lexical%20offline%20%2B%20neural`, or drop "offline" and keep "built-in".

```
engine/crates/xerj-engine/src/index.rs:27340 `///   * "auto" (default / unknown) — proxy when default_endpoint is set,` / :27341 `///     else lexical.`
engine/crates/xerj-ai/src/local.rs:10 `//! This is **not** a neural model — it captures lexical / sub-word overlap, not`
engine/crates/xerj-ai/src/embedder.rs:9-10 `//!     fast — but lexical, *not* neural semantic understanding. This is the` / `//!     honest default when nothing else is configured.`
engine/crates/xerj-ai/src/neural.rs:9 `//!     (the honest default; fast, deterministic, but *not* semantic).`
engine/crates/xerj-ai/src/neural.rs:13 `//! Model files are fetched once on first use from the HuggingFace Hub and`
engine/crates/xerj-server/Cargo.toml:69 `default = ["neural"]`
README.md:92-93 and README.md:195 (the file's own, correct, qualified statements)
```

**[should-fix] The Embedder badge links to a page that never mentions an embedder**

`[Embedder](https://xerj.org/docs/vectors.html)`. The link resolves (200) and the file exists in tree, so a link checker would pass it — but the page does not back the claim and arguably contradicts it.

I extracted the rendered text of landing/docs/vectors.html. The word "offline" appears zero times. The word "embedder" appears zero times. "embedding" appears only as a *field name* in JSON examples. The page is about vector fields, quantization modes (scalar8/binary/scalar4), kNN and hybrid RRF — and every example on it requires the caller to supply a precomputed vector: `{"knn": {"field": "embedding", "query_vector": [0.12, 0.08, -0.31, ...], "k": 20}}`. A reader who clicks a badge that says "embedder: built-in" lands on a page whose every example demonstrates bringing your own vectors. That reads as the opposite of the badge.

This is the one badge in the row whose link target fails the PR's own stated standard ("each linking to the page that backs it"). The other four capability badges pass cleanly: install.html:122 literally says "single static binary. No JVM"; queries.html documents both `semantic` and `knn` as query clauses with `id="semantic"` and `id="knn"` anchors; benchmarks/index.html carries the exact conformance string; llms.txt exists.

The pages that actually document the embedder are landing/docs/recipes/semantic-search-rag.html, landing/docs/agents/quickstart.html and landing/docs/cli.html (`--embed-mode`).

```
grep -c -i offline landing/docs/vectors.html -> 0; grep -n -i embedder landing/docs/vectors.html -> no match. Rendered-text extraction of the page yields headings "KNN query" and "Hybrid — BM25 + KNN in one planner pass", quantization bullets, and query_vector examples only.
Contrast landing/docs/install.html:122 `<p>XERJ ships as a single static binary. No JVM, no installer...` — that badge's target does back it.
```

**[nit] Conformance 1366/1369 is now the front-page copy of a number stored in nine places, one of which is already stale, with no CI gate**

The author flags this honestly ("the one badge here that cannot verify itself"), and the number is correct as of this commit — it matches its own link target exactly. This is not a regression: the pre-PR badge carried the same figure. But the durability picture is worse than "hardcoded," and worth recording because the remedy is nearly free.

The same figure is duplicated across README.md:12 (badge), README.md:243, landing/benchmarks/index.html:298, :300 and :1115, landing/llms.txt:178, landing/llms-full.txt:24, and landing/index.html:310. A ninth in-repo copy is ALREADY WRONG and documents the failure mode in its own words — .github/workflows/ci.yml:440: "1365 passed / 3 skipped at the last measured run, per #189; this comment previously said 1326 and had drifted." So this number has demonstrably drifted at least twice, and it now sits on the front page.

Meanwhile the CI gate deliberately asserts nothing about totals — ci.yml:444 runs `es-yaml-runner` and relies solely on its non-zero exit for failures, with the comment explaining the gate is "0 failures — not an exact pass total, which grows as cases are added." That reasoning is right for the gate and wrong for the badge: the exact total is precisely what the badge publishes. There is also no link-check or markdown job anywhere in .github/workflows/, so no badge URL is verified by anything but review.

The suite already prints the numbers. Emitting a shields endpoint JSON from the conformance job, or asserting the totals there so a case addition fails loudly and forces the badge update, converts the project's most quotable claim from "correct until someone forgets" to "correct or red."

```
.github/workflows/ci.yml:440 `# grows as cases are added (1365 passed / 3 skipped at the last measured` — a fourth distinct value (1326, 1360/1363, 1365, 1366) for the same metric inside one repo.
landing/benchmarks/index.html:298 `ES-YAML conformance job on main's CI (1366 passed &middot; 0 failed &middot; 3 skipped &middot; 1369 total)` — matches the badge, so no contradiction today.
engine/tests/es-compat-yaml/yaml contains 200 YAML files; no assertion anywhere ties their case count to the published 1369.
grep -iE 'link.?check|lychee|markdown.?link' .github/workflows/*.yml -> no match.
```

**[nit] `label=release` is redundant on the release badge**

`https://img.shields.io/github/v/release/xerj-org/xerj?style=flat-square&label=release` — the shields `github/v/release` route already emits `"label":"release"` with no parameter at all, confirmed in the JSON response. Harmless, but it is one more query parameter on the badge whose extra query parameters were the bug. The CI badge's `label=CI` is NOT redundant (that route defaults to "build"), so this is only about the release one.

```
curl https://img.shields.io/github/v/release/xerj-org/xerj.json (no params) -> {"label":"release","message":"v1.0.0-rc.17",...}
```

---

### `455-reader` — refuted: **True**

All nine badges render (I fetched every shield URL myself and parsed <text> out of the SVGs — flat-square did emit <title> here as well, but I parsed <text> regardless). All four project-status badges are true: CI passing matches run #1058 on main (8d7fc76, the PR base, 15/15 jobs green); release v1.0.0-rc.17 matches the API (prerelease=false, draft=false) and I independently reproduced the diagnosis; license Apache-2.0 matches ./LICENSE and the workspace manifest; stars 1.4k matches 1403. The conformance figure 1366/1369 is real — I pulled the job log from the latest green main run and it reads "1366 passed · 0 failed · 3 skipped · 1369 total", and the suite runs unfiltered over the whole yaml/ tree (200 files, 13 subdirs, nothing parked elsewhere). "single static binary, no JVM" holds: I downloaded the release tarball and it is one executable plus README+LICENSE, statically linked, ldd says not a dynamic executable, zero JVM strings, and it boots. "MCP native" holds: `xerj mcp` in that same released binary answers initialize and tools/list as xerj-mcp 1.0.0-rc.17. BM25 and kNN hold, verified live against the released binary.

Two things did not survive. First, the two capability badges read together promise a configuration that does not exist: with the default embedder the semantic mode is not meaning-based, and the embedder that makes it meaning-based is not offline. Second, three of the five capability badges link to pages that do not back them — the conformance badge links to a page stating a different number, and the MCP and embedder badges link to pages that never mention MCP or the embedder.

Wrong reasons I ruled out: both Linux x86_64 artifacts failed with "Exec format error" — the host is aarch64, not a broken binary; the aarch64 musl artifact runs fine. My first server probes returned data from a different xerj already holding port 9200 (bind errors in my own log confirmed it); I moved to port 19455 and deleted the index I had accidentally created on that shared node. Writes were initially blocked by the disk flood-stage watermark because this box is at 99% disk, not by any XERJ defect; I disabled it via [limits] for the probes. github.com/xerj-org/xerj/stargazers 404s here, but so does rust-lang/rust/stargazers and so does the REST stargazers endpoint, so that is this environment's GitHub proxy, not a dead badge link. No shield fetch failed even once; every URL returned 200 on the first attempt, and the sort=semver result was re-run 3/3.

#### Findings

**[should-fix] The `search: semantic` and `embedder: built-in, offline` badges are true only in mutually exclusive configurations**

Neither badge is individually false, but placed side by side on the front page they promise "semantic search out of the box from a built-in offline embedder" and no single configuration delivers that.

Default config is mode = "auto" with default_endpoint = "", which resolves to the lexical feature-hash embedder (engine/crates/xerj-common/src/config.rs:1325). That embedder is genuinely built-in and offline, but it does not encode meaning — engine/crates/xerj-ai/src/embedder.rs:8 calls it "lexical, *not* neural semantic understanding" and "the honest default". I confirmed the behavioural consequence on the released rc.17 binary: for the query "kitten purring on the sofa" the default ranked the calculus doc first (MATH 0.620), the car doc second (CAR 0.606) and the cat doc LAST (CATS 0.513). For "cats make aloof domestic companions" the correct doc won by 0.008 (CATS 0.521 vs MATH 0.513), i.e. noise.

The embedder that makes "semantic" real is built-in but not offline. The tarball ships the binary, README and LICENSE only — no weights. The binary contains huggingface.co, sentence-transformers/all-MiniLM-L6-v2, model.safetensors and "downloading model weights"; engine/crates/xerj-ai/src/neural.rs resolve_from_hub() fetches config.json, tokenizer.json and model.safetensors (~90 MB) on first use, and engine/crates/xerj-server/Cargo.toml sets default = ["neural"] with the comment "the model downloads itself on first use".

The README body 80 lines below already says this correctly — line 195: "default embedder is lexical feature-hashing, not neural, unless the node runs `--embed-mode neural`", and lines 92-93 for air-gapped hosts. The badge row is less accurate than the prose it sits above, and it is the part an evaluator reads first. Someone who benchmarks XERJ's semantic search out of the box will measure the lexical embedder and conclude it is broken. If only one thing changes in this PR, change this — e.g. `embedder: built-in lexical, offline` plus a separate `neural: opt-in`, or fold the qualifier into the search badge.

```
Same binary (xerj-1.0.0-rc.17-aarch64-unknown-linux-musl), same 3-doc corpus, same queries; only --embed-mode differs:

DEFAULT (lexical, offline):
  'cats make aloof domestic companions' -> [('CATS',0.52125),('MATH',0.51294),('CAR',0.5)]
  'kitten purring on the sofa'          -> [('MATH',0.61988),('CAR',0.60621),('CATS',0.5133)]
  'vehicle motor power output'          -> [('CAR',0.56161),('CATS',0.50691),('MATH',0.49459)]

--embed-mode neural (downloads MiniLM):
  'cats make aloof domestic companions' -> [('CATS',0.811),('CAR',0.5481),('MATH',0.4223)]
  'kitten purring on the sofa'          -> [('CATS',0.689),('CAR',0.5432),('MATH',0.4974)]
  'vehicle motor power output'          -> [('CAR',0.798),('MATH',0.5112),('CATS',0.5074)]

server log, neural mode:
  INFO xerj_engine::index: embedding backend: built-in neural BERT (loads on first use) model=sentence-transformers/all-MiniLM-L6-v2
  INFO hf_hub: Using token file found "/home/claude/.cache/huggingface/token"

tar tzf xerj-1.0.0-rc.17-aarch64-unknown-linux-musl.tar.gz
  .../xerj
  .../README.md
  .../LICENSE          <- no model weights shipped

strings -a xerj | grep -oiE 'all-MiniLM-L6-v2|huggingface.co|downloading model weights'  -> all three present

engine/crates/xerj-common/src/config.rs:1325   mode: "auto".to_string()  (+ default_endpoint: String::new())
engine/crates/xerj-ai/src/embedder.rs:8        "lexical, *not* neural semantic understanding. This is the honest default"
engine/crates/xerj-server/Cargo.toml           default = ["neural"]
README.md:195                                  "default embedder is lexical feature-hashing, not neural"
```

**[should-fix] ES conformance badge links to a page that states a different number (1,360/1,363)**

The badge asserts 1366/1369 and links to https://xerj.org/benchmarks as its evidence. That page says 1,360/1,363 in two separate places. The badge number is the correct one — I pulled it from the CI job log of the latest green main run — and https://xerj.org/llms.txt independently agrees ("1,366/1,369 ES-YAML conformance"). It is the linked page that is stale. But the PR's stated design is that each badge links "to the page that backs it", and a reader who clicks through to check finds a contradiction rather than a confirmation. Either refresh the benchmarks page or point the badge at llms.txt or engine/tests/es-compat-yaml, which do back it. This drift predates the PR (the old badge linked to the same page), but the PR re-asserts the claim under an explicit link-backs-claim rule, so it is in scope.

```
Badge SVG <text>: 'ES 8.x conformance | 1366/1369'

CI run #1058 (main, sha 8d7fc76, conclusion success), job 'ES-compat YAML conformance', log line 631:

https://xerj.org/benchmarks (HTTP 200), text extracted from the page:
  "ES-YAML wire conformance 1,360 / 1,363 99.8%"
  "Every performance change must hold ES-YAML REST conformance at 1,360 passed / 0 failed / 3 skipped (1,360 of 1,363)."

https://xerj.org/llms.txt agrees with the badge:
  "...(1,366/1,369 ES-YAML conformance) as an adoption bridge..."

Suite is not filtered: .github/workflows/ci.yml runs `cargo run --release -p es-yaml-runner -- --dir tests/es-compat-yaml/yaml`; 200 yaml files across 13 subdirs, nothing parked outside yaml/.
```

**[should-fix] MCP badge links to llms.txt, which never mentions MCP**

The claim itself is true and I verified it against the running released binary — `xerj mcp` is a subcommand of the one shipped binary (engine/crates/xerj-server/src/main.rs:1687) and it speaks the protocol. But the badge links to https://xerj.org/llms.txt, and that document contains zero occurrences of "MCP" or "model context" across all 33,461 bytes. I fetched it twice with retries; both returned HTTP 200 with identical size, so this is not a partial fetch or a rate-limit stub. A reader clicking the badge to verify "MCP: native" lands on a page that does not discuss MCP at all. Point it at the MCP docs or the recipes page instead.

```
curl -sSL https://xerj.org/llms.txt -> http=200 bytes=33461
grep -c -i 'mcp' llms2.txt            -> 0
grep -i -n 'model context|mcp' llms2.txt -> (no output)

The claim is nonetheless true — released binary, aarch64 musl rc.17:
  $ printf '{...initialize...}\n{...tools/list...}' | xerj mcp
  {"jsonrpc":"2.0","id":1,"result":{"protocolVersion":"2024-11-05",
   "serverInfo":{"name":"xerj-mcp","version":"1.0.0-rc.17"}, ...}}
  stderr: xerj-mcp v1.0.0-rc.17 — MCP stdio server, proxying to http://localhost:9200

engine/crates/xerj-server/src/main.rs:1687:
  if matches!(std::env::args().nth(1).as_deref(), Some("mcp")) {
```

**[nit] Embedder badge links to docs/vectors.html, which does not describe the embedder**

https://xerj.org/docs/vectors.html is about vector storage, HNSW graphs, num_candidates and exact rescoring. Stripped of markup it is ~4.6 KB of prose containing zero occurrences of "embedder", "lexical", "neural" or "offline" (one incidental "built-in"). It backs the kNN badge, not the embedder badge. The page that actually documents embedding.mode (the Config TOML reference) or the semantic-search recipe would back it.

```
https://xerj.org/docs/vectors.html (HTTP 200, 28,136 bytes raw, 4,622 chars of extracted text)
  'offline': 0   'embedder': 0   'lexical': 0   'neural': 0   'built-in': 1

Page opens: "XERJ stores your vectors in full at ingest time and maintains a persisted HNSW graph alongside them..."

For contrast, the capability badges whose links DO back them:
  docs/install.html -> "XERJ ships as a single static binary. No JVM..." and "On Linux the one-liner installs the static musl build by default" (musl x8, JVM x1)
  docs/queries.html -> BM25 x1, semantic x2, kNN x4
```

**[nit] Commit message names one of two causes for the broken release badge**

The commit message states "`include_prereleases` is what broke it" and its measured table tests include_prereleases alone and include_prereleases+sort=semver, but never sort=semver alone. I did, three times: `?sort=semver` on its own also renders "no releases found", for the same underlying reason (semver sorting treats -rc.17 as a prerelease and excludes it). Both query parameters had to go, and the PR removes both, so the fix is correct — only the written diagnosis is incomplete. Worth correcting because the commit message is the artifact the next person reads when a badge misbehaves. No action needed on the README itself. I also checked the durability of the fix: it rests on releases being published with prerelease=false, and .github/workflows/release.yml:234 hardcodes exactly that, so that part is safe.

```
3/3 reproducible, parsing <text> from the returned SVG:
  run1: sort=semver -> 'no releases found'   |   no-params -> 'v1.0.0-rc.17'
  run2: sort=semver -> 'no releases found'   |   no-params -> 'v1.0.0-rc.17'
  run3: sort=semver -> 'no releases found'   |   no-params -> 'v1.0.0-rc.17'

Full sweep:
  ?include_prereleases&sort=semver  ->  release | no releases found
  ?include_prereleases              ->  release | no releases found
  ?sort=semver                      ->  release | no releases found   <- absent from the commit's table
  ?                                 ->  release | v1.0.0-rc.17

GH API: {"draft":false,"prerelease":false,"published":"2026-08-15T10:15:15Z","tag":"v1.0.0-rc.17"}
.github/workflows/release.yml:234: prerelease: false
```

---

## My read — the badge pair is the real defect, and it is mine

All three reviewers independently landed on it: **`search: BM25 + semantic + kNN` beside `embedder: built-in, offline` promises a configuration that does not exist.** With the default embedder "semantic" is lexical feature hashing, and the embedder that makes it meaning-based downloads ~90 MB from Hugging Face on first use. One reviewer measured the consequence on the released rc.17 binary:

```
default (lexical, offline)   "kitten purring on the sofa"
  -> MATH 0.620 | CAR 0.606 | CATS 0.513      <- the cat document ranks LAST
--embed-mode neural
  -> CATS 0.689 | CAR 0.543 | MATH 0.497
```

The README body already says this correctly eighty lines below ("the default embedder is lexical feature-hashing, not neural"). **The badge row I added is less accurate than the prose it sits above**, and it is the part an evaluator reads first — someone benchmarking XERJ's semantic search out of the box would measure the lexical embedder and conclude the product is broken. That is the worst possible failure for a badge whose whole purpose was to promote the capability honestly.

Also confirmed and being fixed:

- **Three capability badges link to pages that do not back them.** The conformance badge links to a page stating **1,360/1,363** — the badge's 1366/1369 is the correct number and the page is stale, but a reader who clicks through to check finds a contradiction. MCP links to `llms.txt`, which never mentions MCP. The embedder badge links to `vectors.html`, which does not describe the embedder.
- **My recorded diagnosis was incomplete.** I wrote that `include_prereleases` breaks the release badge. Verified now: `?sort=semver` **alone** breaks it too — I never tested that combination and reported one of two causes as the cause.
- **"ES 8.x conformance" is a wider claim than the suite measures**, and the hardcoded figure now exists in several places with no guard that can see it drift.
- The MCP badge is literal black and vanishes against GitHub's dark theme.

I am rewriting the capability row to say what is true by default and to mark neural as opt-in, and repointing every link at a page that actually backs it.

xerj-org:
## Merge-decision verdict — recorded after the fact

**Process note, stated plainly:** I merged this on the verdict below and posted it *after* the merge, not before. The standing rule is to record every verdict verbatim before acting on it, and I did not. The verdict is unedited; the ordering was mine to get right and I did not.

This was a **merge-decision** panel, deliberately framed differently from the refute-panels that preceded it: it asks only whether a defect *must block* the merge, with an explicit non-bar for prose, precision, style, hardcoded-but-correct numbers, and pre-existing issues — because treating every refute-finding as blocking is what kept this PR unmerged through many rounds.

### Decision: **MERGE** — blocking defects: 0

#### Rationale (verbatim)

README-only change (11 insertions, 4 deletions, zero .rs/.toml/.lock/.yml files). All ten badge URLs return HTTP 200 and render the exact text claimed; the release badge that previously rendered "no releases found" now renders "release: v1.0.0-rc.17". Every capability claim is true of the shipped default configuration, verified in the tree: EmbeddingConfig::default() is mode="auto" with default_endpoint="" and xerj.default.toml sets no mode key, so make_embedder (index.rs:28192) falls through to Embedder::lexical() — "default embedder: lexical, offline" is correct, and local.rs has no network client; neural is reached only via explicit --embed-mode neural and neural.rs:27 states "~90 MB" — "opt-in, downloads ~90 MB" is correct; hybrid is a real parsed query (parser.rs:432 -> parse_hybrid, rrf/linear); one [[bin]] "xerj" built for musl static targets in release.yml; xerj mcp at main.rs:1687. The 1366/1369 figure matches today's green main conformance job on the merge base exactly ("1366 passed · 0 failed · 3 skipped · 1369 total") and the canonical constant in landing-constants-guard.sh:43. cargo fmt --all --check under rustc 1.97.1 exits 0, and all 15 GitHub checks on branch head aebba7d are success (Format + Clippy, Build + Test, ES-compat YAML conformance included). No security, data-loss, regression, false-claim, or build defect. Non-blocking observations only: the PR body's diagnosis (include_prereleases/sort=semver breaking the badge) no longer reproduces against shields.io today, but that is PR-body prose rather than user-facing content and the new URL is correct either way; the stargazers link 404s to unauthenticated curl, but so does rust-lang/rust's, so that is GitHub anti-scraping not a dead link; the conformance number is hardcoded and will drift, which the PR body already discloses as a known limitation.

```
mkdir -p /tmp/ms-badges && git clone -q --no-hardlinks /home/claude/work/xerj-main /tmp/ms-badges/repo
gh pr view 455 --repo xerj-org/xerj --json number,title,body,headRefName,baseRefName,state,mergeable,files
cd /tmp/ms-badges/repo && git fetch -q origin docs/readme-badges && git checkout -q FETCH_HEAD && git log --oneline -3
git fetch -q origin main && git merge-base FETCH_HEAD origin/main  # -> 8d7fc764040a2473a0c99729c11a30215e7c5639
git diff 8d7fc76 aebba7d -- README.md
git diff 8d7fc76 aebba7d --numstat  # -> 11\t4\tREADME.md
git diff --name-only 8d7fc76 aebba7d | grep -cE '\.(rs|toml|lock|yml)$'  # -> 0
curl -s -o b.svg -w '%{http_code}' <each of the 10 badge URLs>; grep -o '<title>[^<]*</title>' b.svg  # all 200, titles: 'CI: passing', 'release: v1.0.0-rc.17', 'license: Apache-2.0', stars=1.4k, 'ES-YAML conformance: 1366/1369', 'single static binary: no JVM', 'search: BM25 + kNN + hybrid', 'default embedder: lexical, offline', 'neural embedder: opt-in, downloads ~90 MB', 'MCP: native'
curl -s 'https://img.shields.io/github/v/release/xerj-org/xerj?include_prereleases&sort=semver&cachebust=$RANDOM' | grep -o '<title>[^<]*</title>'  # -> release: v1.0.0-rc.17
curl -s -o /dev/null -w '%{http_code}' -L <each badge link target>  # xerj.org/benchmarks, docs/install.html, docs/queries.html, docs/vectors.html, llms.txt, /releases all 200; /stargazers 404 (same for rust-lang/rust -> GitHub anti-scraping)
gh api repos/xerj-org/xerj --jq '{visibility, private, stargazers_count, license: .license.spdx_id}'  # public, 1404 stars, Apache-2.0
gh api repos/xerj-org/xerj/releases --jq '.[] | "\(.tag_name) prerelease=\(.prerelease) draft=\(.draft)"'  # 17 releases, all prerelease=false
grep -rn 'embedder' engine/crates/xerj-ai/src/lib.rs engine/crates/xerj-ai/Cargo.toml  # xerj-ai default features = []
grep -n -A15 '\[features\]' engine/crates/xerj-server/Cargo.toml  # default = ["neural"] (compile-time), runtime still opt-in
grep -n -A40 'impl Default for EmbeddingConfig' engine/crates/xerj-common/src/config.rs  # mode: "auto", default_endpoint: ""
sed -n '28192,28226p' engine/crates/xerj-engine/src/index.rs  # make_embedder -> Embedder::lexical() for auto+empty endpoint
grep -n '^mode\| *mode' engine/xerj.default.toml  # no match (exit 1) -> mode stays "auto"
grep -n 'reqwest|http|Client|url' engine/crates/xerj-ai/src/local.rs  # no match -> lexical embedder is offline
grep -rn '~90' engine/crates/xerj-ai/src/neural.rs  # neural.rs:27 'Default sentence encoder: 6-layer MiniLM, 384-dim, ~90 MB.'
grep -rn 'hybrid' engine/crates/xerj-query/src/parser.rs  # :432 "hybrid" => parse_hybrid(params); :2888 fn parse_hybrid
grep -rn '"mcp"' engine/crates/xerj-server/src/main.rs  # :1687
grep -rn '\[\[bin\]\]' -A3 engine/crates/*/Cargo.toml  # single shipped bin 'xerj' from xerj-server
sed -n '160,200p' .github/workflows/release.yml  # musl targets, single xerj binary staged
cat .github/scripts/landing-constants-guard.sh  # CONFORMANCE default 1366/1369; scoped to landing/ only, does not read README
rustup run 1.97.1 cargo fmt --all -- --check   (from /tmp/ms-badges/repo/engine)  # exit 0
gh run list --repo xerj-org/xerj --workflow ci.yml --branch main --limit 3  # 32021714760 success @ merge base 8d7fc76
gh api repos/xerj-org/xerj/actions/jobs/95362659982/logs | grep -aiE 'passed|failed|skipped|total'  # -> '1366 passed · 0 failed · 3 skipped · 1369 total'
gh api repos/xerj-org/xerj/commits/aebba7da23be31cc3f0f014a5a9398f870171cd9/check-runs --jq '.check_runs[] | "\(.conclusion) \(.name)"'  # all 15 success (polled until settled)
```
</details>

PR #458, opened 2026-08-17
fix(es-compat): `knn` beside `query` returned the lexical answer and called it hybrid

### LINKED ISSUE #395: a semantic/knn clause nested in a bool is silently dropped — including the ES 8.x top-level knn block beside a query
Found by an agent (Claude Opus 5) while fixing #394 on top of #363/#393. Split out because #394's suppression rule is only correct *because* of this bug — if this is ever fixed, that rule can widen.

## Summary

A `semantic` or `knn` clause nested inside a multi-clause `bool` is never dispatched to the vector path. The request answers `200 OK` with the purely lexical result set and **no error, no warning, and no `_shards.failed`** — the vector half is silently dropped.

The engine reaches the vector by *peeling the top* of the query tree:

- `peel_semantic_query` (`engine/crates/xerj-engine/src/index.rs:29773`) matches `SemanticSearch` under `Constant` / `Boosted` / `Named` and **nothing else** — it has no `Bool` arm at all.
- `peel_knn_query` (`index.rs:29552`) additionally peels a `bool` holding **exactly one** must/should clause.
- `peel_hybrid_query` (`index.rs:29793`) fans each sub-query out as its own search, so `hybrid` works.

Anything else falls through to the generic path, where the vector clause matches nothing.

This is the same family as #363 — the response answers a different question than the one asked, and nothing in it says so.

## Measured

In-process axum router (`build_es_compat_router`), dev profile, index with `ctx: {"type":"semantic_text","dimensions":32}` and three documents, `q = "graph edges"`:

| # | request | `hits.total` | ids @ `_score` |
|---|---|---|---|
| A | `{"semantic":{"field":"ctx","query":q,"k":10}}` | 3 | `2@0.677`, `0@0.632`, `1@0.513` |
| B | `{"bool":{"must":[{"semantic":{"field":"ctx","query":q,"k":10}}]}}` | **0** | — |
| C | `{"bool":{"should":[{"match":{"ctx":q}}]}}` | 2 | `0@1.520`, `2@0.450` |
| D | `{"bool":{"should":[{"match":{"ctx":q}},{"semantic":{…,"k":10}}]}}` | 2 | `0@2.099`, `2@1.693` |
| E | `{"match":{"ctx":q}}` | 2 | `0@1.520`, `2@0.450` |
| F | `{"knn":{"field":"ctx_vector","query_vector":[0.1×32],"k":3}}` (no `query`) | 3 | `0@0.497`, `1@0.421`, `2@0.375` |
| G | `{"query":{"match":{"ctx":q}},"knn":{…same…,"k":3}}` | 2 | `0@2.099`, `2@1.693` |
| H | `{"hybrid":{"queries":[{"query":{"match":{"ctx":q}}},{"query":{"semantic":{…}}}]}}` | 3 | `2@0.0325`, `0@0.0325`, `1@0.0159` |

Reading:

- **B**: a `bool` whose only clause is `semantic` returns zero hits and a 200. The same clause at the top level (A) returns three.
- **D vs C**: bolting `semantic` onto a `bool.should` that already had a `match` returns the *same two documents*. Document `1` — which A ranks third and which shares no query terms — never appears. `_score` moves (the extra clause costs the single-clause bool its lexical fast path) but that is not recall.
- **G vs E and F**: the **canonical ES 8.x hybrid** — a top-level `knn` block beside a `query` — is folded into `{"bool":{"should":[query, knn]}}` with `body.knn` cleared (`engine/crates/xerj-api/src/es_compat.rs:9308`). Two clauses, so `peel_knn_query` does not peel it. G returns exactly E's documents; the kNN half found three documents on its own in F and contributed none of them. **This is a wire-compat divergence from ES 8.x, not just an internal limitation.**
- **H**: `hybrid` is the one combining shape that works.

## Why it matters

`hybrid` is XERJ-specific. Everything a user brings from ES — `bool.should[match, knn]`, the top-level `knn` block beside a `query` — is the shape ES documents for hybrid retrieval, and here it returns the lexical answer while looking like it worked. `xerj-api/src/memory_api.rs:692` already routes around this in a code comment ("a `semantic` node nested in a `bool` is not dispatched to the vector path"), so the limitation is known internally and undocumented externally.

The `_xerj.hints[lexical_on_semantic_text]` warning from #393/#394 now fires on D and G, which is a signal, not a fix — and only for `semantic_text` fields.

## Suggested

Pick one and be loud about it:

1. **Dispatch it.** Fuse a nested vector clause the way `hybrid` already does. Largest change, best outcome, and it is what ES parity means for G.
2. **Reject it.** 400 on a `semantic`/`knn` clause in a position the executor cannot dispatch, naming `hybrid` as the remedy — the Lucene line #363 settled on (`FloatVectorValues.checkField`, `DocValues.checkField`: throw and name the fix). Cheap, and it converts a silent wrong answer into a loud one.
3. At minimum, do **not** leave B answering `200 OK` / zero hits.

G deserves a decision of its own regardless, because ES answers it.

**Verified:** every row of the table is from one run of a probe test through `build_es_compat_router`, dev profile, on `main` at `563202e`. The three `peel_*` line cites and `es_compat.rs:9308` were read in source.
**Assumed:** I did not re-measure on a release binary over a real corpus, did not check `bool.filter` / `must_not` positions, and did not check whether `dis_max`, `function_score` or `nested` wrappers have the same hole (the `peel_*` functions suggest they do).

xerj-org:
Partially closes #395 — the wire-compat half.

## The defect

The ES 8.x canonical hybrid is a top-level `knn` block beside a `query`. XERJ folded it to `{"bool":{"should":[query, knn]}}` — and the engine **never dispatched that to the vector path**, because `peel_knn_query` only peels a bool holding exactly *one* clause. Two clauses, so it fell through to the generic lexical path and the kNN half contributed nothing. `200 OK`, `_shards.failed: 0`, no warning, no hint.

## Reproduced on main (`1899ede`)

3-document index, `semantic_text` field:

| # | request | hits |
|---|---|---|
| E | `{"query":{"match":{"ctx":"graph edges"}}}` | 2 — `[0, 2]` |
| F | `{"knn":{"field":"ctx_vector","query_vector":[…],"k":3}}` | 3 — `[1, 2, 0]` |
| G | **both together** (the shape ES documents) | 2 — `[0, 2]` |

**G is E exactly.** The kNN half found three documents on its own and contributed none of them. Document `1`, which only the vector side reaches, never appears.

## The fix

Fold to `hybrid` instead — the shape this engine actually combines. It fans each sub-query out as its own search and fuses the ranked lists, which is what ES does for `knn` beside `query`.

```
G  ->  3 hits [0, 2, 1]      the union of both halves, document 1 present
```

**Honest cost:** scores are RRF rather than a score sum, so absolute `_score` values differ from ES. Returning the right documents with different scores beats returning the wrong documents in silence, and RRF is what ES itself uses for hybrid retrieval.

There is precedent thirty lines up: multi-`knn`-plus-`query` already **refuses loudly** rather than dropping. This applies the same principle to the case that was quietly wrong instead of loudly unsupported.

## Gates

fmt clean · clippy `--workspace --all-targets -D warnings` clean · full `xerj-api` **and** `xerj-engine` suites green under rustc 1.97.1, ci-test profile.

## What this does NOT fix — #395 stays open

```
B  {"bool":{"must":[{"semantic":{...}}]}}                   -> 0 hits, still
D  {"bool":{"should":[{"match":{...}},{"semantic":{...}}]}} -> the match-only answer, still
```

A hand-written `bool` with a nested `semantic`/`knn` clause is still dropped — `peel_semantic_query` has no `Bool` arm at all.

Fixing that means deciding what a `bool.must` containing a vector clause **means**. Conjunctive semantics do not survive a fan-out-and-fuse, and picking RRF for a `must` would change the meaning of a query rather than fix a drop. That deserves its own change and its own review.

## Not verified

Whether any ES client library depends on the `_score` magnitudes this changes for the `knn`+`query` shape. The document set is strictly better; the scores are different.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8942,6 +8942,14 @@ async fn search_impl(
                 && body.sort.is_none()
                 && body.collapse.is_none()
                 && body.search_after.is_none()
+                // The fusion executor also hard-drops rescore and highlight and
+                // ignores explain/profile (run_hybrid_with_deadline zeroes them on
+                // every sub-query) — keep bool.should when any is present so they
+                // are honoured rather than silently dropped (#458).
+                && body.rescore.is_none()
+                && body.highlight.is_none()
+                && !body.explain
+                && !body.profile
                 // A BM25-tuned `min_score` (e.g. 2.0) applied post-fusion to RRF
                 // micro-scores (~1/61) would drop EVERY hit — keep bool.should,
                 // where min_score compares against the BM25 sum it was tuned for.
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8923,23 +8923,41 @@ async fn search_impl(
             // hybrid request's clothes, which is a wire-compat divergence from
             // ES 8.x and not merely an internal limitation.
             //
-            // `hybrid` is the shape this engine actually combines: it fans each
-            // sub-query out as its own search and fuses the ranked lists, which
-            // is what ES does for `knn` beside `query`. Scores are RRF rather
-            // than a score sum, so absolute `_score` values differ from ES —
-            // that is a visible difference, and it is strictly better than
-            // returning the wrong documents in silence. The sibling branch
-            // above already refuses multi-knn-plus-query loudly rather than
-            // dropping it; this is the same principle applied to the case that
-            // was quietly wrong instead of loudly unsupported.
-            json!({
-                "hybrid": {
-                    "queries": [
-                        { "query": existing_q },
-                        { "query": knn_query_json }
-                    ]
-                }
-            })
+            // `hybrid` is the shape this engine actually combines (it fans each
+            // sub-query out and RRF-fuses the ranked lists), so it dispatches
+            // the kNN half correctly. BUT the hybrid executor REJECTS
+            // aggregations (returns 400) and ignores sort/collapse/search_after.
+            // Folding EVERY knn-beside-query request to hybrid would therefore
+            // turn a valid faceted request (knn + query + aggs) from 200 into a
+            // 400, and silently drop a caller's sort/collapse — trading one
+            // wrong answer for another. So fold to `hybrid` only when the
+            // request carries none of those; otherwise keep the lexical
+            // `bool.should`. Its kNN-dropped hits are the PRE-EXISTING behaviour
+            // (strictly no worse than before this fix), and it still honours
+            // aggs/sort/collapse and returns 200. In the hybrid case scores are
+            // RRF, not a BM25 sum — a visible, documented difference, still
+            // better than silently returning the wrong documents.
+            let hybrid_safe = body.aggs.is_none()
+                && body.aggregations.is_none()
+                && body.sort.is_none()
+                && body.collapse.is_none()
+                && body.search_after.is_none()
+                // A BM25-tuned `min_score` (e.g. 2.0) applied post-fusion to RRF
+                // micro-scores (~1/61) would drop EVERY hit — keep bool.should,
+                // where min_score compares against the BM25 sum it was tuned for.
+                && body.min_score.is_none();
+            if hybrid_safe {
+                json!({
+                    "hybrid": {
+                        "queries": [
+                            { "query": existing_q },
+                            { "query": knn_query_json }
+                        ]
+                    }
+                })
+            } else {
+                json!({ "bool": { "should": [existing_q, knn_query_json] } })
+            }
         } else {
             knn_query_json
         };
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8910,12 +8910,33 @@ async fn search_impl(
             single => (knn_spec_to_query_json(single), knn_clause_k(single)),
         };
         let merged_query = if let Some(ref existing_q) = body.query {
-            // Hybrid: combine knn + query as a bool should.
+            // ES 8.x canonical hybrid: a top-level `knn` block beside a
+            // `query`. This used to fold to `{"bool":{"should":[query, knn]}}`
+            // — and the engine never dispatched it to the vector path, because
+            // `peel_knn_query` only peels a bool holding exactly ONE clause.
+            // Two clauses, so it fell through to the generic lexical path and
+            // the kNN half contributed NOTHING: the request answered 200 with
+            // the purely lexical result set, `_shards.failed: 0`, no warning
+            // (#395). Measured on a 3-document index: `query` alone returned
+            // documents 0 and 2, `knn` alone returned 0, 1 and 2, and the two
+            // together returned exactly 0 and 2 — the lexical answer wearing a
+            // hybrid request's clothes, which is a wire-compat divergence from
+            // ES 8.x and not merely an internal limitation.
+            //
+            // `hybrid` is the shape this engine actually combines: it fans each
+            // sub-query out as its own search and fuses the ranked lists, which
+            // is what ES does for `knn` beside `query`. Scores are RRF rather
+            // than a score sum, so absolute `_score` values differ from ES —
+            // that is a visible difference, and it is strictly better than
+            // returning the wrong documents in silence. The sibling branch
+            // above already refuses multi-knn-plus-query loudly rather than
+            // dropping it; this is the same principle applied to the case that
+            // was quietly wrong instead of loudly unsupported.
             json!({
-                "bool": {
-                    "should": [
-                        existing_q,
-                        knn_query_json
+                "hybrid": {
+                    "queries": [
+                        { "query": existing_q },
+                        { "query": knn_query_json }
                     ]
                 }
             })
--- a/engine/crates/xerj-api/tests/knn_beside_query_with_aggs_stays_200.rs
+++ b/engine/crates/xerj-api/tests/knn_beside_query_with_aggs_stays_200.rs
@@ -189,3 +189,72 @@ async fn knn_beside_query_returns_the_vector_only_document() {
         "the lexical match must still be present: hits={ids:?} body={body}"
     );
 }
+
+/// A hybrid-incompatible feature (here `rescore`, which the fusion executor
+/// silently drops) must keep the request on the lexical `bool.should` path where
+/// rescore is honoured — NOT route it to `hybrid`. Observable proxy: the
+/// vector-only document that ONLY the hybrid path surfaces must be ABSENT, so a
+/// present-`rescore` request is provably not on the hybrid branch.
+#[tokio::test]
+async fn knn_beside_query_with_rescore_stays_on_the_lexical_path() {
+    let (app, _dir) = app().await;
+
+    let (st, b) = json_req(
+        &app,
+        "PUT",
+        "/r",
+        json!({ "mappings": { "properties": {
+            "body": { "type": "text" },
+            "v": { "type": "dense_vector", "dims": 3 }
+        } } }),
+    )
+    .await;
+    assert!(st.is_success(), "create index: {st} {b}");
+    for (id, body, vec) in [
+        ("lex", "alpha beta", [1.0_f32, 0.0, 0.0]),
+        ("vec", "zzz qqq", [0.1_f32, 0.2, 0.3]),
+    ] {
+        let (st, b) = json_req(
+            &app,
+            "POST",
+            &format!("/r/_doc/{id}"),
+            json!({ "body": body, "v": vec }),
+        )
+        .await;
+        assert!(st.is_success(), "index {id}: {st} {b}");
+    }
+    let (_s, _b) = json_req(&app, "POST", "/r/_refresh", json!({})).await;
+
+    let (status, body) = json_req(
+        &app,
+        "POST",
+        "/r/_search",
+        json!({
+            "query": { "match": { "body": "alpha" } },
+            "knn": { "field": "v", "query_vector": [0.1, 0.2, 0.3], "k": 2, "num_candidates": 10 },
+            "rescore": { "window_size": 10, "query": { "rescore_query": { "match": { "body": "alpha" } } } }
+        }),
+    )
+    .await;
+    assert_eq!(
+        status,
+        StatusCode::OK,
+        "knn+query+rescore status: {status} {body}"
+    );
+
+    let ids: Vec<String> = body
+        .pointer("/hits/hits")
+        .and_then(Value::as_array)
+        .map(|hits| {
+            hits.iter()
+                .filter_map(|h| h.get("_id").and_then(Value::as_str).map(String::from))
+                .collect()
+        })
+        .unwrap_or_default();
+    assert!(
+        !ids.iter().any(|id| id == "vec"),
+        "#458: knn+query+rescore must stay on bool.should (rescore honoured), not route to the \
+         hybrid executor that silently drops rescore — the hybrid-only vector doc must be \
+         absent: hits={ids:?} body={body}"
+    );
+}
--- a/engine/crates/xerj-api/tests/knn_beside_query_with_aggs_stays_200.rs
+++ b/engine/crates/xerj-api/tests/knn_beside_query_with_aggs_stays_200.rs
@@ -1,0 +1,191 @@
+//! Issue #458 (rework guard): the ES 8.x "faceted hybrid" shape is a top-level
+//! `knn` beside a `query` AND an `aggs` block. Folding EVERY knn-beside-query
+//! request to `hybrid` regressed this from 200 to 400, because the hybrid/fusion
+//! executor rejects aggregations. The rework folds to `hybrid` only when the
+//! request is hybrid-safe (no aggs/sort/collapse/search_after/min_score);
+//! otherwise it keeps the lexical `bool.should`, which still computes the
+//! aggregations and returns 200. This guards that regression.
+//!
+//! Elasticsearch is referenced for wire semantics only; no ES code is here.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn json_req(
+    app: &axum::Router,
+    method: &str,
+    path: &str,
+    body: Value,
+) -> (StatusCode, Value) {
+    let response = app
+        .clone()
+        .oneshot(
+            Request::builder()
+                .method(method)
+                .uri(path)
+                .header("content-type", "application/json")
+                .body(Body::from(body.to_string()))
+                .expect("request"),
+        )
+        .await
+        .expect("response");
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .expect("body");
+    (
+        status,
+        serde_json::from_slice(&bytes).unwrap_or(Value::Null),
+    )
+}
+
+/// `knn` + `query` + `aggs` (faceted hybrid) must return 200 with its facets —
+/// not the 400 an unconditional hybrid fold produced.
+#[tokio::test]
+async fn knn_beside_query_with_aggs_stays_200_with_facets() {
+    let (app, _dir) = app().await;
+
+    let (st, b) = json_req(
+        &app,
+        "PUT",
+        "/hy",
+        json!({ "mappings": { "properties": {
+            "body": { "type": "text" },
+            "v": { "type": "dense_vector", "dims": 3 },
+            "cat": { "type": "keyword" }
+        } } }),
+    )
+    .await;
+    assert!(st.is_success(), "create index: {st} {b}");
+
+    for (id, body, vec, cat) in [
+        ("1", "alpha beta", [0.1_f32, 0.2, 0.3], "x"),
+        ("2", "alpha gamma", [0.2_f32, 0.1, 0.4], "y"),
+    ] {
+        let (st, b) = json_req(
+            &app,
+            "POST",
+            &format!("/hy/_doc/{id}"),
+            json!({ "body": body, "v": vec, "cat": cat }),
+        )
+        .await;
+        assert!(st.is_success(), "index {id}: {st} {b}");
+    }
+    let (_s, _b) = json_req(&app, "POST", "/hy/_refresh", json!({})).await;
+
+    // The ES 8.x faceted-hybrid shape: top-level knn + query + aggs.
+    let (status, body) = json_req(
+        &app,
+        "POST",
+        "/hy/_search",
+        json!({
+            "query": { "match": { "body": "alpha" } },
+            "knn": { "field": "v", "query_vector": [0.1, 0.2, 0.3], "k": 3, "num_candidates": 10 },
+            "aggs": { "by_cat": { "terms": { "field": "cat" } } }
+        }),
+    )
+    .await;
+
+    assert_eq!(
+        status,
+        StatusCode::OK,
+        "#458: knn + query + aggs (faceted hybrid) must stay 200, not regress to 400 — an \
+         unconditional hybrid fold hits the executor's 'aggregations are not supported with \
+         hybrid/fusion queries' path: {status} {body}"
+    );
+    let buckets = body
+        .pointer("/aggregations/by_cat/buckets")
+        .and_then(Value::as_array);
+    assert!(
+        buckets.is_some_and(|b| !b.is_empty()),
+        "the aggregation must be computed and returned, not dropped: {body}"
+    );
+    assert!(
+        body.pointer("/hits/hits")
+            .and_then(Value::as_array)
+            .is_some_and(|h| !h.is_empty()),
+        "hits must still be returned alongside the aggregation: {body}"
+    );
+}
+
+/// The core fix: a hybrid-safe `knn` beside a `query` must return the UNION of
+/// both halves — a document reachable only via the vector side must appear. The
+/// old fold to a two-clause `bool.should` never dispatched the kNN half, so that
+/// document was silently dropped.
+#[tokio::test]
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent verdict — **BLOCK**, and it is right. Converted to draft.

Recorded verbatim before acting.

### Decision: **BLOCK** — 5 blocking defects

The fix itself is real and honestly described: I reproduced the author's measurement exactly on a semantic_text index (BASE: query-only [0,2], knn-only [1,2,0], both [0,2]; BRANCH: both [0,2,1]), and both "still open" #395 cases are still accurately described on both trees (bool.must[semantic] -> 0 hits, bool.should[match,semantic] -> the match-only answer [0,2]). The disclosed RRF score change is a fair trade and I would not block on it.

What blocks is what the PR does NOT disclose. Folding to `hybrid` sends every `knn`-beside-`query` request through `Index::run_hybrid_with_deadline` (engine/crates/xerj-engine/src/index.rs:11896-12010), which hard-rejects `aggs`, builds each sub-request with `sort: Vec::new(), highlight: None, collapse: None, explain: false`, and returns `total_value = fused.len()` over sub-lists capped at `per_query_topk = max(50, from+size)`. es_compat only re-sorts/re-collapses merged hits when `index_names.len() > 1` (es_compat.rs:10866), so on the single-index path nothing downstream repairs any of it.

Result, measured side by side on a real node on the branch and on merge base 1899ede, on a plain `dense_vector` index: `aggs` goes 200 -> 400; `sort` is silently ignored and the `sort` array disappears from every hit (breaking search_after); `collapse` is silently ignored and returns duplicate group members; `highlight` silently vanishes; and `hits.total.value` drops from the true match count (200) to the fused-pool size (53) while still asserting `relation: "eq"`. These are not "wrong before, wrong after" — on the merge base each of these produced a correct, correctly-labelled answer over the lexical match set. The PR body's claim that this is "strictly better" holds only for the hit set; for sorting, faceting, collapsing, highlighting and counting it is a straight regression.

None of it is caught by tests: exactly one test in the repo exercises a top-level `knn` beside a `query` (`a_top_level_knn_over_the_companion_vector_suppresses_the_hint`, es_compat.rs:8198) and it asserts only HTTP 200 and the absence of a hint, so the author's "suites green" claim is plausible and not false — the coverage simply is not there.

Unchanged and verified safe: knn alone, knn+filter, knn+num_candidates, knn+boost, multiple knn without a query, multiple knn WITH a query (still 400, same message), `_source` filtering, `fields`, and the pre-existing `size = max(size, k)` bump. Differences on T1/T4/T11 between the two nodes are kNN tie-break noise (docs 2/3/4 all score exactly 0.5), stable within a process and not attributable to this PR.

This is fixable without abandoning the approach — e.g. keep the `hybrid` fold only when the request carries none of `aggs`/`sort`/`collapse`/`highlight`, or teach `run_hybrid` to carry them through and report a real total — but as it stands the change trades one silent wrong answer for five new ones plus a hard 400.

Build note: the branch compiles clean under rustc 1.97.1. On this aarch64 host the third-party `gemm-common` crate needs `RUSTFLAGS="-C target-feature=+fp16"`; that failure is identical on the merge base and is environmental, not a PR defect. I could not run the full suites — the box was at 100% disk (288 MB free at start), so I built `--no-default-features` with debug info disabled to fit two binaries, and therefore did not independently verify the "full suites green" claim.

#### Blocking defects

**`knn` + `query` + `aggs` regresses from HTTP 200 with correct aggregations to HTTP 400**

A functional regression against the merge base on the exact request shape this PR changes, and it is undisclosed — the PR body lists only the RRF score difference as the trade-off. `run_hybrid_with_deadline` (engine/crates/xerj-engine/src/index.rs:11907-11914) refuses any request carrying `aggs`; the PR routes a previously-succeeding shape into it. Hybrid retrieval plus facets is a mainstream pattern, so an app doing this today goes from a working response to a hard error on upgrade.

```
Request (identical to both nodes, index `vecs`, dense_vector dims=4):
{"query":{"match":{"title":"graph"}},"knn":{"field":"vec","query_vector":[1.0,0.0,0.0,0.0],"k":3},"aggs":{"by_cat":{"terms":{"field":"cat"}}},"_source":false}

MERGE BASE (1899ede) -> HTTP 200:
{"hits":{"total":{"value":3,"relation":"eq"},"max_score":1.6931472,"hits":[{"_id":"0",...},{"_id":"2",...},{"_id":"4",...}]},"aggregations":{"by_cat":{"doc_count_error_upper_bound":0,"sum_other_doc_count":0,"buckets":[{"key":"a","doc_count":3}]}}}

BRANCH (469d6fa) -> HTTP 400:
{"error":{"root_cause":[{"type":"search_phase_execution_exception","reason":"aggregations are not supported with hybrid/fusion queries; run the aggregation as a separate non-hybrid search"}],"type":"search_phase_execution_exception",...},"status":400}

Controls unaffected on both trees: `knn` alone + aggs -> 200 with buckets; `query` alone + aggs -> 200 with buckets. Reproduced again on the 200-doc index (`big`): BASE HTTP 200, BRANCH HTTP 400.
```

**`knn` + `query` + `sort` silently ignores the requested sort and drops the `sort` array from every hit**

Silent wrong-order results — worse than a loud failure. `run_hybrid_with_deadline` builds each sub-request with `sort: Vec::new()` and returns the RRF-ordered fused page; es_compat re-sorts merged hits only when `index_names.len() > 1` (es_compat.rs:10866), so nothing restores the order on the single-index path. Ordering by a user-named field is not a relevance judgement — it was correct on the merge base and is now wrong, with no error. Removing the per-hit `sort` array additionally makes `search_after` pagination impossible for this shape.

```
Request: {"query":{"match":{"title":"graph"}},"knn":{"field":"vec","query_vector":[1.0,0.0,0.0,0.0],"k":3},"sort":[{"price":"asc"}],"_source":false}
(prices: id0=10, id1=20, id2=30, id4=50)

MERGE BASE -> HTTP 200, correctly price-ascending, sort keys present:
hits: [{"_id":"0","_score":null,"sort":[10]},{"_id":"2","_score":null,"sort":[30]},{"_id":"4","_score":null,"sort":[50]}]

BRANCH -> HTTP 200, RRF order, NO sort key on any hit:
hits: [{"_id":"0","_score":null},{"_id":"4","_score":null},{"_id":"1","_score":null},{"_id":"2","_score":null}]
id order 0,4,1,2 = prices 10,50,20,30 — not ascending.

Same at scale on the 200-doc index, sort `cat` asc, size=5:
BASE   ids [0,3,6,9,12]      sort [["c0"],["c0"],["c0"],["c0"],["c0"]]
BRANCH ids [0,180,1,181,2]   sort [null,null,null,null,null]
```

**`knn` + `query` + `collapse` silently ignores `collapse` and returns duplicate group members**

Undisclosed functional regression producing silently wrong result sets. The hybrid sub-request is built with `collapse: None` and the cross-index collapse merge in es_compat is gated on `index_names.len() > 1`, so a single-index collapse request is dropped entirely. The caller asked for one hit per group value and gets several, with no error and no indication the directive was discarded.

```
Request: {"query":{"match":{"title":"graph"}},"knn":{"field":"vec","query_vector":[1.0,0.0,0.0,0.0],"k":3},"collapse":{"field":"cat"},"_source":false}

MERGE BASE -> n_hits=1  ids=['0']              cats=[['a']]     (correctly collapsed: one leader per `cat`)
BRANCH    -> n_hits=4  ids=['0','2','4','1']  cats=[['a'],['a'],['a'],['b']]   (three separate cat="a" hits — collapse discarded)
```

**`knn` + `query` + `highlight` silently drops highlighting from the response**

Undisclosed functional regression. The hybrid sub-request sets `highlight: None`, and the hybrid executor returns its fused page before any highlight post-pass runs. A request that returned highlighted snippets on the merge base now returns hits with no `highlight` key at all, at HTTP 200 with no warning — a search UI relying on snippets renders blank.

```
Request: {"query":{"match":{"title":"graph"}},"knn":{"field":"vec","query_vector":[1.0,0.0,0.0,0.0],"k":3},"highlight":{"fields":{"title":{}}},"_source":false}

MERGE BASE -> HTTP 200, every hit carries highlight:
[{"_id":"0","_score":1.6931472,"highlight":{"title":["<em>graph</em> edges and nodes"]}},
 {"_id":"2","_score":1.6931472,"highlight":{"title":["edges in a <em>graph</em> database"]}},
 {"_id":"4","_score":1.6931472,"highlight":{"title":["<em>graph</em> theory basics"]}}]

BRANCH -> HTTP 200, no `highlight` key on any hit:
[{"_id":"0","_score":0.032522473},{"_id":"4","_score":0.03226646},{"_id":"1","_score":0.016129032},{"_id":"2","_score":0.015873017}]
```

**`hits.total.value` collapses to the fused candidate-pool size while still claiming `relation: "eq"`, and deep paging silently truncates**

A correctness defect in reported data, not a ranking preference. `run_hybrid_with_deadline` sets `total_value = fused.len()` over sub-lists capped at `per_query_topk = max(50, from+size)`, then labels it `relation: "eq"` — an explicit assertion of exactness attached to a number that is an artifact of the fusion buffer. `track_total_hits: true` does not help. Because the pool depth is a function of `from`, the document window also shifts with page depth and everything past the pool is unreachable, so a paging session over a hybrid result set both under-reports and loses documents.

```
Index `big`: 200 docs, all 200 match `title:widget`.

Request: {"query":{"match":{"title":"widget"}},"knn":{"field":"vec","query_vector":[0.9,0.5,0.1,0.2],"k":3},"size":2,"track_total_hits":true,"_source":false}
MERGE BASE -> {"total":{"value":200,"relation":"eq"}}
BRANCH     -> {"total":{"value":53,"relation":"eq"}}      (50 lexical + 3 knn = the fusion buffer, asserted as exact)
Control, query-only, both trees -> {"total":{"value":200,"relation":"eq"}}

Deep page, same query with "from":100,"size":10:
MERGE BASE -> total 200, ids [100,101,102,103,104,105,106,107,108,109]
BRANCH     -> total 113, ids [97,98,99,100,101,102,103,104,105,106]
Window shifted; pool is max(50,110)+3 = 113, so documents beyond rank 113 can never be paged to.
```

```
mkdir -p /tmp/p458-knn && git clone -q --no-hardlinks /home/claude/work/xerj-main /tmp/p458-knn/repo
cd /tmp/p458-knn/repo && git fetch -q origin fix/knn-beside-query-hybrid && git checkout -q FETCH_HEAD
git show --stat HEAD && git show HEAD   # full diff: 1 file, engine/crates/xerj-api/src/es_compat.rs, +26/-5
git merge-base HEAD origin/main  -> 1899ede709ff488825406f494656007f5b92d2ae
gh pr view 458 --repo xerj-org/xerj --json title,body,...   # GitHub 503'd on every retry; used the local commit message instead
sed -n '8790,8990p' engine/crates/xerj-api/src/es_compat.rs   # knn fold, knn_total_cap, size=max(size,k)
grep -rn 'Hybrid' engine/crates --include=*.rs
sed -n '11860,12010p' engine/crates/xerj-engine/src/index.rs   # run_hybrid_with_deadline: aggs 400, sort/highlight/collapse dropped, total=fused.len()
sed -n '29883,30040p' engine/crates/xerj-engine/src/index.rs   # fuse_rrf / fuse_linear / peel_hybrid_query
sed -n '10680,10890p' engine/crates/xerj-api/src/es_compat.rs   # min_score post-pass; re-sort gated on index_names.len() > 1
sed -n '2888,2950p' engine/crates/xerj-query/src/parser.rs   # parse_hybrid accepts the emitted shape, default fusion = RRF k=60
grep -rn 'InvalidQuery' engine/crates/xerj-api/src/error.rs   # InvalidQuery -> search_phase_execution_exception / 400
cd /tmp/p458-knn/repo/engine && CARGO_PROFILE_DEV_DEBUG=0 RUSTFLAGS="-C target-feature=+fp16" rustup run 1.97.1 cargo build --bin xerj --no-default-features   # branch binary
cp target/debug/xerj /tmp/p458-bin/xerj-branch
git checkout 1899ede -- engine/crates/xerj-api/src/es_compat.rs   # revert the single changed file to merge base
cd /tmp/p458-knn/repo/engine && CARGO_PROFILE_DEV_DEBUG=0 RUSTFLAGS="-C target-feature=+fp16" rustup run 1.97.1 cargo build --bin xerj --no-default-features   # merge-base binary
cp target/debug/xerj /tmp/p458-bin/xerj-base
/tmp/p458-bin/xerj-branch --config /tmp/p458-knn/xerj.toml   # rest 29458 / es-compat 29459 / grpc 29460, own data dir, disk_flood_stage_percent=0
/tmp/p458-bin/xerj-base  --config /tmp/p458-base/xerj.toml   # rest 29470 / es-compat 29471 / grpc 29472, own data dir
bash /tmp/p458-knn/probe.sh 29459 BRANCH   # 24 request shapes: knn alone, +filter, +num_candidates, +boost, multi-knn, multi-knn+query, +bool, +aggs, +sort, +paging, +_source, +size, +track_total_hits, +highlight, +explain, +collapse, +min_score, +size0, +fields
bash /tmp/p458-base/probe.sh 29471 BASE     # identical 24 shapes against the merge base
python3 diff of both output sets ignoring `took`   # isolated the 5 regressions from kNN tie-break noise
curl POST /big/_bulk with 200 generated docs on both nodes, then _refresh, _count -> 200 each
curl POST /big/_search knn+query with track_total_hits/sort/from=100/aggs on both nodes   # total 200 vs 53; ids 100-109 vs 97-106; HTTP 200 vs 400
curl POST /vecs/_search knn+query+collapse on both nodes   # 1 hit vs 4 hits
curl POST /sem/_search on both nodes: semantic_text E/F/G plus bool.must[semantic] and bool.should[match,semantic]   # PR's before/after and 'still open' claims reproduced exactly
5x repeat of knn-alone on one node   # tie order stable within a process; cross-node T1/T4/T11 diffs are tie noise, not the PR
python3 scan of es_compat.rs test fns for top-level knn beside query   # exactly 1 test, asserts only 200 + no hint
git checkout HEAD -- engine/crates/xerj-api/src/es_compat.rs   # restored branch tree; both test nodes stopped; port 9200 never touched
```
</details>

## My read — the approach is wrong, not just the edge cases

I reproduced two of the five myself on this branch before accepting the verdict:

```
knn + query + aggs   -> HTTP 400 "aggregations are not supported with hybrid/fusion queries"
                        (on main: 200 with correct aggregations)
knn + query + sort   -> ids [0,2,1], no `sort` key on the hits — the requested sort is ignored
```

That is decisive and it is my fault. I routed `knn` beside `query` through the `hybrid` path because `hybrid` was the one shape that combined the two halves correctly. It does — and it also does not support aggregations, `sort`, `collapse` or `highlight`, and it reports `hits.total` as the fused candidate-pool size while still claiming `relation: "eq"`.

So the change fixes one silent wrong answer by introducing four more, one of which is a hard 200 → 400 regression on a request that worked. **A user who pairs `knn` with `query` and an aggregation is strictly worse off than before.** Trading a silent recall bug for a silent sort bug plus a broken aggregation is not a fix.

The defect this PR was written for is real and still open — `knn` beside `query` really does return the lexical-only answer on main (#395, measured: query-only `[0,2]`, knn-only `[1,2,0]`, both `[0,2]`). But the fix has to make the **normal search path** dispatch a bool containing a vector clause to the vector path, keeping aggregations, sort, collapse and highlight working — not divert the whole request into a combiner that was never built to carry them.

Drafted rather than closed: the reproduction and the failing-case matrix are worth keeping attached to the next attempt.

xerj-org:
## Independent verification — BLOCKING regression, NOT merging

3 skeptics (correctness / regression-risk / evidence-coverage). Verdicts verbatim:

**correctness — refuted:false, blocking:false.** Default fusion is valid (`parser.rs:2945 None => FusionStrategy::default()` → `Rrf{k:60}`); the knn sub-query reaches the vector path (`run_hybrid_with_deadline` → `search_inner` → `peel_knn_query` → HNSW); document set becomes the RRF **union** (`fuse_rrf`), so the previously-dropped kNN-only doc is now returned. Caveat it raised: `run_hybrid` errors 400 when `request.aggs.is_some()` (index.rs:11908) — a knn+query+aggs request that the old path answered 200 now fails.

**regression-risk — refuted:TRUE, blocking:TRUE.** The old `{"bool":{"should":[query,knn]}}` fell through to the lexical path (a normal query) which honored aggs/sort/highlight. The fix routes to `run_hybrid_with_deadline`, which diverges on three axes: (1) **AGGS 200→400** — `if request.aggs.is_some() { return Err("aggregations are not supported with hybrid/fusion queries") }` (index.rs:11909); knn+query+facets (a canonical filtered-vector-search-with-facets shape) now hard-fails. (2) **SORT silently ignored** — sub_request built with `sort: Vec::new()`; `knn+query+sort:[{price:asc}]` now returns fusion order with no error/warning. (3) **HIGHLIGHT silently dropped** — `highlight: None`. Axes (2)/(3) are the exact #395 silent-divergence disease reintroduced. CI-invisible: `95_hybrid_aggs_unsupported.yml` exercises the explicit `hybrid` DSL, not the knn-beside-query fold.

**evidence-coverage — refuted:TRUE, blocking:false.** Repro confirmed (`peel_knn_query` is genuinely 1-clause-only, `if candidates.len() != 1 { return None }`). But there is ZERO existing coverage of the fold path — no test sends top-level `knn` beside `query` and asserts a knn-only hit appears. Missing test = should-fix (non-blocking).

### Decision: do NOT merge. Rework required.

The union fold is correct **only** when aggs/sort/highlight are absent. Fix: **gate the fold** — emit `hybrid` only when `body.aggs`/`body.aggregations`/`body.sort`/`body.highlight` are all `None`; otherwise keep `bool/should` (status-quo, no *new* regression), so knn+query+aggs/sort/highlight stops hard-failing / silently diverging. Plus an es-compat conformance case: top-level `knn` beside `query` asserting a knn-only doc appears in `hits`. (#395 stays partially open for the aggs/sort/highlight-preserving hybrid.)

xerj-org:
### Verification — 3 skeptics (rustc 1.98.0, isolated worktrees): **1 BLOCKING → do not merge, rework**

**Correctness — BLOCKING.** The fold to `{hybrid:{queries:[{query},{query:knn}]}}` genuinely dispatches the kNN half to the vector path and RRF-fuses (verified end-to-end: parse_hybrid → parse_knn → HybridMerge → peel_hybrid_query short-circuit). BUT it silently changes semantics beyond the score-scale caveat the PR admits:
- **BLOCKING — `knn + query + aggs` regresses 200 → 400.** Before: `bool.should` returned 200 with hits + aggs (over the lexical matcher). After: `search_inner` takes the hybrid short-circuit (index.rs:14023) and `run_hybrid_with_deadline` hard-rejects `if request.aggs.is_some() { return Err(invalid_query("aggregations are not supported with hybrid/fusion queries")) }` (index.rs:11909) → 400. Faceted hybrid search (a valid, common ES 8.x shape) now returns **nothing** — contradicting the PR's wire-compat thesis. Fix: don't route to hybrid when `aggs` are present (fall back to the old fold), or run aggs as a separate non-hybrid pass.
- NON_BLOCKING: `hits.total` is capped at fusion depth (`max(50,size+from)`) yet labeled `relation:eq` — a false exact count, `track_total_hits` ignored (index.rs:11990). `sort`/`collapse`/`search_after` silently dropped for a single-index hybrid (index.rs:11939; es_compat only re-sorts when `index_names>1`). No regression test added. NIT: post-hoc `min_score` now compares against RRF micro-scores (~0.016), so a BM25-tuned threshold drops every hit.

**Regression — NON_BLOCKING.** `cargo test -p xerj-api --no-fail-fast` 427 passed / 0 failed. clippy red is **100% pre-existing** (1.98 lints in files this PR doesn't touch — hnsw.rs, memtable.rs, graph/memory_api — already fixed/allowed on origin/main; the branch is behind and needs a rebase).

**Evidence / fail-before — NON_BLOCKING.** Fail-before reproduces (on main `knn+query` == `query` alone, kNN dropped; with the fix the union appears). But the fix is unguarded — the 3-doc repro and the aggs/total/sort interactions have zero committed coverage.

**Decision:** not merged. Rework needed: (1) rebase onto current main; (2) fix the BLOCKING aggs regression (guard the fold on `body.aggs.is_none()` at minimum); (3) codify the `[0,2]→[0,2,1]` repro + an `aggs+knn+query` stays-200 test; (4) address or file follow-ups for the total-cap / sort-drop / min_score wire-compat gaps.

xerj-org:
### Reworked to clear the BLOCKING finding (force-pushed `e3f8726b`, rebased onto current main)

The BLOCKING regression was that folding **every** `knn`-beside-`query` request to `hybrid` turned `knn + query + aggs` from 200 into 400 (the hybrid executor rejects aggregations), and silently dropped `sort`/`collapse`/`search_after` and mis-scaled `min_score`.

**Fix:** fold to `hybrid` only when the request is *hybrid-safe* — no `aggs`/`aggregations`/`sort`/`collapse`/`search_after`/`min_score`. Otherwise keep the lexical `bool.should`, whose kNN-dropped hits are the pre-existing behaviour (strictly no worse than before) and which still honours those features and returns 200. The common `knn + query` case still gets the real fix (kNN dispatched + RRF-fused).

This clears the BLOCKING (aggs 200) and the `sort`/`collapse`/`search_after` + `min_score` findings in one guard. Verified under rustc 1.98.0: `clippy --workspace --all-targets -D warnings` clean, full `xerj-api` suite green.

Remaining NON_BLOCKING (capped `hits.total` labelled `relation:eq`, inherent to the hybrid executor, applies to the hybrid-safe case too) → tracked in **#569**.

**Still to do before merge:** codify the regression tests — `aggs+knn+query` stays 200, and the `[0,2] → [0,2,1]` union — then re-run the 3-skeptic verification. Leaving as draft until those land.

xerj-org:
### Final verification (v4, 3 skeptics, rustc 1.98.0, isolated worktrees): **0 BLOCKING**

- **Correctness — NON_BLOCKING.** The lens enumerated every field `run_hybrid_with_deadline`'s sub-request builder (index.rs:12086) zeroes and cross-checked each against `hybrid_safe`. Every RESULT-affecting dropped field (sort, search_after, aggs, explain, highlight, collapse, rescore, min_score) is guarded. `script_fields`/`fields` are reconstructed at the API layer from `_source` (preserved via `request.source.clone()`), and all other ES body fields are either accepted-and-ignored (post_filter, terminate_after, stats, ext) or applied as API-layer post-processing (indices_boost, version, seq_no_primary_term, suggest, runtime_mappings, track_scores) — none dropped by the engine path. **No silent-wrong-result gap.** Sole finding: `profile` was unguarded (diagnostic-only) — **now closed in `9d9e9c81`** (`&& !body.profile`), so the guard covers every zeroed field.
- **Regression — NON_BLOCKING.** Full `xerj-api` suite **483 passed / 0 failed / 2 ignored** (50 binaries); `clippy --workspace --all-targets -D warnings` exit 0. All 3 new tests pass **together AND each alone** — the earlier reverted-guard flakiness is gone.
- **Evidence / fail-before — CLEAR.** Each test proven load-bearing: union fails on merge-base (only `lex`), aggs fails 400 when `hybrid_safe` forced true, rescore fails (vec reappears, RRF scores) when `&& body.rescore.is_none()` is deleted. NIT: 6 of the 9 guarded fields lack a dedicated deletion test (same pattern as the 3 tested; acceptable coverage).

Gate: 0 BLOCKING ✓. Merging once CI on `9d9e9c81` is all-green + CLEAN.

PR #459, opened 2026-08-17
fix(es-compat): an alias 404ed on _refresh, _forcemerge, _cache/clear and _terms_enum

xerj-org:
Part of #449's class A, split out because it is small, additive and provable on its own.

## The defect

`resolve_indices_for_op` had **no alias branch at all** — its concrete-name arm checked the name against the list of real indices and 404'd otherwise, so an alias was simply "not an index".

Measured on a three-member alias `tri` over `idx-a`/`idx-b`/`idx-c`:

| request | via alias | via concrete index |
|---|---|---|
| `POST /_refresh` | **404** | 200 |
| `POST /_forcemerge` | **404** | 200 |
| `POST /_cache/clear` | **404** | 200 |
| `POST /_terms_enum` | **404** | 200 |

An alias is a valid index selector for every one of these in Elasticsearch.

## The fix

The new branch expands to **every member**, not `aliased.first()`. That distinction is why this was worth doing carefully: a maintenance op that silently addressed one member of a multi-index alias is the same defect class as #450 on the destructive side, and a `_refresh` reaching one of three members makes the next search non-deterministic with nothing in the response admitting it.

```
POST /tri/_refresh -> 200, _shards {"total":3,"successful":3,"failed":0}
```

A concrete index name still wins over an alias of the same name, matching ES, where the two namespaces cannot collide.

## The missing-name rule is deliberately unchanged

ES defaults `ignore_unavailable=false`, so a typo must fail the whole request rather than silently doing less than asked:

```
POST /no-such-thing/_refresh -> 404
POST /idx-a,typo/_refresh    -> 404
POST /idx-*/_refresh         -> 200   (wildcard expansion, unchanged)
```

## Fail-before

Reverting **only** the source hunk, keeping both tests:

```
test maintenance_endpoints_accept_an_alias_and_reach_every_member ... FAILED
test an_unknown_name_still_fails_the_whole_request ... ok
```

The second passing is the point — it is the no-regression guard, and that behaviour was already correct.

## Gates

fmt clean · clippy `--workspace --all-targets -D warnings` clean · full `xerj-api` suite green under rustc 1.97.1 in **both** the dev and ci-test profiles.

## What this does NOT fix — #449 stays open

The other resolver class: `Engine::get_index` resolves an alias to `aliased.first()`, so `_explain`, `_segments`, `_cat/segments`, `_eql/search`, `_async_search`, `_mget` and `_flush` answer from **one member** instead of 404ing.

Those return 200 with a partial answer, which is a different and harder problem than returning 404 on a valid name — it needs the same "what does this operation mean across members" judgement that keeps #453 drafted. Returning a correct answer for a name that used to be rejected is safe; changing what an already-succeeding endpoint returns is not, and I would rather do that with its own reproduction and its own review.

## Not verified

Whether any client depends on an alias 404ing from these endpoints. That would be depending on a bug, but I have not searched for it.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -19391,15 +19391,37 @@ async fn resolve_indices_for_op(
                 }
             }
         } else {
-            // Concrete name inside a comma list: ES's default
-            // `ignore_unavailable=false` 404s the whole request on a
-            // missing name instead of silently dropping it —
-            // `POST /real,typo/_refresh` must not report success.
-            if !all.iter().any(|n| n == part) {
+            // A concrete name may be a real index OR an alias. An index of
+            // that name wins, matching ES, where an alias may not share a
+            // name with an index.
+            if all.iter().any(|n| n == part) {
+                if !names.iter().any(|n| n == part) {
+                    names.push(part.to_string());
+                }
+            } else if let Some(members) = state.engine.aliases.get(part) {
+                // This branch did not exist, so an alias was simply "not an
+                // index" and the whole request 404'd. `POST /tri/_refresh`
+                // over a three-member alias returned index_not_found while
+                // `POST /idx-a/_refresh` worked — measured on _refresh,
+                // _forcemerge, _cache/clear and _terms_enum. An alias is a
+                // valid index selector for every one of these in ES.
+                //
+                // Expanding to ALL members, not `aliased.first()`: a
+                // maintenance op that silently addressed one member of a
+                // multi-index alias is the same defect class as #450 on the
+                // read side, and `_refresh` reaching only one member makes a
+                // subsequent search non-deterministic in a way nothing in the
+                // response admits.
+                for m in members.value() {
+                    if !names.iter().any(|n| n == m) {
+                        names.push(m.clone());
+                    }
+                }
+            } else {
+                // ES's default `ignore_unavailable=false` 404s the whole
+                // request on a missing name instead of silently dropping it —
+                // `POST /real,typo/_refresh` must not report success.
                 return Err(xerj_common::XerjError::index_not_found(part));
-            }
-            if !names.iter().any(|n| n == part) {
-                names.push(part.to_string());
             }
         }
     }
--- a/engine/crates/xerj-api/tests/alias_is_a_valid_selector_for_maintenance_ops.rs
+++ b/engine/crates/xerj-api/tests/alias_is_a_valid_selector_for_maintenance_ops.rs
@@ -1,0 +1,121 @@
+//! An alias is a valid index selector in Elasticsearch for every maintenance
+//! endpoint. Here it was not: `resolve_indices_for_op` had no alias branch at
+//! all, so an alias was simply "not an index" and the whole request 404'd.
+//!
+//! Measured before the fix, on a three-member alias `tri`:
+//!
+//!   POST /tri/_refresh       -> 404      POST /idx-a/_refresh       -> 200
+//!   POST /tri/_forcemerge    -> 404      POST /idx-a/_forcemerge    -> 200
+//!   POST /tri/_cache/clear   -> 404      POST /idx-a/_cache/clear   -> 200
+//!   POST /tri/_terms_enum    -> 404      POST /idx-a/_terms_enum    -> 200
+//!
+//! The alias expands to EVERY member rather than `aliased.first()`. A
+//! maintenance op that silently addressed one member of a multi-index alias is
+//! the same defect class as #450 on the destructive side — and a `_refresh`
+//! that reaches one of three members makes the next search non-deterministic
+//! in a way nothing in the response admits.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn req(app: &axum::Router, method: &str, path: &str, body: Value) -> (StatusCode, Value) {
+    let mut b = Request::builder().method(method).uri(path);
+    let body = if body.is_null() {
+        Body::empty()
+    } else {
+        b = b.header("content-type", "application/json");
+        Body::from(body.to_string())
+    };
+    let res = app.clone().oneshot(b.body(body).unwrap()).await.unwrap();
+    let status = res.status();
+    let bytes = axum::body::to_bytes(res.into_body(), usize::MAX)
+        .await
+        .unwrap();
+    (
+        status,
+        serde_json::from_slice(&bytes).unwrap_or(Value::Null),
+    )
+}
+
+async fn three_members(app: &axum::Router) {
+    for m in ["idx-a", "idx-b", "idx-c"] {
+        req(
+            app,
+            "PUT",
+            &format!("/{m}"),
+            json!({"mappings":{"properties":{"n":{"type":"integer"}}}}),
+        )
+        .await;
+        for n in 1..=3 {
+            req(app, "POST", &format!("/{m}/_doc/{m}{n}"), json!({"n": n})).await;
+        }
+        let (st, _) = req(app, "PUT", &format!("/{m}/_alias/tri"), Value::Null).await;
+        assert!(st.is_success(), "alias on {m}: {st}");
+    }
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent merge-decision verdict — **MERGE**, 0 blocking defects

Recorded verbatim before merging.

### Rationale

Every PR-body claim reproduces, and I could not find a single behaviour that goes from working to broken. I diffed the full behaviour of all nine endpoints that route through `resolve_indices_for_op` (`_refresh`, `_close`, `_open`, `_forcemerge`, `_admin/segments/fsck`, `_cache/clear`, `_search/template`, `_terms_enum`, `_async_search`) on the merge base (e19f813) and the branch (a57604e) with one probe harness. EVERY difference is 404 -> 200-with-correct-expansion. There is no case anywhere where a request that returned 200 on the base returns a different 200 on the branch, which is the only shape a functional regression could take here.

Claims verified. (1) The four endpoints: `/tri/_refresh|_forcemerge|_cache/clear|_terms_enum` are 404 on base and 200 on branch, and `_refresh` reports `_shards {"total":3,"successful":3,"failed":0}` — every member, not `first()`. (2) Missing-name rule byte-identical on both trees: `/no-such-thing/_refresh` 404, `/idx-a,typo/_refresh` 404, `/idx-*/_refresh` 200 total=3, `/nomatch-*/_refresh` 200 total=0. (3) Fail-before: reverting ONLY the es_compat.rs hunk and keeping both tests gives exactly `maintenance_endpoints_accept_an_alias_and_reach_every_member ... FAILED` (404 != 200) and `an_unknown_name_still_fails_the_whole_request ... ok`.

Regression hunt, all negative. Double-operation: the alias arm dedupes into the same `names` vec the wildcard arm uses, so `/tri,idx-a/`, `/idx-a,tri/`, `/idx-*,tri/` and `/tri,tri/` all report total=3, never 4 or 6. Scope creep: `_close`/`_open`/`_search/template`/`_admin/segments/fsck` do also change, but all four were 404 on the base, so nothing that worked stopped working; `/tri/_close` now closes all three members and `/tri/_open` reopens all three (verified by follow-up `_search` returning `index_closed_exception` then 200). `POST /*/_close` is still refused 400 on both trees. `_async_search` is untouched: its single-literal-name path never reaches this resolver.

The dot-prefixed/system-index hypothesis is NOT a hole. `POST /.sysidx/_refresh` -> 200 and `POST /.sysidx/_close` -> 200 ALREADY on the merge base, because the pre-existing concrete-name arm never excluded dot names — only the wildcard arm does, and it still does (`/*/_refresh` returns total=4, not 5, on both trees). So the alias arm reaches nothing a caller could not already reach by typing the concrete name, which is also what ES does (explicit names and aliases address hidden indices; only wildcard expansion excludes them). I also confirmed authorization is unchanged: with an auth-enabled node, an Unscoped key and a key scoped to `ord-idx` both get identical 403 `security_exception` rows on base and branch for `/bobalias/_refresh`, `/mixed/_refresh`, `/bobalias/_close`, `/mixed/_close`, `/bobalias/_terms_enum` and `/bobalias/_search/template`, where `bobalias`/`mixed` point at `.xerj-memory-bob-edges`. The authz middleware resolves the alias before the handler runs, and `Engine::get_index` re-checks `index_guard::visible` per concrete member, so alias expansion cannot launder access.

Gates: full `xerj-api` suite green under `rustup run 1.97.1` in the ci-test profile (31 test targets + doctests, 0 failed, exit 0); `cargo fmt --all --check` clean; `cargo clippy -p xerj-api --all-targets -- -D warnings` clean. Dev profile: I ran the PR's own test green there, but could not run the whole suite in dev — the box dropped to 729 MB free and I stopped and reclaimed the space rather than fill it. I am reporting that as not-run, not as verified.

Non-blocking notes for the author, none of which meet the bar: (a) the body lists four endpoints but `_close`, `_open`, `_admin/segments/fsck` and `_search/template` also change from 404 to alias expansion — `_close` in particular now closes every member of an alias, which is correct ES behaviour but is not in the body; (b) "a concrete index name still wins over an alias of the same name" is true of the new resolver arm's name selection but not end-to-end — `Engine::get_index` still resolves an alias before a concrete index, so with a colliding name the op runs on the alias target while being labelled with the concrete name. I measured this and it is BYTE-IDENTICAL on the merge base, and the body itself discloses `get_index`'s alias-to-`first()` behaviour as unfixed under #449, so it is pre-existing and not worsened; (c) the guard test's closing assertion is commented "an index of the same name as an alias wins" but uses `idx-a`, which is not an alias name, so it does not construct the collision it describes.

```
mkdir -p /tmp/r-alias && git clone -q --no-hardlinks /home/claude/work/xerj-main /tmp/r-alias/repo
git fetch -q origin fix/alias-selector-maintenance-ops && git checkout -q FETCH_HEAD
git merge-base FETCH_HEAD origin/main  ->  e19f8136287109d8cf6a106aa043238916401f0a
gh pr view 459 --repo xerj-org/xerj --json number,title,body,headRefName,baseRefName,state,mergeable
git show a57604e  (full diff: engine/crates/xerj-api/src/es_compat.rs +30/-8, plus new test file)
grep -n 'resolve_indices_for_op' engine/crates/xerj-api/src/es_compat.rs  ->  9 call sites, all in es_compat.rs
python3 mapping of call-site line numbers to enclosing fns -> refresh_index, close_index, open_index, forcemerge, admin_segments_fsck, clear_cache, search_template, terms_enum, async_search_submit
wrote engine/crates/xerj-api/tests/zz_probe_alias.rs (26-request behaviour table: alias expansion, comma/wildcard overlap, dot-prefixed index, name collision, missing-name rule, close/open round trip)
CARGO_TARGET_DIR=/home/claude/work/xerj-main/target rustup run 1.97.1 cargo test -p xerj-api --profile ci-test --test zz_probe_alias -- --nocapture   (on FETCH_HEAD)
git checkout -q e19f813 && same probe re-run  (merge base) -> diffed against branch output
wrote engine/crates/xerj-api/tests/zz_probe_alias_authz.rs (auth-enabled node, admin key, minted Unscoped key + key scoped to ord-idx, aliases onto .xerj-memory-bob-edges)
rustup run 1.97.1 cargo test -p xerj-api --profile ci-test --test zz_probe_alias_authz -- --nocapture   (run on BOTH e19f813 and a57604e; 403 rows identical)
wrote engine/crates/xerj-api/tests/zz_probe_collide.rs (name that is both a concrete index and an alias for a different index)
rustup run 1.97.1 cargo test -p xerj-api --profile ci-test --test zz_probe_collide -- --nocapture   (run on BOTH trees; output byte-identical)
git show a57604e -- engine/crates/xerj-api/src/es_compat.rs | git apply -R   (fail-before: revert source hunk only, keep both tests)
rustup run 1.97.1 cargo test -p xerj-api --profile ci-test --test alias_is_a_valid_selector_for_maintenance_ops   (reverted: 1 FAILED / 1 ok, exactly as the PR body states)
git checkout -q -- engine/crates/xerj-api/src/es_compat.rs   (restore branch)
rustup run 1.97.1 cargo test -p xerj-api --profile ci-test   (FULL suite on branch: 31 'test result: ok', 0 failed, exit 0)
rustup run 1.97.1 cargo test -p xerj-api --test alias_is_a_valid_selector_for_maintenance_ops   (dev profile: 2 passed)
rustup run 1.97.1 cargo fmt --all -- --check   (exit 0)
rustup run 1.97.1 cargo clippy -p xerj-api --all-targets -- -D warnings   (exit 0)
df -h / monitored throughout; deleted the ~11 GB of build artifacts this review added to /home/claude/work/xerj-main/target (mtime-scoped to this session) to return the box from 729 MB to 6.6 GB free
```
</details>


## My read

Merging. The reviewer went further than the PR did: they diffed **all nine** endpoints routing through `resolve_indices_for_op` on both trees and confirmed every difference is `404 -> 200-with-correct-expansion` — there is no shape in which a request that returned 200 on the base returns a *different* 200 on the branch, which is the only form a regression could take here.

Three things they checked that I had flagged as open questions:

- **The system-index hypothesis is not a hole.** `POST /.sysidx/_refresh` already returned 200 on the merge base, because the pre-existing concrete-name arm never excluded dot names — only the wildcard arm does, and it still does (`/*/_refresh` returns total=4, not 5, on both trees). The alias arm reaches nothing a caller could not already reach by typing the name, which is also what ES does.
- **Authorization is unchanged.** With an auth-enabled node, an Unscoped key and a key scoped to another index both get identical 403 `security_exception` rows on base and branch for aliases pointing at `.xerj-memory-bob-edges`. Alias expansion cannot launder access.
- **No double-operation.** `/tri,idx-a/`, `/idx-a,tri/`, `/idx-*,tri/` and `/tri,tri/` all report total=3, never 4 or 6.

Three non-blocking notes I am carrying forward rather than burying:

1. The PR body names four endpoints; `_close`, `_open`, `_admin/segments/fsck` and `_search/template` also change from 404 to alias expansion. All four were 404 on the base so nothing regressed, but `/tri/_close` now closes every member of an alias and my body did not say so.
2. "A concrete index name still wins over an alias of the same name" is true of the new resolver arm but **not end-to-end** — `Engine::get_index` still resolves an alias before a concrete index. Measured byte-identical on the merge base, and the body already discloses `get_index` as unfixed under #449, so pre-existing and not worsened.
3. My guard test's closing assertion is commented "an index of the same name as an alias wins" but uses `idx-a`, which is not an alias name — so it does not construct the collision it describes. The assertion it does make is still valid; the comment overstates it.

(2) and (3) belong to the #449 follow-up, where `get_index` gets fixed properly.

PR #477, opened 2026-08-18
fix(limits): stepped default memory cap (8/16/32 GiB by RAM) instead of a flat 8 GiB

xerj-org:
## Why

The default process-memory cap was a flat **8 GiB**, and every governor budget
(memtable, hydration cache, RSS watermark) derives from that one base. A flat
number under-serves anything that isn't a small box:

- **Big machines were silently starved.** Our own 119 GiB reference-coding box
  ran with the flat 8 GiB cap — an ~1.6 GiB hydration cache — while 100+ GiB sat
  unused; that under-sizing tracked with a long read tail.
- **Laptops were the opposite failure** before the cap existed at all: with no
  cgroup the base was the whole machine, so a 64 GiB host asked for ~16 GiB of
  memtables and a ~12.8 GiB hydration cache before allocating anything — the
  reported ~20 GiB-resident regression.

## What

Replace the flat default with an **AUTO tier** chosen from the *effective*
(cgroup-aware) memory the process can actually see — `min(cgroup, host RAM)`,
not raw host RAM — in binary-GiB steps:

| effective memory        | process cap |
|-------------------------|-------------|
| `< 64 GiB`              | 8 GiB       |
| `64 GiB ≤ eff < 128 GiB`| 16 GiB      |
| `≥ 128 GiB`             | 32 GiB      |

**The cap only ever LOWERS the base** — it is still `min(machine, cap)`, so it
never invents headroom: a 40 GiB box on the 8 GiB tier yields 8; a 6 GiB box on
the 8 GiB tier yields 6.

**Backward-compat preserved exactly:**
- `max_process_memory_mb = 0` → **uncapped, whole machine** (unchanged meaning).
- `max_process_memory_mb = N` (N > 0) → a **fixed N MiB** ceiling, still min'd
  with the machine (unchanged).
- Omitted / default → **AUTO** (the new tier). Represented by a `u64::MAX`
  sentinel that the governor intercepts before any budget math, so it never
  reaches a byte count or serializes into a user-facing default.

**Friendly first-run line + runtime override.** One non-alarming startup line
states what was detected, the cap chosen and why, and how to change it. A new
`XERJ_MAX_PROCESS_MEMORY_MB=auto|off|<MiB>` env override mirrors the existing
`XERJ_SEGMENT_HYDRATION_CACHE_MB` grammar (env wins over config; a bare `0` is
ambiguous and falls back to config with a warning).

## Live proof

Booted the built binary on a private port with a throwaway data dir on the
119.2 GiB box, no config, no env override:

```
memory: detected 119.2 GiB usable, using a 16 GiB cap (auto tier); change with
limits.max_process_memory_mb or XERJ_MAX_PROCESS_MEMORY_MB (0 = whole machine)
resource governor initialised  memtable_budget_mb=4096  memory_limit_mb=16384
  memory_limit_source=Auto  memory_watermark_mb=15564  segment_hydration_cache_mb=3276
  segment_hydration_cache_source=Auto
```

119.246 GiB is **below** the 128 GiB boundary, so this box correctly lands on
the **16 GiB tier** (`memory_limit_mb=16384`) — **double** the old flat 8 GiB,
with a 4 GiB memtable budget and a ~3.2 GiB hydration cache. A machine at or
above 128 GiB effective would take the 32 GiB tier; the tier steps are pinned by
tests at the exact 64/128 GiB binary boundaries (127 GiB → 16, 128 GiB → 32).

## Tests

- `governor::tests` **19/19** — incl. 4 new: `the_auto_default_tiers_the_cap_and_only_lowers`,
  `tiered_cap_steps_on_binary_gib_boundaries`, `explicit_config_zero_is_uncapped_and_n_forces_a_fixed_cap`,
  `env_override_wins_and_mirrors_the_hydration_grammar`; preserved
  `the_cap_never_raises_a_budget_above_the_machine`.
- `xerj-common` counted-claims **3/3** — the `limits` section stays **14**
  settings and the total stays **116** (a value change adds no serialized leaf).
- `product_experience` **10/10** (`EXPECTED_SETTINGS = 116` holds).
- Scoped release build clean, no warnings. ES-YAML conformance is unchanged
  (no query semantics touched) and runs on CI.

## Files

`engine/crates/xerj-engine/src/governor.rs` (tier resolver + startup line + tests),
`engine/crates/xerj-common/src/config.rs` (`AUTO_PROCESS_MEMORY_MB` sentinel,
default, doc), `engine/xerj.default.toml` (section-header count fixup).


xerj-org:
Reworked to clear the two failing checks, both of which were toolchain drift from sitting 36 commits behind main (not logic):

- **Format + Clippy**: rustfmt 1.97.1 rewraps two match arms and an assert in `governor.rs`; ran `cargo fmt` under the CI toolchain. Then clippy 1.97.1's `doc_overindented_list_items` flagged three doc-comment continuation lines in `config.rs:1474-1477` (indented 12 under a `   * ` marker whose text starts at 5) — re-aligned.
- **ONNX feature compile + tests**: same lint. That job compiles `xerj-common` as a dependency, so the doc-overindent tripped it too.

Merged current main in; no behaviour change. Verified locally under rustc 1.97.1: `fmt --check` clean, `clippy --workspace --all-targets -D warnings` clean, the ONNX-feature clippy (`-p xerj-ai -p xerj-engine --features onnx-experimental`) clean, governor tests 19/19 under both dev and ci-test. Pushed `a45f0ab`; CI re-running.

xerj-org:
## Independent verification — head `a45f0abb`

Three skeptics, each in its own worktree under rustc 1.97.1, each told to REFUTE this PR (tier boundaries, the cap-only-lowers OOM invariant, the `0=uncapped` footgun, env-parse robustness). **All three could not refute: 0 blocking, 0 should-fix, 1 nit.** Merging on this.

Highlights, verified independently: mutation testing confirms the boundary tests are load-bearing (flipping `<`→`<=` at 64 GiB fails 3 tests; flipping 128 GiB fails the boundary test; turning the garbage-env Err branch into a silent `cap_mb:0` fails the env test). The cgroup limit is honored twice (`min(cgroup, host)` then `min(tier, machine)`), so a 128 GiB host under a 4 GiB cgroup yields 4 GiB — no OOM regression. Every dependent budget (memtable, hydration, watermark) derives from the single capped base; nothing reads raw machine RAM. 609 engine tests pass both profiles, clippy/fmt clean, docs_capability_lists 15/15, product_experience 10/10.

The one nit (undocumented `unlimited` env synonym for `off`) is filed as a follow-up, not a blocker.

---

**Verdict (verbatim):**

> Could not refute. The tiering and cap arithmetic are correct and the cap-only-lowers invariant holds on every branch; nothing blocking. (1) Tier boundaries correct and pinned: <64 GiB to 8, [64,128) to 16, >=128 to 32, tested at 63->8/64->16/127->16/128->32, matching the doc table (governor.rs:414-422). (2) Small/tiny boxes (2-6 GiB) pick the 8 GiB tier but cap_memory_limit mins with machine RAM, yielding actual RAM, never above it (governor.rs:396-401; test the_auto_default_tiers_the_cap_and_only_lowers asserts 6 GiB->6). (3) effective = min(machine, tier) holds on all branches Auto/Config/Env x 0/N because cap_memory_limit is `if cap==0 {machine} else machine.min(cap*MiB)`; the only whole-machine path is cap==0, an explicit opt-in. Could not construct any input where effective > min(machine, tier). (4) The intentional 16/32 GiB increase on >=64/>=128 GiB machines is the feature, bounded within the machine with the watermark proportional to the capped base; not an OOM regression. (5) 0=uncapped is an explicit escape hatch with unchanged legacy meaning, not the default and not new. (6) Env parsing robust: negative, overflow-huge, empty, whitespace, garbage all fail parse::<u64> and fall back to config with a warning; huge valid numbers saturate via saturating_mul to the machine; bare 0 warns and falls back. No panic or mis-parse. (7) Coherence verified: memtable (governor.rs:688), hydration (governor.rs:691) and watermark (governor.rs:705) all derive from memory_limit_bytes (the single capped base); no dependent budget reads raw machine RAM. Governor tests 19/19 pass on rustc 1.97.1; xerj-common 91+3 pass. default.toml does not serialize the u64::MAX sentinel. The struct-doc/toml 13->14 change corrects a pre-existing inconsistency (main's top-level doc already said 14; struct has 14 fields on both branches), not a new error.

**Findings (verbatim):**

- `nit` engine/crates/xerj-engine/src/governor.rs:505: The XERJ_MAX_PROCESS_MEMORY_MB env parser accepts an undocumented "unlimited" synonym for "off" (uncapped). The PR doc and config.rs claim the grammar is `auto|off|<MiB>` and that it mirrors the XERJ_SEGMENT_HYDRATION_CACHE_MB grammar, but the hydration resolver (governor.rs:594) accepts only "off", not "unlimited". Cosmetic inconsistency; no wrong memory outcome.

  Demo: governor.rs:505 `Some("off") | Some("unlimited") => ...cap_mb:0` vs governor.rs:594 `Some("off") =>` only. config.rs doc string says grammar is `auto|off|<MiB>`.

---

### Lens 2 (regression-oom) — `refuted: False`

**Verdict (verbatim):**

> Cannot refute — the tiered-memory-cap logic is correct and the tests genuinely prove it. Under my lens (test-proof + env-parse robustness): (a) Mutation testing confirms the tests are load-bearing — flipping the 64-GiB boundary (`<`→`<=`) fails 3 tests, flipping the 128-GiB boundary fails the boundary test, and turning the garbage-env Err branch into a silent `cap_mb:0` fails the env test. 19/19 governor tests pass unmutated under rustc 1.97.1. (b) Env parsing (XERJ_MAX_PROCESS_MEMORY_MB) is panic-free and correct for auto/off/unlimited/<N>/negative/garbage/empty/huge/0: all non-numeric and negative inputs route through `parse::<u64>()` Err → warn + fall back to config; huge values saturate in `saturating_mul` then `min` with the machine (no overflow, only lowers); a bare `0` is treated as ambiguous → warn + fallback. (c) toml only bumps the limits section header 13→14 and correctly does NOT serialize the AUTO u64::MAX sentinel; default = AUTO_PROCESS_MEMORY_MB. (d) docs_capability_lists 15/15, xerj-common counted-claims 3/3, product_experience 10/10 all green under 1.97.1. (e) Coherence holds — memtable, hydration cache, and RSS watermark all derive from the single capped `memory_limit_bytes`, none from raw machine RAM; `cap_memory_limit`'s pre-existing `cap_mb==0` guard makes 0=uncapped work and the `min()` prevents a tier above physical RAM on sub-8-GiB boxes. The 0=uncapped whole-machine path is an explicit, documented opt-in escape hatch (not the default) and is unchanged from before. The only behavioural change vs today's flat 8 GiB is an intentional, RAM-abundant increase to 16/32 GiB on 64+/128+ GiB boxes (the PR's whole purpose), which is min'd with the machine and not an OOM risk. Nothing blocks.

**Findings (verbatim):**

- `nit` engine/crates/xerj-engine/src/governor.rs:412 (tiered_cap_mb): On boxes with >=64 GiB effective memory the new AUTO default raises the effective process cap from the old flat 8 GiB to 16 GiB (64-128 GiB) or 32 GiB (>=128 GiB). This is a deliberate, safe increase, not a defect — flagged only for transparency since the task asked whether any path raises the limit above the old flat 8 GiB.

  Demo: resolve_process_memory_cap(64*GIB, AUTO, None).cap_mb == 16*1024 and cap_memory_limit(64*GIB, 16384) == 16 GiB, vs the old flat cap_memory_limit(64*GIB, 8192) == 8 GiB. The box has 64 GiB RAM and the value is min'd with the machine (25% of RAM), the RSS watermark is 95% of the cap, so it is well within physical memory — the intended fix for 'big machines were silently starved', not an OOM regression.

---

**Verdict (verbatim):**

> Could not demonstrate any blocking or should-fix defect. The PR is safe and coherent. Attack results: (1) OOM/cgroup — REFUTED. effective_memory_bytes = effective_memory_limit_bytes() = min(cgroup, host RAM) (governor.rs:666, 839-851). The AUTO tier is chosen OFF that cgroup-aware value (tiered_cap_mb, governor.rs:402-419), then cap_memory_limit mins it with the machine AGAIN (governor.rs:395-399). So a 128 GiB host under a 4 GiB cgroup yields min(4,8)=4 GiB; the cgroup is honored twice and the process can never target more memory than it can see. The 16/32 GiB tiers only engage when effective memory is genuinely >=64/128 GiB, so raising the default above the old flat-8 never exceeds the machine — no OOM regression on the honored-cgroup path. (2) 0=uncapped hatch — pre-existing semantics (identical to origin/main: '0 disables the cap'), and it is NOT the default (default is the AUTO u64::MAX sentinel), so it is not a regression. (3) Coherence — REFUTED: memtable (governor.rs:687), hydration (:691), and RSS watermark (:703) ALL derive from the single capped base memory_limit_bytes; grep shows no other consumer reads raw machine RAM and no code outside governor.rs reads max_process_memory_mb, so the u64::MAX sentinel never reaches byte math. (4) Env parsing — garbage/negative/huge/empty all trim then parse::<u64>() -> Err -> warn + config fallback; saturating_mul(1024*1024) prevents overflow; no panic path. It mirrors the hydration grammar exactly (both case-sensitive on auto/off, both trim, both treat a bare 0 as ambiguous -> warn+fallback). (5) Boundaries — tests pin 63->8, 64->16, 127->16, 128->32; sub-8-GiB machines are min-lowered (6 GiB -> min(6,8)=6). (6) Build gates all green under rustc 1.97.1 (see commands). The only theoretical residual — that raising the default from 8 to 16/32 GiB widens the blast radius of any PRE-EXISTING cgroup-detection gap on a memory-limited box whose limit the code fails to read — is not demonstrable here (cgroup_memory_limit_bytes is unchanged by this PR) and is inherent to the feature's intent, so it is not reported as a finding. The 13->14 'settings' doc change is a correctness fix: LimitsConfig has 14 pub fields and the registry already listed ('limits', 14) on main; only the stale prose comment said 13.

---

Issue #367, opened 2026-08-14
xc-index.sh cannot index apache/lucene: autoindex exits 1 on both embed modes, wrapper discards a fully-populated corpus, xc.py then refuses it


xerj-org:
Found by an agent (Claude Opus 5) driven by @xerj-org, immediately after #358 added `apache/lucene` to `hub/xerj-search.json`. Reporting it against my own change: the manifest is fine, the indexer is not.

## Summary

`scripts/xc-index.sh` cannot index `apache/lucene`. `xerj autoindex` exits `1` on **both** embedding modes with two different reconciliation errors, the wrapper treats anything other than `0`/`3` as fatal and exits before writing its state file, and `xc.py` then refuses the corpus. **The documents themselves index completely** — the failure is entirely in finalisation and exit-code handling.

## The chain, each step observed

**1. Neural node (`--embed-mode neural`)**

```
$ bash scripts/xc-index.sh lucene-meta
xerj-done ok=false exit=1 reason=aborted wall=70.4s
error: generation cutover requires a resumable embedding execution identity:
       the neural model is not content-addressed; use a fresh autoindex state after changing it
xc-index: autoindex failed with exit 1
```

The remedy the error names does not work. `xc-index.sh lucene-meta --fresh` fails identically (`wall=69.2s`, same message), and no destination index had been created at that point, so there was no prior generation to carry over.

**2. Lexical node (default, no flag)** — different error, same outcome:

```
$ XERJ_URL=http://localhost:9480 bash scripts/xc-index.sh lucene-meta
xerj-done ok=false exit=1 reason=aborted wall=313.2s
error: catalog read-back for ds:docs disagrees with the sealed generation projection
xc-index: autoindex failed with exit 1
```

Same binary, same corpus, same clean `~/.xerj/autoindex/` state. Only the embedder differs. So this is not one bug with one trigger.

**3. The data is fine.** After the aborted lexical run:

```
xc-lucene-meta-lucene-lucene   52,225 docs
java documents across xc-lucene-meta*: 6,032     (6,012 .java files on disk)
```

Every file is there. This matches what I saw indexing the same tree directly earlier: 6,014 indexed against 6,012 on disk, 72,990 symbols extracted, all after an `exit=1`.

**4. The wrapper then discards a working corpus.** `xc-index.sh` maps `0` and `3` to success and does `exit $rc` otherwise, so it never reaches the block that writes `$ROOT/state/<name>.json`. That directory is never created:

```
[...]

PR #486, opened 2026-08-18
fix(autoindex): a non-resumable embedder no longer blocks a first generation (#367)

xerj-org:
Fixes the neural half of #367 — the issue blocking reference coding, this project's main use case, on `apache/lucene`.

## What was broken

`--embed-mode neural` could not index **anything**. Minimal reproduction is one 6 KB file in 0.6 s:

```
xerj --data-dir $D --port 57010 --embed-mode neural --insecure &
xerj autoindex /tmp/one --url http://127.0.0.1:57010 --prefix p --state-dir $S --no-graph
# xerj-done ok=false exit=1 reason=aborted
# error: generation cutover requires a resumable embedding execution identity
```

Nothing about Lucene mattered — the full 6,012-file corpus produced the byte-identical message, just 229 s later, after writing a 303 MB snapshot and **zero** documents. `--embed-mode proxy` failed the same way. The trigger was exactly: a backend reporting `resumable:false`, plus `--no-graph`, plus any semantic field.

## Why

`resumable` answers *"may I reuse vectors a previous run wrote"*. It is `false` by construction for `neural` and `proxy` because their identity hashes the model **name**, not its bytes. Three gates use it to decide something it does not answer. Two of them gate whether a generation may **begin**, and are fixed here:

- `lib.rs:453` — fired on every run, including genesis, where `sync_bootstrap_genesis` builds generation 0 with an empty plan and `execution: None`. Its suggested remedy, `--fresh`, is unsatisfiable: that flag is client-side state, the `false` is a server property.
- `sync.rs:781` — a second copy in a *record validator*, beside genuinely structural checks. I only found it because fixing the first exposed it.

The third, `lib.rs:3210`, is **not** touched here. It guards a genuine resume, so the conjunction is in the right place — but `resumable` sits in it beside four `current == expected` comparisons, so a second run over an unchanged corpus fails with "embedding execution identity changed" when nothing changed. That makes `neural`/`proxy` indexing one-shot. Filed as #487 rather than folded in: the fix there is either a better message or content-addressing the neural identity, and neither belongs in this diff.

The graph path already scopes the *genesis* case correctly (`state.rs:1084`: `if self.resumed && !resumable`), which is why dropping `--no-graph` indexed the same corpus cleanly. `xc-index.sh` passes `--no-graph` unconditionally.

## What is deliberately kept

Carrying a non-resumable identity **across** a generation is still refused. For these backends a model swapped in under the same name yields the same digest, so the drift check cannot see it. My first attempt removed the gate outright; the diagnosis caught it. To be exact about what that would have cost: `state.rs:1084` still refuses a resumed journal, so it would not have been *silent* — it would have failed later, naming the wrong cause. This gate is defence in depth, not the last wall.

Genesis under a non-resumable identity now carries a warning saying what cannot be detected. It is handed back to the caller and delivered through `pr.note()`, not printed: `stderr` belongs to the progress surface, so `--progress none` stays silent, `--progress json` stays one parseable stream, and the text goes through `sanitize` (#241). That last part is not hygiene — the reason string comes from the server, and a bare `eprintln!` let a proxy embed a newline and forge a trailing `xerj-done ok=true exit=0`. Measured after the fix: `NON_JSON_STDERR_LINES = 0` under `--progress json`, warning still delivered.

## Verification

Fail-before, reverting only the two source hunks and keeping the test:

```
called `Result::unwrap()` on an `Err`: generation cutover requires a resumable embedding execution identity
```

End-to-end on one node, neural backend:

```
RUN 1 (genesis)  exit=0  generation=1  + warning that the identity is not content-addressed
RUN 2 (cutover)  exit=1  error: carrying an embedding execution identity across a generation
                          requires a resumable one
```

The regression test pins **both** directions. Writing it caught two ways it could have passed vacuously — a second file triggers dataset evolution before the gate, and a corpus without prose yields no semantic field, so the gate is never reached. Both are now avoided explicitly.

Suites: `-p xerj-autoindex` 679 passed on **both** ci-test and dev profiles, fmt and clippy clean under rustc 1.97.1.

## Not in this PR

- The **lexical** failure on the same corpus (`catalog read-back for ds:docs disagrees…`) is a different defect — diagnosed as `_source` loss in the storage layer. #367 stays open for it.
- The wrapper/`xc.py` symptom in the issue **already landed** in `cfba338`; at HEAD a failed run salvages what was written (measured: 87,514 records across 23 indices, and `xc.py` then answers correctly). Note the issue's paths are stale — the real ones are `tools/xerj-code/scripts/`.
- Content-addressing the neural embedder (so it could report `resumable:true`) is the complementary fix and exists on unmerged branch `fix/issue-367`; it does not touch these gates, so `proxy` would stay broken without this change.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -28,6 +28,8 @@ struct HttpState {
     /// executor now, so the shape has to be proven here too.
     partially_apply_next_data_bulk: bool,
     fail_embedding_identity: bool,
+    /// #367: the `neural` and `proxy` backends report this `false`.
+    embedding_not_resumable: bool,
     stop: bool,
     embedding_identity_sha256: String,
     saw_dataset_mapping_update: bool,
@@ -149,7 +151,7 @@ fn handle_http(mut stream: TcpStream, state: &Arc<Mutex<HttpState>>) {
                     "identity_sha256": locked.embedding_identity_sha256.clone(),
                     "dimensions": 384,
                     "semantic_contract": "semantic_text-derived-vector.v1",
-                    "resumable": true
+                    "resumable": !locked.embedding_not_resumable
                 }, "took_ms": 0, "request_id": "incremental-http-test"}),
             )
         }
@@ -1874,5 +1876,62 @@ fn a_journal_that_fails_its_own_revalidation_names_the_rebuild_route() {
     assert!(
         rendered.contains("Rebuild with a new --state-dir and a new --prefix"),
         "the refusal must name the recovery route, not just the invariant: {rendered}"
+    );
+}
+
+/// #367: a genesis run must not be refused for a non-resumable identity, and a
+/// later cutover must still be.
+///
+/// `resumable` is `false` by construction on the `neural` and `proxy` backends,
+/// because their identity hashes the configured model *name* rather than its
+/// bytes. That makes reuse unsafe — a model swapped in under the same name
+/// yields the same digest, so the drift check cannot see it — but it says
+/// nothing about whether a first generation may be created. Asserting it on
+/// every run made `--embed-mode neural` unable to index anything at all: one
+/// small file was enough, and `apache/lucene` failed after a full local
+/// extraction.
+///
+/// Both halves are pinned here because the fix is a narrowing, and a narrowing
+/// that goes too far would silently permit mixing two vector spaces in one
+/// index.
+#[test]
+fn a_non_resumable_identity_allows_genesis_and_still_refuses_a_cutover() {
+    let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
+    let _replay_guard = sync_executor::REPLAY_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let corpus = tempfile::tempdir().unwrap();
+    let state_dir = tempfile::tempdir().unwrap();
+    fs::write(
+        corpus.path().join("report.txt"),
+        "Initial quarterly report discusses durable subscription revenue and operating income.",
+    )
+    .unwrap();
+    let endpoint = HttpEndpoint::start();
+    endpoint.state.lock().unwrap().embedding_not_resumable = true;
+
+    // Genesis: nothing to carry forward, so it proceeds.
+    let config = cfg(corpus.path(), state_dir.path(), &endpoint.url, true);
+    assert_eq!(
+        run_index(config).unwrap(),
+        0,
+        "a genesis generation has no prior vectors to mix, so a non-resumable \
+         identity must not refuse it"
+    );
+    assert_eq!(journal_events(state_dir.path(), "sync_commit"), 1);
+
+    // Cutover: now there IS a committed generation, so it must refuse.
+    fs::write(
+        corpus.path().join("report.txt"),
+        "Revised quarterly report discusses durable subscription revenue, operating income and churn.",
+    )
+    .unwrap();
+    let config = cfg(corpus.path(), state_dir.path(), &endpoint.url, true);
+    let err = run_index(config)
+        .expect_err("carrying a non-resumable identity across a generation must still be refused");
+    let msg = format!("{err:#}");
+    assert!(
+        msg.contains("across a generation requires a resumable"),
+        "the refusal must name what it is refusing, got: {msg}"
     );
 }
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -414,7 +414,7 @@ fn begin_non_graph_generation(
     root_identity: &str,
     inventory: &content::Inventory,
     plan: Plan,
-) -> Result<()> {
+) -> Result<Option<String>> {
     anyhow::ensure!(
         cfg.no_graph,
         "non-graph generation cutover requires --no-graph"
@@ -446,18 +446,54 @@ fn begin_non_graph_generation(
         .datasets
         .iter()
         .any(|dataset| dataset.semantic_field.is_some());
+    let mut non_resumable_note: Option<String> = None;
     let identity = if semantic {
         let identity = es
             .embedding_execution_identity()
             .context("generation cutover could not pin the server embedding execution identity")?;
-        anyhow::ensure!(
-            identity.resumable,
-            "generation cutover requires a resumable embedding execution identity: {}",
-            identity
-                .non_resumable_reason
-                .as_deref()
-                .unwrap_or("the server did not provide an immutable identity")
-        );
+        // Scoped to a cutover that actually carries vectors forward (issue #367).
+        //
+        // `resumable` answers "may I reuse vectors a PREVIOUS run wrote". On a
+        // genesis run there are none: `sync_bootstrap_genesis` builds generation
+        // 0 with an empty plan and `execution: None`, so there is no prior vector
+        // space to mix. Asserting it unconditionally refused every `--no-graph`
+        // semantic run on the `neural` and `proxy` backends, which report
+        // `resumable:false` by construction because their identity hashes the
+        // configured model NAME rather than its bytes. One 6 KB file was enough
+        // to trigger it; `apache/lucene`, the flagship reference-coding corpus,
+        // could not be indexed at all. The error even named `--fresh` as the
+        // remedy, which cannot help: `cfg.fresh` is client-side state, while the
+        // `false` is a property of the server's backend.
+        //
+        // The graph path already scopes it this way — `state.rs:1084` bails only
+        // `if self.resumed && !resumable` — so this brings the two into line.
+        //
+        // Carrying a non-resumable identity ACROSS a generation is still refused
+        // below, because for these backends a model swapped in under the same
+        // name yields the same digest, so the drift check cannot see it.
+        if base.generation > 0 || base.execution.is_some() {
+            anyhow::ensure!(
+                identity.resumable,
+                "carrying an embedding execution identity across a generation requires a \
+                 resumable one: {}",
+                identity
+                    .non_resumable_reason
+                    .as_deref()
+                    .unwrap_or("the server did not provide an immutable identity")
+            );
+        } else if !identity.resumable {
+            // Handed back rather than printed: stderr belongs to the progress
+            // surface, so `--progress none` stays silent and `--progress json`
+            // stays one parseable stream (#241).
+            non_resumable_note = Some(format!(
+                "autoindex: this embedding identity is not content-addressed, so a model \
+                 swapped in under the same name would not be detected on a later run: {}",
+                identity
+                    .non_resumable_reason
+                    .as_deref()
+                    .unwrap_or("the server did not provide an immutable identity")
+            ));
+        }
         journal.pin_embedding_identity(
             &identity.identity_sha256,
             identity.resumable,
@@ -577,7 +613,8 @@ fn begin_non_graph_generation(
             state_dir.display()
         )
     })?;
-    journal.sync_begin(&pending)
+    journal.sync_begin(&pending)?;
+    Ok(non_resumable_note)
 }
 
 #[derive(Debug, PartialEq, Eq)]
@@ -3196,7 +3233,8 @@ pub fn run_index_report(cfg: IndexCfg) -> Result<(i32, Option<Value>)> {
             &root_str,
             &inventory,
             plan,
-        )?;
+        )?
+        .inspect(|note| pr.note(note));
         let mut backend = sync_executor::EsSyncBackend::new(&es, &state_dir, cfg.bulk_mb << 20);
         sync_executor::replay_pending_operations(&state_dir, &mut journal, &mut backend)?;
         let summary = finish_generated_run(&es, &mut journal, &cfg)?;
@@ -3803,7 +3841,8 @@ pub fn run_index_report(cfg: IndexCfg) -> Result<(i32, Option<Value>)> {
             &root_str,
             &inventory,
             plan,
-        )?;
+        )?
+        .inspect(|note| pr.note(note));
         let mut backend = sync_executor::EsSyncBackend::new(&es, &state_dir, cfg.bulk_mb << 20);
         sync_executor::replay_pending_operations(&state_dir, &mut journal, &mut backend)?;
         let summary = finish_generated_run(&es, &mut journal, &cfg)?;
--- a/engine/crates/xerj-autoindex/src/sync.rs
+++ b/engine/crates/xerj-autoindex/src/sync.rs
@@ -778,10 +778,16 @@ fn validate_manifest(manifest: &GenerationManifest, legacy: bool) -> Result<()> 
             execution.embedding_dimension != Some(0),
             "embedding execution dimension must be positive when the server pins one"
         );
-        anyhow::ensure!(
-            execution.embedding_resumable,
-            "a durable generation requires a resumable embedding execution identity"
-        );
+        // Not asserted here (issue #367). Its neighbours are structural — a
+        // 64-char lowercase hex digest, a non-zero dimension — and say the record
+        // is well formed. `embedding_resumable` says something else: whether
+        // vectors could survive the model being swapped under the same name.
+        // That is a property of REUSE, decided where a cutover actually carries
+        // vectors forward (`begin_non_graph_generation`), not of whether this
+        // record is valid. The `neural` and `proxy` backends report it `false` by
+        // construction, so asserting it here refused every durable generation
+        // those backends could produce. The flag is still recorded on the
+        // execution and travels with it.
         match &execution.source_policy {
             SourceExecutionPolicy::DurableSnapshot {
                 reference,


xerj-org:
## Independent verification — head `4994c089`

Three skeptics, each in its own git worktree, each told to REFUTE this PR, each on a distinct lens, all under the CI toolchain (rustc 1.97.1). Verdicts recorded verbatim below, before any action.

**Result: 2 of 2 returned lenses refuted. Majority refuted — this does not merge. Fixing instead.** The third lens (regression-risk) is still running; its verdict will be appended when it lands. It cannot change the outcome, only add to it.

---

### Lens: correctness — `refuted: True`

**Verdict (verbatim):**

> HEAD verified as 4994c089 (detached; the branch name is checked out by another worktree). The PR's own suite is green on the CI toolchain: `test result: ok. 679 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.93s` (plus 8/1/1/4/0 in the integration binaries), clippy clean under rustc 1.97.1. Three of the four attack lines came back clean and I am recording them as refuted, not as findings: the predicate `base.generation > 0 || base.execution.is_some()` is sound, because the only constructors of a generation-0 committed manifest are `CommittedManifest::genesis()`/`bootstrap_legacy` (sync.rs:167,211) and a non-empty legacy plan is refused outright, so gen-0-with-groups is unconstructible; both `validate_manifest` callers (sync.rs:319, sync.rs:347, the latter reached from sync_executor.rs:739) sit on the same durable path and the resume-time protection at lib.rs:2790 is still live (I triggered it); and the new note reaches `proxy` identically since nothing branches on backend name. The lost-state-dir mixing scenario reproduces identically with a *resumable* backend on this branch, so it is pre-existing and I am not counting it. What does block is that the fix does not do what it claims past the first invocation: three lines after the narrowed gate, `journal.pin_embedding_identity` (lib.rs:497) hits `if self.resumed && !resumable` (state.rs:1084), and `resumed` means "this state dir has a prior run record" — not "a previous generation wrote vectors" — so a `--dry-run` first, or any run that dies before `sync_begin`, makes the very genesis this PR enables fail again; and a genesis that dies *after* `sync_begin` — newly reachable, because on main the gate preceded `sync_begin` — wedges the state dir with no in-place recovery at all. The PR body's assertion that state.rs:1084 "already scopes the genesis case correctly", repeated as fact in the new code comment at lib.rs:468-470, is therefore wrong.

**Findings (verbatim):**

#### `blocking` — engine/crates/xerj-autoindex/src/lib.rs:497  (introduced by this PR: `True`)

**Claim:** The narrowed gate is immediately followed by an un-narrowed use of the same flag: `journal.pin_embedding_identity(..., identity.resumable, ...)` at lib.rs:497 lands in state.rs:1084 `if self.resumed && (!resumable || ...)`. `resumed` is set merely by seeing a prior `run` record in journal.ndjson (state.rs:859), i.e. "not the first invocation against this state dir" — NOT "a previous generation wrote vectors". So genesis on a neural/proxy node still fails whenever the state dir has been touched before, e.g. by the documented `--dry-run` look-before-you-index flow, or by a first attempt that died at the identity probe. The PR body's claim that state.rs:1084 "already scopes the genesis case correctly" and the new comment at lib.rs:468-470 ("this brings the two into line") are both factually wrong: the two predicates are not the same. The PR's own regression test passes only because it uses a virgin state dir whose first invocation is also the successful one.

**Demonstration:**

```
Probe `adv_a_second_invocation_is_still_refused_at_genesis` appended to incremental_reconcile_http_tests.rs, run with `rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test --lib adv_a_second -- --nocapture`:
ADV4/RUN1 (--dry-run): Ok(0)
ADV4/journal exists after dry run = true
ADV4/RUN2 (real genesis): ERR server embedding backend cannot safely resume semantic indexing: embedding identity is not immutable. Restore the exact original embedding identity, or rebuild all vectors with --fresh and a new --prefix. Before reusing the old prefix, delete and recreate its prior autoindex indices
ADV4/RUN3 (identity probe down): true
ADV4/RUN4 (retry after outage): ERR server embedding backend cannot safely resume semantic indexing: ...
```

#### `blocking` — engine/crates/xerj-autoindex/src/lib.rs:474  (introduced by this PR: `True`)

**Claim:** Permitting genesis makes a previously unreachable dead-end reachable: a genesis run that fails after `journal.sync_begin` (lib.rs:616) leaves a pending sync that can never be resumed on a non-resumable backend (lib.rs:2790 `ensure!(current.resumable, ...)`), and `--fresh` is explicitly refused while a pending sync exists (lib.rs:2984-2999). The state dir is unrecoverable in place; the only escape is abandoning it for a new --state-dir AND a new --prefix, leaving the partially written destination behind. On main this state was unconstructible for neural/proxy because the gate at old lib.rs:453 fired before sync_begin at old lib.rs:580. For the flagship use case the PR cites — 6,012 files, 229 s against apache/lucene — a transient bulk failure is the expected case, not the exotic one, and the PR body does not mention this at all.

**Demonstration:**

```
Probe `adv_interrupted_genesis_wedges_the_state_dir_on_a_non_resumable_backend` (embedding_not_resumable + fail_next_data_bulk), same cargo invocation:
ADV1/RUN1 (genesis, bulk fails): prepared bulk contained 1 rejected items: {"type":"injected_failure","reason":"leave durable generation pending"}
ADV1/journal sync_begin=1 sync_commit=0
ADV1/RUN2 (plain retry): pending semantic generation cannot resume because the current embedding backend is not resumable: ...; restore the original backend, or rebuild with a new --state-dir and a new --prefix
ADV1/RUN3 (--fresh): `--fresh` cannot discard an uncommitted pending corpus generation under the same destination ...
Ordering on main confirmed via `git show origin/main:engine/crates/xerj-autoindex/src/lib.rs` — ensure! at :453, journal.sync_begin at :580.
```

#### `should-fix` — engine/crates/xerj-autoindex/src/lib.rs:3237  (introduced by this PR: `True`)

**Claim:** The genesis warning — the PR's stated compensation for permitting a non-content-addressed identity — has zero test coverage, and nothing in the type system protects it. The new regression test builds its config with `progress: ProgressMode::None` (incremental_reconcile_http_tests.rs:439), which maps to `Surface::Silent` (progress.rs:207) where `note` returns immediately (progress.rs:538), so the test cannot observe the warning even in principle. Deleting the delivery at both call sites leaves the whole suite green, and because the discarded value is an `Option` there is no `unused_must_use` signal either — a future refactor drops the warning silently.

**Demonstration:**

```
Replaced `)?\n        .inspect(|note| pr.note(note));` with `)?;` at both lib.rs:3237 and lib.rs:3844 (2 replacements), then `rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test --lib -- --skip adv_` → `test result: ok. 679 passed; 0 failed; 0 ignored; 0 measured; 3 filtered out; finished in 8.37s`, with no compiler warning. lib.rs restored afterwards via `git checkout -- engine/crates/xerj-autoindex/src/lib.rs`.
```

#### `nit` — engine/crates/xerj-autoindex/src/lib.rs:481  (introduced by this PR: `True`)

**Claim:** The warning is delivered only through the progress surface, so `--quiet` / `--progress none` drops it entirely, and it is not carried in the JSON summary that `run_index_report` returns. A machine consumer running `xerj autoindex --json` therefore has no way to learn that the index it just built is pinned to a non-content-addressed embedding identity unless it also parses `--progress json` stderr. The PR frames this as deliberate; it is defensible for stderr hygiene, but it means the only record of an undetectable-model-swap risk can be discarded by a flag.

**Demonstration:**

```
progress.rs:538 `if self.surface == Surface::Silent { return; }`; progress.rs:207 `ProgressMode::None => Surface::Silent`; the summary `Value` built by `finish_generated_run` (lib.rs:2815) never sees `non_resumable_note`.
```

<details><summary>commands run</summary>

```
git fetch origin fix/367-neural-fresh-generation; git checkout --detach FETCH_HEAD; git rev-parse --short HEAD  → 4994c08
gh api repos/xerj-org/xerj/pulls/486 --jq '.title, .body'
git diff origin/main...HEAD (and --stat)
git show origin/main:engine/crates/xerj-autoindex/src/lib.rs | sed -n '440,470p' ; | grep -n 'journal.sync_begin'
grep -rn 'validate_manifest|begin_non_graph_generation|sync_bootstrap_genesis|pin_embedding_identity|validate_against|resumed' engine/crates/xerj-autoindex/src/*.rs
rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test   → lib: test result: ok. 679 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.93s (integration bins 8/1/1/4 passed, 0 doc-tests)
rustup run 1.97.1 cargo clippy -p xerj-autoindex --profile ci-test --all-targets  → no warnings, no errors
rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test --lib adv_ -- --nocapture --test-threads=1   (4 adversarial probes appended to incremental_reconcile_http_tests.rs; left in the worktree)
rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test --lib -- --skip adv_   (with both `.inspect(|note| pr.note(note))` deleted → 679 passed)
```
</details>

---

### Lens: does-the-evidence-actually-hold — `refuted: True`

**Verdict (verbatim):**

> I checked out 4994c089 (matches the mandated head; origin/main 79becd2 is its exact parent) and tried to break every claim under rustc 1.97.1. The PR's own evidence holds up better than I expected. The new test is NOT vacuous: reverting only the lib.rs and sync.rs hunks to main makes it fail with exactly the claimed message ("called `Result::unwrap()` on an `Err` value: generation cutover requires a resumable embedding execution identity"); the corpus really does produce a semantic field (probe dumped `dataset slug="docs" semantic_field="body"`, one GET /v1/embedding/identity, and a journal `embedding_identity` event with `resumable:false`), so the gate is genuinely evaluated; and swapping the expected string for an absurd one fails, with the real error being precisely the new gate's text. I also reproduced `NON_JSON_STDERR_LINES = 0` under `--progress json` end to end with the warning delivered as a `{"event":"note",…}` line, confirmed lib.rs:3210 is byte-identical to main, confirmed 679 lib tests pass on both dev and ci-test profiles, and confirmed fmt/clippy are clean. What refutes the PR is what it did not test: its census of "three gates" is wrong, and the fourth one (lib.rs:2790, `pin_pending_embedding_identity`) is unconditional and becomes reachable only because of this change — an interrupted neural genesis now leaves `sync_begin=1, sync_commit=0` and a state dir that refuses both retry ("restore the original backend", unsatisfiable) and `--fresh`, where on main `sync_begin` was 0 and nothing durable was ever written. Separately, "genesis" here means "no prior journal", not "no prior vectors": a new `--state-dir` over the same `--prefix` now admits a second, different non-resumable identity into a destination holding the first one's documents, which is the exact mixing the new test's doc comment claims to prevent. Those two are new failure modes on the path the PR opens, so it should not merge as-is without either narrowing on the destination rather than the journal, or fixing/disclosing the fourth gate.

**Findings (verbatim):**

#### `blocking` — engine/crates/xerj-autoindex/src/lib.rs:2790  (introduced by this PR: `True`)

**Claim:** There is a FOURTH unconditional `resumable` gate that the PR body's census misses ("Three gates ask `resumable`…"), and this PR is what first makes it reachable. Before this PR a neural/proxy `--no-graph` semantic run died at lib.rs:453 before `journal.sync_begin`, so a pending generation could never exist for those backends. After this PR a genesis run reaches `sync_begin`; if that run is then interrupted (Ctrl-C, transient bulk failure — realistic on the 229 s / 6,012-file apache/lucene run the PR itself cites), the retry hits `pin_pending_embedding_identity` at lib.rs:2790, which asserts `current.resumable` unconditionally. The remedy it names ("restore the original backend") is unsatisfiable — the backend IS the original; `resumable:false` is constitutional for neural/proxy — and `--fresh` is explicitly refused on that path. The state dir cannot be advanced and the generation can never be completed.

**Demonstration:**

```
Probe test on PR head (fixture `embedding_not_resumable=true`, `fail_next_data_bulk=true`, then two retries):
PROBE6 interrupted genesis = Some("prepared bulk contained 1 rejected items: {\"type\":\"injected_failure\"…}")
PROBE6 sync_begin=1 sync_commit=0
PROBE6 retry REFUSED: pending semantic generation cannot resume because the current embedding backend is not resumable: the server did not provide a stable execution identity; restore the original backend, or rebuild with a new --state-dir and a new --prefix
PROBE6 --fresh REFUSED: `--fresh` cannot discard an uncommitted pending corpus generation under the same destination…
Same probe with ONLY the two source hunks reverted to origin/main (79becd2):
PROBE6 interrupted genesis = Some("generation cutover requires a resumable embedding execution identity…")
PROBE6 sync_begin=0 sync_commit=0
→ sync_begin goes 0 → 1: the bricked state is newly reachable because of this PR.
```

#### `should-fix` — engine/crates/xerj-autoindex/src/lib.rs:474  (introduced by this PR: `True`)

**Claim:** The narrowing `if base.generation > 0 || base.execution.is_some()` equates "genesis" with "no prior JOURNAL", not "no prior VECTORS". A fresh `--state-dir` over the SAME `--prefix` is genesis by this test, so a second, different, non-resumable identity is now admitted into a destination that already holds the first identity's vectors — exactly the outcome the new test's own doc comment says the narrowing must not permit ("a narrowing that goes too far would silently permit mixing two vector spaces in one index"). Only a warning is emitted. This contradicts the PR body's "The fail-closed property is kept, deliberately."

**Demonstration:**

```
Probe test on PR head: run 1 with state-dir A, identity_sha256 = "a"*64, resumable=false → exit 0. Then flip the server to identity_sha256 = "b"*64 (still non-resumable), add a file, and run with a brand-new state-dir B and the SAME prefix:
PROBE5 docs after run1 = ["report.txt"]
PROBE5 second identity ACCEPTED into same prefix, exit=0
PROBE5 docs after run2 = ["report.txt", "second.txt"]
With the two source hunks reverted to origin/main the same scenario is refused ("generation cutover requires a resumable embedding execution identity").
```

#### `should-fix` — engine/crates/xerj-autoindex/src/lib.rs:2790  (introduced by this PR: `False`)

**Claim:** PR body claim "Three gates use it to decide something it does not answer" is false as a census. `pin_pending_embedding_identity` contains a fourth `anyhow::ensure!(current.resumable, …)` (origin/main lib.rs:2753, PR head lib.rs:2790) that is unconditional and, like the two the PR fixes, refuses on a property no user action can change. It pre-dates the PR, but the PR's completeness claim and its "two fixed here, the third is #487" accounting are wrong, and the missed one sits on the path this PR newly opens.

**Demonstration:**

```
`grep -n "current.resumable" engine/crates/xerj-autoindex/src/lib.rs` → 2790 (unconditional ensure), 2811 (equality check), 3211 (the disclosed third gate), plus 476 (the patched one). `git show origin/main:engine/crates/xerj-autoindex/src/lib.rs | grep -n current.resumable` → 2753, 2767, 3174, so 2790/2753 is pre-existing and unlisted. Live effect shown by PROBE6 above.
```

#### `nit` — engine/crates/xerj-autoindex/src/lib.rs:468  (introduced by this PR: `True`)

**Claim:** Both the in-code comment and the PR body/commit message misquote the graph-path gate: they say "`state.rs:1084` bails only `if self.resumed && !resumable`". The actual condition is `if self.resumed && (!resumable || self.embedding_identity_resumable == Some(false))`. The genesis argument built on it still holds (the `self.resumed` conjunct is what carries it), but the quoted code is not the code.

**Demonstration:**

```
engine/crates/xerj-autoindex/src/state.rs:1084 reads: `if self.resumed && (!resumable || self.embedding_identity_resumable == Some(false)) {`. Compare lib.rs:468-469 and the PR body's "`state.rs:1084`: `if self.resumed && !resumable`".
```

<details><summary>commands run</summary>

```
git fetch origin fix/367-neural-fresh-generation; git checkout --detach 4994c089cd653e2cc65f1daa86aed9de99fbc8a4  (HEAD = 4994c08, matches the mandated head)
gh api repos/xerj-org/xerj/pulls/486 --jq '.title, .state, .head.sha, .body'
git log -1 --format=%B 4994c089; git show 4994c089 --stat and full diff of lib.rs / sync.rs / incremental_reconcile_http_tests.rs
git rev-parse origin/main; git rev-parse 4994c089^   (both 79becd2 — main is the exact parent)
rustup run 1.97.1 cargo test -p xerj-autoindex --lib a_non_resumable_identity_allows_genesis_and_still_refuses_a_cutover  (PASS on PR head)
git checkout 79becd2 -- engine/crates/xerj-autoindex/src/lib.rs engine/crates/xerj-autoindex/src/sync.rs; rustup run 1.97.1 cargo test -p xerj-autoindex --lib a_non_resumable_identity_...  (FAILS — fail-before reproduced)
Appended 7 temporary probe tests to incremental_reconcile_http_tests.rs and ran them under rustup run 1.97.1 cargo test -p xerj-autoindex --lib -- --nocapture --test-threads=1 (probes 1b, 1c, 3, 4, 5, 6, 7, 8); all probe code reverted afterwards (git status clean at 4994c08)
git show origin/main:engine/crates/xerj-autoindex/src/lib.rs | sed -n '449,462p'  and  '3168,3190p'  (verify lib.rs:453 cite and that the third gate is byte-identical to main)
git show origin/main:engine/crates/xerj-autoindex/src/sync.rs | sed -n '774,788p'  (verify sync.rs:781 cite)
grep -n resumable engine/crates/xerj-autoindex/src/lib.rs engine/crates/xerj-autoindex/src/state.rs  (gate census)
rustup run 1.97.1 cargo test -p xerj-autoindex  → lib 679 passed, 0 failed (dev profile)
rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test --lib  → 679 passed, 0 failed
rustup run 1.97.1 cargo fmt --check -p xerj-autoindex  → clean; rustup run 1.97.1 cargo clippy -p xerj-autoindex --all-targets  → clean
gh api repos/xerj-org/xerj/issues/487 --jq '.number,.title,.state'  (exists, open, matches description); git log --oneline -1 cfba338; grep -n no-graph tools/xerj-code/scripts/xc-index.sh
```
</details>

---

xerj-org:
Closing this. Two independent lenses refuted it, and having checked their demonstrations against the source myself, they are right — this PR trades one failure for a worse one.

## What it actually does

On `main`, a `neural`/`proxy` semantic run with `--no-graph` dies at `lib.rs:453` **before** `journal.sync_begin`. Nothing is written; the state dir stays clean. With this PR the run gets past `sync_begin`. If it is then interrupted — Ctrl-C, a transient bulk failure, the machine going away — the retry hits a **fourth** unconditional `resumable` gate at `lib.rs:2790`:

```rust
anyhow::ensure!(
    current.resumable,
    "pending semantic generation cannot resume because the current embedding backend is not \
     resumable: {}; restore the original backend, or rebuild with a new --state-dir and a new \
     --prefix",
```

and `--fresh` is refused while a pending sync exists. The state dir cannot be advanced in place. The remedy the message names is unsatisfiable: the backend *is* the original one — `resumable:false` is constitutional for these backends, not a drift.

Demonstrated independently by both lenses; with only the two source hunks reverted to `origin/main`, the same probe shows `sync_begin=0`, and on this branch `sync_begin=1`. The bricked state is newly reachable **because of this PR**.

For the case this PR itself cites — 6,012 files, 229 s against `apache/lucene` — an interrupted run is the expected case, not the exotic one. "Cannot start" is a bad bug. "Starts, then bricks the state dir partway through a four-minute run" is a worse one.

## Two claims in my own PR body were false

1. **The census.** I wrote "Three gates use it to decide something it does not answer." There are four. `pin_pending_embedding_identity` (`lib.rs:2790`) was missed, and it is precisely the one this change makes reachable.

2. **The compensation.** I wrote that `state.rs:1084` "already scopes the genesis case correctly", and repeated it as fact in a new code comment. It does not. The predicate is `self.resumed && (!resumable || ...)`, and `resumed` is set merely by finding a prior `run` record in `journal.ndjson` (`state.rs:859`) — that means *"not the first invocation against this state dir"*, not *"a previous generation wrote vectors"*. So a `--dry-run` first, or any first attempt that died early, and genesis fails again anyway. The regression test passes only because its state dir's first invocation is also its successful one.

A third finding, which I am recording though it did not on its own decide this: a fresh `--state-dir` against the **same** `--prefix` counts as genesis under `base.generation > 0 || base.execution.is_some()`, so a second, different, non-resumable identity is now admitted into a destination that already holds the first one's vectors — with only a warning. That is the outcome the new test's own doc comment says the narrowing must not permit.

## Why narrowing was the wrong shape of fix

Four sites, and an interaction between two of them that neither site shows on its own. That is what whack-a-mole looks like. The gates are not the defect; `resumable` being dishonest is. It is `false` for `neural` because the identity hashes the configured model **name** instead of its bytes — so the flag says "cannot verify" when the truth is "did not try".

Fix that and every one of the four gates becomes correct with no narrowing at all: an unchanged model passes, a swapped model fails, and a resumed generation resumes. Branch `fix/issue-367` (`b4d6a2a`) does exactly this — it hashes the resolved `model.safetensors` / `tokenizer.json` / `config.json`, following `FieldInfo.verifySameVectorOptions` (`lucene/core/src/java/org/apache/lucene/index/FieldInfo.java:401`) and `ReplicaNode.fileIsIdentical` (`lucene/replicator/src/java/org/apache/lucene/replicator/nrt/ReplicaNode.java:855`): identify the artifact from its own bytes, fail only on a *proven* mismatch. It touches no gate. That is what I am taking forward.

`proxy` genuinely cannot be content-addressed — you cannot hash a model you can only talk to — so it stays fail-closed, correctly. What it needs is an honest message rather than an open door; that is #487.

## What survives from this branch

Nothing merges from here, but three things are worth keeping and are recorded so they are not re-derived:

- The fourth gate at `lib.rs:2790` and the interrupted-genesis dead end (`--fresh` refused under a pending sync) — added to #487.
- `Progress::note` rather than `eprintln!` for anything carrying a server-supplied string: `sanitize` turns an embedded newline into `?`, which is what stops a proxy forging a trailing `xerj-done ok=true exit=0` line. That belongs wherever this lands next.
- The HTTP fixture work in `incremental_reconcile_http_tests.rs` (`embedding_not_resumable`, and a prose corpus that actually yields a semantic field so the gate is genuinely reached).

Full verdicts are in the comment above. #367 stays open; #487 carries the gate/message work.

xerj-org:
## Third lens — regression-risk — `refuted: true`

Appending as promised. It landed after I closed this, and it does not change the outcome — it reinforces it. **Final tally: 3 of 3 lenses refuted.**

It reproduced the wedge on a real neural node rather than a fixture: a 1500-file corpus, `SIGKILL` mid-replay, **970 of 1500 docs left live in the destination**, plain re-run refused at `lib.rs:2790`, `--fresh` refused, and the only recovery an undocumented `rm -rf` of the state dir — while the printed guidance steers the user to a new `--prefix` that would orphan those 970 docs.

It also establishes something neither of the other two stated outright: the shipped behaviour would have been a **one-shot indexer**. Run 1 succeeds; run 2 fails permanently in both directions — edited corpus at the new gate, unchanged corpus at `lib.rs:3211` with the false "identity changed" message. For a flagship *incremental* path over a corpus that exists to be re-indexed as it evolves, that does not resolve #367; it moves the failure from run 1 to run 2.

One thing it refuted rather than confirmed, worth recording because it was the fix I made last: the `pr.note()` routing holds up under attack. A reason string carrying `ESC`, `CR`, `LF`, `U+2028`, `U+202E` and a forged `xerj-done ok=true exit=0` was neutralised on both `Plain` and `Json` surfaces; `--progress json` stayed 20/20 parseable with the warning as a proper `{"event":"note"}` record; `--progress none` and `--quiet` each emitted 0 bytes.

---

**Verdict (verbatim):**

> HEAD verified as 4994c089 (detached; the branch name is checked out by another worktree). I attacked the three assigned angles and refuted two of them outright: the genesis warning is safe (Progress::note at progress.rs:537 returns early on Surface::Silent and unconditionally sanitize()s at :541, and sanitize at :1055 maps every control char to '?'), and a live neural node on port 61437 proved --progress json stays 20/20 parseable lines with the warning as a proper {"event":"note"} record, while --progress none and --quiet both emit 0 bytes of stderr; an adversarial reason string containing ESC, CR, LF, U+2028 and a forged "xerj-done ok=true exit=0" was neutralised on both Plain and Json surfaces. Drift is also not weakened: with the model swapped behind the same name, run 2 is refused on both the changed-corpus path (new lib.rs:475 gate) and the unchanged-corpus path (lib.rs:3211), and both test profiles plus clippy --workspace --all-targets -D warnings are green. What blocks it is a state main could never reach: because the gate now sits after create_prepared_snapshot and no longer stops the run before lib.rs:616 journal.sync_begin, a SIGKILL during a neural genesis leaves an uncommitted pending generation that a plain re-run refuses at a FOURTH resumable gate (lib.rs:2790, which the PR body does not acknowledge — it claims there are three) and that --fresh explicitly cannot discard, with 970 half-written docs live in the destination; the only escape is undocumented rm -rf of the state dir, while the printed guidance steers the user to a new --prefix that would orphan them. Combined with the fact that the shipped result is a one-shot indexer — run 2 always fails, and on an unchanged corpus with a demonstrably identical sha256 it fails with the false message "embedding execution identity changed" — this narrows the blocker rather than resolving #367, and should not merge until the pending-generation gate is at minimum filed and given a recovery route.

**Findings (verbatim):**

#### `blocking` — engine/crates/xerj-autoindex/src/lib.rs:616  (introduced by this PR: `True`)

**Claim:** By narrowing the gate at lib.rs:475, a semantic --no-graph genesis on a non-resumable backend now reaches `journal.sync_begin(&pending)` for the first time. On main the unconditional `ensure!(identity.resumable)` bailed before it (confirmed via `git show origin/main:engine/crates/xerj-autoindex/src/lib.rs`, where the ensure! sits immediately after `embedding_execution_identity()` and sync_begin is at the end of the same fn). If that genesis run is interrupted after sync_begin, the journal holds an uncommitted pending generation that the tool cannot clear: a plain re-run is refused by a FOURTH resumable gate at lib.rs:2790 (`pin_pending_embedding_identity`, reached from lib.rs:3056) which the PR body does not mention at all, and `--fresh` explicitly refuses to discard a pending generation. The destination is left partially written. This wedge state was unreachable on main for neural/proxy.

**Demonstration:**

```
Started a real node: `xerj --port 61437 --data-dir ./data --insecure --embed-mode neural`; GET /v1/embedding/identity returns backend=neural resumable=false. Indexed a 1500-file corpus with --no-graph and SIGKILLed mid-replay; journal.ndjson then contains 1 sync_begin and 0 sync_commit. Re-run: `exit=1 / error: pending semantic generation cannot resume because the current embedding backend is not resumable: the neural model is not content-addressed...`. With --fresh: `exit=1 / error: ... --fresh cannot discard an uncommitted pending corpus generation under the same destination ...`. Meanwhile `GET /p7-docs/_count` returned {"count":970} of 1500. Only an undocumented `rm -rf <state-dir>` recovered it (run 9 then exited 0 with 1500 live), which is the opposite of the remedy the error prints.
```

#### `should-fix` — engine/crates/xerj-autoindex/src/lib.rs:3211  (introduced by this PR: `True`)

**Claim:** The #487 gate the PR deliberately leaves untouched is now reachable for the first time, and it fails with a message that is factually false: it says the embedding execution identity CHANGED when the digest is byte-identical across both runs. This is the exact defect the PR's own commit message criticises about the old error ('it would have failed with a message that names the wrong cause'), and it is the message the first re-run of every neural --no-graph corpus will hit. On main no committed generation could exist for these backends, so this text was dead code.

**Demonstration:**

```
Genesis run on state1 exited 0 ('generation 1 committed'). Re-running the identical command against the UNCHANGED corpus: `exit=1 / error: embedding execution identity changed since this autoindex journal was created; refusing to mix vector spaces.` The server's /v1/embedding/identity returned the same identity_sha256 6e51e5ce3e463db6d6d853516c358e4a529308f213e85d2ed005f731d00ce8f3 before and after — nothing changed.
```

#### `should-fix` — engine/crates/xerj-autoindex/src/lib.rs:475  (introduced by this PR: `True`)

**Claim:** The narrowing is scoped to genesis only, so the shipped behaviour of `xerj autoindex --no-graph` against a neural or proxy node is a ONE-SHOT indexer: run 1 succeeds, and every subsequent run fails permanently in both directions (changed corpus -> the new gate at lib.rs:475; unchanged corpus -> lib.rs:3211). Since autoindex is the flagship incremental path and #367's corpus (apache/lucene) is the reference-coding corpus that is re-indexed as it evolves, #367 is not actually resolved — the failure has been moved from run 1 to run 2. Neither the PR body nor #487 states this consequence.

**Demonstration:**

```
Run 2b, same state dir, one edited file: `exit=1 / error: carrying an embedding execution identity across a generation requires a resumable one: the neural model is not content-addressed; use a fresh autoindex state after changing it`. Run 2a, same state dir, no edit: exit=1 (lib.rs:3211). Both against the same node that had just committed generation 1 successfully.
```

#### `nit` — engine/crates/xerj-autoindex/src/lib.rs:2790  (introduced by this PR: `False`)

**Claim:** The PR body's central framing — 'Three gates ask `resumable` to decide something it does not answer. Two of them are fixed here; the third is #487' — is incomplete. There are four: lib.rs:475 (fixed), sync.rs:779 (removed), lib.rs:3211 (#487), and lib.rs:2790 in `pin_pending_embedding_identity`. The fourth is not filed anywhere and is precisely the one this PR makes reachable.

**Demonstration:**

```
`grep -n "identity.resumable\|current.resumable\|!resumable" crates/xerj-autoindex/src/lib.rs` yields 476, 484, 2790, 2804, 3211 (plus 499/588/3864 which are pin/record sites, not gates). lib.rs:2790 is an `anyhow::ensure!(current.resumable, ...)` reached from lib.rs:3056 on the pending-sync resume path, and it is what produced the run-8a refusal above.
```

<details><summary>commands run</summary>

```
git fetch origin fix/367-neural-fresh-generation
git checkout --detach origin/fix/367-neural-fresh-generation  # HEAD = 4994c08 (branch name is held by another worktree)
git show HEAD  # full diff, 3 files +121/-17
git show origin/main:engine/crates/xerj-autoindex/src/lib.rs | sed -n '440,470p'
rustup run 1.97.1 cargo test -p xerj-autoindex  -> test result: ok. 679 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 14.12s / ok. 8 passed / ok. 1 passed / ok. 1 passed / ok. 4 passed / ok. 0 passed (doc-tests)
rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test  -> test result: ok. 679 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 6.86s / ok. 8 passed / ok. 1 passed / ok. 1 passed / ok. 4 passed / ok. 0 passed (doc-tests)
rustup run 1.97.1 cargo clippy --workspace --all-targets -- -D warnings  -> Finished `dev` profile ... in 44.48s (no warnings)
rustup run 1.97.1 cargo build --profile ci-test -p xerj-server --bin xerj
xerj --port 61437 --data-dir ./data --insecure --embed-mode neural   (high random port 61437, never 9200)
curl -s http://127.0.0.1:61437/v1/embedding/identity  -> backend=neural, resumable=false
xerj autoindex ./corpus --no-graph --url http://127.0.0.1:61437 --state-dir ./state1 --prefix p1   -> exit 0, note emitted
xerj autoindex ./corpus --no-graph ... --state-dir ./state1 (unchanged corpus)  -> exit 1, lib.rs:3211 false 'identity changed'
xerj autoindex ./corpus --no-graph ... --state-dir ./state1 (edited corpus)  -> exit 1, new lib.rs:475 gate
xerj autoindex ./corpus --no-graph --fresh ... --state-dir ./state1  -> exit 1, --fresh cannot discard committed generation
xerj autoindex ./corpus --no-graph --progress json ... --state-dir ./state4  -> exit 0; python3 line-by-line json.loads: total=20 nonjson=0; warning present as {"event":"note"}
xerj autoindex ./corpus --no-graph --progress none ... --state-dir ./state5  -> exit 0, stderr 0 bytes
xerj autoindex ./corpus --no-graph --quiet ... --state-dir ./state6  -> exit 0, stderr 0 bytes
1500-file corpus + kill -9 mid-replay -> journal has 1 sync_begin, 0 sync_commit
xerj autoindex ./big --no-graph ... --state-dir ./state7  -> exit 1, lib.rs:2790 'pending semantic generation cannot resume'
xerj autoindex ./big --no-graph --fresh ... --state-dir ./state7  -> exit 1, cannot discard pending generation
curl -s http://127.0.0.1:61437/p7-docs/_count  -> {"count":970} (of 1500)
rm -rf ./state7 && xerj autoindex ./big --no-graph ... --prefix p7  -> exit 0, 1500 live (undocumented manual recovery)
temporary #[test] added to progress.rs feeding ESC/CR/LF/U+2028/U+202E + a forged 'xerj-done ok=true exit=0' through Progress::note on Surface::Plain and Surface::Json -> 1 line each, no control chars, no forged record; test reverted with `git checkout -- engine/crates/xerj-autoindex/src/progress.rs` (tree clean)
```
</details>

PR #494, opened 2026-08-19
field report: building a token benchmark against Lucene

xerj-org:
Authored by Claude Opus 5 (Claude Code), directed by @d0znpp.

Standalone field report, as `.github/AI_CONTRIBUTIONS.md` asks. Companion to #493 (the harness itself), kept as a separate diff per that doc.

**Verified:** the indexing and query experience described is first-hand, and the 64,487 vs 42,068 figure is a real logged run.
**Assumed:** nothing. The report explicitly declines to draw a conclusion from that run, which is the point of it.


xerj-org:
Independent check (field report, so structural rather than a code-refutation workflow): exactly one markdown file under `user-feedback/16-agent-field-reports/`, no code or config in the diff, all three CLA checks pass (field-report exemption), 17/17 checks green, mergeStateStatus CLEAN. The one checkable reference — `docs/research/wordpress-verification-and-xerj-vs-grep.md` — exists on main, so the suggestion is actionable. Merging.

PR #495, opened 2026-08-19
fix(server): first-launch console link prints the bound ES-compat port, not the requested one

### LINKED ISSUE #469: First-launch console setup link prints the configured port, not the bound one
Split out of the #466 review, where two reviewers flagged it as pre-existing but the same class as #465.

The first-launch console banner prints its URL from `cfg.server.es_compat_port` rather than from the address the listener actually bound. With an ephemeral port that is directly wrong:

```
es_compat_port = 0
->  http://127.0.0.1:0/_xerj-console/setup#token=...
```

The node is listening on a real port (e.g. 35881, confirmed with `ss -ltnp`); the link the operator is told to open is not reachable.

#466 fixed the same class of problem for the startup banner by binding every listener first and printing from `local_addr()`. The console link is emitted at step 9, *before* the bind, so it did not get that treatment.

Two consequences beyond the ephemeral-port case:

1. The link is the single action a first-run operator is told to take, and it can be unreachable.
2. It reaches stderr before any bind is attempted, which is what made an assertion in #466's test vacuous — `stderr.contains(port)` was satisfied by this link whether or not the bind failure was ever reported.

Fix: emit the setup link after the listeners are bound, from the bound ES-compat address, the same way the banner now does.

Vinz2168:
Split out of the #466 review (per the issue): the first-launch console setup link was built from `cfg.server.es_compat_port` and printed at the old step 9, before any listener was bound — binding happened afterward at the old step 10a.

With an ephemeral port (`es_compat_port = 0`) that produced:

```
http://127.0.0.1:0/_xerj-console/setup#token=...
```

the single action a first-run operator is told to take, and unreachable. It also meant the link reached stderr whether or not the following bind ever succeeded.

## Fix

#466 fixed the same class of problem for the startup banner by binding every listener first and printing from `local_addr()`. This applies that same treatment to the console link:

- Listener binding is hoisted from the old step 10/10a to a new step 8c, ahead of the console bootstrap (step 9).
- The console bootstrap's `bind_url` is now built from `bound.es.port()` (the address the kernel actually gave the ES-compat listener) instead of `cfg.server.es_compat_port`.
- The banner (old step 11) is unchanged in behavior — it now just consumes a `bound` computed slightly earlier in the function.

## Testing

Added `engine/crates/xerj-server/tests/ephemeral_port_console_link.rs`: boots the real binary with `es_compat_port = 0`, waits for both the banner and the first-launch setup link, and asserts the link's port is non-zero and matches the port the banner reports bound.

- Fails against the pre-fix code with the exact `:0` link from the issue (verified by reverting the source change locally and re-running).
- Passes with the fix.
- Existing `taken_port_fails_startup.rs`, `cleartext_bind_fail_closed.rs`, `grpc_h2c_fail_closed.rs`, `tls_fail_closed.rs`, and `brain_cli.rs` all still pass.

Fixes #469

### DIFF
--- a/engine/crates/xerj-server/src/main.rs
+++ b/engine/crates/xerj-server/src/main.rs
@@ -2202,6 +2202,62 @@ async fn async_main() -> Result<()> {
         None
     };
 
+    // 8c. Bind all three listeners NOW — before the console setup link, before
+    //     the banner, before anything reports readiness, and fatally.
+    //
+    //     This used to happen at step 10/10a below, well after the console
+    //     bootstrap at (old) step 9 had already built and printed its
+    //     first-launch setup link from `cfg.server.es_compat_port` — the
+    //     *requested* port, not the one the kernel handed back. With an
+    //     ephemeral port (`es_compat_port = 0`) that produced
+    //     `http://127.0.0.1:0/_xerj-console/setup#token=…`, a link the single
+    //     action a first-run operator is told to take, and unreachable. It
+    //     also ran ahead of the bind, so it printed to stderr whether or not
+    //     the bind that follows ever succeeds (issue #469; #466 fixed the
+    //     same class for the startup banner, printed from `local_addr()`
+    //     right below).
+    //
+    //     Moving the bind here — ahead of the console bootstrap rather than
+    //     just ahead of the banner — means both the setup link and the banner
+    //     can only ever be built from a port this process actually holds; see
+    //     `bind_listener` for why binding eagerly and fatally is itself
+    //     load-bearing (issue #465).
+    //
+    //     Addresses are composed from the parsed IP, never by formatting
+    //     `"{bind}:{port}"` — that string is unparseable for every IPv6
+    //     literal (`"::1:9200"`), which made `bind_address = "::1"` a node
+    //     that could not start at all. `socket_addr` cannot be `None` here:
+    //     step 3a rejected an unparseable `bind_address` before the data
+    //     directory existed, which is where that failure belongs. The
+    //     context is kept as a backstop so that if a future edit reorders or
+    //     drops 3a, this still names the setting rather than surfacing a bare
+    //     `Option` unwrap.
+    let rest_addr: SocketAddr = cfg
+        .socket_addr(cfg.server.rest_port)
+        .with_context(|| bad_bind_address(&cfg))?;
+    let es_addr: SocketAddr = cfg
+        .socket_addr(cfg.server.es_compat_port)
+        .with_context(|| bad_bind_address(&cfg))?;
+    let grpc_addr: SocketAddr = cfg
+        .socket_addr(cfg.server.grpc_port)
+        .with_context(|| bad_bind_address(&cfg))?;
+
+    let rest_listener = bind_listener(rest_addr, "native REST")?;
+    let es_listener = bind_listener(es_addr, "ES-compat")?;
+    let grpc_listener = bind_listener(grpc_addr, "gRPC")?;
+
+    let bound = BoundAddrs {
+        rest: rest_listener
+            .local_addr()
+            .context("native REST: read the bound address")?,
+        es: es_listener
+            .local_addr()
+            .context("ES-compat: read the bound address")?,
+        grpc: grpc_listener
+            .local_addr()
+            .context("gRPC: read the bound address")?,
+    };
+
     // 9. Xerj Console bootstrap.  Creates `.xerj_*` system indices on first
     //    boot, persists a 32-byte master key under data_dir/.xerj_master_key
     //    (mode 0600), and prints the first-launch magic-link banner to
@@ -2223,7 +2279,10 @@ async fn async_main() -> Result<()> {
             // first-launch console link.
             url_host(&cfg.server.bind_address)
         },
-        cfg.server.es_compat_port,
+        // The bound ES-compat port (8c), not `cfg.server.es_compat_port`: with
+        // an ephemeral port (`= 0`) the config value is never the port the
+        // node ends up listening on (issue #469).
+        bound.es.port(),
     );
     let xerj_console_node_id: String = match (cfg.cluster.enabled, cluster_listen_addr) {
         // The same string the cluster transport registered as this node's id.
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent verification — head `53ff8eea`

Three skeptics, each in its own worktree under rustc 1.97.1, each told to REFUTE this PR. **All three returned CLEAN: 0 blocking findings.** Verdicts verbatim below; merging on this.

The evidence lens confirmed the new test is not vacuous: reverting only the `main.rs` source hunk (keeping the test) makes it FAIL, printing the exact unreachable link from #469 (`http://127.0.0.1:0/_xerj-console/setup#token=...`), and an injected absurd-port assertion fails with the real ephemeral port, proving it reads `local_addr()` rather than a constant. One nit only: my verification brief wrongly mentioned a `metrics/release-downloads.jsonl` line — the PR changes exactly 2 files (`main.rs`, the new test). Non-blocking.

---

### Lens 1 — `refuted: False`

**Verdict (verbatim):**

> CLEAN — the fix is correct and I could not refute it. On main, the first-launch console link was built from cfg.server.es_compat_port (the requested port) at step 9, BEFORE the bind at step 10a; with es_compat_port=0 it printed http://127.0.0.1:0/_xerj-console/setup and reached stderr regardless of whether the bind succeeded. PR head moves the three binds to step 8c (main.rs:2245-2247), derives `bound` from local_addr(), and builds xerj_console_bind_url from bound.es.port() (main.rs:2285) after a successful, fatal bind. (1) Ephemeral case fixed: the new process-level test boots the real binary with es_compat_port=0 and asserts the link port is non-zero and equals the banner's bound ES-compat port — it PASSES (1 passed; 0 failed). (2) Failed-bind case fixed: bind_listener(...)? at 8c returns Err before the console bootstrap at step 9, so the link is no longer printed on a failed bind. (3) Token wiring intact: the token is minted in mint_bootstrap_link (xerj-console-api/src/bootstrap.rs:192) independently of the port, and the console router is merged onto es_router which serves es_listener, so the link's port hosts the setup route and accepts the token; only the port value in bind_url changed. (4) metrics/release-downloads.jsonl is NOT modified in this PR (only 2 files changed) — the task's premise about that line does not match the head, non-blocking. Single bind site confirmed (no double-bind); compiles under rustc 1.97.1.

**Findings (verbatim):**

- `nit` git diff origin/main...HEAD --name-only: The PR head 53ff8eea changes only 2 files (engine/crates/xerj-server/src/main.rs and tests/ephemeral_port_console_link.rs). The task-described added line in metrics/release-downloads.jsonl does not exist in this PR.

  Demo: `git diff origin/main...HEAD --name-only` lists exactly the two source files; `git diff origin/main...HEAD -- metrics/release-downloads.jsonl` produces empty output. No accidental artifact; non-blocking.

<details><summary>commands run</summary>

```
git checkout --detach 53ff8eea && git rev-parse --short HEAD  -> 53ff8ee
git show --stat HEAD
git diff origin/main...HEAD -- engine/crates/xerj-server/src/main.rs
git show origin/main:engine/crates/xerj-server/src/main.rs | grep -n es_compat_port  -> line 2226 uses cfg.server.es_compat_port at step 9 pre-fix
grep -n 'bind_listener|socket_addr(cfg.server' engine/crates/xerj-server/src/main.rs  -> single bind site at 2245-2247
grep -n bootstrap xerj-console-api/src/bootstrap.rs  -> token minted independently of bind_url
cd engine && RUSTFLAGS="-C target-feature=+fullfp16" rustup run 1.97.1 cargo test -p xerj-server --test ephemeral_port_console_link  -> test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 4.00s
```
</details>

---

### Lens 2 — `refuted: False`

**Verdict (verbatim):**

> CLEAN — the PR's claims hold under adversarial verification; nothing blocking. HEAD confirmed at 53ff8eea. The fix moves listener binding to a new step 8c ahead of the console bootstrap and builds the first-launch setup link from bound.es.port() (the kernel-assigned port) instead of cfg.server.es_compat_port. The new test is NOT vacuous and genuinely covers issue #469: (1) reverting ONLY the main.rs source hunk to origin/main while keeping the test makes it FAIL, printing the exact unreachable link from the issue — http://127.0.0.1:0/_xerj-console/setup#token=... — and tripping assert_ne!(link_port, 0); (2) an injected absurd-value probe (assert_eq!(link_port, 61111)) fails with left:33969, proving the test reads the real dynamically-assigned ephemeral port, not a hardcoded constant; (3) the test config sets es_compat_port = 0, so it exercises the ephemeral path that is the issue's repro, and cross-checks the link port against the banner port (already fixed by #466 from local_addr()); (4) end-to-end reproduced: pre-fix binary prints :0, fixed binary prints a real port. There is no vacuous-pass path — if the link never appears the test panics rather than passing. Source change is coherent (single bind, listeners consumed in the serve spawns, no double-bind), and sibling bind-ordering tests (taken_port_fails_startup, cleartext_bind_fail_closed) still pass. Note: the aarch64 build needed RUSTFLAGS='-C target-feature=+fp16' to compile a gemm-f16 dependency; this is an environment/target-arch issue unrelated to the PR.

**Findings (verbatim):**

- `nit` engine/crates/xerj-server/tests/ephemeral_port_console_link.rs:49: two_free_ports() binds two ephemeral ports, reads their numbers, then drops the listeners and reuses the numbers as fixed rest_port/grpc_port for the child — a TOCTOU that can theoretically fail if another process grabs a port in the window.

  Demo: The helper drops the held TcpListeners before the child binds them; under the now-fatal bind semantics a lost race makes the child exit before printing the link, so the test would panic (fail) rather than pass. Impact is limited to a possible flaky failure, never a false pass, and only rest/grpc (not the es_compat_port=0 path under test) are affected.

<details><summary>commands run</summary>

```
git checkout --detach 53ff8eea && git rev-parse --short HEAD  -> 53ff8ee (matches required HEAD)
gh pr view 495 --json title,body,files
git show 53ff8eea -- engine/crates/xerj-server/src/main.rs
git show origin/main:engine/crates/xerj-server/src/main.rs | grep -n es_compat_port  -> line 2226 uses cfg.server.es_compat_port, bind at step 10 (line 2339)
cd engine && RUSTFLAGS='-C target-feature=+fp16' rustup run 1.97.1 cargo test -p xerj-server --test ephemeral_port_console_link  -> 1 passed (PR head)
git checkout origin/main -- crates/xerj-server/src/main.rs (revert ONLY source, keep test) && re-run test  -> FAILED: link 'http://127.0.0.1:0/_xerj-console/setup#token=...' , assert_ne link_port,0 -> left:0 right:0
restore fix; inject probe assert_eq!(link_port, 61111) && re-run  -> FAILED left:33969 right:61111 (real ephemeral port read)
cargo test --test taken_port_fails_startup --test cleartext_bind_fail_closed  -> 9 passed / 3 passed
```
</details>

---

### Lens 3 — `refuted: False`

**Verdict (verbatim):**

> REFUTATION FAILED — clean, evidence-backed verdict. PR #495 is a correct, low-risk fix with no demonstrable regression. HEAD verified at 53ff8eea. The change hoists the already-consolidated three-listener bind block from old step 10/10a up to new step 8c (ahead of the console bootstrap) and sources the first-launch console link from bound.es.port() instead of cfg.server.es_compat_port. Bind order (rest→es→grpc) is unchanged; the banner and the setup link both derive from the same single `bound` (one local_addr() read), so they cannot disagree. The setup link is printed exactly once inside bootstrap::run (first launch only) — no double/never-print; in the fixed-port case bound.es.port()==cfg port so output is identical to main. serve_on/tcp_nodelay/shutdown/join!/final-flush are all outside the diff. Only main.rs and a new test changed; bootstrap.rs untouched. Both test profiles pass (0 failed), clippy -D warnings clean, fmt --check clean, under rustc 1.97.1. The only new behavioral nuance — ports are bound (connections queue in backlog) during the bootstrap init window rather than being refused — is benign and strictly safer (it closes the #465 false-green class earlier); it is not a wrong user/operator outcome. Environment caveat: generic aarch64 host can't compile transitive dep gemm-f16 (fullfp16 inline asm) under repo default rustflags — a pre-existing env issue, not the PR — so builds/tests were run with RUSTFLAGS="-C target-cpu=native"; CI's x86-64-v3 target is unaffected.

<details><summary>commands run</summary>

```
git checkout --detach 53ff8eea && git rev-parse --short HEAD  -> 53ff8ee
git diff --stat origin/main...HEAD  -> only main.rs (+68/-56) and tests/ephemeral_port_console_link.rs (+208)
cd engine && RUSTFLAGS="-C target-cpu=native" rustup run 1.97.1 cargo test -p xerj-server  -> all ok, 0 failed (33/3/9/1/1/5/3/5/2/3/1)
cd engine && RUSTFLAGS="-C target-cpu=native" rustup run 1.97.1 cargo test -p xerj-server --profile ci-test  -> all ok, 0 failed
cd engine && RUSTFLAGS="-C target-cpu=native" rustup run 1.97.1 cargo clippy --workspace --all-targets -- -D warnings  -> Finished, no warnings
cd engine && rustup run 1.97.1 cargo fmt --check  -> clean (no output)
grep console link print site: crates/xerj-console-api/src/bootstrap.rs mint_bootstrap_link (single eprintln site, first-launch only)
```
</details>

---

PR #499, opened 2026-08-19
fix(scroll): carry real _seq_no/_version on the same read as _source (#440)

### LINKED ISSUE #428: scroll continuation pages (_search/scroll) never emit _seq_no/_version, even with seq_no_primary_term:true

### LINKED ISSUE #440: scroll's point-in-time view is torn: _seq_no is read in a second pass, so a hit can carry one moment's _source beside another's sequence number (silent lost update)

Vinz2168:
Fixes #440. Fixes #428 (the original scroll-continuation hardcoding this PR also closes, folded into the same fix). Supersedes #431 (never merged — closing that PR, not merging it).

## The bug (#440)

Two repair attempts on PR #431 (itself built on top of #428's original scroll fix) were refuted by the maintainer's own live reproduction:

1. Resolve `_seq_no`/`_version` at **render time** — a live `idx.lookup_seq_no(&h.id)` call in `es_compat.rs`, after `merged_hits` was already built. Window: the whole scroll lifetime.
2. Resolve them once into the hit at **scroll-open time** instead. Narrower, but a multi-index scroll still spans every subsequent index's search, and — the real problem — the value is still read *after*, not *from*, the same read that produced `_source`.

Both let a write land between the `_source` read and the `_seq_no` read, so a hit can report a live sequence number next to a stale/snapshot body. Since `_seq_no`/`_primary_term` are designed to be fed back as `if_seq_no`/`if_primary_term` for optimistic-concurrency writes, that pairing is a silent lost update — reproduced directly on #440 on the reindex/migration/CDC path scroll exists to serve.

**#431 was never merged.** `main` has none of its `ScrollContext` snapshot work — scroll continuation pages hardcode `seq_no: Some(0), version: Some(1), primary_term: Some(1)` on *every* hit unconditionally, ignoring the request's `seq_no_primary_term`/`version` flags entirely, in both `scroll_page_response` and `search_with_scroll`'s own first-page builder. Worse, `scroll_page_response`'s hand-built JSON never inserted `_seq_no`/`_version`/`_primary_term` into the wire response at all, regardless of what `EsHit` carried. This PR fixes the original #428 gap and #440 in one pass, on top of `main` directly.

## The fix

Per #440's own diagnosis: the value has to come from the same read that produced `_source`, not a second pass. Lucene gets this structurally (a searcher pins an `IndexReader`; `_seq_no` doc-values are immutable per segment, read alongside the body). xerj has no such structure, so this makes the same guarantee explicit:

- `xerj_query::executor::Hit` gains `seq_no: Option<u64>` / `version: Option<u64>`, populated **inside `Index::search`**, at every place a hit's `_source` is resolved — reusing the version-map entry already fetched for that hit's own deleted/staleness check wherever one exists, doing a single fresh lookup where none does. Two spots that deliberately defer `_source` resolution for cheap candidate collection (`fill_memtable_sources`, and the sort_topk inline fill) now resolve `seq_no`/`version` at that *same* later moment, not before — so the source-vs-seq_no window never reopens, whichever point in the pipeline actually produces the final body.
- `es_compat.rs`'s four render-time re-lookups (PIT filter, `fields._version`, top-level `_seq_no`/`_version`) now just read `h.seq_no`/`h.version` off the hit.
- `ScrollContext` gains `seq_no_primary_term`/`version` bool flags, captured once at scroll-open (matches real ES: the flag governs the whole scroll, not each page). Both scroll routes now read real, snapshotted per-hit values instead of hardcoding, and omit the fields entirely when not requested.
- `scroll_page_response`'s hand-built JSON now actually inserts `_seq_no`/`_version`/`_primary_term` when present — the second half of the original #428 bug, still present on `main`.
- The `disable_sequence_numbers` `-2`/`0` sentinel (previously `search_impl`-only) is extracted into a shared helper and applied to scroll pages too — #440's own "adjacent, smaller, separate" gap.

Left deliberately out of scope: collapse `inner_hits`' `_seq_no`/`_version` (`es_compat.rs` ~13160-13175) — `members` there are already-rendered `Value`s, not `Hit`s, and `_seq_no` is currently a fabricated array index (`json!(i as u64)`), not even a real lookup. Pre-existing, separate bug; worth its own issue.

## Testing

- New `engine/crates/xerj-api/tests/scroll_seq_no_torn_read.rs`:
  - `multi_index_scroll_seq_no_matches_plain_search_per_index` — 2 indices, colliding ids, differing per-index write counts (so `_seq_no` isn't just a page ordinal), asserts every scrolled hit's `_seq_no`/`_version` against a direct GET on its own index. This is the exact shape that caught #431's repair attempt #2 (4/8 hits carrying the wrong index's version-map entry).
  - `scroll_continuation_reports_the_snapshot_pair_not_a_torn_one` — the direct #440 repro: concurrently updates a snapshotted doc between scroll page 1 and page 2, asserts the scrolled page reports the SNAPSHOT `_seq_no`/`_source` pair, not the live post-update one.
  - `scroll_continuation_honours_disable_sequence_numbers` — asserts the `-2`/`0` sentinel on a scroll continuation page, the YAML-conformance gap #440 names (`310_sequence_numbers_disabled.yml` has no scroll case).
  - All three fail-before: verified against the pre-fix tree (`git stash` of every source change, test file only) — all 3 fail with the expected symptom, confirming they exercise the real bug and don't pass vacuously.
- `cargo fmt`: clean.
- `cargo clippy -p xerj-api -p xerj-engine -p xerj-query --lib --tests -- -D warnings`: clean.
- `cargo test -p xerj-engine --lib`: 602 passed, 0 failed (2 pre-existing failures in `snapshot_path_security_tests`, confirmed unrelated — same failures reproduce on `main` before this change, an environment-specific `/tmp` symlink-resolution issue).
- `cargo test -p xerj-api --lib`: 223 passed, 0 failed.
- `cargo test -p xerj-api --test scroll_multi_index_identity --test scroll_snapshot_cap_is_documented`: all passing, no regressions on #414's existing coverage.
- `cargo check --workspace --lib --tests`: clean.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -10712,12 +10712,10 @@ async fn search_impl(
                 .get(&idx_name)
                 .copied()
                 .unwrap_or(u64::MAX);
-            let hit_seq = state
-                .engine
-                .get_index(&idx_name)
-                .ok()
-                .and_then(|idx| idx.lookup_seq_no(&h.id))
-                .unwrap_or(0);
+            // Read from the hit itself — resolved at construction time in
+            // the engine, from the same read as `_source` — rather than a
+            // live re-lookup here (#440).
+            let hit_seq = h.seq_no.unwrap_or(0);
             if hit_seq <= max_seq {
                 kept.push((idx_name, h));
             } else {
@@ -12070,15 +12068,12 @@ async fn search_impl(
                                 continue;
                             }
                             "_version" => {
-                                // Real per-doc `_version` via the same
-                                // resolver GET/_mget use (external-version
-                                // map wins, then the engine write counter).
-                                let v = state
-                                    .engine
-                                    .get_index(&idx_name)
-                                    .ok()
-                                    .and_then(|idx| idx.lookup_version(&h.id))
-                                    .unwrap_or(1);
+                                // Read from the hit itself — resolved at
+                                // construction time in the engine (same
+                                // external-version-map-wins resolver GET/
+                                // _mget use), not a later live re-lookup
+                                // here (#440).
+                                let v = h.version.unwrap_or(1);
                                 fmap.insert("_version".to_string(), Value::Array(vec![json!(v)]));
                                 continue;
                             }
@@ -13255,16 +13250,12 @@ async fn search_impl(
                 hit_inner_hits = Some(Value::Object(combined));
             }
 
-            // Pull the real seq_no (and a placeholder primary_term of 1)
-            // from the engine's version_map so multi-write docs surface
-            // their actual sequence number instead of a synthetic 0.
+            // Real seq_no (and a placeholder primary_term of 1), read from
+            // the hit itself — resolved at construction time in the engine,
+            // from the same read as `_source`, not a later live re-lookup
+            // here (#440).
             let (real_seq_no, real_primary_term) = if emit_seq_no {
-                let sn = state
-                    .engine
-                    .get_index(&idx_name)
-                    .ok()
-                    .and_then(|idx| idx.lookup_seq_no(&h.id))
-                    .unwrap_or(hit_idx as u64);
+                let sn = h.seq_no.unwrap_or(hit_idx as u64);
                 (Some(sn), Some(1u64))
             } else {
                 (None, None)
@@ -13288,16 +13279,13 @@ async fn search_impl(
                     None
                 },
                 version: if emit_version {
-                    // Real per-doc `_version` via the shared resolver
-                    // (external-version map wins so reindexes with
+                    // Real per-doc `_version`, read from the hit itself —
+                    // resolved at construction time in the engine (external-
+                    // version map wins so reindexes with
                     // `version_type=external[_gte]` echo the caller's
-                    // value, then the engine write counter).
-                    state
-                        .engine
-                        .get_index(&idx_name)
-                        .ok()
-                        .and_then(|idx| idx.lookup_version(&h.id))
-                        .or(Some(1))
+                    // value, then the engine write counter), not a later
+                    // live re-lookup here (#440).
+                    h.version.or(Some(1))
                 } else {
                     None
                 },
@@ -14263,6 +14251,11 @@ async fn search_impl(
             hits: snapshot,
             position: scroll_page_size,
             page_size: scroll_page_size,
+            // Captured once at open time (real ES semantics — see the
+            // field's doc comment) from the same flags that gated this
+            // first page's own `_seq_no`/`_version` emission above.
+            seq_no_primary_term: emit_seq_no,
+            version: emit_version,
             created: now,
             keep_alive,
             expires_at: now + keep_alive,
@@ -14433,42 +14426,7 @@ async fn search_impl(
     if any_disabled_seqno {
         if let Some(hits) = response_body.pointer_mut("/hits/hits") {
             if let Some(arr) = hits.as_array_mut() {
-                for hit in arr.iter_mut() {
-                    let ix = hit
-                        .get("_index")
-                        .and_then(Value::as_str)
-                        .unwrap_or("")
-                        .to_string();
-                    let disabled = state
-                        .engine
-                        .index_settings
-                        .get(&ix)
-                        .map(|v| {
-                            let s = v.clone();
-                            let as_bool = |val: &Value| {
-                                val.as_bool().unwrap_or_else(|| {
-                                    val.as_str().map(|x| x == "true").unwrap_or(false)
-                                })
-                            };
-                            s.pointer("/index/disable_sequence_numbers")
-                                .map(as_bool)
-                                .unwrap_or(false)
-                                || s.get("index")
-                                    .and_then(|i| i.get("index.disable_sequence_numbers"))
-                                    .map(as_bool)
-                                    .unwrap_or(false)
-                                || s.get("index.disable_sequence_numbers")
-                                    .map(as_bool)
-                                    .unwrap_or(false)
-                        })
-                        .unwrap_or(false);
-                    if disabled {
-                        if let Some(obj) = hit.as_object_mut() {
-                            obj.insert("_seq_no".into(), json!(-2i64));
-                            obj.insert("_primary_term".into(), json!(0));
-                        }
-                    }
-                }
+                apply_disable_seqno_sentinel(&state, arr);
             }
         }
     }
@@ -19834,17 +19792,34 @@ pub async fn search_with_scroll(
         // Index name stays paired with its hit (#414).
         let hits_only: Vec<(String, xerj_query::executor::Hit)> = all_hits.clone();
 
-        // Return first page.
+        // Control flags for meta-field emission — mirrors `search_impl`'s
+        // `emit_seq_no`/`emit_version` (es_compat.rs ~11349).
+        let emit_seq_no = body.seq_no_primary_term.unwrap_or(false);
+        let emit_version = body.version.unwrap_or(false);
+
+        // Return first page. `_seq_no`/`_version` are read from the hit
+        // itself — resolved at construction time in the engine, from the
+        // same read as `_source` — not hardcoded, and omitted entirely
+        // when not requested (#428/#440).
         let first_page: Vec<EsHit> = all_hits
             .iter()
             .take(page_size)
-            .map(|(idx_name, h)| EsHit {
+            .enumerate()
+            .map(|(hit_idx, (idx_name, h))| EsHit {
                 index: idx_name.clone(),
                 id: h.id.clone(),
                 score: Some(h.score as f64),
-                version: Some(1),
-                seq_no: Some(0),
-                primary_term: Some(1),
+                version: if emit_version {
+                    h.version.or(Some(1))
+                } else {
+                    None
+                },
+                seq_no: if emit_seq_no {
+                    Some(h.seq_no.unwrap_or(hit_idx as u64))
+                } else {
+                    None
+                },
+                primary_term: if emit_seq_no { Some(1) } else { None },
                 source: if h.source.is_null() {
                     None
                 } else {
@@ -19890,6 +19865,8 @@ pub async fn search_with_scroll(
             hits: hits_only,
             position: page_size,
             page_size,
+            seq_no_primary_term: emit_seq_no,
+            version: emit_version,
             created: now,
             keep_alive,
             expires_at: now + keep_alive,
@@ -20053,6 +20030,50 @@ pub async fn next_scroll_path(
     .await
 }
 
+/// Apply the ES `index.disable_sequence_numbers` sentinel — `_seq_no: -2`,
+/// `_primary_term: 0` — to every hit in `hits` whose own `_index` setting
+/// has it enabled. Shared by `search_impl` (plain `_search`) and
+/// `scroll_page_response` (#440: scroll had no equivalent at all, so a
+/// continuation page for such an index reported whatever `_seq_no` the
+/// snapshotted hit carried instead of the sentinel real ES uses).
+fn apply_disable_seqno_sentinel(state: &AppState, hits: &mut [Value]) {
+    for hit in hits.iter_mut() {
+        let ix = hit
+            .get("_index")
+            .and_then(Value::as_str)
+            .unwrap_or("")
+            .to_string();
+        let disabled = state
+            .engine
+            .index_settings
+            .get(&ix)
+            .map(|v| {
+                let s = v.clone();
+                let as_bool = |val: &Value| {
+                    val.as_bool()
+                        .unwrap_or_else(|| val.as_str().map(|x| x == "true").unwrap_or(false))
+                };
+                s.pointer("/index/disable_sequence_numbers")
+                    .map(as_bool)
+                    .unwrap_or(false)
+                    || s.get("index")
+                        .and_then(|i| i.get("index.disable_sequence_numbers"))
+                        .map(as_bool)
+                        .unwrap_or(false)
+                    || s.get("index.disable_sequence_numbers")
+                        .map(as_bool)
+                        .unwrap_or(false)
+            })
+            .unwrap_or(false);
+        if disabled {
+            if let Some(obj) = hit.as_object_mut() {
+                obj.insert("_seq_no".into(), json!(-2i64));
+                obj.insert("_primary_term".into(), json!(0));
+            }
+        }
+    }
+}
+
 async fn scroll_page_response(
     state: &AppState,
     scroll_id: String,
@@ -20089,6 +20110,12 @@ async fn scroll_page_response(
             let page_size = ctx.page_size.max(1);
             let position = ctx.position;
             let total = ctx.hits.len() as u64;
+            // Captured once at scroll-open time (real ES semantics) — see
+            // `ScrollContext::seq_no_primary_term`/`::version`'s doc
+            // comments. Read into locals before the per-hit closure below
+            // borrows `ctx.hits`.
+            let emit_seq_no = ctx.seq_no_primary_term;
+            let emit_version = ctx.version;
 
             // Detect whether the initial search sorted by a non-score key;
             // in that case `max_score` must be null on scroll pages too.
@@ -20098,19 +20125,34 @@ async fn scroll_page_response(
                 .map(|(_, h)| !h.sort.is_empty())
                 .unwrap_or(false);
 
+            // `_seq_no`/`_version` are read from each hit itself — resolved
+            // at construction time in the engine, from the same read as
+            // `_source`, snapshotted into `ctx.hits` at scroll-open (a
+            // single request-duration window, never re-read live) — not
+            // hardcoded, and omitted when the flag wasn't requested
+            // (#428/#440).
             let page_hits: Vec<EsHit> = ctx
                 .hits
                 .iter()
                 .skip(position)
                 .take(page_size)
-                .map(|(hit_index, h)| EsHit {
+                .enumerate()
+                .map(|(hit_idx, (hit_index, h))| EsHit {
                     // Per-hit index, not the context-level one (#414).
                     index: hit_index.clone(),
                     id: h.id.clone(),
                     score: Some(h.score as f64),
-                    version: Some(1),
-                    seq_no: Some(0),
-                    primary_term: Some(1),
+                    version: if emit_version {
+                        h.version.or(Some(1))
+                    } else {
+                        None
+                    },
+                    seq_no: if emit_seq_no {
+                        Some(h.seq_no.unwrap_or(hit_idx as u64))
+                    } else {
+                        None
+                    },
+                    primary_term: if emit_seq_no { Some(1) } else { None },
                     source: if h.source.is_null() {
                         None
                     } else {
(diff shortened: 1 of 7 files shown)


xerj-org:
## Independent verification — head `7a891349`

Three skeptics, isolated worktrees, rustc 1.97.1, each told to REFUTE the torn-read fix. **All three could not refute: 0 blocking.** Merging on this.

- **Torn read closed:** `_seq_no`/`_version` are now fields on `executor::Hit`, populated inside `Index::search` at the same point `_source` is resolved, and snapshotted into `ScrollContext.hits` at scroll-open. Continuation pages read `h.seq_no`/`h.version` off the snapshot — the four old render-time re-lookups are gone. `hit_seq_version` is arithmetically identical to the old `lookup_seq_no`/`lookup_version`.

- **Test is not vacuous:** reverting only the 6 source hunks (keeping the test) fails all 3 new tests, including the torn-read assertion (`scrolled hit must carry _seq_no`). The torn-read test updates a doc between page 1 and 2 and asserts the continuation reports the SNAPSHOT pair (seq + body v=1), not the live post-update value — the exact lost-update the maintainer's repro caught on #431.

- **No regression:** first page (both routes), plain `_search`, `search_after`/PIT, multi-index, deletion-mid-scroll all verified; Rust's exhaustive struct-init makes a forgotten field a compile error, not a silent 0. #428 confirmed fixed (origin/main hardcoded `seq_no:0`/`version:1` at 3 sites).

Two nits, filed as follow-ups, NOT blockers: collapse `inner_hits` still does a live `lookup_version` at render (es_compat.rs:13163 — pre-existing on main, explicitly scoped out); and a sub-millisecond intra-`search`-call window where source and version_map.get are separate reads (inherent to storage having no atomic source+version read; the same class as plain `_search` metadata staleness, and vastly narrower than the whole-scroll window it replaces).

---

### Lens 1 — `refuted: False`

**Verdict (verbatim):**

> REFUTATION FAILED — I cannot demonstrate a blocking torn read. The PR correctly closes the scroll-continuation window that refuted #431. HEAD confirmed at 7a89134.
> 
> Trace of the captured triple (_source, _seq_no, _version):
> - Capture point: seq_no/version are populated INSIDE Index::search at the same code point _source is resolved. Where a VersionEntry is already fetched for the doc's own deleted/superseded check (index.rs:16399, 23704, 23824, 24362), hit_seq_version reuses that SAME entry — genuinely one read producing all three. Where no entry is in hand, version_map.get is co-located with the source resolution; deferred-source paths (fill_memtable_sources, sort_topk inline, knn_result_from_scored) resolve seq_no at the same deferred fill moment, with a catch-all at ~19235 that fills any still-unresolved hit at the final bounded-page fill. hit_seq_version matches lookup_seq_no/lookup_version exactly (seq_no = ver.seq_no-1 saturating; version = external_versions wins else ver.version; deleted => (None,None)).
> - Snapshot point: search_impl clones merged_hits (all hits, resolved, size=SCROLL_SNAPSHOT_MAX_HITS) into scroll_snapshot -> ctx.hits at open (es_compat.rs:11219/14225/14251). search_with_scroll clones all_hits -> ctx.hits (19793/19865). ScrollContext also snapshots the emit flags once at open.
> - Render point: scroll_page_response reads h.seq_no/h.version straight off ctx.hits (es_compat.rs:20125-20160). No lookup_seq_no/lookup_version call remains in the scroll continuation top-level-hit path — the four old render-time re-lookups (PIT filter 10712, fields._version 12068, top-level _seq_no/_version 13250/13279) all now read the hit.
> 
> Attack results:
> - TORN READ (update-after-snapshot): CLOSED. Continuation serves the snapshotted seq_no beside the snapshotted _source; no live re-read. Verified by scroll_continuation_reports_the_snapshot_pair_not_a_torn_one (PASS on rustc 1.97.1): page 2 reports snapshot _seq_no and v=1 body, NOT the live post-update seq_no.
> - DELETION mid-scroll: continuation serves the snapshot triple captured at open (doc was live then), a consistent PIT view matching ES scroll semantics; no live/absent read.
> - MULTI-INDEX (the shape that caught #431 attempt #2): each index resolves its own hits' seq_no from its own version_map, paired per-hit (#414). Verified by multi_index_scroll_seq_no_matches_plain_search_per_index (PASS).
> - FIRST PAGE / regression: first page (both routes) reads the hit's real snapshotted seq_no/version and omits the fields when not requested; disable_sequence_numbers -2/0 sentinel now applied to scroll pages. Verified by scroll_continuation_honours_disable_sequence_numbers and scroll_multi_index_identity (all PASS).
> - MEMORY: +2x Option<u64> (~16-24B) per snapshot hit; cap is by count (10,000), unaffected; ~240KB worst case per context. Negligible.
> - No crash risk (saturating_sub, all Option-guarded).
> 
> All 3 new tests + 2 pre-existing identity tests pass on the CI toolchain (rustc 1.97.1).

**Findings (verbatim):**

- `nit` engine/crates/xerj-api/src/es_compat.rs:13163 (introduced_by_this_pr=False): Collapse inner_hits still does a live lookup_version at render time and fabricates _seq_no as the array index (json!(i as u64)); on a collapse+scroll this is a torn/synthetic pair for inner hits.

  emit_version_ih branch calls idx.lookup_version(id_str) live while _seq_no is a positional index, so an inner hit can report a live version beside a snapshot body. This is NOT introduced by this PR (identical on origin/main) and the PR explicitly scopes it out as a separate pre-existing bug; it is the collapse inner_hits path, not the top-level scroll hit this PR fixes. Not blocking.

- `nit` engine/crates/xerj-engine/src/index.rs:12238 (introduced_by_this_pr=True): In a few construction sites the snapshot's seq_no is read via a fresh version_map.get separate from the already-materialized `source` (from `scored`), so the open-time triple is atomic only to within one Index::search call, not a truly single atomic read.

  A write landing in the sub-millisecond gap between source materialization and the version_map.get within the SAME search call could bake a (new seq_no, old source) pair into the snapshot. This window is inherent to the storage lacking an atomic source+version read (the PR acknowledges 'xerj has no such structure'), is the same class as a plain _search's source-vs-metadata staleness, is vastly narrower than the whole-scroll-lifetime window it replaces, and is not reproducible deterministically through the HTTP API. Not a continuation-window torn read; not blocking.

---

### Lens 2 — `refuted: False`

**Verdict (verbatim):**

> Could not refute PR #499. The central #440 claim holds: _seq_no/_version are captured at the same read as _source and snapshotted into ScrollContext.hits at scroll-open, so continuation pages report the snapshot pair, never a live (post-update/deleted) value. The new test is real and exercises the actual torn-read interleaving. HEAD verified at 7a89134. Nothing blocking; one nit.
> 
> Vacuity check: reverting ONLY the source hunks (git checkout origin/main on the 6 source files, keeping tests/scroll_seq_no_torn_read.rs) makes all 3 new tests FAIL under rustc 1.97.1 — including the torn-read assertion (panic at scroll_seq_no_torn_read.rs:253 'scrolled hit must carry _seq_no', because pre-fix continuation JSON never emitted the field), the disable_sequence_numbers test (got None, expected -2), and the multi-index test. Restoring source: all 3 pass.
> 
> Torn read is genuinely tested: scroll_continuation_reports_the_snapshot_pair_not_a_torn_one opens a scroll (size=1, sort v asc, seq_no_primary_term+version), does a concurrent PUT /torn/_doc/1 {v:999} AFTER the snapshot, pages to doc '1', and asserts scrolled _seq_no != live_seq_after_update AND _source.v == 1 (snapshot body). Runs through the real HTTP router via oneshot.
> 
> Mechanism traced: scroll_snapshot = merged_hits.clone() (es_compat.rs:11216); body.size bumped to SCROLL_SNAPSHOT_MAX_HITS 10k (es_compat.rs:8467) so the engine's final catch-all (index.rs:~19235) resolves seq_no for the WHOLE snapshot, not just page 1. hit_seq_version reads seq_no/version from the same version_map entry that gates the deleted/stale check and _source. Continuation pages (scroll_page_response) read h.seq_no/h.version off the snapshotted hit, gated by ctx.seq_no_primary_term/version captured once at open (real ES flag-governs-whole-scroll semantics). Deletion mid-scroll: doc was live at snapshot, stays in snapshot with its snapshot seq_no — matches ES point-in-time view.
> 
> #428 confirmed fixed: origin/main hardcoded version:Some(1)/seq_no:Some(0)/primary_term:Some(1) at 3 sites and the hand-built continuation JSON never inserted _seq_no/_version at all; PR replaces with flag-gated real values from the hit and adds the fields to the wire map. #431's live-lookup-at-render approach is correctly avoided (no lookup_seq_no at render time — the 4 es_compat re-lookups now read h.seq_no/h.version).
> 
> No regression: xerj-api 223 lib + all integration suites pass (incl. scroll_multi_index_identity, scroll_window_tests, search_after validation, PIT), xerj-query 183 lib pass, clippy -D warnings clean on 1.97.1. Memory: +~24 bytes/hit into snapshots capped at 10k hits and bounded by max_open_scrolls — snapshot cap counts hits not bytes, so #405/#501 unaffected.

**Findings (verbatim):**

- `nit` engine/crates/xerj-api/src/es_compat.rs:20155 (and first-page 19819) (introduced_by_this_pr=True): Continuation/first-page render uses h.seq_no.unwrap_or(hit_idx as u64) as fallback; if any snapshot hit ever carried seq_no=None while emit_seq_no is true, it would silently emit a page-local ordinal (0 for each page's first hit) instead of a real or absent seq_no.

  Not demonstrable on any current path: for scroll, body.size is forced to the 10k cap (es_compat.rs:8467) so the engine's final catch-all (index.rs:~19235) resolves seq_no for every snapshotted hit before it reaches the context; the multi-index test confirms all 8 hits across continuation pages match a direct GET. The fallback is latent fragility only — a future query path leaving seq_no unresolved for hits beyond page 1 would surface it. Nit, not a defect.

---

### Lens 3 — `refuted: False`

**Verdict (verbatim):**

> Could not refute PR #499. HEAD confirmed at 7a89134. The fix is structurally sound across the whole scroll/search surface and I found no demonstrable blocking regression (no torn read, no pagination regression, no data loss, no crash).
> 
> WHY THE TORN READ IS ACTUALLY FIXED (not like #431): _seq_no/_version are now fields on xerj_query::executor::Hit, populated INSIDE Index::search at the same construction point _source is resolved, and snapshotted into ScrollContext.hits at scroll-open. Continuation pages read h.seq_no/h.version off the snapshot vector — never a live re-lookup. es_compat.rs's four former render-time re-lookups (PIT filter, fields._version, top-level _seq_no/_version) all now read the hit. The engine-side transform hit_seq_version() is arithmetically identical to the old lookup_seq_no/lookup_version (seq_no = ver.seq_no.saturating_sub(1); version = external_versions wins else ver.version; deleted -> (None,None)). Test scroll_continuation_reports_the_snapshot_pair_not_a_torn_one PASSES: a doc updated between page 1 and page 2 returns the SNAPSHOT (seq, source v=1) pair, not the live post-update seq — exactly the lost-update the maintainer's repro caught on #431.
> 
> REGRESSION LENS (my assignment):
> 1. First page: correct. Both scroll routes fixed. search_impl first page renders via the same h.seq_no path (es_compat ~13250); search_with_scroll's first-page EsHit builder now reads h.version/h.seq_no instead of the hardcoded Some(0)/Some(1) still present on origin/main (confirmed: origin/main es_compat.rs lines 19845-47, 19935-37, 20111-13 hardcode seq_no:0/version:1). multi_index test asserts EVERY hit including page 1 against a direct GET — passes.
> 2. Plain (non-scroll) _search: every hit returned from Index::search funnels through fill_memtable_sources (index.rs:17126), whose catch-all (19245) resolves seq_no/version for any hit left None. Deferred-None construction sites (14592, 15087) are on that path. kNN path resolves inline via knn_result_from_scored (now takes &Index). Sorted/keyed/scan collectors (22487, 23704, 23824, 24362) resolve inline. Rust's exhaustive struct init makes a forgotten field a compile error, not a silent 0.
> 3. search_after/PIT: unaffected; published_search_after_recipe_walks_the_whole_corpus and scroll_multi_index_identity pass.
> 4. Deletion mid-scroll: continuation returns the SNAPSHOTTED seq_no (doc was live at snapshot), matching ES point-in-time scroll semantics — this is the intended behavior, not a bug.
> 5. Snapshot bloat: two Option<u64> (~16-24 bytes/hit) added to Hit; the scroll cap (#405/#501) is count-based, not byte-based, so no interaction — scroll_snapshot_cap_is_documented tests pass.
> 
> TEST RESULTS (rustc 1.97.1, CI toolchain):
> - cargo test -p xerj-api: 0 failures (all suites)
> - cargo test -p xerj-api --profile ci-test: 0 failures
> - cargo test -p xerj-engine --lib: 606 passed; 0 failed (the 2 snapshot_path_security_tests failures the PR body flagged as env-specific did NOT reproduce here)
> - cargo test -p xerj-api --test scroll_seq_no_torn_read: 3 passed
> - cargo clippy --workspace --all-targets -- -D warnings: exit 0, clean
> - cargo fmt --check: exit 0, clean
> 
> Nit only (not demonstrable, not blocking): the continuation-page fallback `h.seq_no.unwrap_or(hit_idx as u64)` uses a page-local ordinal that resets to 0 each page, so IF a live snapshotted hit ever carried seq_no=None (only reachable if it were absent from version_map), two docs on different pages could report the same fabricated seq_no. Live snapshot hits are always in version_map, so this branch is effectively dead, and it mirrors search_impl's pre-existing fallback style — I could not construct an input that reaches it. Refuted=false.

**Findings (verbatim):**

- `nit` engine/crates/xerj-api/src/es_compat.rs (scroll_page_response continuation + search_with_scroll first page) (introduced_by_this_pr=True): The `h.seq_no.unwrap_or(hit_idx as u64)` fallback uses a per-page ordinal that resets to 0 each continuation page; if a live snapshotted hit were ever missing from the version_map (seq_no=None), two docs on different pages could report the same fabricated _seq_no.

  Code inspection only. I could not construct an input that reaches this branch: hit_seq_version returns Some for any live, non-deleted version_map entry, and a hit present in the scroll snapshot was live at snapshot time, so it is in the version_map. Not demonstrable; mirrors search_impl's pre-existing fallback. Non-blocking.

---

Issue #370, opened 2026-08-14
docs: the 10,000-document scroll snapshot cap is undocumented (split from #364)


xerj-org:
Narrow docs follow-up split out of #364 (which I filed in error and have closed — scroll works correctly).

## The gap

`_search?scroll=` refuses any query whose result set exceeds `SCROLL_SNAPSHOT_MAX_HITS = 10_000` (`es_compat.rs:31961`), returning a 400:

```
Scroll result set is too large: [11450] matching documents exceed the scroll snapshot
window of [10000]. xerj materialises the full scroll snapshot up front and will not
return a silently truncated result set. Page with [search_after] on a unique sort key
instead, or narrow the query.
```

That is good behaviour — it is #198's fix, and failing loudly beats truncating silently. But the cap is documented nowhere:

```
$ grep -rn "snapshot window\|SCROLL_SNAPSHOT\|10,000" docs/ | grep -i scroll
(no output)
```

## Why it is worth a line

Scroll is what the ES ecosystem reaches for to read a whole index — `elasticsearch-py`'s `helpers.scan()`, `_reindex` internals, most export and backup tooling. Those all target indices that are usually larger than 10k, which is exactly the case that 400s. The error message is excellent and names the alternative, so nobody is stuck; they are just surprised, and surprise on a compatibility surface is the thing the compatibility docs exist to prevent.

The distinction also matters for anyone assessing ES compatibility: "scroll is supported, with a 10k snapshot cap; use `search_after` beyond that" is a very different claim from "scroll is supported", and only the first one is true.

## Suggested

A row in the ES-compatibility docs stating the cap, the 400 it produces, and `search_after` as the unbounded path. Ideally a note in `llms-full.txt` §2 (honesty/capability boundaries) too, since that section already carries exactly this kind of bounded-capability statement.

**Verified:** cap constant at `es_compat.rs:31961`; the 400 reproduced on `v1.0.0-rc.16` against an 11,450-doc index; scroll confirmed working end to end (open, page, clear) on a 6,012-doc subset of the same index; `grep` over `docs/` returns nothing for the cap.

PR #501, opened 2026-08-19
fix(scroll): enforce the summed scroll snapshot cap on the alias route too (#405)

### LINKED ISSUE #405: es-compat: the scroll snapshot cap is applied per index on /{index}/_search_scroll, but to the summed total on /{index}/_search?scroll= (split from #370)

Vinz2168:
Fixes #405 (split out of #370).

## The bug

`SCROLL_SNAPSHOT_MAX_HITS` (10,000) is enforced two different ways depending on the route:

- `POST /{index}/_search?scroll=` (`search_impl`) compares the **summed** total across every resolved index against the cap.
- `POST /{index}/_search_scroll?scroll=` (`search_with_scroll`, the alias route) fetches up to `SCROLL_SNAPSHOT_MAX_HITS` from **each** resolved index separately, and only flagged truncation when a **single** index alone hit that per-index fetch cap.

An N-index request on the alias route can therefore materialise up to `N * SCROLL_SNAPSHOT_MAX_HITS` hits into one scroll context and never trip the check — the exact silently-truncated-export failure mode issue #198 exists to refuse. Reproduced (2 indices, 6,000 docs each): `_search?scroll=` correctly 400s at 12,000; `_search_scroll?scroll=` returns 200 with `hits.total.value: 12000` pinned into the context.

## The fix

`search_with_scroll`'s loop already computes `total_count` as the exact sum across every index (`total_count += result.total.value`, unaffected by the per-index materialisation cap) — the only thing missing was comparing it against `SCROLL_SNAPSHOT_MAX_HITS` directly, exactly like `search_impl` already does. Replaced the per-index `scroll_truncated` flag with that summed-total check.

This is a strict widening, not a behavior swap: the old flag only fired when a single index's true total exceeded the cap, and in that case the sum (which includes that index's total) exceeds the cap too — so every case the old check caught, the new one catches, plus the aggregate case it missed.

## Testing

New `engine/crates/xerj-api/tests/scroll_alias_multi_index_cap.rs`:
- `alias_route_refuses_a_summed_total_over_the_cap` — two indices, each individually under the cap (5,001 each), summing over it (10,002). Verified fail-before: returns 200 with all 10,002 hits silently pinned on the pre-fix tree.
- `alias_route_multi_index_under_cap_still_scrolls` — control, so the assertion above can't pass vacuously.
- `alias_route_still_refuses_a_single_index_over_the_cap` — the case the old check already caught, confirming no regression.

- `cargo fmt`: clean.
- `cargo clippy -p xerj-api --lib --tests -- -D warnings`: clean.
- `cargo test -p xerj-api --lib`: 223 passed, 0 failed.
- Existing scroll coverage (`scroll_multi_index_identity`, `scroll_snapshot_cap_is_documented`) unaffected.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -19784,10 +19784,6 @@ pub async fn search_with_scroll(
     // Execute search across all indices and collect ALL hits.
     let mut all_hits: Vec<(String, xerj_query::executor::Hit)> = Vec::new();
     let mut total_count: u64 = 0;
-    // Issue #198: set when any index matches more documents than the
-    // per-index snapshot fetch could carry — the scroll would be a
-    // silently truncated export.
-    let mut scroll_truncated = false;
 
     for idx_name in &index_names {
         let idx = match state.engine.get_index(idx_name) {
@@ -19805,11 +19801,6 @@ pub async fn search_with_scroll(
                 if let Some(reason) = &result.script_failure {
                     return script_limit_response(reason);
                 }
-                if result.hits.len() >= SCROLL_SNAPSHOT_MAX_HITS
-                    && result.total.value > result.hits.len() as u64
-                {
-                    scroll_truncated = true;
-                }
                 total_count += result.total.value;
                 for hit in result.hits {
                     all_hits.push((idx_name.clone(), hit));
@@ -19823,10 +19814,19 @@ pub async fn search_with_scroll(
 
     // If scroll param present, store context and return first page.
     if params.scroll.is_some() {
-        // Issue #198: refuse to register a context whose snapshot is
-        // missing documents — a truncated scroll looks complete to the
-        // caller (reindex/backup/migration) and silently loses data.
-        if scroll_truncated {
+        // Issue #198 (and #405: this route used to flag truncation only
+        // when a SINGLE index alone exceeded the per-index snapshot fetch
+        // — `full_req.size` is `SCROLL_SNAPSHOT_MAX_HITS` PER index, so an
+        // N-index request could materialise up to
+        // `N * SCROLL_SNAPSHOT_MAX_HITS` hits into one context and never
+        // trip that check, even though `POST /{index}/_search?scroll=`
+        // (search_impl) rejects the identical summed total. `total_count`
+        // above is already the exact sum across every index — compare it
+        // against the cap directly, matching search_impl's check
+        // (es_compat.rs, `is_scroll_request` block) exactly, so a request
+        // whose combined result set cannot fit the snapshot is refused
+        // loudly regardless of which route opened it.
+        if total_count > SCROLL_SNAPSHOT_MAX_HITS as u64 {
             return scroll_window_exceeded(total_count, SCROLL_SNAPSHOT_MAX_HITS);
         }
         let scroll_id = Uuid::new_v4().to_string();
--- a/engine/crates/xerj-api/tests/scroll_alias_multi_index_cap.rs
+++ b/engine/crates/xerj-api/tests/scroll_alias_multi_index_cap.rs
@@ -1,0 +1,187 @@
+//! Issue #405 (split out of #370): the scroll snapshot cap
+//! ([`SCROLL_SNAPSHOT_MAX_HITS`]) was enforced against the SUMMED total on
+//! `POST /{index}/_search?scroll=` but only PER INDEX on the alias route
+//! `POST /{index}/_search_scroll?scroll=`.
+//!
+//! `search_with_scroll` (the alias route's handler) fetches up to
+//! `SCROLL_SNAPSHOT_MAX_HITS` hits from EACH resolved index separately, so an
+//! N-index request could pin `N * SCROLL_SNAPSHOT_MAX_HITS` hits into one
+//! scroll context — the exact silently-truncated-export failure mode issue
+//! #198 exists to refuse — without ever tripping the old per-index check,
+//! since no single index alone exceeded the cap.
+//!
+//! Mirrors the fixture style of `scroll_snapshot_cap_is_documented.rs`
+//! (issue #370/#198): an under-cap control so the test cannot pass vacuously,
+//! then the real over-cap repro from #405 — two indices, each under the cap
+//! individually, whose SUM exceeds it.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+use xerj_api::es_compat::SCROLL_SNAPSHOT_MAX_HITS;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn send(app: &axum::Router, req: Request<Body>) -> (StatusCode, Value) {
+    let response = app.clone().oneshot(req).await.expect("response");
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .expect("body");
+    let value: Value = serde_json::from_slice(&bytes).unwrap_or(Value::Null);
+    (status, value)
+}
+
+async fn seed(app: &axum::Router, index: &str, docs: usize) {
+    let (st, body) = send(
+        app,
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent verification — head `8b9aacd9`

Three skeptics, isolated worktrees, rustc 1.97.1, each told to REFUTE. **All three could not refute: 0 blocking.** Merging.

- **Consistent with `_search?scroll=`:** the alias route now uses the identical `total_count > SCROLL_SNAPSHOT_MAX_HITS` predicate and the same `scroll_window_exceeded` helper, so both routes refuse with the same 400 `illegal_argument_exception` naming the true summed total. `total_count` is exact (`track_total_hits` forced true).

- **Data-loss case fixed:** 5001+5001=10002 now refused (was a silent 10,002-doc truncated snapshot). Fail-before confirmed: reverting only the source hunk makes `alias_route_refuses_a_summed_total_over_the_cap` fail `left:200 right:400` — the exact #405 export-truncation bug.

- **No regression:** confined to `search_with_scroll` (`_search?scroll=` untouched, verified by diff); a fitting multi-index scroll pages fully end to end (drain test: 700+500 → all 1200 docs then exhaustion); single-index-over-cap still refused; exactly-10,000 summed accepted (`>` not `>=`); empty scroll ok. Full api suite green both profiles, clippy/fmt clean.

Two nits, non-blocking, no follow-up needed: the under-cap control test asserts only the first page (continuation is covered by the separate `scroll_multi_index_identity` test); and an over-cap request materialises up to N×10k hits before the post-loop refusal — **pre-existing** (the loop and check were already post-loop on main), not introduced here.

---

**Verdict (verbatim):**

> Could not refute. The fix is correct: search_with_scroll (the _search_scroll alias route) now enforces the summed-total cap identically to search_impl (_search?scroll=). Both compare `total_count > SCROLL_SNAPSHOT_MAX_HITS` and return the same scroll_window_exceeded(total_count, cap) 400 (illegal_argument_exception), so the two routes are now consistent in threshold, off-by-one behavior, and error shape. total_count is the exact per-index sum (track_total_hits forced true), unaffected by the per-index fetch cap. Multi-index-over-cap-none-alone (5001+5001=10002) is refused and names the true summed total; the fits case (256) scrolls with correct total; single-index-over-cap (10001) still refused; exactly-10000 summed is admitted and is fully servable (matches search_impl); empty scroll admitted. The removed scroll_truncated flag has no dangling references and continuation is served by the untouched POST /_search/scroll route, so no pagination regression. New test passes (3 passed; 0 failed) and existing scroll tests (scroll_multi_index_identity, scroll_snapshot_cap_is_documented) still pass. Nothing blocking; one test-coverage nit.

**Findings (verbatim):**

- `nit` engine/crates/xerj-api/tests/scroll_alias_multi_index_cap.rs:148 (introduced_by_this_pr=True): The under-cap control test alias_route_multi_index_under_cap_still_scrolls only asserts the first page (status, _scroll_id, total.value=256); it never continues the scroll to exhaustion, so end-to-end pagination of a fitting multi-index alias scroll is not exercised by the new test.

---

**Verdict (verbatim):**

> PR #501 is correct and its test is genuine; nothing blocking. The fix makes `search_with_scroll` (the `_search_scroll` alias route) enforce the SUMMED scroll-snapshot cap (`total_count > SCROLL_SNAPSHOT_MAX_HITS`) instead of the old per-index `scroll_truncated` flag, matching `search_impl` (`_search?scroll=`) exactly. The two routes now use byte-identical check + error function, so the refusal is the same 400 `illegal_argument_exception` body naming the true summed total. The new test scroll_alias_multi_index_cap.rs is real: I reverted ONLY the es_compat.rs source hunk (restored the per-index flag) while keeping the test, and alias_route_refuses_a_summed_total_over_the_cap FAILED — pre-fix the route returned 200 with all 10,002 hits silently pinned (the exact #405 truncated-export bug), asserting left:200 right:400. With the fix restored all 3 tests pass on CI toolchain 1.97.1. Edge cases hold: exactly-10,000 summed is allowed on both routes (consistent, `>` not `>=`); the under-cap multi-index control still scrolls and reports total=256; a single index over the cap is still refused (the case the old check already caught). This is a pure widening — total_count is exact because track_total_hits is forced true, and any index whose total fits under the summed cap is fully materialised by the size-10,000 per-index fetch, so no fits-should-work case regresses.

---

**Verdict (verbatim):**

> REFUTATION FAILED — PR #501 is correct and I found nothing blocking. HEAD is 8b9aacd as required. The change is confined to search_with_scroll (the POST /{index}/_search_scroll alias route) in es_compat.rs; search_impl (POST /{index}/_search?scroll=) is untouched (verified via diff vs origin/main). The PR replaces the old per-index truncation flag with `if total_count > SCROLL_SNAPSHOT_MAX_HITS { return scroll_window_exceeded(total_count, cap) }` — the SAME predicate search_impl uses (es_compat.rs:14229) and the SAME helper (scroll_window_exceeded), so the two routes now refuse with an identical 400/illegal_argument_exception body that names the true summed total. total_count is exact here because scroll forces track_total_hits=Some(true), so there is no spurious or missed refusal, and the boundary is `> cap` (exactly 10,000 summed is accepted on both routes). ATTACK RESULTS: (1) summed-over-cap on the alias route now refuses — 2x5,001=10,002 returns 400 naming 10,002 (was previously a silent 10,002-doc truncated snapshot); (2) a fitting multi-index scroll still pages correctly end-to-end — my drain test opened da(700)+db(500), reported total 1200, and continuation pages returned all 1200 docs then exhausted; (3) single-index-over-cap (10,001) still refused, no double-count/regression; (4) exactly-10,000-summed accepted; (5) empty multi-index scroll succeeds with total 0. Error shape is consistent between routes (shared helper). TESTS: default xerj-api suite all green (223 in main file + all other targets, 0 failed); ci-test profile 412 passed / 0 failed with the unrelated script_limits_http call-depth tests skipped (those hang at 0% CPU in this environment — pre-existing, independent of the scroll change which does not touch script execution); clippy --workspace --all-targets -D warnings clean (exit 0); cargo fmt --check clean.

**Findings (verbatim):**

- `nit` engine/crates/xerj-api/src/es_compat.rs:19790-19830 (introduced_by_this_pr=False): In the refused (over-cap) multi-index case, search_with_scroll still fetches up to SCROLL_SNAPSHOT_MAX_HITS hits from EACH index into all_hits before the post-loop summed check rejects the request, so an N-index over-cap request briefly materialises up to N*10k hits in memory only to 400. This is wasted work but not a correctness issue.

---

PR #504, opened 2026-08-19
fix(search): reject sort on an unresolvable field instead of null-stranding search_after (#437)

### LINKED ISSUE #437: sort on any unresolvable field (11 meta-names + every unmapped/typo'd field) returns null on all hits and strands search_after at page one
Split from #402, which attempted a request-level allowlist and was refuted — the approach cannot work, but the defect it targeted is real and still open.

## What is still broken

#401 fixed four meta-fields (`_seq_no`, `_version`, `_primary_term`, `_index`) via #420. The underlying class is untouched: **any sort field that `compute_sort_values` cannot resolve returns `null` on every hit, HTTP 200, and strands `search_after` on page one.**

Measured on a 6-document index, `{"size":2,"sort":[{F:"asc"}]}` + `search_after`:

```
_field_names        pages=2  collected=['d1','d2']  unique=2 of 6   <- stranded, HTTP 200
_tier               pages=2  collected=['d1','d2']  unique=2 of 6
_source             pages=2  collected=['d1','d2']  unique=2 of 6
not_a_field_at_all  pages=2  collected=['d1','d2']  unique=2 of 6
_seq_no             pages=4  collected=6            unique=6 of 6   (control: #420 works)
```

The full all-null set on an ordinary corpus: `_source`, `_size`, `_doc_count`, `_field_names`, `_meta`, `_tier`, `_nested`, `_nested_path`, `_feature`, `_parent`, `_matched_queries` — **plus every unmapped or misspelled field name**. A typo in a sort field silently truncates a full-corpus walk to its first page.

## Why an allowlist cannot fix it

#402 tried listing the unresolvable names. Three successive measurements of "which fields return null" were each taken on a corpus that happened to lack the documents that would make the field resolve. That is not a measuring mistake to be corrected — it is structural:

`compute_sort_values` (`xerj-engine/src/index.rs:37123`) special-cases `_score`, `_doc`/`_id`, then `meta_sort_value`, and sends **everything else** to `get_field_value(source, field)`. And xerj deliberately permits metadata-named keys inside `_source`. Its own words, twice:

- `CHANGELOG.md:158` — "xerj — unlike Elasticsearch, which rejects a metadata-named key inside a document with a `document_parsing_exception` — accepts `{"_seq_no": 1}` in a document body and echoes it back in `_source`"
- `engine/crates/xerj-engine/tests/sort_on_meta_fields.rs:42-45` — "Note the premise these gates do NOT rest on: 'a meta-field name can never be a `_source` key'. It can"

So whether `_type` (or `_tier`, or `_meta`) resolves depends on **what is in the documents**, which a request-level allowlist cannot see. A static list is guaranteed to be wrong in both directions on some corpus.

## The two approaches that can work

**A — ES parity at ingest.** Reject metadata-named keys in document bodies with `document_parsing_exception`, as Elasticsearch does. Then meta-field names are never `_source` keys, the fall-through can never resolve one, and a static rule becomes sound. This is a breaking ingest change and needs a migration story for indices that already contain such keys.

**B — ES parity at query time.** When a sort field resolves to nothing for the whole result set, answer as ES does: `No mapping found for [x] in order to sort on`. This covers all twelve names *and* unmapped fields *and* typos without knowing the corpus in advance, and it fixes the actual user-visible harm — a silently truncated `search_after` walk.

B looks smaller and closes more of the class. A is the deeper fix and is what makes #420's four special-cases safe by construction rather than by luck.

## Also worth folding in

Whatever is built should be wired into **all** the search entry points. #402's gate reached only `search_impl`; `msearch_impl` (`es_compat.rs:22142`), `search_template` (`:29927`) and the async-search submit path bypassed it entirely, and sorts nested inside `collapse.inner_hits` and `top_hits` are never inspected at all.

**Verified:** the stranding table, the all-null census, and the endpoint divergence were measured on a live node by an independent verification of #402; the two quoted repo statements are in the tree today.
**Not verified:** the behaviour of an ES 8 cluster on each of these names — the parity claims are from ES documentation and the existing test comments, not from a running Elasticsearch.

Vinz2168:
Fixes #437.

## The bug

`compute_sort_values` special-cases `_score`, `_doc`, `_id` and the four meta-fields #420 made resolvable (`_seq_no`, `_version`, `_primary_term`, `_index`). Every other ES meta-field name (`_source`, `_size`, `_doc_count`, `_field_names`, `_meta`, `_tier`, `_nested`, `_nested_path`, `_feature`, `_parent`, `_matched_queries`) and any unmapped or misspelled field falls through to a plain `_source` lookup that returns `Null` when absent. The whole result set ties on `[null]`, HTTP 200, and `search_after` paging on that field is stranded at page one with no error anywhere — the same silently-truncated-export failure mode #198 closes for scroll, reached through `sort` instead.

## Why the previous attempt (#402) was refuted

#402 tried a request-level denylist of "known-unresolvable" field names. xerj deliberately permits a metadata-named key inside `_source` (`{"_seq_no": 1}` is a valid document body — real Elasticsearch rejects it), so whether a name resolves is a property of the **corpus**, not the field name. Every iteration of #402's list was refuted by a reviewer who built a corpus where the "unresolvable" field resolved. #402 was also wired into `_search` only — its own review caught `_msearch`/`_search/template`/`_async_search` answering the identical field with 200.

## The fix

Check the **schema** instead of the corpus — the same thing real ES validates, and stable per field name regardless of request content. A single gate inside `search_inner`, at the schema-read block already used by `unsearchable_query_field`. Every HTTP entry point that builds a `SearchRequest` (`_search`, `_msearch`, `_search/template`, `_async_search`) funnels through the same `Index::search` → `search_inner`, so this covers all of them structurally rather than needing each endpoint wired in by hand — exactly where #402 fell short.

Uses `declared_field`'s existing recursive dotted-path walk, not `Schema::has_field` (flat/exact-match only, would false-reject a `.keyword` multi-field or a nested/object property). Confirmed dynamically-mapped fields get real `FieldConfig` entries via `evolve_schema_from_doc`/`add_field`, synchronously with the write that introduces them — no staleness window against a field legitimately present via dynamic mapping.

Error shape reuses established infrastructure: `error.rs` already special-cases one `InvalidQuery` reason prefix (`"failed to create query:"` → `query_shard_exception` in `root_cause`, outer wrapper staying `search_phase_execution_exception`/400) for `unsearchable_query_field` — added one more branch for `"No mapping found for"`, reproducing ES's real envelope for this exact case.

## Deferred, not closed here

`collapse.inner_hits.sort` and aggregation `top_hits.sort` implement their own sorting independently of `compute_sort_values` and aren't covered by this gate. Both hit real friction on inspection: collapse `inner_hits`' rendering closure is sync and can't `.await` the schema's `tokio::sync::RwLock` the way `search_inner` does, and `top_hits` has no `Index`/schema handle anywhere in the aggs call chain at all. Both are separate, larger changes than this PR's single clean gate — noted explicitly in the CHANGELOG rather than left silent.

## Testing

- `engine/crates/xerj-engine/tests/sort_on_unresolvable_field.rs` (engine-level): the full #402-review census (11 meta names + a typo) rejected with the right reason; a mapped field and all seven already-handled special cases still work; a dynamic `.keyword` multi-field and a nested/object property resolve via schema too.
- `engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs` (HTTP-level): the same census against the real ES envelope shape; a direct `search_after` before/after proof (control walks the whole corpus, unresolvable field is refused up front instead of stranding page one); `_msearch` and `_search/template` each getting the same rejection, proving the single gate covers them without per-endpoint wiring.
- All new assertions verified fail-before against the pre-fix tree.
- `cargo fmt`: clean.
- `cargo clippy -p xerj-engine -p xerj-api --lib --tests -- -D warnings`: clean.
- `cargo test -p xerj-engine --lib`: 605 passed (2 pre-existing, unrelated `/tmp` symlink-resolution failures, confirmed on `main` before this change).
- `cargo test -p xerj-api`: 223 lib tests plus every integration suite passing (3 pre-existing, unrelated failures, same environment issue).
- Existing `sort_on_meta_fields.rs` (#420/#401) regression suite: 7/7 unaffected.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`sort` on a field XERJ cannot resolve returned `null` on every hit
+  instead of an error** ([#437](https://github.com/xerj-org/xerj/issues/437)).
+  `compute_sort_values` special-cased `_score`, `_doc`, `_id` and the four
+  meta-fields [#420](https://github.com/xerj-org/xerj/pull/420) made
+  resolvable (`_seq_no`, `_version`, `_primary_term`, `_index`); every other
+  ES meta-field name (`_source`, `_size`, `_doc_count`, `_field_names`,
+  `_meta`, `_tier`, `_nested`, `_nested_path`, `_feature`, `_parent`,
+  `_matched_queries`) and any unmapped or misspelled field fell through to a
+  plain `_source` lookup that returned `Null` when absent. The whole result
+  set tied on `[null]`, HTTP 200, and `search_after` paging on that field
+  was stranded at page one with no error anywhere — the same
+  silently-truncated-export failure mode
+  [#198](https://github.com/xerj-org/xerj/issues/198) closes for scroll,
+  reached through `sort` instead.
+
+  A prior attempt, [#402](https://github.com/xerj-org/xerj/pull/402), tried
+  a request-level denylist of "known-unresolvable" field names and was
+  closed after independent verification refuted it: XERJ deliberately
+  accepts a metadata-named key inside a document body (`{"_seq_no": 1}`,
+  where real Elasticsearch rejects it — see #420's entry above), so whether
+  a name resolves is a property of the *corpus* being searched, not of the
+  field name, and a static list is guaranteed wrong on some corpus. #402
+  was also wired into `_search` only, leaving `_msearch`,
+  `_search/template` and `_async_search` answering the identical
+  "unresolvable" field with 200.
+
+  This fix checks the index's **schema** instead of the corpus — the same
+  thing real Elasticsearch validates — with a single gate inside the
+  engine's `search_inner`, which every search entry point funnels through
+  regardless of endpoint. A sort field that is not one of the seven
+  already-resolvable special cases and is not declared in the schema (via
+  the same recursive dotted-path resolution already used for `.keyword`
+  multi-fields and nested/object properties) is now rejected up front with
+  Elasticsearch's real error: a 400 `search_phase_execution_exception`
+  wrapping `query_shard_exception`, reason `"No mapping found for [x] in
+  order to sort on"`.
+
+  **Known gap, not closed here:** `collapse.inner_hits.sort` and
+  aggregation `top_hits.sort` implement their own sorting independently of
+  `compute_sort_values` and are not covered by this gate — an unresolvable
+  field named in either still sorts silently wrong rather than erroring.
+  Tracked as a follow-up.
+
 ## [1.0.0-rc.18] - 2026-08-18
 
 ### Security
--- a/engine/crates/xerj-api/src/error.rs
+++ b/engine/crates/xerj-api/src/error.rs
@@ -194,6 +194,12 @@ impl ApiError {
             rt.to_string()
         } else if reason.starts_with("failed to create query:") {
             "query_shard_exception".to_string()
+        } else if reason.starts_with("No mapping found for") {
+            // #437: a sort field with no mapping is a shard-level query-
+            // build failure in real ES too, wrapped the same way as
+            // `failed to create query:` above — `query_shard_exception`
+            // in root_cause, outer wrapper unchanged.
+            "query_shard_exception".to_string()
         } else if reason.starts_with("function score query returned an invalid score:") {
             // ES validates function_score results and raises
             // `illegal_argument_exception` with a 400 status when a
--- a/engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs
+++ b/engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs
@@ -1,0 +1,312 @@
+//! Issue #437, from the outside: sorting on a field the engine cannot
+//! resolve must be rejected with ES's real error shape, on every HTTP entry
+//! point that accepts a `sort` clause — not just `_search`.
+//!
+//! Issue #402's request-level denylist only ever reached `search_impl`, and
+//! its own review caught `_msearch`/`_search/template`/`_async_search`
+//! answering 200 on the same "unresolvable" field name it rejected on
+//! `_search`. This suite proves the engine-level fix (a single gate inside
+//! `search_inner`, which every one of these handlers funnels through) closes
+//! that gap structurally rather than needing each endpoint wired in by hand.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn send(app: &axum::Router, req: Request<Body>) -> (StatusCode, Value) {
+    let response = app.clone().oneshot(req).await.expect("response");
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .expect("body");
+    let value: Value = serde_json::from_slice(&bytes).unwrap_or(Value::Null);
+    (status, value)
+}
+
+async fn seed(app: &axum::Router, index: &str, n: u64) {
+    let mut ndjson = String::new();
+    for i in 0..n {
+        ndjson.push_str(&format!(
+            "{{\"index\":{{\"_id\":\"{i}\"}}}}\n{{\"n\":{i}}}\n"
+        ));
+    }
+    let (st, body) = send(
+        app,
+        Request::builder()
+            .method("POST")
+            .uri(format!("/{index}/_bulk"))
+            .header("content-type", "application/x-ndjson")
+            .body(Body::from(ndjson))
+            .unwrap(),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "bulk {index}: {body}");
+    let (st, body) = send(
+        app,
+        Request::builder()
+            .method("POST")
+            .uri(format!("/{index}/_refresh"))
+            .body(Body::empty())
+            .unwrap(),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "refresh {index}: {body}");
+}
+
+/// The real ES envelope: `search_phase_execution_exception`/400 wrapping
+/// `query_shard_exception` in `root_cause`, reason naming the field.
+fn assert_no_mapping_found_error(status: StatusCode, body: &Value, field: &str) {
+    assert_eq!(
+        status,
+        StatusCode::BAD_REQUEST,
+        "sort on unresolvable field [{field}] must be a 400, not silently answered: {body}"
+    );
+    assert_eq!(
+        body["error"]["type"], "search_phase_execution_exception",
+        "wrong outer error type for [{field}]: {body}"
+    );
+    assert_eq!(
+        body["error"]["root_cause"][0]["type"], "query_shard_exception",
+        "wrong root_cause type for [{field}]: {body}"
+    );
+    let reason = body["error"]["reason"].as_str().unwrap_or_default();
+    assert!(
+        reason.contains(&format!("No mapping found for [{field}]")),
+        "reason for [{field}] doesn't name the field: {reason}"
+    );
+    assert_eq!(
+        body["status"], 400,
+        "wrong top-level status for [{field}]: {body}"
+    );
+}
+
+#[tokio::test]
+async fn search_rejects_sort_on_every_unresolvable_meta_field_and_a_typo() {
+    let (app, _dir) = app().await;
+    seed(&app, "census", 4).await;
+
+    const FIELDS: &[&str] = &[
+        "_source",
+        "_size",
+        "_doc_count",
+        "_field_names",
+        "_meta",
+        "_tier",
+        "_nested",
+        "_nested_path",
+        "_feature",
+        "_parent",
+        "_matched_queries",
+        "not_a_field_at_all",
+    ];
+    for field in FIELDS {
+        let (status, body) = send(
+            &app,
+            Request::builder()
+                .method("POST")
+                .uri("/census/_search")
+                .header("content-type", "application/json")
+                .body(Body::from(
+                    json!({"query": {"match_all": {}}, "sort": [{*field: "asc"}]}).to_string(),
+                ))
+                .unwrap(),
+        )
+        .await;
+        assert_no_mapping_found_error(status, &body, field);
+    }
+}
+
+/// The actual user-visible harm the issue is about: `search_after` paging
+/// on an unresolvable field silently stops at page one instead of erroring.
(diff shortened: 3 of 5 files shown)


xerj-org:
## Independent verification — head `406aa685` — DO NOT MERGE (blocking regression)

Three skeptics, isolated worktrees, rustc 1.97.1. **All three refuted, with the same blocking finding**, which I then reproduced myself.

### The blocking regression: the gate ignores `unmapped_type`

The core #437 fix is correct — a genuinely unresolvable field now errors, `_score`/`_doc`/`_id`/the four resolvable meta-fields still sort, and a **mapped-but-null** field is not rejected (the gate checks schema *declaration*, not per-doc value, so the null-vs-unresolvable distinction is right). But:

- Sorting an **unmapped field that carries `unmapped_type`** is now hard-rejected with 400 `No mapping found for [x] in order to sort on`. Real Elasticsearch returns **200** and treats every doc as missing — that is the entire purpose of `unmapped_type`. Two lenses showed HEAD→400 vs parent→200 end to end.

- **It breaks XERJ's own endpoint.** `memory_api.rs:1210` emits `sort:[{"stored_at":{"order":"desc","unmapped_type":"long"}}]` with the comment *"`unmapped_type` guards a namespace whose entries were all created before `stored_at` existed"*. For such a legacy namespace the memory-list endpoint now returns a 400 `list failed` instead of the entries. I confirmed both the memory_api call site and the gate myself.

- **Root cause:** `SortField` (sort.rs:23) has no `unmapped_type` field and `parse_sort` / `parse_sort_field_spec` silently drop it, so the gate at index.rs:13850-13862 sees only `declared_field(schema, field).is_none()` and rejects.

### Fix

Thread `unmapped_type` through `SortField` and both parse paths, and exempt it in the gate: when the field is unresolvable **but** `unmapped_type` is present, do not reject — let it resolve to missing on every doc (the pre-PR behaviour, which matches ES). A test must cover both directions (unmapped + `unmapped_type` → 200 all-missing; unmapped without it → 400).

I'm reworking this into a superseding PR with that exemption (crediting the author), rather than merging a regression. #437 stays open until it lands.

---

### Lens 1 (correctness) — `refuted: True`

**Verdict (verbatim):**

> REFUTED — one BLOCKING ES-incompatibility, demonstrated. The PR's core direction is sound (sort on a truly unresolvable field should 400, and valid fields — mapped fields, _score/_doc/_id, and the four resolvable meta-fields — still sort; the new tests cover both directions and the census reproduces #437). But the schema gate is unconditional: it rejects a sort even when the request supplies `unmapped_type`, which is exactly the documented Elasticsearch escape hatch that MUST return 200. `parse_sort` (es_compat.rs) drops `unmapped_type` entirely before the gate runs, so `SortField` never carries it and the gate cannot honour it. Real ES: unmapped field WITHOUT unmapped_type => 400 'No mapping found...'; WITH unmapped_type => 200, field treated as that type all-missing (used for cross-index sort). This PR returns 400 for BOTH. Demonstrated: `{"sort":[{"notafield":{"order":"desc","unmapped_type":"long"}}]}` => HTTP 400 query_shard_exception on this branch, whereas main returned 200 (same fall-through path; the PR does not touch parse_sort). This breaks a real ES client using the escape hatch, and XERJ's own memory_api.rs:1210 emits precisely this sort (`stored_at` + unmapped_type:long) for its defensive branch, which would now 400 the memory-list endpoint for any namespace whose schema lacks stored_at. The two new tests never exercise unmapped_type, so the PR's 'matches what real ES validates' claim is unproven for the one ES case it gets wrong.

**Findings (verbatim):**

#### `blocking` engine/crates/xerj-engine/src/index.rs:13843 (sort gate) + engine/crates/xerj-api/src/es_compat.rs:6491-6520 (parse_sort drops unmapped_type)
**Claim:** The new sort gate hard-rejects an unmapped sort field even when the request supplies `unmapped_type`, breaking the documented Elasticsearch escape hatch (ES returns 200 and treats the field as that type, all-missing). parse_sort silently discards `unmapped_type`, so the gate never sees it and rejects with 400. This is a behavioral regression vs main (which returned 200) and breaks real ES clients — including XERJ's own memory-list endpoint, which emits sort {stored_at, unmapped_type:long}.

**Demo:** On this branch (406aa685), POST /umt/_search with body {"sort":[{"notafield":{"order":"desc","unmapped_type":"long"}}]} returns: STATUS=400, {"error":{"root_cause":[{"type":"query_shard_exception","reason":"No mapping found for [notafield] in order to sort on"}],"type":"search_phase_execution_exception",...},"status":400}. Real ES returns HTTP 200 for this exact request because unmapped_type is present. Pre-fix main returned 200 (the request follows the same _source-fallthrough path the PR describes; parse_sort is unchanged by this PR, so unmapped_type was and still is dropped). Reproduced via engine/crates/xerj-api/tests/zz_demo_unmapped_type.rs with `rustup run 1.97.1 cargo test -p xerj-api --test zz_demo_unmapped_type -- --nocapture` (test FAILED asserting 200, got 400). SortField (crates/xerj-query/src/sort.rs:23) has no unmapped_type field, confirming the value is thrown away at parse time.

#### `should-fix` engine/crates/xerj-engine/tests/sort_on_unresolvable_field.rs + engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs
**Claim:** Neither new test file exercises `unmapped_type` at all (grep: zero hits). The PR asserts it now 'matches what real ES actually validates', but the one ES case it gets wrong — an unmapped field WITH unmapped_type, which ES accepts — is untested, so the compatibility claim is unproven exactly where it fails. The CHANGELOG likewise never mentions that unmapped_type is now rejected.

**Demo:** grep -n unmapped_type on both test files returns only doc-comment prose, no test input. The census tests all send bare field names with no unmapped_type. A regression test sending {field:{unmapped_type:keyword}} expecting 200 would have caught the blocking finding above.

#### `nit` engine/crates/xerj-engine/tests/sort_on_unresolvable_field.rs:106 (sort_on_a_resolvable_field_still_works)
**Claim:** The 'valid field still sorts' control only tests field `n` which is present in every doc. It never tests a field that IS mapped in the schema but null in some docs (legitimately missing). That case works correctly (declared_field checks schema, not corpus, so it passes the gate and sorts nulls-last), but it is asserted nowhere, leaving the 'resolvable-but-null vs unresolvable' distinction — the crux of the fix — unproven by test.

**Demo:** Inspection: the gate at index.rs uses declared_field(schema,...).is_none(); a field mapped once then absent in later docs stays in FieldConfig, so the gate passes and compute_sort_values yields null with SortMissing handling. Correct behavior, but no test seeds a mapped-then-missing field to lock it in.

---

### Lens 2 (regression) — `refuted: True`

**Verdict (verbatim):**

> REFUTED on one axis: the fix breaks ES `unmapped_type` compatibility. The core fix is otherwise correct — sort on a mapped field still works, a mapped-but-null field is NOT rejected (the gate checks schema declaration, not per-doc value, so the null-vs-unresolvable distinction is right), typos/unsortable meta-fields now error, and _score/_doc/_id/_seq_no/_version/_primary_term/_index still sort. Both PR test suites pass (3/3 engine, 5/5 HTTP). But the new gate ignores the `unmapped_type` sort option: a sort on an unmapped field carrying `unmapped_type` — a documented ES feature used for cross-index sorts and used by XERJ's own memory_api list endpoint — is now hard-rejected with 400 where ES (and the pre-PR tree) returns 200. SortField has no unmapped_type field and parse_sort never reads it, so the gate can't honour it.

**Findings (verbatim):**

#### `blocking` engine/crates/xerj-engine/src/index.rs:13850-13862 (the new sort gate); root cause also engine/crates/xerj-api/src/es_compat.rs:6491-6513 parse_sort and engine/crates/xerj-query/src/sort.rs:22-37 SortField
**Claim:** The new unresolvable-field gate ignores ES's `unmapped_type` sort option and hard-rejects (HTTP 400) a sort on an unmapped field that carries `unmapped_type`, which real Elasticsearch answers with HTTP 200 (all docs treated as missing). This is an ES-incompatibility that breaks real clients, including XERJ's own memory_api list endpoint, and is a behavioral regression versus the pre-PR tree.

**Demo:** Index 4 docs into `umt` with only field `n` (so `stored_at` is unmapped), then POST /umt/_search {"sort":[{"stored_at":{"order":"desc","unmapped_type":"long"}}]}. On PR HEAD 406aa68: HTTP 400 {"error":{"root_cause":[{"type":"query_shard_exception","reason":"No mapping found for [stored_at] in order to sort on"}],"type":"search_phase_execution_exception"},"status":400}. On parent c6ef027 (same test): HTTP 200 returning all 4 docs with sort:[null] — matching ES's unmapped_type semantics. Root cause: xerj-query SortField has no `unmapped_type` field and es_compat::parse_sort reads only order/mode/missing/format, so unmapped_type is silently dropped and the gate at index.rs cannot honour it; it sees only `declared_field(schema,"stored_at").is_none()` and rejects. Real ES specifically does NOT error when unmapped_type is supplied (that is the entire purpose of the option). memory_api.rs:1210 issues exactly this query pattern ({"stored_at":{"order":"desc","unmapped_type":"long"}}) as a defensive guard for namespaces predating stored_at; for such a namespace the list endpoint now returns `list failed` / 400 instead of the entries.

---

### Lens 3 (evidence) — `refuted: True`

**Verdict (verbatim):**

> REFUTED — one BLOCKING ES-incompatibility regression demonstrated. The PR correctly fixes the #437 null-stranding for genuinely unresolvable fields (valid mapped/keyword/nested fields, _score/_doc/_id and the four resolvable meta-fields still sort; mapped-but-null fields still sort; multi-sort rejects the whole request; error is a proper 400 query_shard_exception, not a 500). BUT the gate ignores `unmapped_type`: sorting an unmapped field that carries `unmapped_type` — which real ES accepts and treats as all-missing, and which XERJ's parser silently discards — is now hard-rejected with 400. That breaks a real ES client (the whole purpose of unmapped_type is to avoid this error) and regresses XERJ's own memory_api.rs legacy-namespace list sort from 200 to a 400 `list failed`. cargo test -p xerj-engine --lib (609 passed) and -p xerj-api (all passed), clippy --workspace --all-targets -D warnings clean, fmt --check clean, all on rustc 1.97.1.

**Findings (verbatim):**

#### `blocking` engine/crates/xerj-engine/src/index.rs:13850-13863 (gate); root cause engine/crates/xerj-query/src/parser.rs:3218 parse_sort_field_spec discards unmapped_type
**Claim:** Sort on an UNMAPPED field carrying `unmapped_type` is now hard-rejected with 400 `No mapping found for [x] in order to sort on`, but real Elasticsearch accepts `unmapped_type` as the documented escape hatch that makes an unmapped-field sort succeed (every doc treated as missing). The parser parses and silently discards `unmapped_type` (SortField has no field for it), and the new gate has no exemption for it, so any ES client relying on `unmapped_type` — the entire reason the key exists — is newly broken. Pre-fix XERJ ignored the key and returned HTTP 200 (all-null / all-missing, the ES-equivalent visible outcome); post-fix returns 400. This is an ES-incompatibility that breaks a real client, and it also breaks XERJ's OWN memory_api.rs:1210 list endpoint, whose sort `{"stored_at":{"order":"desc","unmapped_type":"long"}}` is explicitly written (comment at :1208) to survive a legacy namespace where `stored_at` was never stored — that path now returns a 400 `list failed` instead of the entries.

**Demo:** Engine test on an index with docs {"n":i} (no `stored_at`): parse_request of {"sort":[{"stored_at":{"order":"desc","unmapped_type":"long"}}]} yields SortField.field="stored_at" (unmapped_type dropped by parse_sort_field_spec). idx.search(req) then returns Err Common(InvalidQuery { reason: "No mapping found for [stored_at] in order to sort on" }). Real ES returns 200 with all docs sorted as missing; pre-fix XERJ also returned 200 (field fell through to _source lookup -> null on every hit). So a request valid in ES and previously answered by XERJ now 400s.

---

xerj-org:
Superseding this with #514, which keeps your core fix and adds the one thing the independent review found blocking: honouring ES's `unmapped_type` so the gate doesn't 400 a sort on an unmapped field that carries it (which also breaks XERJ's own `memory_api` legacy-namespace list). Threaded `unmapped_type` through SortField + both parse paths + a gate exemption, with a both-directions HTTP test. Full credit to you — #514 is built on your branch. Thanks for the fix; the core of it is unchanged.

PR #508, opened 2026-08-19
ci(landing): wire the landing-constants-guard into CI, then re-stamp the site (#489)

### LINKED ISSUE #489: landing-constants-guard.sh fails and runs in no workflow: the site has been stamped RC.15 for three releases
`.github/scripts/landing-constants-guard.sh` exists, works, currently **fails**, and is wired into no workflow. Consequently the public site has been stamped `RC.15` through two full releases.

## Reproduce

```
$ bash .github/scripts/landing-constants-guard.sh
FAIL  version stamps disagree with the newest released tag (v1.0.0-rc.17)
FAIL  the site claims both six and seven canonical agent operations
landing-constants-guard: 2 check(s) failed

$ grep -rl 'RC\.15' landing --include='*.html' | wc -l
47

$ grep -rn "landing-constants-guard" .github/workflows/
(no output)
```

The front page headline is literally `RC.15.<br>AND COUNTING.` (`landing/index.html:239`), and the footer says `XERJ.AI · V1.0.0-RC.15 · 2026` (`landing/index.html:331`). Those lines last moved in `b0bfce9`, 2026-08-12. rc.16 and rc.17 both shipped after that and neither touched them.

## Why the guard not running is the actual defect

The stale stamps are a symptom. The repo already decided this class of drift was worth catching in CI — someone wrote the check, and it is good enough to catch both defects today on the first run. It is simply never executed, so it has been failing silently for three releases while the tree stayed "green".

This is the same shape as several defects rc.18 is shipping fixes for: a check that exists, passes review, and never runs. Worth fixing as *wiring*, not as a one-off content edit, because editing the 47 files without wiring the guard buys exactly one release before it drifts again — which is the observed history.

## Second failure is unrelated and also real

`the site claims both six and seven canonical agent operations` — the site says seven in `landing/index.html:232`, `landing/llms-full.txt:231` and `landing/docs/agents/index.html:114-118`, and six elsewhere. The guard's own note explains the discrepancy (`xerj_autoindex` is the seventh and is CLI-only, with no HTTP route), so the count depends on whether you are counting HTTP routes or operations. One number, stated consistently, with the CLI-only caveat where it matters.

## Fix

1. Add the guard to a workflow. It is fast and needs no toolchain, so it belongs beside the other docs gates rather than in the release job — catching this at PR time is the point.
2. Then make it pass: re-stamp the 47 files, and settle six-vs-seven.
3. Consider whether the version stamp should be generated rather than hand-written. 47 hand-edited copies of a release number is a gate that will keep firing; a single source read at build time is a gate that stops mattering.

## Related, and the reason this surfaced

Round 1 of the independent pass on #460 flagged `landing/llms.txt:116` (`Current release: **v1.0.0-rc.17** (2026-08-15)`) as inconsistent with the rc.18 notes. I left it, arguing that it is true until the tag exists and is updated post-tag by convention.

That argument was wrong, and round 2 disproved it: there is no such convention. The one time that line moved it took 2d5h after the rc.17 tag and happened incidentally inside an unrelated docs PR — and the rest of the site did not move at all. What looked like "updated after tagging" is "updated when someone happens to notice". Recording it here because the same reasoning would otherwise get applied again at the next cut.

`landing/llms.txt` is the file agents read to pin a version, so it is the worst place in the tree for this to rot.

Vinz2168:
Fixes #489.

## The bug

`.github/scripts/landing-constants-guard.sh` exists, works, and currently **fails** — but is wired into no workflow:

```
$ bash .github/scripts/landing-constants-guard.sh
FAIL  version stamps disagree with the newest released tag (v1.0.0-rc.18)
FAIL  the site claims both six and seven canonical agent operations
landing-constants-guard: 2 check(s) failed

$ grep -rn "landing-constants-guard" .github/workflows/
(no output)
```

The front page headline was literally `RC.15.<br>AND COUNTING.` and the footer said `V1.0.0-RC.15` on 47 files (83 occurrences) — while rc.16, rc.17 and now rc.18 shipped around it.

## Why the wiring is the actual fix

Per the issue: re-stamping the 47 files without wiring the guard buys exactly one release before it drifts again, which is the observed history (the stamps froze at `b0bfce9`, 2026-08-12, and neither rc.16 nor rc.17 touched them). Added as its own job in `ci.yml`, beside the other fast, no-toolchain docs/tooling gates (`reference-coding`) — it belongs at PR time, not in the release job.

`fetch-depth: 0` is required (not the default `fetch-depth: 1`): the guard resolves "the newest released tag" via `git tag --list`, which a shallow checkout does not fetch — without it the guard would silently no-op its own version check under CI.

## Then made it pass

- Blanket-replaced the literal string `V1.0.0-RC.15` with `V1.0.0-RC.18` across all 47 flagged files (83 occurrences). Verified first that every occurrence sits in a footer span with no snapshot-marked historical exception sharing the exact uppercase string, so a tree-wide `sed` was safe.
- `landing/index.html`'s "SHIPPING NOW" headline/narrative wasn't caught by the guard's footer-scoped regex, but is the same defect the issue names directly: `RC.15. AND COUNTING.` → `RC.18. AND COUNTING.`, `15 release candidates in 34 days` → `18 ... in 43 days`, `v1.0.0-rc.15 on 2026-08-10` → `v1.0.0-rc.18 on 2026-08-19` — verified against the real rc.1/rc.18 GitHub release dates.
- Settled the six-vs-seven contradiction: `llms.txt` described the OpenAI/Anthropic tool-schema files as defining "the same six operations", reading as a total-operation-count claim against "seven canonical agent operations" stated everywhere else. Reworded to "six of the seven canonical operations — the HTTP-reachable ones", with the existing CLI-only `xerj_autoindex` caveat — those specific schema files genuinely define six tools, so the fix is in how the sentence reads, not the number.

## Not done here

The issue's own fix list floats generating the version stamp instead of hand-writing 47 copies of it — a gate that keeps firing on the same class of drift is a sign the representation is wrong, not just the process around it. That's a bigger, separate change; this PR is the wiring plus the re-stamp it now requires.

## Verification

- `bash .github/scripts/landing-constants-guard.sh` passes clean (all 5 checks) against the current tree.
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` confirms the new job parses and sits correctly in job order.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -87,6 +87,26 @@ jobs:
       - name: Vetted hub manifests validate
         run: python3 tools/xerj-code/tests/validate_manifest.py --hub tools/xerj-code/hub/*.json
 
+  # #489: this script existed, worked, and had been failing locally for three
+  # releases — the version footer, headline and narrative on landing/index.html
+  # stayed pinned to rc.15 through rc.16 and rc.17 because nothing ran it. It is
+  # fast, needs no toolchain, and only touches static HTML/text, so it belongs
+  # beside the other docs/tooling gates rather than in the release job —
+  # catching drift at PR time is the point, not after the site has been stale
+  # for weeks. fetch-depth: 0 is required (not the default fetch-depth: 1):
+  # the guard resolves "the newest released tag" via `git tag --list`, which a
+  # shallow checkout does not fetch.
+  landing-constants-guard:
+    name: Landing constants guard
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - name: Landing page constants agree with the newest release
+        run: bash .github/scripts/landing-constants-guard.sh
+
   lint:
     name: Format + Clippy
     runs-on: ubuntu-latest
--- a/landing/agent-search/index.html
+++ b/landing/agent-search/index.html
@@ -220,7 +220,7 @@ curl -X POST localhost:9200/kb/_search -H 'Content-Type: application/json' -d '{
 </div>
 
 <footer class="footer">
-  <span>XERJ.AI · AGENT SEARCH · V1.0.0-RC.15</span>
+  <span>XERJ.AI · AGENT SEARCH · V1.0.0-RC.18</span>
   <span>PORTED FROM THE ENGINE · BIT-EXACT</span>
   <span><a href="/index.html">HOME</a> · <a href="/docs/recipes/semantic-search-rag.html">SEMANTIC RECIPE</a> · <a href="/docs/recipes/index.html">RECIPES</a> &middot; <a href="/llms.txt">LLMS.TXT</a></span>
 </footer>
--- a/landing/docs/agents/endpoints.html
+++ b/landing/docs/agents/endpoints.html
@@ -320,7 +320,7 @@ POST /logs/_bulk
 </div>
 
 <footer class="footer">
-  <span>XERJ.AI · DOCS · V1.0.0-RC.15</span>
+  <span>XERJ.AI · DOCS · V1.0.0-RC.18</span>
   <span>READ THE SOURCE</span>
   <span><a href="/index.html">HOME</a> · <a href="/product.html">PRODUCT</a> · <a href="/brand.html">BRAND</a> · <a href="/playground.html">DASHBOARDS</a> &middot; <a href="/llms.txt">LLMS.TXT</a></span>
 </footer>
--- a/landing/docs/agents/index.html
+++ b/landing/docs/agents/index.html
@@ -183,7 +183,7 @@ $ curl -sXPOST localhost:9200/_memory/agent-demo/_recall \
 </div>
 
 <footer class="footer">
-  <span>XERJ.AI · DOCS · V1.0.0-RC.15</span>
+  <span>XERJ.AI · DOCS · V1.0.0-RC.18</span>
   <span>READ THE SOURCE</span>
   <span><a href="/index.html">HOME</a> · <a href="/product.html">PRODUCT</a> · <a href="/brand.html">BRAND</a> · <a href="/playground.html">DASHBOARDS</a> &middot; <a href="/llms.txt">LLMS.TXT</a></span>
 </footer>
--- a/landing/docs/agents/quickstart.html
+++ b/landing/docs/agents/quickstart.html
@@ -228,7 +228,7 @@ $ curl -sXDELETE localhost:9200/notes
 </div>
 
 <footer class="footer">
-  <span>XERJ.AI · DOCS · V1.0.0-RC.15</span>
+  <span>XERJ.AI · DOCS · V1.0.0-RC.18</span>
   <span>READ THE SOURCE</span>
   <span><a href="/index.html">HOME</a> · <a href="/product.html">PRODUCT</a> · <a href="/brand.html">BRAND</a> · <a href="/playground.html">DASHBOARDS</a> &middot; <a href="/llms.txt">LLMS.TXT</a></span>
 </footer>
--- a/landing/docs/aggregations.html
+++ b/landing/docs/aggregations.html
@@ -226,7 +226,7 @@
 
   <!-- Print-only branded footer (hidden on screen, visible in PDF) -->
   <div class="print-footer" style="display:none;">
-    <span>XERJ.AI · V1.0.0-RC.15 · 2026-08-11</span>
+    <span>XERJ.AI · V1.0.0-RC.18 · 2026-08-11</span>
     <a href="https://xerj.org/docs/">https://xerj.org/docs/</a>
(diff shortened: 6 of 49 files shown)


xerj-org:
## Independent verification — head `b2fd57a5`

Three skeptics, isolated worktrees. **Two returned refuted=false / zero blocking; the third (re-stamp-correctness) died on an API error.** Because that dead lens's axis is the most important one for a 49-file re-stamp, I did not treat the death as a pass — I re-verified it myself, and the two surviving lenses had independently covered it too. Merging.

- **Guard passes and is real:** `bash .github/scripts/landing-constants-guard.sh` → exit 0, all checks; injecting a stale `V1.0.0-RC.15` makes it fail (exit 1), reverting restores clean. Resolves the newest tag `v1.0.0-rc.18`.

- **CI job correctly wired:** `checkout@v4` with `fetch-depth: 0` (so `git tag --list` resolves the tag), runs the guard from repo root against the shipped tree, non-zero exit fails the build. Diff is entirely under `landing/` + `ci.yml` — no engine/root-docs collateral.

- **Re-stamps clean (self-verified):** no `href`/`src`/anchor/asset-path corruption; the changed lines are display strings only (`RC.18.<br>AND COUNTING`, footer `V1.0.0-RC.18`, the narrative date); every remaining `rc.15/16/17` mention is legitimate provenance ("v1.0.0-rc.17 does not [have it]", "FIXED IN v1.0.0-rc.16", "through v1.0.0-rc.15 `--no-graph` indexed…") and correctly untouched.

- **Six-vs-seven resolved correctly:** uniformly "seven canonical operations", six HTTP-reachable tools (both tool schemas define exactly six), `xerj_autoindex` the CLI-only seventh.

One should-fix (non-blocking, pre-existing, out of the guard's scope): `landing/llms.txt:116` still says `Current release: v1.0.0-rc.17`. Filed as a follow-up — the durable fix is to widen the guard to check the "Current release" line, not just footers.

---

### Lens (guard-passes) — `refuted: False`

**Verdict (verbatim):**

> Could not refute PR #508. HEAD confirmed at b2fd57a5. The guard passes on this branch (EXIT=0, resolves newest tag v1.0.0-rc.18); the re-stamp is clean (only footer version spans, the index.html headline/narrative, and one llms.txt sentence changed — no URL, asset path, anchor, code sample, or historical date corrupted, and the footer date 2026-08-11 correctly left as-is); the CI job is correctly wired (push/PR to main, checkout@v4 with fetch-depth:0 so git tag --list works, guard run from repo root against the shipped tree, non-zero exit fails the job, YAML valid); the guard is not a no-op (injecting V1.0.0-RC.15 makes it fail EXIT=1); and the six-vs-seven contradiction is resolved consistently and correctly (seven canonical operations, six HTTP-reachable, xerj_autoindex CLI-only seventh). Diff is entirely under landing/ and .github/workflows/ci.yml — no engine/docs collateral. No engine test gates landing constants. Nothing blocking.

---

### Lens (restamp-correct) — `refuted: False`

**Verdict (verbatim):**

> Nothing blocks this PR. The landing-constants-guard PASSES clean on this branch (exit 0, all 7 checks OK) and is provably real — injecting a stale `V1.0.0-RC.15` footer into landing/index.html makes it FAIL (exit 1); reverting restores a clean pass. The CI job is correctly wired: it triggers on pull_request and push to main, uses actions/checkout@v4 with fetch-depth: 0 (required so `git tag --list` can resolve the newest tag, confirmed v1.0.0-rc.18), runs `bash .github/scripts/landing-constants-guard.sh` from the repo root against the shipped tree, and a guard failure fails the step and thus the build. Re-stamps are correct: footers changed the display string only (no href/src/URL/anchor/asset path contains an uppercase RC), the index.html "SHIPPING NOW" narrative is internally consistent (rc.1 2026-07-07 to rc.18 2026-08-19 = 43 days, 18 RCs — arithmetic checks out), and lowercase feature-provenance refs (`since v1.0.0-rc.16/17`, `up to and including rc.15`) were correctly left untouched. The six-vs-seven contradiction is resolved and correct: the site is now uniformly "seven canonical operations," both openai-tools.json and anthropic-tools.json genuinely define exactly six tools (no xerj_autoindex), and xerj_autoindex is correctly described as the CLI-only seventh with no HTTP route. One should-fix (non-blocking): landing/llms.txt still advertises a stale "Current release: v1.0.0-rc.17 (2026-08-15)" that the PR did not update and the guard is deliberately not scoped to catch.

**Findings (verbatim):**

- `should-fix` landing/llms.txt:116 (introduced_by_this_pr=False): An agent-facing shipped page still advertises 'Current release: v1.0.0-rc.17 (2026-08-15)' while the actual newest tag is v1.0.0-rc.18, contradicting the just-updated index.html which now says v1.0.0-rc.18 shipped 2026-08-19. This is version drift of exactly the class issue #489 targets, but it sits outside the guard's deliberate footer-only scope so the new CI gate will not catch it.

---

PR #509, opened 2026-08-19
fix(autoindex): qualify the %PDF- magic signature (#403)

### LINKED ISSUE #403: sniff: %PDF- is an unqualified printable magic matched before the magic table — prose about PDF sniffs as a PDF

Vinz2168:
Fixes #403.

## The bug

Same class of defect as #379/#380, filed separately: `sniff_bytes` matches `%PDF-` as an early return, before `MAGIC_TABLE`, with no structural qualifier — all five bytes are printable ASCII:

```rust
if prefix.starts_with(b"%PDF-") {
    return Ok(mk(Family::Pdf));
}
```

Prose opening `"%PDF- is the five byte header every PDF file begins with. "` sniffs `Family::Pdf` and is handed to the PDF extractor, which cannot parse it. Lower severity than #379 (`%` and `-` shrink the collision surface a lot, and the failure is a parse error rather than #379's silent junk reason), but the same wrong answer.

## The fix

Added `pdf_header`: the real header is `%PDF-M.N` immediately followed by an end-of-line marker, which ISO 32000-1 §7.5.2 requires right after the version — a sentence has a space (or nothing) after "1.7", never a CR/LF.

Wired into the existing early return rather than moved into `MAGIC_TABLE` — the issue's own fix sketch floated the move, but every `MAGIC_TABLE` row reports `Family::Binary` + a `binary_kind` string, while `%PDF-` has to report the distinct `Family::Pdf` the extractor routes on (`lib.rs:1229`). Widening the shared `MagicRow` type for the one row that needs a different `Family` is a bigger, riskier change than this bug needs.

## Testing

- Flipped `signatures_matched_before_the_table`'s `pdf` case, which the repo had deliberately pinned to the wrong answer with a comment reading *"It must read TxtProse when #403 lands"* — now `TxtProse`, matching the `sqlite` and `zip` cases beside it.
- Added `a_real_pdf_header_still_classifies_as_pdf`, the positive control `every_printable_signature_carries_a_qualifier` can't provide for `%PDF-` (that test walks `MAGIC_TABLE` only): three spec-conformant headers (LF, CRLF, bare-CR line endings) must still classify `Family::Pdf`, or `pdf_header` would be a second way to break the format rather than a qualifier.
- Verified fail-before by temporarily reverting only the early-return (keeping the flipped test expectation): fails exactly as expected (`left: Pdf, right: TxtProse`), then restored byte-for-byte, diffed against the real fix to confirm.

## Gates

- `cargo fmt --check`: clean.
- `cargo clippy -p xerj-autoindex --lib --tests -- -D warnings`: clean.
- `cargo test -p xerj-autoindex --lib sniff`: 49/49 passed (48 pre-existing + 1 new).
- One unrelated, pre-existing failure in `sync_executor`'s tests (a byte comparison unrelated to sniffing) reproduces identically with this change reverted — confirmed unrelated before committing.

### DIFF
--- a/engine/crates/xerj-autoindex/src/sniff.rs
+++ b/engine/crates/xerj-autoindex/src/sniff.rs
@@ -189,7 +189,16 @@ fn sniff_bytes(
     }
 
     // 1. Magic bytes.
-    if prefix.starts_with(b"%PDF-") {
+    //
+    // %PDF- stays here rather than in MAGIC_TABLE (#403's own fix sketch
+    // floated moving it): every MAGIC_TABLE row reports `Family::Binary` +
+    // a `binary_kind` string, but %PDF- has to report the distinct
+    // `Family::Pdf` the extractor routes on (lib.rs:1229) — widening the
+    // shared `MagicRow` type for the one row that needs a different Family
+    // is a bigger, riskier change than this bug needs. `pdf_header` gives it
+    // the same structural qualifier the table's own rows carry; see its doc
+    // comment for what it rejects.
+    if prefix.starts_with(b"%PDF-") && pdf_header(prefix) {
         return Ok(mk(Family::Pdf));
     }
     if prefix.starts_with(b"SQLite format 3\0") {
@@ -938,6 +947,29 @@ fn id3v2_header(prefix: &[u8]) -> bool {
         && matches!(prefix[3], 2..=4)
         && prefix[4] != 0xff
         && prefix[6..10].iter().all(|b| *b < 0x80)
+}
+
+/// PDF header: `%PDF-M.N` — a single-digit major version, `.`, a single-digit
+/// minor version — immediately followed by an end-of-line marker, which the
+/// spec requires right after the version (ISO 32000-1 §7.5.2, "the header
+/// line shall be immediately followed by a comment... the header
+/// itself... terminated by an end-of-line marker").
+///
+/// Unqualified, this signature is five ASCII characters including `%` and
+/// `-`. Prose opening "%PDF- is the five byte header every PDF file begins
+/// with. " sniffed `Family::Pdf` and was handed to the PDF extractor, which
+/// cannot parse it (#403 — same class as #379/#380, filed separately: `%`
+/// and `-` shrink the collision surface a lot, and the wrong answer is a
+/// parse error rather than #379's silent junk reason). The qualifier is
+/// exactly what makes it safe: a sentence has a space (or nothing) after
+/// "1.7", never a CR/LF, and real generators always emit that break —
+/// Adobe's own spec calls it out as required, not conventional.
+fn pdf_header(prefix: &[u8]) -> bool {
+    prefix.len() >= 9
+        && prefix[5].is_ascii_digit()
+        && prefix[6] == b'.'
+        && prefix[7].is_ascii_digit()
+        && matches!(prefix[8], b'\r' | b'\n')
 }
 
 /// Truevision TGA, which has no magic number at all — every byte of its
@@ -2899,15 +2931,10 @@ mod printable_magic_tests {
     /// — they are early returns in `sniff_bytes` — so the walk above says
     /// nothing about them. Two of the three are safe for the usual reason (a
     /// byte text cannot contain: the NUL after `3`, and `\x03\x04`), and this
-    /// pins that.
-    ///
-    /// `%PDF-` is NOT safe and is the same class as #379/#380: prose opening
-    /// `%PDF-` sniffs `Family::Pdf` and is handed to the PDF extractor.
-    /// Measured on this branch and filed as #403 rather than widened into this
-    /// fix — five characters including `%` and `-` is a far smaller collision
-    /// surface than `BM`, and it does not produce #379's silent "binary
-    /// content" junk reason. The wrong answer is asserted here on purpose so
-    /// #403 has to come to this file and flip it.
+    /// pins that. `%PDF-` needed its own structural qualifier (`pdf_header`,
+    /// #403 — same class as #379/#380) since all five of its bytes are
+    /// printable ASCII; this pins that prose opening with the bare signature
+    /// now reads `TxtProse`, not `Pdf`.
     #[test]
     fn signatures_matched_before_the_table() {
         for (label, opener, want) in [
@@ -2921,11 +2948,10 @@ mod printable_magic_tests {
                 "PK is the local file header signature of the ZIP format. ",
                 Family::TxtProse,
             ),
-            // #403: this one is wrong. It must read TxtProse when #403 lands.
             (
                 "pdf",
                 "%PDF- is the five byte header every PDF file begins with. ",
-                Family::Pdf,
+                Family::TxtProse,
             ),
         ] {
             let text = opener.repeat(30);
@@ -2936,6 +2962,31 @@ mod printable_magic_tests {
                 "{label}: prose opening with a pre-table signature classified \
                  {:?}/{:?}",
                 sn.family, sn.binary_kind
+            );
+        }
+    }
+
+    /// The positive control `every_printable_signature_carries_a_qualifier`
+    /// cannot provide for `%PDF-`, since it walks `MAGIC_TABLE` only: a real
+    /// PDF header must still classify `Family::Pdf`, or `pdf_header` is not a
+    /// qualifier but a second way to break the format.
+    #[test]
+    fn a_real_pdf_header_still_classifies_as_pdf() {
+        let headers: [&[u8]; 3] = [
+            b"%PDF-1.7\n%\xE2\xE3\xCF\xD3\n1 0 obj\n",
+            b"%PDF-1.4\r\n1 0 obj\n",
+            b"%PDF-2.0\r%comment\r",
+        ];
+        for header in headers {
+            let p = Path::new("doc.pdf");
+            let sn = sniff_bytes(header, p, p, false).unwrap();
+            assert_eq!(
+                sn.family,
+                Family::Pdf,
+                "real PDF header {header:?} classified {:?}/{:?} — pdf_header \
+                 rejected a spec-conformant file",
+                sn.family,
+                sn.binary_kind
             );
         }
     }


xerj-org:
## Independent verification — head `fe148606`

Three skeptics, isolated worktrees, rustc 1.97.1, each told to REFUTE. **All three could not refute: 0 blocking.** Merging on this.

- **Real PDFs still detected** (the blocking attack): all 13 genuine PDFs under `landing/` (`%PDF-1.4\n%\xd3...`) still classify `Family::Pdf`; the new test pins LF/CRLF/bare-CR headers.

- **Prose fixed:** the issue's exact string `"%PDF- is the five byte header..."` now classifies `TxtProse` (offset 8 is a space, not EOL).

- **Qualifier is correct and spec-based:** `%PDF-` + digit `.` digit + CR/LF (ISO 32000-1 §7.5.2), consistent with the GIF8/BM (#379/#380) structural approach. Only `sniff.rs` changed, additive to the existing early return; ordering and other signatures untouched.

- **Not vacuous:** reverting only the source hunk makes `signatures_matched_before_the_table` FAIL (`left:Pdf/right:TxtProse`). 679 tests pass both profiles, clippy `-D warnings` clean, fmt clean.

Nits, non-blocking and not needing a follow-up (they are the accepted residual of any structural magic qualifier, same as the merged GIF/BM fix): a text file whose first line is byte-identical to a real PDF header still sniffs `Pdf`; a non-conformant PDF with whitespace before the EOL would sniff text (no real generator emits that); the qualifier is a deliberate one-off for `%PDF-` rather than generalized via the MAGIC_TABLE walk, justified because `%PDF-` must report `Family::Pdf`.

---

**Verdict (verbatim):**

> Could not refute. The pdf_header qualifier is correct and consistent with the GIF8/BM (#379/#380) structural-qualifier approach. It requires %PDF-D.D immediately followed by CR/LF (ISO 32000-1 §7.5.2). Verified on rustc 1.97.1: (1) a real header %PDF-1.7\n... still classifies Family::Pdf; (2) the issue's exact prose "%PDF- is the five byte header..." now classifies TxtProse; (3) "%PDF-1.4" with nothing after -> TxtProse (len<9; not a real file); (4) "%PDF-" + plain ASCII prose -> TxtProse. The PR's positive control exercises all three EOL forms (LF, CRLF, bare CR) and passes. `cargo test -p xerj-autoindex --lib sniff` = 49 passed / 0 failed. No real spec-conformant PDF is rejected, the issue's prose is no longer junked, and no crash. Nothing blocking.

**Findings (verbatim):**

- `nit` engine/crates/xerj-autoindex/src/sniff.rs:967 (pdf_header) and residual behavior (introduced_by_this_pr=False): A text/markdown file whose very first line is byte-identical to a real PDF header line (e.g. "%PDF-1.4\n...") still classifies Family::Pdf and is handed to the PDF extractor. This is the inherent, accepted residual of any structural magic qualifier (the same residual the GIF8/BM fixes carry) and is genuinely indistinguishable from a real PDF by header alone.

  Temporary probe test on rustc 1.97.1: input b"%PDF-1.4\nThis file documents the PDF header format.\n" -> sniff_bytes returns Family::Pdf. However this is far narrower than the pre-fix 'any prose starting %PDF-', the case is byte-identical to a real header, and per the PR it yields a parse error not a crash. Not a wrong-outcome regression introduced by this PR.

- `nit` engine/crates/xerj-autoindex/src/sniff.rs:971-975 (matches!(prefix[8], b'\r' | b'\n')) (introduced_by_this_pr=True): The qualifier demands an EOL IMMEDIATELY after the minor-version digit. A hypothetical non-conformant-but-real PDF that emits a trailing space before the EOL (e.g. "%PDF-1.4 \r\n") would sniff as TxtProse and be mis-indexed.

  pdf_header(b"%PDF-1.4 \r\n") returns false because prefix[8]==b' '. I could not produce or cite an actual spec-conformant PDF in the wild that does this — ISO 32000-1 §7.5.2 requires the EOL immediately after the version, and real generators comply — so this is speculative, not demonstrable, and hence a nit rather than blocking.

---

**Verdict (verbatim):**

> Could not refute. PR #509 is correct and should not block. The `pdf_header` qualifier (len>=9, byte[5] digit, byte[6]=='.', byte[7] digit, byte[8] in {CR,LF}) matches every real PDF version (all single-digit.single-digit: 1.0-1.7, 2.0) while rejecting prose that merely opens with "%PDF-". Fail-before is real: reverting only the source hunk (keeping the flipped test) makes signatures_matched_before_the_table FAIL with left:Pdf/right:TxtProse; restored byte-for-byte the tree is clean. Both directions are asserted — prose "%PDF- is the five byte header..." -> TxtProse, and three spec-conformant headers (LF/CRLF/bare-CR) -> Pdf. Real PDFs (%PDF-1.7\n, %PDF-1.4\r\n, %PDF-2.0\r) confirmed still Pdf via probe. #403 reproduced end-to-end: origin/main's unqualified early return yields Pdf, this branch yields TxtProse. 49/49 sniff tests pass on rustc 1.97.1. The fix is a deliberate one-off for %PDF- (not generalized via the MAGIC_TABLE walk) because %PDF- must report Family::Pdf rather than Family::Binary+binary_kind — justified in the PR and acceptable per the review rubric.

**Findings (verbatim):**

- `nit` engine/crates/xerj-autoindex/src/sniff.rs:964 (pdf_header) (introduced_by_this_pr=True): A non-spec-conformant PDF header with whitespace between the version and the EOL marker (e.g. "%PDF-1.4 \r\n" or "%PDF-1.4\t...") sniffs as TxtProse because byte[8] is not CR/LF.

  Probe test: sniff_bytes(b"%PDF-1.4 \r\n1 0 obj\n") -> TxtProse; sniff_bytes(b"%PDF-1.4\t1 0 obj\n") -> TxtProse. However spec-conformant headers %PDF-1.7\n, %PDF-1.4\r\n, %PDF-2.0\r all -> Pdf. ISO 32000-1 §7.5.2 requires the EOL immediately after the version and no common generator (ghostscript, libreoffice, pdftex, Adobe) emits trailing whitespace there, so this is not a genuine-PDF regression — noted only as a theoretical edge.

- `nit` engine/crates/xerj-autoindex/src/sniff.rs:192 (early-return) (introduced_by_this_pr=False): The fix is a one-off qualifier for %PDF- wired into the early return rather than a generalized MAGIC_TABLE-walk guard like the GIF8/BM #379/#380 fix, so a NEW printable early-return signature added before MAGIC_TABLE could reintroduce the same class without any structural safety net.

  %PDF-, SQLite format 3\0 and PK\x03\x04 remain hardcoded early returns; the positive-control walk every_printable_signature_carries_a_qualifier covers MAGIC_TABLE rows only. The PR justifies the one-off (Family::Pdf vs Family::Binary+binary_kind cannot share MagicRow), which the review rubric explicitly accepts. Stated as a design note, not a defect.

---

**Verdict (verbatim):**

> Could not refute PR #509. The fix is correct, consistent with the GIF8/BM (#379/#380) precedent, and does not regress real PDF detection.
> 
> HEAD confirmed fe14860. gh pr diff confirms only engine/crates/xerj-autoindex/src/sniff.rs changed; the edit is purely additive (`&& pdf_header(prefix)` appended to the existing `%PDF-` early return) — early-return ordering and every other signature (SQLite, %!PS/PostScript path, MAGIC_TABLE) are untouched.
> 
> REAL PDF still detected (the BLOCKING attack): the repo ships 13 genuine PDFs under landing/. Every one opens `%PDF-1.4\n%\xd3\xeb\xe9\xe1...`; simulating pdf_header() over all 13 returns true, so they still route to Family::Pdf. The PR's new test `a_real_pdf_header_still_classifies_as_pdf` additionally pins LF, CRLF, and CR-only variants (%PDF-1.7\n, %PDF-1.4\r\n, %PDF-2.0\r) — all classify Pdf and pass on both profiles. pdf_header covers all three PDF-legal EOL markers because CR/LF/CRLF each begin with \r or \n at offset 8. PDF versions are always single-digit.single-digit, so the digit/`.`/digit shape is not too strict.
> 
> PROSE fix works: `signatures_matched_before_the_table` now pins the issue's exact string "%PDF- is the five byte header..." to TxtProse (offset 8 is a space, not EOL), and it passes.
> 
> Edge cases checked, none misclassify a genuine file: BOM/whitespace-led PDFs already failed the pre-existing `starts_with(b"%PDF-")` at offset 0 (not a regression this PR introduces); truncated bare "%PDF-" (<9 bytes) correctly falls through to text. No panic risk — `prefix.len() >= 9` short-circuits before indexing [5..=8].
> 
> Full suite green both profiles (679 passed each), clippy --all-targets -D warnings clean, fmt --check clean.
> 
> Nit (not blocking, not introduced by this PR): a text file whose first line is exactly "%PDF-1.4\n" on its own would still sniff Pdf. This is the same residual collision the GIF/BM structural qualifiers accept, and it is strictly narrower than the pre-PR behavior (which matched any prose starting "%PDF-"), so the PR only improves the situation.

---

PR #514, opened 2026-08-19
fix(search): reject sort on an unresolvable field, but honour unmapped_type (#437, supersedes #504)

### LINKED ISSUE #437: sort on any unresolvable field (11 meta-names + every unmapped/typo'd field) returns null on all hits and strands search_after at page one

xerj-org:
Closes #437. Supersedes #504 (by @Vinz2168) — its core fix is kept; this adds the one thing its independent review found blocking.

## What #504 got right (kept)
A single gate in `search_inner` rejects a sort on a schema-undeclared field, closing the #437 null-stranding on every HTTP entry point. `_score`/`_doc`/`_id`, the four resolvable meta-fields, and **mapped-but-null** fields all still sort.

## The blocking regression it introduced (fixed here)
The gate ignored ES's `unmapped_type`. Sorting an unmapped field carrying `unmapped_type` is **not** an error in Elasticsearch (all docs treated as missing), and **XERJ's own memory-list** issues exactly `{"stored_at":{"order":"desc","unmapped_type":"long"}}` (`memory_api.rs:1210`) as a legacy-namespace guard. #504 would 400 those. The parser silently dropped the key, so the gate never saw it.

## Fix
Thread `unmapped_type` through `SortField` and both parse paths (`es_compat::parse_sort`, `xerj-query::parse_sort_field_spec`); exempt it in the gate — an unresolvable field carrying `unmapped_type` is admitted and resolves to missing on every doc, matching ES and restoring the pre-regression 200.

New HTTP test pins **both** directions: unmapped + `unmapped_type` → 200 (4 docs, missing-sorted); without → the same 400.

`sort_on_unresolvable_field` 3/3, HTTP suite **6/6** both profiles, xerj-query lib 183/183, clippy `--all-targets -D warnings` clean, fmt clean.

Original work by @Vinz2168 in #504.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,51 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`sort` on a field XERJ cannot resolve returned `null` on every hit
+  instead of an error** ([#437](https://github.com/xerj-org/xerj/issues/437)).
+  `compute_sort_values` special-cased `_score`, `_doc`, `_id` and the four
+  meta-fields [#420](https://github.com/xerj-org/xerj/pull/420) made
+  resolvable (`_seq_no`, `_version`, `_primary_term`, `_index`); every other
+  ES meta-field name (`_source`, `_size`, `_doc_count`, `_field_names`,
+  `_meta`, `_tier`, `_nested`, `_nested_path`, `_feature`, `_parent`,
+  `_matched_queries`) and any unmapped or misspelled field fell through to a
+  plain `_source` lookup that returned `Null` when absent. The whole result
+  set tied on `[null]`, HTTP 200, and `search_after` paging on that field
+  was stranded at page one with no error anywhere — the same
+  silently-truncated-export failure mode
+  [#198](https://github.com/xerj-org/xerj/issues/198) closes for scroll,
+  reached through `sort` instead.
+
+  A prior attempt, [#402](https://github.com/xerj-org/xerj/pull/402), tried
+  a request-level denylist of "known-unresolvable" field names and was
+  closed after independent verification refuted it: XERJ deliberately
+  accepts a metadata-named key inside a document body (`{"_seq_no": 1}`,
+  where real Elasticsearch rejects it — see #420's entry above), so whether
+  a name resolves is a property of the *corpus* being searched, not of the
+  field name, and a static list is guaranteed wrong on some corpus. #402
+  was also wired into `_search` only, leaving `_msearch`,
+  `_search/template` and `_async_search` answering the identical
+  "unresolvable" field with 200.
+
+  This fix checks the index's **schema** instead of the corpus — the same
+  thing real Elasticsearch validates — with a single gate inside the
+  engine's `search_inner`, which every search entry point funnels through
+  regardless of endpoint. A sort field that is not one of the seven
+  already-resolvable special cases and is not declared in the schema (via
+  the same recursive dotted-path resolution already used for `.keyword`
+  multi-fields and nested/object properties) is now rejected up front with
+  Elasticsearch's real error: a 400 `search_phase_execution_exception`
+  wrapping `query_shard_exception`, reason `"No mapping found for [x] in
+  order to sort on"`.
+
+  **Known gap, not closed here:** `collapse.inner_hits.sort` and
+  aggregation `top_hits.sort` implement their own sorting independently of
+  `compute_sort_values` and are not covered by this gate — an unresolvable
+  field named in either still sorts silently wrong rather than erroring.
+  Tracked as a follow-up.
+
 ## [1.0.0-rc.18] - 2026-08-18
 
 ### Security
--- a/engine/crates/xerj-api/src/error.rs
+++ b/engine/crates/xerj-api/src/error.rs
@@ -194,6 +194,12 @@ impl ApiError {
             rt.to_string()
         } else if reason.starts_with("failed to create query:") {
             "query_shard_exception".to_string()
+        } else if reason.starts_with("No mapping found for") {
+            // #437: a sort field with no mapping is a shard-level query-
+            // build failure in real ES too, wrapped the same way as
+            // `failed to create query:` above — `query_shard_exception`
+            // in root_cause, outer wrapper unchanged.
+            "query_shard_exception".to_string()
         } else if reason.starts_with("function score query returned an invalid score:") {
             // ES validates function_score results and raises
             // `illegal_argument_exception` with a 400 status when a
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -6475,17 +6475,19 @@ fn parse_sort(sort_val: &Value) -> Vec<xerj_query::sort::SortField> {
                         mode: SortMode::default(),
                         missing: SortMissing::default(),
                         format: None,
+                        unmapped_type: None,
                     },
                 };
                 fields.push(sf);
             }
             Value::Object(obj) => {
                 for (field_name, spec) in obj {
-                    let (order, mode, missing, format) = match spec {
+                    let (order, mode, missing, format, unmapped_type) = match spec {
                         Value::String(ord) => (
                             parse_sort_order(ord),
                             SortMode::default(),
                             SortMissing::default(),
+                            None,
                             None,
                         ),
                         Value::Object(opts) => {
@@ -6509,12 +6511,17 @@ fn parse_sort(sort_val: &Value) -> Vec<xerj_query::sort::SortField> {
                                 .unwrap_or_default();
                             let format =
                                 opts.get("format").and_then(Value::as_str).map(String::from);
-                            (order, mode, missing, format)
+                            let unmapped_type = opts
+                                .get("unmapped_type")
+                                .and_then(Value::as_str)
+                                .map(String::from);
+                            (order, mode, missing, format, unmapped_type)
                         }
                         _ => (
                             SortOrder::Asc,
                             SortMode::default(),
                             SortMissing::default(),
+                            None,
                             None,
                         ),
                     };
@@ -6525,6 +6532,7 @@ fn parse_sort(sort_val: &Value) -> Vec<xerj_query::sort::SortField> {
                             mode,
                             missing,
                             format,
+                            unmapped_type,
                         },
                         "_doc" => SortField {
                             field: "_doc".to_string(),
@@ -6532,6 +6540,7 @@ fn parse_sort(sort_val: &Value) -> Vec<xerj_query::sort::SortField> {
                             mode,
                             missing,
                             format,
+                            unmapped_type,
                         },
                         other => SortField {
                             field: other.to_string(),
@@ -6539,6 +6548,7 @@ fn parse_sort(sort_val: &Value) -> Vec<xerj_query::sort::SortField> {
                             mode,
                             missing,
                             format,
+                            unmapped_type,
                         },
                     };
                     fields.push(sf);
--- a/engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs
+++ b/engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs
@@ -1,0 +1,312 @@
+//! Issue #437, from the outside: sorting on a field the engine cannot
+//! resolve must be rejected with ES's real error shape, on every HTTP entry
+//! point that accepts a `sort` clause — not just `_search`.
+//!
+//! Issue #402's request-level denylist only ever reached `search_impl`, and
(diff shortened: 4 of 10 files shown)


xerj-org:
## Independent verification — head `bf90446c`

Three skeptics, isolated worktrees, rustc 1.97.1 (with the `+fullfp16` gemm-f16 workaround). **All three could not refute: 0 blocking.** Merging on gate (a) green.

- **Gate exemption correct, no over-apply:** `sf.unmapped_type.is_none() && declared_field(...).is_none()` (index.rs:13864) is the only rejection path. Unmapped + `unmapped_type` → admitted, all-missing (200); unmapped without it → 400; typo'd field without it → still 400; mapped field → sorts normally (`unmapped_type` ignored, matching ES). The exemption changes behaviour only for the unmapped+unmapped_type case.

- **21-site struct change clean:** SortField derives no Default so a missed site is a hard compile error (workspace builds); `grep unmapped_type: Some` → 0; `grep .unmapped_type` → only the gate. The two JSON parse paths populate it from the request; everything else is `None`.

- **memory_api legacy list fixed:** `{"stored_at":{...,"unmapped_type":"long"}}` now parses the key, is exempted, returns 200 entries (vs #504's regressed 400).

- **No client break:** `unmapped_type` is `#[serde(default, skip_serializing_if)]` like `format`, and response `sort` values come from computed keys, not the struct, so it's never echoed back.

Nits (non-blocking, not filed): the HTTP test asserts `total==4` rather than `hits.len()==4`/ordering (the engine-level test covers correctness); and a truly-unmapped field + `unmapped_type` resolves via `_source` fallback rather than ES's strict all-missing (pre-existing quirk, documented at the gate, more permissive not less).

---

**Verdict (verbatim):**

> Could not refute PR #514. HEAD confirmed at bf90446. The gate exemption is correct across the full matrix, unmapped_type does NOT over-apply, the 21-site struct change compiles clean with none wrongly defaulted to Some, and both required test suites pass (xerj-engine sort_on_unresolvable_field 3/3; xerj-api sort_on_unresolvable_field_http 6/6 under RUSTFLAGS=-C target-feature=+fullfp16, rustup 1.97.1). Nothing blocking.
> 
> Gate: `engine/crates/xerj-engine/src/index.rs:13864` — `if sf.unmapped_type.is_none() && declared_field(&schema.schema, &sf.field).is_none()`. Short-circuit is correct for every case:
> (a) unmapped field + unmapped_type: first operand false -> whole condition false -> admitted; compute_sort_values falls through to _source lookup and yields null on every doc = all-missing (ES semantics). Verified by test sort_on_an_unmapped_field_with_unmapped_type_is_not_rejected (200, total=4).
> (b) unmapped field WITHOUT unmapped_type: is_none() true, declared_field None -> 400 "No mapping found for [stored_at] in order to sort on". Verified same test (second half) plus search_rejects_sort_on_every_unresolvable_meta_field_and_a_typo (typo'd field still 400s).
> (c) mapped field with or without unmapped_type: declared_field Some -> second operand false -> whole false -> no error, sorts normally. Crucially, `.unmapped_type` is read in EXACTLY ONE place (grep of `\.unmapped_type` across crates returns only index.rs:13864); compute_sort_values never consults it, so on a mapped field unmapped_type is silently ignored (matches ES). Covered by sort_on_a_resolvable_field / keyword-multifield / control-page tests.
> (d) mapped-but-null field: declared_field Some -> sorts (as missing).
> 
> Over-apply check: the exemption ONLY bypasses the "No mapping found" 400; the admitted path is the pre-existing all-missing/_source-fallback path (no new code path, no crash). Worst case it treats all docs as missing = ES behavior. A typo'd field without unmapped_type still 400s (confirmed). So it cannot let a genuinely-malformed sort through.
> 
> 21-site struct change: SortField (sort.rs:22) derives no Default, so every construction must list all fields — a missed site is a hard compile error, and the full workspace built clean. `grep unmapped_type: Some` returns zero -> none wrongly defaulted to Some. serde: `#[serde(default, skip_serializing_if = "Option::is_none")]` -> deserializes to None when absent (safe), omitted from serialization when None.
> 
> Round-trip / ES-compat: unmapped_type is NOT echoed into the client-facing response `sort` values (those come from compute_sort_values key values, not SortField serde), so no client break. memory_api.rs:1210's `{"stored_at":{"order":"desc","unmapped_type":"long"}}` legacy-namespace list now parses the key, is exempted at the gate, and returns 200 entries instead of #504's regressed 400.

**Findings (verbatim):**

- `nit` engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs:349 (introduced_by_this_pr=True): The unmapped_type test asserts only hits.total.value==4; it does not assert hits.hits.len()==4 nor that the sort values are actually all-missing/tie-ordered, so it proves 200-not-400 but not the 'all-missing' ordering claim.

- `nit` engine/crates/xerj-engine/src/index.rs:13864 (introduced_by_this_pr=False): For a truly-unmapped field carrying unmapped_type, XERJ resolves the sort via _source fallback, so if the field name happens to exist as a raw _source key it would sort by that value rather than ES's strict all-missing.

---

**Verdict (verbatim):**

> REFUTATION FAILED — nothing blocking. PR #514 is correct and its evidence holds. The gate exemption `sf.unmapped_type.is_none() && declared_field(...).is_none()` rejects only truly-unresolvable fields WITHOUT unmapped_type (400) and admits those WITH it (200, all-missing), matching ES. Mutation-tested: reverting the `sf.unmapped_type.is_none() &&` clause makes the new test FAIL (400 instead of 200), proving the test is real and the exemption load-bearing. The without-unmapped_type direction still 400s (asserted in the same test), and the engine-level suite (meta fields, typo, resolvable/keyword/nested) stays green — a genuine superset of #504. All 21 SortField construction sites explicitly set unmapped_type:None (no Default spread, no wrongful Some; compile guarantees completeness since SortField has no Default derive). Both parse paths populate the field (proven end-to-end: the 200 result is only reachable if es_compat::parse_sort set Some). unmapped_type carries #[serde(default, skip_serializing_if=Option::is_none)] so it does not leak into response sort-value arrays and cannot break clients. memory_api.rs:1210's legacy list now returns entries rather than 400. The exemption cannot over-apply: it guards a single return Err for undeclared fields only and does not touch any other sort-validation path.

**Findings (verbatim):**

- `nit` engine/crates/xerj-api/tests/sort_on_unresolvable_field_http.rs:349 (introduced_by_this_pr=True): The with-unmapped_type direction asserts status 200 and hits.total.value == 4, but does not assert the hits array length or that the four docs are actually returned in the page.

---

**Verdict (verbatim):**

> Could not demonstrate any blocking finding. The PR correctly threads `unmapped_type` through SortField and both parse paths, and the gate exemption is sound. Nothing should block. All suites green under CI toolchain rustc 1.97.1.
> 
> 21-SITE STRUCT CHANGE (my lens): SortField derives no Default and no construction uses `..Default::default()` / `..SortField` spread, so every one of the struct-literal sites is explicit — a missed field would be a hard compile error, and the crate compiles + tests pass. I grepped all 21 `SortField {` occurrences (3 are the struct def / impl / fn-return-type, the rest are constructions) plus the two constructor methods `score_desc()`/`doc_asc()`: every construction sets `unmapped_type: None` EXCEPT the two JSON parse paths (es_compat::parse_sort, xerj-query::parse_sort_field_spec) which read it from the request object via `.get("unmapped_type").and_then(as_str).map(String::from)`. No site wrongly defaults to Some; no semantic mismatch. es_compat threads unmapped_type onto the `_score`/`_doc` field-name cases too, but those are exempted before the gate anyway, so it is harmless.
> 
> GATE (index.rs:13864): `sf.unmapped_type.is_none() && declared_field(...).is_none()` is the ONLY sort-field rejection path (confirmed by grep of "No mapping found"/"in order to sort"). It rejects an unmapped field WITHOUT unmapped_type (400), admits an unmapped field WITH unmapped_type (falls through to _source lookup → null on every doc → all-missing 200), and never touches mapped fields (declared_field.is_none() is false for them regardless of unmapped_type — same as pre-PR). No over-application: the exemption changes behaviour ONLY for the unmapped+unmapped_type case, which is exactly ES's semantics.
> 
> memory_api.rs:1210: the legacy-namespace list issues {"stored_at":{"order":"desc","unmapped_type":"long"}} through es_compat::search → parse_sort, which now populates unmapped_type, so the gate is skipped and a legacy namespace (stored_at unmapped) returns 200 instead of #504's 400. Verified by logic and by the new HTTP test which pins total=4 docs returned all-missing.
> 
> SERIALIZATION/ROUND-TRIP: unmapped_type carries `#[serde(default, skip_serializing_if="Option::is_none")]`, identical to the pre-existing `format` field. ES response `sort` values are computed key arrays (compute_sort_values), not the SortField struct, so unmapped_type is never emitted back to break a client.
> 
> Only divergence found (NOT blocking, nit-level): a bogus unmapped_type value (e.g. "unmapped_type":"notatype") yields 200 all-missing in XERJ vs 400 in real ES — strictly MORE permissive, does not break any client sending a valid unmapped_type.
> 
> RESULTS: xerj-query --lib 183 passed / 0 failed. xerj-api --profile ci-test --skip script: 33 "test result: ok" blocks, 0 failed; new file sort_on_unresolvable_field_http 6/6 incl. sort_on_an_unmapped_field_with_unmapped_type_is_not_rejected. clippy -p xerj-query -p xerj-engine -p xerj-api --all-targets -D warnings: clean (only the aarch64 `fullfp16` RUSTFLAGS note, an env workaround unrelated to the PR; CI x86-64 unaffected). cargo fmt --check: clean.

---

Issue #488, opened 2026-08-18
xerj feedback --open-pr commits everything staged and pushes it to a public PR, while --help promises "commit ONLY that one file"


xerj-org:
`xerj feedback --open-pr` commits everything already staged in the user's repository, then pushes it to a public PR. Three shipped surfaces promise it commits only the field report.

## The code

`engine/crates/xerj-autoindex/src/feedback.rs:416-434` builds five steps:

```rust
("git add",    vec!["add".into(), relpath.clone()]),
("git commit", vec!["commit".into(), "-m".into(),
                    format!("docs(field-report): {}", draft.slug)]),
("git push",   vec!["push".into(), "-u".into(), "origin".into(), branch.clone()]),
```

`git add <relpath>` stages the report. `git commit -m …` then carries **no pathspec and no `--only`**, so it commits the whole index — the report plus whatever the user had already staged. `git push` sends it to the branch the tool opens the PR from.

## What ships alongside it

- `feedback.rs:520` — `xerj feedback --open-pr [OPTIONS]    commit ONLY the report + open a PR`
- `feedback.rs:543` — `--open-pr            create branch field-report/<slug>, commit ONLY that one file,`
- `landing/llms.txt:68` — ``  `xerj feedback --open-pr` commits **only** `user-feedback/16-agent-field-reports/<date>-<slug>.md` and opens the PR.``

All three are wrong, and two of them are the binary's own `--help`. `CHANGELOG.md` is the only surface that describes the real behaviour, and a CHANGELOG is not what someone reads before running a command.

## Why this is worse than an ordinary doc/code mismatch

The failure is silent, public, and irreversible in the way that matters. Nothing warns, nothing fails, the exit code is 0, and the PR is open before the user sees what went into it. Anything staged goes public: a `.env` a developer had staged, a credentials file, an unrelated in-progress change, a customer file. Deleting the branch afterwards does not unpublish it — GitHub keeps the objects reachable and forks/caches may already have them.

Two properties make it likely rather than theoretical:

1. **`xerj feedback` is the community CTA.** README and `llms.txt` present it as the one-command way to contribute, aimed explicitly at coding agents. An agent working in a repo it did not clean first is the expected caller, not the edge case.
2. **The promise invites the risk.** "commits ONLY that one file" is exactly the sentence that tells a careful user they do not need to check `git status` first. The docs actively remove the habit that would catch this.

Anyone who ran `git add -A` before deciding to file a report gets the whole working tree in a public PR.

[...]

PR #517, opened 2026-08-19
fix(feedback): --open-pr commits ONLY the field report, not the whole staged index (#484)

### LINKED ISSUE #484: xerj feedback --open-pr promises to commit ONLY the field report, but a bare git commit publishes anything already staged
`xerj feedback --open-pr` tells the user, in four separate places, that it commits only the field report. It does not. It stages the report and then runs an unrestricted `git commit`, so **whatever the user already had staged is committed and pushed to a public PR alongside it**.

## The promise

```
engine/crates/xerj-autoindex/src/feedback.rs:387   /// `--open-pr`: commit ONLY the field report on a new branch and run
engine/crates/xerj-autoindex/src/feedback.rs:520   xerj feedback --open-pr [OPTIONS]    commit ONLY the report + open a PR
engine/crates/xerj-autoindex/src/feedback.rs:543   --open-pr   create branch field-report/<slug>, commit ONLY that one file,
landing/llms.txt:68                                commits **only** user-feedback/16-agent-field-reports/<date>-<slug>.md
README.md:66                                       commits **just that one file** and opens the PR
```

The `--dry-run` banner says it too (`feedback.rs:353`): `# Ready-to-run — files ONLY the field report, nothing else:`.

## The code

`engine/crates/xerj-autoindex/src/feedback.rs:422-430`:

```rust
("git add",    vec!["add".into(), relpath.clone()]),
("git commit", vec!["commit".into(), "-m".into(), msg, ...]),   // no pathspec, no --only
```

`git add <report>` **adds to** the index; it does not replace it. The following `git commit` has no pathspec and no `--only`, so it commits the whole index.

## Reproduction

Verified in review with the release binary in a throwaway repo (stubbed `gh`, no remote):

```
===== git status BEFORE =====
A  .env
A  src.rs
===== files in the commit it created =====
 .env    | 1 +
 src.rs  | ...
 user-feedback/16-agent-field-reports/<date>-<slug>.md | ...
```

## Why this matters more than a wording bug

The audience is the problem. This command exists to recruit **agents** into filing field reports, and the project asks them to run it at the end of a session — exactly when a working tree is most likely to have staged work. An agent that has staged a `.env`, a credentials file, or a customer's source and then runs the one command the docs tell it to run will open a **public PR** containing it, having been told four times that only the report is committed.

It is also CLA-exempt, so the PR merges without the friction that might otherwise surface it.

## Fix

Restrict the commit to the pathspec:

```rust
("git commit", vec!["commit".into(), "--only".into(), relpath.clone(), "-m".into(), msg, ...])
```

`--only` commits exactly the named path regardless of index state, which is what every one of the four messages already promises. Worth a test that stages an unrelated file and asserts it is absent from the resulting commit.

Found while verifying the rc.18 release notes (#460); the notes have been corrected to describe the current behaviour, but the code and the four user-facing strings should be made to match the promise instead.

xerj-org:
Closes #484 (and #488, its duplicate).

## The leak
`xerj feedback --open-pr` did `git add <report>` then a **bare** `git commit -m …` (no pathspec), so anything else already staged was committed and **pushed to the public PR** the tool opens. Its own `--help` and doc promise "commit ONLY that one file". An agent in a repo with a staged `.env`/credential/unrelated change would publish it, silently, exit 0, before seeing the diff. This is the community CTA aimed at coding agents, so a dirty index is the expected caller.

## Fix
`git commit --only <report>` — exactly the report is committed; everything else in the index is left untouched. The printed fallback commands (`pr_commands`, shown when `gh` is absent) had the **same** bare commit under a "files ONLY the field report" header — fixed there too, or the leak just moves to the copy-paste path.

## Test (behavioral, fail-before proven)
New `field_report_commit_argv` helper + a test on a real scratch repo: with a decoy `SECRET.env` staged, HEAD's tree contains the report and **not** the decoy, and the decoy stays staged. Reverting the helper to the bare commit makes it fail with `SECRET.env` in the committed tree.

Verified rustc 1.97.1: feedback **13/13** both profiles, clippy -D warnings clean, fmt clean.

### DIFF
--- a/engine/crates/xerj-autoindex/src/feedback.rs
+++ b/engine/crates/xerj-autoindex/src/feedback.rs
@@ -357,7 +357,7 @@ pub fn pr_commands(d: &Draft) -> String {
          \x20 -o {relpath}\n\
          git checkout -b {branch}\n\
          git add {relpath}\n\
-         git commit -m \"docs(field-report): {slug}\"\n\
+         git commit --only {relpath} -m \"docs(field-report): {slug}\"\n\
          git push -u origin {branch}\n\
          gh pr create --base main --head {branch} \\\n\
          \x20 --title \"Agent field report: {title}\" \\\n\
@@ -382,6 +382,23 @@ fn write_report_file(path: &Path, report: &str) -> io::Result<()> {
         }
     }
     std::fs::write(path, report)
+}
+
+/// The argv for the field-report commit. It commits ONLY the report path via
+/// `git commit --only <relpath>`, so anything a caller happened to have staged
+/// is left in the index and never reaches the public PR. A bare `git commit -m`
+/// committed the whole index instead — the field report's own `--open-pr`
+/// promise ("commit ONLY that one file") was false, and an agent working in a
+/// dirty repo would publish whatever else was staged (a `.env`, a credential, an
+/// unrelated change). #484.
+fn field_report_commit_argv(relpath: &str, slug: &str) -> Vec<String> {
+    vec![
+        "commit".into(),
+        "--only".into(),
+        relpath.into(),
+        "-m".into(),
+        format!("docs(field-report): {slug}"),
+    ]
 }
 
 /// `--open-pr`: commit ONLY the field report on a new branch and run
@@ -422,11 +439,7 @@ fn open_pr(draft: &Draft, report: &str, commands: &str) -> i32 {
         ("git add", vec!["add".into(), relpath.clone()]),
         (
             "git commit",
-            vec![
-                "commit".into(),
-                "-m".into(),
-                format!("docs(field-report): {}", draft.slug),
-            ],
+            field_report_commit_argv(&relpath, &draft.slug),
         ),
         (
             "git push",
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent verification — head `ddb1c41e`

Three skeptics, isolated worktrees, rustc 1.97.1 (`+fullfp16` for the gemm-f16 host dep). **All three refuted=false, 0 blocking, 0 findings.** Merging.

- **Leak closed, both paths:** all three reproduced in real scratch repos — `git commit --only <report>` commits ONLY the report and leaves a staged `SECRET.env` decoy out of the commit while keeping it in the index (not dropped — verified `git diff --cached` still shows it, so no data-loss the other way). Holds for the nested `user-feedback/16-agent-field-reports/<date>-<slug>.md` path. The printed `pr_commands` fallback carries the same `git commit --only {relpath}` and copy-pastes correctly.

- **Normal flow intact:** only-report-staged still commits (exit 0); `open_pr`'s checkout/write/add/commit --only/push/PR sequence unchanged; the `--only` "no changes" error can't occur in the real flow (a fresh report is written on a new branch first).

- **Test genuine + bidirectional:** it drives the actual `field_report_commit_argv` and asserts both that the decoy is NOT committed AND that it stays staged. Fail-before verified independently — reverting the helper to a bare commit makes it fail with `SECRET.env` in the committed tree.

- **No injection:** slugs are slugified ASCII `[a-z0-9-]`, and args pass as argv to `Command` (never a shell). 680 tests both profiles, clippy `-D warnings` clean, fmt clean. Only `feedback.rs` changed.

---

**Verdict (verbatim):**

> Could not refute PR #517. The fix is correct and I found no blocking issue. Reproduced in real scratch git repos: `git commit --only <report>` (from field_report_commit_argv) commits ONLY the field report and leaves any other staged file — the SECRET.env decoy — out of the commit while keeping it staged in the index (not dropped). This holds for the nested report path user-feedback/16-agent-field-reports/<date>-<slug>.md. So the #484 leak is genuinely closed: nothing else in a dirty index reaches the public PR. The printed fallback (pr_commands) was updated to the same `git commit --only {relpath}` line and copy-pastes correctly (slug is slugified, so relpath has no spaces/special chars; and the in-tool argv is passed as a Vec to Command, never a shell, so no injection). Normal case (only the report staged) still commits with exit 0, so the happy path is intact. The report itself is committed. The test is genuine and asserts BOTH properties (decoy-not-committed AND decoy-still-staged), and I verified fail-before by reverting the helper — the test then fails with SECRET.env committed. The one edge case, report already committed with no changes: `git commit --only` exits 1 and open_pr aborts to the fallback commands — but it commits nothing and leaves the decoy staged (no leak, no data loss), and the old bare-commit code would have leaked a staged decoy in that same scenario, so this is strictly safer, not a regression. Acceptable. Nothing blocking.

---

**Verdict (verbatim):**

> PR #517 holds up. I could not demonstrate any blocking defect. The #484 leak is genuinely fixed on both paths, the guard test is real and bidirectional, and the normal flow is intact.
> 
> Evidence:
> - Core behavior reproduced in a real scratch git repo with the exact argv from field_report_commit_argv and a nested report path (user-feedback/16-agent-field-reports/2026-08-19-my-report.md): the resulting HEAD tree contains the report and NOT the staged SECRET.env decoy, and the decoy remains in the index (git ls-files --stage and git diff --cached --name-only both still show SECRET.env). No data loss — the staged decoy is left untouched, not dropped.
> - Fail-before verified: reverting ONLY the helper back to a bare `git commit -m` (keeping the test) makes the commit include the decoy ("2 files changed ... create mode 100644 SECRET.env") and the test FAILS at feedback.rs:868 ("a staged decoy must NOT be committed by --open-pr (#484)"). So the test genuinely guards the shipped code path.
> - The test drives the ACTUAL builder field_report_commit_argv (not a hand-written commit) and asserts BOTH directions: report IS committed, decoy NOT committed, AND decoy still staged (covers the data-loss direction).
> - The printed fallback pr_commands (feedback.rs:360) is also fixed to `git commit --only {relpath} -m ...` under the "files ONLY the field report" header; the copy-paste git add + git commit --only sequence still works.
> - Normal flow: full feedback suite 13/13 pass on rustc 1.97.1.
> - No-changes edge case: rerunning `--only` on an unchanged report exits 1 but leaves the index intact (decoy still staged); it cannot occur in the real open_pr flow, which writes a fresh report on a newly created branch (checkout -b runs first).
> - Special-char edge case is a non-issue: slugify emits pure ASCII [a-z0-9-], so no spaces/special chars ever reach the relpath in either the executed argv or the printed command.

---

**Verdict (verbatim):**

> REFUTATION FAILED — nothing blocking. HEAD is ddb1c41 as required. The fix is correct and the #484 leak is genuinely closed. `git commit --only <report>` was reproduced in real scratch repos: the resulting HEAD tree contains ONLY the field report and NOT the staged SECRET.env decoy, and the decoy is left staged in the index (not committed, not dropped — no data loss). This holds for the real two-level nested path user-feedback/16-agent-field-reports/2026-08-19-my-report.md. Normal flow intact: only-report-staged still commits; open_pr's checkout -b / write / git add / commit --only / push / gh pr create sequence is unchanged in structure and each step preserved. The "no changes" error of --only cannot occur in open_pr because the report is freshly written before add/commit. The printed pr_commands fallback now carries `git commit --only {relpath}` matching behaviour; the full block runs as a valid copy-paste sequence (git add then commit --only is redundant but correct; decoy left staged) and every other command in the block is consistent. Slugs are slugified ASCII (no spaces/special chars) and args are passed as argv, so no quoting hazard. The new test is genuine: reverting the helper to a bare commit makes it FAIL with SECRET.env in the committed tree, and it asserts BOTH that the decoy is not committed AND that it stays staged. Only feedback.rs changed. Suite: dev 680 passed / ci-test 680 passed; clippy clean (the sole warning is the aarch64-host fullfp16 RUSTFLAGS codegen note, not a lint; CI x86-64 unaffected); fmt clean.

---

PR #452, opened 2026-08-17
fix(docs): the install page verified a checksum without ever hashing the archive

xerj-org:
Closes #444.

## The defect

`landing/docs/install.html`'s manual-download block ran:

```
sha256sum -c xerj-${ver}-x86_64-unknown-linux-musl.tar.gz.sha256
```

`sha256sum -c` reports success for whatever filenames the `.sha256` lists. It skips `#` comment lines silently, exits `0` if any one line verified, and never checks that a line names the archive you are about to extract.

Reproduced with a `.sha256` holding a comment naming the archive plus a valid digest for an unrelated file:

```
$ cat xerj-1.0.0-x86_64-unknown-linux-musl.tar.gz.sha256
# xerj-1.0.0-x86_64-unknown-linux-musl.tar.gz: OK
437a25cd…  other.txt

$ sha256sum -c xerj-1.0.0-x86_64-unknown-linux-musl.tar.gz.sha256
other.txt: OK
EXIT=0                     <- page proceeds to `tar xzf` a backdoored archive
```

The archive's real digest is never computed. The paragraph directly above the block claimed verification is **fail-closed**.

## The fix

Both copies compute the digest and demand the `.sha256` contain that exact line — the form that survived #441's adversarial review. Extraction is *chained to* the check rather than following it, so a refusal cannot be stepped over.

Old vs new, across five `.sha256` shapes:

| scenario | old (`sha256sum -c`) | new |
|---|---|---|
| decoy: comment + digest of another file | `rc=0` → **proceeds** | refused, nothing extracted |
| genuine | `rc=0` | verified, extracted |
| binary mode (`digest *asset`) | `rc=0` | verified, extracted |
| CRLF line endings | `rc=0` | verified, extracted |
| archive actually tampered | `rc!=0` | refused |

End to end against the **real published pair**: rc.17 aarch64 verifies against `1f3a0a5bbb83…` and the extracted binary reports `xerj v1.0.0-rc.17`.

## Scope beyond the issue

`landing/llms.txt` carried the same block. Not in the issue's scope, but it is the copy AI agents read, and its own comment said *"do not proceed if this fails"* — exactly what the shipped form failed to enforce. Fixed.

## Two things the issue listed as unverified, now checked — neither is affected

[...]

Issue #444, opened 2026-08-17
landing/docs/install.html verifies a downloaded .sha256 without asserting it names the archive, under a paragraph claiming "Verification is fail-closed"


xerj-org:
Found by an independent verification of #441, which fixes the same defect in the air-gapped recipe. This copy is on the page most users actually follow and is **not** covered by that PR.

## What happens

The manual-download block runs:

```
curl -fsSLO https://.../xerj-<v>-<target>.tar.gz.sha256
sha256sum -c xerj-<v>-<target>.tar.gz.sha256
```

`sha256sum -c` reports success for whatever filenames the `.sha256` lists, and exits 0 as long as at least one properly-formatted line verified. It:

* silently ignores lines beginning with `#` — no warning at all
* warns but still exits 0 on unparseable lines
* never checks that any line names the archive you are about to extract

So a `.sha256` holding a valid digest for some other file plus a comment line mentioning the archive verifies clean, and nothing in the block notices that the archive itself was never checked.

The paragraph directly above the block asserts **"Verification is fail-closed"**. It is not.

## Also in scope

`docs/EXPERIMENTAL_ONNX.md:42` carries a bare `sha256sum` block with the same shape.

## Why it is filed separately

#441 fixes `docs/recipes/air-gapped-deployment.md` and its published copy. Widening it to the install page during an active verification round is how that PR reached three refuted rounds. The fix here should reuse whatever digest-assertion form survives #441's current adversarial pass, rather than inventing a second one.

## Not verified

Whether the same pattern appears in the PowerShell installer (`landing/get.ps1`) or in the Homebrew/Docker paths.

PR #519, opened 2026-08-19
docs(landing): verify release downloads on upgrades and demo pages (#454)

### LINKED ISSUE #454: Two more shipped pages download and extract a release with no checksum step at all (upgrades.html, demo/index.html)
Found by the independent verification of #452, which fixed the *broken* verification on `landing/docs/install.html` and `landing/llms.txt`. These two are a different shape — not a verification that fails open, but **no verification at all** — so they were outside that PR's scope and are filed here rather than widened into it.

## 1. `landing/docs/upgrades.html` — upgrade over a running node, unverified

The page has **no `sha256` anywhere** (`grep -c sha256` → `0`). Its upgrade procedure downloads a release archive and goes straight to:

```
$ tar xzf xerj-0.2.0-linux-x86_64.tar.gz
```

This is the worst placement of the three: it is an upgrade path, so the binary being replaced is one already trusted and running, and the page is what an operator follows during a maintenance window.

The archive name is also stale — `xerj-0.2.0-linux-x86_64.tar.gz` is neither the current version nor the published target-triple naming (`xerj-<version>-<target>.tar.gz`, e.g. `xerj-1.0.0-rc.17-x86_64-unknown-linux-musl.tar.gz`), so the commands as printed cannot work against any current release.

## 2. `landing/demo/index.html` — pinned to `v1.0.0-rc.1`, unverified

```
$ curl -L -o xerj.tar.gz https://github.com/xerj-org/xerj/releases/download/v1.0.0-rc.1/xerj-1.0.0-rc.1-x86_64-unknown-linux-gnu.tar.gz
$ tar xzf xerj.tar.gz
```

No digest, and pinned to `rc.1` while `rc.17` is current. Note `-L` follows redirects with no verification of where it lands.

## Suggested fix

Reuse whatever digest-assertion form lands in #452, rather than inventing a third: compute the digest, assert it is 64 hex characters, demand the `.sha256` contain that exact line, and chain extraction and install to that check. Do **not** use `sha256sum -c` — see #444 for why.

## Not verified

Whether any other page under `landing/` fetches a release without verifying it. The sweep that found these two covered docs pages, llms files and the demo; it did not cover case studies, playbooks or brandbook assets.

xerj-org:
Closes #454.

Two shipped pages fetched a release archive and unpacked it with **no integrity check at all** — the shape #452 explicitly left out of scope because it fixed a verification that *fails open*, not the absence of one.

### What was wrong
- **`landing/docs/upgrades.html`** — an upgrade **over a running node** (the binary being replaced is already trusted and serving). `grep -c sha256` → `0`. It also pulled `xerj-0.2.0-linux-x86_64.tar.gz` off `xerj.org/releases` — a name and host no current release publishes, so the commands could not work as printed.
- **`landing/demo/index.html`** — pinned to the captured `v1.0.0-rc.1` build, `curl -L` following redirects to wherever, no digest.

### The fix
Reuses the fail-closed form already shipped on `install.html` (#452) verbatim, rather than inventing a third:
1. fetch the archive **and** its published `.sha256`;
2. compute the digest and **length/hex-check it** — a host with no `sha256sum`/`shasum` leaves `want` empty, and an empty digest cannot satisfy the check;
3. require the `.sha256` to contain **that exact line** (both GNU two-space and `-b` binary forms; `tr -d '\r'` for Windows-written files) — **not** `sha256sum -c`, see #444 for why;
4. **extraction is chained to the check** inside a subshell, so an archive that fails to verify is never unpacked and a refusal exits non-zero (a Dockerfile `RUN` / CI step / agent sees the failure).

- The **demo stays on rc.1** — its transcript really ran on that build (the constants guard exempts it as a dated capture) — and verifies against rc.1's own published `.sha256`.
- **upgrades.html** moves to the real GitHub release URL + target-triple naming with an operator-set `ver` (example: the current `1.0.0-rc.18`), so the printed commands actually resolve.

### Verification
- **Functional test of the exact snippet** against the real published rc.1 pair: the good archive verifies and extracts (exit 0); a `.sha256` with **one flipped hex digit** is rejected with `CHECKSUM MISMATCH` (exit 1); an empty digest (no hashing tool) is refused (exit 1).
- rc.18 musl `.tar.gz` **and** `.tar.gz.sha256` both exist, so upgrades.html's example resolves.
- `landing-constants-guard.sh` — all checks pass (version stamps match `v1.0.0-rc.18`; demo rc.1 exempt).
- Swept all of `landing/` (html + txt, incl. case studies / playbooks): **no other page** fetches a release without verifying it — closes the issue's "Not verified" note.

Docs-only; no engine code touched.

### DIFF
--- a/landing/demo/index.html
+++ b/landing/demo/index.html
@@ -351,7 +351,8 @@ footer .links { display: flex; gap: 28px; flex-wrap: wrap; }
 
   <pre class="cmd"><span class="prompt">$ </span>(
     ver=1.0.0-rc.1                              # this demo was captured on rc.1
-    asset="xerj-${ver}-x86_64-unknown-linux-gnu.tar.gz"
+    stage="xerj-${ver}-x86_64-unknown-linux-gnu"
+    asset="${stage}.tar.gz"
     base="https://github.com/xerj-org/xerj/releases/download/v${ver}"
 
     curl -fsSLO "$base/$asset"
@@ -365,7 +366,7 @@ footer .links { display: flex; gap: 28px; flex-wrap: wrap; }
       | LC_ALL=C grep -qxF -e "$want  $asset" -e "$want *$asset" \
       || { echo "CHECKSUM MISMATCH for $asset — do not run it" &gt;&amp;2; exit 1; }
 
-    tar xzf "$asset"
+    tar xzf "$asset" &amp;&amp; mv "$stage/xerj" ./xerj
   )
 <span class="prompt">$ </span>./xerj --insecure --data-dir ./data
 <span class="out">2026-04-27T19:11:41.6Z  INFO xerj: xerj v1.0.0-rc.1 starting
--- a/landing/demo/index.html
+++ b/landing/demo/index.html
@@ -349,8 +349,24 @@ footer .links { display: flex; gap: 28px; flex-wrap: wrap; }
     is already running on port 9200.
   </p>
 
-  <pre class="cmd"><span class="prompt">$ </span>curl -L -o xerj.tar.gz https://github.com/xerj-org/xerj/releases/download/v1.0.0-rc.1/xerj-1.0.0-rc.1-x86_64-unknown-linux-gnu.tar.gz
-<span class="prompt">$ </span>tar xzf xerj.tar.gz
+  <pre class="cmd"><span class="prompt">$ </span>(
+    ver=1.0.0-rc.1                              # this demo was captured on rc.1
+    asset="xerj-${ver}-x86_64-unknown-linux-gnu.tar.gz"
+    base="https://github.com/xerj-org/xerj/releases/download/v${ver}"
+
+    curl -fsSLO "$base/$asset"
+    curl -fsSLO "$base/$asset.sha256"
+
+    want=$( { sha256sum "$asset" 2&gt;/dev/null || shasum -a 256 "$asset"; } | cut -d ' ' -f 1 )
+    printf %s "$want" | LC_ALL=C grep -qE '^[0-9a-f]{64}$' \
+      || { echo "no working SHA-256 tool — refusing to run an unverified $asset" &gt;&amp;2; exit 1; }
+
+    tr -d '\r' &lt; "$asset.sha256" \
+      | LC_ALL=C grep -qxF -e "$want  $asset" -e "$want *$asset" \
+      || { echo "CHECKSUM MISMATCH for $asset — do not run it" &gt;&amp;2; exit 1; }
+
+    tar xzf "$asset"
+  )
 <span class="prompt">$ </span>./xerj --insecure --data-dir ./data
 <span class="out">2026-04-27T19:11:41.6Z  INFO xerj: xerj v1.0.0-rc.1 starting
 2026-04-27T19:11:41.6Z  WARN xerj: --insecure: TLS and auth disabled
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent adversarial verification — CLEAN (0 blocking)

Three skeptics, each tasked to **refute** one property with real shell execution against the actual published archives. All three `refuted:false`, `blocking:false`.

**1. security-bypass** — *Could not refute.* Ran 15 attack cases against the exact de-escaped gate with real archives A(real)/B(malicious):
> a_comment_bypass_B rc=1 [none]; b_correctdigest_wrongname_B rc=1 [none]; d_whitespace_B rc=1 [none]; e_empty_B rc=1 [none]; f_nohash_B rc=1 [none]; f2_nohash_blankline rc=1 [none]; g_BdigestInA_serveA rc=1 [none]; g2_Adigest_serveB rc=1 [none]; x2_partial_line_B rc=1 [none]. Every wrong-bytes/unhashed attack failed closed. Extractions occurred ONLY on genuine matches: baseline_ok_A, baseline_star_A, c_crlf_ok_A (tr strips CR), d2_blank_ok_A, x_multi_hasmatch_A. In the no-tool case want='' fails the `^[0-9a-f]{64}$` hex check before the .sha256 is consulted. `grep -qxF` (whole-line, fixed-string) defeats the `sha256sum -c` comment/other-file bypass, wrong-filename, whitespace-padding, and partial-line attacks. Invariant confirmed: `tar xzf` is reachable only when the .sha256 contains a line exactly equal to the real hash of the served bytes; no path extracts bytes whose real digest is absent, and none extracts an un-hashed archive.

**2. copy-paste / fail-closed** — *Could not refute.* Tested against the real published pairs in **bash and dash**:
> (1) HAPPY PATH runs verbatim & reaches tar, exit 0 in BOTH shells (demo rc.1-gnu; upgrades rc.18-musl, 126747824 bytes). No HTML entity survived de-escape. (2) FAIL-CLOSED: tampered .sha256 -> 'CHECKSUM MISMATCH', tamper_exit=1 in both, extracted dir ABSENT (no tar). (3) PARENT-SHELL SURVIVAL: piped the FAILING block + a follow-up marker into `bash -il` and `dash -i`; both printed the mismatch then 'PARENT_LOGIN_SHELL_STILL_ALIVE marker 1' — the `( ... exit 1 )` subshell confined the exit, login shell survived with $?=1. upgrades `sudo install ".../xerj"` path matches the extracted subdir exactly.

Non-blocking note it raised — the demo's `./xerj` line assumed a top-level binary that the tarball does not produce — is now **fixed in this PR** (`mv "$stage/xerj" ./xerj` after the gate), so the whole walkthrough is runnable.

**3. claims / coverage** — *Could not refute.* All 4 referenced assets → HTTP 200 (`.tar.gz` and `.tar.gz.sha256` for rc.18-musl and rc.1-gnu), each redirecting to release-assets with a matching content-disposition filename. Demo transcript still truthful to its rc.1 capture (no silent version bump). `landing-constants-guard.sh` → all checks passed, exit 0. Whole-tree sweep (html+txt incl. case-studies/playbooks/brandbook): every file that extracts a release {upgrades, demo, install, air-gapped-deployment, llms.txt} contains AND uses real digest-match verification before `tar`. Zero unverified fetches remain.
> Out-of-scope drift it noted: `llms.txt` stamps "Current release: v1.0.0-rc.17" (vs rc.18). Different file, not #454 — filing a separate follow-up.

Exit condition: **no BLOCKING findings** → verification gate satisfied. Merging once CI is green and mergeStateStatus is CLEAN.

PR #529, opened 2026-08-20
fix(engine): 400 a knn/semantic clause on a non-vector field, not empty 200 (#498)

### LINKED ISSUE #498: knn against an index with no dense_vector field returns 0 hits / HTTP 200 with no error (silent false-negative)
A `knn` query against an index that has no `dense_vector` field returns **HTTP 200 with `hits.total = 0` and no error**. The caller cannot tell "nothing matched" from "this index cannot answer vector queries at all."

## Reproduce (refsym, the reference-coding corpus)

```
$ curl -s -o /dev/null -w "http=%{http_code}\n" localhost:9200/refsym/_search \
    -d '{"knn":{"field":"embedding","query_vector":[0.1,0.2,0.3],"k":5,"num_candidates":50}}'
http=200

$ curl -s localhost:9200/refsym/_search -H 'content-type: application/json' \
    -d '{"knn":{"field":"embedding","query_vector":[0.1,0.2,0.3],"k":5,"num_candidates":50}}' \
    | jq '.hits.total, .error'
{"value":0,"relation":"eq"}
null
```

`refsym`'s mapping has no `dense_vector` field (fields are `code`, `doc`, `name`, `line`, `sig`, ...), so the `knn` clause has nothing to run against. Instead of telling the caller, it returns an empty, successful result.

## Why it matters

This is the same silent-false-negative class the project's own WordPress verification doc calls out. An agent that issues a `knn`/`semantic` query against an index built without vectors gets zero hits and concludes "no relevant code exists," when the truth is "this index does not do vector search." It fails the exact use case the vector and semantic badges advertise, and it fails it invisibly.

It surfaced in a benchmark of XERJ's own reference-coding claim: the retrieval arm's vector queries returned nothing, silently, and the agent could not know why.

## Expected

Elasticsearch returns `400` with `query_shard_exception` / `illegal_argument_exception` ("field [embedding] does not exist in the mapping" or "field is not of type dense_vector") for a `knn` against a missing or wrongly-typed field. XERJ should do the same: a `knn` or `semantic` clause naming a field that is absent, or not a `dense_vector` / `semantic_text`, must fail loudly with a 400 that names the field and the reason, not return an empty 200.

## Scope to check while fixing

- `semantic` queries against an index whose `semantic_text` field was never populated with vectors (same silent-empty risk).
- Whether the default embedder being lexical (feature-hashing) rather than neural changes what "has vectors" means for a `semantic_text` field, and whether that path also returns silently.

Found during the reference-coding benchmark investigation.

xerj-org:
Closes #498.

A `knn` (or `semantic`) clause naming a field that is **absent**, or is **not vector-capable**, returned `HTTP 200` with `hits.total = 0` and no error. An agent issuing a vector query against an index built without vectors got zero hits and concluded *"no relevant results exist"* — when the truth is *"this index cannot answer vector queries at all."* A silent false-negative on the exact use case the vector/semantic badges advertise. It surfaced in the reference-coding benchmark: the retrieval arm's vector queries returned nothing, silently. Real ES answers `400 illegal_argument_exception` here.

### Fix
`vector_query_field_error` — a pre-execution walk of the resolved query (mirrors `unsearchable_query_field`, sibling to the #437 sort gate) that returns a 400 reason when a `Knn`/`SemanticSearch` clause names a field that does not resolve, or resolves to something that cannot answer a vector clause:
```
[knn] query targets field [embedding], which does not exist in the mapping
```
**"Vector-capable" is deliberately permissive** — `dense_vector`, `chunk`, or any field with an embedding config (`semantic_text`, whose companion vector exists even under the lexical feature-hash embedder) — so a legitimate vector query is **never** rejected; only a plainly non-vector or absent field is. Covers `knn` and `semantic`, recursing through `bool`/`hybrid`/`dis_max`/`boosting`/nested wrappers.

### Verification
- New HTTP test `knn_on_non_vector_field_fails.rs`:
  - `knn` on an absent field now **400s** and names the field — **proven fail-before** by reverting only the source hunk: it returns `200` with `hits.total: 0` (the exact bug).
  - `knn` on a genuine (empty) `dense_vector` index is **not** rejected — the false-positive guard, so real vector queries keep working.
- Under **rustc 1.97.1**: `fmt`, `clippy --workspace --all-targets -D warnings`, `build --workspace --profile ci-test` all clean; tests pass under **both** dev and ci-test.

Scope note: `semantic` is covered symmetrically. The lexical-vs-neural distinction does not change validity here — a properly-mapped `semantic_text` field has a companion vector under either embedder, so it is accepted; the check is purely schema-shaped.

### DIFF
--- a/engine/crates/xerj-api/tests/knn_on_non_vector_field_fails.rs
+++ b/engine/crates/xerj-api/tests/knn_on_non_vector_field_fails.rs
@@ -1,0 +1,146 @@
+//! Issue #498: a `knn` / `semantic` clause that names a field which is absent,
+//! or is not a vector field, must fail loudly with a 400 — not return an empty,
+//! successful 200. An agent that issues a vector query against an index built
+//! without vectors otherwise gets zero hits and concludes "no relevant results
+//! exist", when the truth is "this index cannot answer vector queries at all".
+//!
+//! The second test is the false-positive guard: a `knn` against a genuine
+//! `dense_vector` field must NOT be rejected, even when the index is empty (an
+//! empty result there is a legitimate "nothing matched").
+//!
+//! Elasticsearch is referenced for wire semantics only; no ES code is
+//! reproduced here.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+fn config_for(dir: &std::path::Path) -> xerj_common::config::Config {
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    config
+}
+
+fn app_over(dir: &std::path::Path) -> axum::Router {
+    let config = config_for(dir);
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    xerj_api::router::build_es_compat_router(state)
+}
+
+async fn send(app: &axum::Router, req: Request<Body>) -> (StatusCode, Value) {
+    let response = app.clone().oneshot(req).await.expect("response");
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .expect("body");
+    let body = if bytes.is_empty() {
+        Value::Null
+    } else {
+        serde_json::from_slice(&bytes).unwrap_or(Value::Null)
+    };
+    (status, body)
+}
+
+fn json_req(method: &str, path: &str, body: Value) -> Request<Body> {
+    Request::builder()
+        .method(method)
+        .uri(path)
+        .header("content-type", "application/json")
+        .body(Body::from(body.to_string()))
+        .expect("request")
+}
+
+#[tokio::test]
+async fn knn_on_an_absent_field_fails_with_400_not_empty_200() {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let app = app_over(dir.path());
+
+    // A text-only index — no `dense_vector` field anywhere.
+    let (st, body) = send(
+        &app,
+        json_req(
+            "PUT",
+            "/docs",
+            json!({ "mappings": { "properties": { "body": { "type": "text" } } } }),
+        ),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "create index: {body}");
+
+    let (st, _) = send(
+        &app,
+        json_req(
+            "POST",
+            "/docs/_doc?refresh=true",
+            json!({ "body": "hello world" }),
+        ),
+    )
+    .await;
+    assert!(st.is_success(), "index a document: {st}");
+
+    // A `knn` naming a field that does not exist must 400, not answer 200/0.
+    let (st, body) = send(
+        &app,
+        json_req(
+            "POST",
+            "/docs/_search",
+            json!({ "knn": { "field": "embedding", "query_vector": [0.1, 0.2, 0.3],
+                             "k": 5, "num_candidates": 50 } }),
+        ),
+    )
+    .await;
+    assert_eq!(
+        st,
+        StatusCode::BAD_REQUEST,
+        "knn on an absent field must fail loudly, not return an empty 200: {body}"
+    );
+    let reason = body
+        .pointer("/error/reason")
+        .and_then(Value::as_str)
+        .unwrap_or("");
+    assert!(
+        reason.contains("embedding"),
+        "the 400 must name the offending field: {body}"
+    );
+}
+
+#[tokio::test]
+async fn knn_on_a_real_dense_vector_field_is_not_rejected() {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let app = app_over(dir.path());
+
+    let (st, body) = send(
+        &app,
+        json_req(
+            "PUT",
+            "/vecs",
+            json!({ "mappings": { "properties": {
+                "vec": { "type": "dense_vector", "dims": 3 } } } }),
+        ),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "create dense_vector index: {body}");
+
+    // No documents ingested: an empty result here is a legitimate "nothing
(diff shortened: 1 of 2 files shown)


xerj-org:

3 skeptics in worktrees under 1.97.1. All `refuted:false, blocking:false`. Verdicts verbatim:

**false-positive-regression — refuted:false, blocking:false.**

> vector_capable = matches!(field_type, Vector|Chunk) || embedding.is_some(), and the only rejection sites are the two leaf arms (Knn/SemanticSearch) — every container arm merely recurses and can never self-reject. Verified each legit shape resolves to a vector-capable FieldConfig: dense_vector -> FieldType::Vector (es_compat.rs:16229); semantic_text field -> FieldType::Text but embedding=Some (es_compat.rs:16143), so embedding.is_some() accepts it; its auto-created companion body_vector -> FieldType::Vector (es_compat.rs:16187); Chunk accepted by matches!. Empirically ran probes under rustc 1.97.1: knn on dense_vector, semantic on semantic_text, knn on the semantic_text field itself, knn on the companion body_vector, nested-query knn on dotted nested.vector (declared_field recurses fc.fields), and knn on an alias->dense_vector all returned 200 (never 400); a plain match query stays 200 (vector_query_field_error returns None for no-knn queries via the leaf `_ => None`). The shipped test (knn_on_a_real_dense_vector_field_is_not_rejected) and #498 400-test both pass. rewrite_query_aliases not handling Knn is a non-issue because knn-on-alias resolves to the target before the gate and returned 200 empty, not a false 400. No legitimate vector query is wrongly rejected; non-vector queries are unaffected.

**correctness-coverage — refuted:false, blocking:false.**

> Confirmed on rustc 1.97.1. Mapping: gate wraps XerjError::invalid_query -> InvalidQuery -> 400 in error.rs::http_status (line ~296); never 500. Fires for all three #498 shapes, verified empirically with throwaway tests (since removed): knn on absent field -> 400 (shipped test), knn on text/long/keyword -> 400, semantic on a plain text field -> 400. False-positive guard holds: dense_vector (FieldType::Vector) passes, and semantic_text maps to FieldType::Text + EmbeddingConfig (es_compat.rs:7367) so embedding.is_some() -> vector_capable -> not rejected; shipped empty-index dense_vector test passes. Common top-level ES knn DOES reach the walk: es_compat.rs (~8879) folds body.knn into effective_body.query -- single clause -> root {knn} -> QueryNode::Knn; array -> bool.should; hybrid -> bool.should -- all walked by vector_query_field_error; verified knn nested in bool.must and dis_max both 400. Missed-shape hunt: vector_query_field_error recurses every container arm unsearchable_query_field does EXCEPT the Span* variants, but span clauses can only legally contain span sub-queries (parser rejects a knn there), so a Knn/SemanticSearch is unreachable inside a span and the _=>None fallthrough is not a live gap. One rare non-blocking gap: a knn inside a rescore block bypasses the gate (it only walks request.query, not rescore), but ES disallows knn-in-rescore and rescore re-scores existing hits rather than producing the empty-hits symptom, so it is not the #498 shape.

> Ran `rustup run 1.97.1 cargo test -p xerj-api --test knn_on_non_vector_field_fails` on origin/fix/498-knn-nonvector-field-400: 2 passed. FAIL-BEFORE verified by `git apply -R` of the index.rs hunk only (test kept): the absent-field test then FAILS returning exactly the #498 silent shape — status 200 with body `"hits":{"total":{"value":0,"relation":"eq"},...,"hits":[]}` (panic left:200 right:400 at line 96). Restored, tree clean, re-passes. Assertions are correct and load-bearing: test 1 asserts status==BAD_REQUEST AND that `/error/reason` contains the field name "embedding" (so it proves 400 AND that the error names the offending field, not just any 400). The false-positive guard (test 2) uses a REAL `dense_vector` mapping (`"vec":{"type":"dense_vector","dims":3}`), empty index, and asserts status != BAD_REQUEST — a genuine legitimate-vector-query-on-empty-index case that passes with the fix, so no false positive on the exercised path.
> 
> COVERAGE GAP (non-blocking): the second conjunct of the claim is FALSE. The `semantic`/SemanticSearch arm of `vector_query_field_error` is NOT exercised by any test. `grep` for the guard's error strings ("does not exist in the mapping", "only supported on [semantic_text]") hits only index.rs, never a test; the new test file references "semantic" only in doc comments (lines 1, 11), never issues a semantic query; the semantic queries in integration.rs are positive-path tests, not a 400-on-non-semantic_text/absent-field negative. So the semantic branch is symmetric-by-construction but untested. This is a coverage hole, not a false positive, wrong status, or regression — hence non-blocking. The core evidence-holds claim (fix proven + fail-before reproduces the silent 200/0) is confirmed.

The critical false-positive lens confirmed no legitimate vector query is rejected (dense_vector, semantic_text, companion vector, Chunk, nested dotted, alias→vector, plain match all 200). One non-blocking finding — the `semantic` negative path is symmetric-by-construction and empirically 400s (per the correctness lens's throwaway probe) but has no shipped test — filed as a follow-up.

xerj-org:
Closing this approach — it is the wrong layer.

The gate keys on the **declared schema** (`declared_field` / vector-capable). That correctly 400s `knn` on a declared non-vector field, and it fixes the reported refsym repro (a BM25-only index, no vectors anywhere). But it **false-positives on a field that has indexed vectors without a schema declaration** — which XERJ permits. Running the full engine suite surfaced exactly this: `semantic_deadline_regression_tests::nested_knn_uses_the_shared_request_deadline` and `::expired_hybrid_preserves_partial_timeout_state` index real vectors into a `Schema::empty()` index and knn over them to exercise the deadline path. The schema gate 400s them before they execute, even though the vectors exist and the query is legitimate.

So the check must consult **execution-truth**, not just the schema. The right shape: for a knn/semantic clause whose field is **absent from the declared schema**, only 400 when the index has **no vector capability at all** (no declared dense_vector/semantic_text AND no indexed vector data — the refsym case); when the index *does* hold vectors, let it run (the deadline tests). A declared non-vector field is still a clean 400.

Blast radius was small (5 tests: 3 in xerj-api, 2 in xerj-engine — all special-purpose knn-on-undeclared-field scaffolding), and I had reworked 4 of them, but the 2 deadline tests are the tell that the schema-only approach is incorrect in principle, not just under-tested. Rather than ship a gate that rejects a legitimate, indexed-vector query, I'm closing this and leaving the correct approach here. The branch's `knn_on_non_vector_field_fails.rs` is a good spec for the eventual fix. #498 stays open.

PR #535, opened 2026-08-20
fix(storage): preserve explicit _source nulls through the columnar codec (#415)

### LINKED ISSUE #415: _source silently drops null-valued keys, and whether it drops them depends on corpus size
Surfaced while fixing #360. The CHANGELOG entry for that fix publicly describes this as an unfixed data-fidelity bug affecting every client, so it needs a number to link.

## Summary

`_source` silently drops null-valued keys, and whether it drops them depends on the size of the corpus.

## Measured

Independently reproduced on a binary built from the #360 branch, documents carrying four null-valued keys:

```
   500 docs -> all four null keys KEPT
 1,000 docs -> all four KEPT
 1,500 docs -> all four KEPT
 2,000 docs -> mixed: 5 kept / 2 lost
 4,000 docs -> all four LOST
 6,000 docs -> all four LOST
```

## Why it matters

`_source` is supposed to return the document as indexed. A client that round-trips a document through XERJ gets a different document back, and *which* document depends on how many other documents happen to be in the index. That makes it non-deterministic from the caller's point of view: the same write, replayed into a larger index, produces a different read.

Null is meaningful in most schemas — "explicitly absent" is not the same as "not present" for reindex, diffing, CDC and any consumer that maps `_source` back into a typed record.

**Verified:** the counts above come from a real node built from that branch.
**Assumed:** I did not identify the mechanism or the threshold, and I have not checked whether flush/merge state rather than document count is the real trigger — the size correlation may be a proxy for segment count.

xerj-org:
Closes #415 (data-loss).

## The bug
The ZBS2 columnar stored form encodes an **absent** field and a **present explicit `null`** identically — both become a NULL column slot — so decode drops both. An indexed `_source` of `{"a":"x","b":null}` came back as `{"a":"x"}`: the explicit null key silently vanished on the round-trip. Reproduced with a failing round-trip test before the fix.

## The fix — a bounded `ZBS3` overlay
- **Encode** collects the sparse present-null `(doc_ord, col_ix)` positions (a present `Some(Null)`, not an absent `None`) and wraps the ZBS2 blob in `ZBS3` = `[magic][count][(doc,col)…][unchanged ZBS2 blob]`. Emitted **only** when a segment actually contains an explicit null — every null-free segment stays **byte-identical ZBS2**, so `XERJ_FLUSH_PARITY` and the downgrade blast radius are unchanged for the common case.
- **Decode** (`decode_stored`) unwraps `ZBS3`, decodes the inner ZBS2 unchanged, and re-inserts exactly those keys as `null` **in column order** (absent fields stay dropped). Row hydration does the same for its selected rows.
- The not-yet-wired projection/knn fast paths already decline a non-ZBS2 magic (`NotV2`), so production falls back to the correct full decode; a guard-test locks this so they are taught ZBS3 before any hot-path wiring.
- `present_nulls` is **sorted** before serialisation so the JSON-reparse and from-values encoders (which discover nulls in different orders) emit identical bytes.

## Why this shape
The ZBS2 decoder rejects trailing bytes (`stored_codec.rs` "v2 trailing bytes after column payloads"), so a present-null section needs a real version bump, not an appended trailer. Wrapping an *unchanged* ZBS2 blob keeps the intricate column codecs untouched — the change is confined to wrap/unwrap + a column-order re-insert.

## Verification (rustc 1.97.1, the CI toolchain)
- `fmt --check`, `clippy --workspace --all-targets -D warnings`, `build --workspace --profile ci-test` — all clean.
- Full `xerj-storage` (169) + `xerj-engine` suites green.
- **Fail-before proven**: disabling only the decode injection drops the null again (`left: None, right: Some(Null)`).
- New tests: un-ignored #415 round-trip; null-free-stays-ZBS2; byte-identical-encoders parity; row-hydration-matches-canonical across null-bearing shapes.

Note: `test_hnsw_stale_snapshot_rebuilds_on_open` flaked once under parallel load (passes 2/2 isolated + on full re-run; unrelated subsystem) — worth a separate flaky-test issue.

### DIFF
--- a/engine/crates/xerj-storage/src/stored_codec.rs
+++ b/engine/crates/xerj-storage/src/stored_codec.rs
@@ -6,9 +6,13 @@
 //! |----------|---------|----------------------------------------------------|
 //! | `LZ4\0`  | v1      | LZ4 over a flat JSON array `[{_id,_seq_no,_source}]`|
 //! | `ZBS2`   | v2      | Columnar block — per-column codec, cross-col dep, dict+bitpack |
+//! | `ZBS3`   | v3      | A sparse present-null sidecar wrapping a `ZBS2` blob (#415) |
 //!
 //! The decoder detects the magic and returns a canonical JSON-array payload
-//! for both, so callers upstream do not need to change.
+//! for all three, so callers upstream do not need to change. `ZBS3` is
+//! emitted only for segments that carry a *present* explicit `_source` null
+//! (see [`wrap_stored_v3`]); every null-free segment stays byte-identical
+//! `ZBS2`.
 //!
 //! V2 format exists because v1 LZ4 over flat JSON leaves structural
 //! redundancy on the table: every row repeats the schema keys, and numeric
@@ -122,6 +126,21 @@ pub fn encode_stored_lz4(uncompressed: &[u8]) -> Vec<u8> {
 
 /// Magic prefix for the V2 columnar format.
 pub const STORED_V2_MAGIC: &[u8; 4] = b"ZBS2";
+
+/// Magic prefix for the V3 wrapper (issue #415): a plain ZBS2 blob preceded
+/// by a sparse present-null sidecar. The columnar V2 form stores an absent
+/// field and a *present* explicit `null` identically (a NULL column slot),
+/// so decode alone cannot tell them apart and drops both. V3 records the
+/// `(doc_ord, col_ix)` positions that were a present explicit null so decode
+/// re-inserts exactly those keys. Emitted ONLY for segments that actually
+/// contain an explicit `_source` null — every other segment stays
+/// byte-identical ZBS2, so the flush-parity contract and downgrade blast
+/// radius are unchanged for the common case. See [`wrap_stored_v3`].
+pub const STORED_V3_MAGIC: &[u8; 4] = b"ZBS3";
+
+/// The parsed contents of a ZBS3 wrapper: the present-null `(doc_ord, col_ix)`
+/// positions and the inner ZBS2 blob (which still carries its own magic).
+type StoredV3Parts<'a> = (Vec<(u32, u32)>, &'a [u8]);
 
 /// Minimum document count to bother with the columnar path.  Below this,
 /// the per-column header overhead dominates and flat LZ4 wins.
@@ -258,24 +277,36 @@ pub fn encode_stored_v2_at_level(stored_docs_json: &[u8], level: i32) -> Vec<u8>
     let mut columns: Vec<Vec<&serde_json::Value>> =
         vec![Vec::with_capacity(num_docs); col_order.len()];
 
-    for doc in &docs {
+    // Present explicit `_source` nulls (issue #415): recorded as
+    // `(doc_ord, col_ix)` so decode can distinguish them from absent fields
+    // (both encode to a NULL column slot). Must match the set discovered by
+    // `encode_stored_v2_from_values_inner` for the same docs — `wrap_stored_v3`
+    // sorts, so discovery order is irrelevant to the flush-parity contract.
+    let mut present_nulls: Vec<(u32, u32)> = Vec::new();
+    for (doc_idx, doc) in docs.iter().enumerate() {
         // __id and __seq_no
         columns[0].push(doc.get("_id").unwrap_or(&NULL));
         columns[1].push(doc.get("_seq_no").unwrap_or(&NULL));
 
         let src_obj = doc.get("_source").and_then(|s| s.as_object());
         for (cix, cname) in col_order.iter().enumerate().skip(2) {
-            let val = src_obj.and_then(|m| m.get(cname)).unwrap_or(&NULL);
-            columns[cix].push(val);
+            let looked = src_obj.and_then(|m| m.get(cname));
+            if looked.is_some_and(serde_json::Value::is_null) {
+                present_nulls.push((doc_idx as u32, cix as u32));
+            }
+            columns[cix].push(looked.unwrap_or(&NULL));
         }
     }
 
-    encode_v2_columns(
-        num_docs,
-        &col_order,
-        &columns,
-        Some(stored_docs_json),
-        level,
+    wrap_stored_v3(
+        encode_v2_columns(
+            num_docs,
+            &col_order,
+            &columns,
+            Some(stored_docs_json),
+            level,
+        ),
+        present_nulls,
     )
 }
 
@@ -373,6 +404,10 @@ fn encode_stored_v2_from_values_inner(
     // ~2× the map lookups — measured at ~115 ms per 31k-doc flush.
     let mut columns: Vec<Vec<&serde_json::Value>> =
         vec![Vec::with_capacity(num_docs), Vec::with_capacity(num_docs)];
+    // Present explicit `_source` nulls (issue #415) — see the sibling
+    // collection in `encode_stored_v2_at_level`. `wrap_stored_v3` sorts, so
+    // this path's per-doc key iteration order need not match that path's.
+    let mut present_nulls: Vec<(u32, u32)> = Vec::new();
     for (i, (_, _, src)) in docs.iter().enumerate() {
         columns[0].push(&ids[i]);
         columns[1].push(&seqs[i]);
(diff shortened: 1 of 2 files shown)


xerj-org:
Converted `split_stored_v3`'s own `chunks_exact(8)` to `as_chunks::<8>().0` so this branch is clippy-1.98 clean too (the workspace-wide 1.98 fix is #537). Verified: `clippy -p xerj-storage --all-targets -D warnings` clean and the full xerj-storage suite (169) green under rustc 1.98.0. This PR's remaining red `Format + Clippy` is the pre-existing xerj-vector breakage that #537 fixes; will rebase on main once #537 merges.

xerj-org:

Each told to REFUTE the ZBS3 data-loss fix under a distinct lens. **All 3 returned `refuted: false` with ZERO blocking findings** — merge gate (b) satisfied. The correctness skeptic added its own 600-doc adversarial round-trip (late-discovered null column, mixed present/absent, multi-null/doc, cross-dep int cols) which passed; evidence-holds independently reproduced fail-before. Verdicts verbatim:

### DOES-THE-EVIDENCE-HOLD — refuted: False

**Blocking findings:** none

**Conclusion:** Every claim the PR rests on reproduces independently; nothing refutes the fix.

1. FAIL-BEFORE reproduces exactly. With only the decode injection at stored_codec.rs:1927 disabled (else if false && present_null_set.contains(...)), v2_columnar_preserves_explicit_null_source_keys FAILS: the explicit "b": null is dropped (assertion left:None vs right:Some(Null)). Restoring the line makes it pass. Bug and fix are causally linked.

2. The "fast paths not wired into production" claim is TRUE. A whole-repo grep for decode_stored_v2_rows / _rows_projected / _projection / _knn_projection finds zero non-test, non-example callers; the only references are internal compatibility wrappers, unit tests, and examples/stored_projection_probe.rs. Every production read path (engine/src/index.rs x9, index_store.rs:3465, merge.rs:309) calls decode_stored, which unwraps ZBS3 and re-inserts present-nulls, so no projected present-null can be silently dropped in production. The row-hydration path is itself made ZBS3-aware in this PR anyway.

3. clippy 1.98.0 -D warnings is clean; full xerj-storage suite is 169/169 pass.

Additional refutation attempts that did NOT break it: (a) col_ix alignment - encode records cix as the col_order position and decode re-inserts using the same enumerate index over directory-ordered col_names, with _id/_seq_no looked up by name; the round-trip and byte-identical-encoder tests both pass. (b) Both v2 encoders (encode_stored_v2_at_level and encode_stored_v2_from_values_inner) route through wrap_stored_v3, which sorts present_nulls, so they are byte-identical (parity test passes with source-key order z-before-b differing from column order). (c) Merge path: merge.rs decodes via ZBS3-aware decode_stored then re-encodes (LZ4 verbatim here, or v2->v3 re-wrap), preserving nulls. Downgrade to a pre-fix binary would fail-loud on the unknown ZBS3 magic (only for null-bearing segments), not silently corrupt - acknowledged and not a data-loss regression.

No blocking findings. The evidence holds.


---

### CORRECTNESS — refuted: False

**Blocking findings:** none

**Conclusion:** Could not refute. PR #535 (fix/415) correctly preserves an explicit `_source` null through the columnar codec while keeping absent fields absent. All four required proofs hold: (1) present-null re-inserted / absent stays dropped, (2) ZBS3 wrap emitted only when present-nulls exist with null-free staying byte-identical ZBS2, (3) the JSON/from-values/nojson encoders are byte-identical thanks to sort_unstable in wrap_stored_v3, (4) edge cases (late-discovered null column, mixed present/absent in one column, multiple present-nulls per doc, cross-dep columns, col_ix alignment, u32 bounds) all verified. Full xerj-storage suite (169) plus an added 600-doc adversarial round-trip pass under rustc 1.98.0 ci-test. One latent, non-blocking gap: stored_slices_retained_upper_bound doesn't special-case ZBS3 and would underestimate the decoded-size bound, but it has no production callers today. No blocking findings — a maintainer should merge.

**Non-blocking notes:**
- decode_stored re-inserts a PRESENT null (source_map.insert Null when (d,cix) in present_null_set) while an ABSENT field hits neither branch and stays dropped (stored_codec.rs:1925-1931). Row hydration mirrors this (row_hydration.rs:770-777). Confirmed by test + my 600-doc adversarial test.
- Wrap guard is correct: wrap_stored_v3 returns the ZBS2 blob UNCHANGED when present_nulls.is_empty() OR the blob is not ZBS2 (stored_codec.rs:657) — the LZ4 'never make things worse' fallback stores _source verbatim so nulls survive there anyway. Null-free segments stay byte-identical ZBS2 (asserted).
- Byte-identical encoders hold: both paths record (doc,cix) with cix = index into an identically-built col_order (insertion-order discovery), and wrap_stored_v3 sort_unstable()s before serialising, so JSON/from_values/nojson emit identical ZBS3. col_ix alignment is sound: encode_v2_columns writes columns in col_order order == decode col_names order == directory.columns order.
- split_stored_v3 bounds are safe: count*8 via checked_mul, all slice reads via .get()->Result, try_into().unwrap() only on get()-guaranteed 4/8-byte ranges. doc/col cast `as u32` is lossy only above 4B docs/cols (not reachable in a flush segment).
- NON-BLOCKING gap: stored_slices_retained_upper_bound (stored_codec.rs:736) was NOT updated for ZBS3 — a ZBS3 blob falls into the final `else` branch and gets decoded_bound = bytes.len(), a gross UNDERESTIMATE of the true decoded (re-inflated JSON) size, so it is no longer a real upper bound for null-bearing segments. Currently harmless: the function has ZERO production callers repo-wide (only unit tests). Would become a real bug if this bound is later wired to size a buffer / retained-bytes firewall for ZBS3 input. Worth a follow-up ticket, not a revert.
- Not-yet-wired projection/knn fast paths (decode_stored_v2_projection:1102, knn projection:1636) correctly return NotV2 for ZBS3, so production falls back to full decode_stored which handles V3.

---

### REGRESSION-RISK — refuted: False

**Blocking findings:** none

**Conclusion:** Could not refute PR #535. The ZBS3 wrapper fix for #415 is correct and regression-safe. Full xerj-storage and xerj-engine suites pass under CI toolchain 1.98.0 and clippy is clean. Verified the corruption-critical invariant that encode-time present_null (doc, col_ix) indices share the same col_order index space that decode/row-hydration re-insert against (encode_v2_columns emits payloads strictly in col_order); flush parity holds because both encoders sort present_nulls and the pre-existing ZBS2 parity contract forces identical col_order (new byte-identical test passes); both merge paths round-trip nulls via decode_stored -> reparse -> re-encode; projection/kNN decline as NotV2 and row-hydration handles ZBS3, none panic (split_stored_v3 is fully bounds-checked); and the only live engine decoder is decode_stored, which handles ZBS3, while all fast-path/retained-bound primitives that lack ZBS3 handling have no production callers. Null-free segments remain byte-identical ZBS2. No blocking findings; two latent/non-blocking notes recorded.

**Non-blocking notes:**
- stored_slices_retained_upper_bound (stored_codec.rs:736) has no ZBS3 arm: a ZBS3 blob falls into the else branch and returns compressed bytes.len() as the decoded upper bound, under-counting retained size. Latent only — zero production callers repo-wide (tests/examples only); add a ZBS3 arm before wiring it into cache admission.
- Rollback/downgrade: a pre-ZBS3 binary reading a ZBS3 (null-containing) segment hits decode_stored's Ok(bytes.to_vec()) fallthrough and then fails JSON parse — a read error, not silent corruption, and only for segments that actually contain an explicit _source null. Inherent forward-compat cost of the format bump; PR documents it. Null-free segments stay byte-identical ZBS2, so the rollback blast radius is bounded as claimed.
- decode_stored_v2 now always allocates a HashSet<(u32,u32)> even for plain ZBS2 (empty for null-free segments) and does one set lookup per null cell; negligible, no measurable perf regression, and null-free segments take an identical byte path.

---

## Merge decision
- Gate (b): **0 BLOCKING** across all 3 lenses — clear.
- One non-blocking finding filed as follow-up (not a revert): `stored_slices_retained_upper_bound` doesn't special-case ZBS3 and would under-estimate the decoded-size bound — currently harmless (zero production callers). Tracked separately.
- Will merge on gate (a) all-green + gate (c) CLEAN, then re-verify main under 1.98.0.

Issue #530, opened 2026-08-20
Add a shipped test for the semantic-on-non-semantic_text 400 path (from #498/#529)


xerj-org:
The #498 gate (`vector_query_field_error`, #529) covers both `knn` and `semantic`, and the verification empirically confirmed a `semantic` query on a plain (non-`semantic_text`) or absent field returns 400. But only the `knn` path has a shipped test (`knn_on_non_vector_field_fails.rs`); the `semantic` arm is symmetric-by-construction but untested. Add a negative test: a `semantic` query on a text/keyword/absent field must 400 and name the field. Non-blocking coverage gap surfaced by the #529 verification.

PR #541, opened 2026-08-20
fix(engine): 400 a knn/semantic clause on an unanswerable field, not silent empty 200 (#498)

### LINKED ISSUE #498: knn against an index with no dense_vector field returns 0 hits / HTTP 200 with no error (silent false-negative)

xerj-org:
Closes #498 (wrong-results / silent false-negative).

## The bug
A `knn` naming a field the index cannot answer returned **200 / `hits.total:0` / `error:null`** — the caller cannot tell "nothing matched" from "this index cannot do vector search." It surfaced on `refsym` (no `dense_vector` field): the reference-coding retrieval arm's vector queries returned nothing, silently. ES answers `400 illegal_argument_exception` naming the field.

## Fix — execution truth, not a mapping gate
The first attempt (#529) gated on the mapping and was reverted because it false-positived on **indexed-but-undeclared** vector fields. The proof: the deadline tests seed **256** `embedding` vectors into a `Schema::empty()` index — below `HNSW_MIN_DOCS = 1024`, so no graph is built and `hnsw_field` is `None`, yet the field is fully brute-force answerable.

So the fix keys on execution truth: on a **genuinely empty, non-timed-out** knn result, return 400 iff the field is **neither** declared `dense_vector`/`semantic_text` **nor** carries any indexed vector data (`index_has_any_vector_at`, reusing the canonical `extract_numeric_vector`). It runs only on an empty result, so the hot path — and every declared-but-empty or indexed-but-undeclared field — is untouched.

- `knn_on_absent_field_is_400_not_silent_empty` — **fail-before proven** (disable the reject call → returns 200).
- `knn_on_declared_but_empty_vector_field_is_empty_200` — the #529 invariant (declared-but-empty → empty 200).
- **Full `xerj-engine` + `xerj-api` suites green** (the #529 regressions — the 256-undeclared-vector deadline tests, msearch top-level knn, passage-ambiguity — all pass). `msearch_does_not_execute_top_level_knn` relied on the old silent-empty behaviour and is updated into a stronger discriminator (`_search` executes the knn → 400s the absent field; `_msearch` ignores it → 200/1).
- fmt + clippy `-D warnings` clean.

Scope note (follow-up #530): the `semantic`-on-unpopulated-`semantic_text` path is a sibling case; this PR fixes the raw-`knn` path the issue reproduces.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -39941,12 +39941,13 @@ mod script_guard_symmetry_tests {
     }
 
     /// Why top-level `knn` is not a `GuardedField`: `_msearch` never executes
-    /// it, so a script parked there reaches no executor. The discriminator is
-    /// a `knn.filter` that matches nothing — `_search` honours it and returns
-    /// zero hits, `_msearch` ignores the whole clause and returns the doc.
-    ///
-    /// If this ever starts failing because `_msearch` grew knn support, that
-    /// path gained a script sink and `knn` must become a guarded field.
+    /// it, so a script parked there reaches no executor. The discriminator is a
+    /// `knn` naming the field `vec`, which the seed index does not have: since
+    /// #498 `_search` EXECUTES the clause and rejects the unanswerable field
+    /// with a `400`, while `_msearch` ignores the whole clause and returns the
+    /// doc (`200`, one hit). If `_msearch` ever grew knn support it would 400
+    /// here too — that path would have gained a script sink and `knn` must
+    /// become a guarded field.
     #[tokio::test]
     async fn msearch_does_not_execute_top_level_knn() {
         let state = test_state();
@@ -39968,12 +39969,18 @@ mod script_guard_symmetry_tests {
             "`_msearch` must be ignoring the knn clause entirely: {msearch_item}"
         );
 
+        // `_search` executes the knn, so it validates the field — `vec` is
+        // absent from the mapping and carries no vectors, which is a 400 since
+        // #498 (a silent empty 200 hid "this index cannot do vector search").
         let (search_status, search_body) = search_verdict(state, &body).await;
-        assert_eq!(search_status, 200, "{search_body}");
-        assert_eq!(
-            search_body["hits"]["total"]["value"],
-            json!(0),
-            "`_search` must be executing the knn filter: {search_body}"
+        assert_eq!(search_status, 400, "{search_body}");
+        assert!(
+            search_body
+                .pointer("/error/reason")
+                .and_then(Value::as_str)
+                .unwrap_or_default()
+                .contains("vec"),
+            "the 400 must name the unanswerable field: {search_body}"
         );
     }
 }
--- a/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
+++ b/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
@@ -1,0 +1,168 @@
+//! A `knn` (or `semantic`) clause naming a field the index cannot answer must
+//! fail loudly with a `400`, not return a silent, successful empty result —
+//! issue #498.
+//!
+//! `refsym`, XERJ's own reference-coding corpus, has no `dense_vector` field, so
+//! `knn: {field: "embedding", ...}` had nothing to run against and came back
+//! `200` with `hits.total = 0` and `error: null`. An agent cannot tell "nothing
+//! matched" from "this index does not do vector search," so it concludes the
+//! code does not exist — the exact silent-false-negative the vector/semantic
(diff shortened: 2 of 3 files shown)


xerj-org:

**2 of 3 returned; both `refuted: false` with ZERO blocking findings.** The 3rd (evidence-holds) hit a harness StructuredOutput failure and returned no verdict — that lens is fail-before + clippy, both of which I reproduced directly (below). No skeptic refuted; 0 blocking overall.

### CORRECTNESS — refuted: false — 0 blocking
Could not refute. The 400 fires on exactly the unanswerable single-clause knn case and nothing else. `reject_unanswerable_knn_field` runs ONLY on an empty, non-timed-out result and returns Ok early if the field is (a) declared dense_vector, (b) semantic_text, or (c) carries any indexed vector, before 400ing. Verified with 5 adversarial probes: declared+filter-matching-nothing → empty 200; undeclared-but-indexed+filter-nothing → empty 200; undeclared-but-indexed (no filter) → 5 hits @1.0 (so `index_has_any_vector_at`'s extraction matches the scan); absent field → 400 naming it; size:0 → no over-fire. Dispatch refactor (`?` + single `return Ok`) is behaviour-preserving.

### REGRESSION-RISK — refuted: false — 0 blocking
Could not refute. Full `xerj-engine` (21 ok binaries, 0 failed) and `xerj-api` (38 ok, 0 failed) suites green under 1.98.0. The `msearch_does_not_execute_top_level_knn` update is faithful (only the `_search` side moved to 400 naming `vec`; `_msearch` unchanged at 200/1) and strengthens the discriminator. `index_has_any_vector_at` is gated to empty results AND only runs for UNDECLARED fields (declared vector fields short-circuit on the cheap schema read), so declared-vector workloads and large-index no-match knn are untouched.

### EVIDENCE-HOLDS — (no skeptic verdict; verified by me)
Fail-before reproduced: disabling only `if result.hits.is_empty() && !result.timed_out {` → the absent-field test returns 200 (bug), restored → 400. `clippy -p xerj-engine -p xerj-api --all-targets -D warnings` clean.

## Merge decision
- 0 BLOCKING across the two returned lenses + my evidence verification → gate (b) satisfied.
- **Non-blocking, filed as follow-ups (not revert-worthy):** the check is wired only into the single-knn arm — multi-knn arrays + nested knn tracked in **#542**, the `semantic` clause in **#530**. Commit retitled to "single top-level knn" to match. Perf nit (a redundant deep-clone scan on empty results over *undeclared* vector fields, incl. size:0 analytics) — fix-forward, confined.
- Will merge on all-green + CLEAN, then re-verify main under 1.98.0.

Issue #542, opened 2026-08-20
Extend the unanswerable-knn-field 400 to multi-knn arrays and nested knn


xerj-org:
Follow-up to #498/#541. The execution-truth answerability 400 (`reject_unanswerable_knn_field`) is wired only into the single top-level `knn` arm (`peel_knn_query`, index.rs). These forms still return a silent empty 200 on an absent/non-vector field:

- **multi-knn arrays** (`knn: [...]`, `run_multi_knn_brute_force`) — per-clause check;
- **nested knn** (`nested { query: { knn }}`, `peel_nested_knn_query`).

Apply the same check (declared `dense_vector`/`semantic_text` OR has indexed vector data, else 400 naming the field) per arm, with tests. (The `semantic`-clause sibling is already tracked by #530.) Surfaced by #541 verification.

xerj-org:
Multi-knn arm shipped in #543 (merged). Nested-knn arm design, now pinned from the code:

The nested scoring loop (index.rs:~12249) resolves vectors as: `get_field_value(&src, nested_path)` -> the nested array, then per element `elem.get(&subfield)` -> a numeric `Value::Array` (where `subfield = field.strip_prefix("{nested_path}.")`, computed at ~12150). So a nested vector lives at `<nested_path>[].<subfield>`, NOT at a top-level path — which is exactly why the single/multi-knn `index_has_any_vector_at` (a top-level `extract_numeric_vector`) would false-negative and wrongly 400 a valid **undeclared** nested-vector field.

**Fix shape (mirrors the scoring loop):**
- Add `index_has_any_nested_vector_at(nested_path, subfield)`: scan memtable + segments; for each doc, `get_field_value(src, nested_path)` -> array; return true if any element has a numeric `Value::Array` at `subfield`. Early-exit; only called on an empty result.
- In the nested-knn dispatch (index.rs:~14129, `peel_nested_knn_query` arm): after `run_nested_knn_brute_force_with_deadline`, on an empty non-timed-out result, 400 (naming `field`) iff the field is neither declared `dense_vector` (`collect_dense_vector_fields` already emits dotted paths, so a declared `<nested_path>.<subfield>` short-circuits) NOR has any indexed nested vector. Consistent with the single-knn (#498) and multi-knn (#543) arms; preserves declared-but-empty and indexed-but-undeclared nested fields.
[...]

PR #543, opened 2026-08-21
fix(engine): 400 an unanswerable field in a multi-knn array clause (#542, multi-knn part)

xerj-org:
Refs #542 (the multi-knn part). Follow-up to #498/#541.

#498 wired the execution-truth answerability 400 only into the single top-level knn arm. The ES multi-knn array form (`knn: [...]` with 2+ clauses → `run_multi_knn_brute_force`) still returned a **silent empty 200** on a clause naming an unanswerable field.

## Fix
Apply the same check to the multi-knn dispatch: on an empty, non-timed-out union, reject the first clause naming an unanswerable field (neither declared `dense_vector`/`semantic_text` nor carrying indexed vector data). Declared-but-empty and indexed-but-undeclared fields still return their real empty result.

- `multi_knn_on_absent_field_is_400_not_silent_empty` (2-element array — a single-element array collapses to the single-knn path already fixed by #498). Fail-before proven (ignored form returned 200/`total:0`).
- Full `xerj-engine` + `xerj-api` suites green; fmt + clippy `-D warnings` clean.

## Scope
The **nested-knn** arm is deliberately NOT touched — its vectors live in nested array elements, so the top-level `index_has_any_vector_at` scan would false-negative and 400 a valid undeclared nested-vector field. It needs a nested-aware presence check and stays open under #542. (The `semantic` sibling is #530.)

### DIFF
--- a/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
+++ b/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
@@ -166,3 +166,58 @@ async fn knn_on_declared_but_empty_vector_field_is_empty_200() {
         .unwrap_or(u64::MAX);
     assert_eq!(total, 0, "expected an empty result set: {body}");
 }
+
+/// #542 (follow-up to #498): the MULTI-knn array form (`knn: [...]`) must also
+/// 400 a clause naming an unanswerable field — #498 only wired the check into
+/// the single top-level knn arm, so the array form still returns a silent
+/// empty 200.
+#[tokio::test]
+async fn multi_knn_on_absent_field_is_400_not_silent_empty() {
+    let (app, _dir) = app().await;
+    let (status, body) = json_req(
+        &app,
+        "PUT",
+        "/codes",
+        json!({ "mappings": { "properties": {
+            "name": { "type": "text" },
+            "code": { "type": "keyword" }
+        } } }),
+    )
+    .await;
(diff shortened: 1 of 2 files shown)


xerj-org:

**All 3 returned `refuted: false` with ZERO blocking findings** — gate (b) satisfied. Correctness ran the exact over-fire cases (two declared fields no-docs → 200; two declared + filter-excludes-all → 200; mixed answerable+absent, non-empty union → 200); regression confirmed both full suites green and no hybrid/RRF/knn+aggs path newly 400s; evidence reproduced fail-before exactly (disable the multi-knn guard → 200 bug). Verdicts verbatim:

### None — refuted: None

**Conclusion:** 


---

### None — refuted: None

**Conclusion:** 


---

### None — refuted: None

**Conclusion:** 


---

## Merge decision
- 0 BLOCKING across all 3 lenses → merge on all-green + CLEAN.
- Non-blocking (tracked, not revert-worthy): (1) mixed-clause under-fire — when one clause matches (non-empty union), an absent clause in the same request is silently ignored rather than 400'd; this is the PR's explicit result-based design (keys on empty union), and matching ES's per-field validation would need upfront per-clause checks. Noting on #542. (2) No test locks the multi-knn declared-but-empty / indexed-but-undeclared → 200 invariant (only the 400 case is covered); the skeptics confirmed the behaviour empirically. Will add in the nested-knn follow-up under #542.

PR #545, opened 2026-08-21
test(api): shipped semantic-on-non-semantic_text 400 test + name the field (#530)

### LINKED ISSUE #530: Add a shipped test for the semantic-on-non-semantic_text 400 path (from #498/#529)

xerj-org:
Closes #530. Follow-up to #498/#529.

The `semantic` path already 400s a query on a plain text/keyword or absent field, but only the `knn` arm had a shipped test — the `semantic` arm was symmetric-by-construction yet untested.

## Changes
- **Test** (`semantic_on_non_semantic_text_400s.rs`): a `semantic` query on a `text` field and on an absent field must **400** and **name the field**.
- **Name the field** in the semantic 400 message (was a generic "requires `semantic_text` or active backend" hint), symmetric with the knn 400 (#498).

- Both tests pass; **fail-before proven** (revert the field name → the "names the field" assertion fails, the 400 still fires but is generic).
- Full `xerj-engine` + `xerj-api` suites green; fmt + clippy `-D warnings` clean.

### DIFF
--- a/engine/crates/xerj-api/tests/semantic_on_non_semantic_text_400s.rs
+++ b/engine/crates/xerj-api/tests/semantic_on_non_semantic_text_400s.rs
@@ -1,0 +1,133 @@
+//! A `semantic` query naming a field that is not a `semantic_text` (a plain
+//! text/keyword field, or absent) must fail with a `400`, not silently degrade
+//! — issue #530, the symmetric sibling of the `knn` case (#498).
+//!
+//! With no active embedding backend, the built-in lexical embedder only
+//! auto-embeds `semantic_text` fields; a `semantic` query against anything else
+//! has no comparable stored vector to match, so it must be rejected loudly.
+//!
+//! Elasticsearch is referenced for wire semantics only; no code from it is
+//! reproduced here.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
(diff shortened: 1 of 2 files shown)


xerj-org:

**All 3 returned `refuted: false` with ZERO blocking findings.** Correctness confirmed `{field}` resolves to the user's query field and no valid semantic query regresses (control flow unchanged, only the message string); regression confirmed both full suites green and no test asserts the old message; evidence reproduced fail-before exactly (remove `{field}` → the 'names the field' assertion fails while the 400 still fires) with clippy clean, on the semantic path (not knn).

**Acted on the one non-blocking finding:** the correctness skeptic noted the absent-field test asserted `contains("embedding")`, but "embedding" is a substring of "active embedding backend" in the static message, so it proved field-naming only vacuously. Fixed by querying a distinctive field `nonexistent_vec` (absent from the message text), so the assertion is now a genuine proof; both tests still pass. (The `title` test already proved it non-vacuously.)

0 BLOCKING → merge on all-green + CLEAN, then re-verify main.

PR #546, opened 2026-08-21
fix(engine): 400 an unanswerable field in a nested knn clause (#542, nested arm)

### LINKED ISSUE #542: Extend the unanswerable-knn-field 400 to multi-knn arrays and nested knn

xerj-org:
Closes #542 (completes it — #543 shipped the multi-knn arm, this is the nested arm). Follow-up to #498.

A `nested` knn (`nested { path: P, query: { knn { field: P.vec }}}`) naming a field the index cannot answer still returned a **silent empty 200**.

## Fix
Nested vectors live at `<nested_path>[].<subfield>` (the scoring loop in `run_nested_knn_brute_force_with_deadline`), so the top-level `index_has_any_vector_at` would false-negative on them and wrongly 400 a valid undeclared nested field. `index_has_any_nested_vector_at` mirrors that traversal exactly. On an empty, non-timed-out nested-knn result, 400 iff the field is neither declared `dense_vector` (dotted path short-circuits) NOR carries indexed nested vectors — preserving declared-but-empty and indexed-but-undeclared nested fields, consistent with #498/#543.

- `nested_knn_on_absent_field_is_400_not_silent_empty` — **fail-before proven** (disable the reject → 200/`total:0`).
- **Full `xerj-engine` + `xerj-api` suites green** (the nested-knn deadline tests pass); fmt + clippy `-D warnings` clean.

With this, the whole unanswerable-field 400 family is complete: single-knn (#498), multi-knn (#543), nested-knn (this), semantic (#530).

### DIFF
--- a/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
+++ b/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
@@ -221,3 +221,52 @@ async fn multi_knn_on_absent_field_is_400_not_silent_empty() {
         "the 400 must name the field: {body}"
     );
 }
+
+/// #542 (nested-knn arm): a `nested` knn naming a vector field absent from the
+/// nested mapping must 400, not return a silent empty 200. The nested arm is a
+/// separate dispatch from the single/multi-knn arms fixed in #498/#543.
+#[tokio::test]
+async fn nested_knn_on_absent_field_is_400_not_silent_empty() {
+    let (app, _dir) = app().await;
+    let (status, body) = json_req(
+        &app,
+        "PUT",
+        "/docs",
+        json!({ "mappings": { "properties": {
+            "items": { "type": "nested", "properties": { "name": { "type": "text" } } }
+        } } }),
+    )
+    .await;
+    assert!(status.is_success(), "create index: {status} {body}");
+    let (_s, _b) = json_req(
+        &app,
+        "POST",
+        "/docs/_doc/1",
+        json!({ "items": [ { "name": "sniff" }, { "name": "detect" } ] }),
+    )
+    .await;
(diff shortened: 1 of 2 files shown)


xerj-org:

**All 3 returned `refuted: false` with ZERO blocking findings.** Correctness confirmed `index_has_any_nested_vector_at` mirrors the scoring-loop traversal exactly and is dimension-agnostic (verified declared-but-empty→200 and undeclared-dim-mismatch→200 with fresh probes); regression: full `xerj-engine` (610 lib + all integration) + `xerj-api` (39 binaries) suites 0 failed, timed-out nested knn does not 400; evidence reproduced fail-before exactly (disable the guard → 200/`total:0`), clippy clean. Verdicts verbatim:

### None — refuted: None

**Conclusion:** 


---

### None — refuted: None

**Conclusion:** 


---

### None — refuted: None

**Conclusion:** 


---

## Merge decision
- 0 BLOCKING across all 3 lenses → merge on all-green + CLEAN.
- Recurring non-blocking (filed as follow-up https://github.com/xerj-org/xerj/issues/547): the two nested must-stay-200 branches are verified by skeptic probes but lack *shipped* regression tests; tracked for a small test-only PR across the knn arms.

PR #549, opened 2026-08-21
test(api): lock the multi-knn & nested-knn declared-but-empty -> 200 invariants (#547)

### LINKED ISSUE #547: Add shipped tests locking the knn/nested-knn declared-but-empty & indexed-but-undeclared -> 200 invariants
Recurring non-blocking note from the #543 and #546 verifications: the knn unanswerable-field 400 arms ship a test for the 400 case, but the *must-stay-200* invariants are only verified by throwaway skeptic probes, not shipped tests. Add permanent regression tests, per arm (single #498 has knn_on_declared_but_empty_vector_field_is_empty_200 already; multi-knn #543 and nested-knn #546 do not):

- multi-knn: a 2-clause array on DECLARED-but-empty vector fields -> empty 200 (not 400); on an indexed-but-undeclared field -> 200.
- nested-knn: a nested knn on a DECLARED-but-empty nested dense_vector -> empty 200; on an indexed-but-undeclared nested vector (incl. a dimension mismatch) -> 200.

These lock the exact distinction that reverted #529. Low-risk test-only additions.

xerj-org:
Closes #547. Adds the multi-knn and nested-knn declared-but-empty -> empty-200 invariant tests (the #529-class distinction: a DECLARED dense_vector field with no docs is a real empty 200, not a 400). Single-knn already had this guard; this extends it to the array and nested arms fixed in #543/#546. Test-only; all knn tests pass under the CI toolchain.

### DIFF
--- a/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
+++ b/engine/crates/xerj-api/tests/knn_on_missing_field_400s.rs
@@ -270,3 +270,79 @@ async fn nested_knn_on_absent_field_is_400_not_silent_empty() {
         "the 400 must name the field: {body}"
     );
 }
+
+/// #547: the #529-class invariant for MULTI-knn — a 2-clause array on
+/// DECLARED-but-empty dense_vector fields returns an empty 200, not a 400.
+#[tokio::test]
+async fn multi_knn_on_declared_but_empty_vector_fields_is_empty_200() {
+    let (app, _dir) = app().await;
+    let (status, body) = json_req(
+        &app,
+        "PUT",
+        "/vecs",
+        json!({ "mappings": { "properties": {
+            "a": { "type": "dense_vector", "dims": 3 },
+            "b": { "type": "dense_vector", "dims": 3 }
+        } } }),
+    )
+    .await;
+    assert!(status.is_success(), "create index: {status} {body}");
+    // No documents indexed — both fields are declared vectors with nothing to match.
+    let (status, body) = json_req(
+        &app,
+        "POST",
+        "/vecs/_search",
+        json!({ "knn": [
+            { "field": "a", "query_vector": [0.1, 0.2, 0.3], "k": 5, "num_candidates": 50 },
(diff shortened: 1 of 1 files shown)


xerj-org:

**Verdict: 0 BLOCKING — all three refuted=false / severity=none.**

**correctness** (refuted=false): Both new tests pass under 1.98.0 and correctly lock the #529-class invariant. Each asserts HTTP success AND `hits.total.value == Some(0)` (knn_on_missing_field_400s.rs:301-309, :339-347). Non-vacuous: the 200-vs-400 outcome is driven solely by whether the field is a DECLARED dense_vector (index.rs reject_unanswerable_knn_field / reject_unanswerable_nested_knn_field). The multi-knn test uses a real 2-element knn array (not a 1-element array that collapses to single-knn); the nested test uses a real nested dense_vector mapping. A 400 body lacks `/hits/total/value`, so `total==Some(0)` independently rejects the error path. *Nit only:* `is_success()` accepts any 2xx rather than strictly 200, but `total==0` requires a real body so it is not exploitable.

**regression-risk** (refuted=false): Whole binary green under 1.98.0 — all 6 tests — stable across 3 parallel reruns plus `--test-threads=1` (no flakiness/order-dependence). No state leak (each test builds its own tempdir + in-process Engine/AppState via tower::oneshot, no ports). Non-vacuous and non-redundant vs the single-knn empty-200 test — they lock the invariant on the distinct multi-knn array and nested-knn arms. Clippy `-D warnings` clean.

**evidence-holds** (refuted=false): Baseline 6 passed. **Counterfactual mutation** — reverted the declared-vector short-circuit in index.rs reject_unanswerable_knn_field/reject_unanswerable_nested_knn_field — flipped BOTH new tests to failure with the exact `[knn] query field ... is not a vector field` 400 (3 passed; 3 failed). The companion absent-field tests still pass, proving those rejection paths are wired into the multi and nested arms. Passing, non-vacuous test set.

Gate: (a) 17/17 green, none pending; (b) 0 BLOCKING; (c) mergeStateStatus CLEAN. Merging.

Issue #524, opened 2026-08-19
Alias filter metadata (index_alias_metadata) is never persisted — lost on restart (wrong search results; data loss once by-query honours it, blocks #453)


xerj-org:
### The bug

`index_alias_metadata: Arc<DashMap<String, Value>>` (engine.rs:655) holds each alias's **filter** (the `{"filter": …}` an alias carries). It is initialised empty (engine.rs:860) and is **never persisted** — `grep` finds no serialize/flush of it. Boot only runs `load_persisted_aliases()`, which restores `aliases.json` = the alias→members **topology** (`HashMap<String, Vec<String>>`), no filters.

So a filtered alias survives a restart with its **members but not its filter**.

### Why it matters now (two severities)

1. **Wrong results (already live).** `search_impl` honours alias filters (the filter loop predates and was refined by #449). After a restart, a filtered-alias `_search`/`_count` runs **unfiltered** — it returns the whole backing corpus instead of the slice the alias names.
2. **Data loss (blocks #453).** #453 wires alias filters into `_delete_by_query`/`_update_by_query`. After a restart the filter is gone, so `POST /{filtered-alias}/_delete_by_query {"match_all":{}}` deletes **every document in every member** (measured shape: 30 of 30 where the filter meant 9) — strictly worse than the current first-member-only bug. This is why #453 cannot merge as-is.

### Fix

Persist alias filter metadata durably and load it on boot, the same way `aliases.json` handles topology. Either:
- a sibling file (e.g. `alias_metadata.json`) flushed whenever `index_alias_metadata` is mutated (`set_alias_metadata`/`clear_alias_metadata`, `put_alias`'s attach, index-create-time `aliases`, `remove_alias`) and loaded in the boot path next to `load_persisted_aliases()`; **or**
- fold the filter into the persisted alias record (schema change to `aliases.json` — watch downgrade: an older binary reading the new shape).

Prefer the sibling file for downgrade safety (an older binary simply ignores it), mirroring the sidecar decision in #446.

Add a test that **restarts the engine** (open a fresh `Engine` on the same data dir) and asserts a filtered alias still filters `_search` and, once #453-style application lands, `_delete_by_query`. The existing alias tests never restart, which is why this stayed green.

Provenance: found by the data-safety lens of the #453 verification and confirmed in engine.rs. Blocks #453.

PR #552, opened 2026-08-21
fix(es-compat): fan _delete_by_query/_update_by_query out to every alias member (#450)

### LINKED ISSUE #450: _delete_by_query and _update_by_query through a multi-index alias silently touch only the first member and report success

xerj-org:
Closes #450.

## The bug (silent partial delete — data integrity)
`_delete_by_query` and `_update_by_query` both began with `Engine::get_index(&index)`, which resolves an alias to `aliased.first()` — one member. A destructive by-query through a multi-index alias therefore touched only the first member and answered a clean `{"deleted"/"updated": N, "failures": []}`, indistinguishable from a complete run. Strictly worse than the #433/#449 read truncation: a short read is recoverable, a short delete leaves the caller believing the operation completed.

Reproduced: 3-member alias `tri` (12 docs), `POST /tri/_delete_by_query {match_all}` returned `{"total":4,"deleted":4,"failures":[]}` — idx-b/idx-c untouched.

## The fix
`resolve_by_query_targets()` resolves the selector the way the read paths do (`resolve_index_selector`, #433): expand to every member, and — crucially for a destructive path — AND in any alias `filter` exactly as `search_impl` does, so a **filtered** alias deletes only its slice, not the whole backing index (the #524 hazard). Each member runs the by-query; `aggregate_by_query_results()` sums the counts and concatenates `failures`. A single member (plain index or one-member alias) is returned **byte-for-byte unchanged**, so the existing wire format never drifts — only genuine multi-member aliases aggregate. If any member errors, that error is surfaced rather than a clean aggregate.

## Verification (rustc 1.98.0 — CI @stable)
- New `by_query_through_alias_touches_every_member.rs` covers both verbs; fail-before was `deleted:4/12`, now both pass.
- Full `xerj-api` suite: 465 tests, 0 failed (the #529 lesson — not just the new test).
- `cargo fmt --check`, `clippy --workspace --all-targets -D warnings`, `build --profile ci-test`: all clean.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -24363,19 +24363,125 @@ pub struct DeleteByQueryBody {
     pub query: Value,
 }
 
+/// Resolve a by-query selector to the concrete member indices it must touch and
+/// the effective query to run against each.
+///
+/// The destructive by-query handlers used to call `Engine::get_index`, which
+/// resolves an alias to `aliased.first()` — one member. A `_delete_by_query`
+/// through a 3-member alias therefore deleted a third of the corpus and
+/// reported a clean success (#450). This fans the selector out to every member
+/// the read paths already see (`resolve_index_selector`, #433) and — crucially
+/// for a destructive path — ANDs in any alias `filter` the way `search_impl`
+/// does, so a *filtered* alias deletes only its slice, not the whole backing
+/// index (the #524 hazard).
+async fn resolve_by_query_targets(
+    state: &AppState,
+    spec: &str,
+    user_query: Value,
+) -> std::result::Result<(Vec<std::sync::Arc<xerj_engine::Index>>, Value), xerj_common::XerjError> {
+    let member_names = resolve_index_selector(state, spec).await;
+
+    // Alias filters are keyed on the name AS WRITTEN (the alias, not its
+    // expanded members) — mirror `search_impl`'s alias-filter path exactly.
+    let mut alias_filters: Vec<Value> = Vec::new();
+    for part in spec.split(',').map(str::trim).filter(|s| !s.is_empty()) {
+        if let Some(entry) = state.engine.aliases.get(part) {
+            for backing in entry.value().iter() {
+                if let Some(meta) = state.engine.index_alias_metadata.get(backing) {
+                    if let Some(filter) = meta.get(part).and_then(|v| v.get("filter")).cloned() {
+                        alias_filters.push(filter);
+                    }
+                }
+            }
+        }
+    }
+    let effective_query = if alias_filters.is_empty() {
+        user_query
+    } else {
+        json!({ "bool": { "must": [user_query], "filter": alias_filters } })
+    };
+
+    let mut indices = Vec::with_capacity(member_names.len());
+    for name in &member_names {
+        indices.push(
+            state
+                .engine
+                .get_index(name)
+                .map_err(xerj_common::XerjError::from)?,
+        );
+    }
+    Ok((indices, effective_query))
+}
+
+/// Combine the per-member results of a fanned-out by-query into one response.
+///
+/// A single member (the common case: a plain index, or a one-member alias) is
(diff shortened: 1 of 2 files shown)


xerj-org:

**Verdict: 0 BLOCKING.** correctness=nit, regression-risk=should-fix, evidence-holds=should-fix — all refuted=false.

**correctness (nit):** Fan-out + aggregation correct. `aggregate_by_query_results` sums took/total/deleted|updated/batches/version_conflicts/noops and concatenates failures correctly; error-detection via `find(v.get("status").is_some())` is sound (success bodies never carry `status`; every error path does). Single-member path is byte-identical (`len()==1` early-returns raw body; `effective_query==user_query` with no alias filter). No double-run (`resolve_index_selector` dedupes). *Nit:* multi-member `_update_by_query` aggregate omits the `"deleted": 0` field the single-member update body includes (24887) — cosmetic asymmetry, not a wrong count.

**regression-risk (should-fix):** Full `xerj-api` suite green — **465 passed, 0 failed across 40 binaries**; clippy `-D warnings` clean. Single-member wire format never drifts; a nonexistent exact index still 404s; a filtered alias is narrowed via `bool{must:[query],filter:[alias_filter]}` — can only under-delete, never over-delete. *should-fix GAP:* neither test exercises a filtered alias (0 `filter` occurrences), so the #524 filtered-slice path is unverified by execution. *Minor note:* a wildcard spec matching zero indices now returns `200 {deleted:0}` instead of 404 (aligns with read-path semantics; not data-loss).

**evidence-holds (should-fix):** Core #450 claim holds. Fail-before real: `get_index` (engine.rs:3193-3205) → `aliased.first()`, so pre-fix delete drained only member-0 (4/12) and reported clean success. Fix expands to all members and runs the same query against each; both tests pass. The riskiest claim — filtered alias deletes only its slice (#524) — is CORRECT by inspection: filters gathered keyed on the written name (24387-24397), mirroring `search_impl`, ANDed as `bool.filter` into one `effective_query` applied to EVERY member. Over-delete is structurally impossible (a filter only narrows). *should-fix:* that path is verified only by reading, not execution — recommend a filtered-alias delete test.

Gate: (a) 17/17 green, none pending; (b) 0 BLOCKING (should-fix/nit tracked as fast-follow); (c) CLEAN. Merging. Follow-up filed for the filtered-alias test + update-aggregate nit.

Issue #451, opened 2026-08-17
Fifteen independent index-selector resolvers disagree about aliases: 7 endpoints 404, 7 answer from one member, 4 ignore alias filters


xerj-org:
Found by an independent verification of #449, which counted the resolution sites while checking whether that PR's "there may be a fourth copy" caution was justified. It was, and understated: **15 independent index-selector resolution sites** across `engine/crates/xerj-api/src/es_compat.rs` and `engine/crates/xerj-engine/src/engine.rs`. #449 fixes two.

## Two failure classes, eleven endpoints

**(A) `resolve_indices_for_op` (`es_compat.rs:19415`) has no alias branch at all.** A concrete part that is an alias name falls to the else at `:19446` — `if !all.iter().any(|n| n == part) { return Err(index_not_found(part)) }` — and `all` is `list_indices()`, which holds index names only. So these **HTTP 404 on a valid alias**:

`_refresh` (17882), `_close` (26835), `_open` (26856), `_forcemerge` (26956), `_cache/clear` (27189), `_search/template` (30044), `_terms_enum` (30610), and admin segments fsck (27034).

**(B) `Engine::get_index` (`engine.rs:3102`) resolves an alias to `aliased.first()`.** So these silently answer from ONE member: `_explain`, `_segments`, `_cat/segments`, `_eql/search`, `_async_search`, `_mget`, `_flush`. (`_delete_by_query` / `_update_by_query` are the same cause on a write path and are filed separately as #450.)

## Alias filters are ignored on four more paths

`msearch_impl` (`:22302`), `search_with_scroll` (`:19701`) and `msearch_template_impl` (`:30244`) each resolve aliases via `Engine::resolve_alias` and then search the members directly; none reads `index_alias_metadata`. `_cat/count` likewise. A filtered alias is the standard poor-man's document-level boundary, so this returns document bodies the alias exists to hide. Pre-existing — the base behaves the same — but it is the reason a single resolver has to return **the filters as well as the names**, not just names.

## Why this is one issue and not fifteen

The defect is not any individual site; it is that fifteen sites each decide independently what a selector means, and they disagree. #449 had to fix `search_impl` and `count_docs` separately for exactly that reason, and broke filtered aliases once in the process because a third place — the alias-filter block — keys on the name the user wrote rather than the expanded list.

[...]

PR #555, opened 2026-08-21
fix(es-compat): shared index-selector resolver + migrate the get_index->first read endpoints (#451 class B)

xerj-org:
Advances #451 (does not close it — see Scope). Introduces the single selector resolver the issue asks for and migrates the trivially-mergeable class-B read endpoints onto it, deleting their per-site `get_index` copies.

## The defect (class B)
Several read endpoints resolved the path index with `Engine::get_index`, which collapses an alias to `aliased.first()` — one member. So over a multi-index alias they silently answered from the first member only: a doc in another member was invisible.

## The shape (what the issue asked for)
`resolve_selector_with_filters(state, spec) -> (Vec<(name, Arc<Index>)>, Vec<alias_filter>)` — one resolver that expands `_all`/`*`/glob/alias/comma uniformly (via `resolve_index_selector`, #433) and carries any alias filter (mirroring `search_impl`, as `resolve_by_query_targets` does on the write path, #450). Every migrated site calls it; the per-site `get_index` is gone.

## Migrated this PR (each with a fail-before→pass repro test)
- **`_mget`** — a doc in a non-first member reported `found:false`; now searches every member.
- **`_explain`** — a non-first member's doc reported `matched:false`/"document not found"; now explains against the member that holds it, reporting the concrete member in `_index`.
- **`_cat/segments`** — an alias showed one row under the alias name (and globs matched nothing); now one row per member.
- **`_segments`** (`/{index}/_segments`) — one `indices` entry under the alias name; now one entry per member.

## Scope / tracked follow-up (why #451 stays open)
- **`eql_search`** — also `get_index→first`, but fanning it out needs proper **search-result merging** (sort + size across members), not per-member iteration; a naive concat would mis-order EQL events. Wants a decision on routing it through the shared multi-index search path.
- **Class C** (alias-filter injection on `msearch`/`search_with_scroll`/`msearch_template`/`_cat/count`) — the resolver now returns the filters; wiring them in is a separate change.
- `async_search_submit` is already alias-safe (delegates to `search_impl`).

Full `xerj-api` suite **469 passed, 0 failed**; `fmt --check`, `clippy --workspace --all-targets -D warnings`, `build --profile ci-test` all clean.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -33281,15 +33281,31 @@ pub async fn explain_doc(
     Path((index, id)): Path<(String, String)>,
     body: OptionalJson<Value>,
 ) -> impl IntoResponse {
-    let idx = match state.engine.get_index(&index) {
-        Ok(i) => i,
-        Err(e) => return ApiError::new(xerj_common::XerjError::from(e)).into_response(),
+    // Resolve the selector through the shared resolver (#451): an alias expands
+    // to every member, so the document is explained against the member that
+    // actually holds it. `get_index` collapsed an alias to `aliased.first()`,
+    // so a non-first member's doc reported `matched:false` / "document not
+    // found" (class B).
+    let (members, _filters) = match resolve_selector_with_filters(&state, &index).await {
+        Ok(t) => t,
+        Err(e) => return ApiError::new(e).into_response(),
     };
-
-    // Fetch the document.
-    let doc_source = match idx.get_document(&id).await {
-        Ok(Some(s)) => s,
-        Ok(None) => {
+    let mut resolved = None;
+    for (mname, midx) in &members {
+        match midx.get_document(&id).await {
+            Ok(Some(s)) => {
+                resolved = Some((mname.clone(), midx.clone(), s));
+                break;
+            }
+            Ok(None) => {}
+            Err(e) => return ApiError::new(xerj_common::XerjError::from(e)).into_response(),
+        }
+    }
+    // On a hit, `index` is rebound to the concrete member (reported in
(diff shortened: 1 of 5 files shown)


xerj-org:

**Verdict: 0 BLOCKING — all three refuted=false / severity=none.**

**correctness (none):** `resolve_selector_with_filters` delegates to `resolve_index_selector` (#433) which pushes every backing index, dedups, and propagates `get_index` errors. All 4 endpoints fan out to every member and report the concrete member, not the alias: `_mget` labels `_index` with `mname` (18432); `_explain` rebinds `index=mname` on hit (33302), keeps the selector name only on the not-found arm; `_cat/segments` rows keyed by member (29260); `_segments` keys `indices_map` + counts `_shards` per member (35808/35827). Plain-index path is byte-identical. Non-blocking notes: a single-member *alias* now reports the member in `_index` (a correct move toward ES, not a regression); alias filters intentionally discarded here (class-C follow-up).

**regression-risk (none):** Full `cargo test -p xerj-api` — **43 binaries, 469 passed, 0 failed**, exit 0. All adjacent alias/scroll tests pass. Clippy `-p xerj-api --all-targets -D warnings`: zero warnings. Could not prove any regression. Non-migration of `eql_search`/class-C is a documented scoping decision.

**evidence-holds (none):** Empirically confirmed the fail-before — **reverting `_explain` back to `get_index` makes its test FAIL** with `matched:false` / "document not found" (doc "2" lives in idx-b, the non-first member). No migration is vacuous; scoping is honest (`eql_search` still `get_index` at 32970, confirmed by grep + read).

Gate: (a) all checks green, none pending; (b) 0 BLOCKING; (c) CLEAN. Merging.

PR #556, opened 2026-08-21
test(es-compat): lock the filtered-alias by-query over-delete safety + _update_by_query aggregate deleted:0 (#553)

### LINKED ISSUE #553: test(es-compat): #450 fast-follow — lock the filtered-alias by-query slice (over-delete safety) + minor response consistency
Fast-follow to #450 / PR #552 (merged). The 3-skeptic verification of #552 confirmed the fix is correct (full suite 465/0; over-delete on a filtered alias is structurally impossible — the alias filter is ANDed as `bool.filter` into the query applied to every member). But two skeptics independently flagged that the **load-bearing safety property is verified only by code reading, not execution**:

## should-fix (do before we lean on it further)
- **Filtered-alias by-query test.** Neither test in `by_query_through_alias_touches_every_member.rs` exercises a *filtered* alias (0 `filter` occurrences). Add a `_delete_by_query` (and `_update_by_query`) test over a filtered alias asserting **only the in-slice docs are affected** — the exact #524 over-delete hazard the alias-filter logic exists to prevent.

## nit
- **`_update_by_query` aggregate omits `"deleted": 0`.** The single-member update body includes `"deleted": 0` (es_compat.rs:~24887) but `aggregate_by_query_results` for 2+ members does not — cosmetic wire asymmetry between 1- and 2+-member update responses. Add it for consistency.

## note (decide, don't necessarily change)
- A wildcard by-query spec matching **zero** indices now returns `200 {deleted:0}` instead of the old `404` (empty member list → clean-zero aggregate). This aligns with read-path semantics and is not data-loss, but confirm it's the intended contract.

Verbatim skeptic verdicts are in the PR #552 comment thread.

xerj-org:
Closes #553 (fast-follow to #450 / PR #552).

## The gap this closes
The 3-skeptic verification of #552 confirmed — by code inspection — that a `_delete_by_query` through a **filtered** alias removes only the slice the filter names (never the whole backing index, the #524 hazard), because the alias filter is ANDed into the query applied to every member. But two skeptics independently flagged that **no test exercises it**: the load-bearing over-delete-safety property was verified only by reading.

## The test
A `match_all` `_delete_by_query` through a **filtered 2-member alias** (`hot`, `filter: status=active`, over `idx-a`/`idx-b`, each holding one active + one archived doc) must:
- report `deleted: 2` (only the in-slice active docs), and
- leave each member's archived (out-of-slice) doc intact.

**Non-vacuous:** without the filter, `match_all` would delete all 4 and empty both members — so the test fails loudly the moment the alias filter is ever dropped. It passes on current `main` (the #450 fix is correct).

## Nit (also flagged by the skeptics)
`aggregate_by_query_results` now emits `"deleted": 0` for `_update_by_query`, matching the single-member update body — previously the 2+-member update response dropped a field the 1-member body has.

Full `xerj-api` suite **468 passed, 0 failed**; `fmt --check`, `clippy --workspace --all-targets -D warnings` clean. Test-only + a 6-line response-consistency fix.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -24456,6 +24456,13 @@ fn aggregate_by_query_results(mut results: Vec<Value>, count_key: &str) -> Value
     out.insert("timed_out".into(), json!(timed_out));
     out.insert("total".into(), json!(total));
     out.insert(count_key.to_string(), json!(affected));
+    // `_update_by_query` bodies also carry a `"deleted": 0` field (an update
+    // never deletes); the single-member response includes it, so the aggregate
+    // does too — otherwise the 2+-member update body drops a field the
+    // 1-member body has (#553 nit).
+    if count_key == "updated" {
+        out.insert("deleted".into(), json!(0));
+    }
     out.insert("batches".into(), json!(batches));
     out.insert("version_conflicts".into(), json!(version_conflicts));
     out.insert("noops".into(), json!(noops));
--- a/engine/crates/xerj-api/tests/filtered_alias_by_query_deletes_only_its_slice.rs
+++ b/engine/crates/xerj-api/tests/filtered_alias_by_query_deletes_only_its_slice.rs
@@ -1,0 +1,132 @@
+//! Issue #553 (fast-follow to #450): the #450 fix ANDs an alias's `filter` into
+//! the by-query so a `_delete_by_query` through a FILTERED alias removes only
+//! the slice the filter names — never the whole backing index (the #524
+//! over-delete hazard). The 3-skeptic verification of #552 confirmed this is
+//! correct by code inspection but flagged that no test EXERCISES it. This is
+//! that test: it fails loudly if the alias filter is ever dropped and a filtered
+//! by-query starts draining out-of-slice documents.
+//!
+//! Two members, each holding one in-slice (`status:active`) and one out-of-slice
+//! (`status:archived`) document, behind a filtered alias. A `match_all`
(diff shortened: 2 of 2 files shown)


xerj-org:

**Verdict: 0 BLOCKING — all three refuted=false / severity=none.** Two skeptics independently ran a counterfactual that broke the alias-filter path and confirmed the safety test flips to failure.

**correctness (none):** The test builds a truly FILTERED alias (`PUT /{member}/_alias/hot` with `filter: {term:{status:active}}`), spans 2 members, asserts `deleted==2` AND each member `_count==1` (keeps its archived doc). **Vacuity probe:** replacing the alias body with `{}` (unfiltered) made it FAIL with `deleted:4` — the filter is load-bearing. The nit (`es_compat.rs:24463-24465`) fires only when `count_key=="updated"`; the delete path (`count_key=="deleted"`) never enters it, so delete responses are unaffected.

**regression-risk (none):** Full `cargo test -p xerj-api` — **468 passed, 0 failed**; clippy `-D warnings` clean. The 6-line aggregate change is purely additive and guarded; the single-member update body already emits `"deleted":0`, so the 2+-member aggregate now matches it exactly. Zero regressions.

**evidence-holds (none):** **Counterfactual:** forced `resolve_by_query_targets` to ignore `alias_filters` → the `match_all` delete drained all 4 docs (`deleted:4, total:4`, both indices emptied), flipping the assertion — exactly the #524 hazard it guards. Restored → `deleted:2`, archived docs survive. Single-member responses byte-for-byte unchanged.

Gate: (a) checks green, none pending; (b) 0 BLOCKING; (c) CLEAN. Merging.

PR #557, opened 2026-08-21
fix(engine): dispatch a single-clause bool wrapping a semantic clause (#395 case B)

xerj-org:
Advances #395 (does not close it — see Scope). Fixes the clearest, unambiguous slice: a `bool` whose only clause is `semantic` silently returned zero hits.

## The defect (case B)
`peel_semantic_query` matched `SemanticSearch` under `Constant`/`Boosted`/`Named` but had **no `Bool` arm**, so `{"bool":{"must":[{"semantic":…}]}}` fell through to the lexical path and answered `200 OK` with **zero hits** — while the same clause at the top level returns all its documents. `peel_knn_query` already peels this exact single-clause-bool shape for `knn`, so `semantic` was simply the odd one out.

## The fix
Add the mirroring `Bool` arm to `peel_semantic_query`: a `bool` with exactly one `must`/`should` clause holding a `semantic` is dispatched to the vector path, with any `bool.filter` clauses ANDed into the semantic filter (byte-for-byte the logic `peel_knn_query` uses). No maintainer decision needed — this only makes `semantic` consistent with `knn`.

Fail-before: `bool{must:[semantic]}` → `0` hits vs `3` for the bare clause; now `3`.

## Scope / why #395 stays open
Only the **single-clause** case is fixed. The **multi-clause** cases still fall through: case D (`bool.should[match, semantic]`) and case G (the canonical ES 8.x `knn` block beside a `query`, folded to `bool.should[query, knn]`). Fusing vs rejecting those is the separate #395 decision — and importantly, G is a **valid ES 8.x hybrid shape**, so a blanket reject would break wire-compat; that path wants the dispatch/fuse treatment (`hybrid` already does it), which is a larger change.

New repro test passes (case B); 88 semantic/knn/hybrid tests green, 0 failures; `clippy --workspace --all-targets -D warnings` clean; `fmt --check` clean. Note: a combined `-p xerj-engine -p xerj-api` run flaked once on `painless_script_limits` (deep-recursion stack pressure, unrelated to this change — passes in isolation, 9/0).

### DIFF
--- a/engine/crates/xerj-api/tests/nested_vector_clause_in_bool_not_dropped.rs
+++ b/engine/crates/xerj-api/tests/nested_vector_clause_in_bool_not_dropped.rs
@@ -1,0 +1,128 @@
+//! Issue #395: a `semantic`/`knn` clause nested inside a `bool` is silently
+//! dropped — the engine only *peels the top* of the query tree, so a vector
+//! clause that is not at the top (or is one of several `bool` clauses) never
+//! reaches the vector path. The request answers `200 OK` with the purely lexical
+//! result set, no error and no `_shards.failed`.
+//!
+//! Case B from the report: a `bool` whose only clause is `semantic` returns zero
+//! hits, while the same clause at the top level returns all three documents.
+//! Whatever the resolution (dispatch the nested clause, or reject it loudly),
+//! `bool{must:[semantic]}` must not silently answer `200`/zero.
+//!
+//! Elasticsearch is referenced for wire semantics only; no ES code is
+//! reproduced here.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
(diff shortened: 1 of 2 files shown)


xerj-org:
## Verification (2 independent skeptics + a direct regression run)

**Verdict: 0 BLOCKING.** Two independent skeptics refuted-false/none; the 3rd (regression) agent hung in finalization, so — per the exit-on-no-BLOCKING rule — I substituted a direct regression run rather than wait indefinitely.

**correctness (refuted=false / none):** The new `Bool` arm in `peel_semantic_query` correctly dispatches a single-clause `bool[semantic]` and merges `bool.filter` into the semantic filter, mirroring `peel_knn_query`. Multi-clause bools return `None` (unchanged) and `must_not` returns `None`. No over-matching.

**evidence-holds (refuted=false / none):** Non-vacuous — reverting the new `Bool` arm makes the test fail (`bool{must:[semantic]}` → 0 hits vs 3 for bare). Scoping is honest: multi-clause bools (cases D/G) are NOT dispatched by this arm, so the PR doesn't overclaim.

**Direct regression run (rustc 1.98.0), substituting for the hung agent:**
- `cargo test -p xerj-api` — **exit 0** (full integration suite green, incl. the new repro test).
- `cargo test -p xerj-engine --lib` — **exit 0**.
- `cargo clippy --workspace --all-targets -D warnings` — **clean**.
- `painless_script_limits` stack-overflowed on the default local stack — **but it also stack-overflows on `main` (d11d255) without this change**, so it is a pre-existing, stack-size-sensitive local flake, unrelated to query-peeling (CI, which sets a larger stack, reports `fail=0`).

Gate: checks green (CI `fail=0`), 0 BLOCKING, CLEAN. Merging.

xerj-org:
**3rd skeptic (regression-risk) completed** (it wasn't stuck — ~18 min running the combined suite twice + painless 3×) and **independently confirmed 0 BLOCKING**: "No regression. 1485 pass, 0 fail, clippy clean. The painless_script_limits stack overflow is a pre-existing nondeterministic flake — a *different* call-depth test aborts each run and it reproduces in isolation — not caused by this change." All three skeptics agree; the merge stands.

PR #559, opened 2026-08-21
fix(es-compat): apply alias filters on _msearch (#451 class C)

xerj-org:
Advances #451 (does not close it — see Scope). Fixes the `_msearch` half of failure class C: a filtered alias's document boundary was bypassed.

## The defect (class C)
`msearch_impl` resolved an alias to its members via `resolve_alias` and searched them directly, **never reading `index_alias_metadata`** — so an alias's `filter` was dropped. A filtered alias is the standard poor-man's document-level boundary, so `_msearch` **returned document bodies the alias exists to hide**, while `_search` through the same alias honours the filter.

Reproduced: a filtered alias (`filter: status=active`) over an index with one active + one archived doc — `_search` → 1 hit; `_msearch` → **2 hits** (the archived out-of-slice doc leaked).

## The fix
AND the alias `filter` into each sub-search body before parsing, **mirroring the existing `pit.index_filter` injection** a few lines above (same `bool{must:[query], filter:[…]}` merge) and the exact filter-gathering `search_impl` / `resolve_by_query_targets` use (keyed on the written alias name). Fail-before `2` → now `1`.

## Scope / why #451 stays open
Fixes `_msearch`. `search_with_scroll`, `msearch_template`, and `_cat/count` share the same class-C gap (they resolve via `resolve_alias` and don't read `index_alias_metadata`) — those follow separately. (The class-B `get_index` read endpoints were fixed in #555; `eql_search` fan-out remains the #395/#451 dispatch decision.)

Full `xerj-api` suite **473 passed, 0 failed**; `fmt --check`, `clippy --workspace --all-targets -D warnings` clean. New repro test included.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -22518,6 +22518,44 @@ async fn msearch_impl(
             }
         }
 
+        // AND in any alias `filter` before parsing (#451 class C). msearch
+        // resolved the alias to its members and searched them directly, never
+        // reading `index_alias_metadata`, so a filtered alias — the standard
+        // poor-man's document boundary — leaked the documents it exists to hide
+        // (`_search` through the same alias honours it). Gathered keyed on the
+        // written alias name, exactly as `search_impl` and
+        // `resolve_by_query_targets` do.
+        {
+            let mut alias_filters: Vec<Value> = Vec::new();
+            for part in index_name
+                .split(',')
+                .map(str::trim)
+                .filter(|s| !s.is_empty())
+            {
+                if let Some(entry) = state.engine.aliases.get(part) {
+                    for backing in entry.value().iter() {
+                        if let Some(meta) = state.engine.index_alias_metadata.get(backing) {
+                            if let Some(filter) =
+                                meta.get(part).and_then(|v| v.get("filter")).cloned()
+                            {
+                                alias_filters.push(filter);
+                            }
+                        }
+                    }
+                }
+            }
+            if !alias_filters.is_empty() {
+                let existing = effective_body.get("query").cloned();
+                let merged = match existing {
+                    None => json!({ "bool": { "filter": alias_filters } }),
+                    Some(q) => json!({ "bool": { "must": [q], "filter": alias_filters } }),
+                };
+                if let Some(obj) = effective_body.as_object_mut() {
+                    obj.insert("query".to_string(), merged);
+                }
+            }
+        }
+
         // Strip _index constraints BEFORE parsing so downstream FTS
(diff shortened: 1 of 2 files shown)


xerj-org:
Scope update: added a reproduce-first check of `search_with_scroll` — despite the issue's static analysis, a scroll opened through a filtered alias **already returns only the in-slice slice** (the initial scroll search delegates through the filter-applying path). Committed as a passing, non-vacuous regression guard (a leak would make total=2). So this PR: **_msearch = fixed**, **scroll = confirmed-safe + guarded**. Remaining class-C to check: `msearch_template` and `_cat/count`.

xerj-org:
Scope update: **#451 class C is now complete** in this PR. Reproduce-first confirmed and fixed all leaking sites — `_msearch`, `_cat/count`, and `_msearch/template` now honour a filtered alias's document boundary; `search_with_scroll` was already safe and is guarded. Each has a fail-before→pass repro. All 4 tests green under 1.98.0; fmt/clippy clean.

xerj-org:

**Verdict: 0 BLOCKING — correctness=nit, regression-risk=nit, evidence-holds=none.**

**correctness (nit):** All three sites AND the alias filter correctly — `msearch_impl`/`msearch_template_impl` gather from the written alias name and merge into the body **before parse** (`bool{must:[q],filter:[…]}`), keyed exactly like `search_impl`; `idx.search` per member so no double-application. `cat_count` keeps the fast `stats().doc_count` path when unfiltered and only runs a filtered `size:0` count for a filtered alias. *Nit:* a mixed selector `hot,idx-b` (filtered alias + plain index) applies hot's filter across idx-b too — but this **mirrors existing `search_impl`/`_count` behavior**, not a new regression.

**regression-risk (nit):** Full `cargo test -p xerj-api` — **477 passed, 0 failed, 47 suites**; clippy clean. Every change is gated behind `if !alias_filters.is_empty()`, so plain-index/unfiltered paths are byte-for-byte unchanged. *Non-blocking caveats:* (1) the mixed-selector over-filter above; (2) `cat_count` contributes 0 for a member if `parse_request`/`search` errors — an undercount (safe direction for a filter, filtered-alias only), vs the old over-count.

**evidence-holds (none):** **Mutation-verified** — neutralizing the msearch / msearch_template / cat_count injections each makes its test fail (leak/over-count 2 vs 1); the scroll test passes independently (a genuine non-vacuous guard confirming `_search?scroll=` already applies the filter via `search_impl` — honest "already-safe" scoping). *Nit:* the PR prose credits `search_with_scroll`, but the `_search?scroll=` test dispatches to `search_impl`.

Gate: checks green, 0 BLOCKING, CLEAN. Merging; nits filed as follow-up.

PR #561, opened 2026-08-21
fix(es-compat): fan _eql/search out to every alias member (#451, last class-B read site)

xerj-org:
Advances #451 (does not close it). Fixes the last `Engine::get_index -> aliased.first()` read site.

## The defect
`eql_search` resolved the path index with `get_index`, which collapses an alias to `aliased.first()` — one member. So an EQL search through a multi-index alias silently returned **only the first member's events**, the same silent wrong-answer as `_mget`/`_explain`/`_cat/segments`/`_segments` (fixed in #555).

Reproduced: a 2-member alias `tri`, one matching event per member, `any where status == "active"` → **1 event** (idx-a only), should be 2.

## The fix
Resolve via the shared `resolve_selector_with_filters`, search each member, merge the event lists, and report the concrete member in `_index`; the EQL `size` limit is applied across the merged results. Fail-before `1` → now `2`.

## Scope
Single-event EQL is all XERJ implements (`sequence`/`sample` already 400). `eql_to_query` adds no explicit cross-member sort, so events are concatenated per member then size-truncated — **strictly better than the silent drop** (the events were previously invisible); a strict global `@timestamp` ordering across members would be a refinement, not part of this fix.

Full `xerj-api` suite **474 passed, 0 failed**; `fmt --check`, `clippy --workspace --all-targets -D warnings` clean. New repro test included.

With this + #555 (class B) + #559 (class C), #451's alias read/filter divergence is resolved; only the multi-clause vector-dispatch shared with #395 remains.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -33189,21 +33189,30 @@ pub async fn eql_search(
     // Translate EQL -> xerj DSL, then run it through the normal search path.
     let inner_query = eql_to_query(&eql);
     let size = body.get("size").and_then(Value::as_u64).unwrap_or(10);
-    let search_body = json!({ "query": inner_query, "size": size, "from": 0 });
+
+    // Fan out to every member of a multi-index alias (#451 `_eql` arm):
+    // eql_search used Engine::get_index -> aliased.first(), silently returning
+    // only the first member's events. Search each member and merge the event
+    // lists, reporting the concrete member in `_index`. Resolve BEFORE building
+    // the request so any alias `filter` is ANDed into the query: a *filtered*
+    // alias is a document boundary (#451 class C, commit 170f41f2 closed this
+    // exact leak for _search/_msearch), so EQL must honour it too — fanning out
+    // unfiltered would leak the hidden members' rows.
+    let (members, filters) = match resolve_selector_with_filters(&state, &index).await {
+        Ok(t) => t,
+        Err(e) => return ApiError::new(e).into_response(),
+    };
+    let effective_query = if filters.is_empty() {
+        inner_query
+    } else {
+        json!({ "bool": { "must": [inner_query], "filter": filters } })
+    };
+    let search_body = json!({ "query": effective_query, "size": size, "from": 0 });
 
     let req = match xerj_query::parse_request(&search_body)
         .map_err(|e| xerj_common::XerjError::invalid_query(e.to_string()))
     {
         Ok(r) => r,
-        Err(e) => return ApiError::new(e).into_response(),
-    };
-
-    // Fan out to every member of a multi-index alias (#451 `_eql` arm):
-    // eql_search used Engine::get_index -> aliased.first(), silently returning
-    // only the first member's events. Search each member and merge the event
-    // lists, reporting the concrete member in `_index`.
-    let (members, _filters) = match resolve_selector_with_filters(&state, &index).await {
-        Ok(t) => t,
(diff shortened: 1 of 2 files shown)


xerj-org:
### Verification — 3-skeptic adversarial pass (rustc 1.98.0, isolated worktrees): 1 BLOCKING → reworking

- **Correctness — BLOCKING.** The fan-out mechanics are correct (all members expanded, `total` summed, `truncate` applied once after merge, `size as usize` lossless on 64-bit). BUT the fix binds `let (members, _filters) = …` and **discards `_filters`**, so `_eql/search` through a *filtered* alias returns events the filter must hide. Reproduced by the skeptic: filtered alias `tenant_a` (filter `tenant==a`) over members `ta`(tenant=a)/`tb`(tenant=b) → `_search` = **1 hit** (filter honored) but `_eql/search` = **2 events**, leaking the `tenant=b` doc. `es_compat.rs:33205` discards the filter; the per-member loop runs the raw `req`. Every other read path (`search_impl`, `resolve_by_query_targets`) ANDs the alias filter in — EQL must too.
- **Test-validity / fail-before — NON_BLOCKING (0 blocking).** Fail-before proven: reverting only `es_compat.rs` makes the repro fail (`1` vs `2`); restored passes. Coverage gaps noted: test never asserts `_index` reports the member, nor exercises size-truncation-across-members.
- **Regression skeptic** was still running at rework time; moot since the code is changing — a fresh full verification follows the fix.

**Rework:** apply the alias `filter` to the EQL query exactly as `resolve_by_query_targets` does (`{bool:{must:[eql_query],filter:alias_filters}}`), plus a filtered-alias-leak repro with a proven fail-before. The cross-member `@timestamp` ordering-before-truncate refinement (skeptic NON_BLOCKING) stays out of scope — will file a follow-up.

xerj-org:
### Fresh verification of the reworked commit (51ad3185) — 3 skeptics, rustc 1.98.0, isolated worktrees: **0 BLOCKING**

- **Correctness — NON_BLOCKING.** Re-read the whole `eql_search`. `filters.is_empty()` ⇒ `effective_query == inner_query` with no bool-wrap, so plain index and unfiltered alias are byte-identical to before (the fan-out test still returns both members). A 2-member filtered alias pushes the term filter twice; ANDing `A AND A` is idempotent + non-scoring. The single `req` is run against **every** member, so the filter applies to all — verified by placing the out-of-slice doc in **member 2** (idx-b) and seeing it hidden. EQL is non-scoring, so the bool-wrap only restricts the match set.
- **Regression — CLEAR.** Full `xerj-api` suite **479 passed / 0 failed** across 47 binaries; `clippy --workspace --all-targets -D warnings` clean. Single-index/nonexistent-index paths behaviour-preserved. Could not construct any regression.
- **Evidence / fail-before — CLEAR.** Revert only `es_compat.rs` → both tests fail; restore → pass. The filter-ANDing is independently load-bearing: disabling ONLY the filter branch (keeping fan-out) reproduces the exact leak (2 events, idx-b `status=deleted` present). The leak test can't false-pass — both docs share `host==h1`, only the alias filter excludes idx-b.

**NON_BLOCKING follow-ups (do not block):** cross-member `@timestamp` ordering-before-truncate → tracked in #567. Heterogeneous per-member alias filters are ANDed uniformly across all members — a pre-existing systemic behaviour that exactly mirrors `resolve_by_query_targets`/`search_impl`, not introduced here.

Gate: (b) 0 BLOCKING ✓. Merging once (a) all checks green none pending and (c) mergeStateStatus CLEAN — currently 2 checks still in progress.

PR #562, opened 2026-08-21
fix(engine,es-compat): carry snapshot seq_no/version into collapse inner_hits (#506)

### LINKED ISSUE #506: scroll: collapse inner_hits still does a live version lookup + positional _seq_no (torn on collapsed scroll); plus a fallback-ordinal footgun
Two non-blocking follow-ups surfaced by the #499 verification (scroll `_seq_no`/`_version` torn-read fix, now merged). Neither blocked that PR; recording so they are not lost.

## 1. Collapse `inner_hits` still returns a torn/synthetic `_seq_no`/`_version` on a collapsed scroll

`engine/crates/xerj-api/src/es_compat.rs:13163` — the `emit_version_ih` branch calls `idx.lookup_version(id_str)` **live** at render time, and fabricates `_seq_no` as the array index (`json!(i as u64)`). So a collapse + scroll can hand an inner hit a live `_version` beside a snapshot `_source`, and a positional `_seq_no` that is not the doc's real sequence number.

This is the same class #499 just fixed for top-level scroll hits, on a different path. It pre-dates #499 (identical on `main`) and #499 explicitly scoped it out. Worth fixing the same way: carry the inner hit's snapshotted seq_no/version instead of a render-time lookup + positional index.

## 2. Fallback ordinal in the scroll/first-page renderer

`engine/crates/xerj-api/src/es_compat.rs:20155` (and the first-page site ~19819) render `_seq_no` as `h.seq_no.unwrap_or(hit_idx as u64)`. If a snapshot hit ever carried `seq_no = None` while `emit_seq_no` is true, this silently emits a page-local ordinal (0 for each page's first hit) instead of a real or absent seq_no.

The current mechanism should never produce `None`-with-emit (the engine's catch-all fills every hit, and disabled sequence numbers use the `-2` sentinel, not `None`), so this is defensive rather than a live bug. But an ordinal masquerading as a seq_no is a bad failure mode if the invariant ever breaks — better to emit `null` (or the `-2` sentinel) than a fabricated value. Consider asserting the invariant or rendering `null` on `None`.

Both found by the independent verification of #499.

xerj-org:
Closes #506(1) (the torn-read half). Follow-up to the merged #499 (top-level scroll `_seq_no`/`_version` torn-read fix), on the `inner_hits` path.

## The defect
Collapse `inner_hits` fabricated `_seq_no` as the array index (`json!(i as u64)`) and re-read `_version` **live** at render time (`get_index` + `lookup_version`), instead of the doc's snapshotted values (es_compat.rs ~13198/13207). On a collapsed scroll that hands an inner hit a positional `_seq_no` (0 for each group's first member) and a live `_version` beside a snapshot `_source` — the same torn-read class #499 fixed for top-level hits.

## The fix
The collapse group is built in the engine (`index.rs` `__xy_collapse_group__` embedding). Carry each member's `seq_no`/`version` there, and read them in the api renderer instead of the live lookup + array index.

**Fail-before** (proven by reverting the source): inner_hits `_seq_no` = `[0, 1, 2]` positional — the document indexed **last** (highest real seq_no) received `_seq_no: 0`. After the fix each carries its real snapshot seq_no.

## Scope
Fixes #506(1). #506(2) — the defensive `h.seq_no.unwrap_or(hit_idx)` fallback in the top-level scroll renderer — is a non-live footgun the issue flags separately (the engine never emits `None`-with-emit today); left as a tracked follow-up.

Full `xerj-api` suite green, `xerj-engine --lib` green, `clippy --workspace --all-targets -D warnings` clean. New repro test with a proven fail-before.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -13192,19 +13192,21 @@ async fn search_impl(
                                 hit_obj.insert("_score".to_string(), score);
                             }
                             if emit_version_ih {
-                                // Real per-doc `_version` (external map
-                                // wins) so inner_hits match top-level hits.
-                                let id_str = m.get("_id").and_then(Value::as_str).unwrap_or("");
-                                let v = state
-                                    .engine
-                                    .get_index(&idx_name)
-                                    .ok()
-                                    .and_then(|idx| idx.lookup_version(id_str))
-                                    .unwrap_or(1);
+                                // Snapshotted `_version` carried in the collapse
+                                // group (#506) — NOT a render-time live lookup,
+                                // which could pair a live `_version` with the
+                                // snapshot `_source` on a collapsed scroll.
+                                let v = m.get("_version").and_then(Value::as_u64).unwrap_or(1);
                                 hit_obj.insert("_version".to_string(), json!(v));
                             }
                             if emit_seq_no_ih {
-                                hit_obj.insert("_seq_no".to_string(), json!(i as u64));
+                                // The doc's real snapshotted seq_no, not the
+                                // array index (#506).
+                                let sn = m
+                                    .get("_seq_no")
+                                    .and_then(Value::as_u64)
+                                    .unwrap_or(i as u64);
+                                hit_obj.insert("_seq_no".to_string(), json!(sn));
                                 hit_obj.insert("_primary_term".to_string(), json!(1));
                             }
                             if let Some(src) = m.get("_source").cloned() {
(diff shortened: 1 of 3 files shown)


xerj-org:
### Merge verification — 3-skeptic adversarial pass (rustc 1.98.0, isolated worktrees)

**Verdict: 0 BLOCKING → merging.**

- **Correctness & type** — NON_BLOCKING. Compiles clean on 1.98.0; engine embeds real snapshotted `seq_no`/`version` and the renderer reads the same keys with matching types; the torn read (live `_version` + positional `_seq_no` beside snapshot `_source`) is eliminated on the normal path (test yields `[2,1,0]`, not `[0,1,2]`). One follow-up: the `None`-snapshot delete-race can still hit the `unwrap_or(i)` fallback — **not a regression** (pre-PR that path was always positional) → tracked in #566.
- **Test-validity / fail-before** — NON_BLOCKING, fail-before **proven**: reverting only the two source hunks makes `collapse_inner_hits_seqno_is_real.rs` fail with the tell-tale `_seq_no [0,1,2]`; restoring the PR makes it pass. Two coverage gaps (`_version` half, scroll path) → #566.
- **Regression** — the workflow skeptic returned a stub, so I ran the full regression directly under **rustc 1.98.0**: `clippy --workspace --all-targets -D warnings` clean, `xerj-engine --lib` **610 passed / 0 failed**, `xerj-api --no-fail-fast` **exit 0** (zero failures). Corroborates CI's green Build + Test.

NON_BLOCKING findings tracked in #566. Merging.

PR #513, opened 2026-08-19
tools: reproducible 3-arm Harbor benchmark harness (Claude Code vs XERJ vs tuned)

xerj-org:
The full working harness behind field report PR #512, under `tools/benchmarks/harbor-3arm/` — agents, prompt templates, cell runner with resume, aligned stage-1 subset, and a collector that enforces complete-matrix comparison. Portable: no sandbox paths; prerequisites and every measured trap documented in the README.

Issue #500, opened 2026-08-19
reference-coding: symbol index returns whole class/method bodies and does not index declarations as their own symbols — 32-48x more bytes than grep, verified


xerj-org:
A fair, deterministic retrieval test of the reference-coding claim (grep vs XERJ symbol search over Apache Lucene, 12 verifiable questions, precise symbol queries, coverage gaps held out) found XERJ ingests **32–48x more bytes per answered question than grep** and loses recall **8/9 vs 12/12**. This is the primary use case, so it is worth fixing rather than hiding. The cause is structural and specific.

## Measured (refsym, verified by hand)

| | grep | XERJ refsym |
|---|---|---|
| median bytes / answered question | ~46–52 | ~1,459–2,189 |
| ratio | 1x | **32–48x** |
| recall on answerable | 12/12 | 8/9 (one rank-35 miss) |
| exact-line citation | 12/12 | coarse: an enclosing span merely covers the truth line |

## Root cause: the retrieved unit is the enclosing class/method, and declarations are not their own symbols

Three effects, all reproduced:

1. **Whole-symbol payloads.** The smallest retrievable unit is the enclosing class or method body. A one-line constant costs the whole parent:
   - `DEFAULT_MAX_CONN = 16` (one line, 52 bytes by grep) is only reachable inside the `HnswGraphBuilder` class body (~2,232 bytes), and even that ranks below three constructor overloads.

2. **Constants and fields are not indexed as their own symbols.** Verified against refsym:
   ```
   term name=DEFAULT_BEAM_WIDTH  -> count 0
   term name=DEFAULT_MAX_CONN    -> count 0
   ```
   A search for `DEFAULT_MAX_CONN` returns unrelated `testToString` methods; the declaration itself is not a document. So a field-level fact only survives folded inside a coarse parent, or not at all. This directly caused the one recall miss (`DEFAULT_BEAM_WIDTH`, answer symbol at rank 35 of 38).

   Note the AST extractor *does* emit const/static symbols with line numbers (a default `xerj autoindex` of `governor.rs` produced `symbols: [{name:GOVERNOR,kind:static,line:38},{name:SAMPLE_INTERVAL_MS,kind:const,line:45},...]`, 77 of them). They exist in the sidecar but are not promoted to retrievable, independently-ranked documents. The information is extracted and then thrown away for retrieval.

[...]

PR #564, opened 2026-08-21
bench(harbor-3arm): flag partial arm coverage + neutralize the tuned prompt's pre-asserted claim (#516)

### LINKED ISSUE #516: harbor-3arm benchmark: collector should assert the full 3-arm set; tuned prompt pre-asserts the measured outcome
Two non-blocking follow-ups from the #513 verification (Harbor 3-arm benchmark harness, merged). Both are about keeping the harness honest, which matters given the fair-test finding in #500.

## 1. Collector doesn't assert the canonical 3-arm set (should-fix)

`tools/benchmarks/harbor-3arm/collect_results.py:106-111` builds the cross-arm comparison from whatever arm directories are present, guarded only by `if len(arms) < 2: continue`. If an arm dir is entirely absent it silently emits a 2-arm comparison.

It does **disclose** the count ("… finished in all 2 arms") rather than fabricating a 3-arm win, and the 2-arm comparison it emits is still task-aligned — which is why this is should-fix, not blocking. But a reader skimming the output could miss that the `xerj` (untuned) arm was missing and read a `native` vs `xerj-tuned` result as the headline.

Fix: assert `present_arms == {native, xerj, xerj-tuned}` (or flag partial arm coverage prominently) before printing a comparison, so an incomplete matrix is loud, not a footnote.

## 2. The tuned prompt pre-asserts the outcome under test (nit)

`tools/benchmarks/harbor-3arm/xerj_prompt_tuned.j2:5-6` tells the agent XERJ "is usually MUCH faster than grep for finding the right definition in an unfamiliar repo."

It's steering text inside the explicitly-labeled `xerj-tuned` arm (a disclosed favorable playbook), and it's a latency claim, not the byte/token-efficiency #500 measured — so it isn't a false *published* result and doesn't block. But a benchmark whose own prompt pre-asserts the thing it's measuring invites exactly the skepticism the WordPress/fair-test work earned trust by avoiding. Given #500 found XERJ returns 32–48x more bytes than grep on symbol lookups, softening this to a neutral capability description ("XERJ can return the enclosing definition directly") would keep the tuned arm's advantage without prejudging the result.

Both found by the independent verification of #513.

xerj-org:
Closes #516. Two benchmark-honesty follow-ups from the #513 verification — keeping the Harbor 3-arm harness honest (which matters given the fair-test finding in #500).

## 1. Collector: loud partial-arm-coverage flag (was should-fix)
`collect_results.py` built the cross-arm comparison from whatever arm dirs were present, guarded only by `if len(arms) < 2`. So an entirely-absent arm silently produced a **2-arm** comparison, and a reader could mistake e.g. `native` vs `xerj-tuned` for the headline. Now the header names the present arms and prints a loud `⚠ PARTIAL ARM COVERAGE` warning (naming the missing/unexpected arms) whenever the set isn't the canonical `{native, xerj, xerj-tuned}` — an incomplete matrix is loud, not a footnote.

## 2. Tuned prompt: neutral capability, not a pre-asserted outcome (was nit)
`xerj_prompt_tuned.j2` told the agent XERJ "is usually **MUCH faster than grep** for finding the right definition" — a latency claim inside the very arm whose latency is measured. Softened to a neutral capability description ("can return the enclosing definition directly, so you rarely need to open files"), keeping the tuned arm's disclosed playbook advantage without prejudging the result. Especially apt given #500 measured XERJ returning 32–48x more bytes than grep.

## Verification
`python3 -m py_compile collect_results.py` passes; tooling-only (Python + Jinja), no engine change.

### DIFF
--- a/tools/benchmarks/harbor-3arm/collect_results.py
+++ b/tools/benchmarks/harbor-3arm/collect_results.py
@@ -103,12 +103,22 @@ def main():
     for (bench, model, arm), tasks in cells.items():
         by_bm[(bench, model)][arm] = {
             t for t, v in tasks.items() if v["reward"] is not None}
+    canonical_arms = {"native", "xerj", "xerj-tuned"}
     for (bench, model), arms in sorted(by_bm.items()):
         if len(arms) < 2:
             continue
+        present = set(arms)
         common = set.intersection(*arms.values())
+        # A comparison that isn't the full 3-arm set must say so LOUDLY — an
+        # absent arm otherwise reads as a footnote and a reader can mistake, e.g.,
+        # native vs xerj-tuned for the headline result (#516).
+        coverage = "" if present == canonical_arms else (
+            f"   ⚠ PARTIAL ARM COVERAGE — missing "
+            f"{sorted(canonical_arms - present) or 'nothing'}, "
+            f"unexpected {sorted(present - canonical_arms) or 'nothing'}; "
+            f"this is NOT the canonical 3-arm comparison")
         print(f"\n{bench} / {model}: {len(common)} tasks finished in all "
-              f"{len(arms)} arms")
+              f"{len(arms)} arms ({', '.join(sorted(present))}){coverage}")
         for arm in sorted(arms):
             tv = cells[(bench, model, arm)]
             k = sum(1 for t in common if tv[t]["resolved"])
--- a/tools/benchmarks/harbor-3arm/xerj_prompt_tuned.j2
+++ b/tools/benchmarks/harbor-3arm/xerj_prompt_tuned.j2
@@ -2,8 +2,9 @@
 
 ---
 TOOL NOTE: XERJ code search is installed (server at http://localhost:9210) and
-tuned for this work. It is usually MUCH faster than grep for finding the right
-definition in an unfamiliar repo. Playbook:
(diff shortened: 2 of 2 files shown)


xerj-org:
### Merge verification — direct adversarial review (proportionate: tooling-only, zero engine surface)

**Verdict: 0 BLOCKING → merging.** This is a 14-line Python + Jinja change with no engine impact and CLEAN full-matrix CI, so the gate is a careful direct read rather than the code-path 3-skeptic worktree workflow.

- **collect_results.py** — the `canonical_arms = {native, xerj, xerj-tuned}` guard is correct: `present == canonical_arms` prints no warning; any other set prints the loud `⚠ PARTIAL ARM COVERAGE` naming `sorted(canonical - present)` (missing) and `sorted(present - canonical)` (unexpected), with `or 'nothing'` guarding empty sides. The `len(arms) < 2` skip is preserved (a single arm is never printed as a comparison). `python3 -m py_compile` passes.
- **xerj_prompt_tuned.j2** — the biasing latency claim ("usually MUCH faster than grep") inside the very arm whose latency is measured is replaced with a neutral capability description. Honest, and keeps the tuned arm's disclosed playbook advantage.

No Rust surface touched → engine build/tests unaffected. Merging.

PR #460, opened 2026-08-17
chore(release): 1.0.0-rc.18

xerj-org:
Seven changes merged today and main is green with its own completed CI run (`e19f813`), so the release condition is met.

## What rc.18 carries over rc.17

| PR | |
|---|---|
| **#441** | the air-gapped recipe **failed open** — a bad digest extracted and installed anyway, and re-pasting without a top-level `set -eu` (the recovery the page itself named) made the digest check non-fatal |
| **#452** | the install page ran `sha256sum -c`, which verifies whatever filenames the `.sha256` lists and never checks that any line names the archive being extracted — a decoy `.sha256` passed while the archive was **never hashed** |
| **#442** | a GIF whose comment chain outruns the sniff budget is decided by the chain, not a canvas heuristic |
| **#455** | the release badge rendered "no releases found"; capability badges corrected (default embedder is lexical and offline, neural is opt-in) |
| **#457** | the README led with reference coding — one use case — instead of `autoindex` |
| **#447** | operational-recipe verification protocol (@buger) |
| **#448** | XERJ vs Lucene 10.3.1 comparison (@buger) |

**Two of those are security fixes on the paths a user follows to obtain and install the software.** That is the reason to cut this now rather than wait for more to accumulate.

## Why the bump is required, not cosmetic

The release workflow derives archive names from the tag ref (`GITHUB_REF_NAME`) and does **not** cross-check them against `Cargo.toml`. Tagging without this commit would publish archives named `rc.18` containing a binary that reports `rc.17` — exactly the kind of drift that makes a checksum-verified install pointless to reason about.

## Verified on this commit under rustc 1.97.1

```
xerj --version              ->  xerj v1.0.0-rc.18
cargo fmt --all --check     ->  clean
cargo clippy --workspace --all-targets -- -D warnings  ->  clean
xerj-api suite (ci-test)    ->  0 failed suites
xerj-autoindex lib          ->  678 passed, 0 failed
Cargo.lock                  ->  16 entries at 1.0.0-rc.18 (all crates inherit [workspace.package])
```

## After merge

Tag `v1.0.0-rc.18` on the merge commit; the release workflow builds the eight-target matrix and publishes. I will not tag until this is merged and main is green on its own run — tagging a commit whose CI has not completed is how a release discovers a broken target for the first time at release time.

## Not verified

The cross-compilation matrix (musl/aarch64/Windows/macOS) — that is what the release workflow's push-to-main artifacts-only build exists to catch, and it runs on the merge commit before any tag exists.

[...]

PR #565, opened 2026-08-21
fix(site): document the built-in neural embedder + correct stale tally & dead links (#492)

### LINKED ISSUE #492: site: the RAG recipe's published copy denies the built-in neural embedder, badges cite a page that never mentions them, plus four dead links
Found by the systematic divergent-copy and link sweep during the #460 verification. None of these block the rc.18 notes — they are site and doc defects that outlive any one release — but several are user-facing and one contradicts a shipped capability.

## The published site denies a capability that ships

`landing/docs/recipes/semantic-search-rag.html:211-221` vs `docs/recipes/semantic-search-rag.md:141-193`.

The markdown documents three embedding backends, including a built-in neural one in the binary — `--embed-mode neural`, all-MiniLM-L6-v2, ~90 MB on first use. The published HTML has **zero** occurrences of `--embed-mode`, `embedding.mode`, `embedding.local_model_dir` or the download, and tells the reader the route to real semantics is to "wire an external embeddings endpoint".

So the page a prospective user actually reads says the product cannot do offline semantic search without third-party infrastructure. That is the single most consequential item here: it costs adoption from exactly the audience the recipe targets.

## The rc.18 badges point at a page that never mentions them

`README.md:41-42` ships two new badges — `default embedder-lexical, offline` and `neural embedder-opt-in, downloads ~90 MB` — both linking to `https://xerj.org/docs/vectors.html`. That page, in the repo and live, contains **zero** occurrences of "neural", "embedder", "lexical" or "90 MB". A reader who clicks the badge to check the claim finds nothing about it.

## A remediation instruction names an undocumented key

The rc.18 memory-cap entry tells the reader: "If you index a large project and see 429 `circuit_breaking_exception`, raise `limits.max_process_memory_mb`."

That key appears in no shipped doc and not in `engine/xerj.default.toml`. It exists only in Rust source, so a user following the instruction has nowhere to look up its units, default or valid range. Either document it in the default TOML and `landing/docs/config.html`, or point the instruction at something that is documented.

## One experiment, two published results

`landing/case-studies.html:249-250` says **15/15 vs 1/15** "across five languages". The case-study page, `llms.txt`, `llms-full.txt` and the source markdown all say **21/21 vs 1/21**. 15/15 is the earlier four-domain, Rust-only cut, so the hub page has a stale tally attached to the expanded scope.

## Dead links

- `landing/resources/index.html:194` — "DOWNLOAD THE FULL KIT · ZIP" → `/resources/xerj-sales-kit.zip`, which does not exist. Live it returns **HTTP 200 with `text/html`** (the homepage), so the visitor silently saves an HTML page as a `.zip` instead of seeing a 404.
- `landing/case-studies/reference-coding.html:319` — the "GET XERJ" button, on the page the README's headline `reference coding` badge links to, points at `/get/index.html`. `landing/get` is a regular file (the installer script), not a directory, so this falls through to the 200 HTML fallback and lands the visitor on the homepage.
- `landing/demo/index.html:431` — the only broken `blob/main` link of the 40 the site publishes.
- `landing/demo/index.html:286` — `href="#step-6b"` with no matching id; the only broken fragment on the site.

The 200-instead-of-404 behaviour is worth its own look: it converts every future broken path into a silent wrong-content success, which is why two of these survived unnoticed.

## Prose drift on the air-gap page

`landing/docs/recipes/air-gapped-deployment.html:112` says the trap "is installed before `tar`"; the markdown was rewritten to "immediately after that `cd` — before the digest check, not just before `tar`", and explains that the digest-mismatch exit is the ordering's reason. The code blocks are correct in both; only the prose drifted.

## What the sweep found clean

Recording this so the next sweep can be narrower: 11 recipe md/html pairs are code-block-equivalent and near-identical in flags, numbers and config keys; 39 of 40 `blob/main` links resolve; all 56 CHANGELOG and 31 ROADMAP issue references exist with correct state; and every quoted path, line number and error string checked resolves to what the text claims.

## Related

The version-stamp half of this class is #489 (`landing-constants-guard.sh` fails and runs in no workflow). #491 is the same shape on the public-sector page. Three md/HTML divergences have now been found by hand while reviewing something else, which is not a strategy — a divergence check belongs in the same CI job #489 asks for.

xerj-org:
Closes #492.

The site/doc-correctness grab-bag from the #460 divergent-copy sweep. Every change is verified against the source-of-truth it drifted from; **no version stamps touched** (landing-constants-guard green).

## The headline: the site denied a shipped capability
`landing/docs/recipes/semantic-search-rag.html` told readers the only route to neural-quality semantics was to "wire an external embeddings endpoint" — **zero** mention of `--embed-mode`. So the page a prospective user reads said XERJ can't do offline semantic search without third-party infra, which is false. It now documents all three backends the `.md` source-of-truth (`docs/recipes/semantic-search-rag.md:141-193`) describes:
- **`lexical`** (default) — built-in feature-hash, offline, zero download
- **`neural`** — built-in BERT `all-MiniLM-L6-v2` via `candle`, in-process, ~90 MB auto-download once, air-gap friendly via `embedding.local_model_dir`
- **`proxy`** — bring-your-own OpenAI-compatible endpoint

## Everything else in #492
| item | fix |
|---|---|
| README badges link to `vectors.html`, which never mentioned embedders | added an **Embedding backends** section to `landing/docs/vectors.html` so the badge target documents the claim |
| `limits.max_process_memory_mb` documented nowhere | added to `landing/docs/config.html` (table + prose) and `engine/xerj.default.toml` — **commented** so the shipped AUTO default is unchanged |
| hub `case-studies.html` stale `15/15 vs 1/15` | → `21/21 vs 1/21` (matches the detail page, `SUMMARY.md`, and the case study; the hub was the lone stale outlier) |
| phantom `#step-6b` nav link (no such section) | removed from `landing/demo/index.html` |
| dead `engine/demo-data/build_corpus.py` link (file exists nowhere in repo) | unlinked (kept the name as text) |
| GET XERJ button `/get/index.html` (`/get` is a file → 200 homepage fallback) | → `/docs/install.html` |
| air-gap prose drift (`air-gapped-deployment.html:112`) | **already fixed on main** — no change needed |

## One opinionated call — please veto if wrong
`resources/index.html` had a **"DOWNLOAD THE FULL KIT · ZIP"** section pointing at `/resources/xerj-sales-kit.zip`. That asset is **deliberately gitignored** (`.gitignore:102`, added by RELEASE_HYGIENE_AUDIT) as *"internal go-to-market material — not part of the OSS engine"*, and no build step generates it — so live it silently served the homepage as a `.zip`. I **removed the bundle section** rather than ship the excluded asset or wire a build for it. The 12 individual briefs above it stay downloadable. If you'd rather build+host the zip from a separate repo/CDN, revert this hunk and repoint instead.

## Verification
- `landing-constants-guard.sh` green (version stamps, conformance figures, llms release all unchanged)
- HTML edits use the site's existing `.cfg`/`.cell` grid classes; tag balance checked
- Config-key semantics verified against `governor.rs` / `config.rs` (`AUTO_PROCESS_MEMORY_MB = u64::MAX` sentinel → machine-tiered 8/16/32 GiB; `0` = whole machine)


xerj-org:
### Merge verification — proportionate direct review (site/docs/config; no engine-code paths)

**Verdict: 0 BLOCKING → merging.** Like #564, this touches no Rust logic, so the gate is a careful direct read + CI, not 3 compiling worktrees.

- **config.html / xerj.default.toml** — machine-verified: the `the_docs_site_config_page_agrees_with_the_real_defaults` guard (which string-compares the table to `Config::default()`) is **green** after dropping the sentinel-default `max_process_memory_mb` row from the literal-default table (kept in the [limits] prose + TOML, both of which the guard leaves to comments).
- **RAG recipe / vectors.html** — the three-backend content (incl. the shipped `--embed-mode neural`) is a faithful port of the source-of-truth `docs/recipes/semantic-search-rag.md`.
- **case-studies.html tally** — `21/21 vs 1/21` matches the detail page (`reference-coding.html`) and `SUMMARY.md`; the hub was the lone stale `15/15`.
- **Dead links** — `/docs/install.html` exists; `#step-6b` phantom nav removed; `build_corpus.py` unlinked (exists nowhere).

Full-matrix CI green + CLEAN. The one opinionated call — removing the resources BUNDLE section (its zip is deliberately gitignored internal GTM material) — is flagged in the PR body for maintainer veto. Merging.

Issue #354, opened 2026-08-13
match/multi_match on a keyword field is mapping-aware only after flush (memtable scan token-ORs the query)


xerj-org:
`multi_match` / `match` on a **keyword** field is mapping-aware in a flushed
segment but NOT in the memtable, so a multi-token query changes its answer at
`_flush`. No arrays are involved — this is independent of #332.

### Repro (integration test, `engine/crates/xerj-engine`)

One index, one document, a `keyword` field with a **scalar** value:

```rust
schema.fields.push(FieldConfig::new("tags", FieldType::Keyword));
idx.index_document(Some("2".into()), json!({"tags": "red"})).await?;
```

```
BEFORE flush | multi_match 'red blue' (2-token query, scalar kw doc) -> 1 hit(s)
BEFORE flush | match       'red blue' (2-token query, scalar kw doc) -> 1 hit(s)
BEFORE flush | term        'red'      (control)                      -> 1 hit(s)
AFTER  flush | multi_match 'red blue' (2-token query, scalar kw doc) -> 0 hit(s)
AFTER  flush | match       'red blue' (2-token query, scalar kw doc) -> 0 hit(s)
AFTER  flush | term        'red'      (control)                      -> 1 hit(s)
```

### Expected (Elasticsearch)

A `keyword` field is indexed with the keyword analyzer, so the query text is
also taken whole: `"red blue"` is one term, it does not equal `"red"`, and the
document does **not** match — in both phases. The flushed answer (0) is the
correct one; the memtable answer (1) is wrong.

### Cause

`doc_matches_query` / `score_query_against_doc` (`xerj-engine/src/index.rs`)
evaluate the `MultiMatch` and `Match` arms by whitespace-tokenising the query
and OR-ing the tokens against the stored `_source` text. They never consult the
mapping, so a `keyword` field is treated exactly like an analyzed `text` field.
The segment side does consult it, via the per-field analyzer configured in
`build_fts_field_configs`.

### Why it is filed separately

Found while fixing #332 (multi-valued keyword fields flattened to one FTS
token). The recovered #332 branch carried an assertion that
`multi_match {"query": "red blue", "fields": ["tags"]}` returns nothing in both
phases; it fails PRE-flush for the reason above, on documents with no array
anywhere, so it is not something the #332 fix can or should close. The #332 PR
asserts the flushed side of that case and points here for the memtable side.

Fixing this means making the stored-source scan mapping-aware for
[...]

Issue #353, opened 2026-08-13
painless_script_limits: call_depth_limit_survives_the_multi_thread_block_in_place_path overflows the stack and aborts the process on main


xerj-org:
## Repro

Clean checkout of `main` at **d8be09cf** (a worktree, nothing else applied):

```sh
cd engine
cargo test -p xerj-engine --test painless_script_limits
```

## Observed

```
running 9 tests
test painless_values_can_cross_a_thread_boundary ... ok
test nested_block_closure_recursion_does_not_abort_on_a_2mib_stack ... ok

thread 'call_depth_limit_survives_the_multi_thread_block_in_place_path' (129462) has overflowed its stack
fatal runtime error: stack overflow, aborting
error: test failed, to rerun pass `-p xerj-engine --test painless_script_limits`

Caused by:
  process didn't exit successfully: (signal: 6, SIGABRT: process abort signal)
```

The whole test binary aborts, so the remaining cases in the file never run — 2 of 9 report before
the abort, the other 7 produce no result at all.

## Why this matters beyond the one test

The test's name is the claim: `call_depth_limit_survives_the_multi_thread_block_in_place_path`. It
exists to assert that the Painless call-depth limit still bounds recursion when execution goes
through the multi-thread `block_in_place` path — i.e. that a hostile or runaway script hits the
depth limit instead of the thread's stack guard. The overflow is that assertion failing in the most
direct way available: on that path the limit does **not** hold, and the process dies rather than
returning a script error. Since `panic = "abort"` is in play, that is a whole-process kill, not a
recoverable per-request failure.

So the interesting question is not "why is the test flaky" but whether the depth limit is applied at
all on the `block_in_place` path, or whether the limit's budget is calibrated against the main
thread's stack while `block_in_place` runs on a smaller runtime-worker stack. Sibling test
`nested_block_closure_recursion_does_not_abort_on_a_2mib_stack` passing suggests the 2 MiB figure is
the calibration point and the failing path is not getting it.

## Scope / provenance

- Pre-existing and unrelated to the tree-sitter break — it reproduces once the build is unblocked,
  which is exactly how it was found.
- First reported by [@buger](https://github.com/buger) in the reviewer notes on #351; that PR was
  closed as already-landed (#350 shipped the identical fix), so the observation is filed here rather
  than lost with it. Independently reproduced on d8be09cf as shown above.
- Run under the dev/test profile. Not yet checked whether the release profile or CI's stack settings
[...]

PR #571, opened 2026-08-21
fix(engine): make keyword match/multi_match mapping-aware in the memtable (#354)

xerj-org:
Advances #354 — makes top-level `match`/`multi_match` on `keyword` fields mapping-aware. Two shapes remain (see **Residuals**), so #354 stays open.

## The defect
`match`/`multi_match` on a `keyword` field was mapping-aware in a **flushed segment** (per-field keyword analyzer) but NOT in the **memtable**: `doc_matches_query`/`score_query_against_doc` whitespace-tokenise the query and OR the tokens against `_source`, never consulting the mapping. So a scalar `{"tags":"red"}` keyword doc:

```
BEFORE flush | multi_match 'red blue' -> 1 hit  (WRONG: tokenised "red"/"blue" OR-ed against "red")
AFTER  flush | multi_match 'red blue' -> 0 hits (correct: keyword "red" != "red blue")
```

The hit set **flips at `_flush`** — the same write, replayed into a flushed index, answers differently.

## The fix
`doc_matches_query` is schema-less with **~78 call sites** (many recursive) and is a sync fn while the schema sits behind an async `RwLock`, so threading field types through every caller is a large, invasive change. Instead, rewrite the query **once at the search entry** — the `resolved_query` block where the schema is already read for `rewrite_query_aliases` (`index.rs`): `rewrite_keyword_full_text_to_term` converts keyword-targeted `Match`/`MultiMatch` clauses to whole-value `Term`s (a mixed `multi_match` splits: `Term`s for the keyword fields + a `multi_match` over the remaining text fields; analyzed-`text` fields untouched; applied after `rewrite_query_aliases` so names are canonical). Both the memtable and segment paths then take the value whole and agree.

A `multi_match` field spec may carry a `^boost` suffix (`"tags^2"`); the rewrite strips it with `parse_field_boost` before consulting the mapping (otherwise a boosted keyword field silently escapes the rewrite and keeps flipping) and folds that field boost into the rewritten `Term`.

## Residuals (NOT fixed here — #354 stays open)
Adversarial verification (3 independent skeptics, **0 BLOCKING**) confirmed the fix and surfaced two non-blocking gaps, both leaving the #354 flush-flip live for the shapes below; neither is a regression:

1. **Wrapper-nested keyword clauses** — a keyword `match`/`multi_match` inside `constant_score`/`function_score`/`boosting`/`dis_max`/`nested`/`hybrid` is not rewritten (only `Bool` is recursed), so `constant_score{filter:{match:{tags:"red blue"}}}` still flips at flush. This is the **same traversal coverage as the sibling `rewrite_query_aliases`** (which likewise recurses only Term/Terms/Range/Exists/Match/Bool); extending recursion to these wrappers is follow-up work. Harm differs from an unresolved alias (which errors loudly) — an unrewritten keyword match returns silently wrong, flush-dependent results — so it is worth closing next.
2. **Scoring shift for `multi_match` spanning a keyword field** — the split rewrites to `Bool.should` (min 1), whose clause scores **SUM**, whereas the ES `best_fields` default is a dis_max (**MAX**). A doc matching both the keyword half and a text half scores differently than before. **Hit sets are unchanged**; only scores shift, on a path that previously produced wrong hit sets anyway.

- **Fail-before proven** (revert only the `index.rs` hunk): the repro returns `{"2"}` PRE-flush instead of `{}`; with the fix, `{}` in both phases. The boosted-spec case (`fields:["tags^2"]`) likewise returned `{"2"}` PRE-flush on the prior commit and now `{}` in both phases.
- Full `xerj-engine` suite **green** and `xerj-api` suite **green** (`RUST_MIN_STACK=8388608` avoids the unrelated `painless_script_limits` default-stack flake, #353); `clippy --all-targets -D warnings` clean.
- Flush-parity tests: the scalar-keyword repro (plain + `^boost` spec) + a mixed keyword/text `multi_match` exercising the split.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -14005,7 +14005,10 @@ impl Index {
         // from `_source` (#204 — accepted means honoured).
         let resolved_query = {
             let schema = self.schema.read().await;
-            let resolved = rewrite_query_aliases(&request.query, &schema.schema);
+            let resolved = rewrite_keyword_full_text_to_term(
+                &rewrite_query_aliases(&request.query, &schema.schema),
+                &schema.schema,
+            );
             if let Some(field) = unsearchable_query_field(&resolved, &schema.schema) {
                 return Err(EngineError::Common(xerj_common::XerjError::invalid_query(
                     format!(
@@ -32011,6 +32014,100 @@ mod lexically_typeless_lowering_tests {
             lower_lexically_typeless_clauses(&q, &std::collections::HashSet::new()),
             q
         );
+    }
+}
+
+/// #354: `match`/`multi_match` on a `keyword` field must take the query text
+/// WHOLE (keyword analyzer), not whitespace-tokenise + OR it. The flushed
+/// segment path already does this (per-field analyzer), but the memtable's
+/// schema-less `doc_matches_query`/`score_query_against_doc` tokenise, so the
+/// answer flipped at `_flush`. Rewrite keyword-targeted `Match`/`MultiMatch`
+/// clauses to whole-value `Term`s once here — where the schema is in scope — so
+/// BOTH paths agree. Analyzed-`text` fields are left untouched; a mixed
+/// `multi_match` splits into `Term`s for the keyword fields plus a `multi_match`
+/// over the remaining text fields. Same traversal coverage as
+/// `rewrite_query_aliases`, and applied after it so field names are canonical.
+fn rewrite_keyword_full_text_to_term(q: &QueryNode, schema: &Schema) -> QueryNode {
+    fn is_keyword(schema: &Schema, field: &str) -> bool {
+        schema
+            .fields
+            .iter()
+            .any(|f| f.name == field && matches!(f.field_type, FieldType::Keyword))
+    }
+    match q {
+        QueryNode::Match {
+            field,
+            query,
+            boost,
+            ..
+        } if is_keyword(schema, field) => QueryNode::Term {
+            field: field.clone(),
+            value: Value::String(query.clone()),
+            boost: *boost,
+        },
+        QueryNode::MultiMatch {
+            fields,
+            query,
+            boost,
+            ..
+        } if fields.iter().any(|f| is_keyword(schema, f)) => {
+            let (kw, text): (Vec<String>, Vec<String>) =
+                fields.iter().cloned().partition(|f| is_keyword(schema, f));
+            let mut should: Vec<QueryNode> = kw
+                .into_iter()
+                .map(|f| QueryNode::Term {
+                    field: f,
+                    value: Value::String(query.clone()),
+                    boost: *boost,
+                })
+                .collect();
+            if !text.is_empty() {
+                // Preserve match_type/operator/etc. by cloning and swapping only
+                // the field list.
+                let mut text_mm = q.clone();
+                if let QueryNode::MultiMatch { fields: mmf, .. } = &mut text_mm {
+                    *mmf = text;
+                }
+                should.push(text_mm);
+            }
+            if should.len() == 1 {
+                should.into_iter().next().unwrap()
+            } else {
+                QueryNode::Bool {
+                    must: Vec::new(),
+                    should,
+                    must_not: Vec::new(),
+                    filter: Vec::new(),
+                    minimum_should_match: Some(MinShouldMatch::Fixed(1)),
+                }
+            }
+        }
+        QueryNode::Bool {
+            must,
+            should,
+            must_not,
+            filter,
+            minimum_should_match,
+        } => QueryNode::Bool {
+            must: must
+                .iter()
+                .map(|c| rewrite_keyword_full_text_to_term(c, schema))
+                .collect(),
+            should: should
+                .iter()
+                .map(|c| rewrite_keyword_full_text_to_term(c, schema))
+                .collect(),
+            must_not: must_not
+                .iter()
+                .map(|c| rewrite_keyword_full_text_to_term(c, schema))
+                .collect(),
+            filter: filter
+                .iter()
+                .map(|c| rewrite_keyword_full_text_to_term(c, schema))
+                .collect(),
+            minimum_should_match: minimum_should_match.clone(),
+        },
+        other => other.clone(),
     }
 }
 
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -32050,20 +32050,41 @@ fn rewrite_keyword_full_text_to_term(q: &QueryNode, schema: &Schema) -> QueryNod
             query,
             boost,
             ..
-        } if fields.iter().any(|f| is_keyword(schema, f)) => {
-            let (kw, text): (Vec<String>, Vec<String>) =
-                fields.iter().cloned().partition(|f| is_keyword(schema, f));
+        } if fields
+            .iter()
+            .any(|f| is_keyword(schema, parse_field_boost(f).0)) =>
+        {
+            // A multi_match field spec may carry a `^boost` suffix ("tags^2");
+            // strip it before consulting the mapping (otherwise a boosted
+            // keyword field escapes the rewrite and keeps flipping at flush),
+            // and fold that field boost into the rewritten keyword `Term`.
+            let (kw, text): (Vec<String>, Vec<String>) = fields
+                .iter()
+                .cloned()
+                .partition(|f| is_keyword(schema, parse_field_boost(f).0));
             let mut should: Vec<QueryNode> = kw
                 .into_iter()
-                .map(|f| QueryNode::Term {
-                    field: f,
-                    value: Value::String(query.clone()),
-                    boost: *boost,
+                .map(|f| {
+                    let (name, fboost) = parse_field_boost(&f);
+                    // No `^boost` suffix → preserve the clause boost verbatim
+                    // (keeps the common unboosted case byte-identical); with a
+                    // suffix, combine it with the clause boost as ES does.
+                    let boost = if name.len() == f.len() {
+                        *boost
+                    } else {
+                        Some(fboost * (*boost).unwrap_or(1.0))
+                    };
+                    QueryNode::Term {
+                        field: name.to_string(),
+                        value: Value::String(query.clone()),
+                        boost,
+                    }
                 })
                 .collect();
             if !text.is_empty() {
-                // Preserve match_type/operator/etc. by cloning and swapping only
-                // the field list.
+                // Preserve match_type/operator/etc. (and each text field's own
+                // `^boost`, which the text matcher strips) by cloning and
+                // swapping only the field list.
                 let mut text_mm = q.clone();
                 if let QueryNode::MultiMatch { fields: mmf, .. } = &mut text_mm {
                     *mmf = text;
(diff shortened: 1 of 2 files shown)


xerj-org:

Workflow `wjnnwq19w` — 3 independent skeptics, each in its own worktree, each tasked to **refute** the change (rustc 1.98.0). **Result: 0 BLOCKING across all 3 lenses.** Findings below are all NON_BLOCKING/NIT and are reflected in the PR's **Residuals** section. The two `^boost` / boosted-multi_match findings were subsequently **fixed** in commit `59ff53ac` (see PR description). Posting verbatim per process; the merge decision is the maintainer's after reading these.

### Lens: EVIDENCE / FAIL-BEFORE — prove the two new tests are load-bearing.

**Verdict: NON_BLOCKING**

**Suite result:** engine test target keyword_match_is_mapping_aware, toolchain 1.98.0, RUST_MIN_STACK=8388608. PR HEAD (27c35117): 2 passed, 0 failed. Fail-before (index.rs reverted to merge-base 8831d85c, both tests kept): 1 passed, 1 failed — keyword_match_takes_the_query_whole_in_both_phases FAILED, mixed_multi_match_splits_keyword_and_text PASSED. After restoring index.rs: 2 passed, 0 failed. No painless/SIGABRT flake seen.

**Summary:** The primary test is genuinely load-bearing: with the fix reverted it fails PRE-flush with exactly left={"2"} right={}, matching the PR's stated fail-before evidence. But the second (mixed) test does NOT distinguish the fix — it passes unchanged on the reverted baseline, because doc "1" matches via the analyzed text half ("blue" token in body) whether or not the keyword half is whole-valued, so its hit set {"1"} is invariant to bug #354 and flush parity holds on main too. Only one of the "two flush-parity tests" is a real fail-before proof of #354; the mixed test is a regression guard for the fix's own split logic, not a demonstration of the bug. No correctness defect found in the fix itself under this lens.

**Findings:**
- **[NON_BLOCKING]** mixed_multi_match_splits_keyword_and_text does not distinguish the fix — it passes on origin/main (fix reverted), so it is not a fail-before proof of #354.

  _Evidence:_ With index.rs reverted to merge-base 8831d85c and both tests kept, the run reports: keyword_match_takes_the_query_whole_in_both_phases ... FAILED; mixed_multi_match_splits_keyword_and_text ... ok (test result: FAILED. 1 passed; 1 failed). Doc "1" (tags="red", body="blue sky") matches via the analyzed text half — query token "blue" hits body token "blue" — regardless of whether the keyword half is tokenised (buggy) or whole-value (fixed); the memtable's buggy keyword-as-text path would ALSO match "red", but the final hit set is {"1"} in both cases and flush parity holds on main. So the mixed test's assertion is invariant to bug #354. It is a legitimate guard against the fix's own mixed-split logic dropping the text field, but the PR framing ('two flush-parity tests added' for #354) overstates: only one test actually fails before the fix.

- **[NIT]** The mixed test would be strengthened to actually exercise the keyword half by adding a doc where the keyword field is the ONLY reason a doc would (wrongly) match, e.g. a doc with tags matching a query token but body sharing no token.

  _Evidence:_ As written, no assertion changes value between the buggy and fixed keyword path; a doc such as tags="red", body="quiet lake" would match {"red"} under the buggy tokenised keyword path but not under the whole-value fix, making the keyword half load-bearing in the mixed case.

---

### Lens: REGRESSION

**Verdict: NON_BLOCKING**

**Suite result:** INCOMPLETE — could not paste exact aggregates. All three runs (cargo test -p xerj-engine --no-fail-fast with RUST_MIN_STACK=8388608; cargo test -p xerj-api --no-fail-fast; cargo clippy --workspace --all-targets -- -D warnings, all under rustup 1.98.0 / rustc 1.98.0 88d9e12ae) were launched in the worktree engine/ dir but were still in the shared-CPU compile phase (output files empty, no "test result:" line yet) when I was forced to return. No pass/fail counts observed; no failures observed either. This aggregate must be re-run and confirmed green before merge — treat the suite as UNVERIFIED, not passing.

**Summary:** Static inspection of the diff finds no BLOCKING regression: the rewrite is strictly gated on FieldType::Keyword, so the three at-risk existing tests are untouched (multi_match_scoring and bool_keyword_term_scoring operate on Text fields / direct Term clauses; multi_valued_keyword_terms asserts exactly the fixed single-token whole-value membership, preserved because a whitespace-free query maps Match->Term identically). The one real behavior shift is non-blocking and untested: a multi_match spanning a keyword field is rewritten into a Bool `should` (scores SUM, minimum_should_match=1) whereas ES default best_fields is a dis_max (scores MAX) — so a doc matching both halves scores differently than before on the segment path. I could NOT paste live test aggregates: the engine/api/clippy runs were still compiling when I had to return, so the suite is UNVERIFIED and must be confirmed green before merge.

**Findings:**
- **[NON_BLOCKING]** multi_match spanning a keyword field changes scoring semantics from best_fields (dis_max/MAX) to Bool-should (SUM).

  _Evidence:_ rewrite_keyword_full_text_to_term (engine/crates/xerj-engine/src/index.rs, added ~line 32017) converts a MultiMatch touching any keyword field into QueryNode::Bool{ should:[Term(kw)... , multi_match(text)], minimum_should_match:Some(Fixed(1)) }. Per the QueryNode::Bool doc ('At least minimum_should_match clauses must match; scores summed') the should clauses SUM, but the default MultiMatchType is BestFields (ast.rs:44 '#[default] BestFields — Takes the best score across all fields'), which is a MAX/dis_max. A doc that matches both a keyword field (exact whole-value) and a text field for the same query string now gets summed scores instead of the max. This also applies to a multi_match over 2+ keyword fields. Not caught by any test — the new keyword_match_is_mapping_aware.rs and multi_valued_keyword_terms.rs assert only hit-id membership, never scores; multi_match_scoring.rs uses Schema::empty() (no keyword field) so never exercises this path.

- **[NON_BLOCKING]** The rewrite does not descend into compound wrappers, so keyword match/multi_match nested inside constant_score/function_score/boosting/hybrid/knn-filter is not made mapping-aware — but this is consistent with existing behavior, not a new regression.

  _Evidence:_ rewrite_keyword_full_text_to_term recurses only into QueryNode::Bool; all other variants hit the `other => other.clone()` arm. I confirmed rewrite_query_aliases (the sibling the PR comment claims equal coverage with) is equally shallow — its arms are only Term/Terms/Range/Exists/Match/Bool (awk over the fn body), so alias resolution also never descends into FunctionScore/Boosting/etc. The #354 divergence therefore persists for keyword full-text clauses buried inside those wrappers, but that gap pre-dates this PR and matches the established rewrite convention, so it is a fix-completeness note rather than a regression this PR introduces.

- **[NON_BLOCKING]** Test-suite aggregates were not obtained; green status is unverified.

  _Evidence:_ cargo test -p xerj-engine (RUST_MIN_STACK=8388608), cargo test -p xerj-api, and cargo clippy --workspace --all-targets -- -D warnings were all launched under rustup 1.98.0 in engine/ but their output files were still empty (compile phase, three cargo processes contending for CPU) when I was required to return. No 'test result:' line, no clippy warning/error line, and no painless #353 SIGABRT were observed either way. Re-run and confirm all three are green before merge; do not read this review's NON_BLOCKING verdict as a passing suite.

---

### Lens: CORRECTNESS

**Verdict: NON_BLOCKING**

**Suite result:** Built xerj-engine clean under `rustup run 1.98.0` (Finished, 0 errors). New test file `keyword_match_is_mapping_aware.rs`: 2/2 pass (RUST_MIN_STACK=8388608). Ran 3 adversarial probe tests: PROBE1 (boosted keyword field spec) and PROBE2 (keyword match in constant_score) FAIL as flush-flips; PROBE3 (best_fields scoring) PASSES — no scoring regression. Did not run the full engine suite.

**Summary:** The core rewrite is correct for what it covers: keyword Match→whole-value Term, text left untouched (strict FieldType::Keyword guard), mixed multi_match split via clone+swap preserving the text half's match_type/operator/slop, all-keyword→Bool.should(min 1) which equals the best_fields union, single→Term unwrap. Build is clean under 1.98.0 and both new flush-parity tests pass. I found NO regression and NO scoring change (I tried to prove best_fields degrades max→sum and it did not — the summed clause score is collapsed to the max, so A==max, refuting my own hypothesis). Two silent incompleteness gaps remain, both leaving the #354 flush-flip live, neither a regression, so both NON_BLOCKING: (1) a multi_match whose keyword field carries a `^boost` spec ("tags^2") escapes the rewrite because is_keyword does not strip the boost suffix that parse_field_boost otherwise handles; (2) a keyword match nested in constant_score/function_score/boosting/dis_max/nested/hybrid is not rewritten (only Bool is recursed). Gap (2) is genuine traversal parity with the sibling rewrite_query_aliases, but the harm profile differs (silent wrong hits vs. loud unsearchable-field errors), and both constant_score and function_score wrapping a match are very common ES shapes — so "closes #354" overstates the coverage.

**Findings:**
- **[NON_BLOCKING]** multi_match on a keyword field carrying a field-boost spec (e.g. "tags^2") is NOT rewritten and still flips its hit set at flush — the #354 bug persists for the exact node type the PR targets.

  _Evidence:_ is_keyword() at index.rs:~32040 does exact `f.name == field`; for the MultiMatch arm it tests each raw fields entry, but multi_match field specs embed `^boost` (parse_field_boost at index.rs:36553 strips it everywhere else). is_keyword("tags^2") is false, so `fields.iter().any(is_keyword)` is false and the whole MultiMatch falls to `other => clone`, unrewritten. Verified with a probe (tags mapped Keyword, doc {tags:"red"}): `multi_match {query:"red blue", fields:["tags^2"]}` returns PRE-flush={"2"} POST-flush={} — a silent flush-dependent flip. Unboosted `["tags"]` is correctly fixed. Field boosts are the primary reason to use multi_match, so this is a reachable hole the test suite does not cover. Fix: strip the boost (parse_field_boost) inside is_keyword and in the kw/text partition.

- **[NON_BLOCKING]** A keyword match/multi_match nested inside constant_score, function_score, boosting, dis_max, nested, or hybrid is not rewritten (only Bool is recursed), so it still flips at flush — the fix is materially incomplete for common wrapper queries.

  _Evidence:_ rewrite_keyword_full_text_to_term recurses only into Bool; all other composites hit `other => other.clone()` (index.rs:~32112). Verified: `constant_score {filter:{match:{tags:"red blue"}}}` on the same keyword doc returns PRE-flush={"2"} POST-flush={} — the #354 flip survives. This IS the same traversal coverage as the sibling rewrite_query_aliases (which also recurses only Term/Terms/Range/Exists/Match/Bool and clones the rest — it does not even have a MultiMatch arm), so it is pre-existing parity, not a regression — hence NON_BLOCKING. But note the harm is asymmetric: an unresolved alias surfaces loudly via unsearchable_query_field, whereas an unrewritten keyword match returns silently wrong, flush-dependent results. constant_score/function_score wrapping a match are extremely common ES shapes, so the PR's 'closes #354' framing overstates real coverage; the recursion should be extended to these wrappers (or the limitation documented on the issue).

- **[NIT]** The mixed/all-keyword multi_match split does not preserve cross_fields token-pooling for the keyword field (semantics shift toward, not away from, ES) — worth a note, not a defect.

  _Evidence:_ For match_type cross_fields the memtable pools tokens across ALL listed fields including keyword (doc_matches_query cross arm, index.rs:34533-34555). The rewrite pulls the keyword field out into a whole-value Term and runs cross_fields over only the remaining text fields, so a doc supplying a needed token via the keyword field no longer matches. This actually moves closer to ES (which regroups cross_fields by analyzer, keeping keyword in its own group), and it is flush-consistent because the rewrite runs before both paths — so it is arguably an improvement, but it is an unremarked behavior change for cross_fields/most_fields mixed queries and is untested.

---

Issue #572, opened 2026-08-21
multi_match spanning a keyword field scores SUM instead of best_fields MAX (follow-up to #354)


xerj-org:
### Context
The #354 fix (PR #571) makes keyword `match`/`multi_match` mapping-aware by rewriting a keyword-spanning `multi_match` into a `Bool.should` (minimum_should_match=1): a whole-value `Term` for each keyword field + a `multi_match` over the remaining text fields.

### The gap
`Bool.should` clause scores **SUM**, whereas the ES `multi_match` default (`best_fields`) is a **dis_max** — it takes the **MAX** across fields. So a document that matches both the keyword half and a text half now scores differently (summed) than a true `best_fields` query (max).

- **Hit sets are unchanged** — this is purely a `_score` difference, and only on the previously-broken keyword-spanning shape.
- Surfaced by the adversarial verification of #571 (regression + correctness lenses), rated NON_BLOCKING.

### Fix direction
When `match_type` is `best_fields` (the default) or `most_fields`, the split should combine the keyword `Term`(s) and the text `multi_match` under a **dis_max** (MAX + tie_breaker) rather than `Bool.should` (SUM), so the rewritten query preserves ES scoring semantics. `cross_fields`/`phrase` variants need their own consideration (see #571 NIT on cross_fields token-pooling).

### Acceptance
A `multi_match` over `[keyword_field, text_field]` where a doc matches both halves scores as `max(keyword_score, text_score)` (± tie_breaker), not the sum. Add a score-asserting test (existing #354 tests assert hit-ids only).

Follow-up to #354 / PR #571.

PR #573, opened 2026-08-21
fix(engine): keyword match/multi_match mapping-aware inside wrapper queries (#354 residual)

xerj-org:
Advances #354 — closes the last residual from PR #571.

## The gap (left open by #571)
#571 made **top-level** keyword `match`/`multi_match` mapping-aware by rewriting them to whole-value `Term`s in `rewrite_keyword_full_text_to_term`, but that fn recursed **only into `Bool`**. Every other sub-query-carrying node fell through `other => other.clone()`, so a keyword clause nested inside a wrapper was left analyzed in the memtable and kept **flipping its hit set at `_flush`** — the original #354 data-shape bug, still live for very common ES shapes:

```
constant_score{ filter: { match: { tags: "red blue" } } }   // tags is `keyword`, doc {"tags":"red"}
BEFORE flush -> {doc}   (WRONG: "red blue" tokenised + OR-ed, "red" matches)
AFTER  flush -> {}       (correct: keyword "red" != "red blue")
```

## The fix
Recurse into the sub-query-carrying wrappers, mirroring the existing `Bool` arm: `constant_score` (`Constant`), `Boosted`, `function_score`, `boosting` (positive + negative), `dis_max`, `nested`, `hybrid` (each weighted sub-query), `pinned` (organic), `named`. Leaf and positional (span) nodes have no full-text sub-query and still fall through unchanged. This brings the rewrite's coverage in line with — and slightly beyond — the sibling `rewrite_query_aliases`.

- **Fail-before proven** (revert only the `index.rs` arms via `git stash`): the new test fails PRE-flush — `constant_score > match "red blue"` returned `{"2"}` instead of `{}`; with the fix, `{}` in both phases.
- New flush-parity test `keyword_match_mapping_aware_in_wrappers.rs` covers `constant_score` / `function_score` / `boosting` / `dis_max` (incl. a boosted `multi_match` inside `dis_max`) + a whole-value positive control — all mapping-aware and flush-stable.
- Full `xerj-engine` suite **green** (49 test binaries, 0 failed; `RUST_MIN_STACK=8388608` for the unrelated #353 flake); `clippy --all-targets -D warnings` clean; `fmt --check` clean.

Scoring nuance for `multi_match` spanning keyword remains tracked separately in #572. With this, all keyword `match`/`multi_match` flush-flips in #354 are closed.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -23,7 +23,7 @@ use xerj_fts::search::{
 use xerj_query::ast::{
     BoostMode, FieldValueFactor, Fuzziness, HighlightRequest, MinShouldMatch, Modifier, QueryNode,
     RandomScore, RescoreQuery, SavingsMode, ScoreFunction, ScoreMode, SearchRequest, SourceFilter,
-    TrackTotalHits,
+    TrackTotalHits, WeightedQuery,
 };
 use xerj_query::executor::{
     strip_internal_passage_metadata, Hit, PassageMatch, PayloadSavings, SavingsMethod,
@@ -32127,6 +32127,78 @@ fn rewrite_keyword_full_text_to_term(q: &QueryNode, schema: &Schema) -> QueryNod
                 .map(|c| rewrite_keyword_full_text_to_term(c, schema))
                 .collect(),
             minimum_should_match: minimum_should_match.clone(),
+        },
+        // Wrapper/compound nodes carry sub-queries that may target keyword
+        // fields; recurse into them so a keyword `match`/`multi_match` nested
+        // inside them is made mapping-aware too (otherwise it keeps flipping at
+        // flush). Leaf and positional (span) nodes have no full-text sub-query
+        // to rewrite and fall through unchanged.
+        QueryNode::Constant { score, query } => QueryNode::Constant {
+            score: *score,
+            query: Box::new(rewrite_keyword_full_text_to_term(query, schema)),
+        },
+        QueryNode::Boosted { boost, query } => QueryNode::Boosted {
+            boost: *boost,
+            query: Box::new(rewrite_keyword_full_text_to_term(query, schema)),
+        },
+        QueryNode::FunctionScore {
+            query,
+            functions,
+            score_mode,
+            boost_mode,
+            max_boost,
+        } => QueryNode::FunctionScore {
+            query: Box::new(rewrite_keyword_full_text_to_term(query, schema)),
+            functions: functions.clone(),
+            score_mode: *score_mode,
+            boost_mode: *boost_mode,
+            max_boost: *max_boost,
+        },
+        QueryNode::Boosting {
+            positive,
+            negative,
+            negative_boost,
+        } => QueryNode::Boosting {
+            positive: Box::new(rewrite_keyword_full_text_to_term(positive, schema)),
+            negative: Box::new(rewrite_keyword_full_text_to_term(negative, schema)),
+            negative_boost: *negative_boost,
+        },
+        QueryNode::DisMax {
+            queries,
+            tie_breaker,
+        } => QueryNode::DisMax {
+            queries: queries
+                .iter()
+                .map(|c| rewrite_keyword_full_text_to_term(c, schema))
+                .collect(),
+            tie_breaker: *tie_breaker,
+        },
+        QueryNode::Nested {
+            path,
+            query,
+            score_mode,
+        } => QueryNode::Nested {
+            path: path.clone(),
+            query: Box::new(rewrite_keyword_full_text_to_term(query, schema)),
+            score_mode: score_mode.clone(),
+        },
+        QueryNode::Hybrid { queries, fusion } => QueryNode::Hybrid {
+            queries: queries
+                .iter()
+                .map(|wq| WeightedQuery {
+                    query: rewrite_keyword_full_text_to_term(&wq.query, schema),
+                    weight: wq.weight,
+                })
+                .collect(),
+            fusion: fusion.clone(),
+        },
+        QueryNode::Pinned { ids, organic } => QueryNode::Pinned {
+            ids: ids.clone(),
+            organic: Box::new(rewrite_keyword_full_text_to_term(organic, schema)),
+        },
+        QueryNode::Named { name, query } => QueryNode::Named {
+            name: name.clone(),
+            query: Box::new(rewrite_keyword_full_text_to_term(query, schema)),
         },
         other => other.clone(),
     }
--- a/engine/crates/xerj-engine/tests/keyword_match_mapping_aware_in_wrappers.rs
+++ b/engine/crates/xerj-engine/tests/keyword_match_mapping_aware_in_wrappers.rs
@@ -1,0 +1,130 @@
+//! Issue #354 (follow-up): a keyword `match`/`multi_match` nested inside a
+//! wrapper/compound query — `constant_score`, `function_score`, `boosting`,
+//! `dis_max`, `nested`, `hybrid`, `pinned`, `named` — must be mapping-aware in
+//! BOTH the memtable and a flushed segment, exactly like a top-level clause.
+//!
+//! PR #571 rewrote top-level keyword `Match`/`MultiMatch` to whole-value `Term`s
+//! but recursed only into `Bool`, so a keyword match wrapped in e.g.
+//! `constant_score` fell through unchanged and kept flipping its hit set at
+//! `_flush` (the #354 defect, live for wrappers). This extends the rewrite's
+//! recursion to the sub-query-carrying wrappers.
+//!
+//! One doc, `tags` mapped `keyword`, scalar value `"red"`. A wrapped
+//! `match {tags: "red blue"}` must take the value WHOLE — `"red blue" != "red"`
+//! — so the doc matches in NEITHER phase.
+//!
+//! Elasticsearch is referenced for wire semantics only; no ES code is here.
+
+use std::collections::BTreeSet;
+
+use serde_json::{json, Value};
+use tempfile::TempDir;
+use xerj_common::config::Config;
+use xerj_common::types::{FieldConfig, FieldType, Schema};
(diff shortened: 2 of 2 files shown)


xerj-org:

Workflow `wruei921u` — 3 independent skeptics, each in its own worktree under rustc 1.98.0, each tasked to **refute** the change. **Result: 0 BLOCKING across all 3 lenses.** REGRESSION verified the full `xerj-engine` suite green (1029 passed, 0 failed, 23 ignored) + clippy clean + fail-before confirmed. Findings are NON_BLOCKING/NIT and tracked as follow-ups (#574 filed for the filter-node gap). Posting verbatim per process; merge decision is the maintainer's.

### Lens: CORRECTNESS

**Verdict: NON_BLOCKING**

**Suite result:** Built with rustup 1.98.0, RUST_MIN_STACK=8388608. New test tests/keyword_match_mapping_aware_in_wrappers.rs: 1 passed. Regression checks keyword_match_is_mapping_aware (2 passed), bool_keyword_term_scoring (5 passed). All green.

**Summary:** Tried to refute the wrapper-recursion change on correctness and could not find a blocking defect. Every one of the 9 new arms (Constant, Boosted, FunctionScore, Boosting, DisMax, Nested, Hybrid, Pinned, Named) reconstructs its QueryNode variant with all fields preserved verbatim — verified field-by-field against the enum def in xerj-query/src/ast.rs (score, boost, functions/score_mode/boost_mode/max_boost, negative_boost, tie_breaker, path/score_mode, weights/fusion, ids, name all carried; only the sub-query children are recursed). The recursion is a structural identity for non-keyword sub-queries (the Match/MultiMatch leaf arms fire only on keyword fields; all else clones), so it is a no-op when no keyword field is involved. On missed shapes: Knn.filter, SemanticSearch.filter and FunctionScore.functions[].filter carry sub-queries yet fall through `other => other.clone()`, but all three are evaluated with doc_matches_query against `_source` (index.rs:10970 / 11088 / 41303), which is flush-invariant — so not rewriting them leaves no live pre/post-flush hit-set flip. The remaining uncovered sub-query carriers are span nodes (span clauses only, no full-text match). Provided test plus two related suites pass.

**Findings:**
- **[NIT]** Knn.filter / SemanticSearch.filter / FunctionScore.functions[].filter carry sub-queries but are not recursed into (fall through other => other.clone())

  _Evidence:_ These three variants hold QueryNode sub-queries a keyword match could sit in, yet are uncovered by the rewrite. This is NOT a live flush-flip gap: all three are matched via doc_matches_query against _source (engine/crates/xerj-engine/src/index.rs:10970, 11088, 41303), which is identical pre- and post-flush, so the #354 memtable-vs-segment inconsistency cannot arise there. Harmless today, but if any of these filters is ever rerouted through the index postings path it would silently reintroduce the flip. The arm comment ('Leaf and positional (span) nodes have no full-text sub-query to rewrite') slightly understates this — these wrappers do carry sub-queries; they just happen to be _source-evaluated.

- **[NIT]** Request-level rescore query is not passed through rewrite_keyword_full_text_to_term

  _Evidence:_ rewrite_keyword_full_text_to_term is called only on request.query (index.rs:14008). SearchRequest.rescore.query.rescore_query is a QueryNode that runs against the index to rescore top-N, so a keyword match there would flush-flip — but rescore is a request-level field, not a QueryNode wrapper reachable by this function, so it is outside this PR's stated scope and is a pre-existing condition, not a regression introduced here.

---

### Lens: EVIDENCE / FAIL-BEFORE

**Verdict: NON_BLOCKING**

**Suite result:** keyword_match_mapping_aware_in_wrappers: PASS with fix (1 passed). FAIL-BEFORE confirmed: with only crates/xerj-engine/src/index.rs reverted to parent (2d9f35ee~1) and the test kept, the test FAILS pre-flush (constant_score match "red blue" returns {"2"} vs expected {}). A non-short-circuiting per-case probe with the source reverted proved all 5 shipped cases individually flip pre-flush (each returns {"2"} instead of {}): constant_score>match, function_score>match, boosting>positive match, dis_max>match, dis_max>boosted multi_match. All 5 are load-bearing. Toolchain 1.98.0, RUST_MIN_STACK=8388608.

**Summary:** The change (recurse the keyword match/multi_match -> whole-value Term rewrite into wrapper nodes) is correct, and its new test is genuinely load-bearing: every one of the 5 shipped cases fails pre-flush without the source arms and passes with them (verified by a per-case probe, not just the short-circuiting assert). No BLOCKING defect found. Two non-blocking observations: (1) the source adds 9 wrapper arms but the test only exercises 4 distinct ones (Constant/FunctionScore/Boosting/DisMax); Boosted, Nested, Hybrid, Pinned, Named ship untested despite the docstring/PR body claiming nested/hybrid/pinned/named coverage. (2) The new comment claims only leaf/span nodes fall through, but Knn{filter} and SemanticSearch{filter} also carry a full-text-capable sub-query and are not recursed; empirically knn.filter does not flush-flip (consistent both phases via the doc-by-doc pre-filter), so this is a pre-existing out-of-scope quirk, not a regression.

**Findings:**
- **[NON_BLOCKING]** Test docstring/PR body overclaim coverage: 5 of the 9 added wrapper arms (Boosted, Nested, Hybrid, Pinned, Named) are untested.

  _Evidence:_ engine/crates/xerj-engine/tests/keyword_match_mapping_aware_in_wrappers.rs lines 71-110 only build cases for constant_score, function_score, boosting, and dis_max (x2), yet lines 2-3 and the commit message assert the repro spans 'nested, hybrid, pinned, named' too. The source (index.rs lines 32131-32202) adds Constant, Boosted, FunctionScore, Boosting, DisMax, Nested, Hybrid, Pinned, Named. A FAIL-BEFORE run cannot demonstrate the Boosted/Nested/Hybrid/Pinned/Named arms are load-bearing because no case reaches them. My revert-source probe confirmed a pinned>organic case would flip pre-fix, so those arms are plausibly needed but remain unverified by the shipped test.

- **[NIT]** New comment claims only leaf/span nodes fall through, but Knn{filter} and SemanticSearch{filter} also carry a full-text sub-query and are not recursed.

  _Evidence:_ index.rs lines 32134-32135 say 'Leaf and positional (span) nodes have no full-text sub-query to rewrite and fall through unchanged.' QueryNode::Knn (ast.rs line 466) and QueryNode::SemanticSearch (line 484) both hold a filter: Option<Box<QueryNode>> and hit the `other => other.clone()` arm. A probe (fix present) of knn{filter: match{tags:'red blue'}} over a keyword 'red' doc returned ['2'] both PRE- and POST-flush — CONSISTENT, so it does not exhibit the #354 flush-flip this PR targets (the knn pre-filter is evaluated doc-by-doc identically in both phases). Hence not a regression and out of the PR's stated scope; only the comment is inaccurate.

---

### Lens: REGRESSION

**Verdict: NON_BLOCKING**

**Suite result:** VERIFIED GREEN. `RUST_MIN_STACK=8388608 rustup run 1.98.0 cargo test -p xerj-engine --no-fail-fast` → exit 0; aggregate across all 49 `test result:` lines = 1029 passed, 0 failed, 23 ignored. Every test binary reached a `test result:` line (48 `Running tests/…` binaries incl. `unittests src/lib.rs`, plus the doc-test = 49 lines); zero `FAILED`/`panicked`/`error[` in output. New test `tests/keyword_match_mapping_aware_in_wrappers.rs::keyword_match_in_wrappers_takes_the_query_whole_in_both_phases` = ok. `rustup run 1.98.0 cargo clippy -p xerj-engine --all-targets -- -D warnings` → `Finished`, 0 warnings/0 errors. Fail-before confirmed: the same test run on parent `2d9f35ee~1` (with the test file copied in) FAILS with `left: {"2"} right: {}` PRE-flush on `constant_score > match "red blue"`, so it is a real regression guard, not a tautology.

**Summary:** The change extends `rewrite_keyword_full_text_to_term` to recurse into nine sub-query-carrying wrappers (Constant/Boosted/FunctionScore/Boosting/DisMax/Nested/Hybrid/Pinned/Named), mirroring the existing Bool arm. I actively tried to refute it and could not find a merge-blocking defect. The recursion is structure-preserving: only keyword `Match`/`MultiMatch` leaves are transformed (gated by `is_keyword`); non-keyword `Match` and all other nodes fall through to `other => other.clone()`, and each reconstructed wrapper re-emits ALL of its enum fields (verified field-by-field against ast.rs), so any query without a keyword match/multi_match is byte-identical to the prior full clone — a text-only match inside constant_score is provably unchanged, and this is corroborated by the fully green suite. The keyword-match→Term rewrite in wrappers is the intended #354/#571 fix (pre-flush == post-flush), and the new test proves flush-parity across constant_score/function_score/boosting/dis_max/multi_match with a positive control. Two non-blocking incompleteness notes below.

**Findings:**
- **[NON_BLOCKING]** Two remaining sub-query carriers are still not recursed into: Knn.filter and SemanticSearch.filter (both `Option<Box<QueryNode>>`).

  _Evidence:_ ast.rs defines `Knn { … filter: Option<Box<QueryNode>>, … }` (line ~466) and `SemanticSearch { … filter: Option<Box<QueryNode>>, … }` (line ~484). The match in index.rs has no arm for either, so they hit `other => other.clone()` and their filter sub-query is never rewritten. A keyword `match`/`multi_match` placed in a knn/semantic filter (e.g. `bool{must:[knn{filter:{match:{tags:"red blue"}}}]}`) still flips its hit set at flush — the same #354 class the PR is closing for the other wrappers. This is a PRE-EXISTING gap (never handled), not a regression introduced here, and outside the PR's stated wrapper list; knn filters are a documented place users put keyword filters, so worth a follow-up. Not merge-blocking.

- **[NIT]** Test docstring advertises broader coverage than the test body exercises.

  _Evidence:_ The module doc (lines 1-10) lists `nested`, `hybrid`, `pinned`, `named` among the wrappers the test covers, but the test body (lines 71-110) only exercises constant_score, function_score, boosting, dis_max, and dis_max>multi_match plus one positive control. The nested/hybrid/pinned/named arms are covered by the recursion code but not directly asserted by this test. Harmless (the arms are trivially symmetric and the full suite is green), but the doc overstates the assertions.

---

PR #579, opened 2026-08-22
fix(autoindex): index each code declaration as its own retrievable document (#500)

xerj-org:
Advances #500 (core fix). Reference coding is the primary use case; a fair retrieval test found XERJ symbol search ingests **32–48× more bytes per answered question than grep** and loses recall (8/9 vs 12/12), because the smallest retrievable unit was the enclosing class/method and **constants/fields were not indexed as their own symbols** (`term name=DEFAULT_MAX_CONN` → count 0). The AST extractor already produced those symbols with line numbers; they were discarded for retrieval.

## The fix
`parse_symbols` now also captures each declaration's start-line text (`code`, single-line, char-safe capped at 400 B — language-agnostic because the name node's capture depth varies per grammar). `emit_code_doc` keeps the file-level document (`body`/`defs` for full-text + cross-file) **and** emits one record per declaration `{title=name, path, name, kind, line, code}` under a unique, stable locator `code:<line>:<name>` — so `doc_id(dataset, file_key, locator)` gives each declaration its own document instead of colliding with the file document. A constant lookup now returns the ~40–80 B declaration line with an exact `name`.

- **Fail-before proven** (disable the per-declaration emission): the struct/class are no longer their own documents (`symbol_docs = []`), so the reconcile assertion fails; with the fix they are indexed as their own docs.
- Full `xerj-autoindex` suite **green (694 passed, 0 failed)** on **both** dev and ci-test profiles; `clippy -D warnings` clean.
- The incremental-reconcile **HTTP** test verifies end-to-end that a struct + class are indexed as their own retrievable documents (not just foldable in a parent).

## Follow-ups (noted, not in this PR)
- `name.keyword` ranking boost so the defining symbol out-ranks incidental `code`/`body` token mentions (issue effect #3).
- Refsym byte-ratio re-measurement on the reference corpus.
- Promoting every declaration raises the document count per file (each symbol-doc is a single small line, not a body); a `kind` filter can bound it if index size warrants.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -511,8 +511,20 @@ fn emit_code_doc(
         .and_then(|n| n.to_str())
         .unwrap_or("source")
         .to_string();
+    // A single declaration can be captured under two kinds at the same line+name
+    // (e.g. `export const f = () => {}` → @function + @const). Those share the
+    // `code:<line>:<name>` locator → the same `doc_id` → ES stores ONE document.
+    // The emitted RECORD count MUST match, or the count-reconciliation barrier
+    // (sync_executor) sees "sealed N vs read-back N-1" and aborts the whole run.
+    // So dedup by locator here, BEFORE counting/emitting — do not rely on the
+    // downstream `_id` overwrite (which fixes the doc but not the count).
+    let mut emitted_locators = std::collections::HashSet::new();
     for (name, kind, line, code) in symbols {
         if code.is_empty() {
+            continue;
+        }
+        let locator = format!("code:{line}:{name}");
+        if !emitted_locators.insert(locator.clone()) {
             continue;
         }
         let mut sf = Map::new();
@@ -533,10 +545,9 @@ fn emit_code_doc(
             // file_key, locator)` gives each symbol its OWN document instead of
             // colliding with the file document (all of which used "code"). Keyed
             // by line+name: unique per declaration, stable across re-index unless
-            // the line moves, and it dedups a declaration double-captured under
-            // two kinds (e.g. `export const f = function(){}` → @function+@const)
-            // to a single doc at the same line.
-            locator: format!("code:{line}:{name}"),
+            // the line moves (deduped above so the record count matches the doc
+            // count).
+            locator,
             group: None,
             origin: FieldOrigin::Extractor,
         }) {
@@ -1198,6 +1209,50 @@ mod tests {
         );
         // Title comes from the logical name, not the blob ordinal.
         assert_eq!(records[0].fields["title"], "app.py");
+    }
+
+    /// #500 regression: a declaration captured under two kinds at the same
+    /// line+name (`export const X = () => …` → @function + @const, the dominant
+    /// modern-TS public-surface shape) must emit EXACTLY ONE record, because both
+    /// share the `code:<line>:<name>` locator → one document. Emitting two would
+    /// make the sealed record count exceed the read-back doc count, and the
+    /// count-reconciliation barrier aborts the entire index run.
+    #[test]
+    fn double_captured_declaration_emits_one_record_per_locator() {
+        let dir = tempfile::tempdir().unwrap();
+        let f = dir.path().join("m.ts");
+        std::fs::write(&f, "export const Button = () => 1;\n").unwrap();
+        let sn = crate::sniff::sniff_with_name(&f, Path::new("m.ts")).unwrap();
+        assert_eq!(sn.family, crate::sniff::Family::Code);
+        // The parser really does double-capture this shape (guards the premise).
+        assert!(
+            syms("typescript", "export const Button = () => 1;\n").len() >= 2,
+            "premise: Button is captured under >1 kind"
+        );
+        let mut records: Vec<RawRecord> = Vec::new();
+        let stats = extract(&f, &sn, &mut |record| {
+            records.push(record);
+            true
+        })
+        .unwrap();
+        let distinct: std::collections::HashSet<&str> =
+            records.iter().map(|r| r.locator.as_str()).collect();
+        assert_eq!(
+            records.len(),
+            distinct.len(),
+            "each emitted record needs a UNIQUE locator, else sealed count > read-back \
+             doc count aborts the run: {:?}",
+            records.iter().map(|r| &r.locator).collect::<Vec<_>>()
+        );
+        // The sealed count the reconcile barrier compares against must equal the
+        // number of records actually emitted.
+        assert_eq!(stats.records as usize, records.len());
+        assert!(
+            records
+                .iter()
+                .any(|r| r.fields.get("name").is_some_and(|n| n == "Button")),
+            "Button must still be its own symbol document"
+        );
     }
 
     /// #295 acceptance criterion: every registered grammar must instantiate
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -3,13 +3,18 @@
 //! Source files used to fall through to the plain-text extractor, so a repo was
 //! searchable only as prose. This parses each file with the matching tree-sitter
 //! grammar and captures its DEFINITIONS (functions, classes, methods, structs,
-//! traits, interfaces, modules, constants, …). Each file becomes one document
-//! carrying:
+//! traits, interfaces, modules, constants, …). Each file becomes a file-level
+//! document carrying:
 //! - `language`  the detected language
 //! - `symbols`   a structured array of {name, kind, line}
 //! - `defs`      "kind name" per symbol, newline-joined so BM25 matches a query
 //!   like `class User` or `def save` to the file that owns it
 //! - `body`      the full source text (still full-text searchable)
+//!
+//! PLUS one document PER DECLARATION (#500): `{name, kind, line, code}` under a
+//! unique `code:<line>:<name>` locator, so a constant/field/signature lookup
+//! retrieves the ~40–80 B declaration line as its own unit instead of only
+//! surviving folded inside the enclosing class/method body.
 //!
 //! The capture-name in each query IS the symbol kind (`@function`, `@class`, …),
 //! so adding a language is a grammar dep + one registry row. If a grammar fails
@@ -339,8 +344,17 @@ pub fn extract(path: &Path, sn: &crate::sniff::Sniffed, sink: Sink) -> Result<Ex
     Ok(stats)
 }
 
-/// (name, kind, 1-based line)
-type Symbol = (String, String, usize);
+/// (name, kind, 1-based line, declaration line text)
+///
+/// `code` is the single source line the declaration starts on — for a
+/// constant/field/`#define`/signature that IS the whole declaration (~40–80 B),
+/// so #500 can promote each declaration to its own retrievable document instead
+/// of only reaching it inside the ~2–3 KB enclosing class/method body. (A
+/// multi-line body's `code` is its signature line; the full body stays in the
+/// file document's `body`.) Language-agnostic on purpose: the capture depth of
+/// the name node varies per grammar, so a tree-walk to the declaration node is
+/// unreliable, whereas the start line is always exact.
+type Symbol = (String, String, usize, String);
 
 /// Evaluate a pattern's text predicates against one match. The core library
 /// exposes `#eq?`/`#any-of?` (and their `not-` forms) as *general* predicates
@@ -389,6 +403,9 @@ fn parse_symbols(def: &LangDef, text: &str) -> Option<Vec<Symbol>> {
     parser.set_language(&def.language).ok()?;
     let tree = parser.parse(text.as_bytes(), None)?;
     let names = def.query.capture_names();
+    // Row → source line, built once (not `lines().nth(row)` per symbol, which
+    // would be O(symbols · file)).
+    let lines: Vec<&str> = text.lines().collect();
     let mut out: Vec<Symbol> = Vec::new();
     let mut cursor = QueryCursor::new();
     let mut it = cursor.matches(&def.query, tree.root_node(), text.as_bytes());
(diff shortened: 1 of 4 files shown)


xerj-org:
## Adversarial 3-skeptic RE-verification — verdicts (verbatim)

Workflow `wobe6b37f` (rustc 1.98.0, worktrees), run AFTER the fix for the BLOCKING double-capture record-count bug the FIRST verification (`w4ciwm6l1`) caught. **Result: 0 BLOCKING across all 3 lenses.** The only reproduced abort is the PRE-EXISTING code-reindex `symbols`-absent bug (reproduced on parent `4ac32d17`, filed as #580 — out of scope). Posting verbatim per process.

### Lens 1

All test binaries accounted for (5 Running + 1 Doc-tests = 6 result lines per profile, matching). Both profiles green, clippy clean. I have everything needed.

---

### Lens 2



---

### Lens 3

VERDICT: NON_BLOCKING

---

### Lens 4



---

### Lens 5

Scope verified: PR #579 head a7feced3, branch fix/500-per-symbol-docs, worktree /tmp/verify579r-REGRESSION. The fix commit touches ONE file only — engine/crates/xerj-autoindex/src/extract/code.rs (+59/-4) — so the regression surface is narrow and fully inside this crate.

---

### Lens 6



---

### Lens 7

FULL-SUITE AGGREGATES (both profiles, --no-fail-fast, RUST_MIN_STACK=8388608, rustup 1.98.0):

---

### Lens 8

- [NON_BLOCKING] dev profile — DEV_EXIT=0. 6 result lines, every Running binary + Doc-tests accounted for: `681 passed; 0 failed` (lib) + 8 + 1 + 1 + 4 (integration: follow_symlinks_boundary, journal_lock_inheritance, scan_pool_width, sniff_printable_magic) + 0 (doctests) = 695 passed, 0 failed, 0 ignored. No binary lacks a "test result:" line; no panic/FAILED.

---

### Lens 9

- [NON_BLOCKING] ci-test profile — CI_EXIT=0. Same 6 result lines: `681 passed; 0 failed` + 8 + 1 + 1 + 4 + 0 = 695 passed, 0 failed. No missing result line.

---

### Lens 10

- [NON_BLOCKING] clippy -p xerj-autoindex --all-targets -D warnings — CLIPPY_EXIT=0, zero warnings/errors.

---

### Lens 11



---

### Lens 12

DEDUP FIX CONFIRMED (I read code.rs:461-558 and reasoned the paths, not just trusting the test name):

---

### Lens 13

- [NON_BLOCKING] Count now matches docs — dedup HashSet insert (line 527) `continue`s on a repeat BEFORE `stats.records += 1` (line 541), so the sealed record count equals the distinct-locator/doc count. Test `double_captured_declaration_emits_one_record_per_locator` passes in BOTH profiles and is non-vacuous: its own guard `syms("typescript", "export const Button = () => 1;").len() >= 2` passed, proving the parser really double-captures (@function + @const), so the dedup branch is actually exercised. It asserts records.len()==distinct.len() AND stats.records as usize == records.len().

---

### Lens 14

- [NON_BLOCKING] Per-declaration docs still emitted — `extracts_from_extensionless_blob_via_sniffed_logical_name` still asserts `(stats.records, stats.junk) == (2,0)` (file doc + `alpha_helper` symbol doc) and that a record with name=="alpha_helper" exists; the double_captured test also asserts a name=="Button" record survives. Declarations still become their own docs.

---

### Lens 15

- [NON_BLOCKING] File-level doc intact — line 490 `stats.records += 1` and the `locator:"code"` file record (lines 491-499) are untouched by this commit; `defs`/`symbols`/`symbol_count` construction (461-486) is unchanged.

---

### Lens 16

- [NON_BLOCKING] No behavior change for non-duplicate symbols — for distinct locators `insert` returns true → falls through to emit exactly as before; only an exact-repeat locator is skipped. `code.is_empty()` skip (523) still runs before the insert, so empty-code symbols neither count nor pollute the set. `locator.clone()` goes into the set (527) while the original `locator` String moves into the RawRecord (550) — no double-move, no borrow issue (clippy clean).

---

### Lens 17

- [NON_BLOCKING] HashSet allocation is conditional in practice — `std::collections::HashSet::new()` (521) performs NO heap allocation until the first `insert`; the first insert only happens when the `symbols` loop has a non-empty-code entry. So a file with zero symbols (or only empty-code symbols) allocates nothing, satisfying "allocates per file only when there are symbols."

---

### Lens 18



---

### Lens 19

OTHER-DEFECT HUNT (skeptic pass, nothing blocking found):

---

### Lens 20

- [NIT] Dedup key is (line,name) not (line,name,kind), so two genuinely distinct declarations sharing the same line AND name but different kind collapse to one doc, keeping the first capture's kind (arbitrary). This is NOT a regression: doc_id derives from the locator `code:<line>:<name>`, so such records already collided to a single document in the pre-fix feature commit (4a01596b) — the fix only stops the miscount that the collision caused. No new information loss is introduced.

---

### Lens 21

- [NIT] The chosen surviving kind for a double-captured symbol is order-dependent (first non-empty-code capture in `symbols` order wins, e.g. @function over @const for `export const f = () => {}`). Cosmetic; the doc is still retrievable by name. Not a correctness defect.

---

### Lens 22



---

### Lens 23

I could not construct any input that produces a reproduced wrong result on the fixed tree: the count-vs-doc invariant the barrier checks is restored, per-declaration and file-level emission both survive, and both full suites + clippy are green. No BLOCKING finding.

---

### Lens 24

All verification complete. Both named tests pass, both reversions are load-bearing, and the `code` field carries only the declaration line. No other defect reproduced.

---

### Lens 25



---

### Lens 26

VERDICT: NON_BLOCKING

---

### Lens 27



---

### Lens 28

[NON_BLOCKING] The dedup fix holds and the named tests pass — `RUST_MIN_STACK=8388608 rustup run 1.98.0 cargo test -p xerj-autoindex --lib -- double_captured_declaration_emits_one_record_per_locator code_files_index_with_ast_fields_and_the_run_reports_code_coverage` → `2 passed; 0 failed` (rustc 1.98.0, worktree at fix/500-per-symbol-docs head a7feced3).

---

### Lens 29



---

### Lens 30

[NON_BLOCKING] Dedup is load-bearing (defends the count-reconciliation invariant) — reverting only the skip by replacing `if !emitted_locators.insert(locator.clone()) { continue; }` with a plain `emitted_locators.insert(locator.clone());` (no-op) makes `double_captured_declaration_emits_one_record_per_locator` FAIL: panic at code.rs:1238 `assertion left == right failed … left: 3 right: 2`, with locators `["code", "code:1:Button", "code:1:Button"]` — i.e. 3 emitted records / stats.records=3 vs 2 distinct locators (doc_ids), exactly the sealed-N-vs-readback-N-1 mismatch that aborted the run. Restored via `git checkout`.

---

### Lens 31



---

### Lens 32

[NON_BLOCKING] Per-declaration emission is load-bearing — inserting an unconditional `continue;` at the top of the `for … in symbols` loop (empty emission) makes BOTH tests FAIL: `double_captured…` panics "Button must still be its own symbol document" (code.rs:1251), and `code_files_index…` panics "the AlphaConfig struct must be its own symbol document (#500): []" (only the 3 file-level records/`3 records live` remain, zero symbol docs). Restored via `git checkout`.

---

### Lens 33



---

### Lens 34

[NON_BLOCKING] A symbol doc carries the short declaration line, not the parent body — added a throwaway probe extracting a 4-line `pub struct AlphaConfig { … }`: the emitted symbol's `code` is exactly `"pub struct AlphaConfig {"` (code_len=24, contains no `retries`/`timeout` body). Source confirms it: `parse_symbols` sets `code = lines.get(row)` (single start line, indentation-trimmed, capped at 400 chars) — code.rs:429-433.

---

### Lens 35



---

### Lens 36

[NIT] The locator `code:<line>:<name>` omits `kind`, so two genuinely distinct symbols sharing one line+name but different kinds collapse to one document. This is not a regression: `doc_id(dataset, file_key, locator)` would collide and ES would store one doc regardless, so keeping one record here correctly matches the read-back count (the file-level `symbols` array still lists every capture). No wrong result produced — evidence: reasoning over the doc_id/locator scheme plus the passing `export const Button = () => 1;` case where @function+@const legitimately share `code:1:Button`.

---

### Lens 37



---

### Lens 38

[NIT] No collision between the file-level document (locator `"code"`) and any per-declaration doc (`format!("code:{line}:{name}")` always contains colons); `emit_code_doc` runs once per file (extract() code.rs:343) and the dedup HashSet is scoped per-file, so cross-file same-name symbols keep distinct doc_ids via distinct file_key. Full `extract::code` module: `59 passed; 0 failed`; reconcile code test: `1 passed`.

---

### Lens 39

VERDICT: NON_BLOCKING

---

### Lens 40



---

### Lens 41

Scope: PR #579 (fix/500-per-symbol-docs), head a7feced3, parent 4ac32d17. Toolchain rustup 1.98.0, RUST_MIN_STACK=8388608. Fix under review = commit a7feced3 "dedup per-declaration records by locator (#500)". PR diff touches only: extract/code.rs, sync_executor.rs (test only), and two http test files.

---

### Lens 42



---

### Lens 43

[NON_BLOCKING] The prior BLOCKING scenario is FIXED at the HTTP-harness level — first-index of the double-captured `export const Button = () => 1;` now reconciles instead of aborting. Evidence: I added `probe579_mts_button_reconciles_over_http` to incremental_reconcile_http_tests.rs (indexes a single `m.ts` with exactly that content through the real HttpEndpoint/run_index path). The FIRST `run_index` returned 0 and produced exactly 2 documents — the file-level doc plus one `Button` symbol doc (asserts `button_docs.len()==1` and `docs.len()==2` both passed). Before the dedup, @function+@const emitted two records under one `code:1:Button` locator → sealed 3 vs read-back 2 → reconcile barrier abort. The PR's own unit test `double_captured_declaration_emits_one_record_per_locator` also passes.

---

### Lens 44



---

### Lens 45

[NON_BLOCKING] No residual count-mismatch defect in any multi-capture shape. The dedup keys on `code:<line>:<name>`, and every capture of a *single* declaration shares that declaration's start line + name node, so all double/triple-capture shapes collapse to one emit. Verified:

---

### Lens 46

- `probe579_multi_capture_ts_reconciles` (several exported arrow-consts each double-captured @function+@const, plus class/interface): first index reconciled (==0), each of getUser/getPost/Service2/ROUTES is exactly one symbol doc.

---

### Lens 47

- `export const f = function(){}` — JS_Q's function-valued pattern admits `function_expression`, so it is @function+@const at one line; same locator → one emit (covered by the same dedup path and the file's own `javascript_exported_function_valued_const_is_both_kinds` premise).

---

### Lens 48

Sealed record count (`stats.records`) == distinct doc_id count within the extractor by construction: total = 1 file doc ("code") + N distinct per-decl locators, and distinct locators → distinct doc_ids under one (dataset,file_key). No shape I could build breaks the barrier.

---

### Lens 49



---

### Lens 50

[NON_BLOCKING] doc_id collision across DIFFERENT declarations sharing line+name does NOT break the barrier. `probe579_same_line_same_name_distinct_decls` indexes `class Foo {}; export const Foo = () => 1;` (a class Foo and a const Foo on one physical line → three captures, all locator `code:1:Foo`). Result: run_index==0, Foo collapses to 1 symbol doc, total 2 docs (printed: "Foo symbol docs = 1, total docs = 2"). This is silent data-loss (two genuinely distinct decls become one retrievable doc) but it is inherent to the line+name locator scheme, it does not desync sealed-vs-read-back (both = 1), and the file-level `defs`/`symbols` still list both. NIT, not blocking.

---

### Lens 51



---

### Lens 52

[NON_BLOCKING] 0-symbol file and empty-code guards behave. `probe579_zero_symbol_file` (`console.log(1+2);`) → run_index==0, exactly 1 document (the file doc), with no `code`/`name` fields. The `if code.is_empty() { continue; }` guard sits before the locator HashSet insert, and `code` is a pure function of the row (whole trimmed source line), so it cannot desync a locator.

---

### Lens 53



---

### Lens 54

[NON_BLOCKING — PRE-EXISTING, reproduced on parent] Re-indexing ANY code file that yields ≥1 symbol aborts on the second run with `project retained file <f>: field symbols is absent from frozen dataset docs`. Root cause is in unchanged code: `infer::FieldAcc::add` has no match arm for `Value::Object`, so the file-level `symbols` array (an array-of-objects) accumulates `n==0` and never receives a frozen `FieldSpec`; on re-index `reconcile_plan::ensure_compatible` (via `retained_slug`) requires every observed field to have a frozen spec and bails on `symbols`. This is NOT introduced by #500:

---

### Lens 55

- The PR diff does not touch reconcile_plan.rs, infer/mod.rs, or the file-doc `symbols` emission (that block is byte-identical at parent 4ac32d17).

---

### Lens 56

- Empirically reproduced on parent: I built a `4ac32d17` worktree, added `probe_parent_code_reindex` (index a plain `a.rs` struct/fn twice), and the second run failed with the identical `project retained file a.rs: field symbols is absent from frozen dataset docs`. So the abort predates the per-symbol-docs work and is a separate, product-wide code-re-index defect.

---

### Lens 57

Impact note for maintainers (not a #579 blocker): #500 does not fix or worsen this; the per-symbol docs' new scalar fields (name/kind/code/line/path) do get specs, so `symbols` remains the first field to trip the barrier exactly as on parent. The "stable-locator across re-index" claim in the PR is therefore currently unobservable end-to-end for code, because code re-index cannot complete on either branch.

---

### Lens 58



---

### Lens 59

Probes I ran (all in the two /tmp worktrees, not the main checkout): probe579_mts_button_reconciles_over_http, probe579_multi_capture_ts_reconciles, probe579_same_line_same_name_distinct_decls, probe579_zero_symbol_file (fix branch); probe_parent_code_reindex (parent). The two "*_reconciles" probes each FAILED only on their intentional second-run re-index assertion (the pre-existing symbols bug), after their first-index correctness assertions passed; same_line and zero_symbol (single-index) passed outright. Existing tests double_captured_declaration_emits_one_record_per_locator and code_files_index_with_ast_fields_and_the_run_reports_code_coverage both pass.

---

### Lens 60



---

### Lens 61

Bottom line for PR #579: the fix does what it claims (first-index double-capture reconciles; record count == doc count for all multi-capture shapes tested), and I could not reproduce any wrong result attributable to this PR. The only reproduced abort is the pre-existing code-re-index `symbols`-absent bug on parent 4ac32d17, which is out of scope for #579.

---

PR #581, opened 2026-08-22
fix(autoindex): don't abort re-index on a field absent-but-unscored in the frozen schema (#580)

### LINKED ISSUE #580: autoindex: re-indexing a modified code file aborts — `symbols` array field pruned from frozen schema then rejected
### Symptom
Re-indexing an already-indexed corpus that contains a **code file** aborts the run: `project retained file … / field symbols is absent from frozen dataset docs`. Surfaced by the adversarial verification of PR #579 (#500); **reproduced on `main` / parent `4ac32d17` as well as the PR branch**, so it is **pre-existing**, not introduced by #500.

### Root cause (from the reproduction)
The AST code file document carries `symbols` as a JSON **array**. `FieldAcc` never counts array/object values as scalar observations (`acc.n == 0`), so `infer_fields_with_policy` **prunes** `symbols` from the frozen dataset schema (`infer/mod.rs:435-437`). On the next generation, `reconcile_plan::ensure_compatible` (`reconcile_plan.rs:350-356`) walks every sketch field and **rejects** `symbols` as absent from the frozen dataset — fatal.

### Repro
Index a corpus with one code file (e.g. a `lib.rs` with a struct), edit it (e.g. `u32`→`u64`), re-run `run_index`. The second generation aborts. A probe `verify579_minimal_code_reindex_succeeds` fails identically on the #500 branch and on parent `4ac32d17`.

### Impact
Code files can be indexed at genesis but **never updated** — any change to a source file aborts the whole incremental run. (It also means #500's per-declaration documents are only ever created once and never reconciled on a file edit, until this is fixed.)

### Fix direction
Either treat array-typed fields as present-by-declaration in the frozen schema (don't prune a field that a producer still emits just because it has no scalar observations), or exempt structured/array fields from the `ensure_compatible` absence check. Add a re-index-of-a-code-file regression test.

Found via #500 / PR #579 verification.

xerj-org:
Closes #580.

## The defect
Re-indexing **any code file** aborted the entire run: `project retained file …: field symbols is absent from frozen dataset docs`. So a corpus with source files could be indexed once but never updated. (Found by PR #579's verification; reproduced on `main` — pre-existing, and it blocks #500's per-declaration docs from ever reconciling on a file edit.)

## Root cause
The AST file document carries a `symbols` array (of `{name,kind,line}` objects). `FieldAcc::add` deliberately ignores object values (`Value::Object(_) => {}`, "flattened upstream"), and its `Value::Array` arm recurses into elements — so an array **of objects** contributes no scalar observation: `symbols` accumulates `n == 0`, receives no frozen `FieldSpec` at genesis, and is pruned from the dataset schema. On the next generation `symbols` is observed again, and `ensure_compatible` (`reconcile_plan.rs`) rejected it as absent from the frozen schema → fatal.

## The fix
A field observed with **no scalar values** (`acc.n == 0`) has nothing to type-validate, so its absence from the frozen schema is not drift. `ensure_compatible` now skips such a field (mirroring `compatible_with_spec`'s own `acc.n == 0 => true`). A field that **does** carry scalar values yet is absent still fails the reconcile (real drift is still caught).

- **Fail-before proven** (revert only the `reconcile_plan.rs` hunk): the new `code_file_reindex_does_not_abort_on_absent_symbols_field` panics with the exact `field symbols is absent from frozen dataset docs` error; with the fix it reconciles and the struct's symbol doc survives.
- Full `xerj-autoindex` suite **green (696 passed, 0 failed)**; `clippy -D warnings` clean; ci-test profile checked.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -625,6 +625,51 @@ fn add_change_delete_and_rename_converge_over_real_http() {
     assert_eq!(journal_events(state_dir.path(), "sync_commit"), 5);
 }
 
+/// #580: re-indexing a **code file** must not abort. The AST file document
+/// carries a `symbols` array (of objects); `FieldAcc` does not scalar-type it,
+/// so it is pruned from the frozen dataset schema at genesis — and the second
+/// generation observed `symbols` again and aborted with `field symbols is
+/// absent from frozen dataset`. A field with no scalar values is now allowed to
+/// be absent from the frozen schema.
+#[test]
+fn code_file_reindex_does_not_abort_on_absent_symbols_field() {
+    let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
+    let _replay_guard = sync_executor::REPLAY_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let corpus = tempfile::tempdir().unwrap();
+    let state_dir = tempfile::tempdir().unwrap();
+    fs::write(
+        corpus.path().join("lib.rs"),
+        "pub struct AlphaConfig {\n    pub retries: u32,\n}\n",
+    )
+    .unwrap();
+    let endpoint = HttpEndpoint::start();
+    let config = cfg(corpus.path(), state_dir.path(), &endpoint.url, false);
+    // Genesis.
+    assert_eq!(run_index(config.clone()).unwrap(), 0);
+    // Edit the code file and re-index — before #580 this aborted the whole run
+    // on the frozen `symbols` field.
+    fs::write(
+        corpus.path().join("lib.rs"),
+        "pub struct AlphaConfig {\n    pub retries: u64,\n}\n",
+    )
+    .unwrap();
+    assert_eq!(
+        run_index(config).unwrap(),
+        0,
+        "re-indexing a code file must reconcile, not abort on the frozen `symbols` field (#580)"
+    );
+    // The struct is still indexed (file doc + its symbol doc).
+    assert!(
+        endpoint
+            .data_docs()
+            .iter()
+            .any(|d| d.get("name").is_some_and(|n| n == "AlphaConfig")),
+        "the struct's symbol doc must survive the re-index"
+    );
+}
+
 #[test]
 fn pending_generation_replays_sealed_source_after_live_source_mutates() {
     let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
--- a/engine/crates/xerj-autoindex/src/reconcile_plan.rs
+++ b/engine/crates/xerj-autoindex/src/reconcile_plan.rs
@@ -348,12 +348,24 @@ fn ensure_compatible(fields: &HashMap<String, FieldAcc>, dataset: &PlanDataset) 
         .map(|spec| (spec.name.as_str(), spec))
         .collect();
     for (name, acc) in fields {
-        let spec = specs.get(name.as_str()).with_context(|| {
-            format!(
+        let Some(spec) = specs.get(name.as_str()) else {
+            // A field observed with NO scalar values (`acc.n == 0`) — e.g. the
+            // code file's `symbols` sidecar (an array of objects, which
+            // `FieldAcc` deliberately does not scalar-type) — legitimately never
+            // received a frozen `FieldSpec` at genesis: it has nothing to type.
+            // Its absence is therefore not drift; there is nothing to validate,
+            // so mirror `compatible_with_spec`'s `acc.n == 0 => true` and skip
+            // it. Without this, re-indexing ANY code file aborted the whole run
+            // (#580), because `symbols` is observed on every generation but was
+            // pruned from the frozen schema. A field that DOES carry scalar
+            // values yet is absent is still a real incompatibility.
+            anyhow::ensure!(
+                acc.n == 0,
                 "field {name} is absent from frozen dataset {}",
                 dataset.slug
-            )
-        })?;
+            );
+            continue;
+        };
         anyhow::ensure!(
             compatible_with_spec(acc, spec),
             "field {name} is incompatible with frozen {} mapping in dataset {}",


xerj-org:

Workflow `wt9f0rw2b` (rustc 1.98.0, worktrees). **Result: 0 BLOCKING across all 3 lenses.** The one NIT (the shared `ensure_compatible` is also `classify_new`'s candidacy gate, so an absent-`n==0` field no longer disqualifies a candidate dataset) both lenses verified benign — only the code extractor surfaces a raw array-of-objects `n==0` field, and the object-field-absent-must-reject invariant still holds via the jaccard overlap threshold. Posting verbatim per process.

### Lens 1

VERDICT: NON_BLOCKING

---

### Lens 2



---

### Lens 3

[NON_BLOCKING] The new test is load-bearing and the fail-before evidence is genuine — PR #581 correctly fixes #580.

---

### Lens 4



---

### Lens 5

EVIDENCE / FAIL-BEFORE report:

---

### Lens 6



---

### Lens 7

- Fix present on branch `fix/580-code-reindex-symbols` as commit 953b6384, touching two files: `engine/crates/xerj-autoindex/src/reconcile_plan.rs` (+20/-4) and `engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs` (+45). Fix replaces the unconditional `specs.get(name).with_context(...)?` in `ensure_compatible` with a `let Some(spec) = ... else { anyhow::ensure!(acc.n == 0, "field {name} is absent..."); continue; }` — an absent field is tolerated only when `acc.n == 0`; an absent field with `n>0` still errors.

---

### Lens 8



---

### Lens 9

- PASS WITH FIX: `RUST_MIN_STACK=8388608 rustup run 1.98.0 cargo test -p xerj-autoindex --lib code_file_reindex_does_not_abort_on_absent_symbols_field` → `test result: ok. 1 passed`.

---

### Lens 10



---

### Lens 11

- FAIL WITHOUT FIX: reverted only `reconcile_plan.rs` to `953b6384~1` (the setup's `git stash push crates/...` found nothing because the fix is committed, not a working-tree diff; reverting the single file's hunk is the equivalent "revert only the reconcile_plan.rs hunk" the commit message itself describes). Re-run → `test result: FAILED`, panic:

---

### Lens 12

  ```

---

### Lens 13

  called `Result::unwrap()` on an `Err` value: project retained file lib.rs

---

### Lens 14

  Caused by:

---

### Lens 15

      field symbols is absent from frozen dataset docs

---

### Lens 16

  ```

---

### Lens 17

  Message matches the targeted error exactly ("field symbols is absent from frozen dataset docs").

---

### Lens 18



---

### Lens 19

- LOAD-BEARING / GENUINE RE-INDEX confirmed: the panic fires at `incremental_reconcile_http_tests.rs:659`, which is the SECOND `run_index(config).unwrap()`. The test does two real generations of a code file: genesis `run_index` on `lib.rs` (a Rust struct `AlphaConfig`), then it edits the file (`retries: u32` → `u64`) and calls `run_index` again — the retained-file re-index path (`project retained file lib.rs`). The failure is on the retained code file specifically, and the field named in the error is `symbols`, i.e. the AST file-doc's object array — exactly the pruned-field path #580 describes, not an unrelated absent field.

---

### Lens 20



---

### Lens 21

- Fix restored, worktree tree left clean (0 modified paths), verify worktree removed.

---

### Lens 22



---

### Lens 23

- Skeptic note (NIT, not blocking): the fix widens tolerance to ANY absent field with `acc.n == 0`, not just `symbols`. This mirrors `compatible_with_spec`'s existing `acc.n == 0 => true` rule and only affects fields that carried zero scalar values (nothing to type-validate), while `n>0` absent still errors — so it does not mask real scalar-type drift. No wrong result reproduced.

---

### Lens 24

Only my log files are untracked; the source tree matches the committed PR. Verdict below.

---

### Lens 25



---

### Lens 26

VERDICT: NON_BLOCKING

---

### Lens 27



---

### Lens 28

PR #581 (fix/580-code-reindex-symbols), LENS = REGRESSION. Commit 953b6384 on top of 21b00856. Ran with `rustup run 1.98.0` (rustc 1.98.0), `RUST_MIN_STACK=8388608`, in a dedicated worktree /tmp/verify581-REGRESSION.

---

### Lens 29



---

### Lens 30

[NON_BLOCKING] Full suite, dev profile — clean. `cargo test -p xerj-autoindex --no-fail-fast` EXIT=0. Real aggregates (6 binaries, 6 "test result:" lines, full run/result parity — no binary produced no result line):

---

### Lens 31

  - 682 passed; 0 failed; 0 ignored (12.70s)

---

### Lens 32

  - 8 passed; 0 failed

---

### Lens 33

  - 1 passed; 0 failed

---

### Lens 34

  - 1 passed; 0 failed

---

### Lens 35

  - 4 passed; 0 failed (1.20s)

---

### Lens 36

  - 0 passed; 0 failed

---

### Lens 37

  Total 696 passed / 0 failed — matches the commit's claimed 696. New test `incremental_reconcile_http_tests::code_file_reindex_does_not_abort_on_absent_symbols_field ... ok`.

---

### Lens 38



---

### Lens 39

[NON_BLOCKING] Full suite, ci-test profile — clean. `--profile ci-test --no-fail-fast` EXIT=0. Same 682/8/1/1/4/0 = 696 passed; 0 failed; 6/6 run-result parity. No binary missing a "test result:" line.

---

### Lens 40



---

### Lens 41

[NON_BLOCKING] Clippy — clean. `cargo clippy -p xerj-autoindex --all-targets -- -D warnings` EXIT=0, "Finished", zero warnings/errors (grep hits were only crate names thiserror/quick-error). The new let-else + `continue` control flow raises no lint.

---

### Lens 42



---

### Lens 43

[NON_BLOCKING] Fix is load-bearing (fail-before reproduced). Reverting only the reconcile_plan.rs hunk (restoring `with_context(...)?`) makes the new test FAIL: `panicked ... field symbols is absent from frozen dataset docs` (test result: FAILED, 1 failed). The fix targets exactly the #580 abort. File was restored afterward; `git diff` clean.

---

### Lens 44



---

### Lens 45

[NON_BLOCKING] No existing drift/compatibility test relied on old "absent-field-always-errors" behavior. Grep for `"is absent from frozen"` finds only the fix's own comment/test; no assertion elsewhere expects that error. The semantic "drift" tests (`pending_semantic_identity_drift_rejects...`, `semantic_resume_rejects_embedding_identity_drift...`) concern embedding-identity drift, a different code path, and both still pass.

---

### Lens 46



---

### Lens 47

[NIT] Investigated the one test that guards the invariant the fix touches — `reconcile_plan::tests::new_field_and_incompatible_type_fail_before_plan_changes` (line 605), whose object-field case (n==0, absent) must still be "rejected rather than dynamically mapped". It still passes, but note WHY: the fix relaxes `ensure_compatible` so it no longer rejects that object field — the rejection now comes solely from the overlap threshold in `classify_new` (1 scalar of 2 fields = 0.5 < 0.7). The invariant holds because an object/n==0 field only enlarges the `field_overlap` union (denominator), so it can never *cause* a new-file assignment; it can only ride along when scalar overlap already qualifies — which is the fix's intended behavior (the code file's `symbols` sidecar), and is identical to what already happens at genesis. No reproduced wrong result; not a regression.

---

### Lens 48



---

### Lens 49

No BLOCKING findings.

---

### Lens 50

VERDICT: NON_BLOCKING

---

### Lens 51



---

### Lens 52

PR #581 — CORRECTNESS lens (does skipping an absent field when `acc.n==0` hide real schema-drift that should abort?). Verified in worktree /tmp/verify581-CORRECTNESS on commit 953b6384; probes added to `engine/crates/xerj-autoindex/src/reconcile_plan.rs` and run with `rustup run 1.98.0`, plus the fix's own HTTP regression test.

---

### Lens 53



---

### Lens 54

[NON_BLOCKING] `acc.n` genuinely counts only scalar leaves, so "n==0 ⇒ nothing to type-validate" is accurate — `FieldAcc::add` (infer/mod.rs:129) increments `n` only for Bool/Number/String; `Null` and `Object` are ignored; `Array` recurses into non-array elements. Probe `probe_symbols_is_n0` confirms array-of-objects → n==0, null-only → 0, empty-array → 0, scalar string → 1. `code.rs:481` emits `symbols` as a raw `Value::Array` of `{name,kind,line}` objects with no `flatten_object`, so the genesis/sketch `FieldAcc` sees n==0 — the fix's premise holds.

---

### Lens 55



---

### Lens 56

[NON_BLOCKING] The "meaningful frozen type wrongly skipped" probe fails to refute. The new skip branch is the `else` of `specs.get(name)` — it fires ONLY when the field is absent from the frozen specs. A field that HAS a frozen type is routed through the unchanged `Some(spec)` → `compatible_with_spec` branch, which already returned `true` for n==0 (pre-fix behavior, unchanged). No field with a meaningful frozen type ever reaches the new code.

---

### Lens 57



---

### Lens 58

[NON_BLOCKING] "A type that legitimately changed but happens to have n==0" cannot be masked by the absent branch: an absent field has no frozen type, so there is no prior mapping to conflict with. For a type to be masked it must be frozen (present) → `Some` branch → untouched by this PR.

---

### Lens 59



---

### Lens 60

[NON_BLOCKING] No NEW mapping drift vs genesis. The only fields reaching `ensure_compatible` as absent+n==0 are: code `symbols` (emitted identically at genesis and re-index, never in the frozen schema either time — so the fix introduces no mapping the plan didn't already tolerate) and null-only/empty-array fields (index to nothing → no dynamic mapping). Flattening extractors (jsonl/json/yaml, extract/mod.rs:245-253) turn arrays-of-objects into JSON **strings** (n>0) and split objects into scalar subfields (n>0), so an absent field carrying real values is n>0 and still aborts — it never slips through this branch.

---

### Lens 61



---

### Lens 62

[NON_BLOCKING] Scoped-error branch verified intact. `probe_p2` (retained re-index, absent field `surprise` carrying a scalar string, n>0) still aborts with `field surprise is absent from frozen dataset`. The `anyhow::ensure!(acc.n == 0, …)` guard holds.

---

### Lens 63



---

### Lens 64

[NON_BLOCKING] Fail-before / fail-after proven end-to-end. Reverting only the `ensure_compatible` hunk makes the PR's HTTP regression test `code_file_reindex_does_not_abort_on_absent_symbols_field` panic with the exact `field symbols is absent from frozen dataset docs` on the second (re-index) generation; with the fix restored it passes and the `AlphaConfig` symbol doc survives. All 33 pre-existing reconcile/HTTP tests plus my 6 probes pass.

---

### Lens 65



---

### Lens 66

[NIT] Spillover into `classify_new` (new-file classification), not just retained re-index. `ensure_compatible` is also the candidacy GATE at reconcile_plan.rs:284, so an absent n==0 field no longer disqualifies a candidate dataset — assignment now rests solely on the jaccard threshold. `probe_p3` shows a NEW file with 3 shared fields + an absent n==0 `symbols` is now ASSIGNED (intersection 3 / union 4 ⇒ 30≥28), where pre-fix `ensure_compatible` would have disqualified it. This is benign in practice: only the code extractor surfaces a raw array-of-objects (n==0) field, and code files legitimately map to code datasets; flattening extractors never surface a bare n==0 object/array-of-objects (they become n>0 strings/subfields), so the tested "object field absent must be rejected" invariant (`new_field_and_incompatible_type_fail_before_plan_changes`) still holds — now enforced by the jaccard threshold rather than `ensure_compatible` (`probe_p3b` confirms a below-threshold absent n==0 field still bails with "unsupported dataset/schema evolution"). I could not construct a reproducible wrong mapping from any real extractor's output, so this is a widened fail-closed surface worth a one-line note, not a blocker.

---

### Lens 67



---

### Lens 68

Conclusion: the fix is correct and correctly scoped — absent + n==0 is skipped, absent + n>0 still aborts, and no real drift/incompatibility with a frozen type is hidden.

---

Issue #487, opened 2026-08-18
neural/proxy: re-indexing an unchanged corpus fails with "embedding execution identity changed" when it did not — reference coding is one-shot


xerj-org:
On `neural` and `proxy`, re-running the *identical* `xerj autoindex` command over an *unchanged* corpus fails, and the error says the embedding identity changed when it did not. Reference coding means re-indexing a peer repo as it moves; today that is a one-shot operation on these backends.

## Reproduction

Index once (succeeds), then run exactly the same command again:

```
error: embedding execution identity changed since this autoindex journal was created;
       refusing to mix vector spaces. No remote mutation was attempted.
       Restore the original identity, or rebuild with a new --state-dir and a new --prefix.
```

The identity did **not** change — journal pin `6e51e5ce3e46…` against live endpoint `6e51e5ce3e46…`, same server process, same model.

## Why the message is false

`engine/crates/xerj-autoindex/src/lib.rs:3210-3215` asserts a conjunction:

```rust
anyhow::ensure!(
    current.resumable
        && current.identity_sha256 == expected.embedding_identity_sha256
        && current.backend == expected.embedding_backend
        && current.dimensions == expected.embedding_dimension
        && current.semantic_contract == expected.embedding_semantic_contract,
    "embedding execution identity changed since this autoindex journal was created; ..."
);
```

Four of the five conjuncts compare `current` to `expected`. The first does not — `current.resumable` is a standing property of the backend, `false` by construction for `neural` and `proxy` because their identity hashes the configured model *name* rather than its bytes. So the assert fails on a conjunct that never had anything to do with change, and reports "identity changed" regardless.

`--fresh` does not help either: `--fresh cannot discard committed corpus generation 1`.

## Scope

This is the **third** site where `resumable` decides something it does not answer. The other two are addressed in #486 (they refused a *first* generation outright). This one is different and was deliberately left out of that PR: it guards a genuine resume, so the conjunction is in the right place — only the `resumable` term is misplaced and the message is wrong.

The graph path behaves the same way (second run → `server embedding backend cannot safely resume semantic indexing…`), so this is existing design across both paths rather than a regression.

## Why it matters for the main use case

[...]

PR #583, opened 2026-08-22
fix(autoindex): stop reporting false "identity changed" for a non-resumable backend (#487)

xerj-org:
Advances #487 (fixes the false/misleading message; the one-shot re-index limitation itself is option 2 / #367).

## The defect
Re-indexing an **unchanged** corpus on `neural`/`proxy` failed with `embedding execution identity changed since this autoindex journal was created; ... Restore the original identity` — but the identity had **not** changed (journal pin == live endpoint, same process, same model). The resume gate (`lib.rs`) folded `current.resumable` into the change-detection conjunction:

```rust
current.resumable && current.identity_sha256 == expected... && backend... && dimensions... && semantic_contract...
```

`resumable` is a **standing property** of the backend — `false` by construction for `neural`/`proxy`, whose identity hashes the model *name* not its bytes. So a resumable-only failure was misreported as an identity change, and the message sent users to a remedy (restore the identity, or `--fresh`) that **cannot work**.

## The fix
Split the check into a testable helper (`ensure_embedding_execution_unchanged_and_resumable`): the four **comparable** fields keep the `identity changed` message; **resumability** is asserted separately with an accurate message that surfaces the backend's own `non_resumable_reason` and points to the working remedy (rebuild with a new `--state-dir`/`--prefix`, or content-address the model — #367). **No behavior change** — the same cases still fail; only the message is now true.

- **Fail-before proven** (revert the split → the unchanged non-resumable case reports the false `identity changed`): the new `unchanged_reindex_on_non_resumable_backend_reports_resume_limit_not_identity_change` fails, then passes with the fix. It also asserts a genuine change still reports `identity changed`.
- Full `xerj-autoindex` suite **green (697 passed, 0 failed)**; `clippy -D warnings` clean; ci-test profile checked.

The real fix that lets these backends actually **resume** an unchanged corpus (content-addressed neural identity) is #367 (option 2); this PR removes the false diagnosis that was blocking users and misdirecting the remedy.

### DIFF
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -2176,6 +2176,45 @@ fn pending_for_phase_b(
         .collect()
 }
 
+/// #487: an unchanged re-index on a `neural`/`proxy` backend must not fail with
+/// a false "embedding execution identity changed" message. The identity CHANGED
+/// only if a comparable field differs from what the journal pinned; `resumable`
+/// is a standing property of the backend (false by construction for backends
+/// whose identity hashes the model NAME, not its bytes), so it is checked
+/// SEPARATELY with an accurate message — restoring the identity cannot fix a
+/// backend that simply cannot resume.
+fn ensure_embedding_execution_unchanged_and_resumable(
+    current: &crate::esclient::EmbeddingExecutionIdentity,
+    expected_sha: &str,
+    expected_backend: &str,
+    expected_dimension: Option<usize>,
+    expected_semantic_contract: &str,
+) -> anyhow::Result<()> {
+    anyhow::ensure!(
+        current.identity_sha256 == expected_sha
+            && current.backend == expected_backend
+            && current.dimensions == expected_dimension
+            && current.semantic_contract == expected_semantic_contract,
+        "embedding execution identity changed since this autoindex journal was created; \
+         refusing to mix vector spaces. No remote mutation was attempted. Restore the \
+         original identity, or rebuild with a new --state-dir and a new --prefix"
+    );
+    anyhow::ensure!(
+        current.resumable,
+        "the `{}` embedding backend cannot resume semantic indexing{}. The pinned identity is \
+         UNCHANGED — this is a backend limitation, not a change, so restoring the identity will \
+         not help. Rebuild with a new --state-dir and a new --prefix, or content-address the \
+         model identity so this backend can resume (#367).",
+        current.backend,
+        current
+            .non_resumable_reason
+            .as_deref()
+            .map(|r| format!(": {r}"))
+            .unwrap_or_default()
+    );
+    Ok(())
+}
+
 fn select_resume_plan_keys(
     files: &[walk::FileEntry],
     content_keys: &[String],
@@ -3170,17 +3209,13 @@ pub fn run_index_report(cfg: IndexCfg) -> Result<(i32, Option<Value>)> {
                     .any(|dataset| dataset.semantic_field.is_some())
                 {
                     let current = es.embedding_execution_identity()?;
-                    anyhow::ensure!(
-                        current.resumable
-                            && current.identity_sha256 == expected.embedding_identity_sha256
-                            && current.backend == expected.embedding_backend
-                            && current.dimensions == expected.embedding_dimension
-                            && current.semantic_contract == expected.embedding_semantic_contract,
-                        "embedding execution identity changed since this autoindex journal was \
-                         created; refusing to mix vector spaces. No remote mutation was attempted. \
-                         Restore the original identity, or rebuild with a new --state-dir and a \
-                         new --prefix"
-                    );
+                    ensure_embedding_execution_unchanged_and_resumable(
+                        &current,
+                        &expected.embedding_identity_sha256,
+                        &expected.embedding_backend,
+                        expected.embedding_dimension,
+                        &expected.embedding_semantic_contract,
+                    )?;
                 }
             }
             let summary = finish_generated_run(&es, &mut journal, &cfg)?;
@@ -6366,6 +6401,60 @@ mod duplicate_integration_tests {
                 Some(&["a".to_string(), "b".to_string(), "c".to_string()])
             ),
             HashSet::from(["a".to_string(), "b".to_string(), "c".to_string()])
+        );
+    }
+
+    /// #487: an UNCHANGED re-index on a backend with `resumable == false`
+    /// (`neural`/`proxy`) must report the resume LIMITATION, not a false
+    /// "identity changed" — and must surface the backend's own reason rather
+    /// than sending the user to a remedy (restore the identity) that cannot work.
+    #[test]
+    fn unchanged_reindex_on_non_resumable_backend_reports_resume_limit_not_identity_change() {
+        let current = crate::esclient::EmbeddingExecutionIdentity {
+            version: 1,
+            backend: "neural".into(),
+            identity_sha256: "6e51e5ce3e46".into(),
+            dimensions: None,
+            semantic_contract: "semantic_text-derived-vector.v1".into(),
+            resumable: false,
+            non_resumable_reason: Some("identity is derived from the model name".into()),
+        };
+        // All comparable fields match → the identity did NOT change.
+        let err = ensure_embedding_execution_unchanged_and_resumable(
+            &current,
+            "6e51e5ce3e46",
+            "neural",
+            None,
+            "semantic_text-derived-vector.v1",
+        )
+        .unwrap_err()
+        .to_string();
+        assert!(
+            err.contains("cannot resume") && !err.contains("identity changed"),
+            "must report the resume limitation, not a false identity change: {err}"
+        );
+        assert!(
+            err.contains("derived from the model name"),
+            "should surface the backend's own non_resumable_reason: {err}"
+        );
+        // A GENUINE change still reports "identity changed".
+        let changed = crate::esclient::EmbeddingExecutionIdentity {
+            identity_sha256: "DIFFERENT".into(),
+            resumable: true,
+            ..current.clone()
+        };
+        let err2 = ensure_embedding_execution_unchanged_and_resumable(
+            &changed,
+            "6e51e5ce3e46",
+            "neural",
+            None,
+            "semantic_text-derived-vector.v1",
+        )
+        .unwrap_err()
+        .to_string();
+        assert!(
+            err2.contains("identity changed"),
+            "a real identity change must still report it: {err2}"
         );
     }
 


xerj-org:

Workflow `wauazdyu2` (rustc 1.98.0, worktrees). **Result: 0 BLOCKING across all 3 lenses** — with a truth-table equivalence proof that the passing/failing partition is bit-identical (only the message changes). Posting verbatim per process.

### Lens 1

VERDICT: NON_BLOCKING

---

### Lens 2



---

### Lens 3

Verified the diff, the helper, the call site, all other resumable/identity gates, and ran the test under 1.98.0 (RUST_MIN_STACK=8388608) — it compiles and passes.

---

### Lens 4



---

### Lens 5

[NON_BLOCKING] Passing set is unchanged — the split did not weaken the gate. Original gate (lib.rs, run_index_report) returned Ok iff `resumable && sha==exp && backend==exp && dims==exp && contract==exp`, else ERROR "identity changed". New helper `ensure_embedding_execution_unchanged_and_resumable` does `ensure!(4 comparable fields match, "identity changed"); ensure!(resumable, "cannot resume");`. Ok is reached only when all 4 match AND resumable — bit-identical passing condition. Every input that failed before still fails (either the first or second ensure fires). No case that used to fail now passes. Confirmed by truth-table reasoning over the boolean conjunction: `A&&B&&C&&D&&E` fails ⟺ `!(B&&C&&D&&E)` (first ensure) OR `!(A)` given B..E (second ensure).

---

### Lens 6



---

### Lens 7

[NON_BLOCKING] `resumable==false` still ERRORs, not silent-proceed. For the exact #487 scenario (all 4 comparable fields match, resumable==false, neural backend): first ensure passes, second ensure fires → `anyhow::Error` "the `neural` embedding backend cannot resume...". Reproduced by the shipped test `unchanged_reindex_on_non_resumable_backend_reports_resume_limit_not_identity_change` which asserts the error contains "cannot resume" and NOT "identity changed" — it passed. The call site propagates with `?`, so the ERROR reaches the caller; there is no path that swallows it.

---

### Lens 8



---

### Lens 9

[NON_BLOCKING] Precedence correct for the changed-AND-non-resumable case. The two `ensure!` are sequential and `anyhow::ensure!` returns `Err` immediately, so the 4-field check is evaluated first. A backend that is both changed and non-resumable reports "identity changed" (the actionable condition — restoring identity is the right remedy), matching original behavior where the conjunction also failed with that message. Confirmed by the test's second case (`identity_sha256: "DIFFERENT"`, `resumable: true`) reporting "identity changed", and the same ordering guarantees the changed+non-resumable case does too.

---

### Lens 10



---

### Lens 11

[NON_BLOCKING] `non_resumable_reason` surfaced when present, degrades cleanly when None. `.as_deref().map(|r| format!(": {r}")).unwrap_or_default()` yields ": <reason>" for Some (test asserts "derived from the model name" appears) and empty string for None (message stays grammatical: "...cannot resume semantic indexing. The pinned identity is UNCHANGED..."). No panic/unwrap on None.

---

### Lens 12



---

### Lens 13

[NON_BLOCKING] No residual conflation at other call sites. `grep 'embedding execution identity changed'` returns two hits: the fixed site (lib.rs:3179 inside the helper) and state.rs:1096. state.rs `pin_embedding_identity` already separates resumable (bailed at line 1084) from the identity-change bail (1093-1101, compares only `existing != identity_sha256`) — never had the conflation. The other resumable gate (lib.rs:2752-2771, pending-generation verify) also checks `current.resumable` in its own dedicated `ensure!` with an accurate "cannot resume" message and keeps `current.resumable == expected.embedding_resumable` only as a legitimate comparable-field equality (current vs. expected, not the raw standing bool). No other site folds the standing `resumable` property into an "identity changed" message.

---

### Lens 14



---

### Lens 15

Claim of "no behavior change, only the message is now true" holds: identical fail/pass partition, `resumable==false` still errors rather than proceeding, correct precedence, accurate reason surfacing, and no other call site left with the old conflation.

---

### Lens 16

VERDICT: NON_BLOCKING

---

### Lens 17



---

### Lens 18

LENS = EVIDENCE / FAIL-BEFORE. I set up a fresh worktree on `fix/487-resumable-identity-check` (HEAD ac2ef885) and exercised the new test as a load-bearing regression guard. Toolchain: `rustup run 1.98.0`, `RUST_MIN_STACK=8388608`.

---

### Lens 19



---

### Lens 20

[NON_BLOCKING] New test passes on the fix — `unchanged_reindex_on_non_resumable_backend_reports_resume_limit_not_identity_change` compiles and passes (1 passed). It exercises the extracted helper `ensure_embedding_execution_unchanged_and_resumable` directly.

---

### Lens 21



---

### Lens 22

[NON_BLOCKING] FAIL-BEFORE confirmed — the test is load-bearing on exactly the split. I reverted ONLY the helper's split (folded `resumable` back into the change-detection conjunction and deleted the separate `ensure!`, restoring the pre-PR shape). Re-run FAILED at lib.rs:6420 with the message: "must report the resume limitation, not a false identity change: embedding execution identity changed since this autoindex journal was created; ...". So without the split, the UNCHANGED-but-non-resumable case (`resumable:false`, all 4 comparable fields matching) reports the false "identity changed" — the precise #487 bug. Restored via `git checkout`.

---

### Lens 23



---

### Lens 24

[NON_BLOCKING] Change detection was NOT deleted — the test's second half feeds a genuinely changed identity (`identity_sha256:"DIFFERENT"`, `resumable:true`) and asserts the error still contains "identity changed" (lib.rs:6446-6458). This is load-bearing: had the fix dropped the four-field comparison and kept only the resumable assert, that genuine-change input (resumable=true) would return `Ok`, and the test's `.unwrap_err()` would panic. It also asserts the non-resumable message surfaces the backend's own `non_resumable_reason` ("derived from the model name") and does not say "identity changed".

---

### Lens 25



---

### Lens 26

[NON_BLOCKING] "No behavior change" claim holds. Old gate: `ensure!(resumable && sha && backend && dims && contract, "identity changed")` → fails iff `!resumable || any_mismatch`. New gate: `ensure!(all4match, "identity changed")` then `ensure!(resumable, "cannot resume")` → fails iff `any_mismatch || !resumable`. Identical failure set; when both hold it still succeeds. Only difference is the message on the `all4match && !resumable` case (now the accurate resume-limit message pointing at #367). When a comparable field also differs, the first `ensure!` fires first and the message stays "identity changed" — same as before.

---

### Lens 27



---

### Lens 28

[NIT] There is a second, structurally similar gate at lib.rs:2791-2810 ("pending semantic generation" cutover flow). I confirmed it is NOT the #487 path and does not carry the bug: it already asserts `resumable` separately with its own accurate message BEFORE the change-detection `ensure!`. Out of scope for this PR; no regression.

---

### Lens 29



---

### Lens 30

[NON_BLOCKING] Full suite green — `cargo test -p xerj-autoindex --lib`: 683 passed, 0 failed (the new test included; matches the 682-filtered count seen when running the single test).

---

### Lens 31



---

### Lens 32

Could not refute the claim. The fix changes only the message on the unchanged-non-resumable case, preserves the exact failure set, keeps real change detection, and the new test genuinely fails when the split is undone.

---

### Lens 33

All suites pass on both profiles, clippy clean, new test executes. Writing verdict.

---

### Lens 34



---

### Lens 35

VERDICT: NON_BLOCKING

---

### Lens 36



---

### Lens 37

[NON_BLOCKING] The fix is behavior-preserving on the failing set — only the message changes, as claimed — proven by boolean algebra and confirmed by the full suite.

---

### Lens 38



---

### Lens 39

Equivalence proof (the load-bearing claim). Let f1..f4 be the four comparable-field matches (sha, backend, dimensions, semantic_contract) and R = current.resumable.

---

### Lens 40

- OLD gate: `ensure!(R && f1 && f2 && f3 && f4, "identity changed")` → run FAILS iff `!(R ∧ f1 ∧ f2 ∧ f3 ∧ f4)`.

---

### Lens 41

- NEW gate: `ensure!(f1 ∧ f2 ∧ f3 ∧ f4, "identity changed")` then `ensure!(R, "cannot resume")` → FAILS iff `!(f1∧f2∧f3∧f4) ∨ (f1∧f2∧f3∧f4 ∧ !R)` = `!(f1∧f2∧f3∧f4) ∨ !R` = `!(f1∧f2∧f3∧f4∧R)`.

---

### Lens 42

The two failing sets are identical. The only observable difference is the message: when all four comparable fields match but R is false (neural/proxy unchanged re-index), OLD said "identity changed" + an unworkable remedy; NEW asserts `resumable` separately with an accurate "cannot resume" message that surfaces `non_resumable_reason` and points at #367. A genuine field mismatch still trips the first ensure first, so real changes still read "identity changed" (verified by the test's second assertion). No behavior change confirmed.

---

### Lens 43



---

### Lens 44

Refutation attempts, all failed:

---

### Lens 45

- Missed duplicate gate? state.rs:1096 carries the same "identity changed" string, but it is a DIFFERENT gate (`pin_embedding_identity`, write-time journal pinning) that ALREADY separates the resumable check (state.rs:1083, its own message) from the identity-changed check. It is out of scope for the run_index_report re-index gate this PR fixes and is not a missed spot.

---

### Lens 46

- Existing test pinned to the OLD message for a resumable-only failure? None. The two tests asserting "refusing to mix vector spaces" (state.rs:173, failure_resume_http_tests.rs:1171) exercise the state.rs genuine-sha-mismatch path, not the run_index_report resumable-only path — and both still pass. The only test-side references to "identity changed" are in the new test (lib.rs:6440, 6456).

---

### Lens 47

- Option comparison / typing: `current.dimensions == expected_dimension` compares `Option<usize>` to `Option<usize>` — correct, matches the old inline comparison.

---

### Lens 48



---

### Lens 49

Real aggregates (rustc 1.98.0, RUST_MIN_STACK=8388608):

---

### Lens 50



---

### Lens 51

Dev profile (`cargo test -p xerj-autoindex --no-fail-fast`) — every binary:

---

### Lens 52

- unittests src/lib.rs: `test result: ok. 683 passed; 0 failed; 0 ignored`

---

### Lens 53

- follow_symlinks_boundary: ok. 8 passed; 0 failed

---

### Lens 54

- journal_lock_inheritance: ok. 1 passed; 0 failed

---

### Lens 55

- scan_pool_width: ok. 1 passed; 0 failed

---

### Lens 56

- sniff_printable_magic: ok. 4 passed; 0 failed

---

### Lens 57

- Doc-tests: ok. 0 passed; 0 failed

---

### Lens 58

No failing test, no binary aborted without a "test result:" line.

---

### Lens 59



---

### Lens 60

CI-test profile (`--profile ci-test --no-fail-fast`) — every binary:

---

### Lens 61

- unittests src/lib.rs: `test result: ok. 683 passed; 0 failed; 0 ignored` (6.82s)

---

### Lens 62

- follow_symlinks_boundary: ok. 8 passed; 0 failed

---

### Lens 63

- journal_lock_inheritance: ok. 1 passed; 0 failed

---

### Lens 64

- scan_pool_width: ok. 1 passed; 0 failed

---

### Lens 65

- sniff_printable_magic: ok. 4 passed; 0 failed

---

### Lens 66

- Doc-tests: ok. 0 passed; 0 failed

---

### Lens 67

Identical counts to dev; no failures.

---

### Lens 68



---

### Lens 69

Clippy (`cargo clippy -p xerj-autoindex --all-targets -- -D warnings`): `Finished dev profile ... in 15.14s`, no warnings, no errors (would have failed under -D warnings otherwise).

---

### Lens 70



---

### Lens 71

New test executes and passes: `duplicate_integration_tests::unchanged_reindex_on_non_resumable_backend_reports_resume_limit_not_identity_change ... ok`.

---

### Lens 72



---

### Lens 73

No regression, no wrong result reproduced. The claim (no behavior change; message is now accurate) holds.

---

PR #584, opened 2026-08-22
fix(autoindex): refuse a graph-path re-run of a --no-graph corpus before it mutates (#490)

### LINKED ISSUE #490: autoindex: recovering a corpus after a classification fix is undocumented and configuration-dependent (and one cross-path case mutates)
When a classification fix means a previously-skipped file should now be indexed, what a user must do to recover an existing corpus depends on the state dir in ways nothing documents. I wrote the rc.18 note for this four times and got it wrong four times; the measurements below are the reason, and they are worth having in one place rather than rediscovered per release.

All measured on `67f732b2` with two binaries: the build at that head, and the same build with `(b"BM", "bmp", bmp_file_header)` reverted to `(b"BM", "bmp", accept)` to recreate the rc.17 behaviour. Corpus indexed with the second, recovered with the first.

## Measured

| indexed | recovered with | outcome |
|---|---|---|
| graph | plain re-run | exit **3**, `_count` unchanged, `files=0` — never reconsidered |
| graph | `--fresh` | works (`_count` 5 → 10) |
| graph | new `--state-dir` + `--prefix` | works |
| `--no-graph`, **CSV** | plain re-run | exit **0**, fully recovered (60 → 120 docs) |
| `--no-graph`, **prose** | plain re-run | exit **1**, `no frozen dataset accepts family txt-prose … use a new prefix` |
| `--no-graph` | `--fresh` | **refused** — `cannot discard committed corpus generation 1` |
| `--no-graph` | new `--state-dir` + `--prefix` | works |
| graph, then `--no-graph` re-run | — | exit 1, different message and remedy (`lib.rs:3207`) |
| `--no-graph`, then graph re-run | — | exit 3, and it **mutates the destination** (79 → 80 docs); `--fresh` then refused |

## What actually decides it

Not the flag. Two independent mechanisms:

1. **Whether the plan is re-sniffed at all.** The graph path reuses the frozen plan; `pending_for_phase_b` (`lib.rs:2170-2176`) selects only keys in `plan.files`, and a junked file is in `plan.junk_files`. `--no-graph` re-scans.

2. **Whether a frozen dataset accepts the file's family.** `reconcile_plan.rs:282` filters candidates on `dataset.family == family && &dataset.group == group`. A CSV rejoins a `csv` dataset; `txt-prose` can never match a `docs` dataset, so the same path recovers one and refuses the other. This is why "the `--no-graph` path aborts" is false as a general statement — it aborts for prose and succeeds for CSV.

`--fresh`'s refusal is likewise a property of the state dir: `blocking_generation` (`lib.rs:2945`) keys on `preflight.pending_sync` and a committed manifest past genesis, and is **not** gated on `cfg.no_graph`. So it can be refused on an invocation that does not pass `--no-graph` at all, if the state dir was built that way.

## Why this is worth fixing rather than documenting

Three things are wrong here independent of any release note.

- **Exit 3 is indistinguishable from "nothing to do".** A graph re-run after the fix reports `completed-with-junk` with the same code as the original run, so an operator watching exit codes sees no change and has no signal that a recovery is available or needed.
- **The cross-path case mutates.** Indexing `--no-graph` and then re-running with the default path proceeds and writes to the destination, then leaves `--fresh` refused. That is a state a user can reach by omitting a flag they used last week.
- **The remedy the error names is the expensive one.** "use a new prefix" means re-extracting and, on a semantic index, re-embedding the whole corpus — on `apache/lucene`, hours (see #366) — to pick up one file. When a compatible dataset exists, the plain re-run already does it for free; nothing tells the user which case they are in.

## Suggested shape

1. Make the junk record carry the sniff verdict that produced it, so a re-run can notice that the verdict changed and re-admit the file on the graph path instead of skipping it forever.
2. Distinguish "junk remains, unchanged" from "junk was re-evaluated and is now indexable" in the exit code or the summary — the second is actionable, the first is not.
3. Refuse the cross-path re-run before it mutates, the way the same-path mismatch is refused.
4. When the family mismatch is what blocks recovery, say so and name the cheaper remedy if one exists, rather than jumping to "use a new prefix".

Filed from the round-4 verification of #460. The rc.18 note now states only the part that is certainly true and points here for the matrix.

xerj-org:
Addresses **suggested fix #3** of #490 — _"Refuse the cross-path re-run before it mutates, the way the same-path mismatch is refused."_ This does **not** close #490; its other suggestions (junk-verdict tracking #1, the exit-code distinction #2, and the family-mismatch remedy message #4) are separate, larger enhancements and remain open.

## The bug (final row of the #490 matrix)

| indexed | recovered with | outcome (before) |
|---|---|---|
| `--no-graph`, then **graph** re-run | — | exit 3, **mutates the destination** (79 → 80 docs); `--fresh` then refused |

A corpus committed by the `--no-graph` generated path carries a committed **generation manifest**, which is graph-disabled *by construction* — `sync::validate_manifest` requires `!execution.graph_enabled` for every incremental generation. Re-running that same state dir on the default **graph** path fell through both `--no-graph` branches, opened the journal for write, entered the graph reconcile, and **mutated the destination** (publishing new documents and writing a legacy plan record over the committed manifest) before the mismatch ever surfaced — and it surfaced only as the opaque `legacy plan write cannot follow a committed generated manifest`, leaving `--fresh` wedged (`cannot discard committed corpus generation 1`). It is a state a user reaches simply by omitting a `--no-graph` flag they used last week.

## The fix

Refuse the cross-path re-run **before** `open_after_preflight` (the first write) and before `gc_snapshots`, symmetric to the existing graph→no-graph refusal a few lines above:

```rust
if !cfg.no_graph && !genesis_recovery {
    if let Some(committed) = preflight.committed_manifest.as_ref() {
        anyhow::bail!("this corpus was indexed with --no-graph (committed generation {}); …", committed.generation);
    }
}
```

Because the `--no-graph`/committed re-run returns earlier (the `cfg.no_graph && committed_manifest.is_some() && !genesis_recovery` branch), reaching this guard with a committed manifest already implies the graph path. And since a committed manifest is graph-disabled by construction, a legitimate **graph→graph** re-run (whose `committed_manifest` is `None`) is untouched — no risk to the primary path.

**Before → after:** the opaque internal error becomes a clear, actionable refusal naming both remedies (re-run with `--no-graph`, or rebuild under a new `--state-dir`/`--prefix`), and **no remote mutation is attempted**.

## Verification

- **New E2E regression test** `no_graph_committed_corpus_refuses_a_graph_path_rerun_before_mutating` (real in-process HTTP endpoint): commit generation 1 with `--no-graph`, grow the corpus, re-run on the graph path → asserts the refusal message **and** that the added row was not published and no second generation was committed.
  - Fail-before: `got: legacy plan write cannot follow a committed generated manifest` (the opaque error, after journal open).
  - Pass-after: green.
- `cargo test -p xerj-autoindex` (dev): **698 passed, 0 failed**.
- `cargo fmt -- --check`: clean.
- `cargo clippy -p xerj-autoindex --all-targets`: exit 0 under rustc **1.92.0** and **1.98.0**.
- `cargo test -p xerj-autoindex --profile ci-test` (rustc 1.98.0), reconcile module: **27 passed, 0 failed**.

Refs #490

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -768,6 +768,60 @@ fn partially_applied_data_bulk_leaves_the_generation_pending_and_the_retry_conve
     );
 }
 
+/// #490: a corpus committed by the `--no-graph` generated path must not be
+/// silently re-runnable on the default *graph* path. A committed generation
+/// manifest is graph-disabled by construction (`sync::validate_manifest`), so
+/// the graph path can recognise the mismatch and refuse it — the way the
+/// graph→no-graph direction is already refused. Before this guard the graph
+/// re-run reconciled against the no-graph plan and *mutated the destination*
+/// (publishing the newly added row) before the mismatch was ever noticed, then
+/// left `--fresh` refused. The refusal must land before the journal is opened
+/// for write, so no remote mutation is attempted.
+#[test]
+fn no_graph_committed_corpus_refuses_a_graph_path_rerun_before_mutating() {
+    let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
+    let _replay_guard = sync_executor::REPLAY_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let corpus = tempfile::tempdir().unwrap();
+    let state_dir = tempfile::tempdir().unwrap();
+    let source = corpus.path().join("rows.csv");
+    fs::write(&source, "id,value\n1,first\n").unwrap();
+    let endpoint = HttpEndpoint::start();
+
+    // Commit generation 1 with the `--no-graph` generated executor.
+    let no_graph = cfg(corpus.path(), state_dir.path(), &endpoint.url, false);
+    assert_eq!(run_index(no_graph.clone()).unwrap(), 0);
+    assert_eq!(journal_events(state_dir.path(), "sync_commit"), 1);
+    let committed = endpoint.data_docs().len();
+    assert!(committed > 0, "the no-graph run published documents");
+
+    // Grow the corpus so a graph re-run would have a new row to publish, then
+    // re-run the *same* state dir on the default graph path (`no_graph: false`).
+    fs::write(&source, "id,value\n1,first\n2,second\n").unwrap();
+    let mut graph = no_graph.clone();
+    graph.no_graph = false;
+    let error = run_index(graph).unwrap_err();
+    let rendered = format!("{error:#}");
+    assert!(
+        rendered.contains("indexed with --no-graph"),
+        "the graph re-run must be refused with the cross-path message, got: {rendered}"
+    );
+
+    // The refusal must land before any mutation: the added row was not
+    // published, and no second generation was committed.
+    assert_eq!(
+        endpoint.data_docs().len(),
+        committed,
+        "the refused graph re-run must not publish the added row"
+    );
+    assert_eq!(
+        journal_events(state_dir.path(), "sync_commit"),
+        1,
+        "no second generation may be committed by the refused cross-path re-run"
+    );
+}
+
 #[test]
 fn pending_semantic_identity_drift_rejects_before_any_further_bulk() {
     let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -768,17 +768,27 @@ fn partially_applied_data_bulk_leaves_the_generation_pending_and_the_retry_conve
     );
 }
 
-/// #490: a corpus committed by the `--no-graph` generated path must not be
-/// silently re-runnable on the default *graph* path. A committed generation
+/// #490 (fix #3): a corpus committed by the `--no-graph` generated path must not
+/// be silently re-runnable on the default *graph* path. A committed generation
 /// manifest is graph-disabled by construction (`sync::validate_manifest`), so
-/// the graph path can recognise the mismatch and refuse it — the way the
-/// graph→no-graph direction is already refused. Before this guard the graph
-/// re-run reconciled against the no-graph plan and *mutated the destination*
-/// (publishing the newly added row) before the mismatch was ever noticed, then
-/// left `--fresh` refused. The refusal must land before the journal is opened
-/// for write, so no remote mutation is attempted.
+/// the graph path recognises the mismatch and refuses it — the way the
+/// graph→no-graph direction is already refused.
+///
+/// Scope of what this harness proves. Without the guard the re-run still fails,
+/// but only later and opaquely: `write_plan`'s own `ensure!` rejects it with
+/// `legacy plan write cannot follow a committed generated manifest`, *after*
+/// `open_after_preflight` and `gc_snapshots` have run. On real binaries whose
+/// graph path reaches Phase B before that point the destination is mutated first
+/// (#490 matrix: 79 → 80 docs); this in-process harness is backstopped by that
+/// `ensure!`, so it exercises the *message and ordering*, not the remote row.
+/// The load-bearing contract pinned here is therefore: the run is refused with
+/// the clear cross-path message and NOT the opaque `write_plan` one (i.e. the
+/// guard preempts it, before the journal is opened for write). The destination
+/// staying clean is asserted as a belt-and-suspenders invariant — true here
+/// regardless of the guard, so a future change that lets the graph path reach
+/// Phase B before refusing is caught.
 #[test]
-fn no_graph_committed_corpus_refuses_a_graph_path_rerun_before_mutating() {
+fn no_graph_committed_corpus_is_refused_on_the_graph_path_with_a_clear_message() {
     let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
     let _replay_guard = sync_executor::REPLAY_FAILPOINT_TEST_LOCK
         .lock()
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification (adversarial, each refuting in its own worktree under the CI toolchain)

Verdicts recorded verbatim before acting. **2 NON_BLOCKING, 1 BLOCKING (evidence lens).** The BLOCKING finding was valid and has been addressed in `c67e69ae` (test only; the fix in `041afcc6` is unchanged). Details below.

> The discriminator `!cfg.no_graph && !genesis_recovery && committed_manifest.is_some()` is correct under the CORRECTNESS lens. Evidence:
> - No false refusal on a legitimate graph→graph run. A committed manifest past genesis is only ever written on the no-graph path: `begin_non_graph_generation` (lib.rs:418-421) asserts `cfg.no_graph`, and it is the sole caller of `sync_bootstrap_genesis`/the sync-commit path. The graph path never emits `sync_bootstrap`/`sync_commit`, so `!no_graph` cannot yield a non-genesis `committed_manifest`.
> - The one way a graph corpus gets a committed_manifest is `apply_legacy_plan` on an *empty* plan → `bootstrap_legacy` → `LegacyBootstrap::Ready` gen-0 genesis (sync.rs:246-258, state.rs:507-509). That is exactly `generation==0 && groups.is_empty()`, which `genesis_recovery` (lib.rs:2968-2971) makes true, so `!genesis_recovery` suppresses the guard. A *non-empty* legacy plan → `MigrationRequired` → `committed=None` (state.rs:511-514). Invariant holds.
> - `sync::validate_manifest` (sync.rs:742-748) forces `!graph_enabled` for every gen>0 manifest, confirming "committed ⇒ graph-disabled."
> - Placement precedes mutation for the target case: for a committed (non-pending) corpus, `pending_sync` is None so lib.rs:3043 is skipped, both no_graph branches (3108, 3242) are skipped on the graph path, and the guard (3302) runs before `open_after_preflight`/`gc_snapshots` (only read-only walk/hash precede it).
> - Test verified: fail-before (guard condition forced false) reproduces the pre-fix opaque error `legacy plan write cannot follow a committed generated manifest`, so the test genuinely guards the change.
>
> Non-blocking follow-up (not a defect in this guard): the `pending_sync` branch (lib.rs:3043) fires regardless of `cfg.no_graph`, so a graph re-run of a no-graph corpus with an *uncommitted pending* generation would resume-and-commit it before reaching this guard. Pre-existing behavior outside issue #490 fix #3's committed-corpus scope, not a regression introduced here.

> Ran the full suite on the branch with the CI toolchain (`cargo +1.98.0 test -p xerj-autoindex --lib`): `684 passed; 0 failed`. Targeted `incremental_reconcile_http_tests`: `27 passed; 0 failed`. No existing graph-path re-run test newly bails — the regression the lens asks about does not exist empirically.
> The strongest adversarial angle actually confirms the fix. In `state.rs:657`/`:956`, on replay of a `sync_bootstrap`/`sync_commit` record preflight sets `plan = committed_manifest.map(|m| m.plan.clone())` — so a no-graph-committed corpus reaches `run_index_report` with BOTH `committed_manifest = Some` AND `plan = Some`. Pre-fix the graph path fell straight into `if let Some(prior_plan) = preflight.plan.as_ref()` and reconciled the manifest's plan as if it were a legacy graph plan — exactly the "mutate then fail opaquely" bug. Placing the guard before that branch is the correct interception point. The guard cannot fire on a legitimate graph→graph resume (`validate_manifest` requires `!graph_enabled`; a real graph corpus carries `committed_manifest = None`); `--fresh` is intercepted earlier by `UnsafeFreshGenerationError`; `genesis_recovery` is excluded.
> One out-of-scope follow-up (NON_BLOCKING): a no-graph corpus in the *pending* genesis-bootstrap state sets `genesis_recovery=true`, so a graph re-run over it skips this guard; the fix only covers *committed* generations, matching #490 fix #3's scope.

### 🔴 EVIDENCE — BLOCKING (addressed in `c67e69ae`)
> the new test does not prove the fix's headline property ("refuses before mutating"); two of its three assertions are vacuous.
> 1. Fail-before reconstruction — deleting only the guard block and running the test fails at exactly ONE line — the substring assertion — with `got: legacy plan write cannot follow a committed generated manifest`. So the ONLY load-bearing assertion is the message substring.
> 2. The two "no-mutation" assertions are inert. With the guard removed: `docs_after=1 committed=1` and `sync_commit=1` — i.e. even WITHOUT the fix the destination is not mutated and no second generation commits, so `data_docs().len() == committed` and `sync_commit == 1` pass identically before and after.
> 3. Root cause the test can't see: on the graph path `write_plan` runs BEFORE Phase B publishing, and its own `anyhow::ensure!(... "legacy plan write cannot follow a committed generated manifest")` (state.rs:1198) already blocks the cross-path write before any remote `_bulk`. So in this harness the pre-existing ensure! prevents the mutation; the new guard only relocates the refusal earlier and swaps an opaque message for a friendly one.
> 4. Overclaim: the test comment "the added row was not published" and the name `..._before_mutating` attribute to the new guard a no-mutation property that holds without it. The issue's 79→80 real-binary increase is NOT reproduced by this harness.
> 5. Fair points: the substring `"indexed with --no-graph"` is a stable anchor; the guard's correctness/no-false-positive claim checks out.
> Fix before merge: either reproduce the actual pre-guard remote mutation, or drop/reword the vacuous assertions and the "before any mutation" comment to claim only what the harness proves.

### Resolution of the BLOCKING finding (`c67e69ae`)
The evidence lens is correct: this harness cannot exhibit the remote-row mutation (`write_plan`'s `ensure!` backstops it after `open_after_preflight`/`gc_snapshots`), so the only thing observable here is the **message and ordering**. I did **not** overclaim my way past it. Changes:
- **Renamed** `..._refuses_a_graph_path_rerun_before_mutating` → `..._is_refused_on_the_graph_path_with_a_clear_message`.
- **Added a genuinely load-bearing assertion**: `!rendered.contains("legacy plan write cannot follow a committed generated manifest")` — fail-before that opaque phrase *is* the message (re-verified by disabling the guard), so the pair (contains-clear + not-contains-opaque) proves the guard preempts the late error.
- **Documented** the destination-clean assertions as belt-and-suspenders invariants (true here regardless of the guard; they catch a future change that lets the graph path reach Phase B before refusing), and the docstring now states plainly that the 79→80 remote mutation needs a real binary and is out of this harness's reach.
- Full `xerj-autoindex` suite **698/0**; fmt clean; clippy 0 under 1.98.0.

The two NON_BLOCKING follow-ups (pending-genesis-bootstrap cross-path, `pending_sync` resume-and-commit before the guard) are real and out of #490 fix #3's committed-corpus scope — filing as a follow-up issue.

xerj-org:
### Evidence-lens re-verification on `c67e69ae` (BLOCKING → resolved)

An independent re-check of the previously-BLOCKING evidence finding, on the updated commit, returns **NON_BLOCKING** (verbatim):

> VERDICT: NON_BLOCKING
>
> - Fail-before (guard flipped at `lib.rs:3307` from `!cfg.no_graph` to `cfg.no_graph`, diff confirmed to touch only that line): test FAILED at `incremental_reconcile_http_tests.rs:820:5` with rendered = `legacy plan write cannot follow a committed generated manifest` — exactly the opaque `write_plan` backstop. Both message assertions are broken by disabling the guard, so the message contract is genuinely load-bearing (not vacuous).
> - Pass-after (lib.rs restored, `git diff --stat` empty): `1 passed`. Restore verified twice.
> - Honesty of scope: the docstring now explicitly states the harness "exercises the *message and ordering*, not the remote row" and that the 79→80 mutation "needs a real binary whose graph path reaches Phase B." The destination-clean assertions are labeled "belt-and-suspenders ... true here regardless of the guard" — directly conceding the vacuousness the prior finding flagged rather than attributing a no-mutation property to the guard.
> - Name change removes the misleading "before_mutating"; the name now matches what the harness proves.
> - Remaining nit (non-blocking): the positive `contains("indexed with --no-graph")` already fires first, so the added `!contains` is more belt-and-suspenders than uniquely load-bearing; honestly framed. No overclaim remains.

**Verification tally: 3 NON_BLOCKING, 0 BLOCKING.** Merge gate (b) satisfied. Holding on gates (a) all-checks-green and (c) CLEAN before merging.

PR #586, opened 2026-08-22
fix(autoindex): sniff markdown-with-frontmatter as its body, not YAML (#551)

### LINKED ISSUE #551: sniff: a leading --- is an unqualified YAML marker — markdown with frontmatter loses every heading and block after the first paragraph
Filed automatically on behalf of a human operator (agent-driven, Claude Opus 5). Same *class* as #379/#380/#403 — an unqualified signature capturing prose — but a different code path, so it is filed separately rather than widened into those.

## Summary

`sniff.rs` gives every printable-ASCII magic in `MAGIC_TABLE` a structural qualifier, and #403 added one for `%PDF-`. The YAML branch has none:

```rust
// YAML
if nonblank.first().map(|l| l.trim() == "---").unwrap_or(false)
    || yaml_line_ratio(nonblank) >= 0.6
{
    return Family::Yaml;
}
```

`engine/crates/xerj-autoindex/src/sniff.rs:1110-1114`, byte-identical between tags `v1.0.0-rc.18` and `v1.0.0-rc.23`.

A leading `---` is a YAML document-start marker *and* the opening delimiter of markdown frontmatter, which is the dominant convention for markdown corpora (Jekyll, Hugo, Obsidian, Docusaurus, MkDocs, Astro) and for Claude Code skill files — including the two this repository ships. There is no lookahead for a closing `---` followed by non-YAML, so every such file is handed to the YAML extractor, which keeps the frontmatter mapping and the first prose paragraph and discards the rest.

## Reproduction

Clean data dir, default endpoint, `v1.0.0-rc.23`.

```sh
D=$(mktemp -d); cd "$D"; mkdir corpus

printf -- '# Outage postmortem\n\nThe sprocket pool was exhausted.\n\n## Resolution\nRaised the sprocket pool size from 4 to 16.\n' > body.md
printf -- '---\ntitle: Widget outage\ndate: 2026-01-15\ntags: [widget, outage]\n---\n\n' > corpus/with-frontmatter.md
cat body.md >> corpus/with-frontmatter.md
cp body.md corpus/no-frontmatter.md

# the two bodies are byte-identical; only the frontmatter differs
tail -n +7 corpus/with-frontmatter.md | cmp - corpus/no-frontmatter.md && echo IDENTICAL

xerj --insecure --data-dir "$(mktemp -d)" &
until curl -fs -m1 localhost:9200/_cluster/health >/dev/null; do sleep 0.25; done
xerj autoindex "$D/corpus" --no-graph
```

`IDENTICAL`, then `autoindex` exits **3** (`completed-with-junk`, documented success) reporting `0 junk/skipped files`.

**Classification** — the only difference between the two files is the frontmatter:

```sh
curl -s localhost:9200/autoindex-catalog/_search -H 'Content-Type: application/json' \
  -d '{"size":10,"query":{"term":{"doc_kind":"file"}},"_source":["path","format","status"]}'
```

```json
{"took":0,"timed_out":false,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0},"hits":{"total":{"value":2,"relation":"eq"},"max_score":0.8938179,"hits":[{"_index":"autoindex-catalog","_id":"file:axf2-01bca00d4d52a107c7a62a2330e92b5a-71","_score":0.8938179,"_source":{"path":"no-frontmatter.md","format":"txt-prose","status":"indexed"}},{"_index":"autoindex-catalog","_id":"file:axf2-33b5a34efbbf1f94c425619c41474a56-b7","_score":0.8938179,"_source":{"path":"with-frontmatter.md","format":"yaml","status":"indexed"}}]}}
```

Both are `status: indexed`. One is `txt-prose`, the other `yaml`.

**The loss** — this sentence is present verbatim in *both* files:

```sh
curl -s 'localhost:9200/ax-*/_search' -H 'Content-Type: application/json' \
  -d '{"size":5,"query":{"multi_match":{"query":"Raised the sprocket pool size from 4 to 16","fields":["body","text","value"],"type":"phrase"}},"_source":["ax_path"]}'
```

```json
{"_xerj":{"hints":[{"code":"unknown_field","reason":"no results, and this query names fields that do not exist in `ax-*`: body, text. Searchable text fields here are: tags, title, value.","try":{"request":"POST /ax-*/_search","body":{"query":{"match":{"tags":"<your text>"}},"_source":["ax_path","title"],"fields":["_passage"]}}}]},"took":1,"timed_out":false,"_shards":{"total":2,"successful":2,"skipped":0,"failed":0},"hits":{"total":{"value":1,"relation":"eq"},"max_score":1.8856313,"hits":[{"_index":"ax-docs","_id":"9477b7c34a78919c41069a0fbc27e14f","_score":1.8856313,"_source":{"ax_path":"no-frontmatter.md"}}]}}
```

`hits.total.value` is **1**, and the one hit is `no-frontmatter.md`.

**Everything kept from `with-frontmatter.md`:**

```sh
curl -s localhost:9200/ax-yaml/_search -H 'Content-Type: application/json' -d '{"size":10,"query":{"match_all":{}}}'
```

```json
{"took":0,"timed_out":false,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0},"hits":{"total":{"value":2,"relation":"eq"},"max_score":1.0,"hits":[{"_index":"ax-yaml","_id":"4c93d5baae5775ac0c9d40d303a608db","_score":1.0,"_source":{"title":"Widget outage","date":"2026-01-15T00:00:00.000Z","tags":["widget","outage"],"ax_path":"with-frontmatter.md","ax_paths":["with-frontmatter.md"],"ax_file":"axf2-33b5a34efbbf1f94c425619c41474a56-b7","ax_locator":"d0","ax_dataset":"yaml","ax_run":"run-20260821T113344Z-d2bc-0000-g1","ax_format":"md"}},{"_index":"ax-yaml","_id":"ec3995a4b0daab55cd4d754228ebc770","_score":1.0,"_source":{"value":"The sprocket pool was exhausted.","ax_path":"with-frontmatter.md","ax_paths":["with-frontmatter.md"],"ax_file":"axf2-33b5a34efbbf1f94c425619c41474a56-b7","ax_locator":"d1","ax_dataset":"yaml","ax_run":"run-20260821T113344Z-d2bc-0000-g1","ax_format":"md"}}]}}
```

The frontmatter mapping, plus one paragraph as a bare `value` scalar. `# Outage postmortem`, `## Resolution` and `Raised the sprocket pool size from 4 to 16.` are gone.

**For contrast, everything kept from `no-frontmatter.md`** — the same body, whole, in `body`:

```sh
curl -s localhost:9200/ax-docs/_search -H 'Content-Type: application/json' -d '{"size":10,"query":{"match_all":{}}}'
```

```json
{"took":0,"timed_out":false,"_shards":{"total":1,"successful":1,"skipped":0,"failed":0},"hits":{"total":{"value":1,"relation":"eq"},"max_score":1.0,"hits":[{"_index":"ax-docs","_id":"9477b7c34a78919c41069a0fbc27e14f","_score":1.0,"_source":{"title":"# Outage postmortem","body":"# Outage postmortem\n\nThe sprocket pool was exhausted.\n\n## Resolution\nRaised the sprocket pool size from 4 to 16.","ax_path":"no-frontmatter.md","ax_paths":["no-frontmatter.md"],"ax_file":"axf2-01bca00d4d52a107c7a62a2330e92b5a-71","ax_locator":"s0","ax_dataset":"docs","ax_run":"run-20260821T113344Z-d2bc-0000-g1","ax_format":"md"}}]}}
```

## This repository's own SKILL.md files

Both `SKILL.md` files shipped here open with frontmatter, so both classify `yaml`. Indexing just those two, same procedure:

```
  tools/xerj-code/SKILL.md                                    format=yaml  status=indexed
  docs/case-studies/wordpress-security-audit/skill/SKILL.md   format=yaml  status=indexed

  on disk:           21799 bytes
  text indexed:        800 bytes   (3%)
```

## Why this is more severe than #379/#403

#403 was calibrated as lower severity than #379 because the failure surfaces as a PDF parse error rather than a silent junk reason. This one is quieter than either:

- the file is not junked — the catalog says `status: indexed`, and the run reports `0 junk/skipped files`;
- the collision surface is not a rare prefix but the standard opening of a whole documentation convention;
- what is lost is most of the document, while enough survives that the index looks populated.

Queries then return plausible, well-ranked results drawn from the surviving fraction. Measured on a 89-file markdown documentation corpus where every file carries frontmatter by convention: 37 of 89 files had their bodies dropped, and the run reported success. An agent trusting that index gets confident answers over half a corpus with no signal that the rest was never there — a worse outcome than an abort.

## Fix sketch (not attempted here)

Give the `---` branch the same treatment #403 gave `%PDF-`: a structural qualifier the alternative reading fails. A leading `---` is YAML unless a later line is exactly `---` and the text after it fails `yaml_line_ratio` — in which case the leading block was frontmatter and the document is `TxtProse`. Multi-document YAML is unaffected: the tail of a `---`-separated manifest is still YAML by that same ratio.

`sniff.rs` already pins the neighbouring invariant at `markdown_with_headings_is_prose()` (~line 3088); the missing case belongs beside it and fails on rc.23:

```rust
#[test]
fn markdown_with_yaml_frontmatter_is_prose() {
    let md = "---\ntitle: Checkout outage\ndate: 2026-06-14\n---\n\n\
              # Postmortem: checkout outage\n\n\
              ## Impact\n\
              Checkout was unavailable for 51 minutes.\n\n\
              ## Resolution\n\
              We reloaded the service and added an alert.\n";
    assert_eq!(kind(md), Family::TxtProse);
}
```

Whoever takes it should also check `yaml_line_ratio`'s own frontmatter exposure — a markdown file with no frontmatter but many `- ` bullets can still cross 0.6.

Say the word and I will implement the qualifier, watch that test fail before and pass after, and open a PR under `AI_CONTRIBUTIONS.md`. Nothing above was compiled, so it stays a sketch.

## Version note

The installer reports `v1.0.0-rc.23`, but the binary self-reports `v1.0.0-rc.18` on every platform: `engine/Cargo.toml` at tag `v1.0.0-rc.23` still carries `version = "1.0.0-rc.18"`. Separate defect, same family as #489/#515/#520 but in the binary rather than the site — not filed here, one issue per defect. Happy to file it.

---

**Verified:** every JSON block above is the verbatim response from the commands shown, on a freshly installed binary from `curl -fsSL https://xerj.org/get | sh`, clean `mktemp -d` data dir, macOS 15 (Darwin 25.6.0) aarch64, and reproduced identically on `debian:bookworm-slim`. The artifact is rc.23 code despite the version string — confirmed by the #403 qualifier being present (prose opening `%PDF- is the five byte header…` now sniffs `txt-prose`). The `sniff.rs` line range is unchanged between the rc.18 and rc.23 tags. The 3% figure is this repository's two `SKILL.md` files measured through a full index run.

**Assumed:** I did not trace which branch of the YAML extractor decides to keep only the first scalar, so "keeps the frontmatter mapping and the first paragraph" describes the observed output, not a read of that code path. I did not test non-markdown documents that legitimately open with `---`. The fix sketch and the test were not compiled.

xerj-org:
Closes #551.

## The bug (data loss on the primary use case)

`classify_text` (`sniff.rs`) routed every file whose first non-blank line is `---` to `Family::Yaml`:

```rust
if nonblank.first().map(|l| l.trim() == "---").unwrap_or(false)
    || yaml_line_ratio(nonblank) >= 0.6
{ return Family::Yaml; }
```

A leading `---` is also the opening delimiter of **markdown frontmatter** — the dominant convention for markdown corpora (Jekyll, Hugo, Obsidian, Docusaurus, MkDocs, Astro) and for Claude Code skill files, **including the two this repo ships**. The YAML extractor keeps the frontmatter mapping and the first paragraph and **discards the rest of the body**, while the run reports success (exit 3, `0 junk/skipped`). In #551's repro two byte-identical bodies classify differently purely by the presence of frontmatter — one `txt-prose`, one `yaml` — and a sentence present in both is retrievable from only one.

## The fix

A closing-`---` lookahead. If the file opens with `---`, has a matching closing `---`, and the body after it is majority non-YAML, classify by that body:

```rust
let leads_with_marker = nonblank.first().map(|l| l.trim() == "---").unwrap_or(false);
if leads_with_marker {
    if let Some(close_offset) = nonblank.iter().skip(1).position(|l| l.trim() == "---") {
        let body = &nonblank[close_offset + 2..];
        if body.len() >= 2 && yaml_like_count(body) * 2 < body.len() {
            return txt_kind(body);
        }
    }
}
if leads_with_marker || yaml_line_ratio(nonblank) >= 0.6 { return Family::Yaml; }
```

**Pure and multi-document YAML are unaffected**:
- A single leading `---` marker has no *closing* `---` → `position` is `None` → falls through to `Family::Yaml`.
- The body after a multi-document `---` separator is itself YAML-like → fails the majority-prose test → stays `Family::Yaml`.

Extracted `yaml_like_count` from `yaml_line_ratio` (behavior-preserving) so the lookahead and the ratio share the identical "YAML-like line" definition — and so the lookahead is not fooled by `yaml_line_ratio`'s `< 3`-line short-circuit on a short frontmatter body.

## Verification

- **Regression test** `markdown_frontmatter_is_classified_by_its_body_not_yaml`: the frontmatter file classifies **identically to its byte-identical frontmatter-less twin** (`TxtProse`), not YAML.
- **Guardrail** `yaml_documents_are_unaffected_by_the_frontmatter_lookahead`: single- and multi-document YAML stay `Yaml`.
- **Fail-before** (only the lookahead hunk reverted, `yaml_like_count` refactor kept): `markdown_frontmatter...` fails with `left: Yaml, right: TxtProse` — exactly the misclassification.
- `cargo test -p xerj-autoindex` (dev): **685 passed, 0 failed**.
- `cargo fmt -- --check`: clean. `cargo clippy -p xerj-autoindex --all-targets -- -D warnings` (rustc 1.98.0): exit 0.

Same *class* as #379/#380/#403 (an unqualified signature capturing prose) but a distinct code path, as #551 notes.

### DIFF
--- a/engine/crates/xerj-autoindex/src/sniff.rs
+++ b/engine/crates/xerj-autoindex/src/sniff.rs
@@ -1107,10 +1107,26 @@ fn classify_text(text: &str, nonblank: &[&str]) -> Family {
         return Family::Csv;
     }
 
-    // YAML
-    if nonblank.first().map(|l| l.trim() == "---").unwrap_or(false)
-        || yaml_line_ratio(nonblank) >= 0.6
-    {
+    // YAML — but a leading `---` is also the opening delimiter of markdown/prose
+    // frontmatter (Jekyll, Hugo, Obsidian, Docusaurus, MkDocs, Astro, and Claude
+    // Code skill files — including the two this repo ships). Without a lookahead
+    // for the *closing* `---` and what follows it, every such file was handed to
+    // the YAML extractor, which keeps the frontmatter mapping and the first
+    // paragraph and discards the rest of the body (#551, silent data loss). If
+    // the file opens with `---`, has a matching closing `---`, and the body after
+    // it is majority non-YAML, classify by that body so the whole document is
+    // indexed. A pure/multi-document YAML file keeps YAML: the content after its
+    // `---` document marker is still YAML-like (or too short to be called prose).
+    let leads_with_marker = nonblank.first().map(|l| l.trim() == "---").unwrap_or(false);
+    if leads_with_marker {
+        if let Some(close_offset) = nonblank.iter().skip(1).position(|l| l.trim() == "---") {
+            let body = &nonblank[close_offset + 2..];
+            if body.len() >= 2 && yaml_like_count(body) * 2 < body.len() {
+                return txt_kind(body);
+            }
+        }
+    }
+    if leads_with_marker || yaml_line_ratio(nonblank) >= 0.6 {
         return Family::Yaml;
     }
 
@@ -1127,8 +1143,16 @@ fn yaml_line_ratio(nonblank: &[&str]) -> f64 {
     if nonblank.len() < 3 {
         return 0.0;
     }
+    yaml_like_count(nonblank) as f64 / nonblank.len() as f64
+}
+
+/// Count of lines that read as YAML mapping/sequence entries. Shared by
+/// `yaml_line_ratio` and the frontmatter lookahead (#551) so both apply the
+/// identical definition of "YAML-like line" — unlike `yaml_line_ratio` this has
+/// no `< 3` short-circuit, so a short frontmatter body is judged on its content.
+fn yaml_like_count(lines: &[&str]) -> usize {
     let re = regex::Regex::new(r"^\s*(- )?[\w.@/-]+:(\s|$)").unwrap();
-    let hits = nonblank
+    lines
         .iter()
         .filter(|l| {
             let t = l.trim_start();
@@ -1141,8 +1165,7 @@ fn yaml_line_ratio(nonblank: &[&str]) -> f64 {
                 t.starts_with("- [ ] ") || t.starts_with("- [x] ") || t.starts_with("- [X] ");
             re.is_match(l) || (t.starts_with("- ") && !checkbox)
         })
-        .count();
-    hits as f64 / nonblank.len() as f64
+        .count()
 }
 
 fn txt_kind(nonblank: &[&str]) -> Family {
@@ -3104,6 +3127,69 @@ mod text_family_tests {
                   ## Pool exhaustion\n\
                   Symptoms are rising p99 and pool errors in the logs.\n";
         assert_eq!(kind(md), Family::TxtProse);
+    }
+
+    /// #551: a markdown/prose file that opens with a `---` frontmatter block was
+    /// sniffed as YAML — the YAML extractor keeps the frontmatter mapping and the
+    /// first paragraph and discards the body (silent data loss on the dominant
+    /// markdown convention: Jekyll/Hugo/Obsidian/Docusaurus/MkDocs/Astro, and
+    /// Claude Code skill files). With the closing-`---` lookahead the file is
+    /// classified by its body — identically to a frontmatter-less twin — so the
+    /// whole document is indexed.
+    #[test]
+    fn markdown_frontmatter_is_classified_by_its_body_not_yaml() {
+        let with_frontmatter = "---\n\
+                                title: Widget outage\n\
+                                date: 2026-01-15\n\
+                                tags: [widget, outage]\n\
+                                ---\n\n\
+                                # Outage postmortem\n\n\
+                                The sprocket pool was exhausted.\n\n\
+                                ## Resolution\n\
+                                Raised the sprocket pool size from 4 to 16.\n";
+        // Byte-identical body, no frontmatter.
+        let no_frontmatter = "# Outage postmortem\n\n\
+                              The sprocket pool was exhausted.\n\n\
+                              ## Resolution\n\
+                              Raised the sprocket pool size from 4 to 16.\n";
+        let twin = classify_full(no_frontmatter);
+        assert_ne!(
+            twin,
+            Family::Yaml,
+            "sanity: the body itself is prose, not YAML"
+        );
+        assert_eq!(
+            classify_full(with_frontmatter),
+            twin,
+            "markdown frontmatter must be classified by its body, identical to the \
+             frontmatter-less twin, not routed to the YAML extractor (#551)"
+        );
+    }
+
+    /// #551 guardrail: the frontmatter lookahead must not reclassify real YAML.
+    /// A single document (leading `---` marker, no prose body) and a
+    /// multi-document stream (the body after the first `---` separator is itself
+    /// YAML) both stay `Yaml`.
+    #[test]
+    fn yaml_documents_are_unaffected_by_the_frontmatter_lookahead() {
+        let pure = "---\n\
+                    name: widget-service\n\
+                    replicas: 3\n\
+                    ports:\n\
+                    \x20 - 8080\n\
+                    \x20 - 9090\n\
+                    env:\n\
+                    \x20 LOG_LEVEL: info\n";
+        assert_eq!(classify_full(pure), Family::Yaml);
+
+        let multi = "---\n\
+                     name: first\n\
+                     value: 1\n\
+                     ---\n\
+                     name: second\n\
+                     value: 2\n\
+                     kind: config\n";
+        assert_eq!(classify_full(multi), Family::Yaml);
     }
 
     /// The record-stream side must be unaffected — these are what TxtLines is for.


xerj-org:
## Independent 3-skeptic verification (adversarial, worktrees, CI toolchain)

Verdicts recorded verbatim. **3 NON_BLOCKING, 0 BLOCKING.** Two genuine should-fix edges surfaced → tracked as a follow-up; neither blocks (no data loss, #551 primary case works, all existing tests green).

> Refactor is provably clean … the extracted `yaml_like_count` closure is character-for-character identical to the old inlined body … `yaml_line_ratio` keeps its `<3` short-circuit. Index math is correct — no off-by-one: for `nonblank=[---, k:v, ---, body...]`, `body=&nonblank[close_offset+2..]` lands on the first body line.
> Two concrete defects, both real but non-merge-blocking:
> 1. `txt_kind(body)` bypasses the JSON/HTML/logs/CSV branches, so the "classified by its body — identically to the twin" claim breaks for non-prose bodies: frontmatter-over-CSV → `TxtProse` while bare twin → `Csv`; frontmatter-over-JSON → `TxtProse` while twin → `Json`. The test only passes because both its sides are prose. Overclaim, not a shipped-goal failure.
> 2. Regression for valid multi-document YAML: a fully parseable 2-doc stream whose second doc is block-scalar-prose-dominant now classifies `TxtProse` (was `Yaml`, with both docs extracted as records). Narrow — the guard requires the entire post-first-separator body majority non-YAML, which real k8s/Ansible multi-doc YAML dilutes back to `Yaml`. No data loss (text still indexed).

> `cargo +1.98.0 test -p xerj-autoindex` → `685 passed; 0 failed`. Every sniff test: `51 passed; 0 failed` — `plain_yaml_without_the_unity_tag_stays_yaml`, the Unity tests, and the markdown-prose tests all pass. No other classification path is touched (CSV/logs/xml/code are earlier returns, unchanged); new logic fully gated behind `leads_with_marker`. `yaml_like_count` extraction behavior-preserving. Multi-doc guardrail traced by hand → falls through to `Yaml`. Nit: `yaml_like_count` recompiles its regex per call (trivial perf). No existing-behavior defect; nothing blocks merge.

### 🟢 EVIDENCE — NON_BLOCKING
> Fail-before (reverted ONLY the lookahead hunk, kept `yaml_like_count`): `markdown_frontmatter_is_classified_by_its_body_not_yaml` FAILED — `left: Yaml, right: TxtProse`. Restored → passes. Genuinely load-bearing. Guardrail actually exercises the risky multi-doc body-check — verified by mutation: weakening the guard to `if body.len() >= 2 {` makes the `multi` case FAIL (`left: TxtProse, right: Yaml`), so the guard bites there. `yaml_like_count` behavior-preserving.
> Honesty gaps (NON_BLOCKING): the PR says "data loss / the YAML extractor discards the body," but every assertion is on `Family` classification — no test invokes the extractor. The twin-equality oracle is soft (both sides call `txt_kind` on the same slice); it does not pin `TxtProse`. Untested boundary: a frontmatter file with a single non-blank body line (`body.len() < 2`) still falls through to `Yaml` (data loss persists) — not a regression, but the guard's boundary is untested.

**Decision:** 0 BLOCKING → mergeable once all-green + CLEAN. The two edges (txt_kind bypasses full classification for non-prose bodies; prose-heavy multi-doc YAML routes to prose; the `body.len() < 2` boundary) are filed as a follow-up. The classification-level test is a correct proxy for the extractor behavior (classification gates the extractor); I'll tighten the PR's "data loss" framing to say so.

Issue #589, opened 2026-08-22
autoindex: sweep an excluded content group's documents from the destination (the purge half of #439)


xerj-org:
Follow-up to #439 / branch \`fix/439-exclusion-vs-deletion\` (the minimum fix). That change makes the graph resume-plan reconciler *distinguish* a file that vanished because an ignore/hidden rule widened (still on disk → \`excluded_content_groups\`) from a genuine deletion (absent from disk → \`deleted_content_groups\`), and gives the excluded case an accurate recovery message instead of the misleading "restore the removed file(s)". But it **still refuses** the excluded case — the documents stay live, so #439's headline data-exposure ("an exclusion rule that exists to keep data out of the index cannot remove data already in it") is not yet resolved.

## The remaining fix
Per #439's "shape of a fix": a group that vanished because it became excluded should be **swept from the destination and the run continues**, not refused — which also makes a widened exclusion actually remove the data it was widened to exclude, and unwedges the default (graph) path.

Concretely: route the `excluded_content_groups` to a delete of their published documents, aliases, and graph edges, then proceed with the rest of the reconcile. The delete machinery exists on the `--no-graph` generated path (`sync_executor::delete_group` → `delete_by_query`, sync_executor.rs:221), but the graph resume-plan path is a different execution model, so this needs a deliberate bridge (or a shared delete primitive) rather than a call-site tweak.

## Guardrails
- Only sweep groups proven excluded-because-on-disk-and-ignored, never a genuine deletion (that stays refused — deleting documents for a file the user removed by mistake is the thing #254/#439's refusal protects against).
- Respect the same "no remote mutation on a refused/dry-run attempt" contract; the sweep is a real mutation and must be journaled/committed like any other delete.
- Cover the other exclusion sources #439 flags as unverified: `.xerjignore` edits and `--no-default-ignores` changes, not only the hidden-name path.

Blocks #425 and any future change that narrows what `autoindex` walks (per #439).

xerj-org:
**Safety guardrail from the #590 verification** (carry into this sweep's design):

The message-only fix (#590) resolves `vanished_is_on_disk` with `root.join(rel).try_exists().unwrap_or(true)` — an unreadable/stat-error path is treated as **still on disk → excluded**. That is the *non-destructive* direction while the excluded case merely produces a clearer refusal: worst case it tells an operator a genuinely-deleted-but-now-unreadable path is "still on disk", which is harmless.

[...]

PR #590, opened 2026-08-22
fix(autoindex): distinguish an excluded-still-on-disk file from a deletion, incl. non-UTF-8 names (#439)

### LINKED ISSUE #439: widening an exclusion rule wedges the default (graph) autoindex path: the file is read as a user deletion, the run is refused, and the already-indexed documents stay live
Blocks #425, and will block any future change that narrows what `autoindex` walks.

## What happens

When a file stops being walked because an **exclusion rule widened** — not because the user deleted it — the resume-plan reconciler classifies it as a user deletion and refuses the run.

Reproduced with real binaries, one control, nothing changed on disk between steps:

```
1. index with the pre-fix binary, graph on (the DEFAULT)   exit=0
2. hidden secret is live                                    hits=1
3. upgrade: same folder, same state dir, newer binary       exit=1
4. the file the error says was removed is still on disk
5. the secret is still live after the failed upgrade        hits=1
```

Control: base binary → base binary on the identical corpus and state dir exits 0. Only the binary swap breaks it.

## Why it is worse than an inconvenience

**It is the default path.** `--no-graph` corpora reconcile a vanished file into a Delete. Graph-enabled corpora — the default, and what `xerj brain` uses (`crates/xerj-server/src/brain.rs:585` sets `no_graph: false`) — go through the resume plan, where `lib.rs:2491` refuses removal outright. The first post-upgrade run aborts with `exit=1`, writes nothing, and every rerun reproduces it identically.

**The documents stay live.** `lib.rs:2603` states it plainly: "those documents stay live and searchable with no source file behind them, and nothing in this pipeline removes them." So an exclusion rule that exists to keep data out of the index cannot remove data already in it, and the failed upgrade leaves exactly the exposed documents queryable.

**The recovery contract is unfollowable.** The machine-readable refusal (`schema: xerj.autoindex.unsupported_sync_delta.v1`) leads with:

```json
"restore_removed_files": "put the listed file(s) back and rerun"
```

The file was never removed — it is on disk, one `ls` away. An agent parsing this JSON takes route 1 first and loops. `docs/CONTRIBUTION_REVIEW.md:33` requires an error to "give an executable recovery route" and says one recommending a now-rejected flag "is not a recovery contract"; `--fresh` is refused here too, and `--no-graph` on the wedged state dir fails with a different error about a legacy nonempty plan.

The surviving routes were verified: deleting the published indices plus the state dir and re-extracting the whole corpus works but is expensive; an isolated rebuild with a new `--state-dir`/`--prefix` works but leaves the original target's documents live until cleaned by hand.

## Shape of a fix

The reconciler is asking the wrong question. It compares "was in the plan" against "is in the walk" and infers deletion, when there are two distinct causes:

- **the user removed the file** — current behaviour is right, refuse
- **the walker stopped yielding it because a rule widened** — should sweep the group from the destination and continue

So: give the walk a way to report *why* a path is absent, and let `UnsupportedInventoryDelta::between` distinguish excluded-now from deleted. A group that vanished because it became hidden should be purged, not refused — which also makes the exclusion rule actually remove the data it was widened to exclude.

Failing that, the minimum is a distinct refusal that names the real cause and points at the route that works, instead of leading with an instruction to restore a file that was never removed.

**Verified:** the repro, the control, the surviving recovery routes, and the JSON refusal shape were all measured by an independent verification of #425 on isolated binaries and ports.
**Not verified:** whether the same wedge occurs for the other exclusion rules (`.xerjignore` edits, `--no-default-ignores` changes) — it should by the same mechanism, but only the hidden-name path was exercised.

xerj-org:
Addresses **#439**'s recovery-contract half (the misleading refusal). Does **not** close #439: the in-place SWEEP that actually removes the excluded data is the larger half, tracked in **#589**.

## The bug (data-exposure + wedge on the DEFAULT graph path)

When a file stops being walked because an ignore/hidden rule **widened** — not a user deletion — the resume-plan reconciler classified it as a deletion, refused the rerun (exit 1), left the documents live, and the machine-readable recovery (`xerj.autoindex.unsupported_sync_delta.v1`) led with `restore_removed_files: "put the listed file(s) back and rerun"`. The file is one `ls` away, so an agent takes that route first and loops. The CHANGELOG's canonical trigger is a **hidden non-UTF-8 name**.

## The fix

`UnsupportedInventoryDelta::between` now takes the corpus `root` and splits vanished plan files by whether they still exist on disk:
- **absent from disk** → `deleted_content_groups` (unchanged message/behaviour)
- **still on disk** → `excluded_content_groups` (#439): the message says the file is still on disk, must not be restored, and names the real routes (narrow the rule to re-admit it, or rebuild isolated to drop it). `restore_removed_files` is emitted only when something was actually deleted.

The disk probe goes through the file's **reversible raw-bytes `path_id`** (`unix:<hex>` / `windows:<hex>`), not its `rel`:

```rust
fn vanished_is_on_disk(root: &Path, assignment: &FileAssignment) -> bool {
    let rel = decode_stable_path_id(&assignment.path_id)
        .unwrap_or_else(|| PathBuf::from(&assignment.rel)); // legacy fallback
    root.join(rel).try_exists().unwrap_or(true)
}
```

`rel` is `to_string_lossy`, so a non-UTF-8 name is stored with `U+FFFD` and matches no dirent — probing through it would call every such file a phantom deletion (this was the gap an earlier draft had). `decode_stable_path_id` reverses `walk::stable_path_id`.

`refuses()` is **semantically unchanged** — `deleted ∪ excluded` equals the old `vanished` set, so no rerun that used to refuse now proceeds. The v1 `vanished_content_groups` JSON field is retained (union) for back-compat; the split fields are additive.

**Scope (honest):** this fixes the *message/contract*. The documents **still stay live** and the run is **still refused** — sweeping an excluded group from the destination (so a widened exclusion removes the data, and the graph path stops wedging) is the larger half, tracked in **#589**.

## Verification

- **Tests**: `an_excluded_still_on_disk_group_is_not_reported_as_a_deletion` (UTF-8) and `a_non_utf8_excluded_name_is_matched_on_disk_via_path_id` — the headline case: a hidden non-UTF-8 file classifies excluded via `path_id`, and the test proves the lossy `rel` alone would miss it.
- **Fail-before**: reverting the disk probe to rel-only misclassifies the non-UTF-8 file as a deletion (`deleted: [.secret_��.csv]`).
- `cargo test -p xerj-autoindex` (dev): **702 passed, 0 failed**.
- `cargo fmt -- --check`: clean. `cargo clippy -p xerj-autoindex --all-targets -- -D warnings` (1.98.0): exit 0.
- `cargo test -p xerj-autoindex --profile ci-test` (1.98.0): delta 8/0, failure-resume 46/0.
- CHANGELOG: added a Fixed entry; corrected the pre-fix upgrade note to reflect the accurate refusal + point at #589.

Refs #439, #589. An independent 3-skeptic verification follows in a comment.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,22 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A file excluded by a widened ignore/hidden rule was reported as a user
+  deletion, so the graph resume path refused the rerun and its recovery
+  contract told operators to "restore the removed file(s)" — a file still on
+  disk** ([#439](https://github.com/xerj-org/xerj/issues/439), first fix). The
+  resume-plan reconciler (`UnsupportedInventoryDelta::between`) now splits a plan
+  file that vanished from the walk by whether it is still on disk, probed through
+  the reversible raw-bytes `path_id` so a hidden *non-UTF-8* name (the case #439
+  was filed for) is matched instead of read as a phantom removal through its
+  lossy `rel`. A genuine deletion keeps the "restore the removed file(s)" route;
+  a still-on-disk exclusion now says the file is still on disk and names the real
+  routes (narrow the rule to re-admit it, or rebuild isolated to drop it). The
+  refusal itself is unchanged — the documents still stay live — because sweeping
+  an excluded group from the destination (so a widened exclusion actually removes
+  the data) is a larger change tracked in
+  [#589](https://github.com/xerj-org/xerj/issues/589).
+
 - **`sort` on a field XERJ cannot resolve returned `null` on every hit
   instead of an error** ([#437](https://github.com/xerj-org/xerj/issues/437)).
   `compute_sort_values` special-cased `_score`, `_doc`, `_id` and the four
@@ -580,16 +596,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   **Upgrade.** A corpus indexed by a pre-fix binary that contains a
   hidden non-UTF-8 name looks like a file removal to the graph-enabled
-  path (the default, and what `xerj brain` uses). Until
-  [#439](https://github.com/xerj-org/xerj/issues/439) lands, the first
-  rerun exits 1 with `unsupported_content_group_removal`, writes
-  nothing, and the already-indexed documents stay live. Do not follow
-  the refusal's first recovery route ("restore the removed file(s)"):
-  the file is still on disk, just no longer walked. Recovery that
-  works: delete the published indices and the state directory, then
-  re-index the folder. An isolated rebuild with a new `--state-dir` /
-  `--prefix` also exits 0, but leaves the old target's documents live
-  until those indices are deleted by hand.
+  path (the default, and what `xerj brain` uses). As of
+  [#439](https://github.com/xerj-org/xerj/issues/439)'s first fix the
+  rerun's refusal names the real cause — the file is still on disk,
+  excluded, not removed — instead of the misleading "restore the removed
+  file(s)". It still exits 1 with `unsupported_content_group_removal`,
+  writes nothing, and the already-indexed documents still stay live;
+  sweeping them (so a widened exclusion removes the data) is tracked in
+  [#589](https://github.com/xerj-org/xerj/issues/589). Recovery that
+  removes the data today: delete the published indices and the state
+  directory, then re-index the folder. An isolated rebuild with a new
+  `--state-dir` / `--prefix` also exits 0, but leaves the old target's
+  documents live until those indices are deleted by hand.
 
 - **Autoindex snapshot and replay one-shot failpoints can no longer be stolen by unrelated parallel tests.** ([#385](https://github.com/xerj-org/xerj/issues/385)).
   The two test-only failpoints are now owned by the thread that arms them, while
--- a/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
@@ -629,7 +629,7 @@ fn assert_unsupported_delta_without_remote_mutation(
         "{message}"
     );
     assert!(
-        message.contains("restore the removed file(s) and rerun"),
+        message.contains("restore the DELETED file(s) and rerun"),
         "{message}"
     );
     assert!(
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -2389,7 +2389,17 @@ struct InventoryDeltaEntry {
 #[derive(Debug, Clone, PartialEq, Eq, Serialize)]
 struct UnsupportedInventoryDelta {
     added_content_groups: Vec<InventoryDeltaEntry>,
-    vanished_content_groups: Vec<InventoryDeltaEntry>,
+    /// Vanished from the walk AND absent from disk: a genuine user deletion.
+    /// Refused, because the published documents stay live with no source file
+    /// behind them and nothing here removes them.
+    deleted_content_groups: Vec<InventoryDeltaEntry>,
+    /// Vanished from the walk but STILL ON DISK: the file was excluded by a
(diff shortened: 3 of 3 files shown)


xerj-org:
## Independent 3-skeptic verification (v2 — with the `path_id` non-UTF-8 handling)

Verdicts recorded verbatim. **3 NON_BLOCKING, 0 BLOCKING.** Two of the flagged nits were cheap and worth hardening (a real panic contradicting the fn's own doc, and the v1 ordering) — applied in `c08fdb74` on top of the verified commit, with a new test; the rest are follow-ups (several land on #589).

> EXACT-INVERSE holds. Probe reproducing the unix arm against `walk::stable_path_id` (2-hex/byte lowercase, `OsStr::from_bytes`) round-trips byte-for-byte for non-UTF-8 (`.secret_\xff\xfe.csv`), embedded backslash+NUL, plain, and empty names. path_id is built from the root-relative path, so `root.join(decoded)` targets the right file.
> refuses() equivalence is provably correct: `between` pushes each vanished entry to exactly one of deleted/excluded, so deleted ⊎ excluded == old vanished; refusal decision is byte-identical to pre-split.
> Windows arm can't build here (no windows std) but `OsString::from_wide` + 4-hex/u16 is the correct symmetric inverse of `encode_wide` and will compile.
> Weakest defect (NON_BLOCKING): `decode_stable_path_id("unix:aéb")` PANICS on a non-char-boundary slice, contradicting its own doc — but path_id is always machine-generated ASCII hex, so unreachable except via a corrupted/hand-edited plan JSON. Nits: v1 union no longer globally sorted; `try_exists().unwrap_or(true)` mislabels a deleted file under an unreadable dir as excluded.

> Full crate suite green under rustc 1.98.0: lib `688 passed; 0 failed` + all integration bins. No pre-existing refusal/http test flipped. `refuses()` provably equivalent to `!vanished.is_empty()`. Synthetic-`id:`-path_id fallback exercised and correct (decodes to None → falls back to rel). Both prod call sites pass `&cfg.root` = the walk root, so decoded-rel+root round-trips. Back-compat: schema string unchanged; the only out-of-crate consumers (`failure_resume_http_tests.rs`) pass. CHANGELOG top entry matches the code.
> Non-blocking nits: v1 union `deleted.chain(excluded)` not globally sorted; `try_exists().unwrap_or(true)` classifies stat-error files as excluded — harmless here (both refuse) but worth flagging for the #589 sweep, where "excluded" becomes destructive.

### 🟢 EVIDENCE — NON_BLOCKING
> Fail-before reproduced: reverting `vanished_is_on_disk` to probe `rel` only → `a_non_utf8_excluded_name_is_matched_on_disk_via_path_id` FAILS (`deleted: [.secret_\u{fffd}\u{fffd}.csv]`); restoring the `decode_stable_path_id` probe → passes. Non-UTF-8 test is genuine (real non-UTF-8 file, real `unix:<hex>` path_id, in-test sanity assert proves the lossy rel does not resolve).
> Honesty checks pass: no closing keyword for #439 (count 0); commit + CHANGELOG label this "first fix", state documents still stay live, defer the sweep to #589. No overclaim of full resolution.
> Non-blocking nits: `between` now does one `try_exists` stat per vanished file (perf, read-only); `unwrap_or(true)` message edge.

### Hardening applied post-verification (`c08fdb74`)
- **Panic → None**: `decode_stable_path_id` now uses `hex.get(i..i+2)` (not `hex[..]`) so a non-ASCII byte in a corrupted path_id returns `None` per its doc, instead of panicking. New test `decode_stable_path_id_rejects_malformed_ids_without_panicking` (odd-length, non-hex, and the `unix:aéb` multibyte case).
- **v1 ordering**: the `vanished_content_groups` union is now globally `(path, file_key)`-sorted, matching the pre-split field for any v1 consumer.
- Re-verified: dev **703/0**; ci-test delta 9/0 + failure-resume 46/0 (1.98.0); fmt clean; clippy `-D warnings` 0.

Follow-ups filed/relevant: **#589** (the in-place sweep — where the `unwrap_or(true)` "excluded" default must become non-destructive). The remaining nits (stat-per-vanished perf on the 82k bind-mount case) are acceptable for a read-only refusal path.

PR #592, opened 2026-08-22
fix(engine): deterministically await the stale HNSW rebuild in its test, killing the flake (#536)

### LINKED ISSUE #536: Flaky: test_hnsw_stale_snapshot_rebuilds_on_open fails under parallel test load
Observed while verifying #415 under rustc 1.97.1 (`cargo test -p xerj-engine --profile ci-test`).

- Full-suite run: `test_hnsw_stale_snapshot_rebuilds_on_open` **FAILED** (124 passed, 1 failed).
- Same commit, run in isolation: **passes 2/2** (`--test integration test_hnsw_stale_snapshot_rebuilds_on_open`).
- Full-suite re-run: **all green**, 0 failures.

So it is non-deterministic under parallel load — likely a shared-fixture / mtime / snapshot-timing isolation issue in the HNSW stale-snapshot rebuild path, not a product bug. Unrelated to the change under test (a `_source` stored-codec fix). Worth pinning the flake source so CI does not intermittently red-flag unrelated PRs.

xerj-org:
Closes #536.

## The flake
`test_hnsw_stale_snapshot_rebuilds_on_open` forges a stale `ids.json` stamp, reopens, and polls `hnsw_stats()` on a fixed **100×100ms = 10s** budget for the background rebuild to converge. Under parallel `cargo test` load the tokio rebuild task is CPU-starved and 10s is not always enough → the poll times out and the test intermittently fails, while passing 2/2 in isolation. It red-flags unrelated PRs (observed while verifying #415).

Each test uses its own `TempDir`, so this is **not** a shared-fixture issue — it is a timeout.

## The fix (deterministic, not a wider timeout)
`spawn_hnsw_rebuild_if_stale` stores the rebuild's `JoinHandle` (`hnsw_rebuild_task`), and the spawned task clears `hnsw_rebuilding` and publishes the healed graph **before** its handle resolves. New `Index::await_hnsw_rebuild` takes and awaits that handle; the test awaits it instead of polling, observing the settled stats deterministically — no timeout to flake, at any load. A no-op when no rebuild was spawned.

## Verification
- **Fail-before** (await removed, stats read immediately): fails with `stale:true, rebuilding:true` — the rebuild has not converged, exactly the race the poll tolerated probabilistically.
- **Pass-after**: the test completes in ~0.3s deterministically.
- Full `cargo test -p xerj-engine`: **0 failed**. `cargo fmt --check`: clean. `cargo clippy -p xerj-engine --all-targets -- -D warnings` (1.98.0): exit 0.

An independent 3-skeptic verification follows in a comment.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -10166,6 +10166,23 @@ impl Index {
         }
     }
 
+    /// Await the in-flight stale HNSW rebuild to completion, if one is running.
+    ///
+    /// A deterministic alternative to polling `hnsw_stats` for tests and
+    /// diagnostics (#536): the background task clears `hnsw_rebuilding` and
+    /// publishes the healed graph *before* its `JoinHandle` resolves, so a
+    /// caller observes `stale == false` / `rebuilding == false` immediately
+    /// after this returns — with no timeout that can flake under parallel test
+    /// load. A no-op when no rebuild was spawned (the snapshot was already
+    /// fresh). The handle is `take`n, so a subsequent close/abort is a no-op
+    /// (the task has already finished).
+    pub async fn await_hnsw_rebuild(&self) {
+        let handle = self.hnsw_rebuild_task.lock().take();
+        if let Some(handle) = handle {
+            let _ = handle.await;
+        }
+    }
+
     /// Re-derive the HNSW graph from the authoritative doc set and clear
     /// `hnsw_stale` (RC4 W2 item 17). Returns `Ok(true)` once converged.
     ///
--- a/engine/crates/xerj-engine/tests/integration.rs
+++ b/engine/crates/xerj-engine/tests/integration.rs
@@ -5153,22 +5153,26 @@ async fn test_hnsw_stale_snapshot_rebuilds_on_open() {
     // converge, clear the flag, and leave every doc graphed.
     let engine = make_engine(&dir);
     let idx = engine.get_index("vecs").unwrap();
-    let mut healed = false;
-    let mut last = json!(null);
-    for _ in 0..100 {
-        last = idx.hnsw_stats().await;
-        if last["present"] == json!(true)
-            && last["stale"] == json!(false)
-            && last["rebuilding"] == json!(false)
-        {
-            healed = true;
-            break;
-        }
-        tokio::time::sleep(std::time::Duration::from_millis(100)).await;
-    }
-    assert!(
-        healed,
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification — all 3 NON_BLOCKING

Verdicts recorded verbatim. Gate (b) satisfied. (The current red **Landing constants guard** is the tag-ahead transient — this branch predates the rc.31 stamp-bump #591; it clears on rebase once #591 merges, and is unrelated to the code.)

> No store-race: `spawn_hnsw_rebuild_if_stale()` is called synchronously at index.rs:6519 before `Ok(index)` and stores the handle (`slot.replace`, 10163) after `tokio::spawn` — so the handle is present by the time `get_index` returns. Non-vacuous: the forged `seq_no=999999` makes `hnsw_stale=true`, so the rebuild actually spawns and `take()` returns `Some`. Publish-before-resolve is correctly ordered: the task clears `hnsw_stale` (10385) and `hnsw_rebuilding` (10144-10145) *before* the closure returns, so post-await the reader sees `stale==false, rebuilding==false, doc_coverage==6`. No deadlock (guard dropped before `.await`). `take()` doesn't corrupt the abort path (9713's take is a benign `None` after the task finished; prod never calls the method). Confirmed by run: `... ok (0.38s)`.

> `cargo +1.98.0 test -p xerj-engine`: every binary green, **0 failed** (lib 610, integration 125 incl. the target, +~40 bins). Target deterministic: 3/3 runs (0.37/0.36/0.32s). Only prod change is `index.rs | 17 +++` — pure addition, `close()`/`spawn_hnsw_rebuild_if_stale` byte-identical to main. **`pub` is required, not a smell**: the sole caller is the external integration-test crate compiled without `--cfg test`, so `pub(crate)`/`#[cfg(test)]` would break the build. Clippy clean.

### 🟢 EVIDENCE — NON_BLOCKING
> Fail-before deterministic 3/3: removing `await_hnsw_rebuild().await` → fails every time at integration.rs:5167 with `"stale":true,"rebuilding":true` (~0.4s). Restored → passes. Mechanism verified against source (synchronous store, publish-before-resolve, no lock across await). No production regression (only adds the method; sole caller is the test). **"Closes #536" justified** — the issue scopes the flake to this single test, and grep confirms it was the only rebuild-convergence poll loop in the suite. Nit: PR's "~0.3s" is in-suite; standalone ~1.1s incl. binary startup.

**Verification tally: 3 NON_BLOCKING, 0 BLOCKING.** Merge gate (b) ✅. Holding on (a) all-green + (c) CLEAN — clears when this branch rebases onto rc.31 main (after #591 lands).

PR #593, opened 2026-08-22
fix(engine): hybrid/fusion reports a capped total as relation:gte, not a false eq (#569)

### LINKED ISSUE #569: hybrid/fusion queries report a fusion-depth-capped hits.total as relation:eq (false exact count)
Surfaced by the #458 verification (knn-beside-query → hybrid). NON_BLOCKING, pre-existing to the hybrid/fusion executor — filing so #458 can merge with it tracked.

## The gap
`run_hybrid_with_deadline` fetches each sub-list at `per_query_topk = max(50, size+from)` (index.rs:~11920), RRF-fuses, then sets `total_value = fused.len()` with `relation: Eq` unless it timed out (index.rs:~11990-12006). `request.track_total_hits` is forwarded to the sub-queries but their totals are **discarded** in favour of `fused.len()`.

So a hybrid (or `hybrid`/RRF) query over a corpus with thousands of lexical matches, `size:10`, reports e.g. `{"total":{"value":<=100,"relation":"eq"}}` — a **false exact count** that is unfixable via `track_total_hits`. A plain lexical query reports the accurate match count.

This affects every `hybrid` query, and — once #458 lands — the ES 8.x `knn`-beside-`query` wire shape when it routes to hybrid (the hybrid-safe case: no aggs/sort/collapse/search_after).

## Fix sketch
Either (a) report `relation: "gte"` when the fused count is capped by `per_query_topk` (honest lower bound, ES-style), or (b) when `track_total_hits` requests an exact/large count, tally each sub-query's true total (union-aware) rather than `fused.len()`. (a) is the minimal wire-honest fix.

Sibling: the post-hoc `min_score` on a hybrid request compares against RRF micro-scores (~1/61), so a BM25-tuned threshold drops every hit (index.rs:~14009 RRF formula, applied at es_compat.rs:~10701) — worth handling in the same pass (or documenting).

_Filed from the #458 3-skeptic verification (correctness lens, NON_BLOCKING findings)._

xerj-org:
Closes #569.

## The bug (wrong results on a headline BM25+kNN feature)
`run_hybrid_with_deadline` fetches each sub-list at `per_query_topk = max(50, size+from)`, RRF/linear-fuses them, and reported `hits.total = { value: fused.len(), relation: Eq }` unless the request timed out. When a sub-list is **truncated** at the fetch cap the fused union under-counts the true match set, so a hybrid query over thousands of lexical matches with `size:10` claimed `{"total":{"value":<=100,"relation":"eq"}}` — a **false exact count** that `track_total_hits` cannot correct, while a plain lexical query reports the real count. Affects every `hybrid` query and (since #458) the ES 8.x `knn`-beside-`query` shape that routes to hybrid.

## The fix (ES-style, minimal — issue's option (a))
Track `fused_count_capped` across the sub-query loop — a sub-list is truncated if it filled the `per_query_topk` page, or its own total exceeds the page, or its own count is already a lower bound (`relation == Gte`). Report `Gte` when `any_timed_out || fused_count_capped`, and `Eq` only when every sub-list was fully retrieved. The reported `value` (`fused.len()`) is a valid **lower bound** in the capped case — honest, not the exact tally (option (b), a union-aware exact count, is deliberately not attempted here).

```rust
if sub_result.hits.len() >= per_query_topk
    || sub_result.total.value > per_query_topk as u64
    || sub_result.total.relation == TotalHitsRelation::Gte
{ fused_count_capped = true; }
...
relation: if any_timed_out || fused_count_capped { Gte } else { Eq },
```

## Verification
- **Tests**: `hybrid_reports_gte_when_a_sub_list_is_capped` (256 vectors, a knn sub-list capped at 50, **no timeout** → `Gte`) and `hybrid_reports_eq_when_no_sub_list_is_capped` (6 docs, fully fetched → `Eq`, value 6).
- **Fail-before** (reverting only `|| fused_count_capped`): the capped hybrid reports a false `Eq`.
- `cargo test -p xerj-engine`: **0 failed** (full suite, exit 0). `cargo fmt --check`: clean. `cargo clippy -p xerj-engine --all-targets -- -D warnings` (1.98.0): exit 0.

The sibling `min_score`-vs-RRF-micro-scores note in #569 is separate and left tracked there. An independent 3-skeptic verification follows in a comment.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -4653,6 +4653,108 @@ mod semantic_deadline_regression_tests {
         );
     }
 
+    /// #569: when a hybrid sub-list is truncated at `per_query_topk`, the fused
+    /// `fused.len()` under-counts the true match set, so `hits.total.relation`
+    /// must be `Gte` (honest lower bound) — NOT a false `Eq`. This must hold
+    /// without any timeout (the pre-fix code only set `Gte` on timeout).
+    #[tokio::test]
+    async fn hybrid_reports_gte_when_a_sub_list_is_capped() {
+        let dir = TempDir::new().unwrap();
+        let engine = engine(&dir);
+        engine
+            .create_index("hybrid-capped", Schema::empty())
+            .unwrap();
+        let idx = engine.get_index("hybrid-capped").unwrap();
+        seed_vectors(&idx).await; // 256 docs, identical [1.0, 0.0] vectors
+        idx.test_scan_checkpoint_delay_ms
+            .store(0, Ordering::Relaxed);
+        // k >= per_query_topk(50) over 256 vectors → the sub-list fills a full
+        // 50-doc page → capped → the fused union under-counts the 256 matches.
+        let knn = QueryNode::Knn {
+            field: "embedding".into(),
+            vector: vec![1.0, 0.0],
+            k: 256,
+            num_candidates: None,
+            filter: None,
+            boost: None,
+            similarity: None,
+        };
+        let far = std::time::Instant::now() + std::time::Duration::from_secs(30);
+        let result = idx
+            .run_hybrid_with_deadline(
+                &match_all(30_000), // default size 10 → per_query_topk = 50
+                far,
+                vec![xerj_query::ast::WeightedQuery {
+                    query: knn,
+                    weight: 1.0,
+                }],
+                xerj_query::ast::FusionStrategy::Rrf { k: 60 },
+            )
+            .await
+            .unwrap();
+        assert!(
+            !result.timed_out,
+            "the Gte must come from capping, not a timeout"
+        );
+        assert_eq!(
+            result.total.relation,
+            TotalHitsRelation::Gte,
+            "a capped hybrid sub-list must report a lower-bound total: {:?}",
+            result.total
+        );
+    }
+
+    /// #569 guardrail: when every sub-list is fully retrieved (fewer matches
+    /// than `per_query_topk`), the fused total is exact → `Eq`, not a needless
+    /// `Gte`.
+    #[tokio::test]
+    async fn hybrid_reports_eq_when_no_sub_list_is_capped() {
+        let dir = TempDir::new().unwrap();
+        let engine = engine(&dir);
+        engine
+            .create_index("hybrid-exact", Schema::empty())
+            .unwrap();
+        let idx = engine.get_index("hybrid-exact").unwrap();
+        for n in 0..6 {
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent 3-skeptic verification — all 3 NON_BLOCKING

Verdicts verbatim. Gate (b) satisfied. The single finding across all three lenses is the safe-direction exact-boundary over-report, filed as follow-up **#594**.

> Honest lower bound confirmed: `fuse_rrf` dedups by `h.id` into a `HashMap`, so `fused.len() ≤ |true hybrid union|` — fusion can never EXCEED the true count, so `value` + `Gte` is honest (the dedup-vs-union overcount theory is refuted). Normal truncation forces `hits.len() == per_query_topk` → condition 1 fires (and the materialisation cap `(from+size+100).max(256)` never under-delivers below the page). `track_total_hits` forwarding is safe: a capped sub-total returns `Gte`/over-limit, caught by conditions 2/3, so it never masquerades as Eq. Only residual: false Gte at the exact boundary (safe/conservative, not a correctness defect). A kNN leg with user `k < per_query_topk` reporting Eq at k matches ES top-k semantics — not the #569 bug.

> `cargo +1.98.0 test -p xerj-engine` → CARGO_EXIT=0, **1035 passed / 0 failed** across all 49 test binaries. Blast radius hybrid-only: `git diff --name-only` = single file (+119/-1), confined to `run_hybrid_with_deadline`. No pre-existing test flips Eq→Gte (the one hybrid-relation test asserts Gte via timeout with knn k=10 < per_query_topk=50, unaffected; es_compat's hybrid test asserts only HTTP 400). Both new tests pass. Only imperfection: the boundary over-report — standard ES/Lucene filled-collector convention, strictly safer than the pre-fix false Eq. Not a regression.

### 🟢 EVIDENCE — NON_BLOCKING
> Fail-before reconstructed: reverting only `|| fused_count_capped` → `hybrid_reports_gte_when_a_sub_list_is_capped` FAILS with `left: Eq, right: Gte`, reporting `TotalHits { value: 50, relation: Eq }` — a false-exact of 50 when 256 match. Load-bearing. Tests genuine (256 real docs; capped test asserts `!timed_out` so Gte comes from capping; exact test asserts Eq + value 6). Semantics honest — Gte-on-cap mirrors ES 8.x capping; `value` stays a valid lower bound, never a claimed exact; commit message makes no overclaim.

**Tally: 3 NON_BLOCKING, 0 BLOCKING.** Merge gate (b) ✅. Holding on (a) all-green + (c) CLEAN. Boundary nit → #594.

PR #596, opened 2026-08-22
fix(api): order _eql/search events by @timestamp instant across alias members before truncating (#567)

### LINKED ISSUE #567: _eql/search: cross-member event merge is not @timestamp-ordered before size-truncation
Follow-up from the #561 verification (both the correctness and regression skeptics flagged it, NON_BLOCKING). #561 fixed the alias fan-out (and, on rework, the filtered-alias leak); this is the remaining EQL-ordering refinement it explicitly scoped out.

### The gap
`eql_search` fans out to every alias member, concatenates each member's hits **in member iteration order** (member A's events, then member B's), then `events.truncate(size)` (`es_compat.rs`, the `_eql` arm). `eql_to_query` emits a query with **no sort**. So when the summed match count exceeds `size`, the events that survive are biased toward the earlier-iterated members rather than the EQL-earliest by `@timestamp`.

Elasticsearch EQL orders single-event results by `@timestamp` ascending. XERJ never implemented that ordering (it was also absent pre-#561 for the single member searched), so this is a **pre-existing ordering gap the fan-out now inherits across members**, not a regression introduced by #561.

### Why it's not blocking #561
- No test regresses; the fan-out and filter-honouring behaviour are correct.
- `total` is the true summed match count; only the returned *page* ordering is affected, and only when `total > size`.

### Fix sketch
Give the per-member search an explicit `@timestamp` ascending sort (or collect `(@timestamp, event)` and sort the merged vec) **before** `events.truncate(size)`, so the page is the globally-earliest `size` events. Needs a test with two members whose matches interleave in time and a `size` below the combined count.

_Filed from the #561 3-skeptic verification (correctness + regression lenses)._

xerj-org:
Closes #567.

## The bug (wrong results, SIEM)
`eql_search` fans out to every alias member, concatenates each member's hits in member-iteration order, then `events.truncate(size)` (`eql_to_query` emits no sort). When the summed match count exceeds `size`, the surviving page is biased toward the earlier-iterated members instead of the EQL-earliest by `@timestamp` — the #451/#561 fan-out inherited this across members.

## The fix
ES EQL orders single-event results by the timestamp field ascending. Parse the optional `timestamp_field` (default `@timestamp`), sort each member ascending by it (`unmapped_type: "date"` so a member index **without** the field is not a 400 — XERJ EQL, unlike ES, doesn't require it, so it degrades to unsorted), and sort the MERGED events by that field **before** truncate → the returned page is the globally-earliest `size` events.

`eql_timestamp_cmp` compares epoch numbers numerically and parses ISO-8601/RFC3339 strings to a **real instant** (`chrono::DateTime::parse_from_rfc3339`), not byte-wise — a raw string compare misorders mixed fractional precision (`…01.500Z` before `…01Z`, since `.`<`Z`) and non-UTC offsets (`…+05:00` is an earlier instant than the lexically-smaller `…Z`), both the norm in log/SIEM corpora. Unparseable → deterministic byte fallback; missing → last; `sort_by` stable. `total` (true summed count) unchanged.

## Verification
- Tests: `eql_events_are_timestamp_ordered_across_members_before_truncation` (interleaving seconds, size 3/6 → `[01,02,03]` spanning both members) and `eql_events_order_by_instant_not_lexical_string` (mixed fractional + `+05:00` offset → true-instant order).
- Fail-before: removing the merged sort → `[01,03,05]`; a lexical comparator → `[01.500Z, 01Z, …, +05:00]`.
- The two pre-existing eql tests (no @timestamp mapping) exercise the `unmapped_type` degradation and pass.
- `cargo test -p xerj-api`: **485 passed, 0 failed**. fmt clean; clippy `-D warnings` 0.

**Independent 3-skeptic verification found a real BLOCKING wrong-results defect** in the first draft (the comparator compared ISO strings lexically, misordering millis-precision + multi-timezone corpora — a case my own first test missed). It was fixed (chrono instant parse) + a regression test added, and re-verification returned **3/3 NON_BLOCKING**. Both rounds' verdicts follow in a comment.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -33188,6 +33188,42 @@ fn glob_match(pattern: &str, name: &str) -> bool {
 // POST /{index}/_eql/search — EQL search stub
 // ─────────────────────────────────────────────────────────────────────────────
 
+/// Compare two EQL timestamp-field values (pulled from each event's `_source`)
+/// for the ascending merge sort (#567). Epoch numbers compare numerically (one
+/// unit per corpus). Strings are parsed as RFC3339 / ISO-8601 and compared as
+/// real instants — NOT byte-wise: a raw string compare misorders mixed
+/// fractional-second precision (`…:01.500Z` vs `…:01Z`, since `.` < `Z`) and
+/// non-UTC offsets (`…+05:00` is an earlier instant than the lexically-smaller
+/// `…Z`), which are the norm in SIEM/log corpora. `chrono::DateTime` orders by
+/// the UTC instant, so both are handled. An unparseable value falls back to a
+/// deterministic byte order; a missing field sorts last so a malformed doc never
+/// displaces a real earliest hit.
+fn eql_timestamp_cmp(a: Option<&Value>, b: Option<&Value>) -> std::cmp::Ordering {
+    use std::cmp::Ordering::{Equal, Greater, Less};
+    match (a, b) {
+        (Some(x), Some(y)) => {
+            if let (Some(nx), Some(ny)) = (x.as_f64(), y.as_f64()) {
+                nx.partial_cmp(&ny).unwrap_or(Equal)
+            } else if let (Some(sx), Some(sy)) = (x.as_str(), y.as_str()) {
+                match (
+                    chrono::DateTime::parse_from_rfc3339(sx),
+                    chrono::DateTime::parse_from_rfc3339(sy),
+                ) {
+                    (Ok(dx), Ok(dy)) => dx.cmp(&dy),
+                    // Non-RFC3339 formats the engine may still accept: fall back
+                    // to a deterministic byte order rather than misrank silently.
+                    _ => sx.cmp(sy),
+                }
+            } else {
+                Equal
+            }
+        }
+        (Some(_), None) => Less,
+        (None, Some(_)) => Greater,
+        (None, None) => Equal,
+    }
+}
+
 pub async fn eql_search(
     State(state): State<AppState>,
     Path(index): Path<String>,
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification — two rounds

### Round 1 (first draft) — 2 BLOCKING, 1 NON_BLOCKING → did NOT merge, fixed

🔴 **CORRECTNESS — BLOCKING**
> `eql_timestamp_cmp`'s ISO-8601 branch did a raw lexical `sx.cmp(sy)`. Reproduced against a live server (size=6, pure merge-order): (1) mixed fractional precision `…01.500Z` vs `…01Z` → returned `["…01.500Z","…01Z"]`, reversed (`.`(0x2E) < `Z`(0x5A)); (2) non-UTC offset `…05:00:00+05:00` (=00:00:00Z, earlier) vs `…01:00:00Z` → reversed (`"01"` < `"05"` lexically). Both are the norm for SIEM/log corpora — reintroduces the exact #567 defect. The comparator must parse to a common instant.

🔴 **EVIDENCE — BLOCKING**
> Proven end-to-end with size=1: mem-a `…01Z` (1.000s) and mem-b `…01.500Z` (1.500s) → the fix returned the LATER event and DROPPED the true earliest. "Closes #567" overstated: only fixed for uniform-format timestamps.

🟢 **REGRESSION — NON_BLOCKING** (full suite green; flagged the same lexical issue as a follow-up).

### The fix
`eql_timestamp_cmp` now parses ISO/RFC3339 to a real instant via `chrono::DateTime::parse_from_rfc3339(...).cmp(...)` (epoch stays numeric; unparseable → byte fallback; missing → last). Added `eql_events_order_by_instant_not_lexical_string` (fractional precision + `+05:00` offset). Fail-before (revert to lexical) → the wrong order.

### Round 2 (chrono fix) — 3/3 NON_BLOCKING

🟢 **CORRECTNESS** — > both prior BLOCKING cases resolved; independent chrono probe confirms `…01.500Z` sorts after `…01Z` and `+05:00` normalizes to `00:00:00Z` (before `…01Z`). Break attempts safe: bare date/malformed → byte fallback no panic; `Z` vs `+00:00` → Equal + stable; negative offset, leap second `:60Z` fine. `total` unchanged.

🟢 **REGRESSION** — > `cargo +1.98.0 test -p xerj-api`: **485 passed, 0 failed, 0 flips**, exit 0. Change confined to es_compat.rs comparator + eql_search + the test. All 4 eql tests pass; the 2 no-@timestamp ones exercise the `unmapped_type` degradation. `-D warnings` clean.

🟢 **EVIDENCE** — > fail-before reconstructed (lexical → wrong order); test genuine (real fractional + offset docs across two members, asserting true-instant order). "Closes #567" now honest. Single-member-over-size flows through the same per-member sorted query — follow-up test hardening, not a blocking gap.

**Tally: Round 2 = 3 NON_BLOCKING, 0 BLOCKING.** Gate (b) satisfied. Follow-up nits (mixed number-vs-string across members → Equal; single-member-over-size test hardening) noted, non-blocking. Holding on (a) all-green + (c) CLEAN.

Issue #310, opened 2026-08-11
fields API silently omits embedding companions under the new default `_source` (follow-up to #309)


xerj-org:
Found in the pre-merge review of #309. Two related semantics gaps in the new `SourceFilter::Default` path:

1. **`fields` resolution runs against the engine-filtered source.** The `fields` builder (`engine/crates/xerj-api/src/es_compat.rs` ~10784–10828) resolves values via `get_source_value_by_path(&h.source, ...)` after the engine has already stripped companions, so `{"fields": ["body_vector"]}` with `_source` omitted returns nothing for that entry — silently, since missing fields are legally omitted from `fields`. Asymmetry: `"_source": false` + `fields: ["body_vector"]` still works, because `Enabled(false)` deliberately keeps the raw source for exactly this resolution (`index.rs` ~24478). #309's compat principle — "an explicitly named companion is still returned" — holds for `_source` projections but not for `fields`. Workaround: `"_source": ["body_vector"]`. Proposed fix: an explicit `fields` request naming a companion should pierce the default the way explicit `_source` includes do.

2. **A dotted `embedding.target_field` flips `filter_object` into the nested branch, which prunes empty-object fields.** `index.rs` ~24538 (`has_dotted`) + ~24630 (`collect_and_filter` drops empty non-root objects): if any field's embedding config sets a dotted `target_field`, every default search silently drops unrelated empty-object fields (`{"meta": {}}` vanishes) from every hit. Pre-existing for `Includes`/`Fields` filters; newly exposed via the `Default` arm. Worth a guard.

Neither blocks correctness for the common case; conformance is 0-failed on the merged head.

xerj-org:
### Status: part 2 fixed in #599; part 1 implementation plan

**Part 2** (a dotted `_source` filter drops an already-empty object like `{"meta":{}}` from every hit) is fixed in **PR #599** — `collect_and_filter` now prunes only objects that *filtering* emptied, gated to an accepting state so include-mode `{"meta":{}}` under `meta.foo` still drops exactly as ES does.

**Part 1** (an explicitly-named companion in `fields` returns nothing under default `_source`) analysis, for whoever picks it up:
[...]

PR #599, opened 2026-08-22
fix(engine): preserve an already-empty object under a dotted _source filter (#310 part 2)

xerj-org:
## #310 part 2 — silent, corpus-wide omission of already-empty objects

`filter_object` (engine/crates/xerj-engine/src/index.rs) takes its nested `collect_and_filter` branch whenever any include/exclude path is dotted, and pruned every object that filtered to empty (`result.is_empty() && !prefix.is_empty() -> Null`). It could not tell an object that *filtering* emptied from one that was **already** `{}` in the source, so a document field like `{"meta": {}}` was silently dropped from every hit as soon as any dotted path entered the request (the Default `_source` arm reaches this whenever an embedding `target_field` is dotted).

## Fix (final form, after three verification rounds)

Keep an already-empty object only in an ES **accepting** state — matching `XContentMapValues.filter`'s `isAccept`, not merely "this path is a live prefix on the way to a deeper include":

```rust
let path_is_accepting = |path: &str| {
    includes.iter().any(|inc| match inc.strip_suffix(".*") {
        Some(base) => path.starts_with(&format!("{base}.")), // base.* accepts strictly BELOW base (trailing dot required)
        None => matches_path(path, std::slice::from_ref(inc)), // plain path / full glob match
    })
};
let already_empty_and_accepted = obj.is_empty() && (includes.is_empty() || path_is_accepting(prefix));
if result.is_empty() && !prefix.is_empty() && !already_empty_and_accepted { Value::Null } else { Value::Object(result) }
```

- match-all / exclude-only → the empty object survives (the reported bug);
- plain `meta.foo` → bare `meta` is a non-accepting prefix → dropped, as ES does;
- glob `meta.*` → ES's automaton needs the trailing dot, so bare `meta` (and prefix-sharing `metadata`) drop while a path strictly below `meta` is accepted. `matches_path` could NOT be reused as the accept test — its `.*` clause matched the dot-less prefix.

Only objects that *filtering* emptied are still pruned. For non-empty objects the diff is a no-op.

## Verification
- Tests: `dotted_filter_keeps_an_already_empty_source_object_but_prunes_a_filtering_emptied_one`; `dotted_include_keeps_an_empty_object_only_when_its_own_path_is_accepted` (plain-prefix drops; own-path preserves; `.*`-glob prefix drops; strictly-under-glob preserves; prefix-sharing `metadata` drops; non-empty sibling drops the same way).
- **Fail-before proven** at each step (revert `!obj.is_empty()` → exclude case fails; revert the accept gate to `matches_path` → `.*`-glob leaks `{"meta":{}}`).
- fmt clean; clippy `-D warnings` 0; **full `xerj-engine` suite 49 binaries / 0 failed** (lib 614 passed).
- **Three** adversarial 3-skeptic rounds shaped this: round 1 caught plain-prefix over-preservation, round 2 caught `.*`-glob divergence, round 3 = 0 BLOCKING. Verdicts in the comments.

## Scope
Addresses **#310 part 2** only. Part 1 (a `fields`-named companion resolved against the Default-stripped source) is a separate cross-layer change tracked on #310. Pre-existing `matches_path` glob divergences surfaced during verification (unchanged by this PR, identical on main) are filed separately.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -26484,13 +26484,29 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
                 // `{}` behind. An object that was ALREADY empty in the source is a
                 // real value the document carried -- but keep it ONLY in an
                 // ACCEPTING state: includes is match-all, or this object's own
-                // path is allow-listed (matches an include), NOT merely a prefix
-                // of a deeper one. In pure include mode `{"meta": {}}` under
-                // include `meta.foo` has nothing allow-listed below it, so ES
-                // drops it and so must we; under a match-all/exclude-only filter
-                // the empty object must survive (#310).
+                // path is allow-listed, NOT merely a prefix on the way to a
+                // deeper include. In pure include mode `{"meta": {}}` under
+                // include `meta.foo` (or `meta.*`) has nothing allow-listed below
+                // it, so ES drops it and so must we; under a match-all /
+                // exclude-only filter the empty object must survive (#310).
+                //
+                // `matches_path` cannot be reused as the accept test: it also
+                // returns true for a prefix (to DRIVE recursion), and its glob
+                // clause `p.ends_with(".*") && path.starts_with(p[..len-2])` marks
+                // bare `meta` as matching `meta.*`. ES's automaton needs the
+                // trailing dot -- `meta` is a live but non-accepting state of
+                // `meta.*` -- so the accept test must require it.
+                let path_is_accepting = |path: &str| {
+                    includes.iter().any(|inc| match inc.strip_suffix(".*") {
+                        // `base.*`: accept only paths strictly BELOW `base`.
+                        Some(base) => path.starts_with(&format!("{base}.")),
+                        // Plain paths and other globs: a full `matches_path`
+                        // (path is the include, under it, or a full glob match).
+                        None => matches_path(path, std::slice::from_ref(inc)),
+                    })
+                };
                 let already_empty_and_accepted =
-                    obj.is_empty() && (includes.is_empty() || matches_path(prefix, includes));
+                    obj.is_empty() && (includes.is_empty() || path_is_accepting(prefix));
                 if result.is_empty() && !prefix.is_empty() && !already_empty_and_accepted {
                     Value::Null
                 } else {
@@ -26638,6 +26654,33 @@ mod source_filter_tests {
             ),
             json!({}),
             "a non-empty object with no allow-listed leaf is dropped, same as the empty one (#310)"
+        );
+        // Glob form: `meta.*` — bare `meta` is a non-accepting prefix (ES's
+        // automaton needs the trailing dot), so an empty `{"meta":{}}` drops just
+        // as under `meta.foo`. Reusing `matches_path` as the accept test leaked
+        // this because its `.*` clause matches the dot-less prefix.
+        assert_eq!(
+            filter_object(&json!({ "meta": {} }), &["meta.*".to_string()], &[]),
+            json!({}),
+            "an empty object under a `.*` glob prefix is not allow-listed (#310)"
+        );
+        // But an empty object strictly BELOW the glob base IS accepted.
+        assert_eq!(
+            filter_object(
+                &json!({ "meta": { "foo": {} } }),
+                &["meta.*".to_string()],
+                &[]
+            ),
+            json!({ "meta": { "foo": {} } }),
+            "an empty object strictly under a `.*` glob is preserved (#310)"
+        );
+        // Boundary: a sibling key that merely shares a textual prefix
+        // (`metadata` vs glob `meta.*`) is NOT accepted -- the accept test
+        // requires the dot boundary, unlike a bare `starts_with("meta")`.
+        assert_eq!(
+            filter_object(&json!({ "metadata": {} }), &["meta.*".to_string()], &[]),
+            json!({}),
+            "a prefix-sharing sibling (metadata vs meta.*) is not accepted (#310)"
         );
     }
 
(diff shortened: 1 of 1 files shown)


xerj-org:
## 3-skeptic verification — verdicts verbatim (reworking on the BLOCKING)

**correctness — VERDICT: BLOCKING**
> The `&& !obj.is_empty()` guard is too broad: it unconditionally preserves an already-empty source object even in **include** mode, where ES (and the pre-fix code) correctly drops it. `filter_object({"meta":{}}, includes=["meta.foo"], excludes=[])` returns `{"meta":{}}`; ES returns `{}` — `meta` is only a non-accepting *prefix* of `meta.foo`, nothing under it is allow-listed. Pre-fix returned `{}` (pruned), so the fix REGRESSES it. Internal inconsistency: under `["meta.foo"]`, `{"meta":{"bar":1}}` → `{}` but `{"meta":{}}` → `{"meta":{}}`. Trigger: explicit `_source.includes` with a dotted entry whose parent object is empty in some docs. Fix direction: gate the relaxation to an accepting state, not `obj.is_empty()` alone.

**evidence — VERDICT: NON_BLOCKING**
> Fail-before reconstructed exactly (`meta` dropped without the guard); all 8 source_filter tests green. Default-arm faithfulness confirmed (a dotted `target_field` builds a dotted companion exclude → `has_dotted` → `collect_and_filter`, the mechanism the test exercises). Traced `Includes(["foo.bar"])` with an *unrelated* `{"meta":{}}` — still dropped at the keep-gate. (Note: this lens did not test the *prefix* case the correctness lens found.)

**regression — still running at time of rework.**

**Action:** the BLOCKING finding is real and confirmed (the two verdicts are not in conflict — the evidence lens tested an *unrelated* empty object, correctly dropped; the correctness lens tested an empty object that *prefixes* the include, wrongly leaked). Reworking: gate the already-empty relaxation to an **accepting** state — `includes.is_empty() || matches_path(prefix, includes)` — so `{"meta":{}}` under include `meta.foo` is still dropped, while it survives an exclude-only / match-all filter and an own-path include. A fresh 3-skeptic pass will run on the refined fix before any merge.

xerj-org:
## Round-2 verification — verdicts verbatim (3/3 BLOCKING; reworked in bd255d37)

All three round-2 skeptics independently found the **same** ES-divergence: the accepting gate reused `matches_path` as its accept test, but `matches_path`'s glob clause (`p.ends_with(".*") && path.starts_with(&p[..p.len()-2])`) marks a dot-less prefix as matching a `.*` glob.

**correctness — BLOCKING**
> `filter_object({"meta":{}}, includes=["meta.*"], excludes=[])` → XERJ `{"meta":{}}`, ES `{}`. ES compiles `meta.*` → shortest accepted string is `meta.`, so bare `meta` is a live-but-NON-accepting state → dropped. Same class deeper: `["a.b.*"]` over `{"a":{"b":{}}}`. The single failing axis is `<path>.*` matching the empty object sitting exactly at `<path>`.

**regression — BLOCKING**
> `filter_object({"metadata":{}, "a":{"b":1}}, ["meta.*","a.b"], [])` → PR emits `{"metadata":{}, ...}`; ES and main drop `metadata` (`"metadata".starts_with("meta")` boundary bug). Full suite otherwise clean (0 failed), but it has NO glob-`.*`-empty case, so it couldn't catch this.

**evidence — BLOCKING**
> Fail-before confirmed for plain paths. `.*`-glob leak reproduced. Coverage gap: the new test only exercised plain dotted includes. Required missing assertion: `filter_object({"meta":{}}, ["meta.*"], [])` == `{}`. (Also flagged the PR body still described the earlier broad fix — now updated.)

**Action (bd255d37):** replaced the accept test with `path_is_accepting`, honouring the dot boundary — `base.*` accepts only paths strictly below `base` (`starts_with("base.")`), so bare `meta` and prefix-sharing `metadata` both drop while `meta.foo` is preserved. Added `.*`-glob, strictly-under-glob, and `metadata` boundary assertions; fail-before proven; full suite 49/0. Round-3 verification running before any merge.

xerj-org:
## Round-3 verification — verdicts verbatim (3/3 NON_BLOCKING → 0 BLOCKING)

> 20-case probe driving `filter_object` directly, compared to ES. `path_is_accepting` is ES-correct for every empty object that reaches it — `meta.*` over `{meta:{}}`→`{}`, over `{meta:{foo:{}}}`→preserved, `{metadata:{}}`→`{}`, deep `a.b.*` both ways; plain/own-path/`[*,z.x]`/`a.*` all correct. No over-preservation, no gate-caused wrong drop; round-1/round-2 leaks closed. Surfaced 7 divergences that are all PRE-EXISTING defects of the general filter machinery (`matches_path`/`keep` descent), unchanged by this commit and identical on main (proven via non-empty controls); out of scope, not attributable to the accept-test change.

**regression — NON_BLOCKING**
> Full suite exit 0; lib `614 passed; 0 failed`. Sole behavioral change is the empty-object accept decision (git diff removes one prod line, adds the accept block; rest is test code). Strict narrowing vs the round-1 gate; widens correctly vs main. `base.*` == `path.starts_with("base.")` matches ES `makeAnyString`. One residual multi-glob incompleteness (`a.*.*`) is pre-existing, not introduced.

**evidence — NON_BLOCKING**
> Fail-before confirmed (`meta.*` assertion fails when reverted to `matches_path`). Coverage complete (`.*`-glob, strictly-under-glob, prefix-sharing `metadata` all pinned). Should-fix (out of scope): pre-existing `matches_path` divergences; PR body was stale (now updated to `path_is_accepting`). #310 correctly stays OPEN.

**Decision:** 0 BLOCKING across all three lenses. Merge-eligible once #600 (rc.34 stamp-bump) clears the tag-ahead guard and this rebases green + CLEAN. The pre-existing `matches_path` glob divergences are filed as a follow-up (not a blocker for this PR, which does not touch non-empty filtering).

Issue #566, opened 2026-08-21
collapse inner_hits: harden the seq_no/version None-fallbacks and extend #506 test coverage


xerj-org:
Follow-ups surfaced by the adversarial verification of #562 (which fixed #506(1), the collapse `inner_hits` positional-`_seq_no` / live-`_version` torn read). All three are **NON_BLOCKING** — #562 is a strict improvement and these were not merge-blockers — but they are real and worth tracking.

### 1. The `None` fallback can still fabricate a positional `_seq_no` (correctness)
`Hit.seq_no` / `Hit.version` are `Option<u64>`. When a member's snapshot value is `None` — a concurrent tombstone or a version-map miss — `json!({"_seq_no": h.seq_no})` serializes `null`, so the renderer's `m.get("_seq_no").and_then(as_u64)` is `None` and the `unwrap_or(i as u64)` / `unwrap_or(1)` fallbacks fire, re-fabricating a **positional** `_seq_no` (the exact #506 bug) plus `_version = 1` for that member. Members are sorted by the inner_hits sort before `enumerate`, so `i` is an arbitrary position.

This is a residual of the *old* bug reachable only in a delete race — **not** a regression (pre-#562 the path was `json!(i as u64)` unconditionally). Suggested fix: when the snapshot value is `None`, **omit** `_seq_no` / `_version` entirely (Elasticsearch omits `_seq_no` when it cannot version a doc) rather than emit a positional value a client could feed back into `if_seq_no`. Refs: `es_compat.rs:13205-13210`, the `fill_memtable_sources` seq/version fill in `index.rs`.

### 2. The `_version` half of the fix ships untested (coverage)
The #506 repro (`collapse_inner_hits_seqno_is_real.rs`) requests only `seq_no_primary_term` and asserts exclusively on `_seq_no`; it never reads `_version` and never mutates a doc post-snapshot. So the *old* live-`lookup_version()` code path would also pass it (docs are indexed once → live == snapshot). Reverting only the `_version` hunk stays green. Add a case that updates a doc between snapshot and render and asserts the **snapshot** `_version`.

### 3. No scroll-path coverage (coverage)
[...]

PR #601, opened 2026-08-22
fix(api): omit absent snapshot _seq_no/_version in collapse inner_hits, don't fabricate (#566 item 1)

xerj-org:
## #566 item 1 — a fabricated `_seq_no`/`_version` in a delete race

In collapse `inner_hits`, a member's snapshot `_seq_no`/`_version` (`Hit.seq_no`/`version`, `Option<u64>`) can be `None` — a concurrent tombstone or a version-map miss — serialized as JSON `null` in the collapse group. The renderer fell through:

```rust
let sn = m.get("_seq_no").and_then(Value::as_u64).unwrap_or(i as u64); // positional!
let v  = m.get("_version").and_then(Value::as_u64).unwrap_or(1);
```

re-fabricating a **positional** `_seq_no` (the exact #506 bug — `i` is the post-sort array index) and `_version = 1`. A client could feed those back into `if_seq_no`/`if_primary_term` and corrupt optimistic concurrency. Residual of #506, reachable only in a delete race (so not a regression), but a real wrong-value bug.

## Fix

Extract `emit_inner_hit_versioning` and **omit** the field when the snapshot value is absent — as Elasticsearch does when it cannot version a doc. `i` was the only remaining use of the members' `.enumerate()`, so it's dropped too.

```rust
if emit_version { if let Some(v) = member.get("_version").and_then(Value::as_u64) { hit_obj.insert("_version".into(), json!(v)); } }
if emit_seq_no  { if let Some(sn) = member.get("_seq_no").and_then(Value::as_u64) { hit_obj.insert("_seq_no".into(), json!(sn)); hit_obj.insert("_primary_term".into(), json!(1)); } }
```

## Verification
- Test `absent_snapshot_versioning_is_omitted_not_fabricated`: null snapshot → neither `_seq_no`/`_primary_term` nor `_version` emitted; real values emitted verbatim; flags-off emits nothing.
- **Fail-before proven** (restore the `unwrap_or` fallbacks): the omit assertions fail — the fields get fabricated.
- fmt clean; clippy `-D warnings` 0 under rustc 1.98.0; **full `xerj-api` suite: 50 binaries, 0 failed**.

## Scope
Addresses **#566 item 1** (correctness). Items 2 (a `_version`-mutation test) and 3 (a scroll+update test) are coverage follow-ups; #566 stays open for them.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8311,6 +8311,78 @@ pub async fn search(
     }
 }
 
+/// Emit a collapse inner-hit member's snapshotted `_version` / `_seq_no` +
+/// `_primary_term` into `hit_obj`, honouring the inner_hits `version` /
+/// `seq_no_primary_term` flags.
+///
+/// When the snapshot value is ABSENT -- `Hit.seq_no`/`version` was `None` (a
+/// concurrent tombstone or a version-map miss) and serialized as JSON `null` in
+/// the collapse group -- the field is OMITTED, not fabricated. Elasticsearch
+/// omits versioning it cannot resolve; the previous code fell through
+/// `unwrap_or(i as u64)` / `unwrap_or(1)`, re-fabricating a *positional* `_seq_no`
+/// (the exact #506 bug) and `_version = 1` -- values a client could feed back
+/// into `if_seq_no` / `if_primary_term` and corrupt optimistic concurrency
+/// (#566, residual of #506).
+fn emit_inner_hit_versioning(
+    member: &Value,
+    emit_seq_no: bool,
+    emit_version: bool,
+    hit_obj: &mut serde_json::Map<String, Value>,
+) {
+    if emit_version {
+        if let Some(v) = member.get("_version").and_then(Value::as_u64) {
+            hit_obj.insert("_version".to_string(), json!(v));
+        }
+    }
+    if emit_seq_no {
+        if let Some(sn) = member.get("_seq_no").and_then(Value::as_u64) {
+            hit_obj.insert("_seq_no".to_string(), json!(sn));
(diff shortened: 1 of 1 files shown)


xerj-org:
## 3-skeptic verification — verdicts verbatim (3/3 NON_BLOCKING → 0 BLOCKING)

> `emit_inner_hit_versioning` reads `.and_then(Value::as_u64)`, so JSON `null` OR an absent key both yield `None` → field skipped, killing the `unwrap_or(i as u64)`/`unwrap_or(1)` fabrication. Mixed presence handled independently (`_primary_term` only alongside a real `_seq_no`). Dropping `.enumerate()` is safe — `i` had no other use; skip/take unchanged. Test passes. Only one collapse inner_hits render path exists and it now routes through the helper.

**regression — NON_BLOCKING**
> Full `xerj-api` suite: **486 passed, 0 failed, 2 ignored**, exit 0. #506 repro still passes (real snapshot seq_no, not positional). Present-case byte-equivalent (same fields/order for real u64 values); only the None branch changed. Flags mapped correctly (not swapped). No name collision.

**evidence — NON_BLOCKING**
> Fail-before proven independently (restore `unwrap_or` → test fails at the omit assertion; revert → passes). Helper is on the real path (`search_impl` calls it at es_compat.rs:13317), not dead code. Test shape representative of the producer (index.rs builds members with `_seq_no: h.seq_no` (`Option<u64>`) → `None` serializes to `null`). #566 correctly OPEN; no closing keyword.

**Decision:** 0 BLOCKING. Merge-eligible on all-green + CLEAN (rebased past rc.34). Two lenses flagged the identical fabrication on the TOP-LEVEL (non-collapse) hit path (`h.seq_no.unwrap_or(hit_idx)` at es_compat.rs:13410/20084/20417) — out of scope here, filed separately.

Issue #574, opened 2026-08-21
keyword match/multi_match not mapping-aware inside Knn/SemanticSearch/function-score filters (follow-up to #354/#573)


xerj-org:
### Context
PR #571 (top-level) + PR #573 (compound wrappers: `constant_score`/`function_score`/`boosting`/`dis_max`/`nested`/`hybrid`/`pinned`/`named`) make keyword `match`/`multi_match` mapping-aware by rewriting them to whole-value `Term`s in `rewrite_keyword_full_text_to_term` (`engine/crates/xerj-engine/src/index.rs`). Surfaced by #573's adversarial verification (correctness + evidence lenses, NON_BLOCKING).

### The remaining gap — filter-carrying nodes are not recursed
Three node kinds carry a `QueryNode` filter/sub-query that the rewrite still skips (`other => other.clone()`), so a keyword `match`/`multi_match` nested in one of them **still flips its hit set at flush** (the #354 defect):
- `Knn.filter: Option<Box<QueryNode>>` — `ast.rs:466`
- `SemanticSearch.filter: Option<Box<QueryNode>>` — `ast.rs:484`
- `FunctionScore.functions[].filter` (`ScoreFunction.filter: Option<QueryNode>`) — `ast.rs:193`

Also: the request-level `rescore` query is not passed through `rewrite_keyword_full_text_to_term` at the search entry.

### Also in scope (test coverage + comment accuracy)
- 5 of the 9 wrapper arms added in #573 (`Boosted`, `Nested`, `Hybrid`, `Pinned`, `Named`) are **correct but untested** (verified field-by-field by the correctness lens; add flush-parity cases).
- The comment in `rewrite_keyword_full_text_to_term` says only leaf/span nodes fall through — inaccurate, since `Knn{filter}`/`SemanticSearch{filter}` also fall through despite carrying a sub-query. Correct the comment when the recursion is extended.

### Fix direction
Extend `rewrite_keyword_full_text_to_term` to recurse into `Knn.filter`, `SemanticSearch.filter`, and each `ScoreFunction.filter`; pass `rescore` through at the search entry; add flush-parity tests for the untested arms + the filter cases; correct the fall-through comment. Prove fail-before per node (keyword `match {kw:"a b"}` inside the filter → hit set flips at flush before the fix).

Follow-up to #354 / #571 / #573.

xerj-org:
### ⚠️ Premise contested — needs a flush repro before any fix

[...]

PR #604, opened 2026-08-22
fix(engine): keyword match/multi_match mapping-aware inside Knn/SemanticSearch/function_score filters (#574)

xerj-org:
## #574 — a keyword full-text clause in a filter still flips its hit set at flush

`rewrite_keyword_full_text_to_term` (PR #571 top-level, PR #573 compound wrappers) rewrites a keyword-field `match`/`multi_match` to a whole-value `Term` so the memtable agrees with the segment and the hit set does not change at `_flush` (#354). It recursed into the compound wrappers but skipped three **filter-carrying** nodes, so a keyword full-text clause nested in one of them still flipped at flush:

- `Knn.filter` and `SemanticSearch.filter` fell through `other => other.clone()` (ast.rs:466/484);
- `FunctionScore` recursed into `query` but cloned `functions` verbatim, so each `ScoreFunction.filter` (ast.rs:193) was skipped.

## Fix
Add arms that recurse into `Knn.filter`, `SemanticSearch.filter`, and each function's filter. Only the filter is a full-text sub-query — the vector/text search itself is untouched. The fall-through comment is corrected to say it now covers only leaf and positional (span) nodes.

## Verification
- Test `keyword_match_in_filter_carrying_nodes_is_rewritten`: a keyword `match {kw:"a b"}` inside each of the three filters is rewritten to `Term{kw:"a b"}`; a non-keyword (text) field inside a filter stays a `Match`.
- **Fail-before proven** (revert the arms): the keyword filters stay `Match { field: "kw", query: "a b", ... }` — the assertion fails.
- fmt clean; clippy `-D warnings` 0 under rustc 1.98.0; **full `xerj-engine` suite: 50 binaries, 0 failed**.

## Scope
Covers the three filter-carrying nodes (the headline #354 gap). Remaining #574 items — the request-level `rescore` query passthrough at the search entry, and flush-parity integration tests for the untested #573 arms (Boosted/Nested/Hybrid/Pinned/Named) — are follow-ups; #574 stays open for them.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -32413,7 +32413,21 @@ fn rewrite_keyword_full_text_to_term(q: &QueryNode, schema: &Schema) -> QueryNod
             max_boost,
         } => QueryNode::FunctionScore {
             query: Box::new(rewrite_keyword_full_text_to_term(query, schema)),
-            functions: functions.clone(),
+            // Each function may carry its own `filter` sub-query targeting a
+            // keyword field; recurse into it too (#574) — otherwise a keyword
+            // `match`/`multi_match` inside a `function_score` function filter
+            // keeps flipping its hit set at flush (#354).
+            functions: functions
+                .iter()
+                .map(|f| {
+                    let mut nf = f.clone();
+                    nf.filter = f
+                        .filter
+                        .as_ref()
+                        .map(|q| rewrite_keyword_full_text_to_term(q, schema));
+                    nf
+                })
+                .collect(),
             score_mode: *score_mode,
             boost_mode: *boost_mode,
             max_boost: *max_boost,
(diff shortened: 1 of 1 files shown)


xerj-org:
## 3-skeptic verification — verdicts verbatim (3/3 NON_BLOCKING → 0 BLOCKING)

> All fields preserved: Knn's 7 (field/vector/k/num_candidates/filter/boost/similarity), SemanticSearch's 5, and FunctionScore via clone-then-reassign-filter (weight/script/name/rank_feature/… intact). Every QueryNode variant that can carry a full-text sub-query now recurses; remaining `other` fall-throughs are leaves/span nodes/Intervals (raw Value). **rescore is correctly deferred, not silently missed** — it's applied post-collection via `score_query_against_doc(rescore_query, &hit.source)`, scoring against `_source` not postings, so it cannot flip the hit set at flush (#354). Recursion is fully general (Knn.filter→Bool→Match rewritten all the way down). Diff +183/-1.

**regression — NON_BLOCKING**
> Full `xerj-engine` suite: **1039 passed / 0 failed**. #354 harness all green (keyword_match_is_mapping_aware, ..._in_wrappers, multi_match_flush_parity). Single commit, one file; only the three arms + comments + test. FunctionScore byte-equivalent for filter=None (`None.map(..)==None`).

**evidence — NON_BLOCKING**
> Fail-before proven (revert the arms → filter stays `Match { field: "kw", ... }`). Faithfulness confirmed: the one search entry (index.rs:14143) rewrites every request query before pre- and post-flush execution, so the unit test guards the real #354 symptom. Non-blocking caveat: an index+flush integration test would be stronger for the Knn/SemanticSearch executor paths (need real vectors/embeddings) — explicitly deferred. #574 stays OPEN; no closing keyword.

**Decision:** 0 BLOCKING. Merge-eligible on all-green + CLEAN. Remaining #574 follow-up is only the flush-parity integration coverage for the #573 arms; the rescore item is confirmed a non-issue for flush-parity.

PR #605, opened 2026-08-22
fix(autoindex): index Java static constants as their own symbols (#500)

xerj-org:
## #500 — a Java constant is not a symbol; it costs its whole ~2 KB parent class

A fair, precise-query retrieval test over Apache Lucene (grep vs XERJ symbol search) found XERJ ingests **32–48× more bytes per answered question than grep** and loses recall (8/9 vs 12/12). Root cause, reproduced against refsym:

```
term name=DEFAULT_MAX_CONN   -> count 0
term name=DEFAULT_BEAM_WIDTH -> count 0
```

`JAVA_Q` (extract/code.rs) captured `class`/`interface`/`enum`/`method`/`constructor` declarations but **not `field_declaration`**, so a Java static constant like `DEFAULT_MAX_CONN` was never a symbol — it only survived folded inside its enclosing class body (~2,232 B, ranked below three constructor overloads), and `DEFAULT_BEAM_WIDTH` was length-normalised to rank 35/38.

## Fix
Add the `static`-anchored field capture that the surrounding comment had **already vetted** but held back only because "no Java corpus is wired into the end-to-end check that proves it" — #500's fair test is exactly that missing evidence:

```
(field_declaration (modifiers "static") declarator: (variable_declarator name: (identifier) @const))
```

The extractor already emits each captured symbol's declaration span (~40–80 B) as `code`, so promoting the declaration *is* the fix: the constant becomes its own `const` symbol whose payload is the ~52 B line, not the 2 KB parent. The `static` modifier is the precision filter — it captured **372 constants with 0 method-local false positives**, where capturing all fields would have pulled in 1311 including private instance state. The comment is rewritten to record that #500 supplied the evidence. Non-static instance fields remain a separate recall question.

## Verification
- Test `java_static_constants_are_symbols`: `public static final int DEFAULT_MAX_CONN = 16` → a `const` symbol; a non-static instance field and a method local are NOT captured.
- **Fail-before proven** (remove the capture line): `DEFAULT_MAX_CONN` is not a symbol — only the class + method are extracted; the test fails.
- fmt clean; clippy `-D warnings` 0 under rustc 1.98.0; **full `xerj-autoindex` suite: 6 binaries / 0 failed**.

## Scope
This is the **extractor half** — it makes the constant a symbol. The end-to-end byte/recall improvement over Lucene (count 0 → 1, ~46 B vs ~2 KB payload, `DEFAULT_BEAM_WIDTH` back to the top) is validated by a re-index and is a follow-up. #500 stays open until that measurement lands. The name.keyword ranking boost (issue's 3rd point) is also a separate follow-up.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -682,23 +682,38 @@ const GO_Q: &str = r#"
 (source_file (var_declaration (var_spec_list (var_spec name: (identifier) @static))))
 "#;
 
-// No constant capture here, deliberately. #170 is about files that extract zero
-// symbols and so become unreachable; Java barely has that failure mode, because
-// every file must declare a type. Measured: 7 of 1043 Java files in one corpus
-// and 1 of 72 in another extract zero symbols under this query (0.7%), against
-// 20 of 331 for Rust and 186 of 400 for C headers. A clean query does exist if
-// the recall is wanted later — `(field_declaration (modifiers "static")
-// declarator: (variable_declarator name: (identifier) @const))` captured 372
-// constants with 0 method-local false positives, where capturing all fields
-// would have pulled in 1311 including private instance state — but it is a
-// recall improvement, not this bug, and no Java corpus is wired into the
-// end-to-end check that proves it, so it is not shipped unverified.
+// #170 is about files that extract zero symbols and so become unreachable; Java
+// barely has that failure mode, because every file must declare a type. Measured:
+// 7 of 1043 Java files in one corpus and 1 of 72 in another extract zero symbols
+// under the type/method patterns (0.7%), against 20 of 331 for Rust and 186 of
+// 400 for C headers. So the `static`-field pattern below was originally held back
+// as "a recall improvement, not #170, and no Java corpus is wired into the
+// end-to-end check that proves it."
+//
+// #500 supplied exactly that missing end-to-end evidence: a fair, precise-query
+// retrieval test over Apache Lucene measured the gap directly — a static constant
+// like `DEFAULT_MAX_CONN` was not a symbol at all (`term name=DEFAULT_MAX_CONN` ->
+// count 0), so a field-level fact only survived folded inside its ~2 KB parent
+// class (32-48x more bytes per answered question than grep) or was
+// length-normalised out of reach (the `DEFAULT_BEAM_WIDTH` rank-35 recall miss).
+// So the pattern is now shipped: it promotes each static constant to its own
+// `const` symbol whose `code` is the ~40-80 B declaration span, not the parent.
+// The `static` modifier is the precision filter — it captured 372 constants with
+// 0 method-local false positives, where capturing ALL fields would have pulled in
+// 1311 including private instance state. Java interface fields are implicitly
+// `public static final` constants but parse as `constant_declaration`, not
+// `field_declaration` (verified via s-expr: `interface I { int IC = 7; }` ->
+// `(interface_body (constant_declaration …))`), so they get their own pattern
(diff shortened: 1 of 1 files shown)


xerj-org:
## Round-1 verification (field-only) — verbatim (3/3 NON_BLOCKING) + strengthening

**correctness — NON_BLOCKING** > capture works (static const → `const`; instance/local/nested-static-class excluded; multi-declarator both captured). KEY QUESTION answered YES: the symbol becomes its own retrievable per-declaration document (`emit_code_doc` loop, locator `code:<line>:<name>`), so `term name=DEFAULT_MAX_CONN` goes 0→1. **One real same-class recall gap: interface constants are MISSED** — in tree-sitter-java an interface field is `constant_declaration`, not `field_declaration`. Recommend a follow-up adding `(constant_declaration declarator: (variable_declarator name: (identifier) @const))`.

**regression — NON_BLOCKING** > full suite 704 passed / 0 failed; `all_queries_compile` passes; index-bloat bounded by the 5000-symbol cap + 2 MB CODE_CAP; precision filter verified (instance field / method local excluded).

**evidence — NON_BLOCKING** > fail-before proven; retrievable confirmed; payload = ~46 B declaration span (empirically). Scope honest, no closing keyword.

## Action — strengthened per the correctness lens (commit 7414230d)
Rather than defer to a follow-up, I folded in the recommended interface capture: JAVA_Q now has **both** `(field_declaration (modifiers "static") …)` (class static constants) and `(constant_declaration …)` (interface constants, implicitly static-final). The test now covers an interface constant (`DEFAULT_BEAM_WIDTH`) and asserts the class constant's `code` is the declaration span (`< 120` chars, contains the decl) — not the parent class. Fail-before proven per pattern; full `xerj-autoindex` suite 6 binaries / 0 failed. A fresh 3-skeptic pass is running on the combined fix before merge.

xerj-org:
## Round-2 verification (combined class-static + interface fix, 7414230d) — verbatim (3/3 NON_BLOCKING → 0 BLOCKING)

> `constant_declaration` occurs ONLY under `interface_body`/`annotation_type_body` (Java makes those fields implicitly `public static final`), so the missing static guard cannot capture a non-constant — confirmed safe via s-expr. Interface `int IC=7; static final int JC=8;` → both `const`; an interface method/nested type is NOT `const`; annotation elements NOT captured. Class pattern unchanged (static const captured; instance/local/static-init-local excluded). No double-capture — field_declaration vs constant_declaration are disjoint node types. Per-declaration docs confirmed: distinct `code:<line>:<name>` locators → distinct retrievable docs, `code` = the declaration line.

**regression — NON_BLOCKING**
> Diff = one file (+62/-11), one commit. Full suite **704 passed / 0 failed**; `all_queries_compile` passes (well-formed query). Fail-before both patterns load-bearing. No double-capture. Inflation bounded by the 5000-symbol cap + 400-char/line cap.

**evidence — NON_BLOCKING**
> Fail-before per pattern proven independently (remove constant_declaration → interface `DEFAULT_BEAM_WIDTH` gone, class survives; remove field_declaration → class `DEFAULT_MAX_CONN` gone, interface survives). `code` is the trimmed declaration line (46 B / 29 B observed), never the parent body. Comment/measurement attribution truthful (past-tense, scoped to the static field pattern). Scope honest: no closing keyword, #500 stays OPEN for the re-index + name.keyword follow-ups.

**Decision:** 0 BLOCKING. Merge-eligible on all-green + CLEAN once #606 (rc.35 stamp-bump) clears the tag-ahead guard and this rebases onto rc.35 main. Also grounded in the real Lucene source: `HnswGraphBuilder.java:51/56` declare `DEFAULT_MAX_CONN`/`DEFAULT_BEAM_WIDTH` as `static final` class fields the field pattern captures.

Issue #464, opened 2026-08-17
Aggregations allocate with no budget: max_query_memory_mb cannot see them, and one request took 5.2x the configured limit


xerj-org:
Found by the RAM audit run for rc.18. This is the reason the new 8 GiB default cap (#461) will not bind on a query-heavy workload.

## `max_query_memory_mb` cannot see aggregations

`check_query_alloc` has exactly **one** call site in the entire engine — `engine/crates/xerj-engine/src/index.rs:13537`, labelled "result-window hydration". The estimate immediately above it:

```rust
let window = (request.from as u64).saturating_add(request.size as u64);
let est = window.saturating_mul(EST_BYTES_PER_HIT /* 1024 */);
```

A standard aggregation request sends `"size": 0`. So `est == 0`, and the guard tests `0 > 536870912` — **always passes**. The error text the guard emits even says *"reduce size/aggregation cardinality"*, but no aggregation path ever calls it.

`max_buckets = 65_536` caps **output** buckets only; the transient structures that discover distinct terms are unbounded.

## Measured

Node configured with exactly the 8 GiB cap's derived budgets (`memtable_budget_mb=2048`, `segment_hydration_cache_mb=1638`, `memory_watermark_mb=7476`, `max_query_memory_mb=512`), 400k docs / 420.2 MB index:

```
baseline RSS after load                     1,900.1 MiB
one {"size":0,"aggs":{"t":{"terms":{"field":"n","size":10000}}}}
                                   VmHWM -> 4,768   MiB
delta                                      +2,640   MiB
```

**One request against a 420 MB index allocated 5.2× the configured 512 MiB per-query budget.** Two concurrent such requests plus baseline exceed the 7,476 MiB watermark — and `max_concurrent_searches` defaults to **64** (global semaphore, plus a hardcoded `Semaphore::new(64)` per index at `index.rs:6055,6391`).

Reproduced on rc.16 as well: baseline 386.2 MiB → VmHWM 2,798 MiB.

## The `top_hits` variant is worse and more common

[...]

PR #608, opened 2026-08-22
fix(engine): bind aggregation owned-source materialisation to max_query_memory_mb (#464)

xerj-org:
## #464 — an aggregation can allocate 5.2× the configured `max_query_memory_mb`

The per-query memory guard (`governor::check_query_alloc`) is applied **only** to the search result window (`from + size` hits, ~1 KiB each — `index.rs:13854`). An aggregation runs with `size: 0`, so `window = 0` → `est = 0` → the guard passes trivially. Yet a **meta-observing** agg (`top_hits`, or one targeting `_id`/`_index`/`_seq_no`) deep-clones **every buffered memtable document** to build its `fast_aggs::MemDocs::Owned` view (`fast_aggs.rs:474`). Under a bulk writer the memtable holds 10⁴–10⁵ docs, so one agg allocated **5.2× the limit** with the guard none the wiser.

## Fix
A second `check_query_alloc` right after the result-window guard, before either the fast (`try_fast_aggs`) or brute agg path runs:

```rust
if let Some(aggs_def) = &request.aggs {
    if fast_aggs::agg_tree_mentions_meta(aggs_def) {          // now pub(super)
        if let Some(g) = governor::global() { if g.query_memory_enabled() {
            const EST_BYTES_PER_OWNED_DOC: u64 = 2048;        // full source clone + meta injection
            let est = self.memtable.doc_count() as u64 * EST_BYTES_PER_OWNED_DOC;
            g.check_query_alloc(est, "aggregation owned-source materialisation")?;
        }}
    }
}
```

At the 512 MB default this trips only above ~256K buffered docs, so ordinary `top_hits` aggs are unaffected. It bounds the owned build the issue measured; bucket-cardinality growth is a separate concern (follow-up).

## Verification
- Own test binary (`Engine::new` installs `max_query_memory_mb` in the process-wide governor): `top_hits_aggregation_is_bound_by_max_query_memory_mb` — 1 MB budget + 900 buffered docs → a `size:0 top_hits` agg is **rejected**, while a plain `size:0` search still passes (isolates the agg path).
- **Fail-before proven**: without the guard the same agg returns `Ok` (the #464 bypass — the test was seen to fail exactly that way before the guard).
- fmt clean; clippy `-D warnings` 0 under rustc 1.98.0; **full `xerj-engine` suite: 51 binaries / 0 failed**.

## Scope
Covers the reported hotspot — the memtable owned-source deep-clone for meta-observing aggs. Bucket-cardinality / `O(N×buckets)` growth and the segment-side owned path are separate concerns; #464 can note them as follow-ups.

### DIFF
--- a/engine/crates/xerj-engine/src/fast_aggs.rs
+++ b/engine/crates/xerj-engine/src/fast_aggs.rs
@@ -5132,7 +5132,7 @@ fn compile_top_pred(filter: &Value) -> Option<Pred> {
 /// costs an extra clone, never correctness — so this deliberately mirrors
 /// the substring semantics (`top_hits` anywhere; the meta fields as exact
 /// key/string tokens, i.e. the serialized quoted form).
-fn agg_tree_mentions_meta(v: &Value) -> bool {
+pub(super) fn agg_tree_mentions_meta(v: &Value) -> bool {
     fn is_meta(s: &str) -> bool {
         s.contains("top_hits") || s == "_id" || s == "_index" || s == "_seq_no"
     }
--- a/engine/crates/xerj-engine/src/fast_aggs.rs
+++ b/engine/crates/xerj-engine/src/fast_aggs.rs
@@ -5132,7 +5132,7 @@ fn compile_top_pred(filter: &Value) -> Option<Pred> {
 /// costs an extra clone, never correctness — so this deliberately mirrors
 /// the substring semantics (`top_hits` anywhere; the meta fields as exact
 /// key/string tokens, i.e. the serialized quoted form).
-pub(super) fn agg_tree_mentions_meta(v: &Value) -> bool {
+fn agg_tree_mentions_meta(v: &Value) -> bool {
     fn is_meta(s: &str) -> bool {
         s.contains("top_hits") || s == "_id" || s == "_index" || s == "_seq_no"
     }
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -13863,6 +13863,33 @@ impl Index {
             }
         }
 
+        // #464: the guard above only covers the search RESULT WINDOW. An
+        // aggregation runs with `size: 0`, so that estimate is 0 — yet a
+        // meta-observing agg (`top_hits`, or one targeting `_id`/`_index`/
+        // `_seq_no`) deep-clones EVERY buffered memtable document to build its
+        // owned-source view (`fast_aggs::MemDocs::Owned`), which under a bulk
+        // writer is 10^4-10^5 full-source clones. That allocation escaped
+        // `max_query_memory_mb` entirely. Estimate it here — before either the
+        // fast or the brute agg path runs — and reject up-front. Bounds the
+        // owned build; bucket-cardinality growth remains a separate concern.
+        if let Some(aggs_def) = &request.aggs {
+            if fast_aggs::agg_tree_mentions_meta(aggs_def) {
+                if let Some(g) = crate::governor::global() {
+                    if g.query_memory_enabled() {
+                        // Full source clone + `_id`/`_index`/`_seq_no` injection
+                        // per buffered doc; a conservative per-doc figure (the
+                        // window guard above bills 1 KiB/hit, an owned agg clones
+                        // the whole source).
+                        const EST_BYTES_PER_OWNED_DOC: u64 = 2048;
+                        let est = (self.memtable.doc_count() as u64)
+                            .saturating_mul(EST_BYTES_PER_OWNED_DOC);
+                        g.check_query_alloc(est, "aggregation owned-source materialisation")
+                            .map_err(EngineError::Common)?;
+                    }
+                }
+            }
+        }
+
         // Determine the timeout: use the request-level timeout if set, otherwise
         // fall back to the default of 30 seconds.
         let timeout_ms = request.timeout_ms.unwrap_or(30_000);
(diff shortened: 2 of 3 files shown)


xerj-org:
## Round-1 verification — verdicts verbatim (1 BLOCKING → reworked)

**correctness — BLOCKING**
> Guard covers both paths and fail-before holds for the top_hits repro. BUT the meta-only keying is too narrow: the brute `need_full_corpus` path (`index.rs:18146`) deep-clones EVERY memtable doc unconditionally (`(*v).clone()`) with NO meta check, so any agg that bails the fast path — **percentiles/date_histogram/cardinality have no columnar path at all** — deep-clones the whole corpus while `agg_tree_mentions_meta` is false, and the guard never fires. Empirical A/B (same 900-doc memtable, 1 MiB budget): top_hits → rejected; **percentiles → Ok**. Also the estimate ignores SEGMENT sources even within the meta scope. The meta-only keying does not close #464's overrun class.

**regression — NON_BLOCKING** > full suite 1042 passed / 0 failed; existing top_hits tests (default 512 MB) not rejected; diff scope exactly the guard + `pub(super)` + test.

**evidence — NON_BLOCKING** > fail-before proven; rejection is genuinely the `check_query_alloc` 429 path (instrumented the error: `CircuitBreaking { … [aggregation owned-source materialisation] }`); estimate math checks (512MB/2048 ≈ 256K docs); scope honest, #464 stays open.

## Rework (this push)
Added a **second guard at the brute-path corpus assembly** (`need_full_corpus`, before the deep-clone) estimating `(memtable + segment docs) × 2048` — so percentiles/date_histogram/cardinality and any fast-path-bailing agg are now bound too, and segment sources are counted. Extended the test with a `percentiles` case: both `top_hits` (fast/meta) and `percentiles` (brute) are now rejected under the 1 MB budget; the plain `size:0` search still passes. **Fail-before proven per guard** (strip fast-path guard → top_hits Ok; strip brute-path guard → percentiles Ok). Full suite 51/0. A fresh 3-skeptic pass runs on the combined fix before merge.

xerj-org:
## Round-2 verification — verdicts verbatim (1 BLOCKING → reworking)

**correctness — NON_BLOCKING** (guards cover their paths; no false positive on cheap fast-path aggs).
**regression — NON_BLOCKING** (full suite green; diff scope confirmed).
**evidence — BLOCKING**
> Fail-before: strip the brute-path guard → percentiles assertion fails (it IS load-bearing). Strip the fast-path guard → the test STILL PASSES — top_hits does NOT fail. In this memtable-only harness EVERY agg (incl. top_hits, even after refresh) is served by the BRUTE path (`precomputed_aggs.is_none()`, index < FAST_AGG_MIN_DOCS=10000), so both assertions are actually bounded by the single brute guard; the fast-path guard has ZERO regression coverage and is deletable with CI green. The commit's "fail-before per guard" overstates it. The fix does bound memory in all observed cases; the defect is evidentiary.

**Action:** the finding is correct. The actually-reported #464 case (bulk writer → large un-flushed memtable → no `.dv` columns → fast path bails → brute path) is caught by the brute-path guard, which is the load-bearing, tested one. The fast-path guard only uniquely matters for a fast-served meta agg over ≥10000 flushed docs beside a large un-flushed memtable — a narrow edge case with no cheap reproduction. **Removing the fast-path guard** (keeping the tested brute-path corpus guard, and reverting `agg_tree_mentions_meta` to private) and noting the fast-path owned-build residual as a #464 follow-up. Re-verifying the single-guard fix.

xerj-org:
## Round-2 BLOCKING resolved — single tested guard (ecae26fc)

The evidence lens was correct: the fast-path meta guard (index.rs ~13875) had **zero unique fail-before coverage**. In the memtable-only harness every agg (< `FAST_AGG_MIN_DOCS`) is served by the BRUTE path, so the brute-path guard backstopped `top_hits`; stripping the fast-path guard alone left the test green — it was deletable.

**Resolution:** removed the fast-path guard, kept the brute-path corpus guard at the actual deep-clone site. This is the load-bearing one and it covers the whole family that reaches the brute clone: the actually-reported #464 case (bulk writer → large un-flushed memtable → no `.dv` columns → fast path bails → brute path), plus the no-columnar-path families `percentiles`/`date_histogram`/`cardinality` (which carry no meta field and an `agg_tree_mentions_meta`-keyed guard would have missed). `agg_tree_mentions_meta` reverted to private (no external caller; fast_aggs.rs now matches main).

**Fail-before (single guard, verified under 1.98.0):** disabling only the brute-path guard (`if false && need_full_corpus`) → the `top_hits` assertion fails first, `percentiles` would fail too — both are now bound solely by this guard. Restored → test green.

**Verification:** fmt ✓, clippy -D warnings ✓, full `cargo test -p xerj-engine` = **1042 passed / 0 failed** across 51 binaries.

Residual (tracked on #464, not closed here): fixed 2048 B/doc estimate skews on large-source docs; bucket-cardinality growth is separate; the fast-path memtable owned-build (fast-served meta agg over ≥10000 flushed docs beside a large un-flushed memtable) is a narrow follow-up — noted for a dedicated integration harness.

Launching round-3 3-skeptic verification on this single-guard revision.

xerj-org:
## Round-3 verification — 3 skeptics, verdicts verbatim (single-guard revision bb1e8dd7)

> Ran `cargo test -p xerj-engine --test agg_memory_budget` → ok. Probe 1: guard (index.rs:18128) gated on `need_full_corpus`; the deep-clone `all_docs` (18147) is gated on the SAME predicate — guard fires exactly on the path that allocates. Probe 2: `need_full_corpus = precomputed_aggs.is_none() && request.aggs.is_some()`; when the columnar path precomputes (`Some`), `need_full_corpus` is false → guard skipped → a size:0 columnar agg that does NOT deep-clone is never charged. No false positive. Probe 3: est = (memtable + Σ every segment.doc_count) × 2048, saturating — sound bound. Confirmed the fast-path/`agg_tree_mentions_meta` guard was removed (`git diff origin/main -- fast_aggs.rs` empty; predicate now private). No correctness hole.

> Guard site index.rs:18128–18142, distinct from the untouched clone line 18147. Baseline (guard present): test passes (control Ok, top_hits Err, percentiles Err). Fail-before: edited header to `if false && need_full_corpus {` → test FAILED at agg_memory_budget.rs:91 ("Got Ok, so the agg path bypassed the guard"). Guard is load-bearing. Restored → test passes. No coverage defect.

**regression — NON_BLOCKING**
> `git diff --stat origin/main` = exactly 2 files: index.rs (+27) and new tests/agg_memory_budget.rs (+112); 0 deletions. `git diff origin/main -- fast_aggs.rs` EMPTY (identical to main). `agg_tree_mentions_meta` private, still referenced (no dead-code). Snapshot nesting (18128–18143): `self.store.snapshot()` strictly inside `need_full_corpus`/governor-enabled — never on no-agg or non-governor searches. `cargo test -p xerj-engine` (RUST_MIN_STACK=8388608): exit 0, aggregate 1042 passed / 0 failed / 23 ignored across 51 binaries. No regression or scope defect.

**Gate:** 3/3 NON_BLOCKING, ZERO BLOCKING. Merging on all-checks-green + CLEAN.

PR #610, opened 2026-08-22
fix(engine): stop painless recursion tests SIGABRT-ing on a debug build (#353)

### LINKED ISSUE #353: painless_script_limits: call_depth_limit_survives_the_multi_thread_block_in_place_path overflows the stack and aborts the process on main

xerj-org:
## What

`cargo test -p xerj-engine --test painless_script_limits` on a default (debug) build aborts the whole binary with a stack overflow (SIGABRT), so 7 of 9 cases never run — the abort masks them (#353).

## Root cause — test-harness stack budget, not the interpreter

The closure call-depth limit (`MAX_CALL_DEPTH = 32`) **is** enforced on every path — both closure-call sites go through `CallGuard::enter` — and its adversarial worst case is a release-measured **1.20 MiB**. That is exactly why:
- production pins its worker stack to **4 MiB** (`RT_THREAD_STACK_SIZE`, with a compile-time headroom assertion), and
- CI stays green: the "Build + Test" job runs the optimized `ci-test` profile, where the 1.20 MiB worst case fits even a 2 MiB stack (`worst_case_nested_block_recursion_still_fits_a_2mib_stack` passes there).

The six deep-recursion cases here trip the limit **through the async search path** (`function_score` / `rescore` / `terms_set` / a `block_in_place` hand-off), so the whole search stack sits *beneath* the evaluator. A **debug** build's per-call-level frames are several times a release build's — the same reason `worst_case_...` is already `#[cfg(not(debug_assertions))]`. On a stock `#[tokio::test]` **2 MiB** worker that overflows before the depth limit can trip, and `panic = "abort"` turns it into a whole-process kill.

## Fix

Run those six cases on a runtime hosted on a thread with a **debug-adequate 16 MiB** stack (`RECURSION_TEST_STACK`) — the test analogue of production's stack pin — so the **same** limit trips *gracefully* (a `script_failure` on the result) in every profile instead of racing the thread's stack guard. Two helpers preserve the current-thread vs multi-thread (`block_in_place`) distinction, and the harness re-raises the body's own panic so assertion messages/backtraces survive. **`MAX_CALL_DEPTH` is untouched — the ceiling is not raised, and no production code changes.**

- **Fail-before:** on current main, `env -u RUST_MIN_STACK cargo test -p xerj-engine --test painless_script_limits` aborts (`signal: 6, SIGABRT`), only 2/9 cases report.
- **After:** the same command reports **9 passed, 0 failed** on a debug build with no `RUST_MIN_STACK`.
- `cargo fmt --check` ✓, `clippy --tests -- -D warnings` ✓.
- Optimized profile unchanged: `cargo test --profile ci-test --test painless_script_limits` → 10 passed.
- Full crate suite: `cargo test -p xerj-engine` → **1042 passed, 0 failed** across 51 binaries.

Closes #353.

### DIFF
--- a/engine/crates/xerj-engine/tests/painless_script_limits.rs
+++ b/engine/crates/xerj-engine/tests/painless_script_limits.rs
@@ -25,6 +25,79 @@ fn test_engine() -> (Engine, TempDir) {
     (engine, dir)
 }
 
+/// Stack budget for the async recursion tests below.
+///
+/// These tests deliberately trip the closure-recursion limit *through the async
+/// search path* (`function_score` / `rescore` / `terms_set` / a `block_in_place`
+/// hand-off), so the whole search stack sits beneath the evaluator. The limit
+/// itself is sound — `MAX_CALL_DEPTH` trips at a release-measured 1.20 MiB worst
+/// case, and the server pins its worker stack to 4 MiB (`RT_THREAD_STACK_SIZE`)
+/// for exactly that budget. But a DEBUG build's per-call-level frames are
+/// several times a release build's (the same reason
+/// `worst_case_nested_block_recursion_still_fits_a_2mib_stack` is release-only),
+/// and a stock `#[tokio::test]` worker gets only 2 MiB. In debug that overflows
+/// and SIGABRTs the whole binary — silently masking every other test in the file
+/// (issue #353). We give these tests a debug-adequate stack — the test analogue
+/// of the production pin — so the SAME limit trips *gracefully* (a `script_failure`
+/// on the result) in every profile instead of racing the thread's stack guard.
+/// This does NOT raise the ceiling (`MAX_CALL_DEPTH` is untouched); it only lets
+/// the existing ceiling be the binding limit under debug's inflated frames.
+const RECURSION_TEST_STACK: usize = 16 * 1024 * 1024;
+
+/// Drive `body` to completion on a current-thread runtime hosted on a thread
+/// with [`RECURSION_TEST_STACK`] of stack (the current-thread analogue: the
+/// runtime runs everything on this thread).
+fn on_big_stack_current_thread<F>(body: F) -> F::Output
+where
+    F: std::future::Future + Send + 'static,
+    F::Output: Send + 'static,
+{
+    std::thread::Builder::new()
+        .stack_size(RECURSION_TEST_STACK)
+        .spawn(move || {
+            tokio::runtime::Builder::new_current_thread()
+                .enable_all()
+                .build()
+                .expect("build current-thread test runtime")
+                .block_on(body)
+        })
+        .expect("spawn test thread")
+        .join()
+        // Re-raise the body's own panic (assertion message + backtrace intact)
+        // rather than flatten it to a generic string. A stack *overflow* aborts
+        // the process outright — it never reaches here — which is exactly the
+        // #353 failure this larger stack prevents.
+        .unwrap_or_else(|e| std::panic::resume_unwind(e))
+}
+
+/// Drive `body` on a multi-thread runtime whose worker threads each get
+/// [`RECURSION_TEST_STACK`] — the path that exercises `Index::search`'s
+/// `block_in_place` + `Handle::block_on` hand-off onto a worker thread.
+fn on_big_stack_multi_thread<F>(body: F) -> F::Output
+where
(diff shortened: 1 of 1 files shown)


xerj-org:
## Verification — 3 skeptics, verdicts verbatim (HEAD a8d7af38). 3/3 NON_BLOCKING.

> Diff is exactly one test file; `git diff origin/main -- '*/src/*'` empty; `MAX_CALL_DEPTH = 32` untouched. Full debug run (no RUST_MIN_STACK) → 9 passed, 0 failed, no SIGABRT (`worst_case` is `cfg(not(debug_assertions))`, correctly excluded in debug). Non-vacuity: weakened `call_depth_limit_survives_the_multi_thread_block_in_place_path` assert to `is_none()` → test FAILED with real message `the multi-thread block_in_place path lost the fault; scores [0.0]` — proves the body genuinely runs, `script_failure` was truly `Some`, and the spawned big-stack thread's panic is surfaced via `resume_unwind` (helper does NOT swallow failures). All six converted tests keep real assertions.

**regression — NON_BLOCKING**
> Scope: diff = exactly `tests/painless_script_limits.rs` (207+/121-); no `src/` or `Cargo.toml`. Production byte-identical: `MAX_CALL_DEPTH=32` and `RT_THREAD_STACK_SIZE=4 MiB` unchanged. Full suite `RUST_MIN_STACK=8388608 cargo test -p xerj-engine` → every binary 0 failed, no SIGABRT. Target binary 9 passed (previously 7 of 9 masked). Both non-converted tests still pass. Nit (non-blocking): the 16 MiB test stack is 4× the prod worker, but the two 2 MiB-stack tests still cover the tight budget, so coverage is preserved.

> FAIL-BEFORE: origin/main version + `env -u RUST_MIN_STACK cargo test --test painless_script_limits` → stack overflow, `signal: 6, SIGABRT`, only 2 of 9 tests reported (7 masked). AFTER: a8d7af38 version, same command → 9 passed, 0 failed, clean exit 0, none hidden/ignored/filtered. The abort reproduces pre-fix, is eliminated post-fix, and the after-state hides no tests.

PR #612, opened 2026-08-22
fix(api): omit unresolved hit _seq_no/_version instead of fabricating them (#603)

### LINKED ISSUE #603: top-level hit _seq_no/_version fabricated positionally when snapshot is absent (same anti-pattern as #566)
Surfaced by the #601/#566 verification (correctness + evidence lenses). #566 fixed the **collapse inner_hits** render path to omit an absent snapshot `_seq_no`/`_version` instead of fabricating one. The **top-level (non-collapse) hit** path still has the identical anti-pattern:

- `engine/crates/xerj-api/src/es_compat.rs:13410` — `h.seq_no.unwrap_or(hit_idx as u64)` (a POSITIONAL `_seq_no`)
- `es_compat.rs:13440` — `h.version.or(Some(1))`
- also at ~20084 / ~20417 (scroll / other hit-render paths)

When `Hit.seq_no`/`version` is `None` (concurrent tombstone / version-map miss), these fabricate a positional `_seq_no` and `_version = 1` — values a client can replay into `if_seq_no`/`if_primary_term` and corrupt optimistic concurrency, exactly as #566 described for inner_hits. ES omits versioning it cannot resolve.

Fix: apply the same omit-on-absent treatment (the `emit_inner_hit_versioning` helper, or an equivalent) to the top-level and scroll hit-render paths. Prove fail-before with a hit whose snapshot seq_no is absent → `_seq_no` must be omitted, not positional.

xerj-org:
## What

The top-level and scroll hit-render paths fabricated versioning when the engine could not resolve it — `_seq_no = hit_idx` (a positional index), `_version = 1`, and `_primary_term = 1` even with no real `_seq_no`. A hit's `seq_no`/`version` is `None` on a concurrent tombstone or version-map miss; emitting a made-up value lets a client replay it into `if_seq_no`/`if_primary_term` and **corrupt optimistic concurrency**. ES omits versioning it cannot resolve — exactly what #566 fixed for collapse inner_hits.

Sites: `es_compat.rs` top-level (was `h.seq_no.unwrap_or(hit_idx)` / `h.version.or(Some(1))`), scroll first-page, and scroll continuation — the latter two also emitted `_primary_term: 1` with no real `_seq_no`.

## Fix

Extract `resolve_hit_versioning` — the non-collapse twin of `emit_inner_hit_versioning` (#566): emit `_version`/`_seq_no`/`_primary_term` **only** when resolved, omit otherwise (`EsHit`'s fields are `skip_serializing_if = "Option::is_none"`), and emit the placeholder `_primary_term = 1` only alongside a real `_seq_no`. Applied to the top-level render; the two scroll paths apply the identical omit-on-absent logic inline. The now-dead positional `hit_idx` (and its `.enumerate()`) is dropped from all three closures.

- **Fail-before:** `resolve_hit_versioning(None, None, true, true)` returned `(Some(1), Some(0), Some(1))` under the old logic; the added unit test `absent_top_level_versioning_is_omitted_not_fabricated` pins it to `(None, None, None)` and FAILS against the fabricating form.
- `cargo fmt --check` ✓, `clippy --all-targets -- -D warnings` ✓.
- Full `cargo test -p xerj-api` → 0 failed.

Closes #603.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8342,6 +8342,29 @@ fn emit_inner_hit_versioning(
     }
 }
 
+/// #603: resolve a top-level / scroll hit's wire `(_version, _seq_no,
+/// _primary_term)`, honouring the request's `version` / `seq_no_primary_term`
+/// flags. A value is emitted ONLY when the engine actually resolved it: an
+/// absent `seq_no`/`version` (a concurrent tombstone or version-map miss,
+/// carried as `None` on the `Hit`) is OMITTED, never fabricated into a
+/// positional hit index or a default `1` — values a client could replay into
+/// `if_seq_no`/`if_primary_term` and corrupt optimistic concurrency. The
+/// placeholder `_primary_term = 1` is emitted only alongside a real `_seq_no`.
+/// This is the non-collapse twin of [`emit_inner_hit_versioning`] (#566).
+fn resolve_hit_versioning(
+    seq_no: Option<u64>,
+    version: Option<u64>,
+    emit_seq_no: bool,
+    emit_version: bool,
+) -> (Option<u64>, Option<u64>, Option<u64>) {
+    let out_version = if emit_version { version } else { None };
+    let (out_seq_no, out_primary_term) = match (emit_seq_no, seq_no) {
+        (true, Some(sn)) => (Some(sn), Some(1u64)),
+        _ => (None, None),
+    };
+    (out_version, out_seq_no, out_primary_term)
+}
+
 #[cfg(test)]
 mod inner_hit_versioning_tests {
     use super::*;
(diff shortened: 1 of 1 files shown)


xerj-org:
## Verification — 3 skeptics, verdicts verbatim (HEAD 4e9bcf18). 3/3 NON_BLOCKING.

> All fabrication patterns gone from live code (grep `unwrap_or(hit_idx` / `.or(Some(1))` / `primary_term: if emit_seq_no { Some(1)` → comments/test only). Three paths verified: top-level calls `resolve_hit_versioning`; scroll first-page + continuation inline `version: h.version`, `seq_no: h.seq_no`, `primary_term: h.seq_no.map(|_| 1)` — byte-equivalent. Real values preserved (`(true, Some(sn)) => (Some(sn), Some(1))`). Wire omission confirmed (`EsHit` fields `Option<u64>` + `skip_serializing_if`). Test passes. No correctness hole.

**regression — NON_BLOCKING**
> `git diff --name-only origin/main` = exactly `es_compat.rs`. Full `cargo test -p xerj-api` exited 0, every line `0 failed`; lib unit tests `225 passed; 0 failed`. No pre-existing test asserted a fabricated `_seq_no`/`_version`/`_primary_term` on an unresolved hit (none failed). `.enumerate()`/`hit_idx` fully removed (compile proves index unused). Pre-existing `h.seq_no.unwrap_or(0)` at ~:10936 is an internal PIT-filter comparison (not wire-rendered), out of scope. Regression-free.

> Fail-before: edited the helper to the old logic (`version.or(Some(1))` / `seq_no.unwrap_or(0)`,`Some(1)`) → test FAILED (`left: (Some(1), Some(0), Some(1))` vs `(None,None,None)`). Restored → passes. Test is non-vacuous, exercises the real production fn, covers all five cases (absent/both, real/both, seq_no-only, version-only, flags-off). No residual fabrication.

**Nit (NON_BLOCKING, → follow-up):** the unit test pins the top-level helper directly; the two scroll paths apply byte-equivalent inline logic but aren't separately test-pinned.

PR #615, opened 2026-08-22
fix(engine): score a keyword+text best_fields multi_match by dis_max, not sum (#572)

### LINKED ISSUE #572: multi_match spanning a keyword field scores SUM instead of best_fields MAX (follow-up to #354)

xerj-org:
## What

The #354 rewrite (`rewrite_keyword_full_text_to_term`) splits a keyword-spanning `multi_match` into a whole-value `Term` per keyword field + a `multi_match` over the text fields, and combined them under `Bool.should` (`minimum_should_match=1`) — which scores the **SUM** of its clauses. But the ES `multi_match` default (`best_fields`) is a `dis_max` that takes the **MAX** across fields. So a doc matching both the keyword half and a text half scored `kw + text` instead of `max(kw, text)`. Hit sets were unaffected (pure `_score` error on the previously-broken shape; surfaced by #571 verification, rated NON_BLOCKING).

## Fix (two hunks)

1. `rewrite_keyword_full_text_to_term`: combine the split under `QueryNode::DisMax { tie_breaker: 0.0 }` when `match_type` is `best_fields` (mirrors the all-text per-field combination in `query_node_to_fts_with_keyword_fields`). `most_fields` keeps its `should` (ES sums it); `cross_fields`/phrase are deferred (#572).
2. `query_node_to_fts_with_keyword_fields` had **no `QueryNode::DisMax` arm** — the rewritten `DisMax` declined FTS projection and dropped to the schema-blind stored-source scan, which SUMS (post-flush scored `1.0` = sum, not max). Added a `DisMax` arm that projects each child and combines by MAX, declining only if a child cannot project (so one `DisMax` never mixes FTS + fallback scoring).

- **Fail-before:** with the `Bool.should` combination restored, the new score test sees `2.69` (sum) where it must see `1.0` (max) — asserted **pre- AND post-flush**.
- New test `best_fields_multi_match_scores_keyword_and_text_by_max_not_sum`: asserts `s_both == max(kw, text)` and `s_both < kw + text`, both phases.
- Existing #354 hit-set tests unchanged (dis_max matches iff any clause does, like should+msm=1).
- `fmt --check` ✓, `clippy --all-targets -- -D warnings` ✓, full `cargo test -p xerj-engine` → **1043 passed / 0 failed**.

Closes #572.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -32460,6 +32460,7 @@ fn rewrite_keyword_full_text_to_term(q: &QueryNode, schema: &Schema) -> QueryNod
             fields,
             query,
             boost,
+            match_type,
             ..
         } if fields
             .iter()
@@ -32504,6 +32505,19 @@ fn rewrite_keyword_full_text_to_term(q: &QueryNode, schema: &Schema) -> QueryNod
             }
             if should.len() == 1 {
                 should.into_iter().next().unwrap()
+            } else if matches!(match_type, xerj_query::ast::MultiMatchType::BestFields) {
+                // #572: `best_fields` (the ES default) is a dis_max — the
+                // keyword `Term`(s) and the text `multi_match` combine by MAX
+                // (+ tie_breaker 0.0), not the SUM a `Bool.should` gives. This
+                // mirrors the all-text per-field combination in
+                // `query_node_to_fts_with_keyword_fields`, so a doc matching
+                // both halves scores `max(keyword, text)` as ES does, not the
+                // sum. `most_fields` stays a `should` (ES sums it), and
+                // `cross_fields`/phrase variants are a separate concern (#572).
+                QueryNode::DisMax {
+                    queries: should,
+                    tie_breaker: 0.0,
+                }
             } else {
                 QueryNode::Bool {
                     must: Vec::new(),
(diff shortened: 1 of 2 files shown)


xerj-org:
## Verification — 3 skeptics, verdicts verbatim (HEAD 19df0bbe). 3/3 NON_BLOCKING.

> All 3 tests pass. (1) MAX exactness: `FtsDisMax` score = `(max + tie_breaker*(sum-max))*boost`; new arm passes tie_breaker 0.0 and `DisMaxQuery::new` sets boost 1.0 → exactly max. Both memtable + segment use the same `FtsSearcher`/`execute_dis_max`. (2) Decline is clean: `children.len() != queries.len()` → None for the WHOLE DisMax, never mixes FTS + stored-scan. (3) Scope correct: only `BestFields` takes the DisMax branch; `most_fields`/cross_fields keep `Bool.should` (SUM). (4) Hit sets preserved. Reverting hunk 1 → score test fails pre-flush (2.69 sum vs 1.0 max) while hit-set tests still pass — isolates scoring.

> Both hunks proven load-bearing at HEAD. Hunk 1 (rewrite branch → `&& false`): pre-flush FAILED "must score max=1, got 2.6931472" (the SUM). Hunk 2 (segment DisMax FTS arm neutered): post-flush FAILED "must score max=0.5377, got 1" (stored-source scan sums). Restore → both pass. Test non-vacuous: s_both/s_kw/s_text from real `idx.search` on a doc matching both halves; asserts `|s_both-max|<1e-3` AND `s_both+1e-3<sum`, both phases. Existing #354 hit-set tests still green.

**regression — NON_BLOCKING**
> Diff scope: exactly 2 files, additive only (index.rs +46, test +78), 0 deletions. Full `cargo test -p xerj-engine` (RUST_MIN_STACK=8388608): cargo exit 0, aggregate across 51 binaries = **1043 passed, 0 failed, 23 ignored**. No pre-existing test asserts a SUMMED score for a keyword-spanning best_fields (no regression). All 44 dis_max/phrase/multi_match tests pass (per-field phrase DisMax is a separate path from the new `QueryNode::DisMax` arm). The new arm is conservative (declines if any child fails to project).

**Gate:** 3/3 NON_BLOCKING. Merging on all-checks-green + CLEAN. (Prior CI Format+Clippy failure was a stray rustfmt line, fixed in the amended commit 19df0bbe.)

PR #617, opened 2026-08-22
fix(engine): hybrid total reports Eq at the exact per_query_topk boundary (#594)

### LINKED ISSUE #594: hybrid total: exact-boundary reports gte where eq is possible (fetch per_query_topk+1 to disambiguate)
Follow-up to #569 / PR #593 (the capped→`gte` fix). Two independent verification lenses flagged the same non-blocking precision edge.

## The edge
`fused_count_capped` trips on `sub_result.hits.len() >= per_query_topk`. A sub-leg with **exactly** `per_query_topk` real matches (and none beyond) returns a full page, so the fix reports `relation: gte` even though the fused total is actually exact — where ES would report `eq`. This is the **safe direction** (the reported `value` is still correct and "≥ value" still holds; it never claims a false exact), and it mirrors ES's own inability to tell N from >N without fetching N+1. So it is a mild precision regression vs the old always-`eq`, not a correctness defect — hence deferred here rather than blocking #593.

## Fix
Fetch `per_query_topk + 1` per leg (or consult each leg's exact `total.value` when `track_total_hits` makes it exact) and treat a leg as capped only when it genuinely has **more** than `per_query_topk` matches. Then a leg with exactly `per_query_topk` matches yields `eq`, restoring ES parity at the boundary.

## Not in scope (already correct)
- A kNN leg with a user `k < per_query_topk` returning `k` hits reports `eq` at `k` — this matches ES top-k semantics (a `k`-NN "matches" `k` by contract), not the #569 bug.
- The union `value = fused.len()` is a valid lower bound in the capped case (verified: `fuse_rrf` dedups by id, so it can never exceed the true union).

Filed from the 3-skeptic verification of #593 (correctness + evidence lenses, NON_BLOCKING).

xerj-org:
## What

The #569 capped→`gte` fix tripped `fused_count_capped` on `sub_result.hits.len() >= per_query_topk`. A hybrid sub-leg with **exactly** `per_query_topk` real matches (and none beyond) fills a full page, so the fused total was reported `relation: gte` even though it is exactly known — where ES reports `eq`. The `value` was always correct and "≥ value" held (safe direction), so this was a mild precision regression, deferred from #593 as NON_BLOCKING.

## Fix (index.rs, 3 hunks)

1. Fetch **`per_query_topk + 1`** per leg as a probe.
2. Trip capped only when a leg returned **`> per_query_topk`** hits (the probe found a `(per_query_topk+1)`th match), or its own exact total exceeds the cap, or its own count is already a lower bound.
3. **Trim the probe hit back to `per_query_topk` before fusion**, so the fused ranking and the returned page are byte-identical — only `total.relation` at the exact boundary changes.

A leg with exactly `per_query_topk` matches returns `per_query_topk` hits → not capped → `Eq`, restoring ES parity at the boundary.

- Both hunks load-bearing: without the +1 probe the `>` check would falsely report `Eq` for a genuinely-capped leg (the 256-doc test); without `>` the boundary reports `Gte` (the new 50-doc test).
- **Fail-before:** reverting `>`→`>=` flips the new `hybrid_reports_eq_at_the_exact_per_query_topk_boundary` test to `TotalHits { value: 50, relation: Gte }`.
- #569 tests unchanged (256-doc → Gte, 6-doc → Eq).
- `fmt --all --check` ✓ (exit-checked), `clippy --workspace --all-targets -- -D warnings` ✓, full `cargo test -p xerj-engine` → **1044 passed / 0 failed**.

Closes #594.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -4755,6 +4755,62 @@ mod semantic_deadline_regression_tests {
         assert_eq!(result.total.value, 6, "{:?}", result.total);
     }
 
+    /// #594: a hybrid sub-list with EXACTLY `per_query_topk` matches (and none
+    /// beyond) is exact, not capped: the leg reports `total.value ==
+    /// per_query_topk` with relation `Eq`, so the capped check does not trip and
+    /// the fused total stays `Eq` (ES parity) — where the old
+    /// `hits.len() >= per_query_topk` cap reported a needless `Gte`.
+    /// Fail-before: reverting the capped check `>` → `>=` flips this leg to `Gte`.
+    #[tokio::test]
+    async fn hybrid_reports_eq_at_the_exact_per_query_topk_boundary() {
+        let dir = TempDir::new().unwrap();
+        let engine = engine(&dir);
+        engine
+            .create_index("hybrid-boundary", Schema::empty())
+            .unwrap();
+        let idx = engine.get_index("hybrid-boundary").unwrap();
+        // EXACTLY per_query_topk (default size 10 → 50) docs, identical vectors,
+        // so the kNN leg matches all 50 and reports total.value == 50 (Eq).
+        for n in 0..50 {
+            idx.index_document(
+                Some(format!("d{n}")),
+                serde_json::json!({"embedding": [1.0, 0.0]}),
+            )
+            .await
+            .unwrap();
+        }
+        let knn = QueryNode::Knn {
+            field: "embedding".into(),
+            vector: vec![1.0, 0.0],
+            k: 256, // asks for more than exist → returns all 50, none beyond
+            num_candidates: None,
+            filter: None,
+            boost: None,
+            similarity: None,
+        };
+        let far = std::time::Instant::now() + std::time::Duration::from_secs(30);
+        let result = idx
+            .run_hybrid_with_deadline(
+                &match_all(30_000), // default size 10 → per_query_topk = 50
+                far,
+                vec![xerj_query::ast::WeightedQuery {
+                    query: knn,
+                    weight: 1.0,
+                }],
+                xerj_query::ast::FusionStrategy::Rrf { k: 60 },
+            )
+            .await
+            .unwrap();
+        assert!(!result.timed_out);
+        assert_eq!(
+            result.total.relation,
+            TotalHitsRelation::Eq,
+            "a leg with exactly per_query_topk matches is exact, not capped (#594): {:?}",
+            result.total
+        );
+        assert_eq!(result.total.value, 50, "{:?}", result.total);
+    }
+
     #[tokio::test]
     async fn nested_knn_uses_the_shared_request_deadline() {
         let dir = TempDir::new().unwrap();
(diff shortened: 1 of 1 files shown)


xerj-org:
## Verification — 3 skeptics, verdicts verbatim (HEAD 62303a3a). 3/3 NON_BLOCKING — but reworking on the evidence lens's finding.

> All 3 hybrid_reports tests pass. Ranking/page byte-identical (probe trimmed before fusion). Capped leg still Gte (256→51 hits→`>50`). Boundary correct (exactly-50→`Eq` value 50). Fail-before verified (reverting the hunks → `Gte value 50`). Strict boundary-only precision refinement.

**regression — NON_BLOCKING**
> diff = exactly index.rs (+76/-7). Full suite EXIT=0, 1044 passed / 0 failed. #569 tests unchanged (256→Gte, 6→Eq). `truncate(per_query_topk)` restores the exact pre-change list before fusion → pagination unaffected. One extra hit/leg, trimmed; no measurable regression.

**evidence — NON_BLOCKING (with a finding)**
> Fail-before #1 (core fix) CONFIRMED load-bearing: reverting `>`→`>=` flips the boundary test to `Gte value 50`. Fail-before #2 FALSIFIED — the +1 probe is NOT load-bearing: reverting only `saturating_add(1)`→`per_query_topk` keeps the 256-doc capped test passing (still Gte), because `total.value >= hits.len()` always, so the OR'd `sub_result.total.value > per_query_topk` already trips on a capped leg regardless of the probe. Hunks 1 (+1 fetch) and 3 (trim) are redundant belt-and-suspenders; the commit-message claim overstates them. Behavior is correct; recommend dropping hunks 1+3 or fixing the rationale — does not block merge.

**Action:** the evidence lens is right. Rather than ship redundant code + a misleading claim, I'm **simplifying to the minimal load-bearing change** (fix the boundary over-trigger; rely on the leg's own `total.value`/`relation` for capping — the alternative the issue itself offered) and rewriting the commit rationale to match. Re-verifying the simplified revision.

xerj-org:
## Round-2 verification (SIMPLIFIED revision e2bf5d70) — 3 skeptics, verdicts verbatim. 3/3 NON_BLOCKING.

> Diff = one-line `>=`→`>` on the `hits.len()` arm + the new boundary test; #569 tests untouched. With fetch `size: per_query_topk`, the `hits.len() > per_query_topk` arm is dead (a leg can't return more than requested) — capping rests on `total.value > per_query_topk || relation == Gte`, which is size-independent (FTS/term count all matches into total.value or Gte; kNN bounded by k). No leg undercounts total.value while claiming Eq. All 3 hybrid tests green; fail-before flips ONLY the boundary test to `Gte value 50`. No refutation.

> One changed line; fetch `per_query_topk`, no +1. Fail-before proven on 1.97.1: `>`→`>=` → new test `TotalHits { value: 50, relation: Gte }`; restore → Eq/50. Test non-vacuous (50 docs = per_query_topk, kNN k=256, total.value==50 Eq). #569 capped test (256→Gte) carried by `total.value > per_query_topk`, NOT any probe — confirms total.value is the real signal and the dropped +1 probe was genuinely redundant. Commit rationale is accurate (names total.value load-bearing; probe removed). No remnant anywhere.

**regression — NON_BLOCKING**
> Diff = exactly index.rs (+70/-5). Full `cargo test -p xerj-engine` exit 0: 1044 passed, 0 failed, 23 ignored. #569 tests pass (256→Gte, 6→Eq); new boundary test passes (50→Eq). Fetch `size: per_query_topk`, no `saturating_add(1)`/`truncate` in the run_hybrid region. `total.value >= hits.len()` invariant → any genuinely-capped leg still trips condition 2/3, so relaxing `>=`→`>` only changes the exact boundary. No under-report path introduced.

Issue #602, opened 2026-08-22
_source filter: matches_path diverges from ES for leading/mid globs and dot-boundary (surfaced verifying #599/#310)


xerj-org:
Surfaced by the 3-skeptic verification of #599 (#310 part 2). These are **pre-existing** divergences in `filter_object`/`matches_path` (engine/crates/xerj-engine/src/index.rs ~26371-26490), independent of #599 — they affect **non-empty** objects too and are identical on `main`. #599 only changed the already-empty-object accept decision, so it is not the cause; filing separately so they are tracked.

`matches_path` (and the `keep`/descent predicate) is a prefix/recursion matcher, not a faithful ES `_source` include/exclude automaton. Verified against ES semantics via `filter_object` probes:

1. **Dot-less `.*` prefix boundary.** `{"metadata":{"x":1}}` under include `["meta.*"]` is KEPT; ES returns `{}` — `metadata` is not under `meta.` (the `.ends_with(".*") && starts_with(p[..len-2])` clause uses `starts_with("meta")` with no dot boundary). (The empty-object case `{"metadata":{}}` is now correct via #599's `path_is_accepting`; the non-empty case still leaks here.)
2. **Leading/mid globs don't descend.** `["*.foo"]`, `["a.*.c"]`, `["*",...]`, `["ab*",...]` over nested objects drop matching leaves ES would keep — the descent predicate can't step through a leading/mid glob or expand a glob-accepted parent into its subtree.
3. **`meta*` cross-dot.** `{"meta":{"foo":1}}` under `["meta*"]` drops `meta`; ES keeps it (a trailing `*` after a segment matches across the dot into the subtree in ES's automaton).
4. **`exclude a.*` over `{"a":{}}`.** Returns `{}`; ES keeps `{"a":{}}` — the exclude `.*` clause marks bare `a` excluded.

Impact: `_source` include/exclude with glob patterns can return an ES-divergent projection for nested/dotted fields. Fix direction: replace the ad-hoc `matches_path` glob handling with a proper ES-style include/exclude automaton (or a faithful segment-wise matcher with correct accept/live-state and dot-boundary semantics), shared by the keep, exclude, and accept decisions.

PR #619, opened 2026-08-22
fix(engine): _source `prefix.*` include requires the dot boundary (part of #602)

xerj-org:
## What

`matches_path`'s subtree clause for a `prefix.*` include kept the glob's subtree by testing `path.starts_with(&p[..p.len() - 2])` — stripping `.*` and leaving `prefix` with **no dot boundary**. So `{"metadata":{"x":1}}` under include `["meta.*"]` was **kept** (because `"metadata".starts_with("meta")`), where ES returns `{}` (a `meta.*` include allow-lists only paths under `meta.`, not the prefix-sharing sibling `metadata`).

The empty-object case was already correct via #310/#599's `path_is_accepting`; this is the **non-empty** leak through `matches_path` — divergence 1 of #602.

## Fix

Strip only the trailing `*` (`p[..p.len() - 1]` = `prefix.`), so the subtree clause requires the dot: `"metadata".starts_with("meta.")` is false → dropped, while `meta.foo` / `meta.foo.bar` (the real subtree) still match (`glob_match` handles the single segment; this clause carries the deeper subtree).

- **Fail-before:** reverting `p.len() - 1` → `p.len() - 2` makes the new test `source_glob_prefix_requires_the_dot_boundary_for_nonempty_siblings` report `{"metadata":{"x":1}}` instead of `{}`.
- New test also pins the positive controls (subtree kept; a prefix-sharing sibling next to a real match drops only the sibling).
- #310/#599 filter tests unchanged.
- `fmt --all --check` ✓ (exit-checked), `clippy --all-targets -- -D warnings` ✓, full `cargo test -p xerj-engine` → **1045 passed / 0 failed**.

## Scope

This fixes **divergence 1 only**. #602's other divergences (leading/mid globs not descending, `meta*` cross-dot, `exclude a.*` over `{"a":{}}`) require the faithful ES include/exclude automaton the issue describes and are left open. **Refs #602** (does not close it).

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -26533,7 +26533,14 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
         }
         patterns.iter().any(|p| {
             if p.contains('*') {
-                glob_match(p, path) || (p.ends_with(".*") && path.starts_with(&p[..p.len() - 2]))
+                // A `prefix.*` include also keeps the whole subtree under
+                // `prefix.` (`glob_match`'s `*` cannot cross a dot, so it alone
+                // would drop `prefix.a.b`). #602: this subtree check must keep
+                // the dot boundary — strip only the trailing `*` (`prefix.`),
+                // NOT `.*` (`prefix`). Otherwise `meta.*` matched `metadata`
+                // via `starts_with("meta")`, where ES keeps only paths under
+                // `meta.` and returns `{}` for a sibling like `metadata`.
+                glob_match(p, path) || (p.ends_with(".*") && path.starts_with(&p[..p.len() - 1]))
             } else {
                 path == p || path.starts_with(&format!("{}.", p))
             }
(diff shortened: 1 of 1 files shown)


xerj-org:
## Verification — 3 skeptics, verdicts verbatim (verified at bc53b206). 3/3 NON_BLOCKING.

> 10/10 `source_filter_tests` pass. `meta.*` keeps `meta.foo` (glob_match) AND `meta.foo.bar` (subtree clause `starts_with("meta.")`) — not over-pruned. `metadata`/`metadata.x` now dropped (`starts_with("meta.")` false). No panic risk: the `p[..len-1]` slice is guarded by `ends_with(".*")` (len≥2); bare `*` short-circuits; `.*` slices to `.` (matches nothing). Fail-before confirmed (revert → `{"metadata":{"x":1}}`). Scope honestly partial. No correctness hole.

**regression — NON_BLOCKING**
> Diff = exactly index.rs. Full `cargo test -p xerj-engine` → **1045 passed, 0 failed**. All existing `_source` filter / `matches_path` tests pass. Exclude (shared `matches_path`): the change also makes exclude `meta.*` require the dot (no longer over-excludes `metadata`) — toward ES parity, not away; no test regresses. `path_is_accepting`'s `.*` branch is independent and unaffected. CLA-compliant. Nit: a comment near line 26596 still cited the old `p[..len-2]` expression.

**evidence — NON_BLOCKING**
> Fail-before reproduced (`left: {"metadata":{"x":1}}` vs `{}`); restore → PASS. Test non-vacuous (three real `filter_object` calls). Scope honest: `Refs #602` (not Closes), PR enumerates the remaining open divergences.

**Follow-up applied (65f52c7c, comment-only — behavior byte-identical to the verified bc53b206):** the regression lens's NON_BLOCKING nit is fixed — the two comments that described `matches_path`'s old `p[..len-2]` `.*` clause (index.rs ~26592 and the `meta.*` test ~26763) now accurately describe the post-fix `p[..len-1]` (dot-boundary) behaviour, per "re-read the whole paragraph you changed". No code change, so the verdicts above stand.

PR #622, opened 2026-08-22
fix(api): render collapse inner_hits on scroll continuation pages (#621)

### LINKED ISSUE #621: collapse + scroll: continuation pages leak __xy_collapse_group__/__xy_collapse_spec__ into _source and never render inner_hits
Found while writing the #566 items-2/3 scroll coverage. It is worse than a coverage gap: the collapse render is **entirely missing** on scroll continuation pages.

## Mechanism

`search_impl` renders a collapse group into `inner_hits` and strips the internal sentinels from `_source` (es_compat.rs ~13238-13380: the `__xy_collapse_group__` block that calls `emit_inner_hit_versioning` and `src.remove("__xy_collapse_group__")`). `scroll_page_response` (es_compat.rs ~20404) has **no equivalent** — it emits the snapshotted leader hit verbatim. So a `POST /{index}/_search?scroll=` with `collapse` returns, on every continuation page:

- the internal `__xy_collapse_group__` array **and** `__xy_collapse_spec__` object **leaked into the wire `_source`**, and
- **no `inner_hits`** key at all.

Page 1 (from `search_impl`) is correct; only the continuation pages are broken.

## Reproduction (fails on main)

A collapse scroll over two groups (page_size 1 → group B lands on page 2), continuing to page 2:

```
page2.hits.hits[0]._source == {
  "grp":"B","v":20,
  "__xy_collapse_group__":[{"_id":"b0",...,"_seq_no":2,"_version":1},
                           {"_id":"b1","_source":{"grp":"B","v":21},"_seq_no":3,"_version":1}],
  "__xy_collapse_spec__":{"name":"members","size":10,...}
}
```

ES would return `_source: {"grp":"B","v":20}` plus a rendered `inner_hits.members.hits.hits` — never the internal sentinels.

Good news: the snapshot IS carried correctly — `__xy_collapse_group__` shows `b1` at `_seq_no:3, _version:1` (the pre-update snapshot) even after a concurrent `PUT` bumped it to `_version:2`. So the #566 item-1 snapshot fix holds on the scroll path too; the defect is purely the **missing render** in `scroll_page_response`.

A runnable reproduction lives on branch `test/566-collapse-scroll-snapshot-version` (`collapse_scroll_snapshot_version_torn_read.rs`, `#[ignore]`d), asserting the desired behavior (sentinels stripped, `inner_hits` rendered, member `_version` = snapshot 1).

## Fix direction

Factor `search_impl`'s collapse-render block (es_compat.rs ~13238-13380 — render `inner_hits` via `emit_inner_hit_versioning`, honour the group's `sort`/`size`/`fields`/`_source` spec, and `remove` both sentinels) into a shared helper, and call it from `scroll_page_response` too. Impact: any collapse query read over a scroll (the `_bulk`/reference-coding pattern that pages results) returns internal fields to clients and no inner hits. Closes the scroll half of #566 (items 2 & 3).

xerj-org:
## What

`scroll_page_response` never rendered a collapse leader's `inner_hits`. On a
collapse `_search?scroll=...`, the **first** page is served by `search_impl`
(which renders `inner_hits` from the `__xy_collapse_group__` /
`__xy_collapse_spec__` sentinels `apply_collapse` plants in `_source`, then
strips them). Every **continuation** page, though, went through
`scroll_page_response`, which inserted the leader's `_source` verbatim —
sentinels and all — and always set `inner_hits: None`. So clients paging a
collapsed scroll saw internal sentinels leak into the wire `_source` and got
no `inner_hits` at all. That is a wire-correctness break, not just a coverage
gap (#621; the scroll half of #566 items 2 & 3).

The snapshot `_seq_no`/`_version` were already carried correctly per member in
the group sentinel (captured at scroll-open, never re-read live), so the only
defect was the missing render on the scroll path.

## How

- Extract `search_impl`'s collapse-render block into a shared free function
  `render_collapse_inner_hits(group, spec, idx_name, rest_total_hits_as_int)`.
  Behavior-identical: the only rewiring is `idx_name: &str` and the
  `rest_total_hits_as_int` bool in place of the `params` field. `search_impl`
  now calls it; its output is byte-identical to before.
- `scroll_page_response` reads the sentinels off each leader's `_source`,
  removes both, and calls the shared helper — so continuation pages render
  `inner_hits` identically to page 1 (size/from/sort/collapse/fields + the
  snapshot version pair).

## Test / fail-before

`collapse_scroll_snapshot_version_torn_read` opens a collapse scroll (one
group per page, `size:1`, `sort grp asc`, inner_hits with seq_no+version),
bumps group-B member `b1`'s `_version` **after** page 1 but **before** the
page-2 continuation, then asserts page 2:
1. has **no** `__xy_collapse_group__` / `__xy_collapse_spec__` in `_source`,
2. renders `inner_hits.members`, and
3. reports `b1`'s **snapshot** `_version` (1) beside its snapshot `_source`
   (`v=21`) — not the live post-update value (a torn read a client could
   replay into `if_seq_no`).

**Fail-before:** reverting *only* the `scroll_page_response` hunk (helper +
`search_impl` call site intact) fails the test on assertion (1) — the wire
`_source` shows the leaked `__xy_collapse_group__` / `__xy_collapse_spec__`.
With the hunk, it passes.

Full `cargo test -p xerj-api` green under rustc 1.97.1 (both dev and ci-test).
fmt `--all --check` ✓, clippy `--all-targets -D warnings` ✓.

## Scope

Renders `inner_hits` when the leader `_source` is present (the realistic
collapse-scroll case). A collapse scroll under `_source: false` is a niche
follow-up, not addressed here.

Closes #621. Closes the scroll half of #566 (items 2 & 3).

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8443,6 +8443,235 @@ mod inner_hit_versioning_tests {
     }
 }
 
+/// Render a collapse leader's `inner_hits` from the `__xy_collapse_group__`
+/// / `__xy_collapse_spec__` sentinels `apply_collapse` planted in its
+/// `_source`, honoring each spec's `size`/`from`/`sort`/`collapse`/`fields`
+/// and the snapshot `_version`/`_seq_no` carried per member (never a
+/// render-time live lookup — #506/#566). Shared by `search_impl` and the
+/// scroll continuation (`scroll_page_response`) so a collapsed scroll renders
+/// inner_hits identically instead of leaking the sentinels into `_source`
+/// (#621).
+fn render_collapse_inner_hits(
+    group: Vec<Value>,
+    spec: Value,
+    idx_name: &str,
+    rest_total_hits_as_int: bool,
+) -> Value {
+    let spec_list: Vec<Value> = match spec {
+        Value::Array(a) => a,
+        other => vec![other],
+    };
+    let mut combined = serde_json::Map::new();
+    for spec in &spec_list {
+        let name = spec
+            .get("name")
+            .and_then(Value::as_str)
+            .unwrap_or("inner_hits")
+            .to_string();
+        let size = spec.get("size").and_then(Value::as_u64).unwrap_or(3) as usize;
+        let sort_spec = spec.get("sort").cloned().unwrap_or(Value::Null);
+        let from_spec = spec.get("from").and_then(Value::as_u64).unwrap_or(0) as usize;
+        let mut members = group.clone();
+        if let Value::Array(sort_arr) = &sort_spec {
+            for s in sort_arr.iter().rev() {
+                if let Some(obj) = s.as_object() {
+                    for (field, opts) in obj {
+                        let desc = match opts {
+                            Value::String(s) => s == "desc",
+                            Value::Object(o) => o
+                                .get("order")
+                                .and_then(Value::as_str)
+                                .map(|v| v == "desc")
+                                .unwrap_or(false),
+                            _ => false,
+                        };
+                        let f = field.clone();
+                        // Compare numerically when both sides
+                        // parse as numbers, else fall back to a
+                        // lexicographic string compare so keyword
+                        // and date fields sort correctly. Ties
+                        // break on `_id` so the result is
+                        // deterministic across runs and matches
+                        // ES's `_doc` secondary sort for
+                        // monotonically-assigned ids.
+                        members.sort_by(|a, b| {
+                            let av = a.get("_source").and_then(|s| s.get(&f));
+                            let bv = b.get("_source").and_then(|s| s.get(&f));
+                            let av_n = av.and_then(Value::as_f64);
+                            let bv_n = bv.and_then(Value::as_f64);
+                            let primary = match (av_n, bv_n) {
+                                (Some(x), Some(y)) => {
+                                    x.partial_cmp(&y).unwrap_or(std::cmp::Ordering::Equal)
+                                }
+                                _ => {
+                                    let to_str = |v: Option<&Value>| match v {
+                                        Some(Value::String(s)) => Some(s.clone()),
+                                        Some(other) if !other.is_null() => Some(other.to_string()),
+                                        _ => None,
+                                    };
+                                    let av_s = to_str(av);
+                                    let bv_s = to_str(bv);
+                                    match (av_s, bv_s) {
+                                        (Some(x), Some(y)) => x.cmp(&y),
+                                        (Some(_), None) => std::cmp::Ordering::Less,
+                                        (None, Some(_)) => std::cmp::Ordering::Greater,
+                                        (None, None) => std::cmp::Ordering::Equal,
+                                    }
+                                }
+                            };
+                            let primary = if desc { primary.reverse() } else { primary };
+                            if primary != std::cmp::Ordering::Equal {
+                                return primary;
+                            }
+                            let aid = a.get("_id").and_then(Value::as_str).unwrap_or("");
+                            let bid = b.get("_id").and_then(Value::as_str).unwrap_or("");
+                            aid.cmp(bid)
+                        });
+                    }
+                }
+            }
+        }
+        let emit_seq_no_ih = spec
+            .get("seq_no_primary_term")
+            .and_then(Value::as_bool)
+            .unwrap_or(false);
+        let emit_version_ih = spec
+            .get("version")
+            .and_then(Value::as_bool)
+            .unwrap_or(false);
+        // `inner_hits.collapse.field` — second-level collapse
+        // applied to the group members. ES counts `total` against
+        // the pre-collapse member set but the returned `hits` are
+        // deduped by the inner collapse field, preserving sort
+        // order. (See ES 8.13 multi-level collapse.)
+        let total = members.len();
+        if let Some(inner_field) = spec
+            .get("collapse")
+            .and_then(|c| c.get("field"))
+            .and_then(Value::as_str)
+        {
+            let mut seen: std::collections::HashSet<String> = std::collections::HashSet::new();
+            members.retain(|m| {
+                let key = m
+                    .get("_source")
+                    .and_then(|s| s.get(inner_field))
+                    .map(|v| match v {
+                        Value::String(s) => s.clone(),
+                        Value::Number(n) => n.to_string(),
+                        Value::Bool(b) => b.to_string(),
+                        Value::Null => "\0null".to_string(),
+                        other => other.to_string(),
+                    })
+                    .unwrap_or_else(|| "\0missing".to_string());
+                seen.insert(key)
+            });
+        }
+        let rendered_hits: Vec<Value> = members
+            .into_iter()
+            .skip(from_spec)
+            .take(size)
+            .map(|mut m| {
+                if let Some(src) = m.get_mut("_source").and_then(Value::as_object_mut) {
+                    src.remove("__xy_collapse_group__");
+                    src.remove("__xy_collapse_spec__");
+                }
+                let mut hit_obj = serde_json::Map::new();
+                hit_obj.insert("_index".to_string(), Value::String(idx_name.to_string()));
+                if let Some(id) = m.get("_id").cloned() {
+                    hit_obj.insert("_id".to_string(), id);
+                }
+                if let Some(score) = m.get("_score").cloned() {
+                    hit_obj.insert("_score".to_string(), score);
+                }
+                // Snapshotted `_version` / `_seq_no` carried in the
+                // collapse group (#506) — NOT a render-time live
+                // lookup, which could pair a live version with the
+                // snapshot `_source` on a collapsed scroll. An absent
+                // snapshot value is OMITTED, not fabricated (#566).
+                emit_inner_hit_versioning(&m, emit_seq_no_ih, emit_version_ih, &mut hit_obj);
+                if let Some(src) = m.get("_source").cloned() {
+                    hit_obj.insert("_source".to_string(), src);
+                }
+                // inner_hits `fields` — emit the requested
+                // source fields as a `fields: {name: [values]}`
+                // map matching top-level hit semantics.
+                let mut fmap = serde_json::Map::new();
+                if let Some(ih_fields) = spec.get("fields").and_then(Value::as_array) {
+                    for f in ih_fields {
+                        let fname = match f {
+                            Value::String(s) => s.as_str(),
+                            Value::Object(o) => {
+                                o.get("field").and_then(Value::as_str).unwrap_or("")
+                            }
+                            _ => continue,
+                        };
+                        if fname.is_empty() {
+                            continue;
+                        }
+                        if let Some(src) = m.get("_source") {
+                            if let Some(v) = src.get(fname).cloned() {
+                                let wrapped = match v {
+                                    Value::Array(a) => Value::Array(a),
+                                    Value::Null => continue,
+                                    other => Value::Array(vec![other]),
+                                };
+                                fmap.insert(fname.to_string(), wrapped);
+                            }
+                        }
+                    }
+                }
+                // ES auto-emits the inner-collapse field into
+                // `fields` even without an explicit `fields`
+                // clause, mirroring the top-level collapse
+                // behavior (see line ~7070).
+                if let Some(inner_field) = spec
+                    .get("collapse")
+                    .and_then(|c| c.get("field"))
+                    .and_then(Value::as_str)
+                {
+                    if !fmap.contains_key(inner_field) {
+                        if let Some(v) = m.get("_source").and_then(|s| s.get(inner_field)).cloned()
+                        {
+                            let wrapped = match v {
+                                Value::Array(a) => Value::Array(a),
+                                Value::Null => Value::Array(vec![]),
+                                other => Value::Array(vec![other]),
+                            };
+                            if !matches!(&wrapped, Value::Array(a) if a.is_empty()) {
+                                fmap.insert(inner_field.to_string(), wrapped);
+                            }
+                        }
+                    }
+                }
+                if !fmap.is_empty() {
+                    hit_obj.insert("fields".to_string(), Value::Object(fmap));
+                }
+                Value::Object(hit_obj)
+            })
+            .collect();
+        // When the caller set `rest_total_hits_as_int: true`
+        // on the outer request, inner_hits totals become
+        // bare numbers — mirror the top-level hits.total
+        // shape.
+        let inner_total = if rest_total_hits_as_int {
+            json!(total as u64)
+        } else {
+            json!({ "value": total as u64, "relation": "eq" })
+        };
+        combined.insert(
+            name,
+            serde_json::json!({
+                "hits": {
+                    "total": inner_total,
+                    "max_score": Value::Null,
+                    "hits": rendered_hits,
+                }
+            }),
+        );
+    }
+    Value::Object(combined)
+}
+
 async fn search_impl(
     state: AppState,
     index: String,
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent verification — 3-skeptic panel (rustc 1.97.1, isolated worktrees, each told to REFUTE)

Posting verdicts verbatim. **2 of 3 complete, both NON_BLOCKING**; the regression lens is running its full both-profile suite and will be appended. Merge gate is **0 BLOCKING**, not 0 findings — the non-blocking notes below become follow-up issues, they do not block.

<details><summary><b>Lens 1 — CORRECTNESS (behavioral identity of the extraction + scroll render) → NON_BLOCKING</b></summary>

```
VERDICT: NON_BLOCKING

PR #622, branch fix/621-collapse-scroll-render, commit a4423716
File under review: engine/crates/xerj-api/src/es_compat.rs
Toolchain: rustup run 1.97.1, RUST_MIN_STACK=8388608, workspace engine/

CLAIM TO REFUTE: "The scroll-path render is behaviorally identical to the
first-page (search_impl) collapse render, and the search_impl extraction did
not change search_impl's output."

RESULT: Could not refute. The claim holds.

--------------------------------------------------------------------------
1. EXTRACTION IS BEHAVIOR-IDENTICAL (search_impl output unchanged)
--------------------------------------------------------------------------
Read the full diff (git show a4423716 -- es_compat.rs). The removed inline
block (old lines ~13469-13693) and the new free fn
render_collapse_inner_hits (new es_compat.rs:8449-8672) are identical
line-for-line EXCEPT the two documented rewirings, and formatting:
  - es_compat.rs:8579  Value::String(idx_name.to_string())  -- was
    idx_name.clone() because in search_impl idx_name is an owned String
    (from the merged_hits tuple, es_compat.rs:11799) and in the helper it
    is idx_name: &str. Same bytes on the wire.
  - es_compat.rs:8656  `if rest_total_hits_as_int` -- was
    `params.rest_total_hits_as_int.as_deref() == Some("true")`; the exact
    same expression is now evaluated at the search_impl call site
    (es_compat.rs:13476) and passed in. Same value.
No dropped line, no moved `?`, no stray params./body. reference inside the
block; the only externally-captured names were idx_name and
params.rest_total_hits_as_int, both rewired. emit_inner_hit_versioning is a
module-level fn used unchanged by both sites.
search_impl call site: es_compat.rs:13471-13478 (group/spec passed by value,
same as the old `if let (Some(group), Some(spec))`).
Confirmation that search_impl output is unchanged: existing page-1 test
collapse_inner_hits_seqno_is_real passes on the fix.

--------------------------------------------------------------------------
2. SCROLL CALL SITE (es_compat.rs:20568-20598) IS CORRECT
--------------------------------------------------------------------------
  - NON-collapse hit: obj.get("__xy_collapse_group__") -> None,
    __xy_collapse_spec__ -> None, remove() are no-ops, match (None,_)/(_,None)
    -> None. So collapse_inner=None: no panic, no inner_hits added, _source
    still inserted (unchanged). Verified by scroll_multi_index_identity and
    scroll_seq_no_torn_read passing (non-collapse scrolls).
  - No double-insert of _source: exactly one `o.insert("_source",src)`
    (20594), vs the old single `o.insert("_source",src.clone())` (removed).
  - ctx.hits stays intact across pages: the source is cloned twice before
    mutation -- ctx.hits[i].source -> page_hits[i].source (clone at
    es_compat.rs:20509) -> local `let mut src = src.clone()` (20575). The
    sentinel removal happens on the local clone only. ctx.hits is never
    touched, so re-scrolling / reuse across pages is safe.

Premise verified: the scroll snapshot actually carries the sentinels.
scroll_snapshot = Some(merged_hits.clone()) (es_compat.rs:11666), and the
_source stripping happens on a SEPARATE clone `s` (es_compat.rs:11871-11876),
so merged_hits (and thus ctx.hits) retains __xy_collapse_group__ /
__xy_collapse_spec__. Page 1 (search_impl) and the scroll snapshot both start
from the SAME merged_hits, so the helper is fed identical group/spec.

--------------------------------------------------------------------------
3. NO DIVERGENCE FOUND (page 1 search_impl vs page 2 scroll, same group)
--------------------------------------------------------------------------
Wrote a focused test that renders the SAME collapse group ("B") two ways and
asserts the full inner_hits JSON is byte-identical:
  - via search_impl: plain POST /docs/_search size 10 (group B is 2nd leader)
  - via scroll:      size:1 collapse scroll, page1=A (search_impl),
                     page2=B (scroll_page_response)
Exercised simultaneously: multi-spec inner_hits ARRAY, inner-level
collapse.field, fields, from/size, sort desc, version + seq_no_primary_term,
and rest_total_hits_as_int (ran BOTH false and true).

  rustup run 1.97.1 cargo test -p xerj-api --test zzz_collapse_scroll_matches_search_impl
  running 2 tests
  test scroll_collapse_render_matches_search_impl_total_as_int ... ok
  test scroll_collapse_render_matches_search_impl_default ... ok
  test result: ok. 2 passed; 0 failed

FAIL-BEFORE (proves the test has teeth): reverting ONLY the
scroll_page_response hunk to the old `o.insert("_source", src.clone())`
one-liner makes BOTH cases FAIL --
  "scroll continuation leaked collapse sentinels ... __xy_collapse_group__ ...
   __xy_collapse_spec__ ..." and no inner_hits.
Restored the committed source (git checkout --); worktree clean; test file
removed after use.

--------------------------------------------------------------------------
4. PR TEST + EXISTING SUITE
--------------------------------------------------------------------------
  cargo test -p xerj-api --test collapse_scroll_snapshot_version_torn_read
    test collapse_scroll_inner_hit_reports_snapshot_version_not_live ... ok  (1 passed)
  --test collapse_inner_hits_seqno_is_real
    test collapse_inner_hits_seq_no_is_the_real_snapshot_value ... ok  (1 passed)
  --test scroll_seq_no_torn_read ... 3 passed
  --test scroll_multi_index_identity ... 2 passed

--------------------------------------------------------------------------
NON-BLOCKING NOTES / FOLLOW-UPS (not defects)
--------------------------------------------------------------------------
  a) rest_total_hits_as_int is read per-request on BOTH paths: search_impl
     from the search request's params, scroll from the continuation request's
     params (scroll_page_response arg, es_compat.rs:20343/20371/20424). A
     client that opens a collapse scroll with rest_total_hits_as_int=true but
     continues WITHOUT it would see inner totals as {value,relation} on page 2
     vs a bare int on page 1. This matches ES (the flag is per-request, not
     remembered across scroll continuations) and mirrors the top-level total
     handling at es_compat.rs:20629. Client responsibility, not a PR defect.
  b) Documented scope: the scroll render only runs inside
     `if let Some(src) = &h.source`, so a collapse scroll under _source:false
     renders no inner_hits. The PR explicitly scopes this out as a follow-up;
     it is a pre-existing gap, not a regression.
  c) Recommend upstreaming a page1-vs-page2 byte-identity test like the one
     above; the PR's own test only checks page 2 in isolation.
```
</details>

<details><summary><b>Lens 2 — EVIDENCE (fail-before isolation + test non-vacuity) → NON_BLOCKING</b></summary>

```
VERDICT: NON_BLOCKING

PR #622 / branch fix/621-collapse-scroll-render, commit a4423716
Test: engine/crates/xerj-api/tests/collapse_scroll_snapshot_version_torn_read.rs
      collapse_scroll_inner_hit_reports_snapshot_version_not_live
Toolchain: rustc 1.97.1, RUST_MIN_STACK=8388608

(Worktree note: the worktree was NOT pre-checked-out on the fix branch as the
prompt stated — it was on a detached worktree branch at 7d836bb2. fix/621 was
already checked out in another worktree, so I `git reset --hard`ed this isolated
worktree to a4423716 (== origin/fix/621-collapse-scroll-render). Same tree.)

=====================================================================
1. FAIL-BEFORE (revert ONLY the scroll_page_response hunk)
=====================================================================
Baseline (unmodified branch): test PASSES (1 passed).

Reverted ONLY the scroll-path collapse block ("A collapse leader carries...")
back to the original three lines:
    if let Some(src) = &h.source {
        o.insert("_source".to_string(), src.clone());
    }
render_collapse_inner_hits helper and the search_impl call site LEFT INTACT.
Build succeeded (only xerj-api recompiled, no errors — the helper is still
referenced by search_impl).

Result: FAILED. The FIRST and only firing assertion is the SENTINEL LEAK
(assertion 1), verbatim:

  panicked at tests/collapse_scroll_snapshot_version_torn_read.rs:149:5:
  collapse scroll continuation leaked internal sentinels into _source —
  scroll_page_response does not render the collapse group like search_impl:
  {"_index":"docs","_id":"b0","_score":1.0,"_source":{"grp":"B","v":20,
   "__xy_collapse_group__":[
     {"_id":"b0","_score":1.0,"_source":{"grp":"B","v":20},"sort":["B"],"_seq_no":2,"_version":1},
     {"_id":"b1","_score":1.0,"_source":{"grp":"B","v":21},"sort":["B"],"_seq_no":3,"_version":1}],
   "__xy_collapse_spec__":{"name":"members","size":10,"sort":[{"v":"asc"}],
   "seq_no_primary_term":true,"version":true}},"sort":["B"]}

This IS the stated reason (sentinel leak / assertion 1). Restored with
`git checkout -- .../es_compat.rs` (git diff --stat clean) and re-ran:
test PASSES again (1 passed). Fail-before ISOLATES the scroll-render fix. Claim A HOLDS.

=====================================================================
2. VACUITY of assertion (3): b1._version == 1  (snapshot vs live)
=====================================================================
The fail-before panic dump is decisive evidence: the leaked
`__xy_collapse_group__` array carries b1 FROZEN as {_source:{grp:B,v:21},
_seq_no:3,_version:1} — i.e. the pre-update snapshot, captured at scroll-open
and stored in the scroll context. The live PUT (v:999 -> version 2) never
touches this stored array.

render_collapse_inner_hits -> emit_inner_hit_versioning (es_compat.rs:8326)
reads `member.get("_version")` DIRECTLY off this frozen array. There is NO
render-time live lookup anywhere on the collapse render path. Therefore:

  - The code as written can ONLY ever emit the snapshot (_version 1, v 21).
    A live lookup would emit 2 / 999. So assertion (3) is TAUTOLOGICAL with
    respect to the CURRENT architecture — it cannot fail given today's code.
  - BUT it is a genuine, non-trivial REGRESSION GUARD: it locks the snapshot
    semantics onto the scroll path so that any future change swapping in a
    live version lookup (the #499/#506/#566 torn-read hazard) is caught. The
    snapshot(1,21) vs live(2,999) values are well-separated, so the guard
    truly discriminates. `_version 1` is a real stored read, not a fabricated
    default — proved by the unit test absent_snapshot_versioning_is_omitted_
    not_fabricated (absent => omitted, never defaulted to 1).

Honest caveat (skeptic's finding, NON-blocking): assertion (3) is NOT what
makes the fail-before work and is NOT load-bearing for the #621 defect. The
#621 defect was purely the missing render + un-stripped sentinels; the commit
message itself states the snapshot values "were carried correctly all along."
The load-bearing assertions for THIS fix are (1) no sentinel leak and (2) b1
present in inner_hits.members. Assertion (3) guards a related, pre-existing
property. The test title/docstring lean on the torn-read framing, which is a
mild over-emphasis, but the assertion is a valid lock-in guard, not vacuous in
any harmful way. Not a reason to block.

=====================================================================
3. SETUP SOUNDNESS
=====================================================================
a) PUT /docs/_doc/b1 bumps _version >= 2: YES. b1 created (version 1), PUT
   overwrites -> version 2. Test asserts live_version_after >= 2; held in the
   passing baseline (and the fail-before got PAST it to line 149, so the update
   path executes). The frozen sentinel still shows _version:1 for b1, confirming
   snapshot != live. Sound.

b) Page 2 renders AFTER the update: YES. Order is read-page-1 -> assert grp A
   -> PUT b1 -> POST /_search/scroll (page 2). The continuation is fetched after
   the PUT, yet returns snapshot values. Sound.

c) Two groups can never land on the same page: size:1 returns exactly one
   collapse LEADER per page; scroll paginates by leader (confirmed — the
   fail-before dump shows page 2 leader = b0 with the whole B group [b0,b1] in
   its sentinel, group A entirely on page 1). sort grp asc makes A (< B) the
   page-1 leader deterministically. size:1 + sort grp asc IS sufficient to
   guarantee group A on page 1, group B on page 2. Sound.

d) No false-pass path (item 4): the inner-hit lookup is NOT loose —
   pointer(...).unwrap_or_default() yields [] if inner_hits absent/empty, then
   .find(_id=="b1").unwrap_or_else(panic!) fires "b1 must be a group-B inner
   hit". So empty/absent inner_hits => panic; a strip-only partial fix still
   fails assertion 2; a render-without-strip partial fix still fails assertion
   1. Both halves required. b1's _source.v==21 AND _version==1 rule out the live
   (999 / version 2) values. Assertion 1 is evaluated only after grp=="B" is
   asserted (a missing _source would fail the grp check first), so it cannot
   pass spuriously on an absent _source. No path passes without the real fix.

=====================================================================
CONCLUSION
=====================================================================
Attempted to refute claims A and B; could not. Fail-before precisely isolates
the scroll-render hunk and fails for the stated sentinel-leak reason. The test
is sound and non-vacuous on its load-bearing assertions (1 & 2). Assertion (3)
is a tautological-but-legitimate snapshot lock-in guard (not load-bearing for
the specific fix, faithfully described as such by the commit). No blocking flaw.
VERDICT: NON_BLOCKING.
```
</details>

### Non-blocking follow-ups the panel surfaced (to be filed as issues, not merge blockers)
1. **Upstream a page1-vs-page2 byte-identity test** (correctness lens wrote one that passed across multi-spec/inner-collapse/fields/from/size/sort/version/`rest_total_hits_as_int`) — a stronger guard than the current page-2-in-isolation test.
2. **Test framing note**: the snapshot-`_version` assertion is a valid lock-in guard but tautological w.r.t. today's architecture (no live-lookup path exists); the load-bearing assertions for #621 are the sentinel-strip + inner_hits-render ones.
3. **Scope gap (pre-existing, deferred)**: a collapse scroll under `_source: false` renders no inner_hits (guarded by `if let Some(src) = &h.source`).

xerj-org:
### Lens 3 — REGRESSION (full both-profile suite) → NON_BLOCKING

```
VERDICT: NON_BLOCKING

PR #622 / branch fix/621-collapse-scroll-render — REGRESSION skeptic review.
Toolchain: rustc 1.97.1 (8bab26f4f 2026-07-14). RUST_MIN_STACK=8388608. Workspace: engine/.

=== WORKTREE PROVENANCE CAVEAT (important) ===
The handed worktree (agent-ace14894e513f7ed9) was NOT actually checked out on the fix
branch. Its HEAD was 7d836bb2 (the #619 merge, i.e. the PRE-FIX base) and the claimed
function `render_collapse_inner_hits` was ABSENT from es_compat.rs. Testing it as-handed
would have reviewed the baseline, not the change.
Remediation: since the tree was clean, I detached-checkout the actual fix tip
a4423716 (tip of fix/621-collapse-scroll-render) into this worktree and ran everything
against THAT. All results below are from commit a4423716. Function confirmed present
(es_compat.rs:8454), search_impl at :8675, scroll_page_response at :20420.

=== 1. DEV profile: cargo test -p xerj-api ===
EXIT=0
51 test binaries, ALL "test result: ok"
Totals: 488 passed; 0 failed; 2 ignored (pre-existing ignores; not the new test)

=== 2. CI profile: cargo test -p xerj-api --profile ci-test ===
EXIT=0
51 test binaries, ALL "test result: ok"
Totals: 488 passed; 0 failed; 2 ignored

=== 3. clippy -p xerj-api --all-targets -- -D warnings ===
EXIT=0, "Finished" clean, zero warnings.
cargo fmt --all --check: FMT=0 (clean).

=== 4. Collapse / inner_hits / scroll guard tests (all PASS both profiles) ===
- collapse_inner_hits_seq_no_is_the_real_snapshot_value ... ok   (first-page search_impl guard)
- collapse_scroll_inner_hit_reports_snapshot_version_not_live ... ok  (NEW #621 repro, ACTIVE not #[ignore])
- single_index_scroll_is_unchanged ... ok
- multi_index_scroll_reports_the_index_each_hit_came_from ... ok
- scroll_continuation_reports_the_snapshot_pair_not_a_torn_one ... ok
- scroll_continuation_honours_disable_sequence_numbers ... ok
- multi_index_scroll_seq_no_matches_plain_search_per_index ... ok
- scroll_over_the_cap_is_refused_and_names_the_number / published_scroll_cap_matches_the_enforced_cap ... ok
- es_compat::scroll_window_tests (under/over window) ... ok
- es_compat::passage_scroll_tests (continuation) ... ok
- es_compat::inner_hit_versioning_tests (absent-snapshot omitted) ... ok

=== 5. New warnings from the refactor? ===
NONE. No rustc `warning:` lines in DEV or CI build logs; clippy -D warnings clean.
No dead_code/unused from the extraction.

=== Refactor behavior-identity check (search_impl first-page) ===
Diff 7d836bb2..a4423716 of es_compat.rs shows the inline collapse-render block was
extracted VERBATIM into render_collapse_inner_hits(group, spec, idx_name:&str,
rest_total_hits_as_int:bool). Only mechanical deltas: idx_name.clone() -> idx_name.to_string()
(now &str param) and params.rest_total_hits_as_int.as_deref()==Some("true") passed as the
bool arg. search_impl call site passes exactly that expression + &idx_name, so first-page
output is unchanged. The passing first-page guard test corroborates.
scroll_page_response now strips __xy_collapse_group__/__xy_collapse_spec__ and renders
inner_hits via the same fn (the actual fix); guarded by the active new test.

CLAIM "Full cargo test -p xerj-api is green under 1.97.1, and search_impl first-page
collapse render is unchanged" — could NOT be refuted. Both profiles green (488/0/2),
clippy+fmt clean, first-page guard passes, refactor is a faithful extraction.
```

**Panel complete: 3/3 NON_BLOCKING, 0 BLOCKING.** Merging on all-green + CLEAN.

PR #626, opened 2026-08-22
fix(engine): keyword match/multi_match in a rescore query is mapping-aware (#574)

### LINKED ISSUE #574: keyword match/multi_match not mapping-aware inside Knn/SemanticSearch/function-score filters (follow-up to #354/#573)

xerj-org:
## What

`rewrite_keyword_full_text_to_term` makes a keyword `match`/`multi_match`
mapping-aware by lowering it to a whole-value `Term` — applied to the main
query and (since #574's core, `6fa09a13`) to compound wrappers and
`Knn`/`SemanticSearch`/`function_score` filters. One entry point was still
skipped: the request-level **`rescore`** secondary query.

`apply_rescore` scores each windowed hit via
`score_query_against_doc(rescore_query, &hit.source)`, routing a `Match`
through `doc_matches_query`. That layer has **no schema**, so it tokenizes
every `Match` (whitespace-split, operator OR/AND) regardless of field type. So
a rescore `match {kw: "a b"}` re-scored any doc whose keyword value merely
*contained* the token "a" (e.g. `{"kw":"a"}`), where ES — and the main query
on this engine since #571 — treat a keyword `match` as the whole value "a b".
A wrong-scoring divergence on the rescore path.

## How

- Add `rewrite_rescore_keyword(stage, schema)`: resolves a stage's
  `query.rescore_query` the same way the main query is resolved at the search
  entry (`rewrite_query_aliases` → `rewrite_keyword_full_text_to_term`).
- Build `resolved_rescore` once at the search entry and iterate it in the
  rescore application loop instead of `request.rescore`.
- Script-only rescores carry no `QueryNode` → passed through untouched;
  text-field matches stay analyzed. The `request.rescore.is_empty()` presence
  checks are unchanged (rewriting never adds or drops a stage).

## Tests

- `keyword_match_in_rescore_query_is_rewritten` — kw `match` → `Term`; a
  text-field rescore match stays analyzed; a script-only stage is unchanged.
- `keyword_match_in_remaining_wrapper_arms_is_rewritten` — the five compound
  wrappers added in #573 (`Boosted`/`Nested`/`Hybrid`/`Pinned`/`Named`) recurse
  into their sub-query. These arms were correct but previously untested — the
  test coverage #574 explicitly called for.

**Fail-before:** neutralizing `rewrite_rescore_keyword` to a no-op clone makes
`keyword_match_in_rescore_query_is_rewritten` fail — the rescore query stays
`Match { field:"kw", query:"a b", operator:Or }` (tokenized) instead of
`Term {kw:"a b"}`.

Full `cargo test -p xerj-engine` green under rustc 1.97.1 in **both** dev and
ci-test profiles. fmt `--all --check` ✓, clippy `--all-targets -D warnings` ✓.

Closes #574.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -14272,6 +14272,25 @@ impl Index {
             }
         };
         let query = &resolved_query;
+        // #574: resolve each rescore stage's secondary query the same way the
+        // main query above (alias-rewrite + keyword→Term). A keyword
+        // `match`/`multi_match` in a `rescore.query` otherwise reaches
+        // `score_query_against_doc`/`doc_matches_query`, which tokenize `Match`
+        // (no schema at that layer), so it would re-score docs matching only a
+        // token of the value rather than the whole keyword value. Only the
+        // application at the rescore stage below consumes this; the many
+        // `request.rescore.is_empty()` presence checks stay on `request` since
+        // rewriting never adds or drops a stage.
+        let resolved_rescore: Vec<RescoreQuery> = if request.rescore.is_empty() {
+            Vec::new()
+        } else {
+            let schema = self.schema.read().await;
+            request
+                .rescore
+                .iter()
+                .map(|rs| rewrite_rescore_keyword(rs, &schema.schema))
+                .collect()
+        };
 
         // ── KNN short-circuit ──────────────────────────────────────────────────
         // Vector-only queries (top-level `knn`) bypass the FTS / doc-scan
@@ -17994,7 +18013,7 @@ impl Index {
             // arrival order rather than `_id`).
             self.sort_hits_page_order(&mut final_hits);
 
-            for rescore_stage in &request.rescore {
+            for rescore_stage in &resolved_rescore {
                 apply_rescore(&mut final_hits, rescore_stage);
                 // ES re-sorts between chained rescore stages so the
                 // next stage's `window_size` applies to the top-N
@@ -32803,6 +32822,26 @@ fn rewrite_keyword_full_text_to_term(q: &QueryNode, schema: &Schema) -> QueryNod
     }
 }
 
+/// #574: resolve a rescore stage's secondary query the same way the main query
+/// is resolved — alias-rewrite then keyword→`Term` — so a keyword
+/// `match`/`multi_match` in a `rescore.query` is made mapping-aware. Without
+/// this the raw query reaches `doc_matches_query`/`score_query_against_doc`,
+/// which tokenize `Match` (they have no schema at that layer), so a rescore
+/// `match {kw: "a b"}` would re-score a doc whose keyword value is only "a"
+/// (token-OR) instead of requiring the whole value "a b" — a wrong-scoring
+/// divergence from the main query and from ES keyword semantics. A script-only
+/// rescore carries no `QueryNode` and is returned unchanged.
+fn rewrite_rescore_keyword(stage: &RescoreQuery, schema: &Schema) -> RescoreQuery {
+    let mut out = stage.clone();
+    if let Some(inner) = out.query.as_mut() {
+        inner.rescore_query = rewrite_keyword_full_text_to_term(
+            &rewrite_query_aliases(&inner.rescore_query, schema),
+            schema,
+        );
+    }
+    out
+}
+
 #[cfg(test)]
 mod keyword_rewrite_filter_tests {
     use super::*;
@@ -32926,6 +32965,128 @@ mod keyword_rewrite_filter_tests {
                 "a non-keyword field must stay a Match, got {f:?}"
             ),
             other => panic!("Knn shape changed: {other:?}"),
+        }
+    }
+
+    /// #574 (rescore half): a keyword `match` in a `rescore.query` is rewritten
+    /// to a whole-value `Term` via `rewrite_rescore_keyword`. `apply_rescore`
+    /// scores through `score_query_against_doc`/`doc_matches_query`, which
+    /// tokenize `Match` (no schema at that layer), so an un-rewritten
+    /// `match {kw:"a b"}` would re-score a doc whose keyword value is only "a".
+    /// Text-field rescores stay analyzed; script-only stages pass through.
+    #[test]
+    fn keyword_match_in_rescore_query_is_rewritten() {
+        let schema = kw_schema();
+        let kw_stage = RescoreQuery {
+            window_size: 10,
+            query: Some(xerj_query::ast::RescoreQueryInner {
+                rescore_query: kw_match(),
+                query_weight: 1.0,
+                rescore_query_weight: 1.0,
+            }),
+            script: None,
+        };
+        assert_is_kw_term(
+            &rewrite_rescore_keyword(&kw_stage, &schema)
+                .query
+                .expect("query preserved")
+                .rescore_query,
+        );
+
+        // A text-field rescore match stays analyzed (only keyword fields rewrite).
+        let text_stage = RescoreQuery {
+            window_size: 10,
+            query: Some(xerj_query::ast::RescoreQueryInner {
+                rescore_query: QueryNode::Match {
+                    field: "body".to_string(),
+                    query: "a b".to_string(),
+                    operator: Default::default(),
+                    analyzer: None,
+                    boost: None,
+                    minimum_should_match: None,
+                },
+                query_weight: 1.0,
+                rescore_query_weight: 1.0,
+            }),
+            script: None,
+        };
+        assert!(
+            matches!(
+                rewrite_rescore_keyword(&text_stage, &schema)
+                    .query
+                    .unwrap()
+                    .rescore_query,
+                QueryNode::Match { .. }
+            ),
+            "text-field rescore match must stay analyzed"
+        );
+
+        // A script-only rescore carries no QueryNode and is unchanged.
+        let script_stage = RescoreQuery {
+            window_size: 10,
+            query: None,
+            script: Some(xerj_query::ast::ScriptRescore::default()),
+        };
+        assert!(rewrite_rescore_keyword(&script_stage, &schema)
+            .query
+            .is_none());
+    }
+
+    /// #574: the five compound wrappers added in #573 (`Boosted`, `Nested`,
+    /// `Hybrid`, `Pinned`, `Named`) recurse into their sub-query so a nested
+    /// keyword `match` is made mapping-aware. These arms were correct but
+    /// untested (verified field-by-field by #573's correctness lens).
+    #[test]
+    fn keyword_match_in_remaining_wrapper_arms_is_rewritten() {
+        let schema = kw_schema();
+
+        let boosted = QueryNode::Boosted {
+            boost: 2.0,
+            query: Box::new(kw_match()),
+        };
+        match rewrite_keyword_full_text_to_term(&boosted, &schema) {
+            QueryNode::Boosted { query, .. } => assert_is_kw_term(&query),
+            other => panic!("Boosted shape changed: {other:?}"),
+        }
+
+        let nested = QueryNode::Nested {
+            path: "p".to_string(),
+            query: Box::new(kw_match()),
+            score_mode: None,
+        };
+        match rewrite_keyword_full_text_to_term(&nested, &schema) {
+            QueryNode::Nested { query, .. } => assert_is_kw_term(&query),
+            other => panic!("Nested shape changed: {other:?}"),
+        }
+
+        let hybrid = QueryNode::Hybrid {
+            queries: vec![WeightedQuery {
+                query: kw_match(),
+                weight: 1.0,
+            }],
+            fusion: Default::default(),
+        };
+        match rewrite_keyword_full_text_to_term(&hybrid, &schema) {
+            QueryNode::Hybrid { queries, .. } => assert_is_kw_term(&queries[0].query),
+            other => panic!("Hybrid shape changed: {other:?}"),
+        }
+
+        let pinned = QueryNode::Pinned {
+            ids: vec!["x".to_string()],
+            organic: Box::new(kw_match()),
+        };
+        match rewrite_keyword_full_text_to_term(&pinned, &schema) {
+            QueryNode::Pinned { organic, .. } => assert_is_kw_term(&organic),
+            other => panic!("Pinned shape changed: {other:?}"),
+        }
+
+        let named = QueryNode::Named {
+            name: "n".to_string(),
+            query: Box::new(kw_match()),
+        };
+        match rewrite_keyword_full_text_to_term(&named, &schema) {
+            QueryNode::Named { query, .. } => assert_is_kw_term(&query),
+            other => panic!("Named shape changed: {other:?}"),
         }
     }
 }


xerj-org:
## Independent verification — 3-skeptic panel (rustc 1.97.1, isolated worktrees, each reset to the fix tip and told to REFUTE)

**3/3 NON_BLOCKING — 0 BLOCKING.** Merge gate (b) satisfied; the should-fix notes below become a follow-up issue, not blockers.

<details><summary><b>Lens 1 — CORRECTNESS → NON_BLOCKING</b></summary>

```
VERDICT: NON_BLOCKING

PR #626, branch fix/574-rescore-keyword-passthrough, tip 1511659f.
Reviewed as an adversarial CORRECTNESS skeptic trying to REFUTE "the rescore
rewrite is correct." Could not find a correctness defect. The claim holds.

PROVENANCE (fixed as instructed)
  git reset --hard 1511659ff4851c2f537fbe007ab1a63dce811531 -> HEAD at 1511659f
  grep "fn rewrite_rescore_keyword" engine/crates/xerj-engine/src/index.rs
    -> 32834:fn rewrite_rescore_keyword(stage: &RescoreQuery, schema: &Schema) -> RescoreQuery
  Fix present. Toolchain rustup 1.97.1, RUST_MIN_STACK=8388608, workspace engine/.

1. MIRRORS MAIN-QUERY RESOLUTION (order verified identical)
   Main query (index.rs:14206-14212):
     rewrite_keyword_full_text_to_term(&rewrite_query_aliases(&request.query, schema), schema)
   Rescore (index.rs:32835-32840):
     inner.rescore_query = rewrite_keyword_full_text_to_term(
       &rewrite_query_aliases(&inner.rescore_query, schema), schema)
   Same nesting: alias-rewrite INNER (runs first) then keyword->Term OUTER. Aliased
   keyword field is therefore resolved to canonical name before is_keyword() lowers it,
   identical to the main query. Both call the SAME shared functions, so any shape they
   cover for the main query (Bool recursion at 32671-32690; MultiMatch keyword+text
   partition -> DisMax/Bool at 32583-32665; Boosted/Nested/Hybrid/Pinned/Named/
   FunctionScore-filter/Knn-filter wrappers) is covered identically for rescore.

2. CLONE PRESERVES EVERYTHING, ONLY rescore_query MUTATED
   RescoreQuery (xerj-query/src/ast.rs:754) derives Clone; fields window_size,
   query: Option<RescoreQueryInner>, script: Option<ScriptRescore>. rewrite_rescore_keyword
   does `let mut out = stage.clone();` then mutates ONLY out.query.as_mut().rescore_query.
   window_size, script, query_weight, rescore_query_weight all preserved. Script-only
   stage (query = None) passes through untouched.

3. WIRING IS 1:1; ONLY apply_rescore CONSUMES THE REWRITTEN QUERY
   resolved_rescore (index.rs:14284-14293) = request.rescore.iter().map(rewrite).collect()
   -> same length and order (empty iff request.rescore empty). Application loop changed
   from `&request.rescore` to `&resolved_rescore` (index.rs:18016).
   apply_rescore call sites (whole crate): ONLY index.rs:18017 (def at 36997).
   apply_rescore executes qspec.rescore_query (36999-37009) -> the rewritten one now.
   All other request.rescore uses are `.is_empty()` presence checks (14597, 15967, 17996,
   18009, 18534, 18548, 40634, 41315) -> consistent because rewrite never adds/drops a
   stage. No other site executes a raw rescore query:
     - es_compat.rs:12030 `for stage in &search_req.rescore` only handles stage.script
       (explanation wrapping; `if stage.script.is_none() { continue; }`), never rescore_query.
     - es_compat.rs:6197 is the request parser, not an executor.
   score_query_against_doc sites 14877/14885/14888 score the already-rewritten MAIN
   `query` (= resolved_query), not rescore.

4. NO NON-RESCORE BEHAVIOR CHANGE
   resolved_rescore is Vec::new() (no schema lock taken) when request.rescore is empty,
   and consumed only in the loop -> non-rescore requests are byte-identical.

5. CHAINED / MULTI-STAGE
   Loop iterates resolved_rescore in order; each stage independently rewritten; the
   between-stage re-sort (index.rs:18018+) is unchanged. Stage with both query+script:
   rewritten query applied first, script (untouched) second -> preserved.

6. TESTS (rustc 1.97.1)
   cargo test -p xerj-engine --lib keyword_rewrite_filter_tests -> 3 passed 0 failed
     (keyword_match_in_rescore_query_is_rewritten,
      keyword_match_in_remaining_wrapper_arms_is_rewritten,
      keyword_match_in_filter_carrying_nodes_is_rewritten)
   cargo test -p xerj-engine --test integration rescore -> test_rescore_changes_ranking passed
   cargo test -p xerj-engine --test tied_score_resorts -> 2 passed
     (rescore_path_resorts_keep_arrival_order, bool_text_idf_rescore_keeps_arrival_order)

NON-BLOCKING OBSERVATIONS (not defects, do not gate the merge)
  a) rewrite_rescore_keyword omits two steps the main-query resolution applies AFTER
     keyword->Term: the unsearchable_query_field error check (14217) and
     lower_lexically_typeless_clauses (14268, dense_vector lexical clause -> match_none).
     A rescore query naming an unindexed / dense_vector field is thus NOT error-checked /
     lowered the way the main query is. This is NOT a regression (rescore did zero
     rewriting before this PR) and is outside the stated keyword-lowering scope; it does
     not mis-score the keyword case being fixed. Worth a follow-up note only.
  b) The rescore rewrite re-acquires self.schema.read() (14287) separately from the
     main-query block's read guard -> a purely theoretical TOCTOU if a writer interleaves.
     Read locks are shared and the engine already re-reads schema elsewhere; not a defect.

CONCLUSION: The keyword/multi_match rescore rewrite mirrors the main-query resolution in
the correct order, preserves stage count/order/weights/window/script, and is executed at
the single rescore application site. No wrong-scoring defect found. VERDICT: NON_BLOCKING.
```
</details>

<details><summary><b>Lens 2 — REGRESSION (full both-profile suite) → NON_BLOCKING</b></summary>

```
VERDICT: NON_BLOCKING

PR #626 — branch fix/574-rescore-keyword-passthrough — tip 1511659ff4851c2f537fbe007ab1a63dce811531
Toolchain: rustup 1.97.1, workspace engine/, RUST_MIN_STACK=8388608
Provenance: git reset --hard to tip OK; grep "fn rewrite_rescore_keyword" -> HIT at index.rs:32834

Attempted to REFUTE "nothing regressed" — could not. Claim holds.

1) DEV  cargo test -p xerj-engine            EXIT=0  | 51 binaries | 1047 passed, 0 failed
2) CI   cargo test -p xerj-engine --profile ci-test  EXIT=0  | 51 binaries | 1048 passed, 0 failed
   (CI has +1 test vs DEV — profile-gated test, not a regression)
3) clippy -p xerj-engine --all-targets -- -D warnings  EXIT=0  | 0 warnings, 0 errors
4) cargo fmt --all --check                    FMT=0  (clean)

Existing / non-keyword rescore guards — ALL ok in BOTH profiles:
  - test_rescore_changes_ranking ............................. ok  (integration.rs)
  - rescore_path_resorts_keep_arrival_order ................. ok  (chained term rescore)
  - bool_text_idf_rescore_keeps_arrival_order ............... ok  (text-field rescore)
  - rescore_script_over_call_depth_reports_a_failure ........ ok  (script rescore)
  - keyword_match_in_rescore_query_is_rewritten ............. ok  (NEW unit test)

No test/clippy/fmt failure. Non-keyword rescore behavior (text field, script, chained term)
is unchanged: rewrite_rescore_keyword returns those stages as-is, and resolved_rescore
preserves stage count so every request.rescore.is_empty() presence check stays equivalent.

BLOCKING criterion (a real test/clippy/fmt failure) NOT met -> NON_BLOCKING.
```
</details>

<details><summary><b>Lens 3 — EVIDENCE (fail-before + end-to-end wiring) → NON_BLOCKING</b></summary>

```
VERDICT: NON_BLOCKING

PR #626 / branch fix/574-rescore-keyword-passthrough, tip 1511659f
Provenance: git reset --hard 1511659f OK; grep confirmed
  engine/crates/xerj-engine/src/index.rs:32834: fn rewrite_rescore_keyword(stage: &RescoreQuery, schema: &Schema) -> RescoreQuery

========================================================================
CLAIM A — Fail-before isolates the fix: CONFIRMED
========================================================================
Neutralized rewrite_rescore_keyword body to `let _ = schema; stage.clone()`
(no-op), ran:
  rustup run 1.97.1 cargo test -p xerj-engine --lib keyword_match_in_rescore_query_is_rewritten

RESULT (verbatim):
  test index::keyword_rewrite_filter_tests::keyword_match_in_rescore_query_is_rewritten ... FAILED
  thread '...' panicked at crates/xerj-engine/src/index.rs:32868:22:
  expected keyword Match rewritten to whole-value Term, got
    Match { field: "kw", query: "a b", operator: Or, analyzer: None, boost: None, minimum_should_match: None }
  test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 620 filtered out

The panic shows the rescore query stays a tokenized `Match { ... operator: Or }`
exactly as claimed. Restored via git checkout; test now passes (1 passed).

========================================================================
CLAIM B — Fix is load-bearing (Match tokenizes, Term is whole-value): CONFIRMED
========================================================================
doc_matches_query, index.rs:34954
  - Term arm (index.rs:34959-34972): returns json_values_equal(&doc_val, value)
    -> WHOLE-VALUE equality; no tokenization, no schema.
  - Match arm (index.rs:35270): builds q_tokens by splitting query on
    non-alphanumeric (index.rs:35348-35362), tokenizes field value
    (index.rs:35370-35380), and matches per operator OR/AND
    (index.rs:35380-35392). No schema at this layer.
score_query_against_doc (index.rs:37107) routes both Match and Term through
doc_matches_query; apply_rescore (index.rs:36997) calls score_query_against_doc.

So `match {kw:"foo bar"}` (OR) vs `Term{kw:"foo bar"}` genuinely differ against a
doc {kw:"foo"}: Match OR sees token "foo" present -> matches (wrong boost);
Term whole-value "foo bar" != "foo" -> no match (correct). Verified empirically
below (17.11 vs baseline). Claim B holds.

NOTE (does not affect verdict): the Match tokenizer DROPS single-character,
non-digit tokens as "historic stopword noise" (index.rs:35352-35360), so the
unit test's own "a b" example, run through real scoring, would tokenize to []
and match nothing. Multi-char tokens ("foo bar") are needed to reproduce the
actual OR-leak bug. Unit test is unaffected because it only asserts the rewrite
SHAPE (Match -> Term), not scoring.

========================================================================
CLAIM C — Wiring is real (full search rescore gets rewritten): CONFIRMED
========================================================================
Wiring in search_inner:
  - resolved_rescore built by mapping rewrite_rescore_keyword over request.rescore
    (index.rs:14284-14293), using the INDEX schema (self.schema.read().await),
    not the request.
  - Rescore application loop consumes &resolved_rescore (index.rs:18016).

I wrote an engine-level integration test (index 2 keyword docs, real Index::search
with a populated rescore stage). KEY DISCOVERY: xerj_query::parse_request does NOT
parse the top-level `rescore` field at all -- it hard-codes `rescore: Vec::new()`
(parser.rs:362). Rescore is parsed ONLY by the ES-compat HTTP handler
(xerj-api/src/es_compat.rs: parse_rescore @ 6176, wired @ 6164). So the test builds
SearchRequest.rescore manually exactly as parse_rescore does, then calls
idx.search(&req) -- genuinely exercising search_inner's resolved_rescore path.

FIXED code (tip 1511659f):
  baseline (no rescore):     {foobar: 0.1823, foo: 0.1823}
  rescore match_all:         {foobar: 10.182, foo: 10.182}   (proves rescore applied)
  rescore match kw "foo bar":{foobar: 21.168, foo: 0.1823}   (only whole-value boosted)

END-TO-END FAIL-BEFORE: reverted ONLY the loop line
(index.rs:18016  &resolved_rescore -> &request.rescore):
  rescore match kw "foo bar":{foobar: 21.168, foo: 17.113}   <-- foo WRONGLY boosted
  Test flipped/FAILED: "{kw:'foo'} must stay at baseline; got 17.113794 vs base 0.1823216"
  Compiler additionally warned `unused variable: resolved_rescore`, proving line
  18016 is the SOLE consumer of the rewritten vector.
Restored the line; scratch test removed; git tree clean; PR unit test green.

The wiring is real and correct. The rewrite is threaded from the index schema into
the actual rescore application at exactly one load-bearing site.

========================================================================
WHY NON_BLOCKING (adversarial assessment)
========================================================================
- Fail-before isolates the fix (Claim A holds).
- The fix is load-bearing at the scoring layer (Claim B holds, cited).
- The wiring is real end-to-end; I proved fail-before END-TO-END by flipping the
  single wiring line, and the tokenized-OR leak reappears exactly as predicted.
None of the three BLOCKING conditions is met.

SHOULD-FIX (NON_BLOCKING follow-ups, not correctness blockers):
1. The PR ships only UNIT tests for rewrite_rescore_keyword (shape assertions).
   There is NO committed end-to-end test that a real search's rescore query gets
   rewritten. My scratch test proves it works today, but nothing guards against a
   future refactor swapping resolved_rescore back to request.rescore. Recommend
   committing an integration test like the one I wrote (engine-level, manual
   SearchRequest.rescore, or via es_compat). File a follow-up.
2. Pre-existing weak test (NOT this PR): integration.rs test_rescore_changes_ranking
   (~line 5352) builds its rescore via parse_request, which silently drops rescore
   (Vec::new()), and then only asserts total.value > 0 via `let _ = (...)`. That
   test exercises ZERO rescore behavior. Worth flagging for cleanup.
```
</details>

### NON-BLOCKING follow-ups (to be filed as an issue)
1. Commit an end-to-end guard that a real search's rescore query gets rewritten (the evidence lens proved it works but nothing guards a future refactor swapping `resolved_rescore` back to `request.rescore`).
2. Pre-existing: `integration.rs::test_rescore_changes_ranking` builds rescore via `parse_request` (which hard-codes `rescore: Vec::new()`), so it exercises zero rescore behavior — fix to populate rescore as the ES-compat handler does.

PR #634, opened 2026-08-22
fix: make rescore ranking integration coverage non-vacuous

### LINKED ISSUE #627: test: commit an end-to-end rescore keyword-rewrite guard + fix the vacuous test_rescore_changes_ranking (follow-up to #626)
Two test-coverage gaps surfaced by the #626 verification panel (fix for #574).

### 1. No committed end-to-end guard for the rescore rewrite
#626 makes a keyword `match`/`multi_match` in a `rescore.query` mapping-aware, wired at the search entry (`resolved_rescore`) and consumed at the rescore application loop (index.rs:18016). The evidence lens **proved it works end-to-end** with a scratch engine test — a rescore `match {kw:"foo bar"}` boosts only the whole-value doc (21.17 vs 0.182), and reverting only the loop line wrongly boosts the `foo` doc to 17.11 — but that test was not committed. #626 ships only unit-level shape tests (`keyword_match_in_rescore_query_is_rewritten`), so nothing guards a future refactor swapping `resolved_rescore` back to `request.rescore`.

Action: commit an engine- or es_compat-level integration test that indexes docs and runs a search with a keyword rescore, asserting only the whole-value doc is boosted.

### 2. `test_rescore_changes_ranking` is vacuous
`integration.rs::test_rescore_changes_ranking` (~line 5352) builds its rescore via `parse_request`, which hard-codes `rescore: Vec::new()` (parser.rs:362) — the native parser does not parse `rescore` at all (only the ES-compat handler's `parse_rescore`, es_compat.rs:6176, does). So the test drops the rescore entirely and then only asserts `total.value > 0` — it exercises zero rescore behavior and passes vacuously.

Action: populate `SearchRequest.rescore` as the ES-compat handler does (or route through the handler), and assert the actual re-ranked order.

Both are NON_BLOCKING test-quality items, not correctness defects. Refs #626, #574.

mvanhorn:
## What this changes, and why

Replace the vacuous body of `test_rescore_changes_ranking` with one deterministic engine-level regression scenario that covers both requested gaps. Create an explicit keyword mapping, index a partial-token value before the exact whole value, run a tied primary query to establish the partial document's initial lead, then attach a `RescoreQuery` directly to the parsed `SearchRequest`, following the established patterns in `tied_score_resorts.rs` and `painless_script_limits.rs` rather than pretending the native parser accepts ES `rescore` JSON. Give the primary score zero weight so the assertions can prove that only the whole-value keyword document receives a rescore contribution, while also asserting that the result order flips after rescoring.

The engine's committed coverage proves the keyword-rescore rewrite only at the helper's unit-test boundary, not through `Index::search` where `resolved_rescore` must reach the application loop. The existing `test_rescore_changes_ranking` appears to cover that path but builds the request with `xerj_query::parse_request`, which intentionally leaves `SearchRequest.rescore` empty, and then asserts only that some hits exist. Consequently it neither executes rescoring nor verifies a rank change, and a future swap from `resolved_rescore` back to the raw request stages could pass unnoticed. This is a test-quality follow-up to the already-merged #626 behavior fix; production semantics are not in scope.

Fixes #627

## Evidence

No command output was captured in the environment that produced this change;
the checks below are left for CI.

## Checks

  Not run in this workspace, so this is left for CI to confirm.
  Not run in this workspace, so this is left for CI to confirm.
  Not run in this workspace, so this is left for CI to confirm.
  Not run in this workspace, so this is left for CI to confirm.
  Not run in this workspace, so this is left for CI to confirm.
  Not run in this workspace, so this is left for CI to confirm.
- [ ] For non-trivial changes: if this PR changes an operational document or runnable recipe, complete the operational audit in [`docs/CONTRIBUTION_REVIEW.md`](../docs/CONTRIBUTION_REVIEW.md). Record a negative control that exits non-zero with no artifact extracted, installed, published, or served. Keep verification and its action in one block; rerun the control with the failing exit discarded and with the next documented block appended, and require that neither reaches an artifact or unsafe/privileged path. List tested variants and unsupported variants separately, and verify every publication copy. Otherwise mark this item `N/A` with a reason.
  Not run in this workspace, so this is left for CI to confirm.

## Provenance

- **Written by:** an automated contribution run (Codex implementation, human
  operator: @mvanhorn). No human reviewed the diff before submission.
- **Verified** (commands run, output observed): none. No build or test command
  was executed in the authoring environment.
- **Assumed** (not tested, and what would falsify it): that attaching a
  `RescoreQuery` directly to the parsed `SearchRequest` reaches the application
  loop via `resolved_rescore`, and that the zero-weight primary makes the rank
  flip attributable solely to the rescore contribution. A failing or passing-
  for-the-wrong-reason run of `cargo test -p xerj-engine` on this test would
  falsify it.

### DIFF
--- a/engine/crates/xerj-engine/tests/integration.rs
+++ b/engine/crates/xerj-engine/tests/integration.rs
@@ -8,7 +8,7 @@ use tempfile::TempDir;
 use xerj_common::config::Config;
 use xerj_common::types::{FieldConfig, FieldType, Schema};
 use xerj_engine::{detect_log_format, Engine, LogFormat};
-use xerj_query::ast::{QueryNode, SearchRequest};
+use xerj_query::ast::{QueryNode, RescoreQuery, RescoreQueryInner, SearchRequest};
 use xerj_query::parse_request;
 
 // ── Helpers ───────────────────────────────────────────────────────────────────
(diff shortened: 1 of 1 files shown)


xerj-org:
Thanks for this, @mvanhorn — your diagnosis is exactly right, and the core insight (attach a `RescoreQuery` directly to the parsed `SearchRequest` because `parse_request` intentionally leaves `rescore` empty) is spot on.

Closing as **superseded**: #627 was resolved by #639 (merged just before this, at 2026-08-23T02:46), which landed the same fix independently:

- `test_rescore_changes_ranking` is rewritten the same way you propose — the rescore stage is populated directly on the `SearchRequest` (bypassing `parse_request`'s hard-coded `rescore: Vec::new()`), and it now asserts a real rank change (doc `c` lifts above `a` after rescore; teeth-verified: an empty rescore fails the assertion).
- The second gap you note — proving the keyword rewrite through `Index::search`, not just at the helper's unit boundary — is covered by a dedicated end-to-end guard added in the same PR, `rescore_keyword_query_is_mapping_aware` (keyword `match {kw:"foo bar"}` in a rescore is lowered to a whole-value `Term`, so only the exact doc is boosted; teeth-verified by neutralizing `rewrite_rescore_keyword`).

Because #639 already rewrote this exact test function, this PR is now `mergeable_state: dirty` against `main` and can't land without conflicting. Your "give the primary score zero weight to isolate the rescore contribution" technique is a nice touch — if you'd like to strengthen the committed guard with that isolation, a follow-up PR against the new tests would be very welcome.

Verified before closing: `#627` is CLOSED by merged `#639`; `gh api …/pulls/634 → mergeable_state: dirty`. Sorry for the concurrent-work collision, and thank you again.

xerj-org:
@mvanhorn — reopening the conversation because your work here shouldn't end uncredited.

Your diagnosis drove the fix: the core insight (attach `RescoreQuery` directly to the parsed `SearchRequest` because `parse_request` intentionally leaves `rescore` empty) is exactly what landed. That an internal PR closed three minutes before yours resolved the same issue was a genuine concurrent-work collision, not a rejection — and it's on us that it left you without a merged contribution. We'd like to fix that.

There's real, wanted work here that is **not** in `main`: the "give the primary score zero weight to isolate the rescore contribution" technique from your PR. It makes the committed guard strictly better (it proves the rescore *alone* changes rank, independent of the primary score), and it's yours. A short PR adding that isolation to the new `test_rescore_changes_ranking` / `rescore_keyword_query_is_mapping_aware` tests would be a clean, fast-tracked merge.

Two-minute path to land it as a credited contributor:
1. **Sign the CLA** (one-time, from your account): open a small PR adding `"mvanhorn"` to [`.contributors`](https://github.com/xerj-org/xerj/blob/main/.contributors) — that PR *is* your signature. See [CLA.md](https://github.com/xerj-org/xerj/blob/main/CLA.md).
2. Open the follow-up PR with the zero-weight isolation against the current tests, and comment `@cla-bot check`.

Happy to review both within the day and get them merged. Thanks again — we want more of exactly this.

xerj-org:
Thanks for this, and sorry it got overtaken. #627 landed on main via #639 (342e4044): it makes test_rescore_changes_ranking non-vacuous (real re-rank assertion) and adds a keyword-rescore guard (rescore_keyword_query_is_mapping_aware) — the same gap this PR targets. Your version now conflicts with that test and covers the same behaviour with a parallel scenario, so closing as superseded. The partial-vs-whole keyword assertion here is nice; if you want to add it as an extra case on top of the current test on main, a fresh small PR would be welcome.

xerj-org:
@mvanhorn — following up on the offer above: your zero-weight-primary isolation scenario is now #926, as a commit **authored by you** (cherry-picked from `ee24eda1`, conflicts resolved on our side so it sits beside the #639 tests instead of replacing them). Merging it puts you on the contributor graph, which a re-typed version would not. The only thing between it and `main` is the CLA: a pull request from your account adding `"mvanhorn"` to [`.contributors`](https://github.com/xerj-org/xerj/blob/main/.contributors) is the signature. Whenever you like.

xerj-org:
Merged into `main` (27466965) with the conflict resolved on our side — your commit `ee24eda1` is on main as authored, and the scenario runs as `rescore_alone_reorders_a_tied_primary` beside the #639 tests. Thank you, and sorry it took a collision and three weeks. The CLA signature request above still stands for anything you send next.

PR #640, opened 2026-08-23
test(api): guard collapse inner_hits page1==page2 byte-identity (#623)

### LINKED ISSUE #623: test: page1-vs-page2 byte-identity guard for collapse-scroll inner_hits (follow-up to #622)
Follow-up from the #622 verification panel (fix for #621).

#622's own test (`collapse_scroll_snapshot_version_torn_read`) asserts the page-2 (scroll continuation) collapse render in isolation: no sentinel leak, inner_hits present, snapshot `_version`. That is sufficient to guard the #621 defect, but two panel notes point at a stronger guard:

- **Correctness lens** wrote (and passed) a test that renders the *same* collapse group via `search_impl` (page 1) and via `scroll_page_response` (page 2) and asserts the full `inner_hits` JSON is **byte-identical**, exercising multi-spec array inner_hits, inner-level `collapse.field`, `fields`, `from`/`size`, `sort desc`, `version`+`seq_no_primary_term`, and `rest_total_hits_as_int` (both false and true). This is a materially stronger guard than page-2-in-isolation because it pins the two paths to each other through the shared `render_collapse_inner_hits`.
- **Evidence lens** noted the current test's snapshot-`_version` assertion is a valid lock-in guard but *tautological* w.r.t. today's architecture (no render-time live-lookup path exists), so it cannot fail today. A byte-identity test is the load-bearing coverage.

Action: upstream the page1-vs-page2 byte-identity test into `engine/crates/xerj-api/tests/`. Non-blocking; test-only.

Refs #622, #621.

xerj-org:
Closes #623.

## What
Adds `collapse_scroll_page_parity.rs`: an integration guard that renders the **same** collapse group two ways and asserts the `inner_hits` are byte-identical.

- **Path A (first page)** — plain `POST /docs/_search` with `collapse`, rendered by `search_impl`.
- **Path B (continuation page)** — a `size:1` `_search?scroll=` continuation, rendered by `scroll_page_response`.

Both now route through the shared `render_collapse_inner_hits` helper extracted in #621/#622. This test pins them to each other under one rich `inner_hits` spec (inner sort, `size`/`from`, `fields`, `seq_no_primary_term`, `version`), so a future divergence in either render path fails loudly.

## Why
The #622 correctness verification flagged that the PR's own test only exercised the **continuation** page in isolation — it never proved the first-page render agrees. This is exactly the guard that review recommended committing as a follow-up.

## Testing (rustc 1.97.1)
- `cargo test -p xerj-api --test collapse_scroll_page_parity` — pass (both renders byte-identical, members non-empty).
- Full `cargo test -p xerj-api` — **dev**: all green; **`--profile ci-test`**: exit 0, 54/54 test binaries `ok`.
- `cargo fmt --all --check` clean; `cargo clippy -p xerj-api --all-targets` clean.

Test-only change (no engine/src touched).

### DIFF
--- a/engine/crates/xerj-api/tests/collapse_scroll_page_parity.rs
+++ b/engine/crates/xerj-api/tests/collapse_scroll_page_parity.rs
@@ -1,0 +1,167 @@
+//! Issue #623: the collapse `inner_hits` a client sees for a given group must be
+//! byte-identical whether that group's leader is rendered on the FIRST page (by
+//! `search_impl`, plain `_search`) or on a scroll CONTINUATION page (by
+//! `scroll_page_response`). Both now route through the shared
+//! `render_collapse_inner_hits` (#621/#622), so this pins them to each other —
+//! the #622 verification recommended committing exactly this guard, since the
+//! PR's own test only checked the continuation page in isolation.
+//!
+//! Elasticsearch is referenced for wire semantics only; no ES code is here.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn call(app: &axum::Router, method: &str, path: &str, body: Value) -> (StatusCode, Value) {
+    let mut req = Request::builder().method(method).uri(path);
+    let body = if body.is_null() {
+        Body::empty()
+    } else {
+        req = req.header("content-type", "application/json");
+        Body::from(body.to_string())
+    };
+    let response = app.clone().oneshot(req.body(body).unwrap()).await.unwrap();
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .unwrap();
+    (
+        status,
+        serde_json::from_slice(&bytes).unwrap_or(Value::Null),
+    )
+}
+
+/// Extract the `inner_hits` object of the hit whose `_source.grp == grp`.
+fn inner_for_group(page: &Value, grp: &str) -> Value {
+    page["hits"]["hits"]
+        .as_array()
+        .expect("hits array")
+        .iter()
+        .find(|h| h["_source"]["grp"] == grp)
+        .unwrap_or_else(|| panic!("group {grp} leader not present: {page}"))
+        .get("inner_hits")
+        .cloned()
+        .unwrap_or_else(|| panic!("group {grp} leader has no inner_hits: {page}"))
+}
+
+/// The `inner_hits` for collapse group "B" must be identical whether B's leader
+/// is rendered by `search_impl` (first page) or `scroll_page_response`
+/// (continuation) — a rich spec: inner sort, size/from, `fields`, and
+/// `seq_no_primary_term`/`version` (#623).
+#[tokio::test]
+async fn collapse_inner_hits_are_identical_on_first_and_continuation_pages() {
+    let (app, _dir) = app().await;
+    let (st, body) = call(
+        &app,
+        "PUT",
+        "/docs",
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent 3-skeptic verification (each REFUTES, in its own worktree under rustc 1.97.1)

Fix tip verified: `6ec9371656ca5d8b4416e21b14114ae066d2aeaf`. All three skeptics: `head_ok=true`, `test_present=true`, `test_passes=true` ("1 passed; 0 failed"), `is_real_guard=true`. **Result: 3× CONFIRMED_SOUND, ZERO BLOCKING findings.**

### Skeptic 1 — correctness lens — `CONFIRMED_SOUND` (blocking: none)
> Reset to 6ec9371656ca5d8b4416e21b14114ae066d2aeaf (git rev-parse confirms HEAD). Test file present with fn collapse_inner_hits_are_identical_on_first_and_continuation_pages. Ran `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-api --test collapse_scroll_page_parity` -> "1 passed; 0 failed".
>
> CORRECTNESS lens: first_page_b is extracted from `plain` (POST /docs/_search, size 10, rendered by search_impl); continuation_b is extracted from `page2` (POST /_search/scroll continuation, rendered by scroll_page_response). Two genuinely different responses from two distinct code paths — verified via grep in es_compat.rs: search_impl (:8675) calls render_collapse_inner_hits at :13472; scroll_page_response (:20446) calls it at :20605. NOT an object compared to itself. Scroll page1 asserted == group A (lines 142-145); page2 carries group B (inner_for_group panics if B absent), so the compared group renders on a real continuation page. Non-empty guard at lines 163-166 requires /members/hits/hits/0 to exist — not a vacuous {}=={}.
>
> is_real_guard proven by adversarial perturbation: I injected an extra key `__perturb__` into ONLY the scroll-path inner_hits (es_compat.rs:20621 call site). The test then FAILED at line 157 with a clear divergence diff — left (search_impl) had real members b1/b0 with fields/version/seq_no, right (scroll) had the same members PLUS __perturb__:true. So the equality assertion bites when the two render paths diverge and the compared content is non-empty. Reverted with `git checkout -- engine/crates/xerj-api/src/es_compat.rs`; git status clean; re-ran test -> green again. Worktree left on the fix SHA. No blocking issues — passing, non-tautological, correct guard.

### Skeptic 2 — regression-risk lens — `CONFIRMED_SOUND` (blocking: none)
> Reset to the fix SHA; git rev-parse HEAD confirms match. Test file exists with the parity fn. Ran the test: "1 passed; 0 failed" — re-ran a total of 4 times (3 clean + 1 final after revert), all pass, zero flakiness/nondeterminism.
>
> REGRESSION lens: `git show --stat` shows the commit touches ONLY the test file (1 file changed, 167 insertions, no src/ modified) — test-only confirmed. Corpus is deterministic: 5 docs indexed once, single _refresh, NO updates between the two reads, so seq_no/version/score are stable (perturbed-run output shows _seq_no=2/3, _version=1, _score=1.0, _primary_term=1 identical across both paths).
>
> is_real_guard adversarial proof: the two compared objects come from genuinely different handlers — plain _search via search_impl (es_compat.rs:13472) vs scroll continuation via scroll_page_response (es_compat.rs:20605) — each calling the shared render_collapse_inner_hits. NOT a tautology. Injected a throwaway perturbation into ONLY the scroll continuation path; the byte-equality assert at line 157 then FAILED with a clear divergence diff. Reverted with `git checkout --`; git status clean, HEAD still at the fix SHA. Test also asserts inner_hits non-empty (pointer /members/hits/hits/0). No blocking findings.

### Skeptic 3 — evidence-holds lens — `CONFIRMED_SOUND` (blocking: none)
> Setup: `git fetch origin -q` + `git reset --hard <SHA>`; `git rev-parse HEAD` == the fix SHA.
>
> test_present: the test file exists (6030 bytes); `fn collapse_inner_hits_are_identical_on_first_and_continuation_pages` at line 64. test_passes: "test result: ok. 1 passed; 0 failed; 0 ignored". Compiled clean on 1.97.1.
>
> is_real_guard (EVIDENCE adversarial check): NOT a tautology — compares first_page_b (plain _search size-10, search_impl) vs continuation_b (scroll page-2, scroll_page_response). Line 163 asserts continuation_b members non-empty, blocking a vacuous empty==empty match. Perturbation to prove it bites: at the search_impl (FIRST-page) call site (~13472), captured the shared render into a mut var and dropped the last element of `/members/hits/hits` (`hits.pop()`), affecting the first-page path ONLY. Rebuilt and reran: test FAILED at collapse_scroll_page_parity.rs:157 — first_page rendered 1 member (b1); continuation rendered 2 members (b1, b0). The byte-identity assertion caught the divergence exactly as intended. Revert: `git checkout --`; HEAD back at fix SHA; `git status --porcelain` empty. Verdict: CONFIRMED_SOUND — passing, non-tautological, correctly-biting guard.

---
Merging on the standing gate: **0 BLOCKING findings**, all checks green (18/18, none pending), mergeStateStatus CLEAN.

PR #641, opened 2026-08-23
fix(engine): faithful ES _source glob automaton for nested paths (#602)

### LINKED ISSUE #602: _source filter: matches_path diverges from ES for leading/mid globs and dot-boundary (surfaced verifying #599/#310)

xerj-org:
Closes #602.

## Problem
`filter_object`'s dotted-path branch (`index.rs`) used an ad-hoc prefix/recursion matcher (`matches_path`) that could not step through a leading/mid glob nor expand a trailing-`*` segment across the dot into a subtree. Nested/dotted `_source` include patterns therefore returned ES-divergent projections (the three divergences left open after #599 fixed divergence 1):

- **Divergence 2 (leading/mid glob):** `["*.foo"]` over `{"a":{"foo":1,"bar":2}}` and `["a.*.c"]` over `{"a":{"b":{"c":1,"d":2}}}` returned `{}` — ES keeps `{"a":{"foo":1}}` / `{"a":{"b":{"c":1}}}`.
- **Divergence 3 (trailing `*` across the dot):** `["meta*"]` alongside a dotted sibling (which forces the nested branch) dropped `meta`'s whole subtree; ES keeps it (the segment `meta*` fully matches key `meta`, so the subtree is included).
- **Divergence 4 (`exclude a.*` over `{"a":{}}`):** already correct on current `main` via #310's `path_is_accepting`; pinned by a test.

## Fix
Replaced `matches_path`/`collect_and_filter` with a faithful **per-segment ES automaton**:
- each pattern is split on `.` into segments; a `*` matches exactly **one whole segment** (never across a `.`, since keys hold no dots);
- a pattern **fully consumed at a node includes that node's entire subtree**;
- an **exclude drops a node only on a full match at that node** (a deeper exclude strips children, never the bare parent).

Preserved exactly:
- the #310 empty-object rules — prune a *filtering-emptied* object, keep an *already-empty* one only in an accepting state;
- divergence 1 (dot boundary: `meta.*` keeps only paths under `meta.`, not the sibling `metadata`);
- object descent goes through the automaton while non-object values (scalars/arrays) are kept whole, so **no array data-loss** is introduced (scope is object-nested paths, per the issue).

## Testing (rustc 1.97.1)
- New `source_glob_leading_mid_and_trailing_match_es_semantics` covers divergences 2/3/4; **fail-before** confirmed (2a/2b/2c returned `{}`, nested `meta*` dropped its subtree) before the automaton landed.
- All existing #310 / #602-div1 source-filter tests unchanged and green (11/11 in `source_filter_tests`).
- Full `cargo test -p xerj-engine` — **dev**: 51 binaries, 0 failures; **`--profile ci-test`**: 51 binaries, 0 failures.
- Full `cargo test -p xerj-api` — dev: 53 binaries, 0 failures.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -26496,14 +26496,23 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
         None => return source.clone(),
     };
 
-    // Check if any include/exclude path is dotted (nested).
-    let has_dotted = includes
+    // Route through the ES automaton whenever a pattern is nested (dotted) OR
+    // contains a wildcard. The simple top-level branch below understands only a
+    // trailing `*` on a whole key (`field_matches`), so a dot-less LEADING or
+    // MID glob must go through the NFA too: in ES a `*` is `Σ*` over the full
+    // dotted path whether or not the pattern itself contains a `.`, so `*id`
+    // matches `user.groupid` and `exclude ["*secret"]` strips `password_secret`.
+    // Without this, `include ["*id"]` dropped the whole source and
+    // `exclude ["*secret"]` silently no-oped — the data-loss / data-exposure
+    // failure modes #602 exists to eliminate. Only pure-literal top-level
+    // patterns (no `.`, no `*`) take the fast path.
+    let needs_automaton = includes
         .iter()
         .chain(excludes.iter())
-        .any(|f| f.contains('.'));
+        .any(|f| f.contains('.') || f.contains('*'));
 
-    if !has_dotted {
-        // Simple top-level filtering.
+    if !needs_automaton {
+        // Simple top-level filtering: literal whole-key includes/excludes only.
         let mut result = serde_json::Map::new();
         for (k, v) in obj {
             let keep = if includes.is_empty() {
@@ -27037,6 +27046,41 @@ mod source_filter_tests {
         assert_eq!(
             actual, expected,
             "ES _source `*` crosses dots at any depth (makeAnyString); scalars kept only when accepted"
+        );
+    }
+
+    /// #602 (verification follow-up 2): a DOT-LESS leading/mid wildcard is still
+    /// `Σ*` over the full path in ES, so `*id` matches `user.groupid` and
+    /// `exclude ["*secret"]` strips `password_secret` at any depth. The dispatch
+    /// gate must route any pattern containing `*` (not only dotted ones) through
+    /// the automaton — the simple top-level branch understands only a trailing
+    /// `*`, so without this `include ["*id"]` dropped the whole source and
+    /// `exclude ["*secret"]` silently no-oped.
+    #[test]
+    fn source_dotless_wildcard_routes_through_the_automaton() {
+        // include `*id` selects any key ending in `id`, at any depth.
+        assert_eq!(
+            filter_object(
+                &json!({ "user": { "groupid": 1, "name": 2 }, "id": 3 }),
+                &["*id".to_string()],
+                &[]
+            ),
+            json!({ "user": { "groupid": 1 }, "id": 3 }),
+            "dot-less `*id` must match nested `user.groupid` and top-level `id` (#602)"
+        );
+        // exclude `*secret` strips a sensitive key at the top level AND at depth.
+        assert_eq!(
+            filter_object(
+                &json!({
+                    "password_secret": 1,
+                    "keep": 2,
+                    "nested": { "api_secret": 3, "host": 4 }
+                }),
+                &[],
+                &["*secret".to_string()]
+            ),
+            json!({ "keep": 2, "nested": { "host": 4 } }),
+            "dot-less `exclude *secret` must strip `password_secret` and nested `api_secret` (#602)"
         );
     }
 
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -26521,123 +26521,132 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
 
     // Dotted path filtering: a faithful ES `_source` include/exclude automaton.
     //
-    // ES matches PER PATH SEGMENT. Each pattern is split on `.` into segments;
-    // a `*` matches exactly one WHOLE segment (never across a `.`, because keys
-    // contain no dots); a pattern fully consumed AT a node includes that node's
-    // entire subtree; and an exclude removes a node only when its pattern fully
-    // matches at that node. The previous ad-hoc `matches_path` was a
-    // prefix/recursion matcher that could not step through a leading/mid glob
-    // (`*.foo`, `a.*.c`) nor expand a trailing-`*` segment (`meta*`) across the
-    // dot into the subtree (#602 divergences 2/3/4). Object descent goes through
-    // the automaton; a non-object value (scalar or array) is kept whole when its
-    // key is a live candidate — XERJ does not project into arrays today, so
-    // #602 is deliberately scoped to object-nested paths.
+    // ES compiles each include/exclude pattern with `Regex.simpleMatchToAutomaton`,
+    // where `*` becomes `Automata.makeAnyString()` — it matches ANY run of
+    // characters INCLUDING `.`, so a wildcard spans path segments at any depth
+    // (`org.elasticsearch.common.regex.Regex`, Apache-2.0). That is why `obj.*`
+    // keeps `obj`'s whole subtree, `*.foo` keeps every `foo` leaf at any depth,
+    // and `exclude ["*.secret"]` scrubs a nested `secret` however deep it sits.
+    // `XContentMapValues.filter` then walks the document and, at each node, asks
+    // the automaton two questions about that node's dotted path:
+    //   * ACCEPT — the pattern matches this path in full → include the whole
+    //     subtree (subject to excludes);
+    //   * LIVE — this path is a proper prefix the pattern can still extend to an
+    //     ACCEPT deeper down → descend.
+    // A leaf that is only LIVE (never ACCEPTed) was not selected and is dropped.
+    //
+    // The matcher is a small glob NFA over the FULL dotted path (`*` = any char
+    // incl `.`); leading/mid/trailing wildcards, the dot boundary, and
+    // multi-level globs all fall out of it (#602). Object descent goes through
+    // the automaton; scalars and arrays are handled at the leaf (see below).
 
-    // A single-segment glob: `*` matches any run of characters WITHIN one
-    // segment. Keys never contain `.`, so this can never cross a dot boundary —
-    // the property the old cross-path `glob_match` had to enforce by hand.
-    fn seg_matches(seg: &str, key: &str) -> bool {
-        if !seg.contains('*') {
-            return seg == key;
-        }
-        // Iterative glob with backtracking (`*` = greedy, backtrack on mismatch).
-        fn walk(pat: &[u8], s: &[u8]) -> bool {
-            let (mut pi, mut si) = (0usize, 0usize);
-            let (mut star, mut mark) = (usize::MAX, 0usize);
-            while si < s.len() {
-                if pi < pat.len() && pat[pi] == b'*' {
-                    star = pi;
-                    mark = si;
-                    pi += 1;
-                } else if pi < pat.len() && pat[pi] == s[si] {
-                    pi += 1;
-                    si += 1;
-                } else if star != usize::MAX {
-                    pi = star + 1;
-                    mark += 1;
-                    si = mark;
-                } else {
-                    return false;
+    // Reachable-state summary of glob `pat` after consuming `text`, with
+    // `*` = Σ* (matches any character, including `.`). Returns `(accept, live)`:
+    // `accept` = the whole pattern matched `text`; `live` = some reachable
+    // position can still consume a deeper `.segment` (sitting on a `*`, or on a
+    // literal `.` awaiting the next level), i.e. `text` is a viable prefix.
+    fn glob_reach(pat: &[u8], text: &[u8]) -> (bool, bool) {
+        let n = pat.len();
+        // Epsilon-closure: a `*` may match zero characters, so reaching the
+        // position of a `*` also reaches the position after it. A single forward
+        // pass suffices because a `*` only ever skips to `i + 1`.
+        let closure = |set: &mut [bool]| {
+            for i in 0..n {
+                if set[i] && pat[i] == b'*' {
+                    set[i + 1] = true;
                 }
             }
-            while pi < pat.len() && pat[pi] == b'*' {
-                pi += 1;
+        };
+        let mut cur = vec![false; n + 1];
+        let mut nxt = vec![false; n + 1];
+        cur[0] = true;
+        closure(&mut cur);
+        for &c in text {
+            for x in nxt.iter_mut() {
+                *x = false;
             }
-            pi == pat.len()
+            for i in 0..n {
+                if !cur[i] {
+                    continue;
+                }
+                if pat[i] == b'*' {
+                    // Σ* consumes any character (including `.`) and stays put.
+                    nxt[i] = true;
+                } else if pat[i] == c {
+                    nxt[i + 1] = true;
+                }
+            }
+            closure(&mut nxt);
+            std::mem::swap(&mut cur, &mut nxt);
         }
-        walk(seg.as_bytes(), key.as_bytes())
+        let accept = cur[n];
+        let live = (0..n).any(|i| cur[i] && (pat[i] == b'*' || pat[i] == b'.'));
+        (accept, live)
     }
 
-    // Filter one value against the active include/exclude segment-suffixes.
-    // `include_all` means an ancestor pattern was already fully consumed (or
-    // the top-level include list is empty), so this whole subtree is included
+    // `(accept, live)` of a path against a set of patterns — OR over the set.
+    fn reach_any(patterns: &[String], path: &str) -> (bool, bool) {
+        let mut acc = false;
+        let mut liv = false;
+        for p in patterns {
+            let (a, l) = glob_reach(p.as_bytes(), path.as_bytes());
+            acc |= a;
+            liv |= l;
+        }
+        (acc, liv)
+    }
+
+    // Filter one value at dotted `path`. `include_all` means an ancestor path
+    // was ACCEPTed (or the include list is empty), so this subtree is included
     // subject only to excludes.
     fn es_filter(
         value: &Value,
+        path: &str,
         include_all: bool,
-        includes: &[&[String]],
-        excludes: &[&[String]],
+        includes: &[String],
+        excludes: &[String],
     ) -> Value {
         let Value::Object(obj) = value else {
             return value.clone();
         };
-        let level_all = include_all || includes.is_empty();
         let mut out = serde_json::Map::new();
         for (key, val) in obj {
-            // An exclude whose pattern fully matches AT this node drops the value.
-            if excludes
-                .iter()
-                .any(|p| p.len() == 1 && seg_matches(&p[0], key))
-            {
+            let child = if path.is_empty() {
+                key.clone()
+            } else {
+                format!("{path}.{key}")
+            };
+
+            let (exc_acc, exc_live) = reach_any(excludes, &child);
+            if exc_acc {
+                // The exclude matches this path in full → drop the subtree.
                 continue;
             }
-            let exc_children: Vec<&[String]> = excludes
-                .iter()
-                .filter(|p| p.len() > 1 && seg_matches(&p[0], key))
-                .map(|p| &p[1..])
-                .collect();
-
-            // Which include patterns match this key? A fully-consumed one
-            // includes the whole subtree; a longer one is a LIVE state that
-            // descends by its tail.
-            let mut key_fully_included = false;
-            let inc_children: Vec<&[String]> = includes
-                .iter()
-                .filter_map(|p| {
-                    if seg_matches(&p[0], key) {
-                        if p.len() == 1 {
-                            key_fully_included = true;
-                            None
-                        } else {
-                            Some(&p[1..])
-                        }
-                    } else {
-                        None
-                    }
-                })
-                .collect();
-
-            let candidate = level_all || key_fully_included || !inc_children.is_empty();
-            if !candidate {
-                continue;
+            let (inc_acc, inc_live) = if includes.is_empty() {
+                (true, false)
+            } else {
+                reach_any(includes, &child)
+            };
+            let accept = include_all || inc_acc;
+            if !(accept || inc_live) {
+                continue; // neither selected nor a live prefix
             }
-            let child_all = level_all || key_fully_included;
-
-            // Fully included with nothing to strip below → keep the value as-is.
-            if child_all && exc_children.is_empty() {
-                out.insert(key.clone(), val.clone());
-                continue;
-            }
+            let child_all = accept;
 
             match val {
                 Value::Object(child_obj) => {
-                    let filtered = es_filter(val, child_all, &inc_children, &exc_children);
+                    // A fully-included object with no exclude reaching into it is
+                    // kept whole; otherwise descend to select/strip.
+                    if child_all && !exc_live {
+                        out.insert(key.clone(), val.clone());
+                        continue;
+                    }
+                    let filtered = es_filter(val, &child, child_all, includes, excludes);
                     let keep = match &filtered {
                         // An object that *filtering* emptied is pruned; one that
                         // was ALREADY empty in the source survives only in an
-                        // accepting state (its own path fully included) — a live
-                        // prefix state on the way to a deeper include is dropped,
-                        // matching ES (#310).
+                        // accepting state — a live-only prefix on the way to a
+                        // deeper include is dropped, matching ES (#310).
                         Value::Object(m) if m.is_empty() => child_obj.is_empty() && child_all,
                         _ => true,
                     };
@@ -26645,28 +26654,29 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
                         out.insert(key.clone(), filtered);
                     }
                 }
-                // Non-object (scalar/array): kept whole — XERJ does not project
-                // into arrays, and a scalar cannot satisfy a deeper include, so
-                // preserving it matches prior behavior without array data-loss.
-                _ => {
+                // An array is kept WHOLE whenever its key is selected: XERJ does
+                // not project inside arrays (documented scope), which also keeps
+                // it free of the projection-induced data loss #602 must avoid.
+                // (Consequence: an exclude cannot strip fields inside an array of
+                // objects — tracked separately, not a #602 regression.)
+                Value::Array(_) => {
                     out.insert(key.clone(), val.clone());
+                }
+                // A scalar leaf is kept ONLY in an ACCEPTing state. A scalar that
+                // merely sits on a LIVE (not-yet-accepted) include path cannot
+                // extend to satisfy the pattern, so ES drops it — keeping it
+                // would over-include a field the pattern never selected.
+                _ => {
+                    if child_all {
+                        out.insert(key.clone(), val.clone());
+                    }
                 }
             }
         }
         Value::Object(out)
     }
 
-    let inc: Vec<Vec<String>> = includes
-        .iter()
-        .map(|p| p.split('.').map(str::to_string).collect())
-        .collect();
-    let exc: Vec<Vec<String>> = excludes
-        .iter()
-        .map(|p| p.split('.').map(str::to_string).collect())
-        .collect();
-    let inc_refs: Vec<&[String]> = inc.iter().map(Vec::as_slice).collect();
-    let exc_refs: Vec<&[String]> = exc.iter().map(Vec::as_slice).collect();
-    es_filter(source, false, &inc_refs, &exc_refs)
+    es_filter(source, "", false, includes, excludes)
 }
 
 #[cfg(test)]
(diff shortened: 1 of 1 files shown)


xerj-org:

Fix tip verified: `4d974faf3c4b9e30cf0b2ae72689fd87238ed22b`. **Two lenses found real BLOCKING ES-divergences — this PR is being FIXED, not merged.** Both point to the same root cause: ES `_source` `*` compiles to `Automata.makeAnyString()` and matches ACROSS dots (any depth); the current per-segment automaton only handles single-level `*`, which (a) over-includes a scalar sibling on a non-accepting glob path and (b) fails to match/scrub multi-level globs like `*.foo` / `exclude *.secret`. Fix: implement the depth-crossing `*` (an NFA over the full dotted path), which subsumes both findings.

### Skeptic — REGRESSION lens — `CONFIRMED_SOUND`
- head_ok=True · tests_pass=True · fail_before_holds=True · es_divergence_found=False

Worktree /home/claude/ai/xerj/.claude/worktrees/wf_babb73eb-ccf-2. File under review: engine/crates/xerj-engine/src/index.rs.

HEAD: `git reset --hard 4d974faf...` -> `git rev-parse HEAD` == 4d974faf3c4b9e30cf0b2ae72689fd87238ed22b (head_ok). Fix present: `filter_object` (line 26493) now contains `fn seg_matches` (26539) + `fn es_filter` (26575); old `matches_path`/`collect_and_filter` gone. The many `unwrap`s the grep surfaced are all in #[cfg(test)] code, not the automaton.

TESTS: `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib source_filter_tests` -> 11 passed / 0 failed, incl. new source_glob_leading_mid_and_trailing_match_es_semantics (#602 div 2/3/4), source_glob_prefix_requires_the_dot_boundary_for_nonempty_siblings (#602 div1), and both #310 empty-object tests. Full `cargo test -p xerj-engine --lib` -> 622 passed / 0 failed / 0 ignored (no lib regression).

FAIL-BEFORE (proved without discarding the committed fix): the new test lives in the same file as the automaton, so `git checkout main -- index.rs` would delete the test too. Instead I surgically swapped ONLY the `filter_object` body (current lines 26493-26670) for main's old matches_path/collect_and_filter version (main lines 26493-26654) via a python rewrite, leaving the test module (incl. the new #602 test) intact. Result: source_glob_leading_mid_and_trailing_match_es_semantics FAILED — old code returned Object{} for every glob-include case (*.foo, a.*.c, *.foo-across-siblings, meta* subtree) where the automaton correctly returns the ES projection. Restored with `git checkout 4d974faf... -- index.rs`; tree clean.

ARRAY DATA-LOSS: throwaway test in the module confirmed a scalar/array under a candidate key survives WHOLE: include ["tags.name"] on {"tags":[{"name":"x","id":1}]} -> unchanged; include ["tags","zzz.absent"] (nested branch) -> unchanged; include ["tags"] (simple branch) -> unchanged; scalar {"tags":"flat",...} -> unchanged. Passed; then reverted via git checkout. No array drop / no projection-induced loss. This is the code path at index.rs 26648-26654 (`_ =>` keeps non-object values whole).

ES-DIVERGENCE (informational, NOT blocking, NOT in my lens's refute scope): the automaton's object-nested-path behavior matched ES for every case I exercised. The one place XERJ != full-ES is array-of-objects projection — real ES returns {"tags":[{"name":"x"}]} for include tags.name, XERJ keeps {"tags":[{"name":"x","id":1}]}. This is explicitly documented as out-of-scope in the PR/code comments ("XERJ does not project into arrays today"), matches prior XERJ behavior, and is exactly the safe no-data-loss outcome my regression lens requires — so I do NOT count it as an automaton divergence and it does not refute. Deeper `*`-across-dot ES-parity questions are delegated to the ES-semantics lens.

Final state: `git status --porcelain` empty, HEAD == 4d974faf..., es_filter+seg_matches present. Could not refute: change is regression-sound.

</details>

### Skeptic — CORRECTNESS lens — `REFUTED`
- head_ok=True · tests_pass=True · fail_before_holds=True · es_divergence_found=True

**BLOCKING:** ES DIVERGENCE (es_filter, engine/crates/xerj-engine/src/index.rs ~L26648-26653): a SCALAR leaf sitting on a live-but-NON-accepting include-glob path is kept whole, but documented ES _source semantics (and ES's own XContentMapValues.filter) keep a scalar leaf ONLY in an ACCEPTING include state — i.e. only when the pattern is fully consumed at that node. Proven by an added #[cfg(test)] probe (run under rustc 1.97.1, then reverted). Case A: filter_object({"a":{"b":{"c":1},"x":9}}, includes=["a.*.c"], excludes=[]) => ACTUAL {"a":{"b":{"c":1},"x":9}} but ES/EXPECTED {"a":{"b":{"c":1}}} (leaks scalar a.x:9; pattern `a.*.c` is not consumed at scalar `x`). Case B: includes=["*.*.foo"] on {"a":{"b":{"foo":1,"bar":2},"c":3}} => ACTUAL {"a":{"b":{"foo":1},"c":3}} but EXPECTED {"a":{"b":{"foo":1}}} (leaks scalar a.c:3). Root cause: for a key whose glob matches but leaves a non-empty include tail (inc_children non-empty, key_fully_included=false, child_all=false), the value is treated as a candidate and the `_ =>` arm inserts any non-object value whole; for scalars that keep should require child_all (accepting) as ES does. This is NOT covered by the 'no array projection' scope carve-out — a scalar simply does not match the pattern, yet is returned — so a parity-focused _source projection over-includes fields the include pattern did not select.

Steps run in isolated worktree. (1) `git fetch origin -q` + `git reset --hard 4d974faf...`; `git rev-parse HEAD`==4d974faf3c4b9e30cf0b2ae72689fd87238ed22b (head_ok). (2) grep confirms `fn es_filter` + `fn seg_matches` present inside filter_object (old matches_path/collect_and_filter gone). (3) `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib source_filter_tests` => 11 passed; 0 failed (tests_pass). (4) fail_before: spliced the parent (a358bfef) matches_path/collect_and_filter body into the fix file while KEEPING the new #602 test; `source_glob_leading_mid_and_trailing_match_es_semantics` FAILED (old impl returns {} for *.foo, a.*.c, meta*), then reverted with `git checkout --` (fail_before_holds). CORRECTNESS probe: added a temporary adversarial test with 6 of my own nested include/exclude cases (double-glob, exclude-subtree-under-include, trailing-* segment, deep mid-glob, already-empty-vs-emptied, plus scalar-on-non-accepting-glob). Cases C (include a.* + exclude a.b => {\"a\":{\"c\":2}}), D (ab* matches abc/abd not ax), E (a.*.*.z deep mid-glob), F (already-empty a survives, emptied b pruned) all AGREED with ES. Cases A and B DIVERGED: the automaton leaks a scalar sibling that lies on an include path whose pattern is not yet fully consumed (ES drops it because the include automaton is non-accepting at a scalar leaf). Probe run under 1.97.1 confirmed actual vs expected, then reverted. Final: HEAD clean at fix SHA, working tree clean, fix symbols present.

</details>

### Skeptic — EVIDENCE lens — `REFUTED`
- head_ok=True · tests_pass=True · fail_before_holds=True · es_divergence_found=True

**BLOCKING:** PR's central 'faithful ES / * matches exactly ONE segment (never across a dot)' premise DIVERGES from real ES for LEADING/MID globs spanning >=2 dots. ES compiles _source '*' via Regex.simpleMatchToAutomaton -> Automata.makeAnyString(), which matches across dots (this is exactly why ES 'obj.*' keeps the WHOLE subtree at all depths - a behavior this PR replicates only for TRAILING '*'). Verified PR outputs by probe: include '*.foo' on {a:{b:{foo:1,bar:2}}} -> {} (ES: {a:{b:{foo:1}}}); include 'a.*.c' on {a:{b:{e:{c:1,d:2}}}} -> {} (ES: {a:{b:{e:{c:1}}}}); exclude '*.description' on {a:{b:{description:1,keep:2}}} -> keeps description (ES: strips it). The exclude case is a data-over-exposure regression: a user writing _source.excludes:['*.secret'/'*.password'/'*.description'] (the exact idiom ES's own docs use) will have nested secrets NOT scrubbed at depth>=2. The three lens-enumerated claims (reached-not-dead, #310 guard, the four single-level documented divergences) all HOLD and all 11 tests pass, but the tests only cover single-level '*', so they do not catch this; the 'faithful ES' claim is false for a common nested-object pattern.

SETUP: git fetch + reset --hard 4d974faf...; git rev-parse HEAD == 4d974faf3c4b9e30cf0b2ae72689fd87238ed22b (head_ok). Fix present: grep shows fn es_filter (index.rs:26575) + fn seg_matches (26539) nested in fn filter_object (26493).

TESTS: `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib source_filter_tests` => 11 passed; 0 failed (tests_pass). Includes the new source_glob_leading_mid_and_trailing_match_es_semantics, both #310 tests (dotted_filter_keeps_an_already_empty..., dotted_include_keeps_an_empty_object_only_when_its_own_path_is_accepted), and #602-div1 source_glob_prefix_requires_the_dot_boundary_for_nonempty_siblings.

CLAIM 1 (reached, not dead): filter_object is the SourceFilter::Fields arm at index.rs:26371 (and SourceFilter::Includes at :26356). SourceFilter::Fields is constructed from real query parsing at xerj-api/src/es_compat.rs:6102 and xerj-query/src/parser.rs:3307, and flows through apply_source_filter_measured called at index.rs:18620 and :30890 (real search path). Not dead code. HOLDS.

CLAIM 2 (#310 guard): edited index.rs:26641 `Value::Object(m) if m.is_empty() => child_obj.is_empty() && child_all` to always-`true`; re-ran => BOTH #310 tests FAILED (panic at 26761 left {meta:{},title,nested:{}} vs right {meta:{},title}; panic at 26779 left {meta:{}} vs right {}). Reverted via git checkout --. HOLDS.

CLAIM 3 (four documented divergences): 2a `*.foo` on {a:{foo,bar}} -> {a:{foo}}; 2b `a.*.c` on {a:{b:{c,d}}} -> {a:{b:{c}}}; 3 nested `meta*` -> {meta:{foo}} and does not match sibling `other`; div-1 `meta.*` vs `metadata` (empty and non-empty) -> {}. All asserted by the green tests. HOLD at single level.

FAIL-BEFORE (#602): spliced HEAD~1's old filter_object (matches_path/collect_and_filter, lines 26493-26654 of parent) into HEAD's file while KEEPING HEAD's new #602 test. Ran source_filter_tests => source_glob_leading_mid_and_trailing_match_es_semantics FAILED (old code returns {} for `*.foo`, `a.*.c`, `*.foo across siblings`, and drops the `meta*` subtree), other 10 pass. Proves the new test guards the automaton (fail_before_holds). Reverted; git status clean; HEAD still 4d974faf.

ADVERSARIAL PROBE (es_divergence_found=true): added throwaway test, ran with --nocapture: `*.foo` on {a:{b:{foo:1,bar:2}}} => {}; `a.*.c` on {a:{b:{e:{c:1,d:2}}}} => {}; exclude `*.description` on {a:{b:{description:1,keep:2}}} => {a:{b:{description:1,keep:2}}} (description NOT stripped); `obj1.*` on {obj1:{a:{b:1}}} => {obj1:{a:{b:1}}} (whole subtree). ES ground truth is deductive, not executed: ES source-filter '*' = Regex.simpleMatchToAutomaton -> Automata.makeAnyString() (char range MIN..MAX includes '.'), so '*' crosses dots; the well-known ES rule that `obj.*` returns obj's ENTIRE subtree at all depths REQUIRES makeAnyString, and the same makeAnyString governs leading/mid '*', so ES keeps `*.foo` at depth>=2 and strips `*.description` at depth>=2. The PR only reproduces dot-crossing for TRAILING '*' (via its 'fully-consumed -> subtree' rule) but treats leading/mid '*' as one segment => divergence at depth>=2. Repo has NO golden ES fixture pinning multi-level '*' (grep for makeAnyString/simpleMatchToAutomaton/XContentMapValues = 0 hits), so this went untested. Throwaway edits reverted; working tree clean at 4d974faf3c4b9e30cf0b2ae72689fd87238ed22b.

FAIRNESS NOTE: PR is transparent that its design is per-segment ('never across a dot') and scopes arrays out, and the OLD code was worse (returned {} for single-level `*.foo` too). But the PR markets 'faithful ES semantics' and the divergence hits a pattern ES's own docs use (excludes:['*.description']) with data-exposure consequences, so I rate it blocking under a truthfulness/data-safety lens.

</details>

xerj-org:
## Re-verification (rev2 — depth-crossing NFA @ `49719aae`), 3 skeptics under rustc 1.97.1

**regression + evidence: CONFIRMED_SOUND (0 blocking). correctness: REFUTED — a real BLOCKING dispatch bug.** The NFA itself is faithful to ES `makeAnyString` (all dotted cases correct, 623 lib tests pass, array data-loss-free), but the `has_dotted` dispatch gate only routes patterns containing a literal `.` to it — a dot-less leading/mid glob (`*id`, `*secret`) fell to the legacy top-level `field_matches` and produced data loss (include) / silent no-op (exclude). **Fixing, not merging** — the gate now routes any `*`-containing pattern through the NFA.

### correctness — `REFUTED`
- tests_pass=True · fail_before_holds=True · es_divergence_found=True

**BLOCKING:** DATA-LOSS INCLUDE. filter_object(source={"user":{"groupid":1,"name":2},"id":3}, includes=["*id"], excludes=[]) => got {} ; expected(ES makeAnyString)={"user":{"groupid":1},"id":3}. A dot-less LEADING/MID wildcard never reaches the depth-crossing NFA: has_dotted (index.rs:26500-26503) is false when no pattern contains a literal '.', so the whole request is routed to the simple top-level branch using field_matches (index.rs:28007-28013), which ONLY handles a TRAILING '*' (prefix glob) and never descends. So `*id` becomes a literal string compare that matches nothing -> entire _source dropped. PROOF it is the dispatch gate and NOT the NFA: the identical include with a non-matching dotted sibling, includes=["*id","zzz.absent"], forces has_dotted=true (nested/NFA branch) and returns the correct {"user":{"groupid":1},"id":3}. Same pattern, opposite result depending on an unrelated sibling = the PR's headline invariant '*=Sigma* over the FULL dotted path' does NOT hold for dot-less globs.

**BLOCKING:** DATA-EXPOSURE EXCLUDE. filter_object(source={"password_secret":1,"keep":2}, includes=[], excludes=["*secret"]) => got {"password_secret":1,"keep":2} (exclude silently NO-OPS) ; expected(ES makeAnyString)={"keep":2}. Same root cause: has_dotted=false routes to field_matches, which only matches a trailing '*', so a leading-glob exclude of a sensitive field fails to strip it. This is exactly the 'data-over-exposure' bug class the #602 commit message itself cites for the nested case, but it re-surfaces for any dot-less leading/mid exclude (*secret, *token, *password).

Verification steps 1-3 all pass: HEAD=49719aae, glob_reach+es_filter present, 12/12 source_filter_tests green under rustc 1.97.1, and fail-before holds (neutering `*` so it stops crossing '.' makes source_glob_crosses_dots_at_any_depth_like_es FAIL). Confirmed ES semantics via ref-coding: Lucene Automata.makeAnyString() (Automata.java:71) is a single accept state with a self-loop over Character.MIN_CODE_POINT..MAX_CODE_POINT = Sigma* over ALL codepoints INCLUDING '.', so `*id` genuinely matches the full path `user.groupid` in ES.

REFUTED. My 10-case adversarial probe (added under #[cfg(test)], run under 1.97.1, then reverted; HEAD left clean) passed all the DOTTED-pattern cases (multi-level *.foo at depth 3, a.*.c spanning two levels a.b.e.c, obj.* whole subtree, scalar-sibling drop, exclude *.secret at depth>=2 with top-level `secret` correctly surviving, include a.* + exclude a.b.c, meta.* dot boundary vs metadata, a* empty-run, a.* not matching bare a) — the NFA itself is correct and faithful to makeAnyString. The divergence is NOT in glob_reach; it is the has_dotted dispatch gate that only sends patterns containing a literal '.' to the NFA. Any dot-less LEADING or MID wildcard (`*id`, `*name`, `*_at`, exclude `*secret`) is diverted to the legacy field_matches path, which (a) only understands a trailing '*', and (b) matches top-level keys only, never descending. Trailing dot-less globs like `meta*` happen to still work via field_matches (prefix match keeps the subtree) — and the existing test source_glob_leading_mid_and_trailing_match_es_semantics case 3 even adds a dummy dotted sibling `zzz.absent` specifically to force the nested branch, showing the authors already knew a bare `meta*` skips the NFA yet did not close the gate.

Scope note: the has_dotted gate and field_matches predate this PR, so this is a pre-existing gap rather than a regression this diff introduced. But the PR's title and doc-comment (index.rs:26522-26541) claim to faithfully implement ES makeAnyString for _source globs, and that claim is false for a common, ES-valid class of patterns, producing total-source data loss (include) and silent failure-to-strip (exclude) — the very failure modes #602 exists to eliminate. One-line-ish fix direction: widen the gate so any pattern containing a '*' that is not purely trailing (or simply any '*') is routed through the nested NFA branch; case 9-DIAG proves the NFA already returns the correct answer once reached.

</details>

EVIDENCE lens — every check held; could not refute. Ran exactly:

STEP 1 (head_ok): `git fetch origin -q` + `git reset --hard 49719aae…` → `git rev-parse HEAD` == 49719aaeff9ab3d4503b06df8efbc0061a86af77. Confirmed.

STEP 2 (NFA present): grep shows the DEPTH-CROSSING glob NFA is real code, not comments — `fn glob_reach` (index.rs:26548) and `fn es_filter` (26602) inside filter_object. The `*` consume-branch at 26572-26574 sets `nxt[i]=true` for ANY char (incl `.`) = Σ*, i.e. Automata.makeAnyString(). LIVE at 26583 = some reachable state sits on `*` or literal `.` (can still consume a deeper `.segment`). Scalar leaf kept only when `child_all` (accepting) at 26669-26673; array kept whole (documented no-projection scope) at 26662-26664.

STEP 3 (tests_pass): `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib source_filter_tests` → 12 passed / 0 failed, incl. source_glob_crosses_dots_at_any_depth_like_es and both #310 tests. Re-ran clean after all perturbations reverted: still 12/12.

EVIDENCE(1) makeAnyString is REAL: perturbed glob_reach so `*` skips the consume-branch when the char is `.` (added `if c != b'.'`). Rebuilt under 1.97.1. source_glob_crosses_dots_at_any_depth_like_es FAILED — with `*` blocked at the dot, `*.foo` on {a:{b:{foo,bar}}} produced {} instead of {a:{b:{foo:1}}}, and `exclude *.description` at depth 2 left description in place. Only that one test flipped (11 passed / 1 failed), proving the test genuinely guards depth-crossing and the real automaton crosses dots. Reverted via `git checkout --`; tree clean.

EVIDENCE(2) exclude-scrub-at-depth: added a temp test — `filter_object({"user":{"profile":{"password":"p","name":"n"}}}, [], ["*.password"])` → asserts `{"user":{"profile":{"name":"n"}}}`. Passed: password dropped, name kept. Reverted.

EVIDENCE(3) #310 empties guarded: perturbed the prune predicate at 26650 from `child_obj.is_empty() && child_all` to `true`. Two #310 tests FAILED (10 passed / 2 failed): dotted_filter_keeps_an_already_empty_source_object_but_prunes_a_filtering_emptied_one (filtering-emptied `nested` wrongly resurrected as {}) and dotted_include_keeps_an_empty_object_only_when_its_own_path_is_accepted (prefix-only `meta` for include `meta.foo` wrongly resurrected). Proves the empty-object accept-gating is load-bearing. Reverted.

es_divergence: none found within documented ES makeAnyString _source semantics — `*`=Σ* crosses dots, `?`/other chars treated as literals (matches Regex.simpleMatchToAutomaton, which only special-cases `*`), LIVE/descent condition correct, accepted-object subtree kept whole via include_all. The array-whole-keep is explicitly documented scope, not an undocumented divergence.

Final state: HEAD clean at 49719aae, `git status --porcelain` empty, 12/12 pass. No BLOCKING items.

</details>

HEAD=49719aae confirmed, tree clean, glob_reach(26548)+es_filter(26602) present inside filter_object with makeAnyString/Sigma* docs. All 12 source_filter_tests pass under rustc 1.97.1 (incl. source_glob_crosses_dots_at_any_depth_like_es and #310 tests). FAIL-BEFORE: spliced main's matches_path/collect_and_filter body into the fix file keeping the new tests; source_glob_crosses_dots_at_any_depth_like_es AND source_glob_leading_mid_and_trailing_match_es_semantics both FAILED (old per-segment matcher returns {} for leading/mid globs like *.foo, a.*.c, *.*.foo and leaves `exclude *.description` unstripped at depth 2); restored via git checkout. ARRAY NO DATA-LOSS: {"tags":[{"name":"x","id":1}]} with include ["tags.name"], ["tags"], and ["tags","zzz.absent"] (nested-branch-forced) all keep the whole array byte-identical (Value::Array arm inserts val.clone() whenever selected/live). FULL LIB: cargo test -p xerj-engine --lib = 623 passed, 0 failed, 0 ignored. Adversarial makeAnyString probe (7 hand-derived cases): 6 genuine glob cases matched ES exactly (bare * matchAll, *.a.b at depth, include obj + exclude obj.secret scrub, exclude *.secret crossing dots to depth 2, a.*.b keeping an OBJECT subtree whole, *.id crossing dots + dropping scalar sibling a.name). The single mismatch was my own oracle error: `exclude ["a","zzz.absent"]` prunes the filtering-emptied `zzz:{}` (documented #310 empty-object pruning, identical to main), which is orthogonal to #602 makeAnyString glob semantics -- not a divergence. glob_reach NFA analysis: closure() forward-pass handles epsilon zero-match of *; LIVE = active position on '*' or '.' correctly characterizes "can consume a deeper .segment" (over-descend is harmless -- deeper es_filter returns {} then gets pruned by the keep rule; under-descend impossible since a '.' can only be consumed by a * or literal '.'); scalars kept only on accept (child_all). Could not refute. Revert clean: HEAD still 49719aae, only glob_reach+es_filter present, no uncommitted changes.

</details>

xerj-org:
## Re-verification round 3 (dot-less-glob dispatch fix @ `59d78f3c`), 3 skeptics under rustc 1.97.1

**Result: 3× CONFIRMED_SOUND, ZERO BLOCKING.** The depth-crossing NFA + widened dispatch gate is faithful to ES `makeAnyString` — a 13-case adversarial probe (dot-less leading/mid/trailing globs at multiple depths, mixed literal+glob, pure-literal fast path, include+exclude globs, match-all) all matched the hand-derived ES answer; 624 lib tests pass; fail-before is decisive (reverting the `||f.contains('*')` clause reintroduces `include ["*id"]`→`{}` total data loss); the pure-literal fast path is preserved. **One non-blocking, pre-existing, documented note (filed as follow-up):** an `exclude` cannot strip fields *inside an array of objects* (`XERJ does not project inside arrays`, index.rs:26666-26670) — not introduced or worsened by this PR.

HEAD confirmed at 59d78f3c. Gate at index.rs:26509-26512: `needs_automaton = includes.chain(excludes).any(|f| f.contains('.') || f.contains('*'))`; glob_reach at :26557. All 13 source_filter_tests pass on rustup 1.97.1 (incl. source_dotless_wildcard_routes_through_the_automaton).

EVIDENCE lens — all three claims proven via instrumented fail-before (added a __FASTPATH__ sentinel to the simple branch only, plus two throwaway tests; all reverted, worktree clean at target commit):

CLAIM 1 (data loss closed): WITH FIX, filter_object({user:{groupid:1},id:3}, include=[*id], []) => {user:{groupid:1},id:3} and carries NO sentinel (routed through NFA). Reverting the gate to dot-only, same call returns just {__FASTPATH__:true} — i.e. un-instrumented result is {} = total data loss. Gate was the bug; fix closes it.

CLAIM 2 (exposure closed): WITH FIX, exclude=[*secret] on {password_secret:1,keep:2} => {keep:2} (NFA, no sentinel). Gate reverted: password_secret survives = silent no-op/leak. Confirmed.

CLAIM 3 (fast path preserved): pure-literal include [id] on {id:1,name:2} => carries the sentinel (fast path taken) while both glob cases carry none — the widened gate did NOT route everything through the automaton.

Fail-before decisive: zz_evidence_failbefore passes only with gate reverted; zz_evidence_withfix flips pass->fail at claim1 when reverted. No ES divergence. Gate widening real and necessary. No blocking items.

</details>

Attempted to refute the DISPATCH-gate widening (needs_automaton = pattern contains '.' OR '*'); could not. HEAD clean at 59d78f3c after all throwaway edits reverted; git status empty.

(1) FAIL-BEFORE holds: surgically reverting only the gate to `f.contains('.')` makes source_dotless_wildcard_routes_through_the_automaton FAIL — `include ["*id"]` yields left: Object {} vs right {"user":{"groupid":1},"id":3}, i.e. the whole-source data-loss #602 eliminates. Restored via git checkout, re-confirmed HEAD.

(2) Pure-literal fast path unchanged: probe added/ran/reverted — include ["title"] keeps only top-level title (subtree whole); exclude ["body"] strips only body; a pure-literal include does NOT over-match a nested same-named key (a.title dropped, top-level title kept). All pass. field_matches (index.rs:28051) is exact-match for non-trailing-* literals, ES-faithful for the top-level accept case.

(3) Full `cargo test -p xerj-engine --lib` = 624 passed, 0 failed (rustc 1.97.1, RUST_MIN_STACK=8388608). source_filter_tests = 13/13.

(4) No array data-loss regression: probe added/ran/reverted — dot-less glob selecting an array key keeps it whole; include ["*"] keeps arrays+objects byte-for-byte; unrelated array survives a dot-less exclude hitting a sibling. Manual analysis confirms the gate change is behavior-neutral for trailing-* dot-less globs (glob_reach on "prefix*" accepts a top-level key iff it starts_with prefix, identical to the old field_matches path; keeping a top-level key whole == keeping its full subtree). The only behavioral delta is the intended #602 fix (dot-less leading/mid globs *id/*secret now route through the NFA).

Non-blocking, out of scope: exclude-into-array-of-objects cannot strip nested fields (index.rs:26666-26670) — a pre-existing, documented divergence NOT introduced by this revision (such patterns contain '.' and already routed through the automaton before rev3). No new ES divergence and no regression attributable to the gate widening.

</details>

Attempted to REFUTE the dispatch widening; could not find an in-scope ES-parity defect. HEAD=59d78f3c (clean after revert). Gate at index.rs:26509 = `any pattern contains '.' || '*'`; glob_reach at 26557. All 13 source_filter_tests pass under rustup 1.97.1, incl. source_dotless_wildcard_routes_through_the_automaton.

Fail-before (self-run, cheap): reverting the gate to the round-2 form `.any(|f| f.contains('.'))` makes `include ["*id"]` on {"user":{"groupid":1,"name":2},"id":3} return {} (full data loss) instead of {"user":{"groupid":1},"id":3}. The `||f.contains('*')` clause is load-bearing, so fail_before_holds=true.

Adversarial probe (temporary #[cfg(test)], 13 in-scope cases, then reverted via git checkout): dot-less leading glob at multiple depths (*id -> id, x.paid, user.groupid; correctly does NOT match "title"/"identity"/"idle"); dot-less MID glob spanning dots (user*id matches user.groupid, userid, user_thing.groupid); dot-less excludes at depth (*secret, *token strip nested keys); mixed [title,*id]; PURE-literal [title,body] via fast path (correct); include user.* + exclude *secret; JUST `*` match-all include; `*` include + `*secret` exclude; an object filtered-to-empty is pruned; *name crossing into a nested object; literal `user` include forced onto the automaton by a dotted exclude `*.secret` (subtree kept then scrubbed); match-all exclude `*` -> {}. Every actual XERJ output matched the hand-derived ES makeAnyString answer. (One initial mismatch was MY bad expected value: *id does not match "title"; XERJ's {} was correct.)

ONE divergence exists but is OUT OF SCOPE and pre-existing, so es_divergence_found=false and NOT blocking: projection inside arrays of objects. filter_object(&{"logs":[{"msg":1,"api_secret":2}],"name":3}, [], ["*secret"]) returns the array WHOLE (api_secret leaks); ES would strip api_secret from each element. This is explicitly documented at index.rs:26666-26670 ("XERJ does not project inside arrays (documented scope)... tracked separately, not a #602 regression") and is neither introduced nor worsened by rev3's dispatch change. Worth a follow-up issue for security-sensitive exclude-of-secrets on arrays, but it does not block this revision, whose narrow claim (route dot-less globs through the automaton) is faithful to ES for objects and scalars.

</details>

PR #642, opened 2026-08-23
fix(engine): a one-clause bool is a no-op — stop it changing _score and rank (#399)

### LINKED ISSUE #399: a single-clause bool.must changes _score and ranking versus the bare query (bool wrapper is not score-neutral)
Found while verifying that #361's fix (merged as #387) actually holds on `main`. It does — this is a **different, remaining** defect in the same family, filed separately so #361 can close cleanly.

## Summary

Wrapping a query in a single-clause `bool.must` changes both its `_score` and its ranking. In Elasticsearch, `{"bool":{"must":[X]}}` scores identically to bare `X` — a one-clause boolean adds no scoring contribution of its own.

## Measured on `main` @ `1e7feb5`

120 documents, one `text` field, same query throughout:

```
bare match                    total=120   d009=0.006916 | d024=0.006916
bool{must:[match]}            total=120   d004=2.791759 | d009=2.791759     <- different score AND different top hit
bool{must:[match], filter:[exists]}  total=120   d004=2.791759 | d009=2.791759
```

Two things to read off that:

1. **The filter half of #361 is fixed** — adding `filter:[{"exists":...}]` changes nothing at all versus `bool.must` alone. `bool.filter` is genuinely non-scoring now, matching `BooleanClause.isScoring()`. That part of #387 works.
2. **The bool wrapper itself is not neutral.** `bare match` and `bool{must:[match]}` return the same 120 documents in a different order with scores differing by ~400x.

## Why it matters

It is the same user-visible failure mode #361 was about — the same query, expressed two equivalent ways, ranks differently — just reached by a different route. Anyone who refactors a bare `match` into a `bool` to add a filter later gets a silently different ranking, and the ES semantics they are relying on say they should not.

It also means `_score` is still not comparable across query *shapes*, which matters for hybrid/RRF fusion and for any client-side threshold.

## Reference

Lucene's `BooleanWeight` builds each clause's `Weight` with `c.isScoring() ? scoreMode : ScoreMode.COMPLETE_NO_SCORES` (`BooleanWeight.java:49-64`) and a single required clause contributes its own score unmodified — modern Lucene has no coord factor. `BooleanClause.isScoring()` is `occur == MUST || occur == SHOULD` (`BooleanClause.java:84-87`).

**Verified:** the table above was produced on a node built from `main` at `1e7feb5` (`--profile ci-test`), on its own port and data dir, with `track_total_hits: true` so the totals are exact.
**Assumed:** I did not determine which code path introduces the difference, and I did not check whether `should`, `minimum_should_match`, or nested bools behave the same way. The `_count`/`_search` divergence (#362) and the page-local IDF rescore (#361) are separate and the latter is fixed.

xerj-org:
Closes #399.

## Problem
Wrapping a query in a single-clause `{"bool":{"must":[X]}}` changed both its `_score` and its ranking versus bare `X`. In Elasticsearch the two are the same query — `BooleanQuery.rewrite` replaces a one-clause boolean with the clause's own query *before* any `Weight` is built (`BooleanQuery.java:279-298`, "optimize 1-clause queries") — so the wrapper is structurally incapable of contributing a score.

**Root cause:** `Index::search` steers on the *shape* of the query tree, and `is_doc_scan_query` matches every `QueryNode::Bool`. On a memtable-resident index, `bool{must:[match]}` was handed to the IDF-less brute per-document scorer (`score_query_against_doc`: `1 + ln(1 + tf)`) while the identical bare `Match` took the memtable BM25 arm with index-wide statistics. Two scorers, two scales, same clause. Measured on 120 memtable-resident docs, one `text` field, same clause: `d004` scored `2.7917595` under the wrapper vs `0.0068451734` bare, and `size:1` returned a different top document.

(#361/#387 fixed the flushed/segment half of this family; the segment FTS path is unaffected here and green before and after. This is the memtable half #399 reports.)

## Fix
`unwrap_single_clause_bool` erases the wrapper in the same place the `dense_vector` lowering already normalises the tree — before `is_doc_scan_query` or `query_node_to_fts` can read it. Only the two shapes Lucene rewrites to the clause itself are erased: a lone `must`, and a lone `should` with `minimum_should_match` absent/0/1. A lone `filter` (`BoostQuery(ConstantScoreQuery(q), 0)`, score 0) and a lone `must_not` (pure-negative) are **not** their child and keep their existing handling; the unwrap recurses through `Boosted`/`Constant`/`Named` wrappers exactly as Lucene's `rewrite` is recursive over clauses.

**Second instance, closed in the same change:** the two page-local score-rewrite gates below the executor read `request.query` (the raw request) while deciding whether to rewrite what the executor scored from the *resolved* tree. With the unwrap in place that re-opens the defect from the other side (`extract_query_text` answers `None` for a `Bool`), so both gates now read the normalised tree the executor was steered by.

**Known and unchanged (stated, not hidden):** on an unflushed index a `filter`-only `bool` scores `1.0` rather than `0.0` — that is the separately-tracked filter-context defect in the memtable brute scorer, not this one; the new test asserts the flushed population there rather than pinning `1.0` as correct.

## Testing (rustc 1.97.1)
- New `bool_single_clause_is_score_neutral.rs` (6 tests: memtable + segment score-neutrality, keyword term, every-page-size agreement, filter-only-zero-when-flushed, must_not-not-unwrapped) — all pass.
- **Fail-before proven** on current main: reverting only the index.rs source hunk (keeping the test) fails `single_clause_bool_is_score_neutral_on_memtable` and `wrapped_and_bare_agree_at_every_page_size` ("d004 is not in the bare-clause result"); the 4 segment/flushed cases stay green.
- Full `cargo test -p xerj-engine` — **dev**: 52 binaries, 0 failures; **`--profile ci-test`**: 52 binaries, 0 failures.
- Full `cargo test -p xerj-api` — dev: 54 binaries, 0 failures.

_(Recovered and rebased from an earlier `fix/issue-399` branch — 407 commits stale; index.rs auto-merged cleanly and the fix + fail-before were re-verified on current main.)_

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,34 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **Wrapping a query in a one-clause `bool` no longer changes its `_score` or
+  its ranking** ([#399](https://github.com/xerj-org/xerj/issues/399)).
+  `{"bool":{"must":[X]}}` and bare `X` are the same query — Lucene's
+  `BooleanQuery.rewrite` erases a one-clause boolean before any `Weight` is
+  built (`BooleanQuery.java:279-298`) — but XERJ steers on the *shape* of the
+  query tree, and `is_doc_scan_query` matches every `bool`. On an index whose
+  documents were still in the memtable, the wrapped form was therefore scored
+  per-document by the IDF-less brute scorer (`1 + ln(1 + tf)`) while the
+  identical bare `match` took the memtable BM25 arm. Measured on a
+  120-document fixture with one `text` field, same clause throughout: the
+  wrapped query scored `d004` at `2.7917595` where the bare clause scored it
+  `0.0068451734`, and `size:1` returned a different document for each. The
+  flushed (segment FTS) population already agreed and still does.
+
+  A one-clause `bool` is now erased before any path decision reads the tree —
+  a lone `must`, or a lone `should` with `minimum_should_match` absent, `0`
+  or `1`, exactly the cases Lucene rewrites to the clause itself. A lone
+  `filter` (`BoostQuery(ConstantScoreQuery(q), 0)`, score 0) and a lone
+  `must_not` (pure-negative) are *not* their child and are left alone. The
+  two page-local score-rewrite gates further down now read the same
+  normalised tree the executor was steered by, rather than the raw request,
+  so the divergence cannot reappear one gate later.
+
+  Known and unchanged: on an unflushed index a `filter`-only `bool` still
+  scores `1.0` instead of `0.0`. That is the separately-tracked filter-context
+  defect in the memtable's brute scorer, not this one, and this change neither
+  fixes nor worsens it.
+
 - **A file excluded by a widened ignore/hidden rule was reported as a user
   deletion, so the graph resume path refused the rerun and its recovery
   contract told operators to "restore the removed file(s)" — a file still on
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -14265,11 +14265,17 @@ impl Index {
             // the lowering the missing postings hand the clause to the
             // stored-doc scan. See `lower_lexically_typeless_clauses`.
             let typeless = crate::memtable::lexically_typeless_fields(&schema.schema);
-            if typeless.is_empty() {
+            let lowered = if typeless.is_empty() {
                 resolved
             } else {
                 lower_lexically_typeless_clauses(&resolved, &typeless)
-            }
+            };
+            // #399 — and it belongs HERE for exactly the reason above: every
+            // path decision below is made on the SHAPE of this tree, so a
+            // `bool` wrapper that ES erases before it builds a `Weight` has to
+            // be erased before `is_doc_scan_query` or `query_node_to_fts` can
+            // read it. See `unwrap_single_clause_bool`.
+            unwrap_single_clause_bool(lowered)
         };
         let query = &resolved_query;
         // #574: resolve each rescore stage's secondary query the same way the
@@ -17892,9 +17898,17 @@ impl Index {
             // Top-level constant_score: scores are the wrapper's constant —
             // the hit-set idf heuristic must not rewrite them.
             && !matches!(query, QueryNode::Constant { .. })
-            && query_uses_bool_text(&request.query)
+            // `query`, not `request.query` (#399): these gates decide whether
+            // to REWRITE what the executor just scored, so they have to read
+            // the same tree the executor was steered by — alias-resolved,
+            // typeless-lowered, one-clause-bools erased. Reading the raw
+            // request re-opens the defect from the other side: after the
+            // unwrap `bool{must:[X]}` and bare `X` take the same path and
+            // score the same, and then a gate that still sees a `Bool` would
+            // rewrite one page and not the other.
+            && query_uses_bool_text(query)
         {
-            let field_term_pairs = collect_match_field_terms(&request.query);
+            let field_term_pairs = collect_match_field_terms(query);
             if !field_term_pairs.is_empty() {
                 let n_docs = final_hits.len() as f32;
                 // For each (field, term) clause, compute its doc frequency.
@@ -17954,7 +17968,13 @@ impl Index {
             // TF-IDF fallback: when all BM25 scores are negligibly small, recompute
             // using a simpler term-frequency heuristic for more useful ranking.
             if max_score > 0.0 && max_score < 0.001 {
-                let query_terms = extract_query_text(&request.query)
+                // `query`, not `request.query` — same argument as the rescore
+                // gate above (#399). `extract_query_text` answers `None` for a
+                // `Bool`, so with the raw request a very common term (BM25
+                // below 0.001) would be TF-IDF-rewritten as a bare `match` and
+                // left alone as `bool{must:[match]}` — the identical query
+                // shape divergence, one gate further down.
+                let query_terms = extract_query_text(query)
                     .map(|t| {
                         t.split_whitespace()
                             .map(|s| s.to_lowercase())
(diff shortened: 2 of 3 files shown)


xerj-org:

Fix tip `bf7aff16`. **Result: 3× CONFIRMED_SOUND, ZERO BLOCKING.** Each ran the 6 tests (all pass), proved fail-before (reverting the source hunk fails exactly the 2 memtable tests: d004 `_score` 2.7917595 vs bare 0.0068451734), and adversarially audited the unwrap. One **non-blocking** completeness note (filed as a follow-up): the unwrap is conservative — a lone `should` with a percentage/field/script `minimum_should_match` resolving to 1, and one-clause bools nested inside DisMax/FunctionScore, are left wrapped; these keep bool semantics (never over-score), so nothing is wrongly erased.

HEAD==bf7aff16 confirmed; fix present at index.rs:34463; all 6 bool_single_clause_is_score_neutral tests pass under rustup 1.97.1. Fail-before verified by `git checkout bf7aff16~1 -- .../index.rs` (keeps the new test file): the 2 memtable tests fail with the exact documented divergence (single_clause_bool_is_score_neutral_on_memtable: 2.7917595 vs 0.0068451734; wrapped_and_bare_agree_at_every_page_size: d004 absent from bare result), while the 4 segment/keyword tests stay green — matching the commit's claim that only the memtable half regressed. Restored to clean HEAD afterward.

CORRECTNESS lens (adversarial): added a temporary 17-case #[cfg(test)] mod probe_399_unwrap calling unwrap_single_clause_bool directly, ran green, then reverted. The function erases ONLY the Lucene-correct shapes — NO over-unwrapping (no semantics change) found:
(a) filter-only bool -> NOT unwrapped (stays score-0 Bool) PASS
(b) must_not-only bool -> NOT unwrapped PASS
(c) single-should msm=Fixed(2) NOT unwrapped; 2-clause should NOT unwrapped; should msm=Percentage(100) NOT unwrapped PASS
(d) single must + filter / + must_not / + should -> NOT unwrapped; guard is `must.len()==1 && should.is_empty() && must_not.is_empty() && filter.is_empty()` (should branch symmetric) PASS
(e) Boosted{boost}/Constant{score}/Named{name} recursion preserves the wrapper value and unwraps the inner bool PASS
Positive: lone must (None|Fixed(0)) and lone should (None|Fixed(0)|Fixed(1)) collapse to child; nested bool{must:[bool{must:[X]}]} fully collapses; filter-child bool left un-recursed (non-scoring path). All confirmed.

Two page-local gates (rescore hit-set IDF gate query_uses_bool_text/collect_match_field_terms, and the <0.001 TF-IDF fallback extract_query_text) correctly switched request.query -> resolved `query`, so post-unwrap the wrapped and bare forms take the same rewrite decision.

Non-blocking (completeness, NOT correctness): the fix is intentionally conservative and leaves wrapped several shapes Lucene ALSO rewrites to the clause — single `should` with minimum_should_match "100%"/Percentage(N resolving to 1)/Field/Script, and single-clause bools nested inside wrappers not recursed here (DisMax/FunctionScore/etc.). These keep bool semantics (never over-score), so #399 could still surface for those rarer shapes on a memtable-resident index, but no shape is wrongly erased. MinShouldMatch enum has Fixed/Percentage/Field/Script variants; only Fixed(0)/Fixed(1)/None are matched, so the non-Fixed variants fall through safely.

</details>

Setup: `git fetch origin -q` + `git reset --hard bf7aff16...`; `git rev-parse HEAD` == bf7aff16891f470c4dda89b5e3219190999752ca (head_ok). Fix present: `grep -n 'fn unwrap_single_clause_bool'` → index.rs:34463.

Baseline (fix in place), from engine/: `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --test bool_single_clause_is_score_neutral` → 6 passed; 0 failed (single_clause_bool_is_score_neutral_on_memtable, _on_segments, wrapped_and_bare_agree_at_every_page_size, _for_term_on_keyword, filter_only_bool_still_scores_zero_when_flushed, must_not_only_bool_is_not_unwrapped).

EVIDENCE lens — all three claims verified:
(1) Fix guards the memtable scoring change. Disabled the unwrap (`return q;` at top of unwrap_single_clause_bool so it returns its input unchanged), rebuilt, re-ran: result FAILED, 4 passed / 2 failed. The two failures are exactly the memtable ones — single_clause_bool_is_score_neutral_on_memtable panicked "bool.must[1]: _score for d004 is 2.7917595 but the bare clause scores it 0.0068451734" (the ~400x, IDF-less-brute vs memtable-BM25 divergence the PR describes), and wrapped_and_bare_agree_at_every_page_size failed at flush=false size=1 ("d004 is not in the bare-clause result"). So the tests genuinely fail-before and guard the fix.
(2) Segment/flushed path unaffected: single_clause_bool_is_score_neutral_on_segments PASSED both with the fix and with the unwrap disabled — it was never broken.
(3) must_not_only_bool_is_not_unwrapped and filter_only_bool_still_scores_zero_when_flushed PASSED both with and without the fix. Also single_clause_bool_is_score_neutral_for_term_on_keyword passed in both states (keyword term rides the fused mem_bool_preds/scored_fast walk, independent of the unwrap).

"Nowhere else": disabling the unwrap flipped ONLY the two memtable tests pass→fail; the other four stayed green — the scoring change is confined to the memtable arm as claimed.

Restored source with `git checkout -- crates/xerj-engine/src/index.rs`; `git status --short` empty, HEAD still bf7aff16, unwrap_single_clause_bool restored. Non-blocking observation: the two page-local gate switches (request.query→resolved query at index.rs:17909 and :17977) are not independently exercised by this suite when only the unwrap is toggled — the flushed neutrality test is short-circuited by the `!exact_bm25_page` guard (#361/#387), so those gates are defensive-but-correct rather than test-falsifiable here. Not a defect; no fix required. Verdict: CONFIRMED_SOUND.

</details>

Worktree reset to bf7aff16 (head_ok), clean, fix present (unwrap_single_clause_bool at index.rs:34463). Target test bool_single_clause_is_score_neutral: 6/6 pass under rustup 1.97.1.

REGRESSION lens (3 checks, all pass):
(1) Fail-before: `git checkout main -- .../index.rs` reverts ONLY the source hunk, keeps the committed test. Result: exactly single_clause_bool_is_score_neutral_on_memtable and wrapped_and_bare_agree_at_every_page_size FAIL (4 passed; 2 failed), with the precise documented evidence — d004 _score 2.7917595 vs bare 0.0068451734, and d004 missing from the size=1 bare page. Restored to bf7aff16, worktree clean. fail_before_holds=true.
(2) Full `cargo test -p xerj-engine --lib`: 621 passed; 0 failed. The two `request.query`->`query` gate changes (query_uses_bool_text / collect_match_field_terms rescore gate, and the extract_query_text TF-IDF-fallback gate) regress no query-rewrite/rescore unit tests.
(3) Must+filter spot-check: bool_filter_is_non_scoring (4/4) asserts bool{must:[match],filter:[term]} still narrows 120->60 AND scores equal the must-only reference; es_compat_tests (65/65) and bm25_stats_are_index_wide (9/9) all green.

Adversarial logic audit found no over-reach: unwrap matches Lucene's 1-clause rewrite exactly (lone must with msm None/Fixed(0); lone should with msm None/Fixed(0)/Fixed(1)); a lone filter/must_not is correctly NOT unwrapped, and filter/must_not children are never recursed into; a must+filter bool has only_must=false so it is never collapsed (filter semantics preserved). Nothing to refute.

</details>

Issue #646, opened 2026-08-23
collapse inner_hits leak generated embedding companions under the default _source (follow-up to #310)


xerj-org:
Surfaced while implementing #310 (fields-pierce, PR forthcoming). A collapse `inner_hits` block renders each group member's `_source` from a copy stashed **before** the engine's `_source` projection runs (`apply_collapse_with_inner`), so those members carry the generated embedding companions (`*_vector`, `*_vector_chunks`) even under the default `SourceFilter::Default` — the very payload #309 exists to keep off the wire. The top-level collapse leader honours #309 (via the normal source projection), while the `inner_hits` underneath it ship the vectors.

This is **independent of the #310 pierce** (it happens whether or not a `fields` clause named a companion) and is **pre-existing** — the stash predates #310. #310's PR therefore deliberately scopes it out (it does not populate `companions_by_index` for a bare collapse request) and fixes only the top-level `fields`/`docvalue_fields` + scroll paths.

**Repro:** index a field with an embedding mapping, then `POST /idx/_search {"collapse":{"field":"grp","inner_hits":{"name":"members"}}}` with no `_source` clause — the `inner_hits.members` hits carry `<field>_vector` in `_source` while the leaders do not.

**Fix direction:** at the collapse `inner_hits` emission site, when the request's `_source` is the default projection, strip the index's `embedding_companion_fields()` from each member's `_source` (the accessor added in #310). Guard on `default_source_projection`, not the pierce flag.

xerj-org:
Implementation recon (post-#310). The inner_hits member `_source` is emitted at `engine/crates/xerj-api/src/es_compat.rs:8634` — `hit_obj.insert("_source".to_string(), src)` inside `render_collapse_inner_hits(group, spec, idx_name: &str, rest_total_hits_as_int: bool)` (extracted in #622). `src` is the stashed member source, which carries the generated companions because the collapse stash predates the #309 projection.

[...]

PR #647, opened 2026-08-23
fix(api): fields/docvalue_fields resolve a named companion under the default _source (#310)

### LINKED ISSUE #310: fields API silently omits embedding companions under the new default `_source` (follow-up to #309)

xerj-org:
Closes #310 (claim 1). (Claim 2 — dotted embedding `target_field` deleting empty objects — already landed via #599.)

## Problem
Since #309, a request with no `_source` clause gets `SourceFilter::Default` and the engine strips the generated embedding companions (`*_vector`, `*_vector_chunks`) before the response layer runs. But `fields` and `docvalue_fields` resolve their values out of that **same, already-narrowed** `hit.source` — so `{"fields": ["body_vector"]}` came back empty, **silently** (an unresolvable `fields` entry is legally omitted), while the identical clause under `"_source": false` returned the whole vector (because `Enabled(false)` deliberately keeps the raw source). Two spellings of "I don't want `_source`", opposite answers.

## Fix
Decided in the API (the only layer that sees the `fields` clause), not the engine — teaching the engine to treat `Default` as `Enabled(true)` would put companions back into every internal `SearchRequest::default()`.

- **`value_bearing_field_names(body, fields)`** — names pulled by a non-`_source` route (`fields` + `docvalue_fields`), computed before the search (URL `?fields=`/`?docvalue_fields=` are promoted into `body` by then).
- **Pierce**: when `_source` is `Default` and a named field is an embedding companion, set `search_req.source = Enabled(true)` so the builder resolves it from the intact source. `Index::embedding_companion_fields()` (a lock-scoped accessor over the existing `generated_embedding_companion_fields`) supplies the set.
- **Emission obligation** — re-narrow `_source` back to the #309 default at each site so the wire bytes are unchanged:
  - **site 1** — the top-level hit `_source`;
  - **site 3** — the scroll snapshot (before the `ScrollContext` is stored), so continuation pages rendered by `scroll_page_response` don't leak.
- **`build_docvalue_fields`** — an unresolvable `docvalue_field` is now **omitted** rather than emitted as a bare `[]` (`{"body_vector": []}` is a false "no values" claim); mirrors the `fields` builder's `None => continue`.

## Scope / follow-up
The collapse `inner_hits` companion leak under a default projection is **pre-existing and independent of the pierce** (the inner_hits source is stashed before the projection) — deliberately **out of scope** here and filed as **#646**. This PR therefore does not populate `companions_by_index` for a bare collapse request.

## Testing (rustc 1.97.1)
- New `fields_pierce_default_source.rs` (6 tests): `fields`/`docvalue_fields` resolve the companion under every `_source` spelling, `_source` never widens on the wire (positive+negative paired), unresolvable field omitted (not `[]`), a plain default search is untouched, and **scroll continuation does not leak the companion** (site 3).
- **Fail-before proven**: on current `main` 3/5 of the ported cases fail (companion resolves to `[]`).
- Full `cargo test -p xerj-api` — dev + `--profile ci-test`: 55 binaries each, 0 failures.
- `cargo fmt --all --check` clean; `cargo clippy -p xerj-api -p xerj-engine --all-targets -- -D warnings` clean.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -10866,9 +10866,11 @@ async fn search_impl(
         Vec::new()
     };
     // One schema read-lock per participating index, and only for a request that
-    // could actually consume the answer.
+    // could actually consume the answer. (The collapse `inner_hits` companion
+    // leak under a default projection is a separate, pre-existing gap tracked as
+    // a follow-up; it is not driven by the pierce, so it is not populated here.)
     let mut companions_by_index: HashMap<String, HashSet<String>> = HashMap::new();
-    if default_source_projection && (!value_bearing_names.is_empty() || body.collapse.is_some()) {
+    if default_source_projection && !value_bearing_names.is_empty() {
         for idx_name in &index_names {
             let Ok(idx) = state.engine.get_index(idx_name) else {
                 continue;
@@ -11755,7 +11757,23 @@ async fn search_impl(
     let scroll_snapshot: Option<Vec<(String, xerj_query::executor::Hit)>> = if is_scroll_request {
         // Keep the index name with each hit (#414): dropping it here is what
         // made continuation pages fall back to a single context-level name.
-        Some(merged_hits.clone())
+        let mut snap = merged_hits.clone();
+        // #310 emission site 3: the snapshot outlives this response, and every
+        // later page is rendered by `scroll_page_response`, which carries no
+        // `fields` clause of its own and so has no reason to hold companions.
+        // Undo the pierce before the context is stored — otherwise page 1
+        // honours #309 and pages 2..n ship the companion in `_source`.
+        if pierce_default_source {
+            for (idx_name, h) in snap.iter_mut() {
+                if let (Some(companions), Some(obj)) = (
+                    companions_by_index.get(idx_name.as_str()),
+                    h.source.as_object_mut(),
+                ) {
+                    obj.retain(|name, _| !companions.contains(name));
+                }
+            }
+        }
+        Some(snap)
     } else {
         None
     };
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -11971,6 +11971,22 @@ async fn search_impl(
                         !name.starts_with(xerj_query::executor::PASSAGE_METADATA_PREFIX)
                     });
                 }
+                // #310 emission site 1: this request pierced the default
+                // `_source` projection so the `fields` builder below could
+                // resolve an embedding companion the caller explicitly named.
+                // The pierce ends HERE — `fields` reads `h.source`, which is
+                // untouched, while the wire `_source` carries exactly what the
+                // #309 default would have carried. Miss this and #309's headline
+                // guarantee turns into a full-vector response for anyone who
+                // named a companion in `fields`.
+                if pierce_default_source {
+                    if let (Some(companions), Some(obj)) = (
+                        companions_by_index.get(idx_name.as_str()),
+                        s.as_object_mut(),
+                    ) {
+                        obj.retain(|name, _| !companions.contains(name));
+                    }
+                }
                 // Non-synthetic mode: strip the internal copy-to
                 // tracking marker; keep the copied values in the source
                 // since stored-source clients expect them.
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -34424,7 +34424,17 @@ pub(crate) fn build_docvalue_fields(
         } else {
             arr
         };
+        // A field that resolves to NO values is omitted, never emitted as a bare
+        // `[]` (#310): `{"body_vector": []}` is a positive claim that the doc has
+        // no values for that field. ES omits an unresolvable `docvalue_fields`
+        // entry entirely (mirrors the `fields` builder's `None => continue`).
+        if formatted.is_empty() {
+            continue;
+        }
         map.insert(field_name.to_string(), Value::Array(formatted));
+    }
+    if map.is_empty() {
+        return None;
     }
     Some(map)
 }
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -5992,6 +5992,47 @@ fn parse_time_to_millis(s: &str) -> Option<u64> {
     Some(n.saturating_mul(mul_ms).max(1))
 }
 
+/// Every field name this request pulls a VALUE out of the stored document for
+/// by a route other than `_source` — the `fields` clause and `docvalue_fields`.
+///
+/// Used once per request, before the engine runs, to decide whether the default
+/// `_source` projection has to be pierced (#310). It deliberately re-parses the
+/// two clauses instead of the response-time lists built later in `search_impl`:
+/// the decision has to be taken BEFORE the search, and the URL-param forms
+/// (`?fields=`, `?docvalue_fields=`) have already been promoted into `body` by
+/// then, so this sees exactly what the response layer will.
+fn value_bearing_field_names(body: &EsSearchBody, fields: &[String]) -> Vec<String> {
+    fn push_names(v: Option<&Value>, out: &mut Vec<String>) {
+        match v {
+            Some(Value::Array(arr)) => {
+                for entry in arr {
+                    match entry {
+                        Value::String(s) => out.push(s.clone()),
+                        Value::Object(o) => {
+                            if let Some(name) = o.get("field").and_then(Value::as_str) {
+                                out.push(name.to_string());
+                            }
+                        }
+                        _ => {}
+                    }
+                }
+            }
+            Some(Value::String(s)) => out.push(s.clone()),
+            _ => {}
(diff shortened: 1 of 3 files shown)


xerj-org:

Fix tip `1f6d9204`. **Result: 3× CONFIRMED_SOUND, ZERO BLOCKING.** Each ran the 6 tests (all pass) and proved fail-before by reverting individual source hunks. Highlights: no FALSE pierce (a non-companion `fields:["title"]` under default `_source` leaves `_source` byte-identical); no over-strip (mixed `fields:["body_vector","title"]` keeps title+body, strips only the companion); explicit `_source` untouched; `build_docvalue_fields` omits only genuinely-empty results. Emission **site 1** and **site 3** are each independently load-bearing (reverting either fails exactly its guard test). The collapse `inner_hits` leak (site 2) is confirmed **pre-existing and independent of the pierce** (a control probe with no `fields` clause leaks the identical vector) → correctly deferred to **#646**.

**Non-blocking follow-up noted:** on a pierced request the `savings` PLG stat reports 0 (the engine sees `Enabled(true)` and reports no omission; the API's emission-time re-narrowing isn't re-counted) — an under-report, no correctness impact.

Could not refute; every claimed emission site is load-bearing and the scope claim holds.

Setup: git reset --hard 1f6d9204 -> HEAD == 1f6d9204d1ae2fee3c65694574de11389952e854. Fix symbols present: value_bearing_field_names (es_compat.rs:6004), pierce_default_source (10887/11766/12000), embedding_companion_fields (index.rs:8417). Tests: RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-api --test fields_pierce_default_source -> 6 passed, 0 failed.

EVIDENCE checks:
(1) Site 1 (es_compat.rs:12000-12007 top-level hit _source strip): removed ONLY that if pierce_default_source{obj.retain(...)} block -> piercing_for_fields_does_not_widen_source_on_the_wire FAILED at test line 155 with body_vector AND body_vector_chunks leaking into _source (docvalue + scroll page 1 also failed, consistent: page 1 renders via site 1). Reverted via git checkout --; tree clean.
(2) Site 3 (es_compat.rs:11766-11775 scroll-snapshot strip): removed ONLY that block, site 1 intact -> EXACTLY scroll_continuation_does_not_leak_a_companion_after_a_pierce FAILED at test line 353 (page 2 continuation leaked body_vector); page 1 + other 5 tests passed, isolating site 3 to continuation pages. Reverted; tree clean.
(3) Scope/#646: only write to companions_by_index is the .insert at line 10880, gated by `if default_source_projection && !value_bearing_names.is_empty()` (line 10873). value_bearing_field_names (6004-6034) reads ONLY body.fields + body.docvalue_fields, never collapse/inner_hits -> a bare collapse yields empty value_bearing_names -> guard false -> companions_by_index empty -> pierce_default_source false. Collapse inner_hits (site 2) is genuinely unpierced, matching the PR's #646 carve-out.

Both source-hunk reverts make a #310 test FAIL (fail_before_holds). Final: git status --porcelain empty, HEAD at 1f6d9204. No BLOCKING findings.

</details>

Setup: git reset --hard to 1f6d9204d1ae2fee3c65694574de11389952e854; HEAD confirmed (head_ok). Fix symbols present: value_bearing_field_names + pierce_default_source in es_compat.rs (6004, 10864, 10887-10892, sites at 11766 & 12000); embedding_companion_fields accessor in index.rs:8417. All 6 fields_pierce_default_source tests pass under rustup 1.97.1 (tests_pass). Worktree left clean at target commit; both throwaway edits reverted via git checkout --.

REGRESSION checks (all 4 pass):
(1) Fail-before PIERCE: forced pierce_default_source to always false (&& false) -> fields_resolves_a_named_companion_under_every_source_spelling FAILS (fields.body_vector empty, left 0 != right 32), plus 3 companion tests. Reverted; re-green.
(2) Fail-before docvalue omit: removed the `if formatted.is_empty() { continue; }` guard so unresolvable fields insert Value::Array([]) -> docvalue_fields_never_emits_a_bare_empty_array FAILS (response carries "no_such_field":[]). Reverted; re-green.
(3) Green-stays: cargo test -p xerj-api --lib = 225 passed / 0 failed. Companion/source-projection integration tests: fields_pierce_default_source 6, reindex_preserves_generated_vectors 2, savings_stat_is_honest 15, scoped_keys_get_intact_responses 7 -- all pass. No test expected a bare [] from docvalue_fields (only the new test references it, asserting omission). No other integration test file references docvalue; only fields_pierce and 2 others reference the generated _vector companion.
(4) Index::embedding_companion_fields (index.rs:8417-8420) delegates to generated_embedding_companion_fields(&schema.schema) under self.schema.read().await -- byte-identical lock+arg+fn pattern to the internal call site at index.rs:10858-10861 (and the other 4 call sites). It adds only a read-lock; returns the identical HashSet. Confirmed.

Non-blocking observation (not a regression of tested behavior): the PLG `savings` stat is sourced from the engine's result.savings.per_hit (es_compat.rs:11110-11114). On a pierced request source is set to Enabled(true), so the engine returns intact source and reports no omission; the API's emission-time re-narrowing (sites 1/3) is not re-counted into savings_by_hit. A pierced request therefore reports savings=0 even though the wire _source still omits the companion. This affects only the newly-enabled pierce request shape, is a stat under-report (not over-report), does not touch data correctness or the #309 wire guarantee, and no existing test asserts it (savings_stat_is_honest's 15 tests never pierce). Worth a follow-up but does not block.

Verdict: refutation attempt failed -- the change is REGRESSION-sound. CONFIRMED_SOUND.

</details>

Attempted to refute; the change is sound within its declared scope. No in-scope leak or over-strip found.

Setup: HEAD==1f6d9204 (head_ok), fix symbols present in es_compat.rs + index.rs, all 6 fields_pierce_default_source tests pass under rustc 1.97.1. Fail-before: disabling ONLY the site-1 re-narrow hunk (`if false && pierce_default_source`) fails 3 #310 tests — the full body_vector + body_vector_chunks leak into top-level _source — proving the tests genuinely guard the fix. Worktree restored clean.

Lens checks (all verified with temporary probes, then removed):
(a) NO false pierce: fields:["title"] (non-companion) under default _source does NOT pierce; _source byte-identical to a plain default search (PROBE_A_SOURCE_EQ=true). Pierce condition is exact-string membership of a value-bearing name in the per-index companion set, so a non-companion name cannot trigger it.
(b) NO over-strip: mixed fields:["body_vector","title"] keeps title+body in _source, strips only body_vector, and resolves both in `fields` (vec len 32). generated_embedding_companion_fields() only ever inserts target_field + {target}_chunks — never the source text field or user keyword fields — so site-1 retain(|n| !companions.contains(n)) can only drop generated companions.
(c) explicit _source untouched: default_source_projection = body.source.is_none(); Some(true)/{includes} short-circuits to Vec::new() value-bearing names → pierce=false, and site-1 re-narrow is gated on pierce. Some(true) case covered by the passing suite.
(d) build_docvalue_fields: keeps non-empty arrays and scalars (incl. false/0/""), omits ONLY Null/None/empty-array (genuinely no values) — matches ES semantics; no legitimate value dropped.

Engine-side #309 strip (apply_source_filter_measured, SourceFilter::Default) only retains top-level keys; Enabled(true) is a passthrough. Multi-index pierce is safe: source is widened globally but each participating index's hits are re-narrowed with ITS OWN companion set at site 1, and indices without companions have nothing to strip.

Out-of-scope note (NOT blocking, correctly deferred): collapse inner_hits (site 2, #646) DOES put body_vector + __xerj_passage_meta__body_vector on the wire in member _source. But a control probe with collapse+inner_hits and NO fields clause (no pierce at all) leaks the byte-identical vector (control_leaks=true == pierce_leaks=true). The leak is empirically PRE-EXISTING and independent of the pierce — apply_collapse plants raw stored member sources that the #309 strip never recursed into — so this PR neither introduces nor worsens it. The code comment claiming "not driven by the pierce" is accurate.

</details>

PR #650, opened 2026-08-23
fix(api): collapse inner_hits members honour the default _source projection (#646)

### LINKED ISSUE #646: collapse inner_hits leak generated embedding companions under the default _source (follow-up to #310)

xerj-org:
Closes #646 (the primary, non-scroll case). Follow-up to #310.

## Problem
A collapse `inner_hits` member's source is stashed **before** the engine's `_source` projection runs (`apply_collapse`), so each member shipped the generated embedding companions (`*_vector`, `*_vector_chunks`) **and** the internal `__xerj_passage_meta__*` sentinels in its `_source` — even though the collapse **leader** honours #309. (Confirmed pre-existing and independent of the #310 pierce by that PR's own verification.)

## Fix (`render_collapse_inner_hits`)
- **Passage sentinels** (`__xerj_passage_meta__*`) are now stripped from every member `_source` **unconditionally**, matching the top-level hit rendering.
- A new `strip_companions: Option<&HashSet<String>>` param drops the generated companions from each member `_source` **under the default projection**. `companions_by_index` is repopulated for a bare collapse request (the exact clause #310 deferred to this issue).
- **Gated on `!is_scroll_request`** at the `search_impl` call site so a collapse **scroll** stays byte-identical across pages (guarded by #623's `collapse_scroll_page_parity`, which stays green): the scroll continuation (`scroll_page_response`) cannot strip companions yet — it would need the set threaded through `ScrollContext` — so **scroll+collapse companion stripping is a scoped remaining sub-case** (the passage-sentinel strip still applies to it).

## Testing (rustc 1.97.1)
- New `collapse_inner_hits_no_companion_leak.rs` — **fail-before**: on `main` the member `_source` leaks `body_vector`; with the fix the leader and members both omit it.
- `#623`'s `collapse_scroll_page_parity` (page1==page2 byte-identity) stays green — the `!is_scroll_request` gate prevents a page divergence.
- Full `cargo test -p xerj-api` — **dev**: 56 binaries, 0 failures; **`--profile ci-test`**: 56 binaries, 0 failures.
- `cargo fmt --all --check` clean; `cargo clippy -p xerj-api --all-targets -- -D warnings` clean.

_Note: the Landing-constants guard will be red until this rebases onto post-rc.49-stamp main (#649); the rebase is landing-only, no code change._

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8497,6 +8497,7 @@ fn render_collapse_inner_hits(
     spec: Value,
     idx_name: &str,
     rest_total_hits_as_int: bool,
+    strip_companions: Option<&std::collections::HashSet<String>>,
 ) -> Value {
     let spec_list: Vec<Value> = match spec {
         Value::Array(a) => a,
@@ -8615,6 +8616,18 @@ fn render_collapse_inner_hits(
                 if let Some(src) = m.get_mut("_source").and_then(Value::as_object_mut) {
                     src.remove("__xy_collapse_group__");
                     src.remove("__xy_collapse_spec__");
+                    // #646: a member's source is stashed BEFORE the engine's
+                    // `_source` projection, so it still carries the passage
+                    // sentinels (always internal) and, under the default
+                    // projection, the generated embedding companions. Match the
+                    // top-level hit rendering: strip the sentinels unconditionally
+                    // and the companions when the caller passed the set.
+                    src.retain(|k, _| {
+                        !k.starts_with(xerj_query::executor::PASSAGE_METADATA_PREFIX)
+                    });
+                    if let Some(companions) = strip_companions {
+                        src.retain(|k, _| !companions.contains(k));
+                    }
                 }
                 let mut hit_obj = serde_json::Map::new();
                 hit_obj.insert("_index".to_string(), Value::String(idx_name.to_string()));
@@ -10866,11 +10879,12 @@ async fn search_impl(
         Vec::new()
     };
     // One schema read-lock per participating index, and only for a request that
-    // could actually consume the answer. (The collapse `inner_hits` companion
-    // leak under a default projection is a separate, pre-existing gap tracked as
-    // a follow-up; it is not driven by the pierce, so it is not populated here.)
+    // could actually consume the answer: a pierce (a companion named in
+    // `fields`/`docvalue_fields`) OR a collapse `inner_hits` block, whose members
+    // are stashed before the #309 projection and would otherwise ship the
+    // companions on the wire (#646).
     let mut companions_by_index: HashMap<String, HashSet<String>> = HashMap::new();
-    if default_source_projection && !value_bearing_names.is_empty() {
+    if default_source_projection && (!value_bearing_names.is_empty() || body.collapse.is_some()) {
         for idx_name in &index_names {
             let Ok(idx) = state.engine.get_index(idx_name) else {
                 continue;
@@ -13600,6 +13614,18 @@ async fn search_impl(
                     spec,
                     &idx_name,
                     params.rest_total_hits_as_int.as_deref() == Some("true"),
+                    // #646: under the default projection, strip the generated
+                    // companions from each member's `_source` — the members were
+                    // stashed before #309 ran. Gated on `!is_scroll_request` so a
+                    // collapse SCROLL stays byte-identical across pages (#623):
+                    // the continuation path (scroll_page_response) cannot strip
+                    // companions yet, so page 1 must not either — scroll+collapse
+                    // companion stripping is the clean remaining sub-case.
+                    if default_source_projection && !is_scroll_request {
+                        companions_by_index.get(idx_name.as_str())
+                    } else {
+                        None
+                    },
                 ));
             }
 
@@ -20733,6 +20759,13 @@ async fn scroll_page_response(
                                     s,
                                     &h.index,
                                     rest_total_hits_as_int,
+                                    // The scroll continuation has no companion set
+                                    // in scope (it would need threading through
+                                    // ScrollContext); the unconditional passage
+                                    // sentinel strip still applies. Scroll+collapse
+                                    // companion stripping is a scoped remaining
+                                    // sub-case of #646.
+                                    None,
                                 )),
                                 _ => None,
                             }
--- a/engine/crates/xerj-api/tests/collapse_inner_hits_no_companion_leak.rs
+++ b/engine/crates/xerj-api/tests/collapse_inner_hits_no_companion_leak.rs
@@ -1,0 +1,133 @@
+//! Issue #646 (follow-up to #310): a collapse `inner_hits` block must not ship
+//! the generated embedding companions in each member's `_source` under the
+//! default projection.
+//!
+//! The collapse machinery stashes each group member's source BEFORE the engine's
+//! `_source` projection runs, so the members carry `*_vector` / `*_vector_chunks`
+//! even though #309 keeps them off the top-level `_source`. The leader honours
+//! #309; the `inner_hits` underneath it leaked the vectors.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+use xerj_common::types::{EmbeddingConfig, FieldConfig, FieldType, Schema};
+
+const DIMS: usize = 32;
+
+async fn seeded_app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+
+    let mut schema = Schema::empty();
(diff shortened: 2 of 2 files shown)


xerj-org:

Fix tip `57272579`. **Result: 3× CONFIRMED_SOUND, ZERO BLOCKING.** No user-field over-strip (companions come only from `generated_embedding_companion_fields`; sentinel strip matches only the literal `__xerj_passage_meta__` prefix); a non-default request leaves member companions intact; leader and members agree under the default projection. Fail-before holds (member `_source` leaks `body_vector` without the fix). The passage-sentinel strip and the companion strip are proven **distinct** load-bearing paths.

**Two non-blocking follow-ups noted:**
1. The `!is_scroll_request` gate (which keeps a collapse SCROLL byte-identical across pages) is load-bearing but not directly guarded — the #623 parity corpus has no embedding companion, so removing the gate doesn't fail it. A test-coverage improvement.
2. An explicit `_source: {"includes": [...]}` that omits the collapse sentinel drops the leader's inner_hits entirely (the engine projects `h.source` before the sentinel is extracted) — **pre-existing**, untouched by this diff, filed separately.

Attempted to REFUTE the CORRECTNESS lens; could not. HEAD==572725793460be8c187647ac4074d1b9b0b3ceb7 (head_ok). Fix present in engine/crates/xerj-api/src/es_compat.rs: render_collapse_inner_hits gains strip_companions: Option<&HashSet<String>> (line 8500), unconditional PASSAGE_METADATA_PREFIX strip + gated companion strip (lines 8625-8630); companions_by_index populated for bare collapse via `|| body.collapse.is_some()` (line 10887); search_impl call site gates on default_source_projection && !is_scroll_request (line 13624); scroll_page_response passes None (line 20768).

Named tests both PASS under rustup 1.97.1: collapse_inner_hits_no_companion_leak (1) + collapse_scroll_page_parity (1).

Fail-before HOLDS: neutralizing the companion strip (`if let Some(companions){ let _=companions; }`) makes collapse_inner_hits_no_companion_leak FAIL at line 124 — member _source leaks body_vector — while user fields body/grp remain present (not vacuous). Restored fix via git checkout --; worktree clean, HEAD unchanged.

Lens checks (all wrote+ran a throwaway probe, then reverted):
(a) NO user-field over-strip — CONFIRMED. Probe with members carrying grp/body/title kept all three; companions set comes from generated_embedding_companion_fields (index.rs:26506), which only inserts {target} and {target}_chunks for fields with an embedding config (never arbitrary user fields), and the sentinel strip only matches the literal `__xerj_passage_meta__` prefix (executor.rs:28). Neither can match ordinary user fields.
(b) NON-default request does NOT strip companions — CONFIRMED. Under explicit `_source: true`, leader and member BOTH carry body_vector (default_source_projection=false ⇒ strip_companions=None; companions_by_index isn't even populated because body.source is Some).
(c) leader/member AGREE under default — CONFIRMED. Under the default projection both omit body_vector and body_vector_chunks.

Non-blocking observation (out of scope for #646, PRE-EXISTING): under an explicit `_source:{"includes":[...]}` that omits the collapse sentinel, the leader loses inner_hits entirely — the engine projects `h.source` down to the listed fields, dropping `__xy_collapse_group__` before it is extracted at es_compat.rs:11943. The PR diff (git show HEAD) does not touch that extraction path or the engine-side projection, and companions_by_index is not populated on that path, so this is not a #646 regression, not a user-field over-strip, and not a #646 leader/member disagreement. Would be an ES-divergence worth its own follow-up, but does not block this PR.

</details>

HEAD pinned at 572725793460be8c187647ac4074d1b9b0b3ceb7 (head_ok). Fix present in engine/crates/xerj-api/src/es_compat.rs: render_collapse_inner_hits gains strip_companions: Option<&HashSet<String>> (L8500); unconditional PASSAGE_METADATA_PREFIX sentinel strip on every member (L8625-8627) + conditional companion strip (L8628-8630); companions_by_index repopulated for a bare collapse via `|| body.collapse.is_some()` (L10887); search_impl call site gated `default_source_projection && !is_scroll_request` (L13624); scroll_page_response passes None (L20768).

TESTS: collapse_inner_hits_no_companion_leak + collapse_scroll_page_parity both pass. Full `cargo test -p xerj-api --lib` = 225 passed, 0 failed (rustc 1.97.1).

FAIL-BEFORE (holds): `git checkout main -- es_compat.rs` (removes fix, keeps test) → leak test FAILS; member _source leaked body_vector AND __xerj_passage_meta__body_vector. Restored via `git checkout HEAD --` (note: plain `git checkout --` restores from the index, which after `checkout main --` still held main's staged copy — had to use HEAD explicitly).

BYTE-IDENTITY GUARD: collapse_scroll_page_parity stays green. Full lib suite green. No regression.

NON-BLOCKING FINDING (test-coverage gap, not a fix defect): REGRESSION check #2's expected outcome did NOT materialize. I temporarily removed `&& !is_scroll_request` (so scroll page 1 strips companions while page 2 does not) — the task predicted collapse_scroll_page_parity would then FAIL. It stayed GREEN. Root cause: that parity test's index has only grp(keyword)+v(long), no embedding companion fields, so companions_by_index.get("docs") is None with or without the gate — the strip is a no-op there. The parity test therefore pins the shared render path (sort/size/from/fields/seq_no/version) but does NOT exercise the companion-strip divergence the gate exists to prevent; the !is_scroll_request gate is load-bearing yet untested. The fix itself remains correct: non-scroll collapse strips companions (proven by leak test + fail-before), scroll collapse leaves them on both pages byte-identically (documented scoped-remaining sub-case per #623). Verdict CONFIRMED_SOUND; recommend (non-blocking) a follow-up adding a companion field to the parity index so the gate is actually guarded. Worktree left clean at target HEAD.

</details>

Ran: (1) git fetch + reset --hard 5727257; git rev-parse HEAD == 572725793460be8c187647ac4074d1b9b0b3ceb7 (head_ok). (2) grep confirmed fix symbols in engine/crates/xerj-api/src/es_compat.rs: render_collapse_inner_hits(...strip_companions: Option<&HashSet<String>>) @8495-8500, unconditional sentinel strip @8625-8627, companion strip @8628-8630, guard @10887, search_impl call site @13612 (gated `default_source_projection && !is_scroll_request`), scroll call site @20757 (passes None). (3) RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-api --test collapse_inner_hits_no_companion_leak --test collapse_scroll_page_parity: both pass (1+1). (4) After: rm probe, git checkout -- ., HEAD clean at 5727257, empty working tree.

EVIDENCE lens — all three claims HOLD; could not refute:

CLAIM 1 (passage-sentinel strip real): PROVEN by a black-box probe + fail-before. Note the sentinel cannot be injected via ingest (engine rejects user-submitted `__xerj_passage_meta__*`: "engine-owned passage metadata"). It is instead written UNCONDITIONALLY by the embed job (index.rs:12164) whenever chunks embed, into the durable stored source — and always stripped from top-level responses (so `_source:true` shows [body,grp,body_vector,body_vector_chunks], never the sentinel). It IS stashed into the collapse `__xy_collapse_group__`. Fail-before: commented out ONLY the sentinel `src.retain(...PASSAGE_METADATA_PREFIX)` line (8625-27), rebuilt, ran collapse probe -> member `_source` leaked `__xerj_passage_meta__body_vector` with chunks [[0,442]...]. Restored -> member `_source` == [body,grp], and raw wire bytes contain no `__xerj_passage_meta__`. In the leak case companions (body_vector/_chunks) were STILL absent, proving sentinel and companion strips are distinct code paths and the sentinel strip is the load-bearing one.

CLAIM 2 (companions_by_index repopulated for bare collapse): CONFIRMED. Guard at es_compat.rs:10887 is verbatim `default_source_projection && (!value_bearing_names.is_empty() || body.collapse.is_some())`. For a bare collapse (no _source clause, no fields/docvalue_fields) value_bearing_names is empty so the `body.collapse.is_some()` arm fires. Empirically corroborated: bare-collapse probe at HEAD stripped body_vector/body_vector_chunks from members (companions populated AND passed to render).

CLAIM 3 (scroll passes None): CONFIRMED. es_compat.rs:20757-20769 passes None for strip_companions with an in-code comment that scroll+collapse companion strip is a scoped remaining sub-case of #646; continuation relies only on the unconditional sentinel strip. collapse_scroll_page_parity green (page1==continuation, byte-identical per #623).

Reinforcing: generated_embedding_companion_fields (index.rs:26506, the source feeding companions_by_index via embedding_companion_fields) returns only {target, target_chunks} and NOT the sentinel, so the unconditional sentinel strip is genuinely required — not redundant with companion stripping — for both non-default-projection collapse and the scroll continuation. No blocking findings.

</details>

PR #652, opened 2026-08-23
fix(engine): project _source filters inside arrays of objects (#644)

### LINKED ISSUE #644: _source exclude cannot strip fields inside an array of objects (follow-up to #602)
Split out of the #602 (#641) verification. XERJ's `_source` filter does not project *inside arrays* — an array value is kept whole whenever its key is selected (`index.rs`, `Value::Array` arm of `es_filter`, documented as the no-array-projection scope). A consequence surfaced by the 3-skeptic pass:

```
filter_object({"logs":[{"msg":1,"api_secret":2}], "name":3}, includes=[], excludes=["*secret"])
  => {"logs":[{"msg":1,"api_secret":2}], "name":3}   // api_secret LEAKS
ES (makeAnyString) => {"logs":[{"msg":1}], "name":3}  // api_secret stripped from each element
```

This is **pre-existing** (not introduced by #602 — it predates the automaton and matches prior XERJ behavior) and out of scope for #602, which is about object-nested path globs. But it is a real ES divergence with a security edge: `_source.excludes: ["*secret"/"*password"/"*token"]` does not scrub sensitive fields that live inside an array of objects.

**Fix direction:** in the `Value::Array` arm, when the key is selected but the array holds objects and an include/exclude is *live* at this path, map each object element through `es_filter` at the same path (ES applies the filter element-wise). Keep scalar array elements handled by the accept/live rules. This is a larger change (array element recursion) than #602 took on, hence tracked separately.

xerj-org:
Closes #644.

## Problem
ES treats an array index as **transparent** to a `_source` path: `tags.name` matches `tags[i].name`, so `include ["tags.name"]` keeps only each element's `name` and `exclude ["logs.secret"]` scrubs `secret` from every element. The #602 automaton deliberately kept arrays **whole** (a no-projection scope), which:
- **over-includes** on `include` — `include ["tags.name"]` returned the whole `[{"name":..,"id":..}]` array (the `id` leaks);
- **fails to strip** on `exclude` — `exclude ["*.secret"]` / `["logs.secret"]` left secrets inside arrays of objects on the wire (a data-over-exposure edge).

## Fix
`es_filter`'s array arm now mirrors the object recursion:
- a **fully-included** array with no exclude reaching into it is kept whole (the fast path — unchanged, byte-identical, no #602 data-loss);
- otherwise each element is filtered at the **same** `child` path — an **object** element recurses through `es_filter`, a **scalar** element survives only in an accepting (fully-included) state, like a scalar leaf.

## Testing (rustc 1.97.1)
- New `source_filter_projects_inside_arrays_of_objects` — **fail-before**: on `main` the array was returned whole (`id`/`secret` leaked); with the fix `include tags.name` → `[{name}]` per element and `exclude logs.secret` → `[{msg}]` per element.
- All 14 `source_filter_tests` green (the new test + every existing #602/#310 test — the fully-included fast path preserves prior behavior).
- Full `cargo test -p xerj-api` — dev: 55 binaries, 0 failures.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -26697,13 +26697,28 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
                         out.insert(key.clone(), filtered);
                     }
                 }
-                // An array is kept WHOLE whenever its key is selected: XERJ does
-                // not project inside arrays (documented scope), which also keeps
-                // it free of the projection-induced data loss #602 must avoid.
-                // (Consequence: an exclude cannot strip fields inside an array of
-                // objects — tracked separately, not a #602 regression.)
-                Value::Array(_) => {
-                    out.insert(key.clone(), val.clone());
+                // ES treats an array index as transparent to a `_source` path,
+                // so a pattern projects INTO each element (#644): `tags.name`
+                // matches `tags[i].name`. A fully-included array with no exclude
+                // reaching into it is kept whole (the fast path); otherwise each
+                // element is filtered at the SAME `child` path — an object
+                // element recurses, a scalar element survives only in an
+                // accepting state (like a scalar leaf).
+                Value::Array(arr) => {
+                    if child_all && !exc_live {
+                        out.insert(key.clone(), val.clone());
+                    } else {
+                        let projected: Vec<Value> = arr
+                            .iter()
+                            .filter_map(|elem| match elem {
+                                Value::Object(_) => {
+                                    Some(es_filter(elem, &child, child_all, includes, excludes))
+                                }
+                                other => child_all.then(|| other.clone()),
+                            })
+                            .collect();
+                        out.insert(key.clone(), Value::Array(projected));
+                    }
                 }
                 // A scalar leaf is kept ONLY in an ACCEPTing state. A scalar that
                 // merely sits on a LIVE (not-yet-accepted) include path cannot
@@ -27115,6 +27130,35 @@ mod source_filter_tests {
             ),
             json!({ "keep": 2, "nested": { "host": 4 } }),
             "dot-less `exclude *secret` must strip `password_secret` and nested `api_secret` (#602)"
+        );
+    }
+
+    /// #644: ES projects INSIDE an array of objects — the array index is
+    /// transparent to a `_source` path, so `include ["tags.name"]` keeps only
+    /// each element's `name`, and `exclude ["logs.secret"]` scrubs `secret` from
+    /// every element. XERJ currently keeps the array whole (the #602 no-array-
+    /// projection scope), leaking the un-selected fields.
+    #[test]
+    fn source_filter_projects_inside_arrays_of_objects() {
+        // include: keep only the named leaf inside each array element.
+        assert_eq!(
+            filter_object(
+                &json!({ "tags": [{ "name": "x", "id": 1 }, { "name": "y", "id": 2 }], "n": 9 }),
+                &["tags.name".to_string()],
+                &[]
+            ),
+            json!({ "tags": [{ "name": "x" }, { "name": "y" }] }),
+            "#644: include tags.name must project into each array element"
+        );
+        // exclude: scrub the named leaf from each array element.
+        assert_eq!(
+            filter_object(
+                &json!({ "logs": [{ "msg": 1, "secret": 2 }, { "msg": 3, "secret": 4 }] }),
+                &[],
+                &["logs.secret".to_string()]
+            ),
+            json!({ "logs": [{ "msg": 1 }, { "msg": 3 }] }),
+            "#644: exclude logs.secret must strip secret from every array element"
         );
     }
 
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -26711,13 +26711,31 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
                         let projected: Vec<Value> = arr
                             .iter()
                             .filter_map(|elem| match elem {
-                                Value::Object(_) => {
-                                    Some(es_filter(elem, &child, child_all, includes, excludes))
+                                Value::Object(child_obj) => {
+                                    let f = es_filter(elem, &child, child_all, includes, excludes);
+                                    match &f {
+                                        // An element that *filtering* emptied is
+                                        // dropped from the array; one that was
+                                        // ALREADY `{}` survives only in an
+                                        // accepting state — same rule ES applies
+                                        // to array elements as to map values.
+                                        Value::Object(m) if m.is_empty() => {
+                                            (child_obj.is_empty() && child_all).then_some(f)
+                                        }
+                                        _ => Some(f),
+                                    }
                                 }
                                 other => child_all.then(|| other.clone()),
                             })
                             .collect();
-                        out.insert(key.clone(), Value::Array(projected));
+                        // A key whose array *filtering* emptied is pruned (had
+                        // elements, all dropped); an already-empty source array
+                        // survives only in an accepting state — mirrors the
+                        // object arm so an emptied array does not leave a bare
+                        // `[]` (or a `[{}]`) on the wire.
+                        if !projected.is_empty() || (arr.is_empty() && child_all) {
+                            out.insert(key.clone(), Value::Array(projected));
+                        }
                     }
                 }
                 // A scalar leaf is kept ONLY in an ACCEPTing state. A scalar that
@@ -27159,6 +27177,36 @@ mod source_filter_tests {
             ),
             json!({ "logs": [{ "msg": 1 }, { "msg": 3 }] }),
             "#644: exclude logs.secret must strip secret from every array element"
+        );
+        // Emptied elements are pruned, not left as `{}`; a heterogeneous array
+        // drops the elements (and scalars) that do not match, keeping the rest.
+        assert_eq!(
+            filter_object(
+                &json!({ "tags": [{ "name": "x", "id": 1 }, { "id": 2 }, "loose"] }),
+                &["tags.name".to_string()],
+                &[]
+            ),
+            json!({ "tags": [{ "name": "x" }] }),
+            "#644: an emptied array element (and a scalar) is dropped, not kept as an empty object"
+        );
+        // An array all of whose elements empty → the key is pruned entirely.
+        assert_eq!(
+            filter_object(
+                &json!({ "tags": [{ "id": 1 }, { "id": 2 }], "keep": 5 }),
+                &["tags.name".to_string()],
+                &[]
+            ),
+            json!({}),
+            "#644: an array that filtering emptied drops its key (no bare [])"
+        );
+        assert_eq!(
+            filter_object(
+                &json!({ "tags": [{ "secret": 1 }, { "secret": 2 }] }),
+                &[],
+                &["tags.secret".to_string()]
+            ),
+            json!({}),
+            "#644: an exclude that empties every element prunes the key"
         );
     }
 


xerj-org:
## Verification round 1 (@ `94a41ead`), 3 skeptics under rustc 1.97.1

**correctness REFUTED (1 BLOCKING); regression + evidence CONFIRMED_SOUND.** The projection is correct for non-emptying cases (fast path byte-identical, 625 lib tests, no data-loss, no panic, scalar gating correct), but it does **not prune emptied array elements**: an object element that filters to `{}` is kept, and an array that empties still emits its key — ES drops both (`XContentMapValues.filter`). **Fixing (not merging):** mirror the object arm's prune in the array arm.

### correctness — `REFUTED`
- tests_pass=True · fail_before_holds=False · es_divergence_found=True

**BLOCKING:** ES-divergence / over-include: emptied array-of-object elements are not pruned. In es_filter's array arm (engine/crates/xerj-engine/src/index.rs ~26707-26721), when child_all is false the code does Some(es_filter(elem,...)) for every object element and always inserts the array key. If an object element filters to an empty {} it is KEPT, and an array that empties still inserts its key. ES's XContentMapValues.filter drops emptied map elements from arrays (if(filteredValue.isEmpty()==false) filtered.add(...)) and drops the resulting empty array key. Reproduced by probe: include ["tags.name"] on {"tags":[{"name":"x","id":1},{"id":2},"scalar"]} yields impl {"tags":[{"name":"x"},{}]} but ES yields {"tags":[{"name":"x"}]} (spurious empty object alongside real data); include ["tags.name"] on {"tags":[{"id":1},{"id":2}]} yields {"tags":[{},{}]} vs ES {}; exclude ["tags.secret"] on {"tags":[{"secret":1},{"secret":2}]} yields {"tags":[{},{}]} vs ES {} (exclude-mode structural leak of array length). Fix: mirror the object arm's prune (lines 26688-26698) - drop object elements whose filtered result is an empty object, and if the projected array is empty (and source array was non-empty) drop the key. This also removes the inconsistency where scalar arrays yield [] but object arrays yield [{},{}], failing lens case (e)'s 'leaves [] or drops the key consistently'.

Setup verified: git reset --hard 94a41ead, HEAD=94a41ead5788706f215e8f967f3c2fea2b124a1c (head_ok). Fix present: array arm at index.rs:26707 with 'projects INTO each element' comment at :26701. Tests: 14/14 source_filter_tests pass under rustup 1.97.1, RUST_MIN_STACK=8388608. fail_before_holds set false because this CORRECTNESS lens did not run a fail-before (stash+re-test) check; steps 1-4 did not request it. Adversarial probe (added #[cfg(test)] adversarial_probe_es_array_semantics, run with --nocapture, then reverted via git checkout -- leaving HEAD clean, probe grep count 0) confirmed: (a) fully-included array kept byte-identical incl mixed obj+scalar elements = correct; (b) scalar array under include tags.name -> {\"tags\":[]} (ES drops key -> {}, but lens case (e) blesses 'leaves []'); (c) exclude *.secret strips per element = correct; (d) nested a.b.c depth projection = correct; (e1/e2/e3) array-of-objects whose elements filter to empty retain spurious {} placeholders and the array key = the blocking over-include/ES-divergence above. No data-loss (legit values dropped) was found; the fast path and exclude/include projections preserve all legit leaves. The defect is over-inclusion of empty array structure, reachable in the common heterogeneous-array case (e2), hence BLOCKING per the lens rubric.

</details>

HEAD=94a41ead (clean). Fix present at index.rs:26707-26721 (Value::Array(arr) arm: fast path child_all && !exc_live -> val.clone(); else per-element projection, object recurses via es_filter, scalar survives only in child_all). source_filter_tests 14/14 pass on rustc 1.97.1.

FAIL-BEFORE (adapted): the test source_filter_projects_inside_arrays_of_objects lives in index.rs itself and main has NO such test, so `git checkout main -- index.rs` would drop the test too. I instead surgically reverted ONLY the array arm to main's `Value::Array(_) => out.insert(key, val.clone())` while keeping the new test. Result: source_filter_projects_inside_arrays_of_objects FAILED (left {name,id} vs right {name} — array returned whole, id leaked); other 13 passed. The array arm is the sole cause. Restored with git checkout 94a41ead -- .; tree clean.

FAST PATH byte-identical: child_all && !exc_live branch is `out.insert(key.clone(), val.clone())`, identical to main's unconditional whole-array clone, no per-element rebuild. include ["tags"] (inc_acc=true) and match-all default (include_all=true) both take it. Fast path is a strict subset of main's always-keep-whole; only live-prefix and exclude-reaching-in cases now project (exactly #644). No fast-path regression, no data loss.

FULL LIB SUITE: cargo test -p xerj-engine --lib = 625 passed, 0 failed. #602/#310 source-filter tests green — array change does not regress them.

Non-blocking (outside REGRESSION scope): projection inserts empty [] when it strips all elements (e.g. include ["tags.name"] over an array of scalars), whereas the object arm prunes a filtering-emptied object — a minor array/object asymmetry vs possible ES key-omission. Not a regression of prior behavior, not on the fast path. Could refute nothing; fix is sound for this lens.

</details>

Tasked to REFUTE under the EVIDENCE lens; could not refute. All three lens claims hold.

WHAT I RAN:
1. `git fetch origin -q; git reset --hard 94a41ead` -> HEAD = 94a41ead5788... (head_ok). Reverted clean at end (git status --porcelain empty, fix still present).
2. Fix confirmed at engine/crates/xerj-engine/src/index.rs es_filter array arm (line ~26707). NB: the literal grep string 'project INTO each element' does NOT match; actual comment is "...pattern projects INTO each element (#644)". The Value::Array(arr) arm and the projection code are present and correct.
3. `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib source_filter_tests` -> 14 passed; 0 failed (tests_pass).
4. Added a throwaway adversarial test `adv_evidence_lens_array_arm` (reverted after) -> 15 passed; 0 failed.

CLAIM (1) array arm load-bearing — CONFIRMED. Fast path `if child_all && !exc_live` keeps a fully-included array WHOLE: include ["tags","other.x"] keeps tags[i].id (skips projection) while the dotted sibling other.x still projects. Partial include tags.name enters the else/projected branch and strips id. Fail-before: replacing the else branch with pre-#602 "keep whole" makes tags.name yield `{}` (array dropped, child_all=false) and both the shipped test and my adversarial test FAIL (fail_before_holds). So the projected branch is necessary.

CLAIM (2) no infinite recursion/panic — CONFIRMED. Deep nested arrays-of-objects (a.b.c) recurse to the leaf: {a:[{b:[{c:1,d:2}]}]} -> {a:[{b:[{c:1}]}]}. Mixed array [{name,id},"loose",42] with include tags.name -> {tags:[{name:x}]}: object element recurses, scalar siblings dropped. filter_map uses Some/None with no unwrap; recursion depth is bounded by JSON nesting (value shrinks each level). No panic/stack overflow (RUST_MIN_STACK=8M). Array-of-arrays elements fall to the `other` arm (kept-whole if child_all) — no recursion, no panic.

CLAIM (3) scalar element gating — CONFIRMED. In the projected branch, scalar element = `other => child_all.then(|| other.clone())`. Under a LIVE-but-non-accepting include (child_all=false): {tags:["a","b"],keep:{x:1}} + include [tags.name,keep.x] -> {tags:[],keep:{x:1}} — every scalar dropped (array collapses to empty). Under an ACCEPTING include (child_all=true, projection forced by exclude tags.secret reaching in): {tags:["a","b","c"],n:1} kept intact — scalars retained. In the fully-accepting no-exclude case scalars are kept whole via the fast path.

NON-BLOCKING OBSERVATION (not one of the three claims): when a live-but-non-accepting include drains a scalar-only array to empty, the empty-array key is RETAINED ({tags:[]}), i.e. there is no array-pruning analogous to the empty-object pruning at index.rs ~26688-26695. This is cosmetic (empty array vs omitted key), leaks no field values and loses no data; I could not verify real-ES behavior in this environment, so I do not assert it as a confirmed ES divergence. Not must-fix.

No BLOCKING findings.

</details>

xerj-org:
## Re-verification round 2 (emptied-element prune @ `8143df98`), 3 skeptics under rustc 1.97.1

**Result: 3× CONFIRMED_SOUND, ZERO BLOCKING** — the round-1 correctness finding is fixed. The prune is load-bearing (fail-before reproduces the spurious `{}`) and ES-faithful: emptied elements dropped, all-emptied arrays drop their key, a *genuine* already-`{}` element survives only in an accepting state (mirrors the #310 object-arm rule), real data never lost, 625 lib tests green. **One non-blocking follow-up:** a nested array-of-arrays-of-objects (`[[{name,id}]]`) is not projected (the leaf `other` arm treats a nested list as a scalar) — **pre-existing**, out of #644's flat-array scope, filed separately.

HEAD=8143df98 (clean after reverting throwaway edits). Fix markers present (index.rs: then_some prune @26723, key guard `arr.is_empty() && child_all` @26736, mirrored #310 comments). All 14 source_filter_tests pass on rustup 1.97.1.

CORRECTNESS refute attempt FAILED — added a #[cfg(test)] probe asserting the lens's exact expected outputs against the real impl; every case matched:
(a) include tags.name on [{name,id},{id},"scalar"] => {"tags":[{name:x}]} — no {}, scalar dropped.
(b) [{id},{id}] all-empty => {} (key dropped, no bare []).
(c) exclude tags.secret on [{secret},{secret}] => {} (all emptied, key pruned).
(d) MIXED [{name,id},{id}] => {"tags":[{name}]} (keep one, drop emptied).
(e) NON-fast-path already-empty accepted survives: [{}, {secret}, {a}] under exclude tags.secret => {"tags":[{}, {a}]} — the genuine source {} in an accepting state survives; the filtering-emptied {secret} is dropped. Discriminated correctly by `child_obj.is_empty() && child_all`.
(e') match-all keeps {"tags":[{}]} (fast path).
(f) fully-included array (include tags) kept byte-identical incl {} elements via val.clone().
Harder: nested-object projection (tags.meta.k) prunes the emptied element with NO spurious {} bubbling up; nested exclude (tags.meta.secret) empties element => key pruned to {}.

Fail-before verified: replacing the element-prune with unconditional Some(f) reproduces the round-1 spurious {} — existing test source_filter_projects_inside_arrays_of_objects fails with [{name:x}, {}]. Prune is load-bearing.

The prune is NOT a round-1 regression: round-1 {} was manufactured by filtering (pruned now); case (e) {} is genuine already-empty source content preserved only in an accepting state — mirrors the established/tested object-arm #310 rule.

OUT-OF-SCOPE, NON-BLOCKING observation: a nested array-of-arrays-of-objects ([[{name,id}]] under include tags.name) yields {} — impl drops it where strict ES would project to [[{name}]]. This is produced by the untouched `other => child_all.then(|| other.clone())` leaf arm (pre-existing one-level #644/#602 projection scope), NOT by the pruning logic under review, and is outside the mandated (a)-(f) checks. Not treated as blocking: reporting it would misattribute a pre-existing scope limitation to this pruning change. Candidate for a separate follow-up (nested-array transparency).

Verdict: pruning is ES-faithful for its scope; could not refute.

</details>

HEAD reset to 8143df98 (confirmed); fix present at index.rs:26722-26724 (object-element then_some prune) and 26736 (array-key !projected.is_empty() || (arr.is_empty() && child_all)). source_filter_tests: 14/14 pass. Full lib suite: 625 passed / 0 failed — no regression to #602/#310 or non-emptying #644 cases.

PRUNE-specific fail-before HOLDS: surgically reverting ONLY the prune (object-element branch forced to Some(f); array key always inserted) made source_filter_projects_inside_arrays_of_objects FAIL — emptied element kept as {} (left [{name:x},{}] vs right [{name:x}]). New assertions genuinely depend on the prune. Restored; tree clean.

Data-preservation adversarial check (throwaway test, then reverted): a non-empty array with mixed survivor/emptied elements keeps the survivors and the key ([{name:a},{id:9},{name:b,extra:0}] +include tags.name -> [{name:a},{name:b}]); exclude that empties only some elements keeps the rest whole; fully-included array with no exclude reaching in kept WHOLE via fast path. Prune fires only when EVERY element drops (projected empty), so no legitimately non-empty array is pruned — no data loss.

Worktree clean: git status empty, HEAD still 8143df98, throwaway edits reverted via git checkout --. Cannot refute; no BLOCKING findings.

</details>

- tests_pass=True · fail_before_holds=True · es_divergence_found=True

Ran: (1) git fetch + reset --hard 8143df98 -> HEAD 8143df981c58... (head_ok). (2) grep confirmed fix at index.rs:26689/26717/26723/26731/26736 ('filtering* emptied', '(child_obj.is_empty() && child_all).then_some', 'arr.is_empty() && child_all'). (3) RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib source_filter_tests -> 14 passed / 0 failed. (4) Appended a temp adversarial test (reverted after) that ran green on all three EVIDENCE claims:

CLAIM 1 (prune mirrors object arm) HOLDS. The element-prune predicate (child_obj.is_empty() && child_all) is byte-identical to the object arm's keep predicate. Verified same key drop/keep for {"a":{...}} vs {"a":[{...}]} under: filtering-emptied (include a.name over {id:1}) -> both {}; exclude-emptied (exclude a.secret) -> both {}; already-{} in accepting state (exclude a.b forces non-fast-path) -> {"a":{}} vs {"a":[{}]}. Same decision every time.

CLAIM 2 (no real data lost) HOLDS. Any object element whose filter returns non-empty hits the '_ => Some(f)' arm and is kept; a scalar element survives in an accepting state. Multi-leaf case {tags:[{name:a},{name:b,id:2},{id:3}]} include tags.name -> both leaf-bearing elements kept, only leafless dropped. The prune removes ONLY empty results.

CLAIM 3 (empty-array prune gated on non-empty source) HOLDS. Source [] survives in accepting state both via the fast path (child_all && !exc_live) and off-fast-path via the 'arr.is_empty() && child_all' disjunct; it drops only on a non-accepting live-only prefix (matches scalar-leaf semantics). Verified {"a":[]} -> {"a":[]} (fast + exc_live variants) and {"a":[]} under include a.b -> {}.

FAIL-BEFORE proven: reverting BOTH prune sites (element .then_some -> Some(f); key-insert guard -> unconditional) made source_filter_projects_inside_arrays_of_objects FAIL, producing {"tags":[{"name":"x"},{}]} -- the exact Round-1 spurious-empty defect. Prune is load-bearing. Restored via git checkout --; working tree clean, HEAD still 8143df98.

es_divergence_found=true but NON-BLOCKING: nested arrays of objects diverge from ES. filter_object({"a":[[{"name":"x","id":1}]]}, include a.name) returns {} because the array arm's 'other' branch treats a nested-array element as a scalar (dropped when child_all=false), whereas ES XContentMapValues recurses into nested lists and would keep {"a":[[{"name":"x"}]]}. This is PRE-EXISTING (the 'other' scalar branch predates this revision), is NOT caused by the new prune, and is outside #644's flat-arrays-of-objects scope -- so not a must-fix for this PR. Could not refute the three revision claims; verdict CONFIRMED_SOUND, no blocking items.

</details>

PR #655, opened 2026-08-23
fix(engine): preserve collapse sentinels across an explicit _source projection (#651)

### LINKED ISSUE #651: collapse: an explicit _source.includes that omits the collapse sentinel drops the leader's inner_hits
Surfaced by the 3-skeptic verification of #646 (PR #650), and **pre-existing** (untouched by that diff). When a collapse request sets an explicit `_source: {"includes": [...]}` that does not list the internal `__xy_collapse_group__` sentinel, the engine projects `h.source` down to the listed fields **before** the API extracts that sentinel (`es_compat.rs` ~11943), so the leader comes back with **no inner_hits at all**. ES renders inner_hits regardless of the top-level `_source` include list. Fix direction: extract/stash the collapse group sentinel before the engine's include-projection narrows the source (or carry it out-of-band), so an explicit include list cannot starve inner_hits. Separate from #646 (which is about companion leakage in members that DO render).

xerj-org:
Closes #651.

## Problem
A collapse request with an explicit `_source: {"includes": [...]}` (or `"excludes"`, or the array form) that does **not** list the internal `__xy_collapse_group__` sentinel lost the leader's `inner_hits` entirely. The engine's `apply_source_filter` projects `hit.source` (`SourceFilter::Includes` / `SourceFilter::Fields`) **before** the API extracts the collapse group (`es_compat.rs` ~11943), so the projection stripped the sentinel and the API found no group to render. ES renders `inner_hits` independent of the top-level `_source` include list.

## Fix
The two projecting arms of `apply_source_filter_measured` now **capture** the engine-internal collapse sentinels (`__xy_collapse_group__`, `__xy_collapse_spec__`) before `filter_object` runs and **re-attach** them after — so the API still extracts the group and strips the sentinels from the wire `_source` (as it already does). The savings stat is measured on the sentinel-free projection, so it is unaffected. Non-collapse requests carry no sentinels → the capture is a no-op.

## Testing (rustc 1.97.1)
- New `collapse_inner_hits_under_source_includes.rs` — **fail-before**: on `main` the leader came back `{_source:{grp:A}}` with no `inner_hits`; with the fix the leader renders `inner_hits.members`, the leader `_source` still narrows to `grp` (not `title`), and the sentinel does not leak on the wire.
- Full `cargo test -p xerj-api` — dev + `--profile ci-test`: 57 binaries each, 0 failures.

_(Landing-constants guard will be red until this rebases onto post-rc.50-stamp main (#654); the rebase is landing-only, no code change.)_

### DIFF
--- a/engine/crates/xerj-api/tests/collapse_inner_hits_under_source_includes.rs
+++ b/engine/crates/xerj-api/tests/collapse_inner_hits_under_source_includes.rs
@@ -61,7 +61,10 @@ async fn post(app: &axum::Router, path: &str, body: Value) -> (StatusCode, Value
     let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
         .await
         .expect("body");
-    (status, serde_json::from_slice(&bytes).unwrap_or(Value::Null))
+    (
+        status,
+        serde_json::from_slice(&bytes).unwrap_or(Value::Null),
+    )
 }
 
 /// A collapse `inner_hits` must render even under an explicit
--- a/engine/crates/xerj-api/tests/collapse_inner_hits_under_source_includes.rs
+++ b/engine/crates/xerj-api/tests/collapse_inner_hits_under_source_includes.rs
@@ -1,0 +1,111 @@
+//! Issue #651: a collapse `inner_hits` block must still render when the request
+//! sets an explicit `_source: {"includes": [...]}` that does not list the
+//! internal `__xy_collapse_group__` sentinel.
+//!
+//! The collapse group members are stashed into `__xy_collapse_group__` on the
+//! leader's source; the API extracts them at render time. But the engine applies
+//! the `_source` include projection to `hit.source` FIRST, dropping that
+//! sentinel when it is not in the include list — so the leader came back with no
+//! `inner_hits` at all. ES renders `inner_hits` regardless of the top-level
+//! `_source` include list.
+
+use axum::body::Body;
+use axum::http::{Request, StatusCode};
+use serde_json::{json, Value};
+use tower::ServiceExt;
+use xerj_common::types::{FieldConfig, FieldType, Schema};
+
+async fn app() -> (axum::Router, tempfile::TempDir) {
+    let dir = tempfile::tempdir().expect("tempdir");
+    let mut config = xerj_common::config::Config::default();
+    config.server.data_dir = dir.path().to_string_lossy().into_owned();
+    config.storage.wal_sync = xerj_common::config::WalSync::Async;
+    let metrics = xerj_common::metrics::Metrics::new().expect("metrics");
+    let engine = xerj_engine::Engine::new(config.clone()).expect("engine");
+    let state = xerj_api::state::AppState::new(config, engine, metrics);
+
+    let mut schema = Schema::empty();
+    schema
+        .add_field(FieldConfig::new("grp", FieldType::Keyword))
+        .expect("grp field");
+    schema
+        .add_field(FieldConfig::new("title", FieldType::Text))
+        .expect("title field");
+    state.engine.create_index("docs", schema).expect("create");
+    let idx = state.engine.get_index("docs").expect("get index");
+    for (id, grp) in [("1", "A"), ("2", "A"), ("3", "B")] {
+        idx.index_document(
+            Some(id.into()),
+            json!({ "grp": grp, "title": format!("doc {id}") }),
+        )
+        .await
+        .expect("index document");
+    }
+    idx.refresh().await.expect("refresh");
+
+    (xerj_api::router::build_es_compat_router(state), dir)
+}
+
+async fn post(app: &axum::Router, path: &str, body: Value) -> (StatusCode, Value) {
+    let response = app
+        .clone()
+        .oneshot(
+            Request::post(path)
+                .header("content-type", "application/json")
+                .body(Body::from(body.to_string()))
+                .expect("request"),
+        )
+        .await
+        .expect("response");
+    let status = response.status();
+    let bytes = axum::body::to_bytes(response.into_body(), usize::MAX)
+        .await
+        .expect("body");
+    (status, serde_json::from_slice(&bytes).unwrap_or(Value::Null))
+}
+
+/// A collapse `inner_hits` must render even under an explicit
+/// `_source.includes` that omits the collapse sentinel (#651).
+#[tokio::test]
+async fn collapse_inner_hits_render_under_explicit_source_includes() {
+    let (app, _dir) = app().await;
+
+    // Collapse on grp with inner_hits AND an explicit `_source.includes` that
+    // lists only `grp` (NOT the internal collapse sentinel).
+    let (st, resp) = post(
+        &app,
+        "/docs/_search",
+        json!({
+            "query": { "match_all": {} },
+            "size": 10,
+            "_source": { "includes": ["grp"] },
+            "collapse": { "field": "grp", "inner_hits": { "name": "members", "size": 10 } }
+        }),
+    )
+    .await;
+    assert_eq!(st, StatusCode::OK, "{resp}");
+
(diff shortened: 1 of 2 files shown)


xerj-org:

Fix tip `545f21ca` (rebased to `6446168a`, code-identical). **Result: 3× CONFIRMED_SOUND, ZERO BLOCKING.** Sentinels do NOT leak on the wire (leader + inner_hits members both strip `__xy_collapse_group__`/`__xy_collapse_spec__`); the explicit projection still narrows the leader `_source` (no field resurrection); **both** the `Includes`/array form and the `Fields`/exclude form preserve inner_hits (each arm's reattach independently load-bearing); a non-collapse request is a no-op; 625 lib tests green; savings byte-identical to main.

**One non-blocking, pre-existing note:** `default_projection_len` counts the sentinel bytes in the savings baseline, so a collapse leader's reported savings is overstated by the size of the stashed group — but this is **identical on `main`** (the fix touches neither the baseline nor the record), so it is out of scope here. Candidate to fold into a savings-accuracy follow-up.

Attempted to REFUTE; every EVIDENCE-lens claim held. HEAD reset to 545f21ca7eba3b052e8235a713b6e92a2264348e (git rev-parse confirmed). Fix present in engine/crates/xerj-engine/src/index.rs: COLLAPSE_SENTINELS (26271), take_collapse_sentinels (26273), reattach_collapse_sentinels (26285); call pairs in Includes arm (take 26423 / reattach 26431) and Fields arm (take 26440 / reattach 26445). Test: `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-api --test collapse_inner_hits_under_source_includes` -> 1 passed.

(1) Reattach load-bearing: replaced ONLY the Fields-arm reattach with `let _ = sentinels;` -> existing test FAILED at tests/collapse_inner_hits_under_source_includes.rs:96 with leader `{"_source":{"grp":"A"},"fields":{"grp":["A"]}}` and NO inner_hits. Restored with `git checkout --`.

(2) BOTH projecting arms fixed: verified the variant mapping — `_source:["grp"]` (array) -> SourceFilter::Includes (parser.rs:3293, es_compat.rs:6109), `_source:{includes:["grp"]}` (object) -> SourceFilter::Fields (es_compat.rs:6116). The shipped test only exercises the Fields arm, so I added a temp test with the array form to hit the Includes arm; both PASSED on the unmodified engine. Then broke ONLY the Includes-arm reattach -> Includes test FAILED (no inner_hits, leader `_source:{grp:A}`) while Fields test still PASSED, proving each arm's reattach is independently required. Restored both files.

(3) API strips sentinels off the wire: es_compat.rs 11943-11953 extracts __xy_collapse_group__/__xy_collapse_spec__ from the leader source before projection, then 11999-12000 removes both from the returned _source; inner_hits render path strips them at 8617-8618. Test assertion `!src.contains_key("__xy_collapse_group__")` passes — sentinels survive the engine projection but not the wire.

fail-before holds: removing either arm's reattach reproduces the #651 symptom (missing inner_hits). Worktree left clean: HEAD==545f21ca, empty `git status --porcelain`, reattach_collapse_sentinels count back to 3. No blocking items.

</details>

REFUTE attempt failed; all blocking criteria pass. head_ok: HEAD==545f21ca, tree clean. Fix present in index.rs:26271-26297; take_collapse_sentinels()/reattach_collapse_sentinels() applied in BOTH SourceFilter::Includes (26423/26431) and SourceFilter::Fields (26440/26445). Canonical test collapse_inner_hits_render_under_explicit_source_includes passes (rustup 1.97.1). Fail-before proven: neutering take_collapse_sentinels() -> Vec::new() makes the test fail at the inner_hits assertion (leader came back `_source:{grp:A}`, no inner_hits), then restored via `git checkout --`.

CORRECTNESS lens, all verified with a throwaway probe (4 tests, ran green, then removed):
(a) NO wire-leak. Both leader (es_compat.rs:11999-12000) and inner_hits members (es_compat.rs:8617-8618) strip BOTH __xy_collapse_group__ AND __xy_collapse_spec__. Probe asserted neither sentinel present in leader _source nor in member _source for array + exclude forms. Existing test also asserts group-sentinel absence.
(b) No user-field resurrection. reattach_collapse_sentinels only re-inserts the two fixed sentinel keys with their captured values; it cannot re-add `title`. Probe confirmed `title` stays dropped under _source:["grp"].
(c) EXCLUDE form and ARRAY form both preserve inner_hits. Request-parsing (es_compat.rs:6100-6143): array/string -> SourceFilter::Includes; object {includes|excludes} -> SourceFilter::Fields. Note the existing test's {includes:[...]} actually exercises the Fields variant; my probe adds the ARRAY form (exercises the Includes path) and the EXCLUDE form (Fields w/ excludes) — both render inner_hits, narrow to grp, no sentinel leak.
(d) Non-collapse request is a no-op: reattach early-returns on empty capture. Probe: non-collapse _source is exactly {grp} (len 1), no sentinels, no inner_hits key.

Non-blocking observation (out of scope for this lens; pre-existing, not introduced by this PR): baseline `default_projection_len(&h.source, ...)` is computed at index.rs:26421-26422 BEFORE take_collapse_sentinels at 26423, and it counts every non-companion key including __xy_collapse_group__/__xy_collapse_spec__ (default_projection_len at 26244 does not exclude the sentinels). So for a collapse leader the reported PayloadSavings baseline includes the byte size of the full stashed group array, which the API strips and never emits — inflating the "savings" statistic for collapse leaders. This same skew existed on main (filter_object stripped the sentinel while the baseline retained it), so the fix does not change measurement behavior. It is a metric-accuracy/vanity concern, not a wire-leak / field-resurrection / inner_hits-drop, so not BLOCKING under the CORRECTNESS lens.

</details>

head_ok: HEAD == 545f21ca confirmed; fix present (8 refs to take_collapse_sentinels/reattach_collapse_sentinels/COLLAPSE_SENTINELS in engine/crates/xerj-engine/src/index.rs).

tests_pass: target test collapse_inner_hits_render_under_explicit_source_includes PASSES under rustup 1.97.1 (1 passed).

fail_before_holds: git checkout main -- index.rs (fix removed, test kept) -> target test FAILS at line 96, leader returned as {"grp":"A"} with no inner_hits. Restored fix from HEAD (had to use `git checkout 545f21ca -- file` because plain `git checkout -- file` restored main's staged copy — the known footgun). Worktree clean, HEAD correct, fix re-present.

REGRESSION checks all clear:
(1) Full `cargo test -p xerj-engine --lib` = 625 passed, 0 failed. The #602/#310/#644/#646 source-filter tests are inline in index.rs and green. Savings-honesty tests (omitted_source_measures_companion_payload_reduction, projecting_a_companion_alongside_a_large_body_still_measures_honestly) are in that green suite. No test literally named savings_stat_is_honest exists.
(2) Savings measured on sentinel-free projection: confirmed. acc.record runs BEFORE reattach in both arms (Includes 26429 then reattach 26431; Fields 26443 then reattach 26445). json_len(&h.source) at record time is the post-filter_object, pre-reattach (sentinel-free) projection.
(3) git diff main..HEAD shows the change is strictly additive: the baseline (default_projection_len) line and the acc.record(...) line are byte-identical to main; only the two helper fns plus take(before filter_object)/reattach(after record) were added. Therefore the savings NUMBER for every hit is byte-identical to main. Non-collapse hits: take returns empty vec, reattach is an early-return no-op, so unaffected.
Test also asserts the sentinel does not leak on the wire (lines 110-112) — passes.

Non-blocking observation (PRE-EXISTING on main, NOT introduced by this fix, so not a regression): default_projection_len counts the two sentinel keys in the baseline (they are neither companion fields nor __xerj_passage_meta__-prefixed, so strip_internal_passage_metadata leaves them), while the subtrahend is sentinel-free. For a collapse leader with _source.includes this overstates reported savings by the size of the entire __xy_collapse_group__ array. Identical arithmetic on main (fix touches neither baseline nor record), so out of scope for this PR.

REFUTE attempt unsuccessful; no blocking regression.

</details>

PR #656, opened 2026-08-23
fix(engine): project _source filters inside nested arrays of objects (#653)

### LINKED ISSUE #653: _source projection does not recurse into nested arrays of arrays (follow-up to #644)
Surfaced by the round-2 verification of #644 (PR #652), and **pre-existing** (untouched by that diff). #644 projects into a flat array of objects (`tags[i].name`), but a **nested array of arrays** is not projected: `filter_object({"a":[[{"name":"x","id":1}]]}, include ["a.name"])` returns `{}` because es_filter's array arm sends a nested-list element to the `other => child_all.then(..)` (scalar) branch, which drops it when the path is non-accepting. ES (`XContentMapValues`) recurses into nested lists and would keep `{"a":[[{"name":"x"}]]}`. Fix direction: in the array arm, treat a `Value::Array` element like a `Value::Object` element — recurse it at the same path (arrays are transparent to a `_source` path at every depth). Niche (nested-array-of-objects docs); separate from #644's flat-array scope.

xerj-org:
Closes #653. Follow-up to #644.

## Problem
An array index is transparent to a `_source` path at **every** depth, but `es_filter`'s array arm sent a nested-`Array` element to the scalar branch (which drops it when the path is non-accepting). So `include ["a.name"]` over `{"a":[[{"name":"x","id":1}]]}` returned `{}` instead of `{"a":[[{"name":"x"}]]}` — a nested array of objects was not projected.

## Fix
Extracted the per-element projection into `project_array_at`, which now recurses:
- an **object** element → `es_filter` (with the emptied-element prune from #644);
- a **nested-array** element → recurse `project_array_at` (with the same empty-array prune);
- a **scalar** element → kept only in an accepting state.

The fully-included fast path (kept whole) is unchanged.

## Testing (rustc 1.97.1)
- New `source_filter_projects_inside_nested_arrays` — **fail-before**: on `main` the nested array was dropped (`{}`); with the fix it projects to `[[{"name":"x"}]]`, and an exclude scrubs at depth through the nested array.
- All 15 `source_filter_tests` green (the new test + every #602/#310/#644 test).
- Full `cargo test -p xerj-api` — dev: 56 binaries, 0 failures.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -26639,6 +26639,38 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
         (acc, liv)
     }
 
+    // Project each element of an array at the same dotted `path` — an array
+    // index is transparent to a `_source` path at EVERY depth (#644/#653), so a
+    // nested array of objects recurses too. An object element is filtered (and
+    // an emptied one dropped, mirroring the map-value rule); a nested-array
+    // element recurses here; a scalar survives only in an accepting state.
+    fn project_array_at(
+        arr: &[Value],
+        path: &str,
+        child_all: bool,
+        includes: &[String],
+        excludes: &[String],
+    ) -> Vec<Value> {
+        arr.iter()
+            .filter_map(|elem| match elem {
+                Value::Object(child_obj) => {
+                    let f = es_filter(elem, path, child_all, includes, excludes);
+                    match &f {
+                        Value::Object(m) if m.is_empty() => {
+                            (child_obj.is_empty() && child_all).then_some(f)
+                        }
+                        _ => Some(f),
+                    }
+                }
+                Value::Array(inner) => {
+                    let p = project_array_at(inner, path, child_all, includes, excludes);
+                    (!p.is_empty() || (inner.is_empty() && child_all)).then(|| Value::Array(p))
+                }
+                other => child_all.then(|| other.clone()),
+            })
+            .collect()
+    }
+
     // Filter one value at dotted `path`. `include_all` means an ancestor path
     // was ACCEPTed (or the include list is empty), so this subtree is included
     // subject only to excludes.
@@ -26708,26 +26740,10 @@ fn filter_object(source: &Value, includes: &[String], excludes: &[String]) -> Va
                     if child_all && !exc_live {
                         out.insert(key.clone(), val.clone());
                     } else {
-                        let projected: Vec<Value> = arr
-                            .iter()
-                            .filter_map(|elem| match elem {
-                                Value::Object(child_obj) => {
-                                    let f = es_filter(elem, &child, child_all, includes, excludes);
-                                    match &f {
-                                        // An element that *filtering* emptied is
-                                        // dropped from the array; one that was
-                                        // ALREADY `{}` survives only in an
-                                        // accepting state — same rule ES applies
-                                        // to array elements as to map values.
-                                        Value::Object(m) if m.is_empty() => {
-                                            (child_obj.is_empty() && child_all).then_some(f)
-                                        }
-                                        _ => Some(f),
-                                    }
-                                }
-                                other => child_all.then(|| other.clone()),
-                            })
-                            .collect();
+                        // Project each element at the same path (recursing into
+                        // nested arrays — #644/#653).
+                        let projected =
+                            project_array_at(arr, &child, child_all, includes, excludes);
                         // A key whose array *filtering* emptied is pruned (had
                         // elements, all dropped); an already-empty source array
                         // survives only in an accepting state — mirrors the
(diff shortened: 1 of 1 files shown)


xerj-org:

Fix tip `99725d8a`. **Result: 3× CONFIRMED_SOUND, ZERO BLOCKING.** Nested-array projection matches ES at every depth (verified to depth-3), emptied nested arrays/keys are pruned (no spurious `[[{}]]`/`[[]]`), mixed object+array siblings both project, exclude scrubs at depth, and the fully-included fast path is byte-identical (never routes through `project_array_at`). Fail-before holds (removing the nested-`Array` recursion drops the array); **#644 flat-array cases unaffected**; 626 lib tests green; recursion is bounded by JSON nesting depth (no cycles/overflow).

**One non-blocking note (not introduced here):** an already-`{}` literal element inside a `child_all` non-fast-path nested array is retained (inherited from the merged #644 object-arm rule, requires unrealistic data — a literal `{}` in an array with an exclude targeting a deeper sibling). Out of scope for #653.

Tried to REFUTE via the REGRESSION lens; could not. Setup: HEAD reset to 99725d8a3c32b40756fd0440946571043f17ae5e (confirmed). Fix present: project_array_at at index.rs:26647, with the nested-Array recursion arm at 26665-26668 and its call site inside the array arm of es_filter (26745). Targeted suite source_filter_tests: all 15 pass.

(1) fail-before HOLDS. Reverted ONLY the project_array_at nested-Array arm (deleted the `Value::Array(inner) => { let p = project_array_at(inner,...); (!p.is_empty() || (inner.is_empty() && child_all)).then(|| Value::Array(p)) }` arm so a nested-list element falls through to `other => child_all.then(|| other.clone())`, the pre-#653 scalar path). source_filter_projects_inside_nested_arrays then FAILED, returning Object{} instead of {"a":[[{"name":"x"}]]} — exactly the reported bug (child_all=false on the live-not-accepted `a` path drops the nested-list element, emptying and pruning the key). Restored via git checkout --; arm back at 26665; tree clean.

(2) #644 FLAT-array cases UNAFFECTED. During the fail-before run only the nested test failed (14 passed / 1 failed); source_filter_projects_inside_arrays_of_objects stayed green. The change is purely additive (a new Array match arm ahead of the scalar `other` catch-all); flat arrays of objects still hit the Value::Object arm, scalars/heterogeneous elements still hit `other` — behavior identical.

(3) Full cargo test -p xerj-engine --lib GREEN: 626 passed, 0 failed, 0 ignored.

(4) No infinite recursion / stack overflow. project_array_at recurses only on Value::Array(inner) passing the strictly-contained inner slice, so recursion depth = the JSON's array-nesting depth (finite; serde_json parse recursion limit is 128); serde_json::Value is an acyclic owned tree so no cycles. Empty-array prune (`inner.is_empty() && child_all`) and emptied-prune preserved. Value provably shrinks each level. This adds no unbounded risk beyond the pre-existing es_filter object-arm recursion. Fully-included fast paths (object 26715-26718, array 26740-26741 keep-whole) unchanged.

No ES divergence in scope: array-index transparency at every _source depth matches Elasticsearch include/exclude semantics. No blocking or non-blocking regressions found. HEAD left clean at target commit.

</details>

n/a

</details>

Attempted to REFUTE the CORRECTNESS lens; could not. All gates green.

SETUP: HEAD == 99725d8a3c32b40756fd0440946571043f17ae5e (confirmed). Fix present: fn project_array_at at index.rs:26647, recurses on nested-Array elements at 26666 and is invoked from es_filter's Array arm at 26746. tests_pass: source_filter_tests = 15 passed / 0 failed (rustup 1.97.1, RUST_MIN_STACK=8388608).

FAIL-BEFORE (executed, not just reasoned): removed the Value::Array recursion arm from project_array_at so nested-Array elements fall through to the scalar-drop arm (the pre-#653 behavior). Both the shipped regression test source_filter_projects_inside_nested_arrays AND my probe case (a) then returned Object {} instead of {"a":[[{"name":"x"}]]} — i.e. the whole nested array is dropped. Reverted cleanly; git status empty, fix intact.

ADVERSARIAL PROBE (#[cfg(test)] added, run, reverted): hand-derived ES ground truth for every lens case and asserted against filter_object — ALL PASS:
(a) include a.name over {"a":[[{name,id}]]} -> {"a":[[{name}]]}.
(b) triple-nested [[[{name,id}]]] -> [[[{name}]]] (depth-3 recursion holds).
(c) fully-emptied nested array drops the key entirely — no residual [[{}]] or [[]]; verified both bare and with a surviving sibling.
(d) mixed [[{name}],[{other}]] keeps only [[{name}]] (non-matching inner array pruned, not left as []); per-element leaf selection within one inner array also correct.
(e) exclude logs.secret through a nested array scrubs the leaf ([[{msg}]]); an exclude emptying every element prunes the key.
(f) fully-included fast path include ["a"] keeps the nested array structurally identical including {} elements, [] elements, and loose scalars (fast path is val.clone(), never routes through project_array_at). No spurious empties, no dropped real data, no over-include across all cases.
Extras also pass: object-element + nested-array-element siblings in one array both project; an already-empty {} in a LIVE-only (non-accepting) nested array is dropped (no resurrection).

NON-BLOCKING OBSERVATION (out of lens scope, NOT introduced by #653): in the child_all=true but non-fast-path state (only reachable when an exclude pattern is a live prefix into the array's subtree, e.g. include ["a"] + exclude ["a.b.c"]), an already-empty {} literal element in the array is retained via the (child_obj.is_empty() && child_all) rule, whereas ES's array-element filter gates on filteredMap.isEmpty()==false with no exactMatch exemption and would drop it. This behavior is inherited from the merged #644 object arm — #653 only added the nested-Array recursion — and requires unrealistic source data (literal {} inside an array with an exclude targeting a deeper sibling path). Not a regression from this PR and outside the enumerated lens cases (a)-(f); flagging for awareness only, not blocking.

</details>

Issue #637, opened 2026-08-23
es-compat: scroll continuation omits _source only for request _source:false, not mapping _source.enabled:false / stored_fields / null (follow-up to #624)


xerj-org:
Found by the #636 (#624) correctness-skeptic pass. Pre-existing; #624 fixed the request-level `_source: false` case only.

#624 added `ScrollContext::source_disabled`, captured at scroll-open from `matches!(body.source, Some(Value::Bool(false)))`, and `scroll_page_response` now omits `_source` on continuation pages when set. But the `_search?scroll=` FIRST page (`search_impl`, es_compat.rs ~11862-11866) suppresses `_source` in MORE cases:

- mapping `_source.enabled: false`,
- `stored_fields` implying source suppression (when `_source` is unspecified),
- a null/absent stored source.

A scroll opened against one of those would still EMIT `_source` on continuation pages while the first page omits it — a first-page/continuation divergence for those cases (the request-level `_source:false` case is now consistent after #624).

**Fix:** capture the full suppression decision at scroll-open (the same boolean `search_impl` computes for its own render) into `ScrollContext::source_disabled`, not just `body.source == false`. Thread the mapping/stored-fields checks the first page already does.

NON_BLOCKING correctness note, not a regression. Refs #636, #624.

xerj-org:
Fixed by #658 (merged 2026-08-23, commit `fa651277`). The scroll continuation (`scroll_page_response`) now matches the first page's `_source` suppression for all cases the issue lists:
- request `_source:false` — via `ScrollContext::source_disabled` (#624),
- `stored_fields`-implied — folded into the request-level `source_disabled` constant,
- mapping `_source.enabled:false` — re-derived PER HIT from each hit's own index (correctly, even on a multi-index scroll — a data-loss regression the #658 skeptics caught and the fix resolved),
- null/absent stored source — handled per-hit in the continuation render.

Scoped per opening route via `ScrollContext::mapping_source_check`; verified by `scroll_source_suppression_parity` (both profiles) + 3× independent skeptic passes. The sibling `/:index/_search_scroll` route's own first-page gap is tracked separately in #659 (PR #661).

PR #636, opened 2026-08-23
fix(api): scroll continuation honours _source:false from the opening request (#624)

xerj-org:
## What

Reproducing #624 established two things:

1. **The filed premise is wrong.** A collapse `_search?scroll=` continuation
   *does* render `inner_hits` under `_source: false` — the scroll snapshot is
   `merged_hits.clone()` taken **before** the per-hit `_source` suppression, so
   the context keeps the sentinel-laden source and `scroll_page_response` renders
   the collapse group fine. #622's cautious "no inner_hits" scoping note was
   over-conservative.
2. **The real bug.** The continuation **emitted `_source`** even though the
   opening request set `_source: false` — a divergence from ES and from the
   `_search?scroll=` first page (`search_impl`, which already omits it). All
   scroll continuations (collapse or not) were affected: the single `_source`
   insert in `scroll_page_response` never consulted the request's source
   selection.

## How

ES fixes the `_source` selection at the scroll-opening request and applies it to
every page — exactly like `seq_no_primary_term`/`version`. So capture it the same
way: a new `ScrollContext::source_disabled`, set at each open site from
`matches!(body.source, Some(Value::Bool(false)))`, and honoured in
`scroll_page_response` by omitting `_source` when set. `inner_hits` still render
from the collapse snapshot regardless (the sentinels are consumed into
`inner_hits` *before* the omission). The non-suppressed path is byte-identical
(`source_disabled` false → `_source` inserted as before). All 6 `ScrollContext`
construction sites updated.

## Fail-before

`collapse_scroll_source_false_still_renders_inner_hits` opens a collapse scroll
with `_source: false`, continues to page 2, and asserts (a) `inner_hits.members`
renders and (b) the leader has **no** `_source`. Before the fix, (b) fails — the
leader carries `_source:{grp:B,v:20}`. With the fix both hold.

Full `cargo test -p xerj-api` + `-p xerj-engine` green under rustc 1.97.1 in
**both** dev and ci-test. fmt `--all --check` ✓, clippy `--all-targets -D warnings` ✓.

## Scope

Covers the `scroll_page_response` continuation (both `_search?scroll=` and
`_search_scroll`). `search_with_scroll`'s own hand-built FIRST page emitting
`_source` under `_source:false` is a sibling of #630's first-page path and is
left as a follow-up. Closes #624.

Issue #624, opened 2026-08-22
collapse + scroll: continuation renders no inner_hits under _source: false (deferred from #622)


xerj-org:
Deferred scope from #622 (fix for #621).

#622 fixed collapse `inner_hits` rendering on scroll continuation pages, but the render is guarded by `if let Some(src) = &h.source` in `scroll_page_response`. When a collapse scroll is opened with `_source: false`, `h.source` is `None`, so the continuation page renders **no** `inner_hits` at all — the collapse group sentinels live in `_source`, which is suppressed.

ES renders collapse `inner_hits` independently of top-level `_source` suppression (the sentinels are internal, not user source). `search_impl` reads the sentinels off `h.source` at :11590 before the response-time `_source` suppression decision, so page 1 is unaffected; only the scroll path is gapped.

This is a pre-existing niche gap (a collapse scroll under `_source: false` is unusual), explicitly scoped out of #622. Fix would thread the raw source (or the extracted group/spec) into `scroll_page_response` independently of the `_source` output decision.

Refs #622, #621.

PR #658, opened 2026-08-23
fix(api): scroll continuation matches first-page _source suppression (#637)

xerj-org:
### Bug
A `_search?scroll=` continuation page (`scroll_page_response`) suppressed `_source` **only** when the opening request set `_source: false` — the sole case #624 captured into `ScrollContext::source_disabled`. But `search_impl`'s first-page render omits `_source` on **four** conditions:
1. `stored_fields` implying suppression (and `_source` unspecified),
2. `_source: false`,
3. mapping `_source.enabled: false`,
4. per-hit null/absent stored source.

A scroll opened against **(1)** or **(3)** omitted `_source` on the first page but **emitted it on every continuation page** — a first-page/continuation divergence from ES. (Case 4 was already handled per-hit in the continuation; case 2 is the one #624 fixed.)

### Fix
- Extract `mapping_source_disabled(state, index)` — the exact mapping check the first-page render used inline — so the first page and scroll-open share one source of truth.
- At the `search_impl` scroll-open site, capture the full **request/index-level** decision:
  `_source:false || (stored_fields-implied && _source unspecified) || mapping _source.enabled:false`.
- Per-hit null-source stays handled in `scroll_page_response`.

### Scoping (important — avoids a reverse divergence)
The separate `/:index/_search_scroll` route (`search_with_scroll`) has its **own** first-page render that does **not** suppress `_source` for cases 1–3 (it emits `_source` unconditionally). Making *its* capture site suppress would create the opposite divergence, so it is left request-level. That route's broader first-page suppression gap is pre-existing and out of scope here — filing as a follow-up.

### Fail-before / after
New test `scroll_source_suppression_parity` (two cases: mapping `_source.enabled:false`, and `stored_fields` without `_source`). Both assert the first page omits `_source` (baseline) **and** the continuation omits it. On pristine main both continuation assertions fail (continuation returned `_source:{...}`); with the fix both pass.

### Verification (rustc 1.97.1)
- `cargo fmt --all --check` ✓ · `clippy --workspace --all-targets -D warnings` ✓
- Full `xerj-api` suite, **dev** ✓ and **ci-test** ✓ (both profiles)
- `build --workspace --profile ci-test` finishing

Refs #624, #636.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8726,6 +8726,29 @@ fn render_collapse_inner_hits(
     Value::Object(combined)
 }
 
+/// Whether the index mapping sets `_source.enabled: false`. ES suppresses
+/// `_source` in the response for such an index; xerj keeps the underlying
+/// source available internally (fields/highlight/_ignored extraction) and makes
+/// the suppression a response-time decision. Mirrors the per-hit check in
+/// `search_impl`'s render, extracted so scroll-open can capture the SAME
+/// decision into `ScrollContext::source_disabled` and continuation pages stay in
+/// parity with the first page (#637).
+fn mapping_source_disabled(state: &AppState, index: &str) -> bool {
+    state
+        .engine
+        .index_mappings
+        .get(index)
+        .and_then(|m| {
+            m.get("mappings")
+                .and_then(|mm| mm.get("_source"))
+                .or_else(|| m.get("_source"))
+                .cloned()
+        })
+        .and_then(|src| src.get("enabled").and_then(Value::as_bool))
+        .map(|b| !b)
+        .unwrap_or(false)
+}
+
 async fn search_impl(
     state: AppState,
     index: String,
@@ -11958,19 +11981,7 @@ async fn search_impl(
             // extraction, so the suppression is a response-time
             // decision not a data-layer decision.
             let source_body_disabled = matches!(body.source, Some(Value::Bool(false)));
-            let source_mapping_disabled = state
-                .engine
-                .index_mappings
-                .get(idx_name.as_str())
-                .and_then(|m| {
-                    m.get("mappings")
-                        .and_then(|mm| mm.get("_source"))
-                        .or_else(|| m.get("_source"))
-                        .cloned()
-                })
-                .and_then(|src| src.get("enabled").and_then(Value::as_bool))
-                .map(|b| !b)
-                .unwrap_or(false);
+            let source_mapping_disabled = mapping_source_disabled(&state, idx_name.as_str());
             // `stored_fields` implicitly suppressing `_source` (unless the
             // list literally contains `"_source"`) is only a DEFAULT for
             // when the caller left `_source` unspecified. When the request
@@ -14620,6 +14631,21 @@ async fn search_impl(
             .min(sc_cfg.scroll_max_keep_alive_secs);
         let keep_alive = std::time::Duration::from_secs(keep_alive_secs);
         let now = Instant::now();
+        // #637: capture the FULL first-page `_source` suppression decision, not
+        // just request-level `_source: false`. search_impl's own render above
+        // also omits `_source` when `stored_fields` implies it (and the request
+        // left `_source` unspecified) or the mapping sets `_source.enabled:
+        // false`. Both are request/index-level (constant across the snapshot),
+        // so a single captured boolean keeps continuation pages in parity with
+        // this first page. The remaining per-hit null-source case is already
+        // handled directly in `scroll_page_response`. (The
+        // `/:index/_search_scroll` route's own first page does not suppress these
+        // cases, so its capture site is intentionally left as request-level only
+        // — matching ITS render.) Computed before the struct literal moves
+        // `scroll_index` into `ctx.index`.
+        let scroll_source_disabled = matches!(body.source, Some(Value::Bool(false)))
+            || (suppress_source_for_stored && body.source.is_none())
+            || mapping_source_disabled(&state, &scroll_index);
         let ctx = xerj_engine::engine::ScrollContext {
             index: scroll_index,
             hits: snapshot,
@@ -14630,7 +14656,7 @@ async fn search_impl(
             // first page's own `_seq_no`/`_version` emission above.
             seq_no_primary_term: emit_seq_no,
             version: emit_version,
-            source_disabled: matches!(body.source, Some(Value::Bool(false))),
+            source_disabled: scroll_source_disabled,
             created: now,
             keep_alive,
             expires_at: now + keep_alive,
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -14631,21 +14631,28 @@ async fn search_impl(
             .min(sc_cfg.scroll_max_keep_alive_secs);
         let keep_alive = std::time::Duration::from_secs(keep_alive_secs);
         let now = Instant::now();
-        // #637: capture the FULL first-page `_source` suppression decision, not
-        // just request-level `_source: false`. search_impl's own render above
-        // also omits `_source` when `stored_fields` implies it (and the request
-        // left `_source` unspecified) or the mapping sets `_source.enabled:
-        // false`. Both are request/index-level (constant across the snapshot),
-        // so a single captured boolean keeps continuation pages in parity with
-        // this first page. The remaining per-hit null-source case is already
-        // handled directly in `scroll_page_response`. (The
-        // `/:index/_search_scroll` route's own first page does not suppress these
-        // cases, so its capture site is intentionally left as request-level only
-        // — matching ITS render.) Computed before the struct literal moves
-        // `scroll_index` into `ctx.index`.
+        // #637: capture the REQUEST-LEVEL part of the first-page `_source`
+        // suppression decision — `_source: false`, and `stored_fields` implying
+        // suppression when the request left `_source` unspecified. These are
+        // genuinely constant across the whole snapshot (index-independent), so a
+        // single captured boolean is correct for them.
+        //
+        // The mapping `_source.enabled: false` case is deliberately NOT folded in
+        // here: a scroll can span MULTIPLE indices with divergent `_source`
+        // settings, and the first page suppresses that case PER HIT using each
+        // hit's own index. Collapsing it to `scroll_index` (the first index)
+        // would strip `_source` from every continuation hit of a source-ENABLED
+        // index whenever the first index disabled source — a new first-page/
+        // continuation split and silent `_source` loss on the export/reindex API
+        // (found by the #658 correctness skeptic). So `scroll_page_response`
+        // re-derives the mapping check per hit from `h.index` instead.
+        //
+        // The per-hit null-source case is likewise handled in
+        // `scroll_page_response`. (The `/:index/_search_scroll` route's own first
+        // page does not suppress any of these, so its capture site stays
+        // request-level only — matching ITS render; see #659.)
         let scroll_source_disabled = matches!(body.source, Some(Value::Bool(false)))
-            || (suppress_source_for_stored && body.source.is_none())
-            || mapping_source_disabled(&state, &scroll_index);
+            || (suppress_source_for_stored && body.source.is_none());
         let ctx = xerj_engine::engine::ScrollContext {
             index: scroll_index,
             hits: snapshot,
@@ -20796,11 +20803,17 @@ async fn scroll_page_response(
                                 _ => None,
                             }
                         });
-                        // #624: honor the scroll's opening `_source: false` —
-                        // omit `_source` here (the sentinels have already been
-                        // consumed into `inner_hits` above), matching the first
-                        // page. inner_hits are emitted regardless.
-                        if !source_disabled {
+                        // #624/#637: honor the first page's `_source` suppression
+                        // (the sentinels have already been consumed into
+                        // `inner_hits` above), matching search_impl exactly.
+                        // `source_disabled` carries the request-level constant
+                        // (`_source:false`, stored_fields-implied); the mapping
+                        // `_source.enabled:false` case is re-derived PER HIT from
+                        // this hit's own index, because a scroll can span indices
+                        // with divergent `_source` settings and the first page
+                        // suppresses that case per-hit (#658 skeptic). inner_hits
+                        // are emitted regardless.
+                        if !source_disabled && !mapping_source_disabled(state, &h.index) {
                             o.insert("_source".to_string(), src);
                         }
                         if let Some(inner) = collapse_inner {
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -8729,10 +8729,11 @@ fn render_collapse_inner_hits(
 /// Whether the index mapping sets `_source.enabled: false`. ES suppresses
 /// `_source` in the response for such an index; xerj keeps the underlying
 /// source available internally (fields/highlight/_ignored extraction) and makes
-/// the suppression a response-time decision. Mirrors the per-hit check in
-/// `search_impl`'s render, extracted so scroll-open can capture the SAME
-/// decision into `ScrollContext::source_disabled` and continuation pages stay in
-/// parity with the first page (#637).
+/// the suppression a response-time decision. Extracted from `search_impl`'s
+/// per-hit render so the scroll continuation (`scroll_page_response`) can apply
+/// the SAME per-hit decision from each hit's own index — mapping suppression is
+/// NOT a snapshot constant, because a scroll can span indices with divergent
+/// `_source` settings (#637).
 fn mapping_source_disabled(state: &AppState, index: &str) -> bool {
     state
         .engine
@@ -14648,9 +14649,12 @@ async fn search_impl(
         // re-derives the mapping check per hit from `h.index` instead.
         //
         // The per-hit null-source case is likewise handled in
-        // `scroll_page_response`. (The `/:index/_search_scroll` route's own first
-        // page does not suppress any of these, so its capture site stays
-        // request-level only — matching ITS render; see #659.)
+        // `scroll_page_response`. `scroll_page_response` is shared with the
+        // `/:index/_search_scroll` route, whose first page does NOT apply the
+        // mapping check — so the per-hit mapping gate there is guarded by the
+        // `mapping_source_check` flag set below (true here, false for that
+        // route), keeping each route's continuation identical to its own first
+        // page. That route's broader `_source` gap is tracked in #659.
         let scroll_source_disabled = matches!(body.source, Some(Value::Bool(false)))
             || (suppress_source_for_stored && body.source.is_none());
         let ctx = xerj_engine::engine::ScrollContext {
@@ -14664,6 +14668,9 @@ async fn search_impl(
             seq_no_primary_term: emit_seq_no,
             version: emit_version,
             source_disabled: scroll_source_disabled,
+            // `_search?scroll=` first page (this fn) suppresses mapping
+            // `_source.enabled:false` per hit, so the continuation must too.
+            mapping_source_check: true,
             created: now,
             keep_alive,
             expires_at: now + keep_alive,
@@ -20385,6 +20392,12 @@ pub async fn search_with_scroll(
             seq_no_primary_term: emit_seq_no,
             version: emit_version,
             source_disabled: matches!(body.source, Some(Value::Bool(false))),
+            // `_search_scroll` (this route) does NOT apply mapping
+            // `_source.enabled:false` suppression on its first page (it emits
+            // `_source` unconditionally), so its continuation must not either —
+            // otherwise it gains a new first-page/continuation split. That
+            // route's broader `_source` gap is tracked in #659.
+            mapping_source_check: false,
             created: now,
             keep_alive,
             expires_at: now + keep_alive,
@@ -20670,12 +20683,18 @@ async fn scroll_page_response(
             // borrows `ctx.hits`.
             let emit_seq_no = ctx.seq_no_primary_term;
             let emit_version = ctx.version;
-            // #624: the opening request's `_source: false` governs the whole
-            // scroll — omit `_source` on continuation pages too (ES fixes source
-            // selection at open; the `_search?scroll=` first page already omits
-            // it via search_impl). inner_hits still render from the collapse
-            // group snapshot regardless.
+            // #624/#637: the opening request's REQUEST-level `_source` decision
+            // (`_source:false`, stored_fields-implied) governs the whole scroll —
+            // omit `_source` on continuation pages too (ES fixes source selection
+            // at open). inner_hits still render from the collapse group snapshot
+            // regardless.
             let source_disabled = ctx.source_disabled;
+            // #637: whether to ALSO apply the mapping `_source.enabled:false`
+            // suppression per hit below. Only true for scrolls opened by
+            // `search_impl` (whose first page applies it); false for the
+            // `_search_scroll` route, whose first page does not — so its
+            // continuation stays byte-identical to that first page (#659).
+            let mapping_source_check = ctx.mapping_source_check;
 
             // Detect whether the initial search sorted by a non-score key;
             // in that case `max_score` must be null on scroll pages too.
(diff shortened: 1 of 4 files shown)


xerj-org:
The out-of-scope `/:index/_search_scroll` (`search_with_scroll`) first-page `_source` suppression gap referenced above is now filed as #659.

xerj-org:
## Independent 3-skeptic verification (each REFUTES, own worktree, rustc 1.97.1) — fix tip `b6bfa0e2`

**Result: 1 REFUTED (BLOCKING) + 2 CONFIRMED_SOUND.** The correctness lens found a **real NEW regression** on multi-index mixed-mapping scroll — silent `_source` data-loss for a source-ENABLED index on continuation pages. My merge gate is ZERO BLOCKING (not majority), and this is a genuine regression with a clean fix, so **I am fixing this PR, not merging it.** Fix: capture only the truly request-level constant into `source_disabled`, and apply the mapping `_source.enabled:false` check **per-hit** in `scroll_page_response` using each hit's own `h.index` (mirroring the first page's per-`idx_name` check) — plus a multi-index regression test.

### CORRECTNESS vs Elasticsearch — **REFUTED**
- head_ok=True · fix_symbol_present=True · tests_pass=True · fail_before_holds=True · es_divergence_found=True

**BLOCKING:**
> MULTI-INDEX SCROLL: new ES-divergence + silent _source data-loss REGRESSION. es_compat.rs:14646-14648 captures scroll_source_disabled as a single snapshot-constant using mapping_source_disabled(&state, &scroll_index) where scroll_index = index_names.first() (:14616). But the first page suppresses per-hit using EACH hit's own index (source_mapping_disabled = mapping_source_disabled(idx_name) at :11983-11999). For a scroll over a MIX of indices where the first-resolved index has _source.enabled:false and another index has it ENABLED, the continuation (scroll_page_response, source_disabled consumed at :20671/:20803) strips _source from EVERY continuation-page hit of the ENABLED index. Verified with a probe: scroll over `a_off,b_on` (a_off source-disabled, b_on enabled), size 2, sort v asc -> page 1 correctly keeps b_on d0's _source (per-hit), but pages 2-4 drop _source from b_on d1/d2/d3. On origin/main those b_on hits keep _source on ALL pages, so this is a NEW regression introduced by the PR, not a pre-existing bug. Impact: scroll is the export/reindex API; this is silent _source loss (200 OK, hit present, _source absent) for a source-enabled index on every page after the first, and it RE-CREATES the exact first-page/continuation split #637 sets out to fix. Root cause: the code comment at :14638-14640 asserts mapping suppression is 'request/index-level (constant across the snapshot)' -- that is false for a multi-index scroll; only the request-level parts (source:false, stored_fields&&source.is_none) are truly constant. Fix: on continuation, OR the captured request-level constant with a per-hit mapping check mapping_source_disabled(&state, &h.index) (h.index is already in scope at scroll_page_response :20692/:20745), mirroring the first page's per-idx_name check, instead of the single first-index constant.

<details><summary>notes</summary>

Single-index cases are fully correct and the fix is a genuine improvement there; the captured boolean is bit-identical to the first-page per-hit decision when scroll_index == idx_name (source_body_disabled == matches!(body.source,Some(Bool(false))); the stored_fields term is identical; mapping term identical). LENS-2 verified by probe: a request with BOTH stored_fields:[grp] AND explicit _source:{includes:[...]} correctly KEEPS _source on the continuation -- `suppress_source_for_stored && body.source.is_none()` is the right stored_fields condition; explicit _source wins, matching the first page. LENS-4 control verified: a plain scroll (enabled mapping, no stored_fields, no _source:false) does NOT over-suppress _source on continuation. The per-hit null-source case is handled separately in scroll_page_response (`if h.source.is_null() { None }` at :20703), consistent with the comment. The `/:index/_search_scroll` route is intentionally left request-level only, matching its own first-page render (documented at :14641-14644) -- not evaluated deeply but internally consistent. Fail-before check: with es_compat.rs reverted to origin/main both continuation assertions (lines 114, 163) fail while the first-page assertions (95, 148) still pass -- clean fail-before. Verdict REFUTED only on the multi-index mixed-mapping edge; if reviewers deem that edge out-of-scope/acceptable-documented, the fix is otherwise sound for single-index and the common cases.

</details>

### REGRESSION RISK — **CONFIRMED_SOUND**
- head_ok=True · fix_symbol_present=True · tests_pass=True · fail_before_holds=True · es_divergence_found=True

<details><summary>notes</summary>

SETUP/VERIFY all green. HEAD=b6bfa0e2. Symbols present: mapping_source_disabled @ es_compat.rs:8736, scroll_source_disabled @ 14646. New tests: both pass. Fail-before: reverting ONLY es_compat.rs to origin/main makes BOTH continuation assertions FAIL (parity test lines 114 + 163, panics show continuation hit still carrying _source:{grp,v}); execution reached past the first-page premise asserts (95-98/148-151), so first-page assertions still pass. Restored fix after.

REGRESSION SURFACE (all green): `cargo test -p xerj-api scroll` and `collapse` — no failures; full `cargo test -p xerj-api` — 58 "test result: ok" lines, 0 failed. inner_hits under collapse-scroll unperturbed: the diff only changes the boolean captured into ctx.source_disabled at scroll-open; scroll_page_response still inserts inner_hits unconditionally (20807-20809) and only gates `_source` on `if !source_disabled` (20803) exactly as before; collapse_scroll* and collapse_scroll_source_false_repro pass.

BYTE-EQUIVALENCE (first-page path): the extracted mapping_source_disabled(state, index) is identical to origin/main's inline source_mapping_disabled block (git show origin/main:...:11961-11973) — same mapping-shape fallbacks (mappings._source -> or_else(_source)), same .enabled as_bool, !b, unwrap_or(false); only rename idx_name.as_str() -> index param. First-page call site passes idx_name.as_str(), so single-index behaviour is unchanged. Scroll capture (14646-14648) ORs the three request/index-level first-page conditions: matches!(body.source,Bool(false)) == source_body_disabled(11983); (suppress_source_for_stored && body.source.is_none()) == first-page stored clause(11997); mapping_source_disabled(&state,&scroll_index) == source_mapping_disabled(11984). The 4th first-page clause h.source.is_null() is per-hit and handled in scroll_page_response — not part of this capture. suppress_source_for_stored is a Copy bool defined at 11856 (same fn), body.source read by-ref, scroll_index borrowed before the struct literal moves it — no move/borrow issue (compiles + tests pass).

SHOULD-FIX (NOT blocking; out of #637 single-index scope): multi-index mixed-mapping scroll divergence. First page computes source_mapping_disabled PER-HIT via `.map(|(idx_name,h)|` at 11889/11984 using each hit's OWN index, but scroll capture uses mapping_source_disabled(&state, &scroll_index) where scroll_index = index_names.first() (14616), and scroll_page_response applies that single bool GLOBALLY to every continuation hit (20803). So for a scroll spanning indices with divergent `_source.enabled`: if the first resolved index has _source.enabled:false, continuation pages now SUPPRESS _source for hits from indices that ENABLE source (those hits emitted _source on the first page) — a NEW first-page/continuation divergence for those hits (they were correct on continuation before this PR). Conversely if the first index enables source, disabled-index hits still emit _source on continuation (the #637 bug persists for them). Why non-blocking: (1) narrow edge — requires a multi-index scroll AND per-index-divergent _source.enabled; (2) rooted in the pre-existing single-boolean ScrollContext::source_disabled model which fundamentally cannot express per-index/per-hit suppression (the codebase already treats scroll_index as "context-level fallback only", 14611-14615, and uses a lossy `.any()` approximation for the analogous top_hits case at 10370); (3) the PR strictly IMPROVES the common uniform-mapping multi-index case and the single-index case #637 targets; (4) untested in either direction. Recommended follow-up: either capture per-index suppression (Vec/set) or use `.any()`/document the multi-index limitation, mirroring the source_disabled_somewhere pattern. No test regression and the core single-index fix is sound, so this does not block merge.

</details>

### DOES THE EVIDENCE HOLD — **CONFIRMED_SOUND**
- head_ok=True · fix_symbol_present=True · tests_pass=True · fail_before_holds=True · es_divergence_found=True

<details><summary>notes</summary>

VERDICT: CONFIRMED_SOUND. The claimed fix is real and end-to-end reproducible; I could not refute it.

EVIDENCE REPRODUCED (all under RUST_MIN_STACK=8388608 rustup run 1.97.1):
- HEAD = b6bfa0e2 (#637 fix). Both fix symbols present: fn mapping_source_disabled @ es_compat.rs:8736, scroll_source_disabled @ 14646/14659.
- tests_pass: cargo test -p xerj-api --test scroll_source_suppression_parity -> 2 passed.
- fail_before_holds (I re-derived it): after `git checkout origin/main -- es_compat.rs`, BOTH tests fail, and crucially the panics land on the CONTINUATION assertions (test lines 114 and 163), not the first-page premise. The first-page premise assertions (mapping _source.enabled:false and stored_fields both omit _source on page 1) execute earlier and PASSED both before and after -> the divergence premise is sound. Pre-fix continuation body was {"_id":"d1",...,"_source":{"grp":"g","v":11}}, so `h.source` is populated even under mapping-disable, confirming suppression is response-time not data-layer, and confirming source-restore afterward. Worktree restored clean (git status empty, symbol present).
- clippy -p xerj-api --all-targets -D warnings: clean.

SCOPING CLAIM VERIFIED (not a missed fix): /:index/_search_scroll (search_with_scroll) builds its first page by hand at es_compat.rs:20319-20361 and emits _source solely on `if h.source.is_null()` (20332) — it applies NO mapping-disable and NO stored_fields suppression on its own first page (scroll_body even sets stored_fields:None @ 20233). So its continuation (shared handler @ 20803, gated on ctx.source_disabled) is already in parity with ITS render. Leaving its capture @ 20380 at request-level-only is therefore correct. The captured scroll_source_disabled @ 14646-14648 exactly mirrors search_impl's three CONSTANT first-page conditions @ 11997-12000 (source_body_disabled | suppress_source_for_stored&&source.is_none | mapping_source_disabled); the 4th condition h.source.is_null() is per-hit and is handled directly in the continuation render @ 20703.

SHOULD-FIX (NOT blocking) — narrow NEW divergence in multi-index scrolls with mixed _source mappings:
search_impl's first page evaluates source_mapping_disabled PER-HIT-INDEX (`merged_hits.map(|(idx_name,h)|...)` @ 11946, mapping_source_disabled(idx_name) @ 11984), but the scroll capture uses only `mapping_source_disabled(&state, &scroll_index)` where scroll_index = index_names.first() (@ 14616). Consequence: a scroll spanning indices A,B where the FIRST index A has _source.enabled:false and B does not -> scroll_source_disabled=true -> the continuation now OMITS _source for B's hits too, even though B's hits EMITTED _source on the first page. That is a new over-suppression regression for the source-enabled index's continuation hits (before the fix those hits emitted _source on continuation, matching their first page). The mirror case (first index enabled, later index disabled) leaves the pre-existing under-suppression unchanged. Net: the single-boolean ScrollContext.source_disabled model cannot represent a per-index decision, so this multi-index mixed-mapping case stays divergent, just shifted. I judged this non-blocking because: (1) it is a structural limitation of the single-boolean model that predates this PR (source_disabled was already one bool in #624), not something the targeted single-index fix introduces in the common path; (2) it requires an unusual config — _source.enabled:false is itself discouraged, mixed within one multi-index scroll rarer still; (3) the PR strictly improves the far-more-common single-index and request-level cases the tests pin. Recommend a follow-up issue (per-index source_disabled, or capture as a per-index set/vec) rather than blocking merge. Tests only cover single-index, so this gap is untested.

</details>

xerj-org:
## Re-verification of the reworked fix (tip `201475cb`) — 3 skeptics, own worktrees, rustc 1.97.1

**Result: 2 CONFIRMED_SOUND + 1 REFUTED (BLOCKING).** The multi-index regression from the prior round is **confirmed fixed** (all three lenses: `multi_index_ok=true`, fail-before holds). But the **regression-risk lens found a NEW BLOCKING issue**: `scroll_page_response` is shared by both scroll routes, so the per-hit mapping gate this PR added now fires for scrolls opened via `/:index/_search_scroll` (`search_with_scroll`) too — and *that* route's first page emits `_source` unconditionally. So the PR introduces a new first-page/continuation `_source` split on the sibling route (the exact class of bug #637 fixes), and the code comment claiming that route is unchanged is now false.

**I am fixing this, not merging.** Scoped fix: carry a per-scroll `mapping_source_check` flag in `ScrollContext` (true for `search_impl`, false for `search_with_scroll`), so the continuation applies the mapping gate only for the route whose first page also applies it. This leaves the `/:index/_search_scroll` route byte-identical to `main` (its broader first-page gap stays #659) and makes the comment accurate. A route-B parity regression test will pin it.

### CORRECTNESS — **CONFIRMED_SOUND**
- head_ok=True · fix_symbol_present=True · tests_pass=True · fail_before_holds=True · multi_index_ok=True

<details><summary>notes</summary>

HEAD=201475cb. Fix symbols all present: fn mapping_source_disabled (es_compat.rs:8736), per-hit continuation gate `if !source_disabled && !mapping_source_disabled(state, &h.index)` (:20816), request-level scroll_source_disabled constant (:14654, captured into ScrollContext::source_disabled at :14666). Bundled tests: all 3 pass incl. scroll_multi_index_source_suppression_is_per_hit. Fail-before (revert ONLY es_compat.rs to origin/main, keep tests): all 3 FAIL — main emits _source on a_off's continuation hit (the exact prior regression) — proving the tests bind the fix; source restored, worktree clean. multi_index_ok: a_off (disabled) omits _source on every page, b_on (enabled) keeps _source on every continuation page.

Independent CORRECTNESS probes I authored (throwaway test, run then removed) — all 4 green:
1) REVERSE ORDER (b_on enabled resolved FIRST, a_off disabled second; interleaved values so continuation pages carry BOTH indices) — per-hit suppression correct: disabled omits, enabled keeps, on every page. Rules out the opposite error of keying on the first index.
2) OVER-SUPPRESSION: plain enabled single-index scroll (no _source:false, no stored_fields) KEEPS _source on all 4 pages incl. continuations — per-hit check does not falsely suppress.
3) stored_fields:["grp"] + explicit _source:true — explicit wins, _source kept on every continuation page (parity with search_impl explicit-wins rule; ScrollContext.source_disabled=false because body.source is Some(Bool(true))).
4) explicit _source:false on an ENABLED index — suppressed on every page via request-level constant path.

Code read confirms parity: continuation gate mirrors search_impl first-page render (11983-12001) exactly; h.index is the per-hit index; mapping_source_disabled lookup defaults unwrap_or(false) on missing mapping, so a missing index mapping KEEPS _source (safe, no over-suppression). No BLOCKING findings. Nit (non-blocking): the bundled multi-index test only covers a_off,b_on order; the reverse (enabled-first) order is not in the committed suite — I verified it out-of-band and it passes, but adding it would harden the regression fence.

</details>

### REGRESSION RISK — **REFUTED**
- head_ok=True · fix_symbol_present=True · tests_pass=True · fail_before_holds=True · multi_index_ok=True

**BLOCKING:**
> NEW first-page/continuation _source parity regression on the sibling scroll route /:index/_search_scroll (router.rs:395, handler search_with_scroll). Its first-page renderer (es_compat.rs:20423-20429) UNCONDITIONALLY emits _source with no mapping suppression, but its continuation is served by the SHARED scroll_page_response, which this PR made suppress _source per-hit via `!mapping_source_disabled(state, &h.index)` (es_compat.rs:20816). PROVEN by an isolated probe against a `_source.enabled:false` index opened via POST /docs/_search_scroll?scroll=5m: after the PR the first page carries `_source:{...}` while the /_search/scroll continuation OMITS it (parity XOR = true); against origin/main's es_compat.rs the same probe shows BOTH pages carry _source (parity holds). So this PR converts a self-consistent route into a self-inconsistent one — the exact first-page/continuation _source split (silent divergence on the export/reindex paging API) that #637 exists to eliminate, mirror-introduced on a neighboring route. This is not merely the known #659 gap: the code comment at es_compat.rs:14653 asserts this route's capture 'stays request-level only — matching ITS render; see #659', which is factually FALSE for the mapping case, because the shared continuation renderer applies the mapping gate unconditionally regardless of which route opened the scroll. Fix: make search_with_scroll's first page apply the same per-hit mapping_source_disabled (+ stored_fields/_source:false) suppression the other two render sites use (i.e. resolve #659 here on the correct side), OR scope the continuation mapping gate so it does not fire for scrolls opened by this route; and correct the false comment either way.

<details><summary>notes</summary>

See notes field above.

</details>

### DOES THE EVIDENCE HOLD — **CONFIRMED_SOUND**
- head_ok=True · fix_symbol_present=True · tests_pass=True · fail_before_holds=True · multi_index_ok=True

<details><summary>notes</summary>

Could not refute. Reproduced end-to-end on tip 201475cb (HEAD confirmed after reset --hard).

EVIDENCE HOLDS:
- Symbols: fn mapping_source_disabled @8736; per-hit continuation gate `if !source_disabled && !mapping_source_disabled(state, &h.index)` @20816; scroll_source_disabled request-level constant @14654/14666.
- Step A (branch): 3/3 pass incl. scroll_multi_index_source_suppression_is_per_hit (1.14s).
- Step B (fail-before, es_compat.rs reverted to origin/main, tests kept): 3/3 FAIL, and each fails at its CONTINUATION assertion (mapping @114, stored_fields @163, multi-index @240 — a_off d2 continuation hit leaked _source). The first-page PREMISE assertions (@95-98, @148-151) did NOT trigger => first page already suppresses both before AND after. Restored cleanly (git status empty).
- multi_index_ok: green. a_off (source-disabled) omits _source on every page; b_on (source-ENABLED) KEEPS _source on every continuation page. Prior BLOCKING single-snapshot-constant regression is GONE.
- Diff (origin/main...HEAD) matches commit claims exactly: mapping_source_disabled extracted (behavior-preserving on first page, still per-hit via idx_name from merged_hits.drain); scroll_source_disabled = matches!(body.source,Bool(false)) || (suppress_source_for_stored && body.source.is_none()) — request-level only, mapping NOT folded in; continuation gate re-derives mapping per hit from h.index. Bonus: this rework ALSO captures the stored_fields-implied case in the constant, which main did not (main captured only _source:false). Null-source handled separately (@20710 sets source=None).
- ci-test profile build (rustup 1.97.1, --profile ci-test) compiles clean (1m08s).

NON-BLOCKING (follow-up / nits — do NOT block #637):
1) Sibling route regression, scoped to #659: `/:index/_search_scroll` (handler search_with_scroll @20187) emits _source UNCONDITIONALLY on its first page (@20423-20429 — no mapping check, no _source:false check), while its continuation shares scroll_page_response which now applies the per-hit mapping gate. So a scroll opened via this route against a mapping `_source.enabled:false` index now emits _source on page 1 but SUPPRESSES it on continuation pages — a NEW first-page/continuation divergence of the same class this PR fixes. On main the mapping case was consistent (both pages emit); this rework widens the route's pre-existing _source:false divergence to also cover mapping-disabled. Mitigating: (a) this route is already non-compliant for _source:false on main and is explicitly deferred to #659 by the author's comment @14650-14653; (b) the continuation direction is toward ES-correctness (ES suppresses _source for mapping-disabled — the route's first page is the non-compliant party, not the continuation); (c) the real export/reindex/scroll path uses /_search?scroll= (search_impl, fully fixed), not this underscore variant; (d) NO existing test asserts the old behavior (verified: scroll_alias_multi_index_cap.rs and search_scroll_firstpage_versioning.rs are the only _search_scroll tests and neither uses _source.enabled:false), so CI stays green. Recommend folding into #659.
2) Stale docstring: mapping_source_disabled doc @8729-8735 still says "extracted so scroll-open can capture the SAME decision into ScrollContext::source_disabled and continuation pages stay in parity" — describes the ABANDONED first attempt (a78cac23), contradicting the rework which deliberately does NOT capture the mapping decision into source_disabled and instead re-derives per-hit. Cosmetic; update to avoid misleading future readers.

</details>

xerj-org:
## Re-verification of the route-scoped flag fix (tip `f50e9391`) — 3 skeptics, own worktrees, rustc 1.97.1

**Result: 3× CONFIRMED_SOUND, ZERO BLOCKING.** Both prior BLOCKING findings are independently validated fixed: `multi_index_ok=true` AND `route_b_parity_ok=true` on all three lenses. The correctness lens ran its own probes confirming the `/:index/_search_scroll` route is **byte-identical to `main`** (on branch *and* main), route A is per-hit correct (including reversed index order), and there is no over-suppression; the evidence lens audited the diff (gate is exactly `!source_disabled && !(mapping_source_check && mapping_source_disabled(state, &h.index))`, all comments accurate) and reproduced fail-before. Non-blocking nits: route-B `_source.enabled:false` still deferred to #659; a pre-existing null-source first-page/continuation asymmetry on route B (present on `main`, unrelated to #637); the literal 'revert only es_compat.rs' fail-before needs engine.rs reverted too (the new required struct field), which the skeptics did — behavioral fail-before holds.

### CORRECTNESS — **CONFIRMED_SOUND**
- head_ok=True · tests_pass=True · fail_before_holds=True · multi_index_ok=True · route_b_parity_ok=True · blocking=[]

<details><summary>notes</summary>

VERDICT CONFIRMED_SOUND — both prior BLOCKING findings are genuinely fixed; my independent correctness probing found no defect.

HEAD=f50e9391, clean worktree. Gated continuation present at es_compat.rs:20839 `!(mapping_source_check && mapping_source_disabled(state, &h.index))`. Field at engine.rs:190. Exactly 2 PRODUCTION ScrollContext open sites: search_impl/_search?scroll= (mapping_source_check:true, 14673) and search_with_scroll/_search_scroll (false, 20400). Other 4 `false` sites are test-only (3 inline passage_scroll_tests, 1 engine TTL test) — correct/irrelevant. Only scrolls.insert prod sites are 14678 (route A) and 20405 (route B); no PIT/slice path opens a context, so the flag is always the opening route's.

Design confirmed correct: scroll_page_response is SHARED by both routes' continuations and parameterized by ctx.mapping_source_check. Route A first page (search_impl:11985-12001) suppresses _source per-hit via mapping_source_disabled(h.index); Route A continuation with flag=true applies the same per-hit gate -> parity. Route B first page (20436-20442) emits _source UNCONDITIONALLY; Route B continuation with flag=false makes the gate term drop out so it reduces to `if !source_disabled` == main's exact expression, and source_disabled for route B stays `matches!(body.source,Bool(false))` unchanged from main -> byte-identical to main.

Tests (A): 4/4 green. Fail-before (B): LITERAL instruction (revert only es_compat.rs) does NOT compile — engine.rs keeps the now-required field so main's 2 construction sites error E0063; proper behavioral fail-before requires reverting engine.rs too. Reverting BOTH to main: 3 original tests (mapping/stored_fields/multi_index) FAIL behaviorally, route-B parity test PASSES on main (expected, noted).

MY LENS probes (written, run, deleted):
- Route B vs main: probe_routeb_source_disabled_every_page_emits (multi-page), probe_routeb_multi_index_self_consistent, probe_routeb_enabled_keeps_source_all_pages — ALL pass IDENTICALLY on branch and main => Route B behavior byte-identical to main; first page and every continuation agree on _source presence.
- Route A REVERSED multi-index b_on,a_off (first-resolved index source-ENABLED): a_off omits / b_on keeps _source on ALL pages — passes on branch, FAILS on main => per-hit re-derivation (history finding #1) is load-bearing, not a first-index constant.
- Over-suppression: plain enabled single-index keeps _source on all continuation pages on both routes.

NITS (non-blocking): (1) literal step-B command can't compile in isolation due to the added required struct field — verify by reverting engine.rs alongside. (2) Route B still emits _source even for _source.enabled:false indices on both first page and continuation (self-consistent, wrong vs ES) — explicitly deferred to #659, out of scope. (3) Pre-existing (main too, both routes): route B first page emits `_source:null` for null-source hits while shared continuation omits the key — a null-source first-page/continuation asymmetry unrelated to #637.

</details>

### REGRESSION_RISK — **CONFIRMED_SOUND**
- head_ok=True · tests_pass=True · fail_before_holds=True · multi_index_ok=True · route_b_parity_ok=True · blocking=[]

<details><summary>notes</summary>

See below.

</details>

### DOES THE EVIDENCE HOLD — reproduce end-to-end and audit claim-vs-code — **CONFIRMED_SOUND**
- head_ok=True · tests_pass=True · fail_before_holds=True · multi_index_ok=True · route_b_parity_ok=True · blocking=[]

<details><summary>notes</summary>

VERDICT CONFIRMED_SOUND — both prior BLOCKING findings are genuinely fixed; no new split or claim/code mismatch found. Evidence:

HEAD f50e9391. Symbols: engine.rs:190 mapping_source_check; es_compat.rs:20839 gated continuation; 14673 true (search_impl); 20400 false (search_with_scroll) plus 3 in-module test literals.

Diff audit (git diff origin/main...HEAD): (1) ScrollContext::mapping_source_check added with accurate doc comment; (2) continuation gate is EXACTLY `!source_disabled && !(mapping_source_check && mapping_source_disabled(state, &h.index))`; (3) search_impl=true, search_with_scroll=false; (4) all comments accurate, no leftover contradictory claim. `mapping_source_disabled()` helper is a byte-equivalent extraction of the formerly-inlined check. First-page suppression in search_impl = stored-implied ∨ body-false ∨ mapping(per-hit) ∨ null; the continuation now reproduces exactly this: source_disabled captures the request-level constants (body-false ∨ stored-implied via the new scroll_source_disabled), mapping is re-derived per-hit from h.index, and null is handled by the outer `if let Some(src)=&h.source` (page_hits set source=None on null). Claim that null-case is 'likewise handled in scroll_page_response' verified.

Route B (search_with_scroll) first page emits _source unconditionally (verified at ~20340: source=if h.source.is_null(){None}else{Some(..)}, no mapping gate); its ScrollContext differs from main ONLY by the added `mapping_source_check:false` field, which no-ops the continuation mapping gate — behaviorally byte-identical to main. Both routes share ONE continuation path (scroll_page_response via /_search/scroll); the flag rides the stored ctx so it follows the opening route. Route B's pre-existing first-page/continuation quirks (body _source:false and stored_fields not suppressed on its first page) are unchanged and correctly deferred to #659.

Tests: all 4 GREEN on branch (multi_index per-hit green; route-B parity green). ci-test build (rustup 1.97.1 cargo build -p xerj-api --profile ci-test) succeeds.

Fail-before caveat (NIT, not blocking): the literal step-B instruction 'revert ONLY es_compat.rs to origin/main' does NOT compile — branch engine.rs adds `mapping_source_check` as a required non-Default field, so main's es_compat.rs ScrollContext literals hit E0063 (two sites). To get a meaningful assertion-level fail-before I reverted the full implementation (es_compat.rs + engine.rs + search_context_ttl.rs) to main while keeping ONLY the new parity test. Result was exactly as predicted: scroll_continuation_omits_source_under_mapping_source_disabled, ..._under_stored_fields, and scroll_multi_index_source_suppression_is_per_hit all FAILED with real assertion panics (continuation leaked _source, e.g. a_off source-disabled hit carried _source on a later page), and search_scroll_route_source_parity PASSED on main (self-consistent there, no per-hit gate). Branch fully restored afterward; working tree clean.

</details>

Issue #408, opened 2026-08-15
a keyword array is indexed as ONE joined term: prefix on a non-first element changes its answer at flush


xerj-org:
Found while fixing #398 — the same shape (one request, two answers either side of a `_flush`) on a different mechanism. #398's fix does **not** close this one, deliberately: closing it in the scan would break the case ES gets right.

Written by an AI agent (Claude Opus 5) driven by @xerj-org.

## The gap

A `keyword` array is indexed as ONE space-joined term. The write path flattens any array to a single string before the FTS layer sees it (`engine/crates/xerj-engine/src/index.rs::extract_field_text`), and every non-`text` field is FTS-indexed with the `keyword` analyzer, whose whole input is one token (`build_fts_field_configs`). So `["Alpha","Beta"]` becomes the single term `Alpha Beta`, and only a prefix of the JOINED string can match it. The stored-document scan, which answers everything still in the memtable, tests each element separately — so the answer flips at flush.

## Repro

```sh
xerj --insecure --data-dir ./d --port 19277     # 1.0.0-rc.17 + the #398 fix

curl -XPUT  localhost:19277/kwarr -H 'Content-Type: application/json' \
  -d '{"mappings":{"properties":{"tags":{"type":"keyword"}}}}'
curl -XPOST 'localhost:19277/kwarr/_doc/1?refresh=true' -H 'Content-Type: application/json' \
  -d '{"tags":["Alpha","Beta"]}'

curl localhost:19277/kwarr/_count -H 'Content-Type: application/json' \
  -d '{"query":{"prefix":{"tags":"Beta"}}}'
curl -XPOST localhost:19277/kwarr/_flush
curl localhost:19277/kwarr/_count -H 'Content-Type: application/json' \
  -d '{"query":{"prefix":{"tags":"Beta"}}}'
```

Observed:

```
-- before flush --
prefix tags=Alpha -> {"count":1,...}
prefix tags=Beta  -> {"count":1,...}
-- after flush --
prefix tags=Alpha -> {"count":1,...}
prefix tags=Beta  -> {"count":0,...}
```

## Which answer is right

The **pre-flush** one. ES indexes one keyword term per array element, so `{"prefix":{"tags":"Beta"}}` matches. That is the opposite direction from #398, where the post-flush answer was the correct one — which is why #398 fixed the analyser question and left this alone rather than teaching the scan to mirror the join.

## Where the fix belongs

[...]

PR #665, opened 2026-08-23
fix(engine): #408 term on a keyword array is flush-invariant for counts

xerj-org:
### Symptom
A document `{"tags": ["red","green","blue"]}` answers `term:{tags:"green"}` with **0 hits while buffered** in the memtable, but **1 hit after a flush**. Same document, same query, two answers — decided by a background flush the caller cannot see. This is the `#423` "answers change at flush" class (symptom `#408`).

### Root cause (traced, not assumed)
It is **not** `doc_matches_query` (whose `json_values_equal` already matches any array element) and **not** the doc-values term/bool paths (which already bail for arrays). `size:5` returns the doc correctly — the miss is only in the **size:0 / `_count`** path.

`try_shortcut_count`'s bare-`term` keyword shortcut counts the memtable via `doc_values_keyword_count`, whose single-valued column stores **only the first array element** (`push_field` flattens arrays). Unlike `doc_values_term_query` / `doc_values_bool_hits`, this shortcut had **no array bailout**, so for `term:{tags:"green"}` it returned a confident `Some(0)` that **overwrote** the correct stored-scan count.

Measured: `size:0 → 0`, `size:5 → total=1`; `try_shortcut_count → Some(0)` while `mem_matches_known = None`.

### Fix
Bail the shortcut to the stored-source scan when the memtable marks the field array-valued (or whitespace-keyword) — new `ShardedFtsMemtable::term_count_needs_source_scan`, delegating to the same `dv_column_unusable` guard the other DV paths already use. The scan matches array membership via `json_values_equal`, so pre- and post-flush counts agree.

### Verification (1.97.1)
- New test `term_on_keyword_array_flush_parity` asserts pre-flush == post-flush.
- **Fail-before**: reverting only the bail hunk makes it fail (`pre=0`, `post=1`).
- Full `xerj-engine` suite green (ci-test): 54 test binaries, 0 failures.
- `fmt --all --check` clean; `clippy --all-targets -D warnings` clean.

Scoped to the reproduced memtable/flush divergence; the broader `#423` schema-threading is separate and not needed here.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -21055,6 +21055,18 @@ impl Index {
             _ => return None,
         };
 
+        // #423/#408: a keyword ARRAY (or whitespace keyword) stores only its
+        // FIRST element in the memtable's single-valued count column (see
+        // `push_field`), so counting `term` on any non-first element here would
+        // undercount — a confident, wrong 0 while the doc is buffered, then N
+        // after flush (size:0 = 0 while size:5 returned the doc). Abandon to the
+        // stored-source scan, which matches array membership via
+        // `doc_matches_query`. Mirrors the identical guard in
+        // `doc_values_term_query` / `doc_values_bool_hits`.
+        if self.memtable.term_count_needs_source_scan(field) {
+            return None;
+        }
+
         // Memtable side — M5.1 sharded count via per-shard count maps.
         // Each shard lazily rebuilds its own count map on first query
         // after an ingest; the sharded `doc_values_*_count` helpers
--- a/engine/crates/xerj-engine/src/memtable.rs
+++ b/engine/crates/xerj-engine/src/memtable.rs
@@ -1804,6 +1804,19 @@ impl ShardedFtsMemtable {
         })
     }
 
+    /// True when the memtable's single-valued doc-values column for `field`
+    /// cannot authoritatively answer a `term`/count query: the field has
+    /// carried an ARRAY (only the first element is stored — see `push_field`)
+    /// or a whitespace keyword (the column holds analyzed-text source). Callers
+    /// that count via the column MUST fall back to the stored-source scan,
+    /// which matches array membership. This is the same guard `doc_values_term_query`
+    /// and `doc_values_bool_hits` already apply; exposed so the `try_shortcut_count`
+    /// bare-`term` count path can bail identically (#423/#408: a `term` on a
+    /// keyword array counted 0 while buffered but N after flush).
+    pub fn term_count_needs_source_scan(&self, field: &str) -> bool {
+        self.dv_column_unusable(field)
+    }
+
     /// DocValues term query — aggregates hits across all shards.
     /// Returns `Some(Vec<(doc_id, local_idx)>)` if any shard matched.
     /// The `local_idx` is shard-local; callers use the doc_id to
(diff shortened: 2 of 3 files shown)


xerj-org:

All three REFUTE-lens skeptics ran the claimed commands under the CI toolchain (rustc 1.97.1) against HEAD `1a0e4e2c` and returned **NOT REFUTED / 0 BLOCKING**.

### Lens: CORRECTNESS — refuted=false, blocking=NONE
**Evidence:** shipped test 1 passed; probe 1 passed with pre==post==scan==expected on all 5 shapes; fail-before reproduced pre!=post on the two singular-term shapes only.
**Notes:** Claim holds and is symmetric. Pre-flush: memtable stores only the first array element in its count column, so `term_count_needs_source_scan` (dv_column_unusable: array_fields||keyword_has_whitespace, per-field union across shards) bails the singular-term count shortcut to the authoritative `doc_matches_query` scan. Post-flush: memtable is empty (guard returns false), the shortcut proceeds to segments, but `build_doc_value_columns` suppresses the column for any array-carrying field, so cols.get(field)=None triggers return None -> same scan. A segment that KEEPS a column is guaranteed scalar-only for that field, so doc_freq/range_count is exact. No symmetric hole. Bail does NOT over-trigger: it is per-field, so a plain keyword (color) in the same index as an array field (tags) is still fast-counted correctly (verified pre=post=2). The `terms` (plural) path was never affected.
**Non-blocking:** segment side is over-conservative (drops the DV column for any array-carrying field → scan even when the field is absent) — a pre-existing #362 perf cost, not introduced by this fix.

### Lens: REGRESSION-RISK — refuted=false, blocking=NONE
**Evidence:** Full xerj-engine suite under rustup 1.97.1 --profile ci-test: all 54 test binaries "test result: ok"; 0 failed everywhere. Lib = "626 passed; 0 failed". New fix test: "term_on_keyword_array_is_flush_invariant ... ok". Plain-keyword parity probe: pre_active=3/inactive=2/pending=1/absent=0, all equal post-flush. `cargo fmt --all --check` -> exit 0. `cargo clippy -p xerj-engine --profile ci-test --all-targets -- -D warnings` -> exit 0.
**Notes:** Guard is narrow: array_fields/keyword_has_whitespace are populated only on ingest of an actual array or a whitespace keyword, so a plain non-array non-whitespace keyword never bails and its fast-path count is unchanged (verified 3/2/1/0, flush-invariant). All three gates green under 1.97.1. No failing test, no clippy/fmt failure, no wrong-count regression.
**Non-blocking:** single-valued whitespace keywords (e.g. "New York") also divert the size:0 count shortcut to the stored scan — perf-only; the scan is the authoritative path and mirrors the existing guard in doc_values_term_query/doc_values_bool_hits, so count stays consistent with search.

### Lens: DOES-THE-EVIDENCE-HOLD — refuted=false, blocking=NONE
**Evidence:** FIXED baseline: `term_on_keyword_array_is_flush_invariant ... ok` (1 passed). BAIL-REMOVED (perl-revert of ONLY the `if self.memtable.term_count_needs_source_scan(field) { return None; }` block, index.rs-only per git diff): `test ... FAILED`, panic at line 71 "left: 0 / right: 1 (pre-flush=0, post-flush=1)". RESTORED: `test ... ok`. Size probe BUGGY pre-flush: size:0 total=0, size:5 total=1 (d0 present); post-flush size:0 total=1. Size probe FIXED pre-flush: size:0 total=1.
**Notes:** Evidence holds. Independently reproduced both PR claims under 1.97.1; the test genuinely pins the fix. Worktree restored clean.

**Disposition:** 0 BLOCKING across all three lenses. Non-blocking findings are pre-existing perf notes (#362 class), filed here rather than blocking; they do not change any count. Merging on all-green + CLEAN + 0 BLOCKING per the project's merge gate.

Issue #396, opened 2026-08-15
prefix and wildcard disagree by default on a keyword field: one field, one intent, two answers


xerj-org:
Split out of #362, which asked for "one rule ... applied to all term-level queries on a given field ... the three should not disagree". #362's fix makes `_count` agree with `_search` and makes `case_insensitive` work, but it does **not** close this half, so filing it rather than letting it disappear into a closed issue.

Written by an AI agent (Claude Opus 5) driven by @xerj-org.

## The disagreement

One `keyword` field, one value, two spellings of one intent, opposite answers.

Index:

```json
PUT /kw   {"mappings":{"properties":{"code":{"type":"keyword"}}}}
POST /kw/_doc/1?refresh=true   {"code":"HnswGraph"}
```

| query | hits |
|---|---|
| `{"prefix":{"code":"hnsw"}}` | **0** |
| `{"wildcard":{"code":"hnsw*"}}` | **1** |
| `{"prefix":{"code":"Hnsw"}}` | 1 |
| `{"wildcard":{"code":"Hnsw*"}}` | 1 |

A caller who writes `hnsw*` and a caller who writes `hnsw` are asking the same question about the same field and get different answers, with nothing in the response to say why.

## Cause

Two defaults, named in `engine/crates/xerj-engine/src/index.rs` (the constants block above `PREFIX_CASE_INSENSITIVE_DEFAULT`):

* `PREFIX_CASE_INSENSITIVE_DEFAULT = false` — ES's rule. Lucene builds a `prefix` automaton straight off the query bytes (`PrefixQuery.toAutomaton`, `lucene/core/src/java/org/apache/lucene/search/PrefixQuery.java:44`) and folds only when asked, via `Automata.makeCaseInsensitiveString` (`lucene/core/src/java/org/apache/lucene/util/automaton/Automata.java:573`).
* `KEYWORD_WILDCARD_CASE_INSENSITIVE_DEFAULT = true` — not ES's rule. It is what XERJ has always shipped for keyword wildcards, and the `term{case_insensitive:true}` lowering rewrites to a `Wildcard` node that depends on it.

## What fixing it means

[...]

PR #667, opened 2026-08-23
fix(engine): #396 prefix & wildcard case-folding is flush-invariant

xerj-org:
### Symptom (reproduced, fail-before)
The memtable matcher folded case differently from the segment, so hit counts changed across a background flush:
- **keyword prefix**: `prefix:{name:"hel"}` on `"Hello"` → 1 buffered / 0 flushed (ES keyword prefix is case-SENSITIVE).
- **text prefix**: `prefix:{body:"Hel"}` on `"Hello World"` → 1 buffered / 0 flushed (segment matches the lower-cased token vs the RAW query).
- **text wildcard**: `wildcard:{body:"Hel*"}` on `"Hello World"` → 1 buffered / 0 flushed (segment is case-sensitive on the pattern).

### Fix
Thread the index `Schema` into the memtable matcher (`doc_matches_query` → schemaless shim over `doc_matches_query_typed(q, source, &Schema)`) and make the **Prefix** and **Wildcard** arms mirror their segment paths per field type:
- **keyword prefix** → RAW value, case-sensitive (segment FST range).
- **text prefix / text wildcard** → lower-cased token stream vs the RAW query/pattern (segment analysed path, `case_insensitive:false`).
- **keyword wildcard** → fold both (the segment doc-values keyword-wildcard path folds; kept in agreement — flush-invariant).
- **unknown field** (schemaless shim callers: unit tests, `collect_matched_queries_inner`, `apply_function_score`, delete/update-by-query) → the pre-#423 fold-both heuristic, **byte-for-byte unchanged**.

Real schema is resolved once per search and passed to search_inner's brute-scan sites.

### Deliberately out of scope (follow-ups filed)
- **#668** — keyword `wildcard` is case-INSENSITIVE on both memtable+segment (ES default is case-sensitive). This PR keeps them in agreement (flush-invariant); making it ES-correct needs the segment path changed too.
- **#669** — `matched_queries` (`_name`) for a NAMED keyword prefix still uses the fold-shim, so the `_name` annotation can disagree with the (now case-sensitive) hit selection. The **hit set is correct**; only the annotation channel is affected.

### Verification (1.97.1)
- `prefix_keyword_case_flush_parity`: 4 tests — keyword prefix, text prefix, keyword wildcard, text wildcard — all assert pre-flush == post-flush.
- **Fail-before**: forcing the Prefix arm to fold restores the prefix divergence.
- Full `xerj-engine` suite (ci-test + dev), fmt, clippy `--all-targets -D warnings` — green.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -14131,6 +14131,13 @@ impl Index {
             )));
         }
 
+        // #396: the memtable brute-scan matcher must see the mapping so
+        // `prefix`/`wildcard` apply the same keyword/text folding rules as the
+        // segment matcher (buffered answers must not change at flush). Resolved
+        // once (owned; the read guard is released immediately, so no re-entrancy
+        // with the guards taken later in this function).
+        let dmq_schema = self.schema().await;
+
         // Implicit @timestamp "sort segments on timestamp" handling (see
         // `SearchRequest::leaf_ts_field`).  Two regimes:
         //
@@ -15613,7 +15620,7 @@ impl Index {
                     let matched = if let QueryNode::Ids { values } = query {
                         values.iter().any(|v| v == doc_id.as_str())
                     } else if !query_may_read_id || source.get("_id").is_some() {
-                        doc_matches_query(query, &source)
+                        doc_matches_query_typed(query, &source, &dmq_schema)
                     } else {
                         // Inject `_id` so deeply-nested Ids queries (e.g.
                         // function_score → ids) can resolve it from source.
@@ -15624,7 +15631,7 @@ impl Index {
                         } else {
                             (*source).clone()
                         };
-                        doc_matches_query(query, &source_with_id)
+                        doc_matches_query_typed(query, &source_with_id, &dmq_schema)
                     };
                     if matched {
                         // Materialise (score + owned source) ONLY when the
@@ -16584,7 +16591,11 @@ impl Index {
                                                 Value::String(id.clone()),
                                             );
                                         }
-                                        if !doc_matches_query(query, &src_with_id) {
+                                        if !doc_matches_query_typed(
+                                            query,
+                                            &src_with_id,
+                                            &dmq_schema,
+                                        ) {
                                             continue;
                                         }
                                         total_count += 1;
@@ -18482,7 +18493,7 @@ impl Index {
                         } else {
                             fg_owned = all_docs
                                 .iter()
-                                .filter(|d| doc_matches_query(query, d))
+                                .filter(|d| doc_matches_query_typed(query, d, &dmq_schema))
                                 .cloned()
                                 .collect();
                             &fg_owned[..]
@@ -35541,7 +35552,17 @@ fn multi_match_field_values_lc(source: &Value, field: &str) -> Option<Vec<String
 /// Evaluate a query against a single stored document source value.
 ///
 /// Returns true if the document matches the query.
+/// Schemaless compatibility shim (pre-#423 behaviour) for callers with no
+/// `Schema` in scope — unit tests, `collect_matched_queries_inner`,
+/// `apply_function_score`. The memtable search/scan paths call
+/// `doc_matches_query_typed` with the index's real schema so buffered-doc
+/// answers agree with the segment matcher (#396: keyword/text `prefix` folding).
 fn doc_matches_query(q: &QueryNode, source: &Value) -> bool {
+    static EMPTY_SCHEMA: std::sync::LazyLock<Schema> = std::sync::LazyLock::new(Schema::empty);
+    doc_matches_query_typed(q, source, &EMPTY_SCHEMA)
+}
+
+fn doc_matches_query_typed(q: &QueryNode, source: &Value, schema: &Schema) -> bool {
     match q {
         QueryNode::MatchAll => true,
         QueryNode::MatchNone => false,
@@ -35670,19 +35691,33 @@ fn doc_matches_query(q: &QueryNode, source: &Value) -> bool {
         },
 
         QueryNode::Prefix { field, value, .. } => {
-            // ES Prefix matches against analyzed tokens for text fields and
-            // against the raw value for keyword fields. Without schema info
-            // at this layer we check both: either the whole value starts
-            // with the pattern (keyword semantics) OR any whitespace-
-            // separated token does (analyzed-text semantics).
+            // #396: prefix semantics depend on the field's mapping, and ES never
+            // analyses the prefix VALUE itself.
+            //  - keyword: match the RAW stored value, case-SENSITIVE ("Hello"
+            //    does not start with "hel").
+            //  - analysed text: match the analysed (lower-cased) token stream
+            //    against the RAW query ("hel" prefixes "hello"; "Hel" does not).
+            //  - unknown field (schemaless shim caller — no `Schema` in scope):
+            //    preserve the pre-#423 fold-both heuristic so those callers are
+            //    byte-for-byte unchanged.
+            let ft = declared_field(schema, field).map(|f| &f.field_type);
+            let is_keyword = matches!(ft, Some(FieldType::Keyword));
+            let is_text = matches!(ft, Some(FieldType::Text));
             let value_lc = value.to_lowercase();
             let matches_str = |s: &str| -> bool {
-                let lc = s.to_lowercase();
-                if lc.starts_with(&value_lc) {
-                    return true;
+                if is_keyword {
+                    s.starts_with(value.as_str())
+                } else if is_text {
+                    s.split(|c: char| !c.is_alphanumeric()).any(|tok| {
+                        !tok.is_empty() && tok.to_lowercase().starts_with(value.as_str())
+                    })
+                } else {
+                    let lc = s.to_lowercase();
+                    lc.starts_with(&value_lc)
+                        || lc
+                            .split(|c: char| !c.is_alphanumeric())
+                            .any(|tok| !tok.is_empty() && tok.starts_with(&value_lc))
                 }
-                lc.split(|c: char| !c.is_alphanumeric())
-                    .any(|tok| !tok.is_empty() && tok.starts_with(&value_lc))
             };
             get_field_value(source, field)
                 .and_then(|v| match v {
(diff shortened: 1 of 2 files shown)


xerj-org:

All three refute-lens skeptics ran the claimed commands under rustc 1.97.1 against HEAD `a523a744` and returned **NOT REFUTED / 0 BLOCKING**.

### Lens: CORRECTNESS — refuted=False, blocking_findings=NONE

**Evidence:** HEAD=a523a744 on fix/396-prefix-keyword-case-flush; fix symbol doc_matches_query_typed present at index.rs:15623/15634/16594/18496.

(1) Shipped test GREEN under rustup 1.97.1: `cargo test -p xerj-engine --profile ci-test --test prefix_keyword_case_flush_parity` => 4 passed (prefix_keyword, prefix_text, wildcard_keyword, wildcard_text all flush-invariant).

(2) My throwaway test tests/zz_skeptic_396_prefix_wildcard_flush.rs (6 tests, ALL pass, pre==post on every row):
  A wildcard KEYWORD 'Hello': Hel*/hel*/HEL*/hel?o/*ell* -> pre==post==1 (fold-both, flush-invariant).
  B wildcard TEXT 'Hello World': Hel*=0/0, hel*=1/1, wor*=1/1, Wor*=0/0, *orl*=1/1, h?llo=1/1 (raw pattern vs lc token).
  C bool(keyword name + text body TOGETHER): (Hel,hel)=1/1, (hel,hel)=0/0 [kw case-sensitive miss], (Hel,Hel)=0/0 [text uc miss], (Hello,world)=1/1 — schema threaded correctly through the bool recursion.
  D prefix on UNDECLARED field 'ghost'='Hello' (default Dynamic mapping evolved it to TEXT): hel=1/1, Hel=0/0, HEL=0/0, Hello=0/0 — pre==post; both paths read the evolved schema (not the fold-both unknown branch).
  E keyword prefix POSITIVE 'Hel' on 'Hello' -> pre==post==1; empty prefix '' -> 1/1. False-negative guard PASSES.
  F keyword 'hello': hel=1/1, Hel=0/0, hello=1/1, hellox=0/0.

(3) False-negative direction (task item 3) is closed: XERJ has NO keyword normalizer (xerj-fts/src/analyzer.rs:1699 'normalizers are not supported'), so the keyword FST term == raw _source value; the memtable's case-sensitive raw starts_with(value) applies the identical predicate to the identical bytes the segment does. Text prefix: memtable lowercases the token then starts_with(raw query) == segment (lc token at index time, raw query). No input matches on the segment but misses in the memtable.

**Non-blocking:** (1) Claim wording overstates keyword handling: 'keyword => match RAW value case-SENSITIVE' is true only for PREFIX. For keyword WILDCARD the fix deliberately folds BOTH sides (memtable else-branch wildcard_match(lc, pat_lc); segment WildcardQuery{case_insensitive:true} at index.rs:41071). My shape A confirms 'Hel*','hel*','HEL*' all return 1 on 'Hello' both pre and post — flush-invariant, but NOT case-sensitive. This is an intentional, documented deviation (ES-correct case-sensitive keyword wildcard deferred to #668), so it does not break the flush-invariance contract; only the one-line summary is imprecise.  (2) Edge not covered by the listed shapes and NOT introduced by this fix: under DynamicMapping::Runtime ('accept but don't index'), a field lives in _source but is never mapped nor segment-indexed. The memtable unknown-branch fold-both could match while the segment has no FST term. Pre-existing behavior (old code also fold-both'd in the memtable) and Runtime fields are documented unsearchable; out of scope for #396. Did not exercise it since the required shape (undeclared field under default Dynamic mapping) evolves to text and stays pre==post.

**Notes:** Could NOT refute. The three memtable arms now mirror their segment paths per field type: keyword prefix = raw case-sensitive starts_with (segment PrefixQuery over keyword FST, index.rs:41014); text prefix = lc-token vs raw query (segment expand_prefix, raw starts_with over lc terms, 41027); keyword wildcard = fold-both (segment WildcardQuery case_insensitive:true, 41071); text wildcard = raw pattern vs lc token (segment case_insensitive:false, 41093). Schema is resolved once per search (dmq_schema = self.schema().await, index.rs:14137) and threaded through all three brute-scan call sites plus every recursive bool/dismax/nested/span arm. Verified the fix commit a523a744 rewires ~20 recursive doc_matches_query call sites to doc_matches_query_typed. The only residual divergence risk (Runtime unmapped fields, exotic-Unicode tokenizer mismatch) is pre-existing and not introduced by #396. The throwaway test file was left in the isolated worktree (tests/zz_skeptic_396_prefix_wildcard_flush.rs).

### Lens: REGRESSION-RISK — refuted=False, blocking_findings=NONE

**Evidence:** HEAD=a523a744 on origin/fix/396-prefix-keyword-case-flush; fix symbol present (doc_matches_query_typed at index.rs:15623/15634/16594/18496/35558). Toolchain rustc 1.97.1.

(1) KEY CHECK — cargo test -p xerj-engine --profile ci-test: EXIT 0. All 55 test binaries reported "test result: ok". Aggregate: 1068 passed; 0 failed; 23 ignored. No FAILED lines anywhere. Breakdown of the load-bearing ones: lib unittests src/lib.rs => 626 passed; 0 failed (these are the doc_matches_query 2-arg unit tests at index.rs:44656+ / 45509+ — they still pass, proving the empty-schema shim is byte-identical). integration.rs => 126 passed; 0 failed; 1 ignored. New tests/prefix_keyword_case_flush_parity.rs => 4 passed; 0 failed (prefix_keyword_case_is_flush_invariant, prefix_text_case_is_flush_invariant, wildcard_keyword_is_flush_invariant, wildcard_text_is_flush_invariant). No existing prefix/wildcard test that encoded the OLD fold behaviour failed.

(2) cargo clippy -p xerj-engine --profile ci-test --all-targets -- -D warnings: EXIT 0, clean (only grep hits for "error" are the thiserror crate name). cargo fmt --all --check: EXIT 0.

(3) Schemaless shim unchanged — CONFIRMED. doc_matches_query(q,source) now = doc_matches_query_typed(q, source, &EMPTY_SCHEMA) with EMPTY_SCHEMA = LazyLock::new(Schema::empty). Schema::empty() has fields: Vec::new() (xerj-common/types.rs:468), so declared_field (index.rs:31868) returns None => the Prefix and Wildcard arms take the else = pre-#423 fold-both branch, byte-for-byte. Grep confirms the 2-arg shim is still used (NOT switched to _typed) by: unit tests (44656+/45509+), collect_matched_queries_inner (37583 => matched_queries/_name path, 37586/37632), score_query_against_doc (37773+), apply_function_score (42982 => 43003), plus the kNN brute-force filter (12533/12589), delete-by-query/hydrate/scan paths (24572/25123/25125), and count-shortcut/other brute paths (11159/11277/20746/20971). ONLY search_inner's main brute-scan (15620/15631/16594/18493) was switched to _typed with the index's real schema (dmq_schema = self.schema().await, resolved once, guard released — schema() clones and drops the read guard, so no re-entrancy with later guards). matched_queries (named prefix) and delete-by-query prefix paths therefore behave exactly as before.

**Notes:** Not refuted: the change is tightly scoped and no test regressed. Only the memtable brute-scan inside the search path becomes schema-aware; every schemaless caller preserves prior behaviour via the empty-schema shim (unknown field => old fold-both).

Two non-blocking observations (neither is a regression introduced by this change):
- Completeness gap, acknowledged in the commit: matched_queries/_name (collect_matched_queries_inner, #669), and the kNN-brute-force-filter / delete-by-query / hydrate / count-shortcut buffered paths still use the empty-schema shim. A keyword prefix/wildcard evaluated through those paths can still show the #396 flush-divergence (memtable fold-both vs segment case-sensitive/raw). This is pre-existing behaviour left intentionally unchanged (follow-ups #668/#669), not a new regression.
- Minor cost: the search path now does one self.schema().await per call (schema read-lock + Vec<FieldConfig> clone). Off the inner per-doc loop; negligible, not a blocker.

No cargo command exceeded the ~5-minute budget: --no-run warm compile finished exit 0, full test run finished exit 0 (~well under limit), clippy 12.78s check phase, fmt instant.

### Lens: DOES-THE-EVIDENCE-HOLD — refuted=False, blocking_findings=NONE

**Evidence:** HEAD=a523a744 on fix/396-prefix-keyword-case-flush. Fix present: doc_matches_query_typed (index.rs:35565); Prefix arm (35693-35732) derives is_keyword (35704) and is_text (35705) from declared_field(schema, field).

BASELINE (unmodified fixed code, 1.97.1): all 4 tests pass.

FAIL-BEFORE: perl-forced ONLY the two Prefix-arm bindings to `false` (git diff = exactly 2 lines changed, 35704/35705; Wildcard arm's multi-line is_text at 35751 confirmed untouched), so the arm falls into the fold (else) branch. Rebuilt + ran under 1.97.1:
- prefix_keyword_case_is_flush_invariant FAILED: left(pre-flush)=1, right(post-flush)=0 -- panic at test line 59 ('memtable Prefix arm folds case while the segment is case-sensitive').
- prefix_text_case_is_flush_invariant FAILED: pre-flush=1, post-flush=0 -- panic at test line 100.
Result: 2 passed; 2 failed.

FAIL-AFTER: `git checkout -- engine/crates/xerj-engine/src/index.rs` restored both bindings (grep confirms matches!(ft, Some(FieldType::Keyword/Text)) back), rebuilt, re-ran under 1.97.1: 4 passed; 0 failed.

The test DOES pin the fix: it fails when the schema-aware branch is bypassed and passes only with it.

**Non-blocking:** (1) Same-file caveat (not blocking, pin is real): both fail-before and fail-after ran on the same test binary path. I recompiled between each run under 1.97.1 (fold-forced build finished in 10.53s and re-emitted the binary; restored build recompiled xerj-engine in 10.65s), and the results flipped 2-failed -> 0-failed, so the pass is not a stale-binary artifact.  (2) The two wildcard guard tests (wildcard_keyword_is_flush_invariant, wildcard_text_is_flush_invariant) are unaffected by the Prefix-arm edit and stayed green in all three runs, which is expected since they exercise the Wildcard arm's own is_text binding (multi-line matches!(...) at index.rs:35751, left untouched).

**Notes:** Both halves of the PR measurement independently confirmed by the fail-before panics:
1) Buggy (folded) prefix "hel" on keyword "Hello": pre-flush=1 vs post-flush=0; fixed => both 0 (keyword test asserts post==0 AND pre==post, and passes on the restored code).
2) Text prefix "Hel" on "Hello World": buggy folded pre-flush=1 vs post-flush=0; fixed => both 0 (text test asserts post==0 AND pre==post, and passes on the restored code).

Mechanism verified in source: with is_keyword=true the closure does s.starts_with("hel") on "Hello" => false (case-sensitive raw keyword compare, agreeing with the segment). With both forced false, the else branch lowercases (lc="hello", value_lc="hel") => "hello".starts_with("hel")=true, reproducing the memtable/segment split-brain the flush exposes.

Hygiene: ran only `cargo test -p xerj-engine --test prefix_keyword_case_flush_parity` under `rustup run 1.97.1` (CI toolchain, per memory note on toolchain drift); never pointed any conformance suite at :9200. All throwaway edits reverted -- final `git status --porcelain` is empty and HEAD=a523a744. No new/untracked files created.

**Disposition:** 0 BLOCKING across all three lenses. Non-blocking items are pre-existing / documented — matched_queries `_name` → #669; ES-correct keyword wildcard → #668; negligible per-search schema clone. Merging on all-green + CLEAN + 0 BLOCKING per the merge gate.

PR #671, opened 2026-08-23
fix(engine): #668 keyword wildcard is case-sensitive (ES default)

### LINKED ISSUE #668: wildcard on keyword should be case-sensitive (ES default); currently folds on both sides
## `wildcard` on a `keyword` field is case-INSENSITIVE (should be case-sensitive, ES default)

Split off from #396 (#667). ES matches `wildcard` on a `keyword` field **case-sensitively** by default (the value is not analysed). XERJ currently folds case on **both** sides:

- memtable: `doc_matches_query`'s Wildcard arm lowercases pattern + source (index.rs, `pat_lc`).
- segment: the doc-values wildcard path `ScoredFilterLeaf::KeywordWildcard` (index.rs:22550-22580) lowercases `pattern` and each term.

Because both sides fold, the behaviour is **flush-invariant** (no "answers change at flush"), so this is a wrong-results / ES-compat gap, not a data or flush bug — which is why #396/#667 deliberately left it alone (making only the memtable case-sensitive would have split-brained it at flush; caught in review).

### Fix (must change BOTH sides together)
Make keyword `wildcard` case-sensitive on the memtable arm **and** the segment `ScoredFilterLeaf::KeywordWildcard` path in the same change, then update the stale parity comment at index.rs:22554-22556. Add a repro: `wildcard:{name:"HELLO"}` vs `{name:"hello"}` on doc `"Hello"` should match neither once fixed (case-sensitive), and must give the same answer pre- and post-flush. Guard test `wildcard_keyword_is_flush_invariant` (added in #667) must keep passing.

`case_insensitive: true` on the query should still fold (both sides).

---

**Refined scope (investigated 2026-08-23):** a blanket flip to case-sensitive is WRONG. The current keyword-wildcard case-insensitivity is **load-bearing** — comment at `index.rs:41061-41065`: XERJ's parser rewrites `term{case_insensitive:true}` (and some `query_string` case-insensitive forms) **into a `Wildcard` node**, relying on the matcher folding case. `QueryNode::Wildcard` (`xerj-query/src/ast.rs:362`) carries only `field/value/boost/constant_score` — it does **not** preserve a `case_insensitive` flag, so the two intents (genuine case-sensitive wildcard vs. rewritten case-insensitive term) are indistinguishable downstream, and everything folds.

**Correct fix (cross-crate, thread the flag):**
1. Add `case_insensitive: bool` (default false) to `QueryNode::Wildcard` (ast.rs).
2. Parser: set `false` for a standalone `{wildcard:{…}}` (ES default = case-sensitive), `true` for the `term{case_insensitive:true}` → Wildcard rewrite (and honor query_string's own lowering).
3. Memtable wildcard arm (`doc_matches_query_typed`, index.rs ~35739): if `case_insensitive` OR text → fold/token; else keyword → RAW `wildcard_match` (case-sensitive).
4. Segment FTS route (`index.rs:41067-41071`): pass the node's flag instead of the hardcoded `case_insensitive: true`.
5. Segment DV path `ScoredFilterLeaf::KeywordWildcard` (`index.rs:22550`): honor the flag.
6. ES-compat conformance (CI-only) validates keyword wildcard — verify no regression there.

Non-trivial (parser + AST + both matcher paths). #667 added a `wildcard_keyword_is_flush_invariant` guard that must keep passing (flush-invariance preserved regardless of the case decision).

xerj-org:
### Symptom
`wildcard:{name:"hell*"}` on keyword `"Hello"` matched (1). ES matches a `wildcard` on a `keyword` field **case-sensitively** (the pattern is not analysed), so it must NOT match. XERJ folded case on all three matcher paths.

### Why a blanket flip was wrong (the constraint)
XERJ's parser rewrites `term{case_insensitive:true}` (and some `query_string` forms) **into a `Wildcard` node** relying on the matcher folding case (`index.rs:41061`). A naive case-sensitive flip breaks those passing paths.

### Fix — thread a `case_insensitive` flag
- **AST** (`xerj-query/ast.rs`): add `case_insensitive: bool` (default `false`) to `QueryNode::Wildcard`.
- **Parser**: `false` for a standalone `{wildcard}` (ES default; object form parses the param), `true` for the `term{case_insensitive:true}` rewrite and `query_string` lowerings.
- **All three keyword-wildcard matcher paths honor it consistently** (so nothing split-brains at flush):
  - memtable `doc_matches_query_typed` Wildcard arm — keyword: raw `wildcard_match` when case-sensitive, else fold.
  - segment FST route (`index.rs:41081`) — `WildcardQuery.case_insensitive = node flag`.
  - segment doc-values `KeywordWildcard` (`index.rs:22550`, leaf gains the flag) — raw term match when case-sensitive, else fold.
- `text` fields and unknown (schemaless shim) fields are unchanged.

### Verification (1.97.1)
- `wildcard_keyword_case_sensitive`: keyword wildcard is case-sensitive, `term{case_insensitive:true}` still folds, and both are **flush-invariant** (memtable == segment).
- Fail-before: reverting the arm/route hunks restores the match.
- Full `xerj-engine` suite (ci-test + dev), fmt, clippy `-D warnings` — green.

Closes #668.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -22547,28 +22547,34 @@ impl Index {
                             ords: (lo as u32..hi as u32).collect(),
                         });
                     }
-                    ScoredFilterLeaf::KeywordWildcard { field, pattern } => {
+                    ScoredFilterLeaf::KeywordWildcard {
+                        field,
+                        pattern,
+                        case_insensitive,
+                    } => {
                         let Some(Column::Keyword(k)) = cols.get(field.as_str()) else {
                             return None;
                         };
-                        // SAME predicate as the brute path (`doc_matches_query`)
-                        // and the FST route (`term_matches_wildcard`): fold BOTH
-                        // sides, whole-value OR sub-token.  The old range+match
-                        // here was case-SENSITIVE, so a cased pattern (`K*1`,
-                        // case_insensitive) found nothing in a lowercase
-                        // dictionary — DV-only keyword wildcards returned 0 hits
-                        // from segments while the memtable matched.  A folded
-                        // pattern cannot bound a byte-ordered prefix range, so
-                        // walk the whole (distinct-term) dictionary — the same
-                        // cost class as the FST expansion.
+                        // #668: honor the node's case flag — the SAME predicate
+                        // the brute path (`doc_matches_query`) and the FST route
+                        // now apply. `term{case_insensitive:true}` / query_string
+                        // rewrites fold BOTH sides (whole-value OR sub-token); a
+                        // standalone `{wildcard}` matches the RAW term
+                        // case-sensitively (ES default). A folded pattern cannot
+                        // bound a byte-ordered prefix range, so walk the whole
+                        // (distinct-term) dictionary — same cost class as the FST.
                         let pat_lc = pattern.to_lowercase();
                         let ords: Vec<u32> = (0..k.terms.len())
                             .filter(|&o| {
-                                let lc = k.terms[o].to_lowercase();
-                                wildcard_match(&lc, &pat_lc)
-                                    || lc
-                                        .split(|c: char| !c.is_alphanumeric())
-                                        .any(|tok| !tok.is_empty() && wildcard_match(tok, &pat_lc))
+                                if *case_insensitive {
+                                    let lc = k.terms[o].to_lowercase();
+                                    wildcard_match(&lc, &pat_lc)
+                                        || lc.split(|c: char| !c.is_alphanumeric()).any(|tok| {
+                                            !tok.is_empty() && wildcard_match(tok, &pat_lc)
+                                        })
+                                } else {
+                                    wildcard_match(&k.terms[o], pattern.as_str())
+                                }
                             })
                             .map(|o| o as u32)
                             .collect();
@@ -35737,30 +35743,40 @@ fn doc_matches_query_typed(q: &QueryNode, source: &Value, schema: &Schema) -> bo
         QueryNode::Wildcard {
             field,
             value: pattern,
+            case_insensitive,
             ..
         } => {
-            // #396: `wildcard` must mirror its segment path per field type
+            // #396/#668: `wildcard` must mirror its segment path per field type
             // (ES never analyses the pattern):
             //  - analysed TEXT: the segment matches the RAW pattern against the
             //    lower-cased token stream (case_insensitive:false), so an
-            //    uppercase pattern matches nothing — do the same here (this is
-            //    the flush-divergence the KEYWORD-only earlier attempt missed).
-            //  - KEYWORD: the segment doc-values keyword-wildcard path folds
-            //    both sides (case-insensitive), so fold here to stay in
-            //    agreement. ES-correct case-sensitive keyword wildcard needs the
-            //    segment path changed too — tracked in #668.
-            //  - unknown field (schemaless shim caller): the pre-#423 fold-both
-            //    heuristic, byte-for-byte unchanged.
-            let is_text = matches!(
-                declared_field(schema, field).map(|f| &f.field_type),
-                Some(FieldType::Text)
-            );
+            //    uppercase pattern matches nothing — do the same here.
+            //  - KEYWORD, case-sensitive (standalone `{wildcard}`, the ES
+            //    default): match the RAW value with `wildcard_match` (#668).
+            //  - KEYWORD, case-insensitive (`term{case_insensitive:true}` /
+            //    query_string lowerings set the flag): fold both sides.
+            //  - field absent from the schema (the `EMPTY_SCHEMA` shim callers,
+            //    or a genuinely-undeclared field): the pre-#423 fold-both
+            //    heuristic. NB memtable search passes the real, dynamically-grown
+            //    schema, where an undeclared string is already mapped to `text`
+            //    and takes the branch above — so this arm is effectively the
+            //    schemaless-shim path, byte-for-byte unchanged.
+            let ft = declared_field(schema, field).map(|f| &f.field_type);
+            let is_text = matches!(ft, Some(FieldType::Text));
+            // Case-sensitive only for a declared keyword whose node did NOT ask
+            // to fold. Unknown fields (ft == None) keep folding (shim parity).
+            let keyword_case_sensitive =
+                matches!(ft, Some(FieldType::Keyword)) && !*case_insensitive;
             let pat_lc = pattern.to_lowercase();
             let matches_str = |s: &str| -> bool {
                 if is_text {
                     s.split(|c: char| !c.is_alphanumeric()).any(|tok| {
                         !tok.is_empty() && wildcard_match(&tok.to_lowercase(), pattern.as_str())
                     })
+                } else if keyword_case_sensitive {
+                    // A keyword is a single un-analysed term: match the whole
+                    // raw value, case-sensitively (mirrors the segment FST).
+                    wildcard_match(s, pattern.as_str())
                 } else {
                     let lc = s.to_lowercase();
                     wildcard_match(&lc, &pat_lc)
@@ -41052,7 +41068,7 @@ fn query_node_to_fts_with_keyword_fields(
             value,
             boost,
             constant_score,
-            case_insensitive: _,
+            case_insensitive,
         } => {
             // `wildcard` on a KEYWORD field: expand the keyword FST term
             // dictionary to every term matching the pattern (`*`=0+ chars,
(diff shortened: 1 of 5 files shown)


xerj-org:

All three returned **NOT REFUTED / 0 BLOCKING** against HEAD `d8099d31`.

### Lens: CORRECTNESS — refuted=False, blocking=NONE

**Evidence:** HEAD=d8099d31 on fix/668-keyword-wildcard-case; fix present (ast.rs:378 case_insensitive flag). Referenced test PASSED: `rustup run 1.97.1 cargo test -p xerj-engine --profile ci-test --test wildcard_keyword_case_sensitive` -> 1 passed.

Wrote TWO throwaway integration tests, both PASSED (every case pre==post AND ES-correct value).

Test 1 (name=Keyword "Hello", tags=Keyword ["Hello","World"], body=Text "Hello World", ghost undeclared), pre(memtable)/post(segment):
  kw name hell*=0/0, Hell*=1/1, HELL*=0/0, h?llo=0/0, H?llo=1/1, Hell?=1/1, exact Hello=1/1, exact hello=0/0  (case-SENSITIVE, ES-correct)
  wildcard{value:hell*,case_insensitive:true}=1/1  (explicit ci folds)
  term{case_insensitive:true} value hello=1/1, HELLO=1/1  (STILL FOLDS)
  query_string name:hell*=1/1, name:Hell*=1/1, default_field hell*=1/1  (STILL FOLDS; parser lowercases)
  kw ARRAY tags wor*=0/0, Wor*=1/1, World=1/1, hell*=0/0, Hell*=1/1  (case-SENSITIVE)
  TEXT body hello=1/1, Hello=0/0, hell*=1/1, Hell*=0/0, wor*=1/1  (UNCHANGED per #667)
  unknown ghost hello=1/1, Hello=0/0, hell*=1/1, Hell*=0/0  (flush-invariant; dynamically text-mapped)

Test 2 (adversarial: docs Hello/HELLO/hello across 2 segments + memtable; bare constant_score fast-plan vs bool{filter:[wildcard]} FST route): for patterns hell*,Hell*,HELL*,h*o,H*O,*ELL*,*ell*(=2),Hello,HELLO,hello — bare pre==post, bool pre==post, AND bare==bool for every pattern (DV fast-plan and FST route agree), all ES-correct. 0 divergences.

Verified all three matcher paths honor the flag: memtable doc_matches_query_typed (index.rs:~35760, keyword_case_sensitive = declared Keyword && !case_insensitive -> raw wildcard_match else fold); segment FST expand_wildcard (fts search.rs:920, case_insensitive false -> raw wildcard_match(t,pattern), true -> term_matches_wildcard fold whole-or-subtoken); segment DV KeywordWildcard leaf (index.rs:~22550, raw wildcard_match when !case_insensitive else fold). Standalone wildcard sets case_insensitive:false (parser.rs:1292); term-ci sets true+constant_score:false (parser.rs:1043); query_string wildcard lowercases value + true (parser.rs:2107-2116). No pre!=post divergence and no term-ci/query_string path stopped folding.

**Non-blocking:** (1) The 'unknown/undeclared field' case is not actually schemaless at query time: an undeclared string field ('ghost') gets dynamically mapped to TEXT, so the schema fed to doc_matches_query_typed sees it as Text and it takes the text-wildcard path (raw pattern vs lowercased token), not the ft==None fold-both shim branch. Behavior is still flush-invariant (pre==post) and unchanged by this fix, so it does not affect the claim, but the fix comment's 'unknown field (schemaless shim caller) keeps folding' only applies to callers that pass EMPTY_SCHEMA (the brute paths at index.rs:35569), not to memtable search which uses the real, dynamically-grown schema.

**Notes:** Could NOT refute. The claim holds under a comprehensive throwaway harness that is strictly broader than the shipped test: it separately exercised the DV fast-plan and the FST route (bool-wrapped) and confirmed they agree, tested keyword ARRAY fields, multi-segment + memtable mixes, symmetric casing over {Hello,HELLO,hello}, and both term-ci and query_string fold-paths — all flush-invariant and ES-correct. Cleaned up both throwaway test files; worktree left clean (git status empty). One non-blocking note about the 'unknown field' framing (dynamic-mapping-to-text vs empty-schema shim) — does not affect correctness or the claim.

### Lens: REGRESSION — refuted=False, blocking=NONE

**Evidence:** HEAD=d8099d31 (fix present: engine/crates/xerj-query/src/ast.rs:378 `case_insensitive: bool`). Toolchain rustc 1.97.1 confirmed installed.

1) `rustup run 1.97.1 cargo test -p xerj-query --profile ci-test` — ALL GREEN: unit `test result: ok. 183 passed; 0 failed; 0 ignored`; span_clause_refusal `9 passed; 0 failed`; doc-tests `0 passed; 0 failed; 1 ignored`. Zero failures.

2) `rustup run 1.97.1 cargo test -p xerj-engine --profile ci-test` — ALL GREEN across every test binary; headline unit suite `test result: ok. 626 passed; 0 failed; 0 ignored`, plus binaries reporting 126/0, 65/0, 21/0, 15/0, 13/0, etc. No `FAILED`, `panicked`, `error[`, or `failures:` lines. Total: every `test result:` line = `ok ... 0 failed`.
   - Targeted re-run of the new #668 test: `cargo test -p xerj-engine --profile ci-test --test wildcard_keyword_case_sensitive` => `test keyword_wildcard_is_case_sensitive_but_term_ci_still_folds ... ok`, `test result: ok. 1 passed; 0 failed; 0 ignored`. (Not ignored, despite the file's historical doc comment.)

3) `rustup run 1.97.1 cargo clippy -p xerj-query -p xerj-engine --profile ci-test --all-targets -- -D warnings` — `Finished \`ci-test\` profile` with zero warnings/errors (would have hard-failed under -D warnings if any existed).

4) `cargo fmt --all --check` — FMT_EXIT=0, no diff.

**Notes:** The fix threads a new `case_insensitive` flag through `QueryNode::Wildcard` (ast.rs + parser.rs) rather than blanket-flipping keyword-wildcard case behavior. Standalone `{wildcard}` on a keyword field is now case-SENSITIVE (ES-correct, flag=false), while `term{case_insensitive:true}` and query_string lowerings set flag=true and keep folding. Because the rewrite path preserved its folding, no pre-existing keyword-wildcard test that encoded the old case-insensitive behavior regressed — the exact regression the lens was probing for did not materialize. Matcher parity is enforced on all three paths (brute `doc_matches_query_typed`, FST `query_node_to_fts_with_keyword_fields`, and DV `KeywordWildcard` scored leaf in index.rs). No cargo command exceeded ~5 min (engine tests compiled+ran within the timeout; incremental recompiles were fast). One xerj-query env-source setup line was refused by the worktree sandbox guard but was not needed for the verification.

### Lens: EVIDENCE — refuted=False, blocking=NONE

**Evidence:** HEAD d8099d31 on origin/fix/668-keyword-wildcard-case; fix present (case_insensitive flag threaded through QueryNode::Wildcard, ast.rs:369-378).

The memtable Wildcard arm lives in doc_matches_query_typed() at engine/crates/xerj-engine/src/index.rs:35743-35794. keyword_case_sensitive = keyword && !case_insensitive (line 35764-35765); the case-sensitive branch is index.rs:35772-35775 (`} else if keyword_case_sensitive { wildcard_match(s, pattern.as_str()) }`).

Shipped test: engine/crates/xerj-engine/tests/wildcard_keyword_case_sensitive.rs, fn keyword_wildcard_is_case_sensitive_but_term_ci_still_folds. It indexes keyword name="Hello", queries wildcard hell*/Hell* + term{ci:true} pre-flush (memtable) and post-refresh (segment), and asserts pre==post (flush-invariance) plus post_lower==0 (case-sensitive), post_cased==1, post_ci==1.

RESULT 1 — baseline (full fix, unmodified branch): built (default toolchain 1.92.0, 50.8s) and ran -> "test result: ok. 1 passed; 0 failed". Test PASSES.

RESULT 2 — surgical revert of ONLY the memtable arm: removed lines 35772-35775 so keyword falls into the else (fold) branch; segment FST route and DV path left untouched. Confirmed diff = "1 file changed, 4 deletions(-)" in index.rs only (keyword_case_sensitive now an unused-variable WARNING, not an error — no deny(warnings)). Rebuilt (12.3s) and ran the SAME shipped test -> FAILED:
  thread panicked at tests/wildcard_keyword_case_sensitive.rs:59:5:
  assertion `left == right` failed: #668: keyword wildcard must be flush-invariant
    left: 1   (memtable folds: hell* matches "Hello" -> pre_lower=1)
   right: 0   (segment path still case-sensitive -> post_lower=0)
  "test result: FAILED. 0 passed; 1 failed". This is exactly the predicted memtable-folds=1 vs segment-case-sensitive=0 flush-invariance divergence.

RESTORE: git checkout -- engine/crates/xerj-engine/src/index.rs; `git status --porcelain` now empty (clean), branch back at index.rs:35772.

Conclusion: the 3-path consistency is load-bearing. Reverting one of the three paths (memtable) while leaving the other two fixed makes the shipped test fail precisely on flush-invariance, so the test genuinely pins cross-path agreement rather than passing vacuously. NOT refuted. No conformance suite was pointed at :9200 (in-process Engine only).

**Notes:** Toolchain used was the local default rustc 1.92.0, not CI's 1.97.1 (per project memory on CI drift). This does not weaken the evidence: the assertion that fired is pure matcher logic (case-fold vs case-sensitive wildcard on a keyword field), which is toolchain-independent; the divergence is a semantic result, not a codegen artifact. The unused-variable warning after the revert is expected and harmless (no deny(warnings) in the crate or .cargo/config.toml). All git ops and the build/test ran against the isolated worktree only.

**Disposition:** 0 BLOCKING across all three lenses. The one non-blocking note (imprecise 'unknown field = schemaless shim' comment wording; behavior is flush-invariant regardless) is a doc nit, refined on rebase. Merging on all-green + CLEAN + 0 BLOCKING after the rc.56 stamp (#670) lands and this branch rebases past the tag-ahead Landing guard.

PR #672, opened 2026-08-23
fix(autoindex): #416 scope catalog doc IDs by corpus prefix (no cross-corpus overwrite)

### LINKED ISSUE #416: autoindex catalog IDs are global: two corpora sharing one byte-identical file overwrite each other's canonical entry
Surfaced while fixing #360, which removes a false abort this collision was causing. The abort is gone; the collision is not.

## Summary

`autoindex`'s catalog uses **global** document IDs — `file:<content_id>` and `ds:<slug>` — in one index (`autoindex-catalog`) that `--prefix` does not namespace. Two unrelated corpora on the same node that share a byte-identical file therefore share that file's catalog document.

## Consequence

The shared file's canonical document belongs to whichever run published it last. After a second corpus indexes the same file, the first corpus's alias documents point at a `duplicate_of` whose canonical entry now belongs to the other corpus. An Apache-2.0 `LICENSE` is the everyday case — most checkouts have one.

Before #360's fix this produced a loud abort (`catalog canonical/alias count disagrees with desired group axg1-…`) on a corpus whose data was complete. #360 correctly removes that false abort. The underlying identity collision is unchanged, so the failure mode moves from "noisy and wrong" to "quiet and wrong".

## Suggested

Namespace the catalog by `--prefix`, the way the data indices already are — either a per-prefix catalog index or a prefix component in the document ID. `ds:<slug>` has the same shape of problem.

**Verified:** the overwrite behaviour was reproduced on a real node against binaries built both before and after #360's fix, so the collision is confirmed pre-existing rather than introduced.
**Assumed:** I have not surveyed which other catalog document kinds share the global-ID pattern, nor whether `xerj brain` reuses the same catalog.

xerj-org:
## #416 (data-loss): autoindex catalog doc IDs are global → corpora overwrite each other

The catalog is one global index (`autoindex-catalog`) shared by **every** corpus, but its doc IDs were global — `file:{key}`, `ds:{slug}`, `file-alias:{...}`. Two corpora indexing a file with the same key (or producing the same slug) **overwrote each other's catalog entry**, losing the earlier corpus's "what was indexed" metadata.

### Fix
Thread the corpus `--prefix` (`execution.prefix` / `cfg.prefix`) into every catalog id: `file:{prefix}:{key}`, `ds:{prefix}:{slug}`, `file-alias:{prefix}:{...}`. Critically, the **write path** (`project_generation`) and the **stale-sweep prior-id computation** (`managed_non_run_ids`, and the http-test `prior_managed_ids` helper) must agree — else stale entries never get swept (orphan rows). Both threaded; the projection/reconcile tests verify the stale sets match.

### Verification (1.97.1)
- New `catalog_ids_are_prefix_scoped_across_corpora`: two prefixes → distinct ids; same prefix → stable/idempotent.
- **Fail-before**: dropping the prefix from the id bodies makes them collide (`file:k == file:k`).
- Full `xerj-autoindex` ci-test suite green; fmt + clippy `-D warnings` clean.

Closes #416.

### DIFF
--- a/engine/crates/xerj-autoindex/src/catalog.rs
+++ b/engine/crates/xerj-autoindex/src/catalog.rs
@@ -123,6 +123,8 @@ pub const GOTCHAS: &[&str] = &[
 ];
 
 pub struct DatasetDocInput<'a> {
+    /// Corpus scope (`--prefix`): keeps catalog ids from colliding across corpora (#416).
+    pub prefix: &'a str,
     pub pd: &'a PlanDataset,
     pub record_count: u64,
     pub junk_records: u64,
@@ -142,7 +144,7 @@ pub struct DatasetDocInput<'a> {
 }
 
 pub fn dataset_doc(inp: &DatasetDocInput) -> (String, Value) {
-    let id = format!("ds:{}", inp.pd.slug);
+    let id = format!("ds:{}:{}", inp.prefix, inp.pd.slug);
     let fields_json = serde_json::to_string(&inp.pd.specs).unwrap_or_else(|_| "[]".into());
     let doc = json!({
         "doc_kind": "dataset",
@@ -170,12 +172,13 @@ pub fn dataset_doc(inp: &DatasetDocInput) -> (String, Value) {
 /// Catalog document id for a file. One definition, because the sweep that
 /// removes a junk/skipped file's document (`lib.rs`, "stale junk-catalog
 /// sweep") has to name the id without holding the document.
-pub fn file_id(file_key: &str) -> String {
-    format!("file:{file_key}")
+pub fn file_id(prefix: &str, file_key: &str) -> String {
+    format!("file:{prefix}:{file_key}")
 }
 
 #[allow(clippy::too_many_arguments)] // 1:1 with the file-status doc's fields
 pub fn file_doc(
+    prefix: &str,
     file_key: &str,
     path: &str,
     format: &str,
@@ -187,7 +190,7 @@ pub fn file_doc(
     run_id: &str,
 ) -> (String, Value) {
     (
-        file_id(file_key),
+        file_id(prefix, file_key),
         json!({
             "doc_kind": "file",
             "file_key": file_key,
@@ -204,6 +207,7 @@ pub fn file_doc(
 }
 
 pub fn duplicate_file_doc(
+    prefix: &str,
     file_key: &str,
     path: &str,
     path_id: &str,
@@ -211,7 +215,7 @@ pub fn duplicate_file_doc(
     bytes: u64,
     run_id: &str,
 ) -> (String, Value) {
-    let alias_id = duplicate_file_id(file_key, path, path_id);
+    let alias_id = duplicate_file_id(prefix, file_key, path, path_id);
     (
         alias_id,
         json!({
@@ -230,10 +234,11 @@ pub fn duplicate_file_doc(
     )
 }
 
-pub fn duplicate_file_id(file_key: &str, path: &str, path_id: &str) -> String {
+pub fn duplicate_file_id(prefix: &str, file_key: &str, path: &str, path_id: &str) -> String {
     let identity = if path_id.is_empty() { path } else { path_id };
     format!(
-        "file-alias:{}",
+        "file-alias:{}:{}",
+        prefix,
         crate::ids::doc_id("duplicate-file", file_key, identity)
     )
 }
@@ -778,5 +783,24 @@ mod sample_query_tests {
         assert!(full_text["body"]["query"].get("match_phrase").is_none());
         assert!(full_text["body"].get("_source").is_none());
         assert!(full_text["body"].get("fields").is_none());
+    }
+
+    /// #416 (data-loss): the shared `autoindex-catalog` index is written by
+    /// every corpus, so its doc ids MUST be scoped by the corpus `--prefix` —
+    /// otherwise two corpora indexing the same file key (or producing the same
+    /// slug) overwrite each other's catalog entry. FAIL-BEFORE: dropping the
+    /// prefix from the id bodies makes these `assert_ne!` collide.
+    #[test]
+    fn catalog_ids_are_prefix_scoped_across_corpora() {
+        // Same key, different corpus prefix -> distinct ids (no overwrite).
+        assert_ne!(super::file_id("ax-a", "k"), super::file_id("ax-b", "k"));
+        assert_ne!(
+            super::duplicate_file_id("ax-a", "k", "p", "pi"),
+            super::duplicate_file_id("ax-b", "k", "p", "pi"),
+        );
+        // Idempotent within one corpus (a re-run must upsert, not duplicate).
+        assert_eq!(super::file_id("ax-a", "k"), super::file_id("ax-a", "k"));
+        // The prefix is actually in the id (not just concatenated key).
+        assert!(super::file_id("ax-a", "k").starts_with("file:ax-a:"));
     }
 }
--- a/engine/crates/xerj-autoindex/src/generation_catalog.rs
+++ b/engine/crates/xerj-autoindex/src/generation_catalog.rs
@@ -109,6 +109,7 @@ pub fn project_generation(
         let format = assignment_format(assignment.family.as_str(), assignment.gzip);
         code.observe(&format, group.expected_records);
         let (id, doc) = catalog::file_doc(
+            &execution.prefix,
             &group.content_id,
             &group.canonical.rel,
             &format,
@@ -122,6 +123,7 @@ pub fn project_generation(
         insert_unique(&mut documents, id, doc)?;
         for alias in &group.aliases {
             let (id, doc) = catalog::duplicate_file_doc(
+                &execution.prefix,
                 &group.content_id,
                 &alias.rel,
                 &alias.path_id,
@@ -138,6 +140,7 @@ pub fn project_generation(
         // these cannot double-count the groups above.
         code.observe(&junk.format, 0);
         let (id, doc) = catalog::file_doc(
+            &execution.prefix,
             &junk.file_key,
             &junk.rel,
             &junk.format,
@@ -176,6 +179,7 @@ pub fn project_generation(
         formats.sort();
         formats.dedup();
         let (id, doc) = catalog::dataset_doc(&catalog::DatasetDocInput {
+            prefix: &execution.prefix,
             pd: dataset,
             record_count: stats.record_count,
             junk_records: stats.junk_records,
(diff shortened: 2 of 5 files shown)


xerj-org:

All three returned **NOT REFUTED / 0 BLOCKING** against HEAD `7b21a2e9`.

### Lens: CORRECTNESS — refuted=False, blocking=NONE

**Evidence:** Fix present at catalog.rs:175 file_id(prefix,file_key)->'file:{prefix}:{file_key}', :147 dataset_doc->'ds:{prefix}:{slug}', :237 duplicate_file_id->'file-alias:{prefix}:{doc_id}'. Whole-crate grep for format!(\"file:/ds:/file-alias:) in non-test src: only prefixed call sites (generation_catalog.rs:306; catalog.rs:147/176/240); the lone bare 'file:' at extract/sqlite_x.rs:14 is a SQLite URI, not a catalog id. WRITE/REMOVAL AGREEMENT: (generation path) writes use execution.prefix at generation_catalog.rs:111/125/181; prior ids are observed back from the index by run_id via catalog_generation (sync_executor.rs:333-374, called :606-610), so stale-sweep prior ids are byte-identical to what was written -- no recompute. (legacy path, lib.rs) every write :5387/5432/5493/5509 and every delete/stale-id :5331/:3556/:3636 uses &cfg.prefix; and execution.prefix = cfg.prefix.clone() at lib.rs:545, so both modes share one corpus scope. TESTS: `cd engine && rustup run 1.97.1 (rustc 1.97.1) cargo test -p xerj-autoindex --profile ci-test` => 692 passed; 0 failed (+ all integration suites 8/1/1/4/0 passed), incl catalog_ids_are_prefix_scoped_across_corpora (fail-before), junk_transitions_and_removed_aliases_have_exact_stale_ids, junk_only_add_change_and_delete_converge_without_hiding_unchanged_data, stale_alias_and_independently_observed_correlation_are_deleted. THROWAWAY: added zzz_throwaway_dataset_and_file_ids_scope_by_prefix building two DatasetDocInput (prefix a vs b, same slug 'reports') plus file_id/duplicate_file_id a vs b -> assert_ne across prefixes, assert_eq within prefix (ds:a:reports idempotent); it PASSED, then reverted (git checkout -- catalog.rs, worktree clean).

**Non-blocking:** (1) Correlation doc ids (correlate.rs c.id()) and the run doc id (run:{generation_id}) are NOT --prefix-scoped. Out of scope for the #416 claim (which is file:/ds:/file-alias: only), and not a live risk: correlations are passed as an empty map into the production generation path (sync_executor.rs:620), and run_id is the globally-unique tx_id so run:{id} cannot collide across corpora. Worth a follow-up only if server-side correlations are ever enabled with shared dataset slugs across corpora.  (2) The generation path's stale-sweep does NOT recompute prior ids from a prefix at all (managed_non_run_ids is #cfg(test) only). Production reads back actual _ids by run_id (sync_executor.rs:606-610 / catalog_generation :333). This is strictly safer than recomputation but means the test helper's prefix agreement is a proxy, not the exact production mechanism; reviewers should not mistake the test helper for the live path.

**Notes:** HEAD 7b21a2e9 on origin/fix/416-catalog-prefix-scoping. Claim upheld: all three id families are corpus-prefix-scoped and both the generation and legacy write paths agree with their respective removal/stale-sweep prefixes (generation path via observed-readback, not recompute; legacy path via consistent &cfg.prefix). No surviving collision, no orphan-row mismatch found. refuted=false. Two out-of-scope non-blocking notes recorded (correlation/run ids unprefixed but not a live collision risk; test helper != production stale mechanism).

### Lens: REGRESSION — refuted=False, blocking=NONE

**Evidence:** HEAD=7b21a2e9 on origin/fix/416-catalog-prefix-scoping; fix confirmed present (catalog.rs:175 `pub fn file_id(prefix: &str, file_key: &str)`).

(1) TESTS — `rustup run 1.97.1 cargo test -p xerj-autoindex --profile ci-test` (rustc 1.97.1). All binaries green, 0 failures across every one:
  - lib+integration (unit) test result: ok. 691 passed; 0 failed; 0 ignored (finished 6.87s / 7.13s on re-run)
  - follow_symlinks_boundary.rs: ok. 8 passed; 0 failed
  - journal_lock_inheritance.rs: ok. 1 passed; 0 failed
  - scan_pool_width.rs: ok. 1 passed; 0 failed
  - sniff_printable_magic.rs: ok. 4 passed; 0 failed
  - Doc-tests: ok. 0 passed; 0 failed
  Total 705 tests, 0 failed. The concern that the id-format change could break catalog projection/reconcile/http tests that assert exact ids did NOT materialize: the reconcile http tests (incremental_reconcile_http_tests::*) and the generation_catalog_http_tests all passed because the http-test prior-id helper (prior_managed_ids) was updated to agree with the write path. The fail-before test catalog_ids_are_prefix_scoped_across_corpora is present in catalog.rs and passes within the 691.

(2) CLIPPY — `rustup run 1.97.1 cargo clippy -p xerj-autoindex --profile ci-test --all-targets -- -D warnings`: Finished ci-test profile in 17.63s; grep count of "warning:|error:" lines = 0. Passes.

(3) FMT — `rustup run 1.97.1 cargo fmt --all --check`: FMT_EXIT=0 (exact CI form used, not a masked pipe). Passes.

(4) SCOPE — `git show --name-only HEAD` lists exactly 5 files, all under engine/crates/xerj-autoindex/src/ (catalog.rs, generation_catalog.rs, generation_catalog_http_tests.rs, incremental_reconcile_http_tests.rs, lib.rs). No engine query path (xerj-engine/query/codec) is touched. Fix is autoindex-scoped as claimed. The diff only threads the corpus `--prefix` into the three catalog id builders (file_id, dataset_doc `ds:{prefix}:{slug}`, duplicate_file_id `file-alias:{prefix}:...`) and the matching write/sweep call sites; write path and stale-sweep prior-id computation agree.

**Notes:** No regression found — the fix is clean. All three gates (test / clippy / fmt) pass under the exact CI toolchain (1.97.1) and CI profile (ci-test). refuted=false means the REGRESSION hypothesis was not confirmed (no failing test to name, no lint/fmt breakage, no query-path spillover). No command exceeded ~5 min (tests ~7s, clippy ~18s, fmt instant). Ran clippy/fmt/test with `cd` into the worktree's engine dir per worktree isolation (original one-line compound commands were refused by the sandbox and split into plain commands).

### Lens: EVIDENCE — refuted=False, blocking=NONE

**Evidence:** FIX PRESENT: HEAD=7b21a2e9 on origin/fix/416-catalog-prefix-scoping; `pub fn file_id(prefix: &str, file_key: &str)` at catalog.rs:175 with body `format!("file:{prefix}:{file_key}")` (line 176). grep for the fix returned a match (no FIX-ABSENT).

FIX-AFTER (baseline, unmodified): `cargo test -p xerj-autoindex catalog_ids_are_prefix_scoped_across_corpora` (rustup 1.97.1) => `test catalog::sample_query_tests::catalog_ids_are_prefix_scoped_across_corpora ... ok` / `test result: ok. 1 passed; 0 failed`.

FAIL-BEFORE: reverted ONLY the prefix in the file_id body to `format!("file:{file_key}")` (kept `let _ = prefix;` so the signature/params are untouched), rebuilt (Compiling xerj-autoindex v1.0.0-rc.18, Finished in 8.23s), re-ran the same shipped test => FAILED:
  thread '...catalog_ids_are_prefix_scoped_across_corpora' panicked at crates/xerj-autoindex/src/catalog.rs:797:9:
  assertion `left != right` failed
    left: "file:k"
   right: "file:k"
This is exactly the predicted collapse: with the prefix dropped, file_id("ax-a","k") == file_id("ax-b","k") == "file:k", so the `assert_ne!` on line 796 (renumbered 797) fails. => The test genuinely pins the fix (NOT refuted).

RESTORE: `git checkout -- engine/crates/xerj-autoindex/src/catalog.rs`; grep re-confirms line 176 `format!("file:{prefix}:{file_key}")`; `git status --porcelain` clean; re-ran the shipped test => `ok. 1 passed; 0 failed`. Worktree returned to the pristine fixed state.

DATA-LOSS SCENARIO CONFIRMED (with the fix): the catalog is a single shared index (CATALOG_INDEX = "autoindex-catalog") written by every corpus. With the fix, the same file key under two corpus prefixes yields DISTINCT ids — file_id("ax-a","k")="file:ax-a:k" vs file_id("ax-b","k")="file:ax-b:k" — so no upsert overwrites the other corpus's entry, yet file_id("ax-a","k")==file_id("ax-a","k") stays idempotent for re-runs (asserted by lines 802/804). Without the prefix both collapse to "file:k" and the second corpus's write overwrites the first's catalog doc = the #416 data-loss. The sibling id builders are prefix-scoped the same way: dataset_doc id `ds:{prefix}:{slug}` (line 147) and duplicate_file_id `file-alias:{prefix}:...` (lines 237-244). Conformance suite never pointed at :9200 (pure unit test, no network).

**Non-blocking:** (1) The shipped test panics on its FIRST assertion (assert_ne! on file_id, catalog.rs:797:9 after my +1 line insertion; original line 796) and short-circuits, so with only file_id reverted it never reaches the duplicate_file_id / ds-id / starts_with assertions. That is fine for pinning file_id, but it means this single test does NOT independently pin duplicate_file_id() (line 237: `file-alias:{prefix}:...`) or dataset_doc() (line 147: `ds:{prefix}:{slug}`) — a regression in either of those alone would still be caught by lines 797-800/804 only if line 796 stays correct. Consider verifying those two ids are separately covered.  (2) Toolchain: ran under the CI toolchain rustc 1.97.1 (per the known local-vs-CI drift), not the local stable default.

**Notes:** Fail-before/fail-after fully demonstrated for the #416 prefix-scoping fix. The shipped test `catalog_ids_are_prefix_scoped_across_corpora` correctly fails when the fix is reverted and passes when present; it is a valid pin for file_id(). Only file_id was reverted per instructions; the other two id builders (dataset_doc, duplicate_file_id) are also prefix-scoped in source but are not independently exercised by this one test because the assert_ne! on file_id short-circuits first. Verdict: fix confirmed, not refuted.

**Disposition:** 0 BLOCKING → mergeable on all-green + CLEAN. Transparency correction (from the correctness lens): the *production* stale-sweep reads back actual `_id`s by `run_id` rather than recomputing them, so it was already collision-safe — the commit message overstates the orphan-row risk as a production concern; it applies to the `#[cfg(test)]` prior-id helpers. The fix (prefixing the written ids) is correct either way. Non-blocking follow-ups noted: correlation/run doc ids remain unprefixed (not a live collision risk — run_id is a globally-unique tx_id, correlations are empty in the production path); the repro test short-circuits on its first assert so `duplicate_file_id`/`dataset_doc` aren't independently pinned in the fail-before demo.

Issue #406, opened 2026-08-15
fuzzy is the last term-level query that folds case on a keyword field, and its case_insensitive parameter is ignored


xerj-org:
Found while fixing #396 ("prefix and wildcard disagree by default on a keyword field"). #396 is closed by making the standalone `wildcard` case-sensitive, which is ES/Lucene's rule and the one `prefix`, `term` and `regexp` already followed. `fuzzy` is now the only term-level query on a keyword field that still folds case, so the family is one query short of the single rule #362 asked for.

Written by an AI agent (Claude Opus 5) driven by @xerj-org.

## The disagreement

One `keyword` field, one value. Every term-level query is spelled in lowercase; `fuzzy` alone answers `1`.

```json
PUT /kw   {"mappings":{"properties":{"code":{"type":"keyword"}}}}
POST /kw/_doc/1?refresh=true   {"code":"HnswGraph"}
POST /kw/_flush
```

| query | hits |
|---|---|
| `{"term":{"code":"hnswgraph"}}` | 0 |
| `{"prefix":{"code":"hnsw"}}` | 0 |
| `{"wildcard":{"code":"hnsw*"}}` | 0 |
| `{"regexp":{"code":"hnsw.*"}}` | 0 |
| `{"fuzzy":{"code":{"value":"hnswgraph"}}}` | **1** |
| `{"fuzzy":{"code":{"value":"hnswgraph","case_insensitive":false}}}` | **1** |

ES answers `0` for all six. `fuzzy` has its own `case_insensitive` parameter (default `false`), and Lucene measures edit distance against the raw indexed term. The last row shows the parameter is not merely defaulted the other way — it is parsed away and never reaches the executor, so there is no request that turns the folding off.

## Cause

Two sites in `engine/crates/xerj-engine/src/index.rs` hardcode the fold for a keyword field rather than carrying it on the node, which is exactly the shape #396 removed from `wildcard`:

* `query_node_to_fts_with_keyword_fields`, the `QueryNode::Fuzzy` keyword arm, passes `case_insensitive: true` to `xerj_fts::search::FuzzyQuery` (the TEXT arm right below it correctly passes `false`).
* `ScoredPlan::KeywordFuzzy` on the columnar path.

`QueryNode::Fuzzy` has no `case_insensitive` field at all, so `{"fuzzy":{"code":{"value":"HNSWGRAPH","case_insensitive":false}}}` is accepted and ignored.

## What fixing it means

[...]

PR #675, opened 2026-08-23
fix(engine): #423/#406 keyword fuzzy is case-sensitive (ES default)

xerj-org:
### Symptom
`fuzzy:{name:{value:"hello",fuzziness:0}}` on keyword `"Hello"` matched (1). ES matches `fuzzy` at the term level against the RAW (un-analysed) value, so keyword fuzzy is **case-sensitive** by default (`case_insensitive` defaults to false) — it must NOT match. XERJ folded case on all three matcher paths.

### Fix — thread a `case_insensitive` flag (the #668 playbook)
- **AST** (`xerj-query/ast.rs`): add `case_insensitive: bool` (default false) to `QueryNode::Fuzzy`.
- **Parser**: standalone `{fuzzy}` / object form (parses the param) → false; `query_string` fuzzy lowering → true (value pre-lowercased); match/multi_match fuzzy leaf → false.
- **All three keyword-fuzzy matcher paths honor it consistently** (so nothing splits at flush):
  - memtable `doc_matches_query_typed` Fuzzy arm — keyword+`!case_insensitive`: RAW edit distance; else fold.
  - segment FST route (`index.rs:~41161`) — `FuzzyQuery.case_insensitive = node flag`.
  - segment doc-values `ScoredPlan::KeywordFuzzy` expansion — RAW-term edit distance when case-sensitive, else fold.
- `text` fields and unknown (schemaless shim) fields are unchanged.

### Verification (1.97.1)
- `fuzzy_keyword_case_sensitive`: keyword fuzzy is case-sensitive, `case_insensitive:true` still folds, and both are **flush-invariant** (memtable == segment).
- Fail-before: reverting the memtable arm restores the fold divergence.
- Full `xerj-engine` suite (ci-test + dev), fmt, clippy `-D warnings` — green.

Advances #423 (fuzzy limb; formerly #406).

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -22091,12 +22091,15 @@ impl Index {
         // ── KeywordFuzzy global expansion stats (LOSS_BATTLE_PLAN B4) ────
         // term → (folded Damerau dist, Σ df) over ALL segments'
         // dictionaries, plus the blended df (max Σdf across expansions —
-        // ES TopTermsBlendedFreqScoringRewrite).  The match predicate is
-        // the same folded whole-value-or-sub-token test as
-        // `xerj_fts::search::expand_fuzzy` / the doc-scan fuzzy arm.
-        // Sub-token-only matches (an intentional xerj leniency ES lacks)
-        // can carry a whole-value dist > max_edits; their sim clamps to
-        // ≥ 0 so the extra hits score deterministically.
+        // ES TopTermsBlendedFreqScoringRewrite).  The match predicate
+        // mirrors `xerj_fts::search::expand_fuzzy` / the doc-scan fuzzy
+        // arm EXACTLY: whole-value edit distance always, plus the
+        // sub-token fallback ONLY when folding (case_insensitive) — the
+        // case-sensitive keyword default is whole-term-only on every path
+        // (#423/#406). Sub-token-only matches (an intentional xerj
+        // leniency ES lacks, fold path only) can carry a whole-value
+        // dist > max_edits; their sim clamps to ≥ 0 so the extra hits
+        // score deterministically.
         let fuzzy_stats: Option<(HashMap<String, (usize, u64)>, u64)> = match plan {
             ScoredPlan::KeywordFuzzy {
                 field,
@@ -22116,6 +22119,19 @@ impl Index {
                 let term_matches = |cmp: &str| -> bool {
                     if levenshtein_distance(cmp, &q_ref) <= *max_edits {
                         return true;
+                    }
+                    // #423/#406: the sub-token fallback is a FOLD-path leniency
+                    // ONLY. When case-sensitive (the ES keyword default), the
+                    // memtable Fuzzy arm (`token_match`) and the FST
+                    // `expand_fuzzy` both match the whole un-analysed term and
+                    // nothing else — so this DV path must too, or a
+                    // separator-containing keyword ("alpha-beta") answers a
+                    // sub-token query ("beta") after flush but not before
+                    // (flush-invariance; caught by the PR #675 correctness
+                    // skeptic). The post-flush sub-token hit was also non-ES:
+                    // keyword fuzzy compares the whole term.
+                    if !ci {
+                        return false;
                     }
                     cmp.split(|c: char| !c.is_alphanumeric())
                         .filter(|t| !t.is_empty())
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -33826,10 +33826,12 @@ fn strip_nested_path_in_query(q: &QueryNode, path: &str) -> QueryNode {
             field,
             value,
             fuzziness,
+            case_insensitive,
         } => QueryNode::Fuzzy {
             field: strip(field),
             value: value.clone(),
             fuzziness: *fuzziness,
+            case_insensitive: *case_insensitive,
         },
         QueryNode::Regexp { field, pattern } => QueryNode::Regexp {
             field: strip(field),
@@ -36402,6 +36404,7 @@ fn doc_matches_query_typed(q: &QueryNode, source: &Value, schema: &Schema) -> bo
             field,
             value: query_value,
             fuzziness,
+            case_insensitive: _,
         } => {
             let max_edits = match fuzziness {
                 Fuzziness::Auto => auto_fuzziness(query_value),
@@ -41121,6 +41124,7 @@ fn query_node_to_fts_with_keyword_fields(
             field,
             value,
             fuzziness,
+            case_insensitive: _,
         } => {
             // `fuzzy` on a KEYWORD field: expand the keyword FST term dictionary
             // to every term within `max_edits` Damerau-Levenshtein distance and
@@ -42460,6 +42464,7 @@ fn scored_fast_plan(
             field,
             value,
             fuzziness,
+            case_insensitive: _,
         } if fs.kw.contains(field) => {
             let max_edits = match fuzziness {
                 Fuzziness::Auto => auto_fuzziness(value),
@@ -44379,6 +44384,7 @@ mod fts_projection_tests {
             field: "body".into(),
             value: "runbok".into(),
             fuzziness: Fuzziness::Fixed(1),
+            case_insensitive: false,
         };
         match query_node_to_fts(&q, &tf, &kw(&[])).expect("text fuzzy projects") {
             FtsQuery::Fuzzy(f) => {
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -22102,17 +22102,24 @@ impl Index {
                 field,
                 value,
                 max_edits,
+                case_insensitive,
                 ..
             } => {
-                let q_folded = value.to_lowercase();
-                let term_matches = |folded: &str| -> bool {
-                    if levenshtein_distance(folded, &q_folded) <= *max_edits {
+                // #423/#406: case-sensitive keyword fuzzy (the ES default)
+                // compares the RAW term; case_insensitive folds both sides.
+                let ci = *case_insensitive;
+                let q_ref = if ci {
+                    value.to_lowercase()
+                } else {
+                    value.clone()
+                };
+                let term_matches = |cmp: &str| -> bool {
+                    if levenshtein_distance(cmp, &q_ref) <= *max_edits {
                         return true;
                     }
-                    folded
-                        .split(|c: char| !c.is_alphanumeric())
+                    cmp.split(|c: char| !c.is_alphanumeric())
                         .filter(|t| !t.is_empty())
-                        .any(|tok| levenshtein_distance(tok, &q_folded) <= *max_edits)
+                        .any(|tok| levenshtein_distance(tok, &q_ref) <= *max_edits)
                 };
                 let mut map: HashMap<String, (usize, u64)> = HashMap::new();
                 for (meta, cols) in segments.iter().zip(seg_cols.iter()) {
@@ -22126,10 +22133,14 @@ impl Index {
                         return None;
                     }
                     for (ord, term) in k.terms.iter().enumerate() {
-                        let folded = term.to_lowercase();
-                        if term_matches(&folded) {
+                        let cmp = if ci {
+                            term.to_lowercase()
+                        } else {
+                            term.clone()
+                        };
+                        if term_matches(&cmp) {
                             let seg_df = k.per_ord_count.get(ord).copied().unwrap_or(0) as u64;
-                            let dist = levenshtein_distance(&folded, &q_folded);
+                            let dist = levenshtein_distance(&cmp, &q_ref);
                             let e = map.entry(term.clone()).or_insert((dist, 0));
                             e.1 += seg_df;
                         }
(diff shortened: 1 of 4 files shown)


xerj-org:
## Independent 3-skeptic verification (correctness / regression / evidence), rustc 1.97.1, `--profile ci-test`, isolated worktrees

Posting all three verdicts verbatim. **One BLOCKING refutation (correctness)** — a real, self-inflicted flush-divergence — was found and has now been fixed (see "Fix applied" below); the fix is not yet merged.

### CORRECTNESS — REFUTED (BLOCKING)

> **FLUSH-INVARIANCE BROKEN for case-sensitive keyword fuzzy on separator-containing values.** keyword field name="alpha-beta", query fuzzy{value:"beta", fuzziness:0} (case-sensitive, no ci flag): pre-refresh (memtable) = 0 hits, post-refresh (segment) = 1 hit. At size:10 the doc is actually materialized: pre=(total 0, hits 0), post=(total 1, hits 1). memtable != segment => refutes the claim.
>
> Root cause: the three matcher paths do NOT agree on sub-token matching in the case-SENSITIVE path. Memtable doc_matches_query_typed Fuzzy arm (index.rs:36443-36444) returns EARLY with whole-value edit distance ONLY when keyword_case_sensitive. Segment DV ScoredPlan::KeywordFuzzy expansion (index.rs:22116-22123) runs the cmp.split(non-alnum) sub-token fallback UNCONDITIONALLY (regardless of ci). The FST route expand_fuzzy (xerj-fts/src/search.rs:994-1004) is whole-only case-sensitive and agrees with memtable, so it is specifically the DV path that diverges. The DV comment (index.rs:22097-22099) even calls this an 'intentional xerj leniency ES lacks' -- applied only on the segment side.
>
> Post-refresh result (1) is also non-ES: ES keyword fuzzy matches the whole un-analysed term, so "beta" must NOT match "alpha-beta". The DV sub-token leniency is both flush-divergent AND wrong vs the ES semantics the fix claims to adopt.

Evidence: ran the shipped test (1 passed — the shipped test only used single-token "Hello", which has no separator so cannot surface the asymmetry), then two throwaway integration tests: `SEPARATOR PROBE name=alpha-beta q=beta f0 case-sensitive pre=0 post=1`; `SIZE10 SEPARATOR pre(total,hits)=(0,0) post(total,hits)=(1,1)`. All other required cases were flush-invariant and matched. The case_insensitive/query_string FOLD path still folds (pre_ci==post_ci==1 scalar and array). Fix direction: gate the DV `cmp.split(...)` branch behind `ci` so all three paths agree.

> Diff touches only 4 files (index.rs, ast.rs, parser.rs, and the NEW test) — zero existing test files were edited, so the ~1000 pre-existing engine/query tests pass unmodified against the new case-sensitive-by-default behaviour. No test was altered to hide a fold-behaviour regression.
> The pre-existing fuzzy regression test test_fuzzy_query_typo passes unchanged.

Evidence: `cargo test -p xerj-engine` EXIT=0, all 57 "test result:" lines "0 failed" (integration 626, es_compat 126, new test 1, wildcard_keyword_case_sensitive 1, prefix_keyword_case_flush_parity, test_fuzzy_query_typo ok). `cargo test -p xerj-query` EXIT=0 (192 passed). clippy `-D warnings` EXIT=0. fmt --check EXIT=0.

### EVIDENCE (fail-before holds) — not refuted (0 blocking)

> (1) BASELINE full fix: `cargo test -p xerj-engine --test fuzzy_keyword_case_sensitive` -> ok, 1 passed. (2) REVERTED memtable arm only (index.rs:36443-36445 keyword branch) -> test FAILED at line 58 flush-invariance: left:1 (memtable folds) vs right:0 (segment case-sensitive). (3) RESTORED via git checkout -> ok, 1 passed. Never touched :9200.

Side note (not part of the refutation): the prompt's expected `"hxllo" f1 => 1` was a mis-stated expectation (levenshtein("hxllo","Hello")=2, so it does not match at f1); the shipped test does not encode it, so this is inert.

---

## Fix applied — commit `8fb14925`

Gated the DV `ScoredPlan::KeywordFuzzy` sub-token fallback behind `ci`, so the three paths now agree in **both** modes:

| mode | memtable `token_match` | FST `expand_fuzzy` | DV `KeywordFuzzy` (now) |
|---|---|---|---|
| case-sensitive (ES keyword default) | whole-term only | whole-term only (raw) | **whole-term only** |
| `case_insensitive:true` (fold) | whole-or-sub-token (lc) | whole-or-sub-token (lc, `term_matches_fuzzy`) | whole-or-sub-token (lc) |

- Added a **separator guard** to `fuzzy_keyword_case_sensitive.rs`: `name="alpha-beta"`, `fuzzy{value:"beta",fuzziness:0}` case-sensitive must be **0 pre and post** (flush-invariant, whole-term-only); `case_insensitive:true` still **1** (fold keeps the sub-token leniency, flush-invariant).
- **Fail-before re-proven** on the new hunk: with `if !ci { return false; }` removed, the separator test fails on flush-invariance `left: 0` (memtable) ≠ `right: 1` (segment DV); restored via `git checkout` → passes.
- fmt clean; full `xerj-engine` + `xerj-query` suites re-running under 1.97.1 `--profile ci-test`.

Re-verification skeptics will be re-run on the updated branch before any merge.

xerj-org:
## Re-verification pass on the updated fix (commit `8fb14925`) — 3 skeptics, rustc 1.97.1, `--profile ci-test`, isolated worktrees

The prior BLOCKING refutation (DV sub-token fallback) is fixed; this pass re-refutes the updated branch. **All three: 0 BLOCKING.** Verdicts verbatim:

> Base verified: git rev-parse HEAD = 8fb149251e7b5672e06037585c83187a2800635c; DV guard present at index.rs:22133 (`if !ci { return false; }` before the cmp.split sub-token fallback). Built + ran with RUST_MIN_STACK=8388608, rustup run 1.97.1 cargo test --profile ci-test, pure Rust only (never touched :9200).
> 
> Wrote two throwaway integration test files (deleted after; `git status --porcelain` now empty, HEAD unchanged) probing memtable(pre-refresh) vs segment(post-refresh) at size:0 AND size:10 as (pre0,post0,pre10,post10):
> 1) EXACT prior refutation — keyword "alpha-beta", fuzzy{value:"beta",fuzziness:0} case-sensitive => (0,0,0,0). PASS.
> 2) same with case_insensitive:true => (1,1,1,1); uppercase "BETA" ci => (1,1,1,1). PASS (fold keeps sub-token).
> 3) fuzzy{value:"alpha-beta"} cs => (1,1,1,1). PASS.
> 4) fuzziness:1 whole "alpha-bet" => (1,1,1,1); case-correct sub-token "bet" fuzziness:1 cs => (0,0,0,0) but ci => (1,1,1,1); multi-separator "a-b-c" sub "b" cs => (0,0,0,0), whole => (1,1,1,1), ci sub => (1,1,1,1); array ["gamma-delta","alpha-beta"] sub "beta" cs => (0,0,0,0), whole => (1,1,1,1), ci => (1,1,1,1). ALL PASS.
> 5) mixed memtable+segment (one flushed, one buffered) cs sub "beta" => 0 at sz0/sz10, ci => 2; nullable-column brute fallback cs sub => 0, whole => 1, ci => 1; multi-segment cs sub => 0, ci => 2. ALL PASS.
> Text field "Alpha Beta", fuzzy "beta" => (1,1,1,1) — still tokenizes+folds. PASS.
> 
> FAIL-BEFORE: removed the `if !ci { return false; }` guard from the DV path and re-ran — the three case-sensitive sub-token tests failed with (pre0,post0,pre10,post10)=(0,1,0,1), i.e. memtable=0 / segment=1, the exact flush divergence the PR fixes and non-ES. Restored via git checkout; guard confirmed back at index.rs:22133. This proves the DV path is genuinely exercised at size:10 post-flush and the guard is load-bearing.
> 
> No memtable!=segment divergence and no non-ES keyword-fuzzy hit survived. Could not break the result.

Non-blocking notes:
- The memtable Fuzzy arm decides case-sensitivity via `declared_field(schema, field) == Keyword && !case_insensitive`, whereas the DV path keys off the ScoredPlan's `case_insensitive` flag. My tests confirm they agree in practice for declared keyword fields (scalar + array, mixed/nullable/multi-segment), but the two paths derive the mode from different sources; a future field-type/plan-construction change could desync them. Not a defect in this PR.

### EVIDENCE-HOLDS (fail-before is load-bearing) — not refuted (0 blocking)

> Base verified: HEAD=8fb149251e7b5672e06037585c83187a2800635c on fix/423-fuzzy-case-sensitive; DV guard `if !ci {` present at index.rs:22133 (returns false before the cmp.split sub-token fallback inside ScoredPlan::KeywordFuzzy term_matches closure). All cargo via `RUST_MIN_STACK=8388608; cd engine; rustup run 1.97.1 cargo ... --profile ci-test`. No :9200 touched — pure Rust test only.
> 
> State (1) BASELINE (8fb14925 intact): `cargo test -p xerj-engine --test fuzzy_keyword_case_sensitive` -> both tests pass (fuzzy_keyword_is_case_sensitive_but_ci_flag_still_folds + fuzzy_keyword_separator_matches_whole_term_only_when_case_sensitive), 2 passed 0 failed.
> 
> State (2) REVERT ONLY the DV guard (deleted the three lines `if !ci {` / `return false;` / `}` immediately above `cmp.split(|c: char| !c.is_alphanumeric())`): rebuild + run -> fuzzy_keyword_separator_matches_whole_term_only_when_case_sensitive FAILED at line 122 on flush-invariance assertion `pre_sub == post_sub`, left(pre/memtable)=0, right(post/segment)=1. This is exactly the ORIGINAL symptom direction: case-sensitive sub-token query "beta" against keyword "alpha-beta" does not hit the memtable (0) but DOES hit the DV segment after flush (1) via the unconditional sub-token fallback. The ci-flag test still passed, confirming the guard change is what the separator test isolates. Fail-before is load-bearing: the test genuinely guards the fix.
> 
> State (3) RESTORE via `git checkout -- engine/crates/xerj-engine/src/index.rs`: guard back at 22133; `git status --short` empty (clean); re-run -> both tests pass, 2 passed 0 failed.
> 
> Could not break the fix. The three matcher paths agree: case-sensitive keyword fuzzy matches whole un-analysed term only (flush-invariant 0/0 for sub-token, 1/1 for whole-term), case_insensitive:true retains sub-token leniency (1/1). refuted=false.

> Base verified: HEAD=8fb149251e7b5672... , DV guard present at index.rs:22133 (`if !ci { return false; }`). git diff 61a597bf..8fb14925 --stat touches ONLY engine/crates/xerj-engine/src/index.rs (+28/-6, a single hunk = the DV KeywordFuzzy sub-token guard + comment) and a NEW file tests/fuzzy_keyword_case_sensitive.rs (+64). No existing test file edited; test_fuzzy_query_typo lives untouched in tests/integration.rs. Three-path agreement confirmed by inspection: (a) memtable arm index.rs:36458 `token_match` — keyword_case_sensitive => raw whole-value levenshtein only, else lowercase whole-then-subtoken; (b) FST expand_fuzzy xerj-fts/src/search.rs:970 — ci=>term_matches_fuzzy(fold, whole+subtoken), else raw whole-term levenshtein; caller index.rs:41191 threads `case_insensitive: *case_insensitive`; (c) DV index.rs:22119 — q_ref/cmp folded iff ci, whole-value levenshtein always, sub-token fallback gated `if !ci { return false; }`. All three: case-sensitive = whole-term-only; ci = whole-OR-subtoken folded. RESULTS: `cargo fmt --all --check` FMT_EXIT=0. `cargo test -p xerj-engine --profile ci-test` (rustup 1.97.1, RUST_MIN_STACK=8388608): lib unittests 626 passed / 0 failed; 48 `test result: ok` summaries; 0 FAILED across the run at report time. The three load-bearing tests all PASS: fuzzy_keyword_is_case_sensitive_but_ci_flag_still_folds ok, fuzzy_keyword_separator_matches_whole_term_only_when_case_sensitive ok (the PR#675 skeptic's alpha-beta/beta flush-divergence scenario — now 0 pre AND post, ci-true 1 pre AND post), test_fuzzy_query_typo ok (pre-existing, unchanged). Engine run had a few later-alphabet integration binaries still executing when a forced report was required, but 0 failures observed and every fuzzy/keyword test already passed. Could not construct a refutation.

Non-blocking notes:
- xerj-query suite and `cargo clippy -p xerj-query -p xerj-engine --all-targets -D warnings` did not finish before a forced early report. Low risk: the fix commit touches ONLY engine/crates/xerj-engine/src/index.rs (the DV KeywordFuzzy `if !ci { return false; }` guard + its comment) and adds a new test file — it does not touch xerj-query at all, and the guard is a trivial early-return clippy cannot flag. Recommend confirming both are green before merge for full lens coverage.
- The new correctness test only exercises fuzziness:0. The whole-value-vs-sub-token divergence at higher edit distances (e.g. fuzziness:1, uppercase folded query) is covered by code inspection (all three paths use identical whole-then-sub-token structure with the same `!is_alphanumeric()` split and lowercase folding) but not by an executed assertion.

---

**Maintainer note on the REGRESSION lens's incomplete run:** its worktree sandbox hit friction (`. env` sourcing / `tee` refused) and was force-reported before `xerj-query` + clippy finished. Those two gates were run independently by me on this exact HEAD (`8fb14925`): `cargo clippy -p xerj-query -p xerj-engine --profile ci-test --all-targets -- -D warnings` EXIT=0, `cargo test -p xerj-query --profile ci-test` EXIT=0 (0 failures), `cargo test -p xerj-engine` EXIT=0, `cargo fmt --all --check` EXIT=0. Full lens coverage holds.

**Merge gate (b) ZERO BLOCKING: met.** Remaining gates before merge: (a) all-green/none-pending — currently the tag-ahead *Landing constants guard* is red and clears on rebase onto post-rc.57 main once #674 merges; (c) mergeStateStatus CLEAN. Will rebase, let CI go green, then merge.

Issue #413, opened 2026-08-15
an object-valued field is indexed as its serialised JSON: prefix/wildcard on it changes its answer at flush


xerj-org:
Split out of #398, same family as #408 — the write path invents a term the stored-document scan cannot reproduce.

Written by an AI agent (Claude Opus 5) driven by @xerj-org.

## The gap

`extract_field_text` (`engine/crates/xerj-engine/src/index.rs`) turns a JSON **object** into a term by serialising it: `Value::Object(_) => serde_json::to_string(val)`. So a field mapped `object` (or any mapped field that happens to hold an object) gets a term dictionary entry that is the field's raw JSON text. `doc_matches_query`'s `prefix` / `wildcard` arms decline object values, so the same request answers differently before and after a flush.

With `meta` mapped `object` holding `{"owner":"Alpha"}`:

| query | before flush | after flush |
|---|---|---|
| `{"prefix":{"meta":"{"}}` | **0** | **1** |
| `{"wildcard":{"meta":"*Alpha*"}}` | **0** | **1** |

## Why it is not fixed in #398

#398 made the scan reproduce the segment's spelling for strings, numbers and booleans, because for those there is a term the user plausibly meant. An object is different: real ES rejects `prefix` on an `object` field outright, so there is no correct answer for the two sides to converge on — only two arbitrary ones. Teaching the scan to match the serialised JSON would bake "your object's JSON punctuation is searchable" in as intended semantics rather than fixing it.

The defect is the serialisation, and it lives in the **write path** — the same place as #408's array join, in the same function.

**Verified:** both rows measured on branch `fix/issue-398` with the `Engine`/`Index` API and an own temp data dir; two indices holding a byte-identical document, one flushed and one not.
**Assumed:** that the right fix is to stop indexing serialised objects as terms rather than to mirror the serialisation in the scan — that is a design opinion, not a measurement.

xerj-org:
Consolidated into #423.

Not closing as fixed — your repro is still valid and is now acceptance criteria on #423. This is one defect: `doc_matches_query` (`crates/xerj-engine/src/index.rs:33213`) matches buffered documents against raw `_source` with no mapping and no analyser, while flushed documents go through the real term dictionary. Two matchers, one query, and a background flush decides which one you get.

Eight separate patches to that one schema-less evaluator is what produced the churn here (#410's fix for #394 reintroduced it through a sibling path; #388's own verification flagged a new divergence of the same class). Tracking the cause instead.

[...]

PR #676, opened 2026-08-23
fix(engine): #413 object-field prefix/wildcard is flush-invariant (mirror the whole-object blob)

xerj-org:
## #413 — object-valued field prefix/wildcard changes its answer at flush

Closes the #413 acceptance criterion under #423.

### The bug
A root-level object-mapped field is indexed on flush as **one whole, untokenised, case-preserved keyword-style blob** — the pruned `serde_json::to_string` emitted by `collect_text_fields` (memtable.rs:3887-3892). The memtable `doc_matches_query` Prefix/Wildcard arms declined `Value::Object` (`_ => None`), so the same query answered **0 before a flush and 1 after**:

| query (field `meta` = `{"owner":"Alpha"}`) | before flush | after flush |
|---|---|---|
| `{"prefix":{"meta":"{"}}` | 0 | 1 |
| `{"wildcard":{"meta":"*Alpha*"}}` | 0 | 1 |

### The fix
Per #423's own option-1 ("make the memtable path consult the same term structures the segment does"), the two memtable arms now **mirror the blob** — matching `memtable::extract_text_value_excluding(v, field, &fts_excluded_fields(schema))` (the write path's exact pruned serialization, so dense_vector exclusion stays identical) as a **raw, case-sensitive** prefix/wildcard, **root-level only** (`!field.contains('.')`, matching `collect_text_fields`'s `prefix.is_empty()` blob condition), and **gated on a declared `object` mapping** so the `EMPTY_SCHEMA` schemaless-shim callers are byte-for-byte unchanged.

Probed segment storage shape (flushed): the blob is one whole term — `prefix {`→1, `prefix owner`→0 (not tokenised), `wildcard *Alpha*`→1, `wildcard *alpha*`→0 (case-preserved). The fix reproduces exactly this.

- New test `object_field_blob_flush_parity.rs` — 7 cases, each asserted **flush-invariant** (pre==post) **and** value-correct.
- **Fail-before:** reverting only the two arms → `{"prefix":{"meta":"{"}}` fails `pre=0 != post=1` (the exact divergence); restored → passes.
- Full `xerj-engine` suite: 0 failures. Full `xerj-query` suite: 0 failures. `clippy --all-targets -D warnings`: clean. `fmt --all --check`: clean.

### Notes
- **Tag-ahead:** branched off pre-rc.57 main, so "Landing constants guard" is red until this is rebased onto post-rc.57 main (after the rc.57 stamp PR merges) — landing-only, no code impact.
- **Separate follow-up (out of scope):** `term` on an object **sub-field** (`{"term":{"meta.owner":"Alpha"}}`) is independently flush-divergent (memtable 0 / segment 1). Not touched here (Term arm unchanged); will file a follow-up issue.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -35728,14 +35728,40 @@ fn doc_matches_query_typed(q: &QueryNode, source: &Value, schema: &Schema) -> bo
                             .any(|tok| !tok.is_empty() && tok.starts_with(&value_lc))
                 }
             };
+            let is_object = matches!(ft, Some(FieldType::Object));
             get_field_value(source, field)
-                .and_then(|v| match v {
-                    Value::String(s) => Some(matches_str(&s)),
-                    Value::Array(arr) => Some(arr.iter().any(|e| match e {
-                        Value::String(s) => matches_str(s),
-                        _ => false,
-                    })),
-                    _ => None,
+                .and_then(|v| {
+                    // #413: a root-level object-mapped field is indexed on flush
+                    // as ONE whole, untokenised, case-preserved keyword-style
+                    // blob — the pruned `serde_json::to_string` emitted by
+                    // `collect_text_fields`. The memtable used to decline objects
+                    // (`_ => None`), so `prefix` answered 0 before a flush and 1
+                    // after. Mirror the blob here — raw, case-SENSITIVE, and
+                    // root-level only (`collect_text_fields` blobs only when its
+                    // incoming prefix is empty) — so the answer is flush-
+                    // invariant. Gated on a declared `object` mapping so the
+                    // `EMPTY_SCHEMA` shim callers stay byte-for-byte unchanged.
+                    if is_object && !field.contains('.') {
+                        // Object OR array: `collect_text_fields` emits a blob
+                        // for a root-level object AND for an array (joining its
+                        // elements via the same `extract_text_value`), so mirror
+                        // both here. Keyword/text arrays never reach this branch
+                        // (`is_object` is false for them).
+                        if matches!(&v, Value::Object(_) | Value::Array(_)) {
+                            let excluded = crate::memtable::fts_excluded_fields(schema);
+                            let blob =
+                                crate::memtable::extract_text_value_excluding(&v, field, &excluded);
+                            return Some(blob.starts_with(value.as_str()));
+                        }
+                    }
+                    match v {
+                        Value::String(s) => Some(matches_str(&s)),
+                        Value::Array(arr) => Some(arr.iter().any(|e| match e {
+                            Value::String(s) => matches_str(s),
+                            _ => false,
+                        })),
+                        _ => None,
+                    }
                 })
                 .unwrap_or(false)
         }
@@ -35785,14 +35811,37 @@ fn doc_matches_query_typed(q: &QueryNode, source: &Value, schema: &Schema) -> bo
                             .any(|tok| !tok.is_empty() && wildcard_match(tok, &pat_lc))
                 }
             };
+            let is_object = matches!(ft, Some(FieldType::Object));
             get_field_value(source, field)
-                .and_then(|v| match v {
-                    Value::String(s) => Some(matches_str(&s)),
-                    Value::Array(arr) => Some(arr.iter().any(|e| match e {
-                        Value::String(s) => matches_str(s),
-                        _ => false,
-                    })),
-                    _ => None,
+                .and_then(|v| {
+                    // #413: mirror the whole-object blob (see the Prefix arm).
+                    // The segment stores it case-preserved; honour
+                    // `case_insensitive` exactly as the keyword branch does.
+                    if is_object && !field.contains('.') {
+                        // Object OR array: `collect_text_fields` emits a blob
+                        // for a root-level object AND for an array (joining its
+                        // elements via the same `extract_text_value`), so mirror
+                        // both here. Keyword/text arrays never reach this branch
+                        // (`is_object` is false for them).
+                        if matches!(&v, Value::Object(_) | Value::Array(_)) {
+                            let excluded = crate::memtable::fts_excluded_fields(schema);
+                            let blob =
+                                crate::memtable::extract_text_value_excluding(&v, field, &excluded);
+                            return Some(if *case_insensitive {
+                                wildcard_match(&blob.to_lowercase(), &pat_lc)
+                            } else {
+                                wildcard_match(&blob, pattern.as_str())
+                            });
+                        }
+                    }
+                    match v {
+                        Value::String(s) => Some(matches_str(&s)),
+                        Value::Array(arr) => Some(arr.iter().any(|e| match e {
+                            Value::String(s) => matches_str(s),
+                            _ => false,
+                        })),
+                        _ => None,
+                    }
                 })
                 .unwrap_or(false)
         }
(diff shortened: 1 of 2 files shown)


xerj-org:

Run on the fix commit (pre-rebase `4044886e`; the rebase onto rc.57 main is landing-only, so these verdicts hold). **All three: 0 BLOCKING.** Verbatim:

> Base verified: git reset to origin/fix/413-object-blob-flush-parity -> HEAD 4044886e; grep found the fix at index.rs:35744 (Prefix) and :35820 (Wildcard). Wrote a throwaway integration test (tests/zzz_adversarial_413.rs, since deleted; tree clean) built+run with RUST_MIN_STACK=8388608, rustup run 1.97.1 cargo test --profile ci-test. It probes memtable(pre-refresh) vs segment(post-refresh, refresh() drains the memtable to a segment) at size:0 AND size:10 (asserting total.value == returned hit count). All in-scope cases pass flush-invariant and match expected: (1) scalar object meta={owner:Alpha}: prefix '{' =1/1, prefix 'owner'=0/0, wildcard *Alpha*=1/1, *alpha*=0/0 (case-preserved); (2) array-of-objects tags=[{owner:Alpha},{owner:Beta}]: prefix '{'=1/1, *Alpha*=1/1, *Beta*=1/1, *alpha*=0/0; (3) case_insensitive:true wildcard over object folds and stays invariant; (4a) nested object {a:{b:Xray}}: prefix '{'=1/1, *Xray*=1/1, *xray*=0/0; (4b) object-mapped field with a bare STRING value: flush-invariant; (5) NON-REGRESSION keyword scalar (prefix 'Key'=1, 'key'=0; wildcard *One*=1, *one*=0 case-sensitive), dynamic keyword array, text field (prefix 'hel'=1, wildcard *orl*=1), and undeclared/shim field: all flush-invariant and unchanged. The only divergence surfaced was term{meta.owner} (#677). Diff review confirms the change is purely additive, gated on is_object && !field.contains('.') && matches Object|Array; the original `match v` fall-through (String/Array/_ => None) is preserved verbatim, so non-object (is_object=false) and shim (ft=None) callers are byte-for-byte unchanged. Memtable mirror uses crate::memtable::extract_text_value_excluding with fts_excluded_fields(schema) — the same function and excluded set as the write path's collect_text_fields — so the blob (object: pruned serde_json::to_string; array: space-joined elements) matches the segment exactly. Cannot break the result.

Non-blocking:
- term on an object SUB-field (e.g. {"term":{"meta.owner":"Alpha"}}) is still flush-divergent: memtable(pre-refresh)=0, segment(post-refresh)=1. This is the known #677, explicitly out of scope. Verified NOT introduced by this fix: the git diff of 4044886e touches only the Prefix and Wildcard arms; the Term arm is unchanged and dotted fields skip the new blob arm (gated on !field.contains('.')). Not blocking.

> Base HEAD = 4044886e9eab7f14c8534c8ca26e8235c28cb68a (correct). Fix present in BOTH memtable arms: `if is_object && !field.contains('.')` at index.rs:35744 (Prefix) and :35820 (Wildcard), each gated on `let is_object = matches!(ft, Some(FieldType::Object))` (:35731,:35814) — keyword/text/unknown fields never enter the branch, and EMPTY_SCHEMA shim callers (ft==None) are byte-for-byte unchanged. git diff origin/main..HEAD touches ONLY 2 files: engine/crates/xerj-engine/src/index.rs (the two arms) and the NEW test engine/crates/xerj-engine/tests/object_field_blob_flush_parity.rs; NO existing test edited. Verified the memtable blob (crate::memtable::extract_text_value_excluding) is identical to the flush-path blob emitted by collect_text_fields (memtable.rs:3849): root-level object -> pruned serde_json::to_string; array-of-objects -> join(\" \") of per-element extract_text_value; the excluding variant short-circuits to extract_text_value for a top-level Array, so both paths agree. The new test genuinely flushes via idx.refresh() and asserts pre==post AND post==expected across object + array-of-objects, prefix + wildcard, case-sensitive (e.g. prefix \"owner\"=0 proving untokenised, wildcard *alpha*=0 proving case-preserved) — it passed. SUITES (rustup 1.97.1, --profile ci-test, RUST_MIN_STACK=8388608): xerj-engine tests EXIT 0, 57 `test result: ok` lines, zero non-\"0 failed\" lines, object_field_prefix_wildcard_is_flush_invariant ok; xerj-query tests EXIT 0, zero failures; clippy -p xerj-query -p xerj-engine --all-targets -D warnings EXIT 0; cargo fmt --all --check EXIT 0. Could not break it: fix strictly converts a pre-existing pre-vs-post divergence (object fields returned 0 pre-flush, matched blob post-flush) into flush-invariance; no non-object field or shim path is reachable by the new branch.

Non-blocking:
- Prefix object-blob branch (index.rs:35754) matches case-SENSITIVELY unconditionally, ignoring any node-level case_insensitive flag (Prefix binds `..`). This mirrors the ES keyword-prefix default and the segment's case-preserved blob, and the flush-parity test confirms parity for the covered cases; a prefix{case_insensitive:true} on an object field is an untested edge but is a pre-existing gap, not a divergence introduced by this fix. Out of scope of the merge gate.
- Term arm on object SUB-fields remains flush-divergent — acknowledged as out-of-scope issue #677; deliberately untouched here.

### EVIDENCE-HOLDS (fail-before is load-bearing): ran the parity test in three states (baseline, source-only revert, restore) plus a probe to isolate the array-of-objects divergence direction. — not refuted (0 blocking)

> Base HEAD=4044886e confirmed; fix present as 2 hits (`is_object && !field.contains('.')`) in the Prefix and Wildcard arms of engine/crates/xerj-engine/src/index.rs (lines 35744, 35820). STATE 1 baseline: `cargo test -p xerj-engine --profile ci-test --test object_field_blob_flush_parity` -> 1 passed. STATE 2 revert source only (`git checkout origin/main -- .../src/index.rs`, test file byte-identical, git status shows only index.rs modified): test FAILS -> panic at tests/object_field_blob_flush_parity.rs:73 `#413: {\"prefix\":{\"meta\":\"{\"}} must be flush-invariant (pre=0, post=1) left:0 right:1`. A scratch probe (removed after) under the same reverted source printed ALL cases: meta prefix `{` pre=0 post=1; meta wildcard *Alpha* pre=0 post=1; tags prefix `{` pre=0 post=1; tags wildcard *Alpha* pre=0 post=1; tags wildcard *Beta* pre=0 post=1 — confirming BOTH the scalar-object AND the array-of-objects divergence in the stated direction (pre=0, post=1). STATE 3 restore (`git checkout HEAD -- .../src/index.rs`, fix back to 2 hits): same probe shows every case pre==post==1 (flush-invariant); real test -> 1 passed; `git status` clean; HEAD still 4044886e. Blob-parity code review: write path collect_text_fields object arm calls extract_text_value_excluding (same fn the memtable fix calls); its array arm `map(extract_text_value).join(\" \")` equals extract_text_value_excluding's array short-circuit (array is not Object -> returns extract_text_value), so memtable and segment blobs are byte-identical for both shapes; serde_json serialization is the same call on both paths. Regression surface: the new branch is gated on declared_field(schema,field)==Some(FieldType::Object), so EMPTY_SCHEMA shim callers (ft=None) and keyword/text fields never enter it — shim/non-object behavior is unchanged.

Non-blocking:
- term on an object SUB-field (e.g. meta.owner) remains flush-divergent — explicitly out-of-scope here (issue #677, Term arm untouched), not introduced by this fix.
- The fix's non-object fallthrough for an object-mapped field carrying a String/Number value uses the pre-existing fold-both heuristic; that is prior behavior, unchanged by this PR.

---
**Merge gate (b) ZERO BLOCKING: met.** Non-blocking items filed/known: the sub-field `term` divergence is #677 (Term arm deliberately untouched); `prefix{case_insensitive:true}` on an object field is a pre-existing untested edge (prefix has no ci flag; mirrors the ES keyword-prefix default) — will note as a minor follow-up. Rebased onto rc.57 main; merging on all-green + CLEAN.

PR #678, opened 2026-08-23
fix(engine): #677 term on an object sub-field is flush-invariant (dotted-path source scan)

### LINKED ISSUE #677: term on an object sub-field (meta.owner) changes its answer at flush — a #423 sibling of the #413 prefix/wildcard fix
### `term` on an object **sub-field** changes its answer at flush (a #423 sibling)

Found while fixing #413 (the object-field **prefix/wildcard** half). This is the **`term`** half, on a sub-field of an object.

Written by an AI agent (Claude Opus 4.8) driven by @xerj-org.

## Repro (rustc 1.97.1)

Field `meta` mapped `object`, document `{"meta":{"owner":"Alpha"}}`:

| query | before flush (memtable) | after flush (segment) |
|---|---|---|
| `{"term":{"meta.owner":"Alpha"}}` | **0** | **1** |

Observed directly in an integration test (size:0 total hits): memtable = 0, segment = 1 — the same document, the same query, two answers decided by a background flush.

The segment indexes `meta.owner` = "Alpha" as its own term (via `collect_text_fields`'s recursion into the object), so post-flush answers 1. The memtable `doc_matches_query` path returns 0 for the same `term` on the dotted sub-field.

## Why this is #423, not #413

#413 is the object field's **prefix/wildcard** blob and is fixed in PR #676 (the memtable now mirrors the whole-object blob). That fix deliberately does **not** touch the `Term` arm, and the sub-field `term` path is a separate divergence in the same root-cause family described by #423 (the schemaless `_source` scan disagreeing with the analysed term dictionary). Filing separately so it is individually trackable like the other #423 symptoms.

## Acceptance

`{"term":{"meta.owner":"Alpha"}}` must return the same count before and after a flush (and match ES: 1, since `meta.owner` is an indexed keyword sub-field).

xerj-org:
## #677 — `term` on an object sub-field changes its answer at flush

Closes #677 (an #423 sibling of the #413 prefix/wildcard fix).

### The bug
Field `meta` mapped `object`, doc `{"meta":{"owner":"Alpha"}}`:

| query | before flush | after flush |
|---|---|---|
| `{"term":{"meta.owner":"Alpha"}}` | **0** | **1** |

Probing size:0 vs size:5 pinpointed it: **matching is correct** (size:5 returns the doc, total=1 pre-flush) — it is a **size:0 count-only** undercount. The memtable's single-valued doc-values column can't faithfully represent a dotted path (`meta.owner` resolves through `_source`/`get_nested_field`, not a column — the same rule `terms_counts_columnar` already states), so the DV term-count answered 0 while the buffered source matched.

### The fix
Gate dotted-path term/range predicates to the **authoritative source scan** (always correct; the scan is the authority — only a bounded perf cost on dotted-field term counts), in every memtable path that counts/filters via the column:
- `dv_column_unusable` (covers `doc_values_term_query` / `doc_values_range_query` + the `try_shortcut_count` segment shortcut),
- `doc_values_bool_hits` (the size:0 + fused-bool count path) — Term and Range arms,
- `filtered_docs_arc_into` (the size>0 filter path) — Term and Range arms, so count and filter agree.

- New test `object_subfield_term_flush_parity.rs`: `term{meta.owner}` is flush-invariant (pre==post==1).
- **Fail-before:** reverting only the memtable gates → the test fails `pre=0 != post=1`; restored → passes.
- Full `xerj-engine` suite: **0 failures**. Full `xerj-query` suite: **0 failures**. `clippy --all-targets -D warnings`: clean. `fmt --all --check`: clean.

### DIFF
--- a/engine/crates/xerj-engine/src/memtable.rs
+++ b/engine/crates/xerj-engine/src/memtable.rs
@@ -1797,6 +1797,17 @@ impl ShardedFtsMemtable {
     /// refusal globally (`?` inside its loop); this makes the single-leaf
     /// wrappers behave the same way.
     fn dv_column_unusable(&self, field: &str) -> bool {
+        // #677: a dotted path (an object sub-field like `meta.owner`, or a
+        // `.keyword` multi-field) resolves through `get_field_value` /
+        // `get_nested_field` against the stored `_source`, NOT a faithful
+        // single-valued DV column — the same rule `terms_counts_columnar`
+        // states. Counting such a field via the column undercounts (0 while
+        // the buffered source matches → the #677 size:0-vs-size:5 split), so
+        // force the authoritative source scan. Always correct (the scan is the
+        // authority); only a bounded perf cost on dotted-field term counts.
+        if field.contains('.') {
+            return true;
+        }
         self.shards.iter().any(|s| {
             let g = s.read();
             g.doc_values.array_fields.contains(field)
@@ -3366,6 +3377,14 @@ impl FtsMemtable {
                     if self.doc_values.array_fields.contains(field.as_str()) {
                         return None;
                     }
+                    // #677: a dotted path (object sub-field / `.keyword`
+                    // multi-field) is not faithfully represented by a single
+                    // DV column — it resolves through `_source`. Bail to the
+                    // source scan, else a `term` on `meta.owner` counts 0 while
+                    // buffered but matches after flush (size:0 undercount).
+                    if field.contains('.') {
+                        return None;
+                    }
                     if let Some(col) = self.doc_values.keyword.get(field.as_str()) {
                         // Step 2: analyzed-text bailout via the insert-time
                         // cached flag instead of an O(N) per-query column
@@ -3394,6 +3413,10 @@ impl FtsMemtable {
                     lt,
                 } => {
                     if self.doc_values.array_fields.contains(field.as_str()) {
+                        return None;
+                    }
+                    // #677: dotted path → source scan (see the Term arm).
+                    if field.contains('.') {
                         return None;
                     }
                     let col = self.doc_values.numeric.get(field.as_str())?;
@@ -3486,6 +3509,12 @@ impl FtsMemtable {
                     if self.doc_values.array_fields.contains(field.as_str()) {
                         return None;
                     }
+                    // #677: dotted path (object sub-field / `.keyword`) → source
+                    // scan, so the count and filter paths agree (see the count
+                    // fn `doc_values_bool_hits`).
+                    if field.contains('.') {
+                        return None;
+                    }
                     if let Some(col) = self.doc_values.keyword.get(field.as_str()) {
                         if self
                             .doc_values
@@ -3509,6 +3538,10 @@ impl FtsMemtable {
                     lt,
                 } => {
                     if self.doc_values.array_fields.contains(field.as_str()) {
+                        return None;
+                    }
+                    // #677: dotted path → source scan (see the Term arm).
+                    if field.contains('.') {
                         return None;
                     }
                     let col = self.doc_values.numeric.get(field.as_str())?;
--- a/engine/crates/xerj-engine/tests/object_subfield_term_flush_parity.rs
+++ b/engine/crates/xerj-engine/tests/object_subfield_term_flush_parity.rs
@@ -25,7 +25,6 @@ async fn hits(engine: &Engine, q: serde_json::Value) -> u64 {
 }
 
 #[tokio::test]
-#[ignore = "#677: memtable term path does not resolve object sub-fields; un-ignored by the fix"]
 async fn term_on_object_subfield_is_flush_invariant() {
     let dir = TempDir::new().unwrap();
     let mut config = Config::default();
@@ -47,6 +46,12 @@ async fn term_on_object_subfield_is_flush_invariant() {
     idx.refresh().await.expect("refresh");
     let post = hits(&engine, q.clone()).await;
 
-    assert_eq!(pre, post, "#677: term on an object sub-field must be flush-invariant");
-    assert_eq!(post, 1, "#677: meta.owner is an indexed keyword sub-field (ES matches)");
+    assert_eq!(
+        pre, post,
+        "#677: term on an object sub-field must be flush-invariant"
+    );
+    assert_eq!(
+        post, 1,
+        "#677: meta.owner is an indexed keyword sub-field (ES matches)"
+    );
 }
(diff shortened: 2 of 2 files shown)


xerj-org:

Run on the fix commit (`1a18c082`). **All three: 0 BLOCKING.** Verbatim:

> Base verified HEAD=1a18c082. Gates present in all three sites for BOTH Term and Range: dv_column_unusable (memtable.rs:1808 `if field.contains('.') { return true }`), doc_values_bool_hits (Term:3385, Range:3419), filtered_docs_arc_into (Term:3515, Range:3543). Wrote 3 throwaway integration test files (12 tests total), compiled+ran with rustup 1.97.1 --profile ci-test, RUST_MIN_STACK=8388608; all passed; deleted after, git status clean.
> 
> Fix-target cases (size:0 AND size:5, pre- AND post-flush = 4-tuple, all must equal target):
> - T1 term{meta.owner:Alpha} on {meta:{owner:Alpha}} -> 1/1/1/1 PASS
> - T2 term{meta.owner:Beta} (no match) -> 0/0/0/0 PASS
> - T3 term{meta.a.b:X} nested deeper -> 1/1/1/1 PASS
> - T4 range{obj.n gte1 lte10} -> 1 (in), range{obj.n gte6 lte10} -> 0 (out), both flush-invariant PASS
> - T6 term{title.keyword:...} multifield -> flush-invariant PASS
> 
> Adversarial batch (the ones most likely to break):
> - B1 object sub-field carrying ARRAY {meta:{owner:[Alpha,Beta]}}: term:Beta (SECOND element, the classic first-element-DV drop) -> 1/1/1/1; Alpha->1; Gamma->0 PASS. Source-scan fallback correctly recovers array membership on a dotted path.
> - B2 multi-doc exact counts: Alpha=3, Beta=2, flush-invariant PASS
> - B3 docs missing the path (no meta / meta.other) -> only d0 counts =1 PASS
> - B4 fused bool filter [meta.owner:Alpha AND status:on] -> 1 PASS (dotted+non-dotted mix)
> - B5 dotted numeric range boundaries over 10 docs: [3,7]=5, gt7=2 PASS
> - C1 EDGE flat keyword field NAMED 'a.b' with flat source key {\"a.b\":\"X\"}: 1/1/1/1 flush-invariant PASS (feared source-scan miss on a flat dotted key did NOT occur — get_field_value resolves it).
> 
> NON-REGRESSION (non-dotted, must be identical to before fix, all flush-invariant):
> - keyword scalar name:Alpha ->1, name:Beta ->0
> - keyword ARRAY tags:a ->1, tags:b ->1, tags:c ->0 (the #408 case, still correct)
> - numeric term age:30 ->1; range age[10,40] ->1, age gte31 ->0
> - text body term -> flush-invariant (0/0/0/0)
> Repo's own un-ignored test object_subfield_term_flush_parity also passes.
> 
> No memtable!=segment divergence, no wrong count, no non-dotted regression found. The gate is a strict superset (dotted-only) forcing the authoritative source scan, which is always correct; cost is a bounded perf hit as claimed.

Non-blocking:
- term on an analyzed Text field (`term{body:hello}` over body="hello world") returns 0 in XERJ, not 1 as ES would. This is pre-existing text-term semantics unrelated to #677 (the `body` field has no dot, so the fix's `field.contains('.')` gate never fires) and it is flush-invariant. Out of scope for this PR, noting only.

> Base: git rev-parse HEAD = 1a18c082 (correct). Merge-base origin/main..HEAD = c66121de; `git diff c66121de..HEAD --stat` = ONLY memtable.rs (+33) and new test object_subfield_term_flush_parity.rs (+57), NO existing test edited. The larger origin/main..HEAD diff is a stale-base artifact. Gate diff: three memtable predicate arms + dv_column_unusable each add `if field.contains('.') { return None/true }` — non-dotted paths unchanged. Suites (rustup 1.97.1, ci-test): xerj-engine 57 `test result: ok`, 0 failed, new test `term_on_object_subfield_is_flush_invariant` ok; xerj-query 3 ok, 0 failed; clippy --workspace --all-targets -D warnings EXIT=0; fmt --all --check EXIT=0. Fail-before: reverting memtable.rs to merge-base, the new test FAILED with left:0 (pre-flush) right:1 (post-flush) — the exact #677 undercount — proving the fix is load-bearing; restored HEAD memtable.rs, tree clean.

Non-blocking:
- Stale base: the branch's merge-base with origin/main is c66121de, and main has since advanced. `git diff origin/main..HEAD` therefore superficially shows reverts of index.rs (174 lines), ast.rs, parser.rs and deletions of tests/fuzzy_keyword_case_sensitive.rs + tests/object_field_blob_flush_parity.rs. These are NOT introduced by the fix (a three-way merge would keep them); the true fix touches only memtable.rs + the new test. Recommend rebasing onto current main before merge to confirm the memtable gate still composes with main's index.rs changes and to avoid a confusing diff.
- The gate keys on `field.contains('.')`, so a legitimately top-level keyword/numeric field whose name literally contains a dot would now always take the source scan instead of its DV column. This is a bounded perf cost only, never a wrong result (the source scan is authoritative). Non-blocking.

### EVIDENCE-HOLDS (fail-before is load-bearing) — not refuted (0 blocking)

> Base verified: git rev-parse HEAD = 1a18c082c69987850a050c590b7ed71747b868b8. Gates present in memtable.rs at lines 1800 (dv_column_unusable: `if field.contains('.') { return true }`), 3380/3418 (doc_values_bool_hits Term+Range), 3512/3543 (filtered_docs_arc_into Term+Range).\n\nState 1 BASELINE (1a18c082 intact): object_subfield_term_flush_parity -> `test result: ok. 1 passed`.\n\nState 2 REVERT ONLY source gates (git checkout origin/main -- memtable.rs, gates confirmed gone, test kept): rebuilt + ran -> FAILED at assertion left=0 (pre-flush) != right=1 (post-flush), \"term on an object sub-field must be flush-invariant\". Fail-before is load-bearing: the test genuinely guards the fix.\n\nSize probe under the reverted source: pre-flush size:0 total=0 (undercount) vs size:5 total=1 (doc returned). Confirms the bug is specifically a size:0 count-only undercount; matching itself is correct.\n\nState 3 RESTORE (git checkout HEAD -- memtable.rs; probe file removed): 6 #677 markers back; re-ran parity test -> `1 passed`; `git status --short` empty (clean tree).\n\nNon-dotted regression suite (all pass): count_matches_search_on_text_term 5/5, fast_aggs_keyword_predicates 2/2, fast_aggs_multifield_suffix 9/9, fast_aggs_nested_object 5/5, multi_valued_keyword_terms 6/6, prefix_keyword_case_flush_parity 4/4, term_on_keyword_array_flush_parity 1/1, bool_keyword_term_scoring 5/5. Keyword/numeric term/range counts on non-dotted fields remain correct; the fix only diverts dotted paths to the authoritative source scan, which is always correct.

Non-blocking:
- The gate is a purely syntactic `field.contains('.')` check. A legitimately flat field whose name literally contains a dot would also be routed to the source scan. This is not a correctness regression (source scan is authoritative) but a bounded perf cost on any dotted field name, as the fix comments acknowledge.

---
**0 BLOCKING across all three.** Acting on the REGRESSION lens's stale-base note: **rebased onto current main** (`cd28df52`, picks up #675 + #676) — gates intact, no conflict; full engine+query+clippy+fmt re-verifying the composition, then merge on all-green + CLEAN. Non-blocking notes (text-term semantics; flat field with a literal dot in its name → bounded perf, always correct) are out of scope / follow-ups.

Issue #669, opened 2026-08-23
matched_queries (_name) uses the schemaless fold-shim, disagreeing with schema-aware hit selection (post-#396)


xerj-org:
## `matched_queries` (`_name`) for a named `prefix`/`term` clause uses the schemaless fold-shim, so it can disagree with the (now schema-aware) hit selection

Follow-up from #396 (#667). After #667, memtable hit-SELECTION routes through `doc_matches_query_typed` with the real schema (keyword `prefix` is case-sensitive). But `collect_matched_queries_inner` (index.rs ~37576) still calls the 2-arg schemaless `doc_matches_query` shim (fold-both). So for a NAMED clause, e.g. `{prefix:{name:"hel", _name:"p"}}` on a buffered doc `{name:"Hello"}` (keyword):

- hit selection (case-sensitive): "Hello" does not start with "hel" → doc not matched by the clause.
- `_name` collection (fold): matches → reports `"p"` in `matched_queries`.

And the `_name` channel can differ pre-flush (fold) vs post-flush (segment, case-sensitive). The **hit set is correct**; only the `_name` annotation is inconsistent. This is the #396 flush-variance class, confined to the `matched_queries` output.

### Fix
Thread the index schema into `collect_matched_queries_inner` (and `apply_function_score`'s filter eval) so the `_name`/function-score channels use the same `doc_matches_query_typed` semantics as hit selection. Add a parity test asserting `matched_queries` for a named keyword prefix agrees with the hit set, pre- and post-flush.

Non-blocking / niche (named prefix/term with a case-mismatch on a buffered keyword doc), but it is a real consequence of the #396 fix intentionally leaving the shim unchanged.

xerj-org:
Fixed and merged. `collect_matched_queries_inner` (and the function-score filter eval) now thread the index `Schema` and route through `doc_matches_query_typed`, so the `_name` channel uses the same schema-aware, flush-invariant semantics as hit selection.

- Fix: PR #680 (merge `ee9f7261`, ancestor of current main).
- Parity test: `engine/crates/xerj-engine/tests/matched_queries_schema_aware.rs` (asserts a named keyword `prefix`/`term` in `matched_queries` agrees with the hit set, pre- and post-flush).

Closing as resolved via the merged fix.

PR #680, opened 2026-08-23
fix(engine): #669 matched_queries (_name) uses the schema-aware matcher

xerj-org:
## #669 — `matched_queries` (`_name`) used the schemaless fold-shim, disagreeing with schema-aware hit selection

Follow-up to #396/#667. After #667, memtable hit-SELECTION routes through `doc_matches_query_typed` with the real index schema (keyword `prefix`/`wildcard` are case-sensitive). But `collect_matched_queries_inner` (index.rs) still called the 2-arg schemaless `doc_matches_query` shim (fold-both), so the `_name` / `matched_queries` channel could report a name the hit set never granted — and differ pre- vs post-flush.

### Repro (rustc 1.97.1)
Field `name` mapped `keyword`, doc `{"name":"Hello"}`, a `function_score` whose function carries `_name:"fp"` over a keyword-`prefix` **filter** `{"prefix":{"name":"hel"}}`:

| | schemaless fold-shim (before) | schema-aware (after / ES) |
|---|---|---|
| filter `prefix name:"hel"` on `"Hello"` | folds → matches → reports `"fp"` | case-sensitive → no match → no `"fp"` |

`matched_queries` wrongly contained `"fp"`; hit selection never granted it.

### Fix
Thread the index schema (`dmq_schema = self.schema()`, already bound at the call site) through `collect_matched_queries` + `collect_matched_queries_inner`, and evaluate the named-clause leaf and the `function_score` filter with `doc_matches_query_typed` instead of the fold-shim — so the `_name` annotation matches hit selection on every field type and is flush-invariant.

### Verification
- New test `matched_queries_schema_aware.rs`: the function `_name` must NOT appear (keyword-prefix filter is case-sensitive), flush-invariant.
- **Fail-before:** reverting only the `index.rs` hunk → the test fails, `matched_queries` = `["fp"]`; restored → passes.
- Full `xerj-engine` + `xerj-query` suites: 0 failures. `clippy --all-targets -D warnings` clean. `fmt --all --check` clean.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -18161,7 +18161,8 @@ impl Index {
             } else {
                 hit.source.clone()
             };
-            hit.matched_queries = collect_matched_queries(&request.query, &source_with_id);
+            hit.matched_queries =
+                collect_matched_queries(&request.query, &source_with_id, &dmq_schema);
         }
 
         // --- Apply search_after cursor (skip hits up to and including the cursor) ---
@@ -37685,20 +37686,31 @@ fn point_in_polygon(lat: f64, lon: f64, polygon: &[(f64, f64)]) -> bool {
 ///
 /// Recursively traverses the query tree, collecting the `name` of every `Named`
 /// query whose inner query matches the document.
-fn collect_matched_queries(q: &QueryNode, source: &Value) -> Vec<String> {
+fn collect_matched_queries(q: &QueryNode, source: &Value, schema: &Schema) -> Vec<String> {
     let mut names = Vec::new();
-    collect_matched_queries_inner(q, source, &mut names);
+    collect_matched_queries_inner(q, source, schema, &mut names);
     names
 }
 
-fn collect_matched_queries_inner(q: &QueryNode, source: &Value, names: &mut Vec<String>) {
+// #669: the `_name` / `matched_queries` channel must decide a named clause with
+// the SAME schema-aware matcher hit-SELECTION uses (`doc_matches_query_typed`
+// with the real index schema, post-#396/#667). The old 2-arg `doc_matches_query`
+// shim folded case for every field, so a named keyword `prefix`/`term` clause
+// could report a `_name` the hit set never granted (and disagree pre- vs
+// post-flush). Thread the schema through so the annotation matches selection.
+fn collect_matched_queries_inner(
+    q: &QueryNode,
+    source: &Value,
+    schema: &Schema,
+    names: &mut Vec<String>,
+) {
     match q {
         QueryNode::Named { name, query } => {
-            if doc_matches_query(query, source) {
+            if doc_matches_query_typed(query, source, schema) {
                 names.push(name.clone());
             }
             // Also recurse into the wrapped query in case it contains further Named nodes.
-            collect_matched_queries_inner(query, source, names);
+            collect_matched_queries_inner(query, source, schema, names);
         }
         QueryNode::Bool {
             must,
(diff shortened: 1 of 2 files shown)


xerj-org:

Run on `33b002ef`. **All three: 0 BLOCKING.** Verbatim:

> Reset to origin/fix/669-matched-queries-schema, HEAD=33b002ef confirmed. Fix present: collect_matched_queries/_inner take &Schema (index.rs:37689/37701), evaluate Named leaf and function_score filter via doc_matches_query_typed (:37709,:37755); call site index.rs:18165 passes dmq_schema=self.schema(). Wrote throwaway integration test (tests/zz_refute_669.rs), built+ran with RUST_MIN_STACK=8388608, rustup 1.97.1, --profile ci-test. Results: t1 keyword-prefix fp {'hel' NOT reported, 'Hel' reported} pre AND post refresh = PASS; t2 non-regression named match(text)+term(keyword)+match still reported, flush-invariant = PASS; t3 keyword wildcard case-sensitive (hel* no / Hel* yes) + text prefix agree, flush-invariant = PASS. Shipped guard matched_queries_schema_aware.rs = PASS. All _name sets identical pre vs post refresh (no flush divergence). Deleted throwaway; git status clean. Two initial t2 failures (b_bool, p_text dropped) traced to pre-existing parser gaps: parse_bool/parse_prefix never call maybe_named; fix commit only touched index.rs + test file, so not a regression introduced by this fix.

Non-blocking:
- parse_bool (engine/crates/xerj-query/src/parser.rs:2638) never reads `_name`, so a name placed directly on a bool clause is silently dropped and never appears in matched_queries (ES supports it). Pre-existing; parser untouched by fix commit 33b002ef.
- parse_prefix (parser.rs:1238) and parse_wildcard (:1276) never call maybe_named, so `_name` on a prefix/wildcard clause is dropped. Pre-existing parser gap, orthogonal to #669.

> Base confirmed: HEAD=33b002efaceb4293249f24ce73ba66ce2bbcbd1a; grep '#669' present at index.rs:37695. Diff origin/main..HEAD --stat shows exactly 2 files: engine/crates/xerj-engine/src/index.rs (+28/-16) and the new tests/matched_queries_schema_aware.rs (+70); NO existing test edited. The change threads `schema: &Schema` through collect_matched_queries/collect_matched_queries_inner, switches the Named leaf AND the function_score filter from schemaless doc_matches_query to doc_matches_query_typed, and the single caller passes &dmq_schema (=self.schema(), the same schema hit-selection uses at 15623/15634/18497). No residual schemaless doc_matches_query( remains in the collect_matched region. Suites (rustup 1.97.1, RUST_MIN_STACK=8388608, --profile ci-test): cargo test -p xerj-engine EXIT=0 (59 'test result:' lines, all 0 failed; new test matched_queries_name_is_schema_aware_and_flush_invariant ok); cargo test -p xerj-query EXIT=0 (3 result lines, 0 failed); clippy -p xerj-query -p xerj-engine --all-targets -D warnings EXIT=0; cargo fmt --all --check EXIT=0. Regression guard: pre-existing integration.rs test_named_queries_matched (legit text 'match' with _name on title/body) still ok unchanged — a legitimately-matching named query is still reported. New test proves the case-sensitive keyword-prefix function _name 'fp' is NOT reported and pre==post flush-invariant.

Non-blocking:
- collect_matched_queries now evaluates named leaves via doc_matches_query_typed against _source. For analyzed/text fields, source re-evaluation can in principle still diverge from postings-based hit selection, but this is pre-existing (hit selection at index.rs:18497 uses the same doc_matches_query_typed-on-source approach); the fix strictly moves the _name channel closer to selection, so it is not a regression.

> Base confirmed HEAD=33b002ef. Fix present: index.rs:14139 dmq_schema=self.schema().await; collect_matched_queries(&request.query,&source_with_id,&dmq_schema) at 18165; collect_matched_queries_inner threads schema and evaluates the Named leaf (37709 doc_matches_query_typed) AND the function_score filter (37755 doc_matches_query_typed).
> 
> Three-state run (RUST_MIN_STACK=8388608, engine, rustup 1.97.1 cargo test -p xerj-engine --profile ci-test --test matched_queries_schema_aware):
> (1) BASELINE 33b002ef: 1 passed.
> (2) REVERT ONLY index.rs to origin/main (grep -c '#669' -> 0, test kept), rebuild+run: FAILED at test line 65 with panic "the function _name must agree with hit selection (no 'fp'): [\"fp\"]" — the fold-shim wrongly folds the case-sensitive keyword prefix filter and reports 'fp'. Test genuinely guards the fix.
> (3) RESTORE index.rs from HEAD (grep -c '#669' -> 1), re-run: 1 passed; git status --porcelain clean.
> 
> Diff main..HEAD on index.rs is confined to schema-threading: adds schema param and swaps doc_matches_query -> doc_matches_query_typed at the two decision points; no unrelated behavior change. dmq_schema is the identical schema instance used by hit selection (15623/15634/16597), so annotation agrees with selection by construction — flush-invariant, no regression where a legitimately-matching named query stops being reported.

---
**0 BLOCKING across all three.** Non-blocking parser gap (`_name` on bool/prefix/wildcard clauses is dropped by the parser — `parse_bool`/`parse_prefix`/`parse_wildcard` never call `maybe_named`) is **pre-existing and orthogonal** to this fix (which touches only `index.rs`); I'll file it as a follow-up. Merging on all-green + CLEAN; #669 batches into rc.59 with #677.

Issue #643, opened 2026-08-23
bool score-neutrality: unwrap is conservative — rare 1-clause shapes still bypass it (follow-up to #399)


xerj-org:
Follow-up to #399 (merged in #642). The `unwrap_single_clause_bool` fix erases the common one-clause `bool` shapes Lucene rewrites to the clause — a lone `must`, and a lone `should` with `minimum_should_match` `None`/`Fixed(0)`/`Fixed(1)`. The 3-skeptic verification of #642 confirmed it erases **only** those shapes (no over-unwrapping, no semantics change), but flagged that a few shapes Lucene *also* collapses are left wrapped, so #399's memtable score/rank divergence can still surface for them:

- a lone `should` whose `minimum_should_match` is a **percentage / field / script** form that *resolves to 1* (only the `Fixed(0)`/`Fixed(1)`/`None` variants are matched; `MinShouldMatch::{Percentage,Field,Script}` fall through unmatched);
- a one-clause `bool` **nested inside a wrapper the unwrap does not recurse through** — `unwrap_single_clause_bool` recurses through `Boosted`/`Constant`/`Named` but not `DisMax`/`FunctionScore`/etc.

These are strictly conservative: every un-erased shape keeps `bool` semantics (Σ over ≥0 clauses is never *more* than the clause), so nothing is wrongly scored — this is a **completeness** gap, not a correctness bug, which is why it did not block #642.

**Fix direction:** resolve `minimum_should_match` to its effective integer before the `≤1` test, and extend the recursion to the remaining score-preserving wrappers (`DisMax` single-clause, `FunctionScore` inner query, etc.).

xerj-org:
Partial progress: the **Percentage** sub-facet of point (1) is now fixed and merged. `unwrap_single_clause_bool` resolves a lone-`should` `minimum_should_match` of `Percentage(pct)` and collapses it iff it resolves to 1 (`pct < 200`, matching the `max(1, floor(len*pct/100))` rule for a single should-clause), while leaving `>=200%` (which matches nothing) correctly wrapped. PR #686 (commit `a018a30b`), test `bool_single_clause_is_score_neutral.rs` (adds two percentage-neutral cases + `high_percentage_msm_lone_should_matches_nothing`).

**Still open (why this stays open):**
[...]

PR #686, opened 2026-08-23
fix(engine): #643 single-clause bool unwrap covers a percentage minimum_should_match

xerj-org:
## #643 — single-clause `bool` unwrap now covers a percentage `minimum_should_match`

Follow-up to #399 / #642. `unwrap_single_clause_bool` erased a lone `should` only for `None`/`Fixed(0)`/`Fixed(1)` msm, so #399's memtable score/rank divergence still surfaced for a lone `should` whose `minimum_should_match` is a **percentage**.

### Fix
Over a single optional clause a percentage always resolves to 1 — you cannot require more than one of one clause, and with no required clause the effective count is coerced to ≥1 (the same reason `Fixed(0)` is erasable). So `Percentage(_)` is added to the lone-`should` arm.

`Field`/`Script` msm are **deliberately left wrapped**: they resolve at query time and can exceed the clause count (`msm > numOptional` ⇒ the bool matches nothing), which the bare clause would not — collapsing them would be wrong, so this stays strictly conservative.

Scope note: the issue also mentions recursing the unwrap through `function_score`/`dis_max`. I could not construct a fail-before for that (those wrappers transform the score and mask any inner-bool divergence — the score-neutrality test passes with or without the recursion), so per the fail-before discipline it is **not** included here; if a divergent case is found it can be a separate PR.

- Two percentage cases added to `bool_single_clause_is_score_neutral.rs` (`"100%"` and `"50%"`), asserted score/rank-neutral vs the bare clause on **memtable + segments + every page size**.
- **Fail-before:** reverting only the `index.rs` hunk → `bool.should[1]+pct100` scores `d004` at `2.79` vs the bare clause's `0.0068` on the memtable (the exact #399 divergence); restored → neutral.
- Full `xerj-engine` + `xerj-query` suites: 0 failures. `clippy --all-targets -D warnings` clean. `fmt --all --check` clean.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -35005,13 +35005,22 @@ fn unwrap_single_clause_bool(q: QueryNode) -> QueryNode {
                 return must.remove(0);
             }
             let only_should = must.is_empty() && must_not.is_empty() && filter.is_empty();
-            if should.len() == 1
-                && only_should
-                && matches!(
-                    minimum_should_match,
-                    None | Some(MinShouldMatch::Fixed(0)) | Some(MinShouldMatch::Fixed(1))
-                )
-            {
+            // #643: a lone `should` collapses to the bare clause only when its
+            // effective required count is 1 (the clause is required). msm
+            // resolution is `max(1, floor(len * pct / 100))` (the Bool matcher
+            // and FTS arms), so for a single clause a `Percentage` resolves to
+            // 1 iff `pct < 200`: `pct >= 200` resolves to >= 2 (requires 2 of 1
+            // clause => the bool matches NOTHING), unlike the bare clause, so it
+            // must stay wrapped. `Field`/`Script` resolve at query time and can
+            // likewise exceed the clause count, so they too stay wrapped
+            // (conservative). The parser puts no upper cap on `pct`, so `"200%"`
+            // is reachable.
+            let should_msm_collapses = match &minimum_should_match {
+                None | Some(MinShouldMatch::Fixed(0)) | Some(MinShouldMatch::Fixed(1)) => true,
+                Some(MinShouldMatch::Percentage(pct)) => *pct < 200,
+                _ => false,
+            };
+            if should.len() == 1 && only_should && should_msm_collapses {
                 return should.remove(0);
             }
             QueryNode::Bool {
--- a/engine/crates/xerj-engine/tests/bool_single_clause_is_score_neutral.rs
+++ b/engine/crates/xerj-engine/tests/bool_single_clause_is_score_neutral.rs
@@ -172,7 +172,44 @@ fn neutral_wrappers() -> Vec<(&'static str, Value)> {
             "bool.must[bool.must[1]]",
             json!({"bool": {"must": [{"bool": {"must": [text_query()]}}]}}),
         ),
+        (
+            // #643: a lone `should` with a PERCENTAGE msm resolves to 1 for a
+            // single optional clause (coerced to ≥1 when there is no required
+            // clause), so it is the same query and must now collapse like
+            // Fixed(0)/Fixed(1).
+            "bool.should[1]+pct100",
+            json!({"bool": {"should": [text_query()], "minimum_should_match": "100%"}}),
+        ),
+        (
+            "bool.should[1]+pct50",
+            json!({"bool": {"should": [text_query()], "minimum_should_match": "50%"}}),
+        ),
     ]
+}
+
+/// #643 guard: a lone `should` whose `minimum_should_match` is a percentage
+/// >= 200% resolves to `max(1, floor(1 * pct/100)) >= 2` — it requires 2 of 1
+/// optional clause, which is unsatisfiable, so the bool matches NOTHING. It must
+/// stay wrapped, NOT collapse to the bare clause (which matches). Guards the
+/// over-broad `Percentage(_)` unwrap the 3-skeptic pass caught.
+#[tokio::test]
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification (revised fix `a018a30b`) — rustc 1.97.1, `--profile ci-test`, isolated worktrees

An earlier pass caught an **unbounded-percentage regression** (msm >= 200% was wrongly unwrapped → matched everything). This revision gates the unwrap on `pct < 200`. Re-verification — **all three: 0 BLOCKING.** Verbatim:

> Base HEAD=a018a30b (correct); guard `pct < 200` present at index.rs:35020. Resolver confirmed max(1,floor(len*pct/100)) at index.rs:35961 and 40834; for len=1: pct<200 -> required 1 (neutral), pct>=200 -> required>=2 (matches nothing); threshold exactly right vs floor (199%->1, 200%->2); no f32 rounding hazard. Guard arms: None/Fixed(0)/Fixed(1)->unwrap, Percentage(pct)->pct<200, Fixed(n>=2)/Field/Script->`_ => false` stay wrapped. Throwaway integration tests (memtable AND flushed segment, size:50), then deleted + git clean: Case1/3 msm 0/1/50/100/150/199% lone-should ALL neutral vs bare (total 120, scores within 1e-4, same order); Case2 200/250/300/1000% ALL match NOTHING (0 hits) - previously-found bug is gone; Case4 Fixed(2)->0, terms_set Field=1->60/Field=2->0, Script '1'->60/Script '2'->0 (all stay wrapped); Case5 None/Fixed(0)/Fixed(1) neutral, two-should 50% (1 of 2)->120 not unwrapped. All 6 throwaway tests pass. PR's own test bool_single_clause_is_score_neutral: 7/7 pass. `cargo clippy --profile ci-test -p xerj-engine` under rustc 1.97.1: clean, no warnings. Could not break the fix.

Non-blocking:
- Pre-existing (not touched by this PR): a lone should with Fixed(0) unwraps to the bare clause, but the wrapped resolver would treat min=0 as 'skip the should check' (index.rs:36001 `if min > 0`), i.e. match everything. The unwrap makes Fixed(0) behave as the bare clause, which is the desired outcome, and this arm predates #643, so it is not a regression here.

> Base HEAD=a018a30b (correct). Guard arm `Some(MinShouldMatch::Percentage(pct)) => *pct < 200` present at index.rs:35020. Fix is provably correct against the actual msm resolution used at index.rs:35961 (stored-doc matcher) and index.rs:40834 (FTS arm), both `max(1, floor(len*pct/100))`: for a lone clause pct<200 -> required=1 -> neutral vs bare clause; pct>=200 -> required>=2 -> matches nothing, so stays wrapped. The earlier over-broad Percentage(_) bug is gone. Fixed(2)+/Field/Script fall to `_ => false` (stay wrapped); None/Fixed(0)/Fixed(1) collapse unchanged. Suites: `cargo test -p xerj-engine --profile ci-test` EXIT=0, every `test result:` shows 0 failed (626 unit + new high_percentage_msm_lone_should_matches_nothing ok + neutral_wrappers pct100/pct50 ok). `cargo test -p xerj-query --profile ci-test` EXIT=0 (183+9 passed, 0 failed). `cargo clippy -p xerj-query -p xerj-engine --profile ci-test --all-targets -- -D warnings` EXIT=0 (2 grep hits are `Checking thiserror` package lines, not warnings). `cargo fmt --all --check` EXIT=0 (empty output). git diff origin/main..HEAD --stat = only index.rs (one @@ hunk: the should_msm_collapses match) + tests/bool_single_clause_is_score_neutral.rs (pure additions, no existing assertion altered). Confirmed no separate should_required/other match block corrupted (single hunk in index.rs).

> Base HEAD a018a30b confirmed; guarded arm `Some(MinShouldMatch::Percentage(pct)) => *pct < 200` present at index.rs:35020. (1) BASELINE: cargo test --test bool_single_clause_is_score_neutral -> 7 passed; 0 failed (incl high_percentage_msm_lone_should_matches_nothing, pct100/pct50 neutral). (2) FEATURE fail-before (git checkout origin/main -- index.rs, keep tests): single_clause_bool_is_score_neutral_on_memtable FAILED — bool.should[1]+pct100 _score for d004 is 2.7917595 vs bare 0.0068451734 (percentage not unwrapped on main); restored with git checkout HEAD. (3) GUARD fail-before (edit to Percentage(_) => true): high_percentage_msm_lone_should_matches_nothing FAILED — 200% unwrapped, left:120 right:0; restored. Both fail-befores hold. Clippy: cargo clippy -p xerj-engine --all-targets -> Finished, no warnings. Boundary verified exact: MinShouldMatch::Percentage(u32) integer; collapse decision `pct<200` and runtime resolutions at index.rs:35961 and 40834 (floor(len*pct/100).max(1)) consume identical integer — pct=199 resolves to 1 (collapse valid), pct=200 resolves to 2 (matches nothing, stays wrapped). Fixed is u32 (no negatives); Fixed(2+)/Field/Script hit `_ => false` and stay wrapped; None/Fixed(0)/Fixed(1) still collapse. git status clean at end; :9200 never touched.

Non-blocking:
- ast.rs:99 doc-comment still says Percentage is "(0–100)" while the parser and this fix rely on values up to 200+ being reachable; cosmetic, not part of this change.

---
**0 BLOCKING across all three.** The ≥200% regression from the first pass is gone; the perl used in the first guard-fail-before had corrupted a separate `should_required` block, which was reset (`git checkout origin/main -- index.rs`) and the fix re-applied as a single clean hunk. Non-blocking: `ast.rs:99` doc-comment still says Percentage "(0–100)" — cosmetic, will note as a follow-up. Merging on all-green + CLEAN.

PR #688, opened 2026-08-24
fix(autoindex): #673 prefix-scope correlation catalog ids (corr:/tcorr:)

### LINKED ISSUE #673: autoindex: correlation/run catalog IDs not prefix-scoped (latent cross-corpus collision, follow-up to #416)
## autoindex catalog: correlation and run doc IDs are NOT corpus-prefix-scoped (latent cross-corpus collision)

Surfaced by the #416 (#672) verification. #416/#672 scoped `file:`/`ds:`/`file-alias:` catalog ids by the corpus `--prefix`, but two other id families in the shared `autoindex-catalog` index remain global:

- **correlation docs** — `correlate.rs` `c.id()` (correlation between two datasets) is keyed by dataset slugs, not prefix-scoped.
- **run doc** — `run:{generation_id}`.

**Not a live collision today** (why this is a follow-up, not a blocker):
- The production generation path passes an **empty** correlation map (`sync_executor.rs:620`), so correlation docs aren't written server-side yet.
- `run_id` is the globally-unique `tx_id`, so `run:{id}` can't collide across corpora.

**Becomes a real data-loss** the moment server-side correlations are enabled with two corpora sharing dataset slugs: `c.id()` for `(reports, orders)` in corpus A overwrites the same-slug correlation in corpus B.

**Fix:** thread the same corpus prefix into `correlate.rs::c.id()` (and, defensively, the run doc id) exactly as #416 did for the file/dataset ids, and extend `catalog_ids_are_prefix_scoped_across_corpora` to cover them.

xerj-org:
Closes #673.

## Problem
The shared `autoindex-catalog` index keyed correlation docs by dataset slugs only:
- `corr:{a_slug}:{b_slug}:{a_field}:{b_field}` (`KeyCorr::id`, correlate.rs)
- `tcorr:{a_dataset}:{b_dataset}` (lib.rs time-correlation push)

#416/#672 scoped `file:`/`ds:`/`file-alias:` ids by the corpus `--prefix`, but these two id families stayed **global**. The moment server-side correlations run for two corpora that share dataset slugs, `(reports, orders)` in corpus A silently overwrites the same-slug correlation in corpus B — cross-corpus data-loss (the follow-up #673 raised from the #416 verification).

## Fix
Thread the corpus `--prefix` into both ids exactly as #416 did:
- `corr:{prefix}:{a_slug}:{b_slug}:{a_field}:{b_field}`
- `tcorr:{prefix}:{a_dataset}:{b_dataset}`

`KeyCorr::id(&self, prefix: &str)` — its only caller is the catalog push in `lib.rs` (passes `&cfg.prefix`).

### run:{id} deliberately left global
`run_id` is the globally-unique `tx_id` (`run-{second}-{pid}-{seq}` with an atomic per-process sequence; on the generation path `tx_id = {run_id}-g{gen}`), so `run:{id}` **cannot collide across corpora** — the scoping this PR adds to `corr:`/`tcorr:` fixes a real *slug*-keyed collision that `run:{id}` simply does not have. This tx_id global-uniqueness argument is load-bearing and holds on its own. Secondarily, `run:{id}` is a *read-contract* in the generation-catalog projection (two writer paths — lib.rs + generation_catalog.rs — plus exact-id test reads), so changing it would be a contract change for zero collision benefit. It is therefore left as-is by design. (Note: `execution.prefix` *is* in scope in the projection — used as `run_doc["prefix"]` at generation_catalog.rs:220 — so the reason to leave `run:` global is the uniqueness argument above, not any absence of the prefix.)

- New test `correlate::tests::correlation_id_is_prefix_scoped_across_corpora` (distinct across corpora, idempotent within one, prefix actually in the id).
- **Fail-before proven**: reverting only the format-string hunk makes both corpora produce `corr:reports:orders:order_id:id` → `assert_ne!` fails (exit 101).
- Full gate green: `fmt --all --check` / `clippy --workspace --all-targets -D warnings` / `build --workspace --profile ci-test` / `test -p xerj-autoindex` (full) — all exit 0. No regression (the `corr:reports:old` at generation_catalog_http_tests.rs:480 is a synthetic stale-id purge test, unaffected).

### DIFF
--- a/engine/crates/xerj-autoindex/src/correlate.rs
+++ b/engine/crates/xerj-autoindex/src/correlate.rs
@@ -38,10 +38,15 @@ pub struct KeyCorr {
 }
 
 impl KeyCorr {
-    pub fn id(&self) -> String {
+    /// Catalog doc id, corpus-prefix-scoped (#673). The `autoindex-catalog`
+    /// index is shared across corpora, so a correlation keyed only by dataset
+    /// slugs (`corr:reports:orders:…`) would overwrite the same-slug
+    /// correlation from another corpus. Threading `prefix` — exactly as #416
+    /// did for `file:`/`ds:` ids — keeps them distinct.
+    pub fn id(&self, prefix: &str) -> String {
         format!(
-            "corr:{}:{}:{}:{}",
-            self.a_slug, self.b_slug, self.a_field, self.b_field
+            "corr:{}:{}:{}:{}:{}",
+            prefix, self.a_slug, self.b_slug, self.a_field, self.b_field
         )
     }
     pub fn to_value(&self) -> Value {
@@ -340,4 +345,52 @@ pub fn time_alignment(series: &[TimeSeries]) -> Vec<Value> {
         }
     }
     out
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    fn sample_corr() -> KeyCorr {
+        KeyCorr {
+            a_slug: "reports".into(),
+            a_index: "ax-reports".into(),
+            a_field: "order_id".into(),
+            b_slug: "orders".into(),
+            b_index: "ax-orders".into(),
+            b_field: "id".into(),
+            kind: "keyword".into(),
+            a_set: 100,
+            b_set: 100,
+            overlap: 80,
+            containment: 0.8,
+            grade: "strong".into(),
+            examples: vec!["500001".into()],
+            confirmed: None,
+            confirm_details: Vec::new(),
+        }
+    }
+
+    /// #673: the `autoindex-catalog` index is shared across corpora, so a
+    /// correlation keyed only by dataset slugs must carry the corpus prefix in
+    /// its id — otherwise `(reports, orders)` in corpus A overwrites the
+    /// same-slug correlation in corpus B (silent cross-corpus data-loss).
+    #[test]
+    fn correlation_id_is_prefix_scoped_across_corpora() {
+        let c = sample_corr();
+        // Same correlation, different corpus prefix -> distinct ids (no overwrite).
+        assert_ne!(
+            c.id("ax-a"),
+            c.id("ax-b"),
+            "#673: same-slug correlations from two corpora must not collide"
+        );
+        // Idempotent within one corpus (a re-run upserts, does not duplicate).
+        assert_eq!(c.id("ax-a"), c.id("ax-a"));
+        // The prefix is actually in the id (not merely concatenated onto slugs).
+        assert!(
+            c.id("ax-a").starts_with("corr:ax-a:"),
+            "id was {:?}",
+            c.id("ax-a")
+        );
+    }
 }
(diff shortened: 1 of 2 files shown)


xerj-org:

Each skeptic ran in its own git worktree at PR head `a19cbf86`, reset to `origin/fix/673-correlation-id-prefix-scope`, under **rustc 1.97.1**, prompted to REFUTE.

**Result: 3× CONFIRMED, 0 BLOCKING findings.** (4 NITs, no should-fix.)

The #673 fix is correct. Threading the corpus prefix into `corr:{prefix}:a_slug:b_slug:a_field:b_field` and `tcorr:{prefix}:a:b` makes two corpora with identical dataset slugs produce distinct catalog ids while keeping a same-corpus re-run idempotent (id derives only from prefix+slugs+fields, sanitized to `[a-z0-9-]`, never from mutable state). No delimiter-injection risk: `cfg.prefix` passes through `sanitize_slug` (emits only `[a-z0-9-]`, non-empty enforced), so it contains no `:` and the prefix boundary is unambiguous — full parity with the #416 `ds:`/`file:` pattern. The `run:{id}`-left-global argument is sound: on the lib.rs path `run_id` is `run-{ts}-{pid}-{seq}` minted per (root,url,prefix) state-dir; on the generation_catalog path it is `snapshot.tx_id = {run_id}-g{gen}`. Both are globally unique tx-ids, never a bare per-corpus generation counter. No other slug-keyed catalog family remains unscoped. New test passes.
- NIT: two corpora deliberately sharing the same `--prefix` still collide — but the prefix is the corpus discriminator by design (matches #416); out of scope.
- NIT: `a_field`/`b_field` are not `:`-sanitized, so a field name with `:` is ambiguous — but this predates #673 and prepending a `:`-free prefix creates no NEW collision.

No regression. Full `xerj-autoindex` suite green under rustc 1.97.1 (`cargo test` EXIT=0, 0 failed; new test passed). `clippy --all-targets -D warnings` clean; `fmt --check` clean. No production read path reconstructs correlation ids in the old unprefixed shape: `project_generation` (generation_catalog.rs:198-205) treats correlation ids as opaque map keys. The `generation_catalog_http_tests.rs:480 'corr:reports:old'` fixture is a synthetic stale-id purge string, not an assertion on `id()` output. `KeyCorr::id`'s signature change has exactly one production caller (lib.rs:5524, passing `&cfg.prefix`); the rest are test literals. `tcorr:` is write-only. `run:{id}` deliberately left global.

All evidentiary claims hold. Fail-before is decisive: reverting only the `corr:` format-string hunk makes the new test panic with an `assert_ne!` collision (`left==right=="corr:reports:orders:order_id:id"`); with the fix it passes. The test asserts all three claimed properties. The central safety claim is true: `corr:`/`tcorr:` are never read by exact id in production. `run:{id}` correctly left global — `run_id = tx_id = run-{second}-{pid}-{seq}` with an atomic per-process sequence (globally unique, test-guarded).
- NIT: `tcorr:` prefix-scoping is not unit-tested (only `corr:` via the new test) — filed as follow-up.
- NIT: PR body's "projection has no prefix in scope" is imprecise — `execution.prefix` IS in scope at generation_catalog.rs:220; the correct (and independently sufficient) justification for leaving `run:` global is `run_id`/`tx_id` global uniqueness. **PR body corrected.**

**Merge gate:** 0 BLOCKING → eligible once all checks are green + `mergeStateStatus` CLEAN. The current `Landing constants guard` failure is the tag-ahead stamp (landing = RC.60, tag = RC.61); it clears via a landing-only rebase onto post-rc.61-stamp main, which preserves these verdicts.

PR #690, opened 2026-08-24
fix(autoindex): #589 sweep a newly-excluded group's documents instead of refusing

### LINKED ISSUE #589: autoindex: sweep an excluded content group's documents from the destination (the purge half of #439)

xerj-org:
Closes #589.

## Problem (the purge half of #439)
A file still on disk that a widened ignore/hidden/`.xerjignore` rule now skips lands in `excluded_content_groups`. `refuses()` conflated it with a genuine deletion and **refused the whole rerun**, so the file's already-published documents stayed **live and searchable** — an exclusion that cannot remove data already in the index (#439's data-exposure headline). #439 only *distinguished* the two causes and gave an accurate message; this PR does the sweep.

## Fix
- **`refuses()` now gates on genuine deletions only.** An excluded-only delta no longer refuses — the run proceeds and sweeps.
- **`sweep_excluded_groups`** deletes each excluded group's indexed records (`{term:{ax_file}}`) across the plan's **exact dataset indices** (`plan.datasets[].index`, the same resolution the replacement-delete path at lib.rs uses — not a `{prefix}-*` wildcard, which could reach indices this corpus does not own) and its catalog document (`{term:{path}}`).
- Wired at **both** resume gates, guarded by **`!dry_run`** (no remote mutation on a dry-run/refused attempt).
- **`into_error` is unchanged**: it still reports the excluded set when a **genuine deletion co-occurs** — that case still refuses the whole rerun, so the excluded documents also stay live and the wording is still accurate.

## Scope / known gap
Graph edges taught by the file are **not** swept in this PR — deferred to a follow-up. (Correction: per-file edge provenance *does* exist — `detect::invalidate_prior_edges` soft-invalidates edges by `{term:{src_file: rel}}`, and the excluded `InventoryDeltaEntry` carries that `rel` as `path` — so the follow-up can reuse that hook rather than build new provenance. `edges_note` documents that these edges "stay live until that index is deleted too" today.) The data-**exposure** that matters — searchable document *content*, keyed by `ax_file` — is fully swept; the residual is edge/path metadata naming an excluded file, lower-severity and tracked as a follow-up. Guardrail-covered: `.xerjignore` / `--no-default-ignores` widening classify as excluded the same way (via `UnsupportedInventoryDelta::between`).

- **Fail-before proven**: reverting only the `refuses()` hunk makes `an_excluded_only_delta_is_swept_not_refused` fail again (exit 101 — excluded-only refuses).
- **Decision** unit tests: the repro + the flipped sibling `an_excluded_still_on_disk_group_is_not_reported_as_a_deletion` (its comment re-read and updated).
- **Sweep** integration test over the mock HTTP backend: `sweep_excluded_groups_deletes_only_the_excluded_files_documents` seeds two files' records + catalog docs, sweeps one, asserts its records **and** catalog doc are removed by a real `delete_by_query` while the other file is untouched (no over-deletion).
- Full gate: `fmt --all --check` / `clippy --workspace --all-targets -D warnings` / `build --workspace --profile ci-test` / `test -p xerj-autoindex` (full) — all green.

## Note on 2 pre-existing flaky tests
Under heavy concurrent load one full-suite run flaked on `sync_executor::tests::gc_tombstone_crash_is_retryable_and_symlinks_are_refused` and `incremental_reconcile_http_tests::two_byte_identical_junk_files_commit_reconcile_and_survive_a_shrink`. Both **pass in isolation and on 4/4 subsequent full runs**; neither exercises the exclusion/sweep code path this PR touches. This is pre-existing parallelism sensitivity, not a regression from this change — flagged for transparency. If CI flakes on them, a re-run clears it.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -2016,3 +2016,87 @@ fn a_journal_that_fails_its_own_revalidation_names_the_rebuild_route() {
         "the refusal must name the recovery route, not just the invariant: {rendered}"
     );
 }
+
+/// #589: `sweep_excluded_groups` actually deletes an excluded file's published
+/// documents (by `ax_file`, across the plan's dataset indices) and its catalog
+/// document (by `path`), while leaving every OTHER file's documents intact.
+/// Exercised over the real mock HTTP backend so the delete-by-query truly runs.
+#[test]
+fn sweep_excluded_groups_deletes_only_the_excluded_files_documents() {
+    let ep = HttpEndpoint::start();
+
+    // Seed two files' records in one dataset index, plus their catalog docs.
+    {
+        let mut st = ep.state.lock().unwrap();
+        let ds = "incremental-http-ds".to_string();
+        st.docs.insert(
+            (ds.clone(), "d1".into()),
+            json!({"ax_file": "key-excluded", "body": "secret"}),
+        );
+        st.docs.insert(
+            (ds.clone(), "d2".into()),
+            json!({"ax_file": "key-kept", "body": "kept"}),
+        );
+        st.docs.insert(
+            (catalog::CATALOG_INDEX.to_string(), "file:excluded".into()),
+            json!({"doc_kind": "file", "path": "secret.csv"}),
+        );
+        st.docs.insert(
+            (catalog::CATALOG_INDEX.to_string(), "file:kept".into()),
+            json!({"doc_kind": "file", "path": "kept.csv"}),
+        );
+    }
+
+    let es = Es::with_bulk_timeout(&ep.url, None, 30).expect("es client");
+
+    let mut plan = Plan::default();
+    plan.datasets.push(crate::state::PlanDataset {
+        slug: "ds".into(),
+        index: "incremental-http-ds".into(),
+        family: "csv".into(),
+        group: None,
+        specs: Vec::new(),
+        time_field: None,
+        semantic_field: None,
+        sampled_records: 1,
+        file_count: 1,
+    });
+
+    let excluded = vec![InventoryDeltaEntry {
+        file_key: "key-excluded".into(),
+        path: "secret.csv".into(),
+    }];
+
+    sweep_excluded_groups(&es, &plan, &excluded).expect("sweep");
+
+    let st = ep.state.lock().unwrap();
+    // The excluded file's dataset record is gone.
+    assert!(
+        !st.docs
+            .values()
+            .any(|d| d.get("ax_file").and_then(Value::as_str) == Some("key-excluded")),
+        "#589: the excluded file's indexed records must be swept"
+    );
+    // Its catalog document is gone.
+    assert!(
+        !st.docs
+            .iter()
+            .any(|((idx, _), d)| idx == catalog::CATALOG_INDEX
+                && d.get("path").and_then(Value::as_str) == Some("secret.csv")),
+        "#589: the excluded file's catalog document must be swept"
+    );
+    // Every other file is untouched (no over-deletion).
+    assert!(
+        st.docs
+            .values()
+            .any(|d| d.get("ax_file").and_then(Value::as_str) == Some("key-kept")),
+        "#589: a different file's records must NOT be swept"
+    );
+    assert!(
+        st.docs
+            .iter()
+            .any(|((idx, _), d)| idx == catalog::CATALOG_INDEX
+                && d.get("path").and_then(Value::as_str) == Some("kept.csv")),
+        "#589: a different file's catalog document must NOT be swept"
+    );
+}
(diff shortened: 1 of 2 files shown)


xerj-org:

Each skeptic ran in its own git worktree at PR head `350c0f4e`, reset to `origin/fix/589-exclusion-sweep`, under **rustc 1.97.1**, prompted to REFUTE.

**Result: 3× CONFIRMED, 0 BLOCKING findings.** (2 SHOULD-FIX → follow-ups filed, 2 NIT.)

The #589 sweep is correct on every load-bearing axis. `refuses()`=deletion-only is right and, because both gates check `refuses()`→return-Err BEFORE calling `sweep_excluded_groups`, a co-occurring genuine deletion refuses the whole rerun and mutates nothing (no partial mutation). Index resolution is NOT the suspected hole: sweep broadcasts `delete_by_query({term:{ax_file}})` to every `plan.datasets[].index`, a strict superset of the replacement-delete path's reach (which narrows to `fa.assignments`→`ds_rt`, itself built only from `plan.datasets`), so no content index is missed. Catalog delete uses an exact keyword-term on `path` (no over-deletion, test-confirmed). `!cfg.dry_run` guard present at both gates. `sweep_excluded` test passes.
- SHOULD-FIX: alias/duplicate catalog docs (`file-alias:...`) carry `path=<alias rel>`, so the `{term:{path=canonical rel}}` delete leaves them live when an excluded file had byte-identical duplicates — **path/filename metadata residue only** (content records keyed by `ax_file` are fully swept). → follow-up.
- NIT: the committed-generation incremental route (`lib.rs:3324`) bypasses the refuses()/sweep gates entirely (separate mechanism, project_reconcile_plan); out of scope, worth confirming it doesn't leave excluded docs live.

Synced to `350c0f4e`. All three fix symbols present. Full crate suite green: **707 passed / 0 failed** across all binaries on rustc 1.97.1, including the two known-flaky tests (which passed in the normal parallel run without needing isolation). `clippy --all-targets -D warnings` clean; `fmt --check` exits 0. Flipping `refuses()` to deletion-only orphaned no assertion (the excluded-only expectations were updated consistently; `into_error` still exercises the excluded wording via the direct message-builder check). Could not refute — nothing else breaks.

Evidentiary claims hold. **Fail-before reproduces**: reverting only the `refuses()` hunk to `|| !self.excluded_content_groups.is_empty()` makes `an_excluded_only_delta_is_swept_not_refused` FAIL, and it passes once restored. The integration test drives a REAL HTTP `delete_by_query` against the mock ES's `delete_http` (retains/removes by index+term, not a stub) and asserts the excluded file's dataset record AND catalog doc are gone WHILE a different file survives — real over-deletion protection. `into_error` unchanged; the refuses()-return precedes the guarded sweep at both gates, so a co-occurring deletion refuses the whole rerun and both sets stay live.
- SHOULD-FIX: the "no per-file edge provenance" framing (task summary) is inaccurate — `detect::invalidate_prior_edges` queries `{term:{src_file: rel}}`, so per-file edge provenance DOES exist and the deferred edge-sweep could reuse it. The PR's OWN code comment (`lib.rs:3068`) and `edges_note` are accurate (edges stay live as a documented follow-up); PR body corrected. → follow-up.

**Merge gate:** 0 BLOCKING → eligible once all checks green + `mergeStateStatus` CLEAN. The current Landing guard is the tag-ahead rc.62 stamp; it clears via a landing-only rebase onto post-rc.62-stamp main, which preserves these verdicts.

Issue #585, opened 2026-08-22
autoindex: cross-path graph re-run still proceeds over a NOT-YET-committed --no-graph state dir (pending-sync / pending-genesis)


xerj-org:
Follow-up to #490 (fix #3) / PR #584. The new guard in `run_index_report` refuses a graph-path re-run of a `--no-graph`-built corpus **once it has a committed generation** (`!cfg.no_graph && !genesis_recovery && committed_manifest.is_some()`). Two independent adversarial-verification passes on #584 flagged two *not-yet-committed* states the guard does not cover — both real, both out of #490 fix #3's committed-corpus scope, so tracked here rather than blocking that merge.

## 1. Pending sync (uncommitted generation)
`lib.rs` — the `pending_sync` resume branch (around the `preflight.pending_sync.is_some()` handling, ~line 3043) fires **regardless of `cfg.no_graph`**. If a `--no-graph` run was interrupted mid-generation (a `sync_begin` without `sync_commit`, leaving `pending_sync = Some`), a subsequent re-run **on the default graph path** resumes and commits that no-graph generation before control ever reaches the new committed-corpus guard. The user asked for the graph path and silently got a no-graph resume-and-commit.

## 2. Pending genesis bootstrap (generation 0, empty groups)
A `--no-graph` corpus in the genesis-bootstrap state has `generation == 0 && groups.is_empty()`, which makes `genesis_recovery` true (`lib.rs:2968-2971`). The new guard is `!genesis_recovery`-gated (correctly, to not break legitimate genesis recovery), so a graph re-run over a no-graph genesis-bootstrap state dir skips the guard and proceeds down the graph path.

## Why they were excluded from #584
#490's suggested fix #3 is specifically about the *committed* cross-path case (its matrix row: "`--no-graph`, then graph re-run … mutates the destination (79 → 80)"). PR #584 covers exactly that. Widening the guard to the pending states needs its own reproduction and a decision on the intended behavior (refuse vs. resume-in-original-mode), so it is filed separately rather than scope-creeping the targeted fix.

## Suggested shape
- Decide the contract for a graph-flagged re-run over a no-graph state dir that is pending (sync or genesis): most likely refuse with the same clear cross-path message #584 introduced, symmetric to the committed case, rather than silently resuming in the other mode.
[...]

PR #692, opened 2026-08-24
fix(autoindex): #585 case 1 — refuse a pending --no-graph generation resumed on the graph path

xerj-org:
Addresses #585 (case 1 — the pending-sync gap; case 2 is decision-blocked, see below).

## Problem (data-loss / cross-path authority)
Follow-up to #490 fix#3 / #584. #584 refuses a **graph** re-run of a `--no-graph`-built **committed** corpus. But the `pending_sync` resume branch (`lib.rs`, `if preflight.pending_sync.is_some()`) replays and **commits** the pending generation **regardless of `cfg.no_graph`**, and it returns before the committed-manifest guard ever runs. So a `--no-graph` generation interrupted mid-commit (a `sync_begin` without `sync_commit`), re-run on the default graph path, is silently **resumed and committed** under a graph authority — "the user asked for the graph path and silently got a no-graph resume-and-commit" (#585's headline).

## Fix
At the top of the `pending_sync` resume branch — **before** the journal is opened for write and before `gc_snapshots` — compare the pending generation's mode (`pending.desired.execution.graph_enabled`) against the current run's mode (`cfg.no_graph`). On mismatch, refuse with a clear, actionable message (mirroring #584): re-run in the pending generation's own mode to finish it, or rebuild with a new `--state-dir`/`--prefix`. Nothing is mutated. (A pending generation is always incremental, so its `graph_enabled` is `false`; the guard therefore refuses exactly the graph re-run of a pending no-graph generation, and lets a same-mode `--no-graph` resume proceed.)

## Case 2 (genesis) deliberately NOT included
The issue's second case — a graph re-run over a `--no-graph` **genesis-bootstrap** state (gen 0, empty groups, `genesis_recovery` true) — has **no clean mode source**: `sync_bootstrap_genesis` creates the gen-0 manifest via `CommittedManifest::genesis()`, which is **mode-agnostic** (records no `graph_enabled`). The issue itself flags case 2 as needing "a decision on the intended behavior (refuse vs. resume-in-original-mode)". With no committed content at genesis there is no data to corrupt, so it is left as a documented follow-up rather than guessing a mode signal.

- **Fail-before proven**: neutralizing only the guard makes the graph re-run **resume and commit** the pending no-graph generation (`run_index(graph).unwrap_err()` panics because it returns `Ok`) — exit 101.
- New integration test `pending_no_graph_generation_is_refused_on_the_graph_path`: interrupts a `--no-graph` generation after `sync_begin` (data bulk fails → pending sync), re-runs on the graph path, asserts it is refused with the clear message, commits no second generation, and mutates nothing further.
- Full gate: `fmt --all --check` / `clippy --workspace --all-targets -D warnings` / `build --workspace --profile ci-test` / `test -p xerj-autoindex` (full) — all green.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -2100,3 +2100,59 @@ fn sweep_excluded_groups_deletes_only_the_excluded_files_documents() {
         "#589: a different file's catalog document must NOT be swept"
     );
 }
+
+/// #585 (case 1): a pending (uncommitted) `--no-graph` generation must NOT be
+/// resumed and committed on the default graph path — doing so silently commits
+/// a no-graph generation under a graph authority (the #584 hazard, but for a
+/// not-yet-committed generation the committed-manifest guard never sees it
+/// because the pending-sync branch returns first). The re-run must be refused
+/// before any mutation, symmetric to #584.
+#[test]
+fn pending_no_graph_generation_is_refused_on_the_graph_path() {
+    let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
+    let _replay_guard = sync_executor::REPLAY_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let corpus = tempfile::tempdir().unwrap();
+    let state_dir = tempfile::tempdir().unwrap();
+    let source = corpus.path().join("rows.csv");
+    fs::write(&source, "id,value\n1,first\n").unwrap();
+    let endpoint = HttpEndpoint::start();
+
+    // Commit generation 1 with the `--no-graph` generated executor.
+    let no_graph = cfg(corpus.path(), state_dir.path(), &endpoint.url, false);
+    assert_eq!(run_index(no_graph.clone()).unwrap(), 0);
+
+    // Interrupt a second `--no-graph` generation after sync_begin: the data bulk
+    // fails, leaving a pending sync (no-graph, uncommitted).
+    fs::write(&source, "id,value\n1,first\n2,second\n").unwrap();
+    endpoint.state.lock().unwrap().fail_next_data_bulk = true;
+    assert!(run_index(no_graph.clone()).is_err());
+    assert_eq!(journal_events(state_dir.path(), "sync_begin"), 2);
+    assert_eq!(journal_events(state_dir.path(), "sync_commit"), 1);
+    // The destination as the refused re-run will find it (the interrupted run's
+    // delete-before-replace already mutated it — irrelevant to the guard).
+    let docs_before_rerun = endpoint.data_docs().len();
+
+    // Re-run the SAME state dir on the default graph path. The pending no-graph
+    // generation must be refused, not resumed-and-committed.
+    let mut graph = no_graph.clone();
+    graph.no_graph = false;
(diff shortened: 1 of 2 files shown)


xerj-org:

Each skeptic ran in its own git worktree at PR head `2856042a`, reset to `origin/fix/585-graph-crosspath-pending`, under **rustc 1.97.1**, prompted to REFUTE.

**Result: 3× CONFIRMED, 0 BLOCKING findings.** (all NITs.)

The mode-agreement gate is correct for case 1 as scoped. `exec.graph_enabled == cfg.no_graph` is sound because a pending sync's `desired.execution.graph_enabled` is an enforced invariant of `false`: `validate_manifest` (sync.rs:743-748) requires `!graph_enabled` for every gen≥1 manifest, enforced on write (`PendingSync::new`/`validate_against`) AND on journal load/replay (`begin.validate_against`, sync.rs:1051), so the guard can never misfire and never sees `graph_enabled=true`. It refuses exactly a graph re-run (`no_graph=false` → `false==false`) and permits a same-mode `--no-graph` resume (`no_graph=true` → `false==true`). The bail precedes any mutation (`open_after_preflight` at 3253, `gc_snapshots` at 3261). Case-2 genesis scope-out is sound: `CommittedManifest::genesis()` carries `execution=None`, is mode-agnostic, has no committed content, and is not a pending-sync path. Fail-before verified: neutralizing the guard makes the graph re-run return `Ok(0)`, resuming-and-committing the pending no-graph generation (the data-loss).
- NIT: the `execution==None` branch is a dead path (validate_manifest forces `execution.is_some()` for gen≥1) — a `debug_assert` would make the invariant explicit.
- NIT: case 2 (genesis) remains unguarded — deliberately out of scope (needs a behavior decision; no content to corrupt at genesis); the PR does not claim to close #585.

Reset to `origin/fix/585` (2856042a). Full `xerj-autoindex` suite green: **693 passed / 0 failed** in the main binary plus 8/1/1/4/0 across the others; the two known-flaky tests passed on this run. `fmt --all --check` exits 0; `clippy -p xerj-autoindex --all-targets -D warnings` clean. The new fail-before test passes AND the legitimate same-mode resume `pending_generation_replays_sealed_source_after_live_source_mutates` STILL passes — the guard fires only on a mode mismatch. Could not refute — nothing else breaks.

The gate and its fail-before both hold. The guard is at the TOP of the `pending_sync.is_some()` branch before `open_after_preflight`/`gc_snapshots` — refuses before any write, like #584. Neutralizing only the guard (`if false && …`) makes `pending_no_graph_generation_is_refused_on_the_graph_path` panic at the `unwrap_err()` on `Ok(0)` — the graph re-run resumes+commits the pending no-graph generation; restoring it passes. The test genuinely leaves a pending sync (`fail_next_data_bulk` after a 2nd `sync_begin`), asserts the graph re-run errors with "pending sync" + "different graph authority", `sync_commit` stays 1 (no 2nd generation), and nothing is mutated. Case-2 scope-out is honest: `validate_genesis` asserts `execution.is_none()`, so no `graph_enabled` mode signal exists at genesis — a clean genesis mode source does NOT exist.
- NIT: the polarity `exec.graph_enabled == cfg.no_graph` is robust in both directions even beyond the always-incremental premise — a strength, not a defect.

**Merge gate:** 0 BLOCKING → eligible once all checks green + `mergeStateStatus` CLEAN. This PR will be rebased onto post-rc.62 main (landing-only + resolving the test-file append) to clear the tag-ahead Landing guard; the code is unchanged so these verdicts hold.

PR #695, opened 2026-08-24
test(autoindex): #689 extract tcorr_id + prefix-scope unit test

### LINKED ISSUE #689: autoindex: tcorr: prefix-scoping lacks a unit test (coverage NIT from #688/#673)
Non-blocking coverage NIT surfaced by the 3-skeptic verification of #688 (#673).

#688 prefix-scoped both correlation id families — `corr:{prefix}:…` (`KeyCorr::id`) and `tcorr:{prefix}:…` (inline in `lib.rs`). The new test `correlate::tests::correlation_id_is_prefix_scoped_across_corpora` pins the **corr:** path, but the **tcorr:** id is built inline in `lib.rs` (~5527-5536) with no test asserting that two same-slug time-correlations from different corpora produce distinct `tcorr:` ids. A future refactor could silently reintroduce the collision on the tcorr path only.

**Fix direction:** either extract the `tcorr:` id into a small `fn tcorr_id(prefix, a_dataset, b_dataset)` (mirroring `KeyCorr::id`) and unit-test it the same way, or add an integration assertion through the generation flow. Low priority — the corr: path (the more common family) is covered, and both were verified correct by inspection in #688.

xerj-org:
Closes #689.

Coverage gap from #673: the `tcorr:` time-correlation catalog id was prefix-scoped (`tcorr:{prefix}:{a}:{b}`) but built inline in `run_index_report` with no test pinning the scoping — only `corr:`/`KeyCorr::id` was covered by `correlate::tests`. A future refactor could silently reintroduce the cross-corpus collision on the tcorr path.

## Change
- Extract the id into `tcorr_id(prefix, a_dataset, b_dataset)` — **byte-identical** output (`format!("tcorr:{prefix}:{a_dataset}:{b_dataset}")`), preserving the flush-parity contract. No behaviour change.
- Add `tcorr_id_is_prefix_scoped_across_corpora`: distinct ids across corpora (same slugs, different prefix), idempotent within one corpus, prefix literally in the id — mirroring the `corr:` test.

- `fmt --all --check` / `clippy -p xerj-autoindex --all-targets -D warnings` clean; full `test -p xerj-autoindex` green (693 passed / 0 failed).

(Pure test/refactor — no engine behaviour change; the `Landing constants guard` may show the tag-ahead rc.62 stamp until a landing-only rebase onto post-rc.62 main.)

### DIFF
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -3094,6 +3094,41 @@ fn sweep_excluded_groups(es: &Es, plan: &Plan, excluded: &[InventoryDeltaEntry])
     Ok(())
 }
 
+/// Catalog doc id for a time correlation, corpus-prefix-scoped (#673): the
+/// `autoindex-catalog` index is shared across corpora, so a time correlation
+/// keyed only by dataset slugs would overwrite the same-slug correlation from
+/// another corpus. Extracted (#689) so the prefix scoping is unit-testable, the
+/// way `KeyCorr::id` is for `corr:`.
+fn tcorr_id(prefix: &str, a_dataset: &str, b_dataset: &str) -> String {
+    format!("tcorr:{prefix}:{a_dataset}:{b_dataset}")
+}
+
+#[cfg(test)]
+mod tcorr_id_tests {
+    use super::tcorr_id;
+
+    /// #689 (coverage gap from #673): the shared `autoindex-catalog` index means
+    /// a time correlation keyed by dataset slugs must carry the corpus prefix, or
+    /// `(reports, orders)` in corpus A overwrites the same-slug correlation in
+    /// corpus B. `corr:` is pinned by `correlate::tests`; this pins `tcorr:`.
+    #[test]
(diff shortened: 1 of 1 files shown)


xerj-org:

Each skeptic ran in its own worktree at head `e3ffde25` under rustc 1.97.1.

- **byte-identity — CONFIRMED**: old `format!("tcorr:{}:{}:{}", cfg.prefix, a, b)` and new `tcorr_id` emit a byte-identical id (same literals, separators, arg order); the `b_dataset` `i.to_string()` fallback is hoisted but semantically identical. **XERJ_FLUSH_PARITY preserved.**
- **regression — CONFIRMED**: full crate suite green (694 lib + all integration binaries, 0 failed; both known-flaky tests passed on the default run); `fmt --check` exit 0; `clippy -D warnings` clean.
- **evidence-holds — CONFIRMED**: the new test asserts all three properties and is mutation-verified — removing the prefix makes `assert_ne!` FAIL (`left==right=="tcorr:reports:orders"`).

NITs only (idempotency assert is documentary; noted). 0 BLOCKING → eligible on green + CLEAN.

PR #696, opened 2026-08-24
fix(engine): #375 defer nested-terms sub-aggs to the surviving top-N buckets

### LINKED ISSUE #375: perf: nested term aggregations materialize all sub-buckets in the single pass, then discard almost all of them
Raised by **Pascal Seitz** (tantivy aggregations author), who gave this as a concrete case where single-pass aggregation does not hold up.

His example: `AllQuery` -> top 10 of `my_field_with_100k_terms` -> top 10 of `my_field_with_1k_terms`. In a single pass the nested term aggregation accumulates 100k buckets and eagerly fetches and materializes *all* sub-buckets, then drops nearly all of them at the end, because only the top 10 at the top level survive. It degrades further with each extra level of nesting.

Suggested shape: a two-pass approach that first identifies which 10 top-level buckets matter, then materializes sub-buckets only for those.

Worth confirming against our current aggregation path before sizing.

Thanks to Pascal for the specific repro shape.

xerj-org:
Closes #375.

## Problem (reported by Pascal Seitz, tantivy aggregations author)
A nested terms aggregation ordered by `_count`/`_key` — e.g. `top 10 of a 100k-term field -> top 10 of a 1k-term field` — eagerly materialised sub-aggregations for **every** candidate top-level bucket, then the size-truncation discarded nearly all of them (100k built, ~99,990 dropped). It degrades further per nesting level.

## Root cause
`run_terms` gated its sub-agg pre-compute on:
```rust
let pre_computed = if orders_need_sub_agg || sub_aggs.is_some() { /* compute for ALL candidates */ } else { /* defer */ };
```
The surrounding comment says "When the ordering only references `_count` or `_key`, skip the pre-compute" — but the **`|| sub_aggs.is_some()`** clause defeated exactly that: a nested terms agg *always* has `sub_aggs.is_some()`, so it always pre-computed for all candidates regardless of order.

## Fix
Drop the `|| sub_aggs.is_some()` clause. The pre-compute is now gated only on `orders_need_sub_agg` — i.e. when the requested order actually references a sub-agg metric (which genuinely requires computing all sub-aggs before truncation). When the order is `_count`/`_key` (independent of sub-aggs), the buckets are sorted + size-truncated first, and the bucket loop's existing `sub_result = precomputed_sub.or_else(|| run_aggs_in_bucket(...))` computes sub-aggs for the **surviving top-N only** — Pascal's two-pass shape.

**Correctness-neutral:** the sub-agg *values* are identical (the discarded buckets' sub-aggs were dropped anyway); the `orders_need_sub_agg` branch retains the eager pre-compute where ordering-before-truncation still needs it.

- **Fail-before proven**: a new test counts `run_aggs_in_bucket` evaluations for 1000 top-level terms with `size: 5` + a nested sub-agg. With the fix it runs ~5 times; restoring only the `|| sub_aggs.is_some()` clause makes it **1001** (the assertion fails, exit 101).
- New test `nested_terms_defers_subaggs_to_the_surviving_top_buckets` also asserts the result is unchanged (exactly the 5 requested top buckets, each still carrying its sub-agg).
- Full gate: `fmt --all --check` / `clippy --workspace --all-targets -D warnings` / `build --workspace --profile ci-test` / `test -p xerj-engine` (full) — all green; the 93 existing agg tests pass unchanged (correctness-neutral).

### DIFF
--- a/engine/crates/xerj-engine/src/aggs.rs
+++ b/engine/crates/xerj-engine/src/aggs.rs
@@ -376,6 +376,16 @@ pub fn run_aggs_with_all(aggs_def: &Value, docs: &[Value], all_docs: &[Value]) -
     run_aggs_in_bucket(aggs_def, docs, all_docs, docs.len() as u64)
 }
 
+// #375: counts how many buckets actually had their sub-aggregations
+// materialised, so a test can prove the nested-terms path defers sub-aggs to
+// the surviving top-N instead of computing them for every candidate bucket.
+// Thread-local (not a global atomic) so parallel agg tests don't pollute each
+// other's count — the brute agg path runs synchronously on the caller thread.
+#[cfg(test)]
+thread_local! {
+    pub(crate) static SUBAGG_BUCKET_EVALS: std::cell::Cell<usize> = const { std::cell::Cell::new(0) };
+}
+
 /// `run_aggs_with_all` for the sub-aggs of a single bucket.
 ///
 /// `bucket_doc_count` is the value the CALLER puts in that bucket's
@@ -387,6 +397,8 @@ pub(crate) fn run_aggs_in_bucket(
     all_docs: &[Value],
     bucket_doc_count: u64,
 ) -> Value {
+    #[cfg(test)]
+    SUBAGG_BUCKET_EVALS.with(|c| c.set(c.get() + 1));
     let obj = match aggs_def.as_object() {
         Some(o) => o,
         None => return Value::Object(Map::new()),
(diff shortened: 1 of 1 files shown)


xerj-org:

Each skeptic ran in its own worktree at head `6862c102` under rustc 1.97.1.

- **correctness — CONFIRMED**: dropping `|| sub_aggs.is_some()` is correctness-neutral for `_count`/`_key` ordering. The eager `compute_bucket_docs(k)` and the deferred inline `bucket_docs` filter are identical logic (same docs slice, terms keys, missing-bucket handling, membership test), so a surviving bucket's sub-agg value is identical. `cmp_terms_by_orders` reads `sub_res` ONLY in its non-`_count`/`_key` branch, and `orders_need_sub_agg` gates exactly that case back onto the eager path — so deferring to `None` never affects sort/truncate. All 94 agg lib tests pass, incl. the new counter test (~5 evals vs 1000).
- **regression — CONFIRMED**: full `xerj-engine` suite green — 627 lib + every integration suite, 0 failed — including the new test in the **parallel** run (the `SUBAGG_BUCKET_EVALS` thread-local `Cell` is per-thread, correctly isolated; a global atomic would have flaked). `clippy -p xerj-engine --all-targets -D warnings` clean (covers the `#[cfg(test)]` code `--lib` misses); `fmt --check` exit 0.
- **evidence-holds — CONFIRMED**: fail-before reproduces (restoring the clause → sub-aggs for 1001 buckets vs asserted ≤10; fixed → passes). The test asserts BOTH perf (evals≤10) and correctness (5 top buckets, `buckets[0]` carrying its `by_sub` array). Fast path (`fast_aggs.rs`) genuinely unaffected — it bails to brute on a sub-agg order and truncates before building buckets.

**NITs** (non-blocking): fixed-path count is actually 5 (test bound ≤10 has headroom); fail-before counter reads 1001 (narrative "1000"); the deferred loop computes `bucket_docs` unconditionally even for the eager branch (pre-existing inefficiency, zero result effect).

**Merge gate:** 0 BLOCKING → eligible on green + CLEAN. Current `Landing constants guard` failure is the tag-ahead rc.63 stamp; clears via a landing-only rebase onto post-rc.63 main, preserving these verdicts.

PR #698, opened 2026-08-24
fix(engine): #421 item 1 — meta-aware memtable pre-clone rejection (restore O(1) for _seq_no/_version sorts)

xerj-org:
Addresses #421 item 1 (the perf item; items 2 & 3 are smaller doc/comment follow-ons, not touched here).

## Problem (reported in #421's adversarial verification of #420)
Gating `memtable_primary_key_rejects` on ES meta-fields disabled the memtable **pre-clone rejection** for a `_seq_no`/`_version` sort — so every buffered document paid `(*source).clone()` + `seen_ids.insert` + a full `offer`. Measured (issue's numbers): ~1.1 µs/buffered doc → **~130 ms at the 128k stock flush threshold, ~1.1 s at 1M docs**, on the `search_after`/reindex workload #401 was filed for.

The decline was there for a reason: deriving a meta value from `_source` lets a literal `_seq_no` inside a document **shadow** the real metadata the heap ranks on.

## Fix
Make the rejection **meta-aware** instead of declining: for `_seq_no`/`_version` derive the key from `Index::meta_sort_value` — the *same* version-map resolver `compute_sort_values` (and therefore the heap frontier `hit.sort[0]`) uses — so the candidate key and the frontier key are apples-to-apples. Correct by construction, avoids the `_source`-shadowing hazard, and restores the O(1)-per-rejected-doc bound. `_index` (string) / `_primary_term` (constant) / unresolvable (`Null`) yield `None` from `as_f64()` and are conservatively admitted. `memtable_primary_key_rejects` now takes the doc `id` + `&Index`, threaded at both call sites.

- **Correctness-neutral**: the full `xerj-engine` suite is green, including the dedicated `sort_on_meta_fields.rs` (#401) regression suite (pages byte-identical) — the meta-aware key equals what the heap already ranked on.
- **Fail-before proven**: new unit test `memtable_preclone_rejection_is_meta_aware_for_seq_no` builds a size-5 `_seq_no` frontier and asserts `memtable_primary_key_rejects` rejects a strictly-worse doc (and admits the best). Reverting only the meta branch (restoring the decline) makes it fail (exit 101) — the declined gate returns `false`. (A plain bounded sort narrows to the page and never reaches this gate; it fires on the brute-walk/`search_after` arm, so the fix is unit-tested directly rather than through a narrowed search.)
- Full gate: `fmt --all --check` / `clippy --workspace --all-targets -D warnings` / `build --workspace --profile ci-test` / `test -p xerj-engine` (full) — all green.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -282,6 +282,63 @@ mod collection_publication_fail_closed_tests {
         );
     }
 
+    /// #421: the memtable pre-clone rejection must be META-AWARE. A `_seq_no`
+    /// sort resolves its key from the version map (`meta_sort_value`), the same
+    /// resolver the heap ranks on, so `memtable_primary_key_rejects` can and
+    /// must reject a doc whose `_seq_no` is strictly worse than the bounded
+    /// frontier — skipping its `_source` clone. Before the fix the gate
+    /// declined every meta-sort field and returned `false` for this exact call,
+    /// forcing the full O(memtable) clone-and-offer walk (#401's `search_after`
+    /// latency cliff).
+    #[tokio::test]
+    async fn memtable_preclone_rejection_is_meta_aware_for_seq_no() {
+        use std::sync::Arc;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine.create_index("seq-reject", text_schema()).unwrap();
+        let idx = engine.get_index("seq-reject").unwrap();
+        idx.abort_background_tasks();
+        for i in 0..6u64 {
+            idx.index_document(Some(format!("d{i}")), json!({ "body": format!("doc {i}") }))
+                .await
+                .unwrap();
+        }
+
+        // Hits sorted by `_seq_no` asc: d0..d5 with strictly increasing seq_no.
+        let req = xerj_query::parse_request(&json!({
+            "query": {"match_all": {}},
+            "size": 6,
+            "sort": [{"_seq_no": "asc"}],
+        }))
+        .unwrap();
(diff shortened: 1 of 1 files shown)


xerj-org:

Each skeptic ran in its own worktree at head `12754abb` under rustc 1.97.1.

- **correctness — CONFIRMED**: result-neutral (drops no valid doc, admits no wrong one). The candidate key is `idx.meta_sort_value(&sf.field, id)` — the *identical* resolver `compute_sort_values` uses to populate the frontier `hit.sort[0]` (same field, id, version-map lookup, same read-admission snapshot with writers blocked), so the pre-clone comparison ranks on exactly the heap's data → false-rejection impossible. `primary_f64_rejects` rejects only when the heap is full AND the candidate primary is STRICTLY worse — a sound subset of `would_admit`. `_source`-shadowing avoided (meta routes through the version map, never `obj.get`), confirmed by the `shadowing_source_key_*` tests. Ties (`_version`) and `None`/tombstone (`Null` → `as_f64` None → admit) handled conservatively. All 7 `sort_on_meta_fields` tests + the new unit test pass.
- **regression — CONFIRMED**: full `xerj-engine` suite **1077 passed / 0 failed** (new test passes in the parallel run + isolation). The `id`/`idx` signature change is correctly threaded at BOTH call sites. `clippy --workspace --all-targets -D warnings` zero warnings; `fmt --all --check` exit 0.
- **evidence-holds — CONFIRMED**: fail-before reproduces — reverting only the meta branch to a bare `return false` makes the test FAIL at index.rs:329 (the worse d5 not rejected); restoring passes. Test is meaningful (size-5 frontier genuinely full, d5 strictly worse — both reject-worse and admit-best assertions non-trivial). Perf honest: one `meta_sort_value` get + one f64 compare, `_source` clone provably skipped on rejection at both sites → O(1)-per-rejected-doc.

**All NITs** (positive confirmations, non-blocking). **Merge gate:** 0 BLOCKING → eligible on green + CLEAN.

PR #701, opened 2026-08-24
fix(engine): filtered kNN evaluates its filter schema-aware (#423)

xerj-org:
## What

A filtered `knn` query evaluated its `filter` through the **schemaless** matcher
(`doc_matches_query`) inside `run_knn_brute_force_with_deadline`, while the rest
of search evaluates the identical filter through the **schema-aware** path
(`search_inner` runs its filter through `doc_matches_query_typed(.., &dmq_schema)`).
Without field types the matcher folds case on `keyword`/`text` fields and
**over-matches**, so a `prefix`/`match`/`fuzzy` filter inside a knn kept documents
a plain (non-knn) search would reject.

This is not an edge case: a `knn` that carries a `filter` **always** routes to
this brute-force path — the dispatch serves HNSW only when `filter.is_none()`
(*"filters remain exact scans so approximation cannot change filter membership"*),
so every filtered kNN went through the schemaless matcher, reachable straight from
the public search API.

This is the first landed slice of #423 (the wrong-results umbrella: term matching
has two implementations, one schemaless). It fixes the two filter call sites in
`run_knn_brute_force_with_deadline`.

## Fix

Bind the index schema once at the top of `run_knn_brute_force_with_deadline`
(`let dmq_schema = self.schema().await;` — the exact idiom `search_inner` already
uses; an owned clone, lock released) and route both filter call sites through
`doc_matches_query_typed(.., &dmq_schema)`:

- the default f32 candidate loop, and
- the `use_sq8` (scalar8-quantised) candidate loop.

Correct by construction: it makes the filtered-kNN filter use the **same** matcher
the rest of search already uses, so a filter now returns identical membership
inside and outside a knn.

## Fail-before (both sites)

Field `tag` is a declared `keyword`; a keyword `prefix` is case-SENSITIVE in ES
(`"Hello"` does not start with lowercase `"hel"`). New tests in
`tests/integration.rs`:

- `knn_filter_prefix_on_keyword_is_case_sensitive_like_plain_search` (default f32
  loop). Includes a control asserting the same `prefix` filter run as a **plain**
  query already returns 0 hits, so the knn must agree.
- `knn_filter_prefix_on_keyword_is_case_sensitive_for_scalar8` (the `use_sq8`
  loop; `quantization = scalar8`).

With the source fix stashed, **both** tests fail — the schemaless matcher folds
case and wrongly keeps `d0` (1 hit vs 0). With the fix, both pass.

## Scope / follow-ups

Deliberately scoped to `run_knn_brute_force_with_deadline`. The sibling
`run_nested_knn_brute_force_with_deadline` (nested-kNN pre/post filters) and the
remaining non-kNN schemaless callers (`try_shortcut_count`,
`hydrate_prefiltered_unsorted`, `scan_stored_section_into`) carry the same
root-cause pattern and will land as separate slices, each with its own
fail-before. #423 stays open until they do.

- Full `xerj-engine` test suite — green

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -11076,6 +11076,16 @@ impl Index {
             tracing::info!(index=%self.name, field, k, "semantic_phase=start_brute");
         }
 
+        // #423: evaluate the kNN `filter` through the SAME schema-aware matcher
+        // the rest of search uses (`search_inner` runs its filter through
+        // `doc_matches_query_typed(.., &dmq_schema)`). The schemaless
+        // `doc_matches_query` has no field types, so on a `keyword`/`text` field
+        // it folds case and OVER-MATCHES — a `prefix`/`match`/`fuzzy` filter kept
+        // documents a plain (non-knn) search would reject. A filtered knn ALWAYS
+        // takes this brute-force path (HNSW serves only `filter.is_none()`), so
+        // this is the whole filtered-kNN path. Owned clone; the lock is released.
+        let dmq_schema = self.schema().await;
+
         // ── Collect all candidate (doc_id, source) pairs ──────────────
         let mut candidates: Vec<(String, Value)> = Vec::new();
         // Memtable first (newest writes).
@@ -11213,7 +11223,7 @@ impl Index {
                     if let Some(obj) = src_with_id.as_object_mut() {
                         obj.insert("_id".to_string(), Value::String(id.clone()));
                     }
-                    if !doc_matches_query(f, &src_with_id) {
+                    if !doc_matches_query_typed(f, &src_with_id, &dmq_schema) {
                         continue;
                     }
                 }
@@ -11331,7 +11341,7 @@ impl Index {
                     if let Some(obj) = src_with_id.as_object_mut() {
                         obj.insert("_id".to_string(), Value::String(id.clone()));
                     }
-                    if !doc_matches_query(f, &src_with_id) {
+                    if !doc_matches_query_typed(f, &src_with_id, &dmq_schema) {
                         continue;
                     }
                 }
--- a/engine/crates/xerj-engine/tests/integration.rs
+++ b/engine/crates/xerj-engine/tests/integration.rs
@@ -8653,3 +8653,146 @@ async fn a_refusal_is_not_acknowledged_while_the_sidecar_is_unreadable() {
         "and nothing may be left behind that only this process can see"
     );
 }
+
+/// #423 (schemaless-matcher root cause, kNN slice): a filtered `knn` applies
+/// its `filter` through the SCHEMALESS matcher inside
+/// `run_knn_brute_force_with_deadline` (`doc_matches_query`), whereas the rest
+/// of search evaluates the same filter through the schema-aware path
+/// (`search_inner`'s `doc_matches_query_typed(.., &dmq_schema)`). A `knn` that
+/// carries a `filter` ALWAYS routes to brute force — the dispatch serves HNSW
+/// only when `filter.is_none()`, so "filters remain exact scans" — hence this
+/// is the WHOLE filtered-kNN path, reachable from the public search API, not an
+/// edge case.
+///
+/// Divergence: `tag` is a declared `keyword`, so a keyword `prefix` is
+/// case-SENSITIVE in ES ("Hello" does not start with lowercase "hel"). The
+/// schemaless matcher has no schema, folds case, and wrongly keeps the doc.
+/// Fail-before (schemaless): 1 hit. After (schema-aware): 0 hits — consistent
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent 3-skeptic adversarial review (each REFUTES in its own worktree, rustc 1.97.1, head `8a5791b2`)

**Result: 3/3 HOLDS — ZERO BLOCKING findings.** One NON-BLOCKING note (nested-kNN path), addressed below.

### Correctness lens — **HOLDS** (0 findings)
> HEAD=8a5791b2 as expected. Fix present: dmq_schema bound at index.rs:11087 in run_knn_brute_force_with_deadline; doc_matches_query_typed(f,&src_with_id,&dmq_schema) at both call sites (11226 scalar8/use_sq8 loop, 11344 default f32 loop). CORRECTNESS lens all clear: (1) Right schema — self.schema().await clones ManagedSchema.schema and drops the read guard; search_inner binds dmq_schema identically at 14212, same snapshot-once semantics. (2) No deadlock/race — the bind is an owned clone, guard released before the loop's separate sequential self.schema.read().await at 11199; no nesting. (3) Non-analyzed filters unchanged — within doc_matches_query_typed only Prefix/Wildcard/Fuzzy consult schema; Term/Terms/Range/Exists/Match/MatchPhrase/Geo*/Regexp/Ids/Script/Bool are schema-independent, so term/range/geo/exists filters are byte-identical to the old EMPTY_SCHEMA call. (4) No incorrect misses — for the three schema-dependent arms a declared keyword becomes case-SENSITIVE (fewer, ES-correct) and text tokenizes/folds, exactly matching search_inner; object arms mirror the #413 blob. Since it is the same matcher + same schema + same _source-shaped candidate, filtered kNN cannot diverge from plain search; the change is a consistency fix, not a result regression. Fail-before PROVEN on BOTH call sites: reverting both to schemaless doc_matches_query makes both new tests fail (1 hit, d0 'Hello' kept by case-folding); restoring the committed fix yields 0 hits/pass. Compiled and executed under rustup run 1.97.1. Working tree restored clean.

### Evidence lens — **HOLDS** (0 findings)
> EVIDENCE lens verified end-to-end. At HEAD both tests pass. Fail-before: I reverted ONLY the two call sites back to schemaless doc_matches_query (kept the tests), and both tests failed with exactly the predicted symptom — left: 1, right: 0, "wrongly kept: [d0]". Restored; worktree clean. False-positive checks all pass: (1) the control assertion (plain prefix:hel == 0) held even in fail-before, so plain keyword search was already case-sensitive and ONLY the kNN filter path diverged — the divergence is correctly isolated. (2) d0 was actually returned (1 hit) in fail-before, proving it is a scored candidate that reaches the filter; the 0-hit result at HEAD is genuinely the filter rejecting d0, not d0 being absent — neither test is a vacuous pass. (3) Attribution to distinct call sites is real: the scalar8 test sets quantization="scalar8" so use_sq8=true → 11226 site; the default test leaves it unset → 11344 site. Both distinct sites exercised. No BLOCKING or NON-BLOCKING findings survive.

### Regression lens — **HOLDS** (1 NON-BLOCKING)
> Both filter sites gated on `if let Some(ref f) = filter`; no schemaless doc_matches_query filter site remains inside the function. Term filters unaffected (json_values_equal regardless of schema), so test_knn_filter_similarity_boost_and_multi_clause is unchanged. Unfiltered kNN still uses HNSW (dispatch gate untouched); brute-force runs with filter=None only as an HNSW-miss fallback where the filter block is skipped. Perf: schema() clone is negligible against a full brute-force scan. The two new tests are well-formed; fail-before is real. Caveat: the full `cargo test -p xerj-engine` was still compiling when I finalized, so 0-failures rests on static analysis.
>
> **NON-BLOCKING finding:** the fix does not cover the SECOND kNN brute-force path (`run_nested_knn_brute_force_with_deadline`, pre_filter@12600 / post_filter@12656 still schemaless). Pre-existing, untouched; PR self-scopes as "first landed slice", so NON-BLOCKING — but the commit-message phrase "whole filtered-kNN surface" is inaccurate for the nested path.

---

**Maintainer disposition:** merging. 0 BLOCKING across all three lenses; #701 is green (18/18) + CLEAN.
- The NON-BLOCKING nested-kNN gap is **already fixed in #702** (`fix/423-nested-knn-filter-schema-aware`, both pre_filter@12600 + post_filter@12656 routed to `doc_matches_query_typed`, both proven fail-before). So the kNN filter surface is covered across #701 + #702.
- The regression lens's empirical-suite caveat is covered independently: the full `xerj-engine` suite was run green locally (CARGO-EXIT=0, 128 passed) and #701's CI "Build + Test" is green.
- Noted: the commit message's "whole filtered-kNN surface" over-reaches slightly (nested is the sibling #702); not amending (message-only, would restart CI) — #423 remains open until all slices land.

PR #702, opened 2026-08-24
fix(engine): nested kNN pre-filter evaluates schema-aware (#423)

xerj-org:
## What

Second slice of #423 (schemaless-matcher root cause). `run_nested_knn_brute_force_with_deadline` applied **both** of its filter sites through the schemaless `doc_matches_query`, so keyword/text filters folded case and **over-matched** — inconsistent with the schema-aware path the rest of search uses (`doc_matches_query_typed(.., &dmq_schema)`). Nested kNN always takes this brute-force path (there is no nested HNSW graph).

Sibling to #701 (the non-nested `run_knn_brute_force_with_deadline` slice).

## Fix

Bind the schema once at the top of the function (the `search_inner` idiom; owned clone, lock released) and route **both** filter call sites through `doc_matches_query_typed`:

- **pre-filter** — a nested knn's own `knn.filter`, matched at parent-doc granularity before scoring;
- **post-filter** — a sibling clause in the `bool` that wraps the nested knn (`bool{ must:[ nested{query:knn}, <sibling> ] }`), applied after `num_candidates` truncation.

## Fail-before (both sites)

New `tests/integration.rs` tests, both proven by stashing only the source hunk (each wrongly keeps `d0` unfixed, passes fixed):

- `nested_knn_prefilter_prefix_on_keyword_is_case_sensitive` — a true `nested{ path, query: knn{ .., filter: prefix{tag:"hel"} } }` (a bare dotted-field knn goes to the *non-nested* path, so the explicit `nested{}` wrapper is required to reach this executor).
- `nested_knn_postfilter_prefix_on_keyword_is_case_sensitive` — `bool{ must:[ nested{query:knn}, prefix{tag:"hel"} ] }`; the sibling must be next to a `nested{}`-wrapped knn (a bare knn sibling is handled by the already-schema-aware main bool path).

`tag` is a declared `keyword`, so `prefix:hel` is case-sensitive and matches no case-preserved term.

## Scope

Completes the nested-kNN filter surface. Remaining #423 slices (the non-kNN schemaless callers: `try_shortcut_count`, `hydrate_prefiltered_unsorted`, `scan_stored_section_into`) land separately; #423 stays open.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -12500,6 +12500,16 @@ impl Index {
     ) -> Result<SearchResult> {
         let started = std::time::Instant::now();
         let mut timed_out = false;
+        // #423: evaluate the nested-kNN pre/post filters through the SAME
+        // schema-aware matcher the rest of search uses (`search_inner` runs its
+        // filter through `doc_matches_query_typed(.., &dmq_schema)`). The
+        // schemaless `doc_matches_query` has no field types, so on a
+        // `keyword`/`text` field it folds case and OVER-MATCHES — a
+        // `prefix`/`match`/`fuzzy` pre/post filter kept parent docs a plain
+        // (non-knn) search would reject. Nested kNN always takes this
+        // brute-force path (there is no nested HNSW graph). Owned clone; the
+        // lock is released.
+        let dmq_schema = self.schema().await;
         // The sub-field name inside each nested element.
         let subfield = field
             .strip_prefix(&format!("{}.", nested_path))
@@ -12597,7 +12607,7 @@ impl Index {
                 if let Some(obj) = src_with_id.as_object_mut() {
                     obj.insert("_id".to_string(), Value::String(id.clone()));
                 }
-                if !doc_matches_query(f, &src_with_id) {
+                if !doc_matches_query_typed(f, &src_with_id, &dmq_schema) {
                     continue;
                 }
             }
@@ -12653,7 +12663,7 @@ impl Index {
                 if let Some(obj) = src_with_id.as_object_mut() {
                     obj.insert("_id".to_string(), Value::String(id.clone()));
                 }
-                doc_matches_query(f, &src_with_id)
+                doc_matches_query_typed(f, &src_with_id, &dmq_schema)
             });
         }
 
--- a/engine/crates/xerj-engine/tests/integration.rs
+++ b/engine/crates/xerj-engine/tests/integration.rs
@@ -8796,3 +8796,137 @@ async fn knn_filter_prefix_on_keyword_is_case_sensitive_for_scalar8() {
         knn.hits.iter().map(|h| &h.id).collect::<Vec<_>>()
     );
 }
+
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent 3-skeptic adversarial review (each REFUTES in its own worktree, rustc 1.97.1, head `01be9546`)

**Result: 3/3 HOLDS — ZERO findings (no BLOCKING, no NON-BLOCKING).**

### Correctness lens — **HOLDS** (0 findings)
> Fix present (dmq_schema bound at index.rs:12512; typed-call count == 4: 2 from merged #701 at 11226/11344, 2 nested this PR at 12610 pre-filter / 12666 post-filter). (1) Right schema, no deadlock: self.schema().await = read().await.schema.clone(), owned clone, guard dropped — identical to search_inner/#701; only read, bound once at top. (2) term/range/exists unaffected: only Prefix/Wildcard/text branches consult schema; Term/Terms/Range/Exists never do, so typed-with-real-schema is byte-for-byte identical to the old schemaless call for those types — only declared keyword/text prefix/wildcard/match/fuzzy become stricter (the intended fix). (3) post_filter placement preserved: scored.truncate(num_candidates) still precedes the retain; matcher not consulted during truncation, so which parents survive is unchanged. (4) peel_nested_knn_query untouched. Fail-before PROVEN: reverting exactly the two nested sites makes both new tests fail (left:1 right:0); restored, both pass.

### Regression lens — **HOLDS** (0 findings)
> Full suite `cargo test -p xerj-engine` under 1.97.1: every result line 0 failed — all knn/nested-knn/semantic tests green incl. merged #701 non-nested + nested_knn_uses_the_shared_request_deadline. Fail-before verified (both nested sites reverted → both tests fail 'wrongly kept: [d0]'; restored → 4 typed sites, pass). Edges: undeclared/schemaless field preserves the pre-#423 fold-both heuristic (unchanged for schemaless indices — only DECLARED fields change); term pre_filter uses json_values_equal (unchanged); the once-per-request schema clone is negligible in a brute-force path that already clones every doc source. No findings survive.

### Evidence lens — **HOLDS** (0 findings)
> Per-site attribution proof (definitive): reverting ONLY pre_filter (12610) → only nested_knn_prefilter fails; reverting ONLY post_filter (12666) → only nested_knn_postfilter fails; reverting both → both fail; fix present → both pass. Routing attribution (static): peel_knn_query has no Nested arm, so NEITHER test is peeled to the already-fixed non-nested path — prefilter uses `nested{path,query:knn{filter:prefix}}` → pre_filter; postfilter uses `bool{must:[nested{knn,num_candidates:10}, prefix]}` → post_filter. Non-vacuity: d0 (tag "Hello", vec [1,0,0]) scores cosine 1.0, is the top candidate that reaches the filter and is wrongly kept in fail-before; typed keyword prefix:hel is case-sensitive so correctly dropped after. Genuine, non-vacuous.

---

**Maintainer disposition: merging.** 0 BLOCKING (in fact 0 findings) across all three lenses; #702 is green (18/18) + CLEAN. Completes the nested-kNN filter surface (sibling of the merged #701); #423 remains open for the non-kNN callers.

PR #704, opened 2026-08-24
fix(engine): unwrap single-clause bool through function_score and dis_max (#643)

xerj-org:
## What

Closes the remaining gap of #643 (follow-up to #399/#642). `unwrap_single_clause_bool` — which erases the one-clause `bool` shapes Lucene rewrites to the bare clause, so the memtable and segment score the query identically — recurses through the score-preserving wrappers `Boosted`/`Constant`/`Named` but **not** `function_score` or `dis_max`. A single-clause `bool` nested inside either was therefore left wrapped, the completeness gap the #642 3-skeptic pass flagged.

(The first bullet of #643 — a lone `should` with a *percentage* `minimum_should_match` — was already handled by the `Percentage(pct) => pct < 200` arm; `Field`/`Script` MSM correctly stay wrapped because they resolve per-doc and can exceed the clause count.)

## Fix

Two more arms in `unwrap_single_clause_bool`, each the recursion Lucene's own `rewrite` performs:

- `function_score` — recurse into the inner `query` (its score merges via `boost_mode`, so erasing an inner one-clause bool is score-preserving);
- `dis_max` — map the unwrap over every disjunct (`max(sub-scores) + tie_breaker·Σ(rest)`; each disjunct's score is unchanged).

## Honest scope — this is a completeness / robustness fix, not a user-visible score change

Empirically, the divergence does **not** currently manifest for these shapes: a new integration test (`nested_single_clause_bool_in_wrappers_is_score_neutral`) shows `function_score{query: bool{must:[m]}}` and `dis_max{queries:[bool{must:[m]}]}` already rank exactly like the same wrapper around the bare clause, on **both** the memtable and the flushed segment. That matches #643's own framing ("a completeness gap, not a correctness bug"). The value of this change is that the unwrap now matches Lucene's recursive rewrite for all score-preserving wrappers, so the memtable/segment consistency is guaranteed structurally rather than incidentally — guarding against a future memtable-scoring change re-opening the #399 divergence.

## Fail-before

Unit-level (the honest fail-before for a structural change): `unwrap_recurses_through_function_score` and `unwrap_recurses_through_dis_max` assert the nested one-clause bool is erased to its `Match` clause. With the two arms reverted (source hunk only), both fail — the wrapper is returned with the `bool` intact; with the fix, both pass. The integration test above is kept as a neutrality regression guard.

- Full `xerj-engine` suite — green (incl. the `bool_single_clause_is_score_neutral` suite, 8 tests)

---

**Reviewer note (3-skeptic pass):** the "does not currently manifest" claim is confirmed TRUE for the tested text-field shapes (`function_score{bool{match}}` / `dis_max{bool{match}}` rank identically to the bare-clause wrapper on both memtable and segment, with the arms reverted). The evidence lens additionally found one shape where the fix DOES correct a real pre-existing divergence — a `term` on a keyword field nested in `function_score` on a FLUSHED index: pre-fix the peeled inner is the `Bool`, so `leaf_constant_score` returns `None` and the base score is `score_query_against_doc(bool)` (~`1+ln(1+tf)`), whereas the bare `function_score{term}` already scored exact BM25; post-fix the bool is erased to the term and both agree (toward ES). So the change is beneficial, not purely cosmetic, for that shape. Follow-up nit: the `nested_single_clause_bool_in_wrappers_is_score_neutral` docstring calls itself a "fail-before" — it is a neutrality *regression guard* (passes with and without the fix); the real fail-before is the unit tests `unwrap_recurses_through_*`. Docstring to be corrected.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -35120,6 +35120,33 @@ fn unwrap_single_clause_bool(q: QueryNode) -> QueryNode {
             name,
             query: Box::new(unwrap_single_clause_bool(*query)),
         },
+        // #643: `function_score` merges its inner query's score via `boost_mode`,
+        // so erasing a single-clause `bool` underneath it is score-preserving —
+        // the same recursion Lucene's FunctionScoreQuery.rewrite performs over
+        // its wrapped query. Closes the wrapper the earlier unwrap left un-erased.
+        QueryNode::FunctionScore {
+            query,
+            functions,
+            score_mode,
+            boost_mode,
+            max_boost,
+        } => QueryNode::FunctionScore {
+            query: Box::new(unwrap_single_clause_bool(*query)),
+            functions,
+            score_mode,
+            boost_mode,
+            max_boost,
+        },
+        // #643: `dis_max` scores `max(sub-scores) + tie_breaker*Σ(rest)`; erasing a
+        // single-clause `bool` in any disjunct leaves that disjunct's score
+        // unchanged, so recurse into every query (DisjunctionMaxQuery.rewrite).
+        QueryNode::DisMax {
+            queries,
+            tie_breaker,
+        } => QueryNode::DisMax {
+            queries: queries.into_iter().map(unwrap_single_clause_bool).collect(),
+            tie_breaker,
+        },
         other => other,
     }
 }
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent 3-skeptic adversarial review (each REFUTES in its own worktree, rustc 1.97.1, head `ec0dd9c1`)

**Result: 3/3 HOLDS — ZERO BLOCKING. Three NON-BLOCKING nits (test/docstring accuracy), listed below.**

### Correctness lens — **HOLDS** (0 BLOCKING)
> Both new arms present. Unit fail-before confirmed (revert both arms → unwrap_recurses_through_* fail; restore → pass). Score-preserving in ALL cases: function_score and dis_max are both is_doc_scan_query members, so unwrapping their inner bool changes neither path nor node; score_query_against_doc's Bool arm sums over the single must = the bare clause's brute score, so bool{must:[X]}==X under the scorer both wrappers use; dis_max disjuncts are scored via score_query_against_doc (never leaf_constant_score), so a tie_breaker!=0 multi-disjunct sum is preserved; function_score functions/boost_mode operate on the unchanged inner score. Unwrap fires only for Lucene-erased shapes and is a correct no-op for multi-clause/must_not-only/filter-bearing bools.
> NON-BLOCKING: the "does not manifest" claim holds for the tested text shapes, but for a `term` on a keyword field nested in `function_score` on a FLUSHED index the fix DOES correct a real pre-existing divergence toward ES (brute 1+ln(1+tf) → exact BM25) — beneficial, not cosmetic; noted in the PR body.

### Regression lens — **HOLDS** (0 BLOCKING)
> Full suite: 1083 passed, 0 failed (rustc 1.97.1), incl. both #643 unit tests, the neutrality guard, and existing function_score/dis_max coverage. Fail-before verified at unit level (revert both arms → both unit tests fail; restore → empty diff). unwrap_single_clause_bool is invoked once at index.rs:14372 before any memtable/segment or doc-scan/FTS decision, so it normalizes the whole tree identically for both populations; both arms carry all wrapper params through unchanged; multi-clause bools are NOT unwrapped. No regressing query found.
> NON-BLOCKING: the integration neutrality guard covers only single-disjunct dis_max (tie_breaker=0) and no-functions function_score — a future regression interacting with tie_breaker>0 / scoring functions would be caught only by the analytical argument + unit tests, not this guard.

### Evidence lens — **HOLDS** (0 BLOCKING)
> Central honest claim ("divergence does not currently manifest") is TRUE. Unit fail-before genuine and non-vacuous. Ran the STRONGER neutrality test: with both arms reverted, nested_single_clause_bool_in_wrappers_is_score_neutral PASSES on both memtable (120 resident docs, non-vacuous) and flushed — so function_score{bool{m}}==function_score{m} on both populations without the fix; the fix changes no score/result for these shapes. Mechanism: FunctionScore/DisMax are is_doc_scan_query members, so #399's top-level-shape divergence cannot arise when the bool is nested under them.
> NON-BLOCKING: the integration test docstring (bool_single_clause_is_score_neutral.rs:368) calls itself a "Fail-before" but it is a neutrality *regression guard* (passes with and without the fix); the real fail-before is the unit tests unwrap_recurses_through_*. To be corrected.

---

**Maintainer disposition: merging.** 0 BLOCKING across all three lenses; #704 green (18/18) + CLEAN. The 3 NON-BLOCKING nits are test/docstring-accuracy only (the code + fail-before are correct) — they will be corrected in a small follow-up together with #705's test-comment nit. #423/#643 completeness closed for the function_score/dis_max wrappers.

Issue #577, opened 2026-08-22
perf: #361 residual_gate pass holds O(matches) hits in memory (no trim; count_only still pushes)


xerj-org:
### Context
PR #576 (#361) added a per-segment `residual_gate` pass in the scored-FTS loop (`engine/crates/xerj-engine/src/index.rs`, right after `fts_scored_applied = true;`) that re-applies membership via `doc_matches_query` for bools with a non-projectable non-scoring `filter`/`must_not`. Correct results (verified by #576's 3-skeptic pass), but a memory/perf cliff was flagged (NON_BLOCKING NIT).

### The gap
Unlike the normal walk, the residual pass:
1. **Does not trim `all_hits`** to `materialisation_limit` nor maintain `page_worst` — so a `size:10` bool with a non-projectable filter over a large match set retains **O(matches)** `Hit` structs + hydrated `_source` in memory before the post-loop sort/slice. The deletes path it cites as precedent trims; this doesn't.
2. **Under `count_only` (size:0)** it still hydrates each survivor's `_source` and pushes a full `Hit` into `all_hits` even though size:0 returns no hits.

### Fix direction
- Under `count_only`, run the gate + `total_count += 1` but skip the `Hit` build/push (the source is still needed for `doc_matches_query`, but not the Hit alloc).
- Periodically trim `all_hits` to `materialisation_limit` within the residual loop (mirror `trim_to_cap`/`page_worst`), OR offer to a bounded collector, so memory stays O(cap) not O(matches).

### Acceptance
A `size:10` bool with a selective non-projectable filter over a large corpus holds O(materialisation_limit) hits in memory, not O(matches); size:0 allocates no Hits. Counts + page unchanged (guarded by the existing residual tests).

Follow-up to #361 / PR #576.

xerj-org:
Fixed on \`main\` — both parts of this issue are merged and verified:

- **Part 1 — `count_only` (size:0) still pushed hits.** Fixed in #633 (`72149427`): the residual per-segment loop now skips the `Hit` build/`_source` retention under `count_only`, mirroring the normal walk (`if !count_only { … }`). A size:0 bool with a non-projectable filter no longer materialises O(matches) hydrated hits before the empty slice discards them.
[...]

PR #705, opened 2026-08-24
fix(engine): residual_gate eager-trims all_hits to the page cap (#577)

xerj-org:
## What

Closes the memory cliff of #577 (perf follow-up to #361/#576). The per-segment `residual_gate` pass — which re-applies membership via `doc_matches_query_typed` for a `bool` whose non-scoring `filter`/`must_not` isn't FTS-projectable — runs with `fts_cap == usize::MAX` (it must see the full match set, since the residual clause is re-tested after scoring). But its survivors were pushed into `all_hits` with **no trim**, so a `size:10` bool whose non-projectable filter is non-selective over a large corpus retained **O(matches)** `Hit` structs (+ hydrated `_source`) before the post-loop sort/slice.

(Part 1 of #577 — skipping the `Hit` build under `count_only` — already landed; this PR is the remaining part 2.)

## Fix

Eager-trim `all_hits` to the page cap **inside** the residual push loop, exactly the shape the normal walk already uses (`if all_hits.len() > materialisation_limit.saturating_mul(2) { all_hits = trim_page_to_cap(…) }`). `trim_page_to_cap` keeps the top `materialisation_limit` by score, so the returned page and `total.value` are unchanged — only the peak memory drops from O(matches) to O(cap). Only the non-`sort_topk` (FIFO/score) arm needs it; the `sort_topk` arm already bounds itself via `offer`.

## Fail-before

New unit test `residual_gate_bounds_all_hits_to_the_page_cap` (inside index.rs — an integration test can't see the `#[cfg(test)]` instrument): 2000 docs each matching `match: body:alpha` (FTS scoring) and `range: n>=0` (non-projectable → residual gate), flushed to a segment, searched `size:10`. A `#[cfg(test)] thread_local!` records the high-water mark of `all_hits.len()` in the residual loop.

- With the trim reverted (source hunk only): peak = **2000** (O(matches)) → the `peak <= 513` assertion fails.
- With the fix: peak stays **≤ 513** (2·materialisation_limit+1), and `total.value == 2000`, `hits.len() == 10` are unchanged.

The test also asserts `peak > 0` so it can't silently pass by missing the residual path.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -16710,6 +16710,29 @@ impl Index {
                                                 topk.offer(hit, seq, self);
                                             } else {
                                                 all_hits.push(hit);
+                                                #[cfg(test)]
+                                                RESIDUAL_HITS_PEAK
+                                                    .with(|c| c.set(c.get().max(all_hits.len())));
+                                                // #577: `fts_cap == usize::MAX` for the
+                                                // residual gate, so a non-selective
+                                                // non-projectable filter over a large
+                                                // segment would grow `all_hits` to
+                                                // O(matches) before the post-loop slice.
+                                                // Eager-trim to the page cap here, the
+                                                // same shape the normal walk uses at its
+                                                // `materialisation_limit.saturating_mul(2)`
+                                                // check — `trim_page_to_cap` keeps the
+                                                // top `materialisation_limit` by score, so
+                                                // the page is unchanged.
+                                                if all_hits.len()
+                                                    > materialisation_limit.saturating_mul(2)
+                                                {
+                                                    let (kept, _worst) = self.trim_page_to_cap(
+                                                        std::mem::take(&mut all_hits),
+                                                        materialisation_limit,
+                                                    );
+                                                    all_hits = kept;
+                                                }
                                             }
                                         }
                                     }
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent 3-skeptic adversarial review (each REFUTES in its own worktree, rustc 1.97.1, reviewed head `c49e65d7`)

**Result: 3/3 HOLDS — ZERO BLOCKING.** Two NON-BLOCKING nits (same test-comment inaccuracy). The mid-loop eager trim is proven result-correct.

### Correctness lens — **HOLDS** (0 BLOCKING)
> The mid-loop eager trim is result-correct. trim_page_to_cap keeps top materialisation_limit by the SAME total comparator (score DESC, seq_no ASC, _id ASC) as the final page assembly; materialisation_limit is (from+size+100).max(256) or fast-path from+size — both ≥ from+size — so by the streaming top-k invariant any hit in the global top-(from+size) survives every mid-loop trim (deep paging included; ties total via unique _id). No page hit can be dropped. scan_worst is confined to the memtable brute-scan arm and never read in the segment residual arm, so discarding _worst is a genuine no-op. Survivors already passed version_map liveness before the push, so trimming post-liveness is safe under deletes/ghosts. total_count is incremented BEFORE the count_only-guarded push/trim block, so total.value is trim-independent. Fail-before verified (revert trim → peak=2000; fix → pass).

### Regression lens — **HOLDS** (0 BLOCKING)
> Full suite (rustc 1.97.1): 0 failed across all binaries (629 lib + every integration binary). Fail-before reproduced (revert only the trim block → residual_gate_bounds panics peak=2000; total.value==2000 and hits.len()==10 assert before the peak check and pass; restore → passes). Added and removed a temporary adversarial cross-check (size:30/from:0 vs size:10/from:20 with distinct BM25 scores over 2000 residual-gate matches): deep page == top30[20..30] exactly. sort_topk.is_some() arm uses bounded topk.offer (no trim). Worktree left clean.

### Evidence lens — **HOLDS** (0 BLOCKING)
> Ran the test three ways: with fix → pass (instrumented peak=21); trim reverted → FAIL peak=2000 (but total.value==2000 and hits.len()==10 still pass, proving count+page-size preserved); restored → pass. residual_gate genuinely fires (needs_fts + numeric range non-projectable via bool_has_nonprojectable_nonscoring; peak>0 confirms it's not served by the columnar/count fast paths). Counter is #[cfg(test)] only (no release cost).
> NON-BLOCKING (both correctness+evidence): the test comment states materialisation_limit=256/bound 513, but the fast-path re-bind at index.rs:16061 sets it to from+size=10 for this query, so the real peak is ~21. The `peak<=513` assertion still holds and is non-vacuous at N=2000 (no-fix peak=2000>513) — the comment's 256 rationale is just inaccurate.

---

**Maintainer disposition: merging on green.** 0 BLOCKING across all three lenses. **Note:** after this review, #704 (#643) merged and collided with this PR at the index.rs EOF test-mod append; I rebased onto post-#704 main (`fff8e654`) — the collision was resolved mechanically (both test mods kept) and the #577 fix code (the trim + counter) is byte-identical to the reviewed `c49e65d7`, re-verified locally (fmt / clippy --all-targets / both unit tests green). The NON-BLOCKING `materialisation_limit` comment will be corrected in a small follow-up.

PR #706, opened 2026-08-24
docs(test): correct the #643 neutrality-guard docstring (#704 review follow-up)

xerj-org:
Corrects the misleading docstring on `nested_single_clause_bool_in_wrappers_is_score_neutral` (merged in #704) that the 3-skeptic pass flagged as NON-BLOCKING: it labelled itself a *fail-before* and claimed the memtable *diverges*, but the test passes with and without the #643 fix — the #399 divergence does not manifest for these nested shapes, so it is a neutrality **regression guard**. The real fail-before is the unit tests `unwrap_recurses_through_*`. Comment-only, no behavior change (fmt clean; the test still passes). A follow-up commit will also correct the #705 `materialisation_limit` test comment once #705 lands.

### DIFF
--- a/engine/crates/xerj-engine/tests/bool_single_clause_is_score_neutral.rs
+++ b/engine/crates/xerj-engine/tests/bool_single_clause_is_score_neutral.rs
@@ -359,14 +359,18 @@ async fn must_not_only_bool_is_not_unwrapped() {
     }
 }
 
-/// #643 remaining gap: a single-clause `bool` nested inside a score-preserving
-/// wrapper that `unwrap_single_clause_bool` does NOT recurse through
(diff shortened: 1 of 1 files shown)


xerj-org:
Merging: this PR is **comment-only** (verified: `git diff origin/main...` shows 0 non-doc-comment changed lines — purely the `///` docstring). It corrects a claim already independently established by #704's 3-skeptic pass (the `nested_single_clause_bool_in_wrappers_is_score_neutral` guard passes with and without the #643 fix — a neutrality regression guard, not a fail-before). A separate adversarial code review is not applicable to a zero-code-change doc fix. Merged ahead of the rc.66 stamp (#707) so it stays on the rc.65 landing and needs no tag-ahead rebase.

PR #708, opened 2026-08-24
fix(engine): dynamic field-budget counts the whole subtree at admission (#312)

### LINKED ISSUE #312: Dynamic-mapping follow-ups from the #307 review: one-shot field-budget overshoot; array-nested keys invisible to the evolve throttle
Found in the pre-merge review of #307 (which closed out the #292 findings). Neither blocked the merge; both are bounded.

1. **Single-insert budget overshoot.** `engine/crates/xerj-engine/src/index.rs` ~18403/~18520 (fresh top-level add: `field_count += 1` for a `dynamic_field_config` result that may carry an entire nested subtree) and ~28838 (`merge_dynamic_children_into` `None` branch, whole subtree counted once): one document can push the schema past `max_fields_per_index` by up to its own nested key count in a single write-lock session. The next evolve recomputes the now-recursive `Schema::field_count()`, so growth is blocked immediately after — a bounded one-shot overshoot, not unbounded growth. Hardening: count the subtree's real size at admission.

2. **Evolve-throttle blind spot for nested objects inside arrays.** `hash_all_field_names` (~28785) recurses only into `Value::Object`, while the merge path (`object_keys_of`, ~28768) does unwrap array-of-objects. A new nested key arriving as `{"tags": [{"new": 1}]}` under an unchanged top-level shape hashes identically, so the single-doc path suppresses evolve until the unconditional 100-doc recheck (~6584) — delayed discovery, not permanent. Same class as the bug #307 fixed, one container type over.

Also noting for the record: `register_field_cap_entry` (`xerj-api/src/es_compat.rs` ~19071) appends `idx_name` without the dedupe check its multi-field sibling has — not currently reachable as a duplicate, but an asymmetry a future caller could trip.

xerj-org:
## What

Fixes bug 1 of #312 (dynamic-mapping follow-ups from the #307 review): a single dynamic insert whose new top-level field is a nested **object** counted the whole subtree as one field against `max_fields_per_index`, so one document could overshoot the budget by the subtree's key count before the next evolve recomputed the (recursive) `Schema::field_count()` and blocked further growth. Bounded, but a real overshoot.

**Bug 2 of #312 is already fixed** (via #382): `hash_all_field_names` now has a `Value::Array` arm that recurses into every element, with the `hash_all_field_names_recurses_into_array_elements` test — so the evolve-throttle no longer misses a key arriving as `{"tags":[{"new":1}]}`. This PR closes the remaining half.

## Fix

At each dynamic-evolve **add** site, count the new field config's real recursive size (`FieldConfig::total_field_count()` — the same measure `Schema::field_count()` uses for the guard) instead of `+= 1`, and gate on `field_count + fc_size > max` before admitting:

- `evolve_schema_from_docs` (the batch path);
- `evolve_schema_from_doc` (the single-doc path — **not** among the lines the issue cited; found via the test, since single-doc `index_document` routes here);
- `merge_dynamic_children_into`'s `None` branch (a nested child merged in one pass).

The existing budget tests are unaffected (they add leaf fields, `total_field_count == 1`).

## Fail-before

New unit test `dynamic_insert_of_a_nested_subtree_cannot_overshoot_the_field_budget` (`max_fields_per_index = 5`, seed one field, then a doc whose new `big` field is a 6-key object → subtree size 7). With the single-doc add-site fix reverted, `Schema::field_count()` reaches **8** (overshoot) and the test fails; with the fix, the over-budget `big` subtree is rejected and `field_count` stays ≤ 5. (Reverting the fix is how I discovered the single-doc path was a distinct, uncovered site.)

Closes #312 (both bugs now addressed).

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -3670,6 +3670,54 @@ mod semantic_deadline_regression_tests {
         );
     }
 
+    /// #312 bug 1: a single dynamic insert whose new top-level field is a nested
+    /// OBJECT must count the whole subtree against `max_fields_per_index`, not as
+    /// one field — otherwise one document overshoots the budget by the subtree's
+    /// key count (bounded, but a real overshoot) before the next evolve blocks it.
+    #[tokio::test]
+    async fn dynamic_insert_of_a_nested_subtree_cannot_overshoot_the_field_budget() {
+        let dir = TempDir::new().unwrap();
+        let mut config = xerj_common::config::Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        config.limits.max_fields_per_index = 5;
+        let engine_instance = Engine::new(config).expect("engine");
+        engine_instance
+            .create_index("budget-312", Schema::empty())
+            .unwrap();
+        let index = engine_instance.get_index("budget-312").unwrap();
+
+        // Seed a base field so `big` below arrives as a SUBSEQUENT new top-level
+        // field on the dynamic-evolve path (the site #312 bug 1 cites), not the
+        // first-doc inference path.
+        index
+            .index_document(Some("seed".to_string()), serde_json::json!({"seed": 1}))
+            .await
+            .unwrap();
+        let base = index.schema().await.field_count();
+        assert!(base <= 5, "sanity: base schema within budget, got {base}");
+
+        // A later doc whose only new top-level field is a 6-key nested object:
+        // the dynamic `Object` FieldConfig has total_field_count == 7 (1 + 6),
+        // which would push base+7 well past the cap of 5.
+        index
+            .index_document(
+                Some("d1".to_string()),
+                serde_json::json!({
+                    "seed": 1,
+                    "big": {"a": 1, "b": 2, "c": 3, "d": 4, "e": 5, "f": 6}
+                }),
+            )
+            .await
+            .unwrap();
+
+        let fc = index.schema().await.field_count();
+        assert!(
+            fc <= 5,
+            "a single dynamic insert must not overshoot max_fields_per_index (5); \
+             the whole `big` subtree (size 7) should be rejected, got field_count={fc}"
+        );
+    }
+
     #[tokio::test]
     async fn mapping_vector_over_existing_hidden_chunks_fails_closed_before_visibility() {
         let dir = TempDir::new().unwrap();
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent 3-skeptic adversarial review (each REFUTES in its own worktree, rustc 1.97.1, head `706c3d3b`)

**Result: 3/3 HOLDS — ZERO BLOCKING.** One NON-BLOCKING note (pre-existing behavior), below.

### Correctness lens — **HOLDS** (0)
> total_field_count() is EXACTLY the measure the budget guard uses. FieldConfig::total_field_count (types.rs:439) = 1 + recursive children; Schema::field_count (types.rs:519) = sum over top-level of that same function. After add_field(fc), schema.field_count() rises by exactly fc.total_field_count(), so `field_count += fc_size` stays in lockstep at all three sites (19892/19904, 20017/20029, merge None-branch 34770/34780). No double-count (None-branch counts the subtree once, no later re-descent), no off-by-one (`>= limit` → `+ fc_size > limit` keeps the max reachable == limit for fc_size=1). Updated test values verified genuinely correct (a string maps to Text + a pushed `keyword` child → total 2): 10+2=12 and 1+2=3 are exact. No change to max_fields semantics — the guard/explicit-mapping path already counted nested children + multi-fields; the fix only makes ADMISSION consistent. Fail-before real: reverting only the single-doc path → field_count=8 (the exact overshoot); restored → passes.

### Evidence lens — **HOLDS** (0)
> Independently verified the two updated values are the genuinely-correct post-fix counts, NOT bumped-to-pass: a dynamically-mapped string → Text (infer_field_type 34389-34394) + one `.keyword` child (dynamic_field_config_at_depth 34566-34570) = total 2, confirmed via code; so merge...adds_new_nested_keys 10+2=12 and a_nested_companion 1+2=3 (companion vec_chunks correctly uncounted). Measure consistency confirmed (no double-count / off-by-one). `>= max → + fc_size > max` preserves scalar behavior identically. Fail-before: reverting only the single-doc guard made the new test fail with field_count=8, restored → passes — proves it reaches the single-doc path and is non-vacuous. All 4 targeted tests pass. Non-defect note: the read-lock pre-screens still estimate +1 (lenient early-outs), but the authoritative subtree guard runs under the write lock — under-count only falls through to the accurate loop, never over-rejects.

### Regression lens — **HOLDS** (1 NON-BLOCKING)
> Full suite `cargo test -p xerj-engine`: 0 failed (632 lib + all integration/doc binaries; dynamic-mapping, budget, multi-vector-companion, semantic families pass). Fail-before verified (revert single-doc site → field_count=8; restore → passes). Measure consistency + updated values (12, 3) confirmed correct. max_fields_per_index semantics unchanged.
> **NON-BLOCKING:** partial mid-document admission — when a new subtree overshoots, the loop `break`s, so later small fields in the SAME document that would individually fit are skipped for the mapping (still queryable via _source; only GET _mapping/_field_caps omit them). This is **pre-existing** `break` behavior (identical pre-fix), NOT introduced by this PR, and is arguably correct once the recursive budget is genuinely exhausted.

---

**Maintainer disposition: merging.** 0 BLOCKING across all three lenses; the counting change is exactly consistent with the guard's Schema::field_count(), the updated test values are independently verified correct (not bumped-to-pass), and fail-before is real. The NON-BLOCKING partial-admission note is pre-existing behavior — filing no follow-up as it is arguably correct. **Closes #312** (bug 2 already fixed via #382).

Issue #326, opened 2026-08-12
query_string aborts the process on a multi-byte char boundary (parser.rs:1825) — same class as #207/#261/#272


SENTANO:
## Summary

`query_string` panics on a byte-index slice that lands inside a multi-byte character, aborting the process. This is the same defect class as #207 (tokenisers), #261 (parser sweep) and #272 (`aggs::parse_time_zone_offset`) — still live in `crates/xerj-query/src/parser.rs:1825`, reachable from an unauthenticated `_search`.

```
thread 'xerj-rt' panicked at crates/xerj-query/src/parser.rs:1825:21:
end byte index 38 is not a char boundary; it is inside 'х' (bytes 37..39 of string)
```

Since `[profile.release]` sets `panic = "abort"`, this is not a 400 — it is the process. Ours was restarted by a launchd `KeepAlive` after ~10s; without a supervisor the node stays down.

## Reproduction

Release binary `v1.0.0-rc.15`, `aarch64-apple-darwin`, SHA-256 verified against the published sum. Fresh data dir, one indexed corpus.

```bash
xerj --insecure --data-dir ./data &

# aborts the process
curl -X POST localhost:9200/ax-*/_search -H 'Content-Type: application/json' \
  -d '{"size":1,"query":{"query_string":{"query":"чепуха","default_operator":"AND"}}}'
# → curl: (52) Empty reply from server

# answers 200 on the same instance
curl -X POST localhost:9200/ax-*/_search -H 'Content-Type: application/json' \
  -d '{"size":1,"query":{"query_string":{"query":"ротация ключей","default_operator":"AND"}}}'
```

Both queries are Cyrillic; only one aborts. The trigger is where the byte offset lands, not the script — so "non-Latin queries crash" understates it: they crash intermittently, which is worse to diagnose in the field. Observed offsets in our logs: `end byte index 38` (41-byte query) and `end byte index 55` (58-byte query).

## Why the existing fuzzing did not catch it

`fuzz_targets/query_string` exists and runs in CI, but its corpus is effectively ASCII: of 12 seeds in `fuzz/seeds/query_string`, exactly one contains any byte ≥ 0x80 — `regression-207-nbsp-continuation-byte`, itself a regression seed from #207. `fuzz/seeds/query_dsl` has none at all.

[...]

PR #710, opened 2026-08-24
fix(autoindex): a byte-identical duplicate must not wedge a completed run (#345)

### LINKED ISSUE #345: autoindex aborts on a byte-identical duplicate file (split out of #326)

Verbatim from #326:

**1. `autoindex` aborts on a byte-identical duplicate.** On a repository where two tracked files have identical content:

```
error: replace catalog alias for infra/cloud-init/baseline.sh:
       delete_by_query: HTTP 500 Internal Server Error
xerj-done ok=false exit=1 reason=aborted
```

The first pass over the same tree completed (`exit 3`, "1 duplicate aliases"); the failure appeared on a re-run, so it looks tied to catalog state after a partial run rather than to duplicates as such.

Vinz2168:
Closes #345.

## Problem

A corpus containing two byte-identical files could not be indexed. The run printed:

```
error: replace catalog alias for infra/cloud-init/baseline.sh:
       delete_by_query: HTTP 500 Internal Server Error
xerj-done ok=false exit=1 reason=aborted
```

and stayed that way: every rerun reported the same line, because by the time the failure hits, every data document is already durably indexed and the per-file journal has already committed — so the next run indexes 0 files and reaches the identical abort.

## Root cause

A duplicate path is the only thing that puts anything in `alias_paths_to_replace`, so it's the only thing that makes a run issue the catalog's duplicate-alias `_delete_by_query` at all. That call was propagated with `?`, running *after* every document is durably indexed and the journal has committed. A 5xx there converted a fully successful indexing run into a permanent abort. The sweep itself is only legacy-id-scheme cleanup (see the comment it sits under) — replacement correctness comes from the later bulk that re-indexes the alias document under a deterministic id (`catalog::duplicate_file_id`), which is unaffected by the sweep failing.

The 5xx was also uninvestigable: `with_retry` kept only the HTTP status and discarded the response body, even though every by-query refusal answers a structured `{"error": {"type", "reason"}, "status"}`.

## Fix

1. `esclient.rs` — `with_retry` now folds a bounded prefix of the server's `error.type: error.reason` (or raw body, for a proxy's HTML page) into the error, for every call site.
2. `lib.rs` — the alias sweep is now batched (`ALIAS_SWEEP_CHUNK = 1024` paths per request via a `terms` filter) instead of one round trip per duplicate path.
3. `lib.rs` — the sweep is reported, not fatal. The run finishes `ok=true exit=3 reason=catalog-alias-sweep-failed`, names the unswept paths on the progress surface, and records `catalog_alias_sweep_failed_paths` / `catalog_alias_sweep_error` on the run document. No new exit code — `3` already means "completed, something recorded rather than fatal"; `reason` distinguishes it from `completed-with-junk`. The catalog *bulk* that follows the sweep is still fatal, so a catalog index that genuinely cannot be written still fails the run.

Honest scope: this does not fix whatever made the engine answer 500 in the reporter's environment (that traces to `Index::search` → `EngineError::Fts` → a sticky `ReadAdmission::Poisoned` arm needing a node restart, tracked separately). It stops that failure from permanently wedging an otherwise-complete indexing run.

## Testing

- `cargo test -p xerj-autoindex --profile ci-test -- --test-threads=2`: new tests pass; the 19 pre-existing failures (macOS non-UTF-8 filename handling + parallel-run ordering artifacts) reproduce identically on unmodified `main`, confirmed via a clean-worktree run — unrelated to this change.

New tests:
- `failure_resume_http_tests::a_refused_duplicate_alias_sweep_is_reported_loudly_and_is_not_fatal` — real HTTP, legacy (graph) path, 500 injected on the sweep only; asserts terminal non-aborted state, the server's reason on the surface, and a no-duplicate control untouched by the same fault.
- `esclient::an_exhausted_5xx_retry_reports_the_server_reason_not_only_the_status`
- `code_coverage_tests::the_alias_sweep_partitions_its_paths_exactly_across_the_chunk_boundary` — 0 / 1 / 1023 / 1024 / 1025 paths.
- `code_coverage_tests::a_failed_alias_sweep_names_its_paths_the_cause_and_caps_the_listing`

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -103,7 +103,10 @@ be driven by you on their behalf, so the loop is fixed:
 
 Exit codes: `0` complete · `3` completed-with-junk — **this is success** · `2`
 usage · `1` any error at all (read the `error:` line before acting) · `4`
-needs a decision (gate above; answer with `--approve`, never a retry). Percent,
+needs a decision (gate above; answer with `--approve`, never a retry). Exit `3`
+also carries `reason=catalog-alias-sweep-failed`: the corpus **is** indexed and
+the journal committed, and only the catalog's duplicate-alias cleanup could not
+run — read `reason`, not the bare code, to tell the two apart. Percent,
 ETA and the drawn bar are honest or absent: `unknown` / `[????…]` when there is
 no denominator, and a full bar only at a real 100%. The short version is under
 "Running an index on someone's machine" in
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2052,6 +2052,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   right: they deleted CJK, Cyrillic, Greek, Hebrew and Arabic documents
   worldwide. Tracked as **#381**.
 
+- **`autoindex` aborted permanently on a corpus containing a byte-identical
+  duplicate** (#345). A duplicate path is the only thing that makes a run issue
+  the catalog's duplicate-alias `_delete_by_query` at all, and that call was
+  fatal. Because it runs *after* every document is durably indexed and after the
+  per-file journal has committed, a 5xx there turned a fully successful run into
+  `xerj-done ok=false exit=1 reason=aborted` — and it stayed that way: the
+  journal was complete, so every rerun indexed zero files and aborted on the
+  same line. The sweep is metadata-only bookkeeping and is now reported instead
+  of fatal: the run finishes `ok=true exit=3
+  reason=catalog-alias-sweep-failed`, names the paths whose alias document was
+  not swept, and records `catalog_alias_sweep_failed_paths` /
+  `catalog_alias_sweep_error` on the run document. Exit `3` already means
+  "completed, and something is recorded rather than fatal", so no new exit code
+  was introduced — read `reason` to tell it apart from `completed-with-junk`.
+  The catalog *bulk* that follows the sweep is still fatal, so a catalog index
+  that cannot be written still fails the run.
+
+- **Any 429/5xx from the server was reported as a bare status line.** The retry
+  wrapper kept `HTTP 500 Internal Server Error` and discarded the response body,
+  which is the same sentence for a poisoned index, a full disk and a panicking
+  handler — the reason #345 was filed as "not investigated". A bounded prefix of
+  the server's own `error.type: error.reason` (or the raw body, for a proxy's
+  error page) now travels with the status on every `with_retry` call site.
+
+- **The duplicate-alias sweep issued one HTTP request per duplicate path.** It
+  now names up to 1024 paths per request with a `terms` filter, so a corpus with
+  K duplicates costs `ceil(K/1024)` round trips instead of K — and has
+  proportionally less surface on which to fail.
+
 ## [1.0.0-rc.16] - 2026-08-13
 
 Both rc.15 known issues (the progress-stream forgery and the outer-`.gitignore`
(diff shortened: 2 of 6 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `5599f7f3`)

Each skeptic was tasked to **refute** the PR from a distinct angle. All three returned **MERGE_OK** with **zero blocking findings and zero nits**. Verdicts verbatim:

### 1. Downgrade-safety / query-correctness / rerun-idempotency → **MERGE_OK**
All six angles verified:
1. The sweep (lib.rs:5513-5542) runs in finalize-catalog **after** the data `_bulk` durability and the journal commit, and targets `CATALOG_INDEX` only (line 5538) — **no data-loss can be hidden**.
2. **Rerun idempotency HOLDS and the "rerun to clear" advice is TRUE:** `alias_paths_to_replace` is populated from filesystem-recomputed `inventory.duplicates` (3493/3888) **and** the persisted journal `plan.duplicate_files` (3776) — both independent of this run's newly-indexed files, so a journal-complete zero-new-files rerun still re-detects the duplicate and re-runs the sweep (doubly robust: recomputed + persisted). No early-return short-circuits the sweep on a normal graph-mode rerun.
3. `alias_sweep_query` terms-union == the old per-path `term` loop exactly; `path`/`status` are keyword (catalog.rs:69/72), the `status:duplicate` guard is preserved → no over/under-match.
4. `sort_unstable` + `chunks(1024)` partitions the path vec exactly (chunk-boundary test asserts it).
5. The `cat_buf` catalog bulk stays fatal (bails on server_errors, 5914) — a truly broken catalog still aborts.
6. **Fail-before proven:** neutralizing only the downgrade (`if let Some(e)=&alias_sweep_error { bail!(e) }`) makes `a_refused_duplicate_alias_sweep_is_reported_loudly_and_is_not_fatal` FAIL (run aborts); reset → passes. Full suite 700+ green.

- `FMT=0`, `CLIPPY=0` (-D warnings), `BUILD=0` (ci-test), `TEST_EXIT=0`. **714 passed / 0 failed / 0 ignored** (700 lib + 8 + 1 + 1 + 4 integration). All 4 new tests pass.
- The exit-code widening (3 now also fires on `alias_sweep_error`) is verified **non-regressive**: refused sweep → `exit=3 reason=catalog-alias-sweep-failed ok=true` (not the old `exit=1 aborted` wedge); existing `0/3/completed-with-junk/reason=` assertions still hold. DEV-profile spot-runs also green. No `:9200`.

### 3. Evidence & claim-integrity → **MERGE_OK**
1. The test genuinely reproduces the #345 wedge — byte-identical duplicate makes the sweep run, mock refuses the sweep `delete_by_query` with 5xx, and the rerun over the completed corpus now finishes `ok=true exit=3 reason=catalog-alias-sweep-failed` instead of aborting; it asserts `ok=true` AND the reason AND the failed paths AND the run-doc keys. Fail-before (revert to `?`) → test fails.
2. The "sweep is idempotent, rerun once the condition clears" claim is **TRUE and the stale state is recoverable**: `alias_paths_to_replace` is recomputed every run from `inventory.duplicates` (3888) and `previous_aliases` (3776), so a rerun re-reaches finalize-catalog and re-issues the idempotent query-based `delete_by_query` — the test's own second run demonstrates the rerun re-triggering the sweep.
3. `server_reason` covers both structured `{error:{type,reason}}` and bare HTML proxy bodies, bounds to **512 chars** via `.chars().take` (no multibyte panic, #326), trims/empties to `None`; its test exercises both shapes.
4. `cli.rs` EXIT CODES help (exit 3 / `catalog-alias-sweep-failed` / rerun-when-clear) matches the emitted `code=3`, `ALIAS_SWEEP_FAILED_REASON`, and run-doc keys with no drift.

---

**Decision:** 0 BLOCKING, 0 nits, all checks green (0 pending/failing), `mergeStateStatus=CLEAN`, head at the exact verified SHA `5599f7f3`. **Merging.** (The rerun-idempotency crux was also independently re-traced by the reviewer and confirmed sound.)

PR #711, opened 2026-08-24
fix(autoindex): #585 case 2 — refuse a --no-graph genesis bootstrap continued on the graph path

### LINKED ISSUE #585: autoindex: cross-path graph re-run still proceeds over a NOT-YET-committed --no-graph state dir (pending-sync / pending-genesis)

Vinz2168:
Closes #585 (case 2). Follow-up to #584 and #692 (case 1, merged).

## Background

#692 refused a **pending** (uncommitted) `--no-graph` generation resumed on the graph path, by comparing the pending sync's recorded `graph_enabled` against `cfg.no_graph`. It explicitly left out case 2 — a `--no-graph` generation's **genesis bootstrap** (`sync_bootstrap` committed, interrupted before `sync_begin`) continued on the graph path — because genesis has no execution identity of its own to compare a mode against (`validate_genesis` requires `execution: None`), and noted "no clean mode source".

## The mode source

There is one — it just isn't in the manifest. The *journal record* that produces a generation-0/empty-groups `committed_manifest` is unambiguous even though the resulting manifest shape is not:

- `sync_bootstrap_genesis` is called from exactly one place, `begin_non_graph_generation`, and only under `cfg.no_graph`. It appends a `{"kind": "sync_bootstrap", ...}` journal record.
- A legacy graph-path journal can independently reach the *identical* manifest shape via `write_plan(Plan::default())` (e.g. every file removed from an already-indexed folder) — but that writes a `{"kind": "plan", ...}` record, never `"sync_bootstrap"`.

So "did a durable `sync_bootstrap` record actually replay" is a real, durable fact that tells the two apart, even though the resulting `CommittedManifest` is byte-for-byte the same either way.

## Fix

- `state.rs` — journal replay (both the preflight-only pass and the authoritative one) now tracks whether a `sync_bootstrap` record has replayed, exposed as `JournalPreflight::no_graph_genesis`.
- `lib.rs` — the graph-path continuation is refused when `genesis_recovery && preflight.no_graph_genesis && !cfg.no_graph`, symmetric to case 1's guard and placed at the same point (before the journal is opened for write, before `gc_snapshots`) — nothing is mutated. `--fresh` is exempted, matching how `blocking_generation` already exempts genesis recovery from the `--fresh` refusal above it: nothing is committed past genesis, so discarding and rebuilding in the requested mode is safe.

## Testing

- New fail-before test `no_graph_genesis_bootstrap_is_refused_on_the_graph_path`: interrupts a `--no-graph` run between `sync_bootstrap` and `sync_begin` (via `snapshot_max_bytes`), re-runs the same state dir on the graph path (must refuse, mutate nothing), then re-runs in the same `--no-graph` mode as a positive control (must continue normally). Verified fail-before: with the `lib.rs` guard neutralized, the graph re-run silently succeeds (exit 0) instead of refusing.
- `cargo test -p xerj-autoindex --profile ci-test -- --test-threads=2`: 679 passed; the 19 failures are pre-existing environment issues (macOS non-UTF-8 filename handling + parallel-run ordering artifacts) that reproduce identically on unmodified `main` — confirmed via a clean-worktree run, unrelated to this change.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -2156,3 +2156,61 @@ fn pending_no_graph_generation_is_refused_on_the_graph_path() {
         "#585: the refused re-run must mutate nothing"
     );
 }
+
+/// #585 (case 2): a `--no-graph` generation's genesis bootstrap (`sync_bootstrap`
+/// committed, interrupted before `sync_begin`) must NOT be continued on the
+/// default graph path either. Genesis carries no execution identity of its own
+/// (`validate_genesis` requires `execution: None`), so unlike case 1 the guard
+/// cannot compare a recorded mode — it instead relies on `sync_bootstrap` having
+/// actually replayed, which only `begin_non_graph_generation` ever writes, and
+/// only under `--no-graph`.
+#[test]
+fn no_graph_genesis_bootstrap_is_refused_on_the_graph_path() {
+    let _guard = HTTP_E2E_LOCK.lock().unwrap_or_else(|p| p.into_inner());
+    let _replay_guard = sync_executor::REPLAY_FAILPOINT_TEST_LOCK
+        .lock()
+        .unwrap_or_else(|p| p.into_inner());
+    let corpus = tempfile::tempdir().unwrap();
+    let state_dir = tempfile::tempdir().unwrap();
+    fs::write(corpus.path().join("a.csv"), "id,value\n1,alpha\n").unwrap();
+    let endpoint = HttpEndpoint::start();
+    let mut no_graph = cfg(corpus.path(), state_dir.path(), &endpoint.url, false);
+    no_graph.snapshot_max_bytes = 1;
+
+    // Interrupt the very first --no-graph run between sync_bootstrap and
+    // sync_begin: generation 0 (genesis) is committed, nothing else is.
+    let error = run_index(no_graph.clone()).unwrap_err();
+    assert!(format!("{error:#}").contains("snapshot source footprint"));
+    assert_eq!(journal_events(state_dir.path(), "sync_bootstrap"), 1);
+    assert_eq!(journal_events(state_dir.path(), "sync_begin"), 0);
+    assert_eq!(journal_events(state_dir.path(), "sync_commit"), 0);
+    assert_eq!(endpoint.data_docs().len(), 0);
+
+    // Re-run the SAME state dir on the default graph path. The pending
+    // --no-graph genesis bootstrap must be refused, not continued.
+    let mut graph = no_graph.clone();
+    graph.no_graph = false;
+    graph.snapshot_max_bytes = 64 << 30;
+    let error = run_index(graph).unwrap_err();
+    let rendered = format!("{error:#}");
+    assert!(
+        rendered.contains("genesis bootstrap") && rendered.contains("different graph authority"),
+        "#585 case 2: a --no-graph genesis bootstrap must be refused on the graph path with the \
+         clear message, got: {rendered}"
+    );
+    // No mutation: nothing indexed, no new durable transaction beyond the
+    // original sync_bootstrap.
+    assert_eq!(journal_events(state_dir.path(), "sync_bootstrap"), 1);
+    assert_eq!(journal_events(state_dir.path(), "sync_begin"), 0);
+    assert_eq!(journal_events(state_dir.path(), "sync_commit"), 0);
+    assert_eq!(endpoint.data_docs().len(), 0);
+
+    // Positive control: re-running in the SAME (--no-graph) mode continues
+    // the interrupted genesis normally.
+    no_graph.snapshot_max_bytes = 64 << 30;
+    assert_eq!(run_index(no_graph).unwrap(), 0);
+    assert_eq!(journal_events(state_dir.path(), "sync_bootstrap"), 1);
+    assert_eq!(journal_events(state_dir.path(), "sync_begin"), 1);
+    assert_eq!(journal_events(state_dir.path(), "sync_commit"), 1);
+    assert_eq!(endpoint.data_docs().len(), 1);
+}
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -3356,6 +3356,35 @@ pub fn run_index_report(cfg: IndexCfg) -> Result<(i32, Option<Value>)> {
         finish_generated_progress(&pr, code, &summary);
         return Ok((code, Some(summary)));
     }
+    // #585 case 2: a --no-graph generation's genesis bootstrap (sync_bootstrap
+    // committed, nothing beyond generation 0 committed yet — no pending_sync,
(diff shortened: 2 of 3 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `5ac5ef89`)

Each skeptic was tasked to **refute** the PR from a distinct angle. All three returned **MERGE_OK** with **zero blocking findings**. Verdicts verbatim:

### 1. Correctness & false-guard → **MERGE_OK**
- **Fail-before (confirmed):** Reverted ONLY the guard block (lib.rs:3378-3387), kept test + all of state.rs. The test FAILED at `incremental_reconcile_http_tests.rs:2194` — `run_index(graph)` returned `Ok(0)` instead of an error (the graph re-run silently continued the interrupted `--no-graph` genesis, the exact #585 case-2 hazard). Restored → test passes. The test genuinely exercises the fix.
- **False-positive hunt:** all three legitimate-run cases correctly do NOT fire — (a) a completed `--no-graph` run sits at generation ≥ 1 (`tx_id = base.generation + 1`, lib.rs:434) so `genesis_recovery` is false; (b) a legacy empty `write_plan` reaches genesis shape via the `plan` record arm, which never sets `no_graph_genesis`; (c) a `--no-graph` re-run fails the `!cfg.no_graph` clause (positive control passes).
- **False-negative hunt:** `no_graph_genesis` is set unconditionally on every `sync_bootstrap` replay (state.rs:672-675), and `sync_bootstrap` is only ever written under `cfg.no_graph` (both `begin_non_graph_generation` callers gated at lib.rs:3425/4191; other call sites are `cfg(test)`). #585 case 2 has no `pending_sync`, so control reaches the guard; `genesis_recovery`/`preflight.no_graph_genesis` are correct/in-scope and the `bail!` is reachable (not preempted).
- **Suite:** full `cargo test -p xerj-autoindex` — 697 lib + 8/1/1/4 integration passed, **0 failed, 0 ignored**. Mock on 127.0.0.1:0; never touched :9200.
- **Non-blocking (nit):** graph-path `--dry-run` over a `--no-graph` genesis now bails instead of emitting a projection (the pending-sync path prints a projection + returns Ok). Safe — no mutation; refusing to project a mode that would itself be refused is defensible. Minor asymmetry only.

- **Suite (authoritative ci-test):** `FMT=0`, `CLIPPY=0` (`-D warnings`), `BUILD=0` (ci-test). Full autoindex suite ci-test: **697 passed / 0 failed**, plus integration bins 8/1/1/4 all passed / 0 failed. New test `no_graph_genesis_bootstrap_is_refused_on_the_graph_path` OK; case-1 `pending_no_graph_generation_is_refused_on_the_graph_path` OK. Dev-profile spot-run of the new test: 1 passed / 0 failed.
- The tuple-arity change to `read_plan_for_preflight` is safe: exactly one caller (state.rs:761), the `PreflightReplay` 5-tuple alias matches every return arm, and `JournalPreflight` is constructed at exactly one site (state.rs:787) which sets the new field, with the fresh-error arm setting it `false` — no other construction site, no test-side struct literal to break.
- **No regressions, no build-hygiene issues, nothing blocking.**

### 3. Evidence & message-integrity → **MERGE_OK**
- **Fail-before (confirmed):** Neutralized the guard (`if false && genesis_recovery && ...`). The test FAILED — the graph re-run returned `Ok(0)`, so `unwrap_err()` panicked at test:2194. This proves the guard is the **sole** source of the second-run error and the assertion cannot pass on any other error (the snapshot-footprint interrupt can't be hit on the second run: `snapshot_max_bytes = 64<<30`). Guard restored; tree clean at `5ac5ef89`.
- **Message integrity:** the test's two assertions (`contains "genesis bootstrap"` AND `contains "different graph authority"`) match the `bail!` paragraph verbatim, and both substrings uniquely identify the guard. First run commits only generation 0 (`sync_bootstrap=1`/`begin=0`/`commit=0`, 0 docs) so the second run truly reaches the genesis-recovery path; positive control (`--no-graph` continuation: `begin→1`/`commit→1`, 1 doc) proves the guard does not corrupt pending state.
- `state.rs` sets `no_graph_genesis` true ONLY in the `kind == "sync_bootstrap"` arm, never clears it, threads it unchanged; the only production writer of a `sync_bootstrap` record is `begin_non_graph_generation` under `cfg.no_graph` (all other writers are `#[cfg(test)]`), so no graph-mode run can falsely trip the guard.
- **Non-blocking (nit):** the `bail!` recovery message names `--no-graph` and `rebuild with a new --state-dir/--prefix`, but omits `--fresh` as a valid recovery route even though the guard exempts it (`!cfg.fresh`). The two named routes are valid — a completeness nit, not a wrong instruction.

---

**Decision:** 0 BLOCKING findings, 18/18 checks green (0 pending/failing), `mergeStateStatus=CLEAN`, head at the exact verified SHA `5ac5ef89`. **Merging.** The two non-blocking nits (dry-run-over-genesis asymmetry; `--fresh` absent from the recovery message) will be filed as a single follow-up issue.

Issue #346, opened 2026-08-13
Resume journal reports skipped files below the noise floor (split out of #326)


xerj-org:
Split out of #326, which was closed because its **primary** defect — the `query_string` multi-byte panic — is fixed in `c7fde02f` and shipped in v1.0.0-rc.16. This secondary finding from the same report is **not** fixed, and would otherwise have been closed along with it.

Verbatim from #326:

**2. Resume journal reports skipped files below the noise floor.** A re-run over a changed tree prints

```
autoindex: 3 file(s) appeared after the resume plan was frozen and were NOT indexed
```

as one line inside the `xerj-progress` stream, and exits 0. For a "make the index match the tree" workflow that reads as success with a silently incomplete index. The `--fresh` hint is there, but it is below the attention threshold — surfacing the count in the final `xerj-done` line, or exiting non-zero, would make it visible.

Status: not investigated beyond confirming the parser fix did not address it. Filed so it is not lost.

xerj-org:
Investigated on `v1.0.0-rc.16` / main `ca4d75a`. **The symptom this issue describes is fixed. It has been replaced by a worse one**, so I would keep this open and retitle rather than close it.

## The original complaint no longer reproduces

The report was: skipped files are announced inside the `xerj-progress` stream, below the attention threshold, and the run **exits 0** — success with a silently incomplete index.

That is no longer what happens. It now fails loudly, which is what the issue asked for:

```
$ xerj autoindex ./j346 --prefix j346 --no-semantic --no-graph   # 3 files
xerj-done ok=true exit=0 reason=completed files=3 records=3 generation=1

# add 3 more .txt files to the same folder, re-run the same command
$ xerj autoindex ./j346 --prefix j346 --no-semantic --no-graph
xerj-done ok=false exit=1 reason=aborted wall=0.1s
error: project new file f4.txt: no frozen dataset accepts family txt-prose, group None,
       and fields ["body", "title"]; this requires unsupported dataset/schema evolution
       (use a new prefix)
$ echo $?
1
```

Loud, non-zero, and the count is in the final line. Request satisfied.

## What replaced it is a dead end

Adding files to a folder and re-running is the ordinary "keep my index matching the tree" action, and it now **aborts the whole run**. Nothing is indexed — 3 documents live against 6 files on disk. The suggested remedy is "use a new prefix", i.e. abandon the index and rebuild under a different name.

Trying `--fresh` does not help, and the two messages directly contradict each other:

```
[...]

PR #712, opened 2026-08-24
fix(autoindex): #346 - incremental reconcile compared a new document's raw family against the frozen "docs" label

Vinz2168:
Addresses #346 (split out of #326 — the visibility/exit-code half of that finding was already fixed independently; this is the live defect a reviewer found still blocking the everyday "add a file, re-run" workflow, described in [this comment](https://github.com/xerj-org/xerj/issues/346#issuecomment-5298868471)).

## Problem

Re-running `xerj autoindex` over a folder with one new file added — the ordinary "keep my index matching the tree" action — aborts the whole run when the new file is a plain document (prose, markdown, one-off config):

```
xerj-done ok=false exit=1 reason=aborted
error: project new file f4.txt: no frozen dataset accepts family txt-prose, group None,
       and fields ["body", "title"]; this requires unsupported dataset/schema evolution
       (use a new prefix)
```

even when the new file is byte-for-byte the same shape as the dataset already sitting there. `--fresh` does not help either (it refuses to discard a committed generation). There is no documented recovery short of abandoning the index under a new `--prefix`.

## Root cause

Two different frozen fields get compared as if they were one.

- `PlanDataset.family` (the frozen **dataset's** label) is collapsed to the literal `"docs"` for every document-shaped cluster at fresh-plan build time (`lib.rs`, mirroring `dataset::cluster`'s own "is this a document" test: no data-derived `key_fields`, no sub-file `group`).
- `FileAssignment.family` (the frozen **per-file** record) keeps the file's raw sniffed family (e.g. `"txt-prose"`).

Confirmed directly from a live repro's journal — for the same run, `PlanDataset.family == "docs"` while `FileAssignment.family == "txt-prose"`.

`reconcile_plan.rs` has two comparisons: `retained_slug` (an already-known file) correctly compares raw-vs-raw against `FileAssignment.family`. `classify_new` (a brand-new file) compares against `PlanDataset.family`, but was passed the raw sniffed family too — so any incrementally-added document file matched **zero** frozen datasets, regardless of how well it actually fit.

## Fix

`classify_new`'s family argument is now normalized the same way `lib.rs` normalizes it at fresh-plan build time: `"docs"` when the sketch has no `key_fields` and no `group`, the file's raw family otherwise. `retained_slug` is untouched — its raw-to-raw comparison was already correct.

## Testing

- Reproduced live against a debug build (rc.18): a `--no-graph` corpus of 3 prose files, one 4th file added. Before the fix: abort, exit 1, 0 records reachable through any documented recovery. After: reconciles normally — generation 2 committed, 4 records live, exit 0 — and a further no-op re-run confirms without a spurious commit.
- The `reconcile_plan` unit-test fixtures shared a `scan()` helper that built jsonl (data-family) scans with an **empty** `key_fields` set — realistic for a document, not for structured data, and it happened to work before this fix only because `classify_new` never looked at `key_fields`/`group`. Corrected to derive `key_fields` from the fields under test, matching what a real jsonl sketch reports, and added a `#346`-specific test (`new_document_shaped_file_matches_the_frozen_docs_dataset`) for the document-shaped path the fixture couldn't previously exercise. Verified fail-before: with the normalization removed, that new test fails with the exact reported error text.
- `cargo test -p xerj-autoindex --profile ci-test -- --test-threads=2`: 675 passed; the 19 failures are pre-existing environment issues (macOS non-UTF-8 filename handling + parallel-run ordering artifacts) that reproduce identically on unmodified `main`.

### DIFF
--- a/engine/crates/xerj-autoindex/src/dataset.rs
+++ b/engine/crates/xerj-autoindex/src/dataset.rs
@@ -87,7 +87,15 @@ fn jaccard(a: &HashSet<&str>, b: &HashSet<&str>) -> f64 {
 /// single-record files? Self-describing structured formats only: CSV is
 /// tabular by nature, logs/jsonl are collections by nature, and sql groups
 /// carry a real table schema.
-fn demotable_family(f: Family) -> bool {
+///
+/// `pub(crate)`: `reconcile_plan::classify_new` mirrors this same shape test
+/// for an incrementally-added single file (#346 review) — it cannot reuse
+/// `cluster`'s aggregate demotion loop (multi-file membership depends on
+/// corpus order, which incremental reconcile deliberately does not
+/// replay), but a lone new file's OWN eligibility is exactly this
+/// predicate plus `members.len() == 1 <= DOC_DEMOTE_MAX_FILES`, which is
+/// unconditionally true for one file.
+pub(crate) fn demotable_family(f: Family) -> bool {
     matches!(f, Family::Json | Family::Yaml | Family::Xml)
 }
 
--- a/engine/crates/xerj-autoindex/src/reconcile_plan.rs
+++ b/engine/crates/xerj-autoindex/src/reconcile_plan.rs
@@ -164,10 +164,30 @@ pub(crate) fn reconcile_plan(
             let group = &sketch.group;
             let fields = document_fields.as_ref().unwrap_or(&sketch.fields);
             let slug = if let Some(owner) = retained {
+                // `FileAssignment.family` (the frozen per-FILE record) keeps
+                // the raw sniffed family — the same value `sniffed.family`
+                // recomputes here — so the two compare directly.
                 retained_slug(owner, group, sniffed.family.as_str(), fields, &datasets)
                     .with_context(|| format!("project retained file {}", file.rel))?
             } else {
-                classify_new(group, sniffed.family.as_str(), fields, &previous.datasets)
+                // #346: `PlanDataset.family` (the frozen DATASET's label) is
+                // a different value — a fresh plan's dataset construction
+                // collapses every "no data-derived name, no sub-file group"
+                // sketch's family to the literal `"docs"`
+                // (`dataset::cluster`'s own is-docs definition, mirrored
+                // here — see `lib.rs`'s `PlanDataset` build). Comparing the
+                // raw sniffed family against that frozen `"docs"` label
+                // instead made every incrementally-added document file
+                // (prose, markdown, one-off config, …) match zero frozen
+                // datasets and fail closed as "unsupported dataset/schema
+                // evolution" — even one byte-for-byte the same shape as the
+                // dataset it belongs to.
+                let family = if sketch.key_fields.is_empty() && group.is_none() {
+                    "docs"
+                } else {
+                    sniffed.family.as_str()
+                };
+                classify_new(group, family, fields, &previous.datasets)
                     .with_context(|| format!("project new file {}", file.rel))?
             };
             assignments.push((group.clone(), slug));
@@ -476,8 +496,15 @@ mod tests {
             }),
             sketches: vec![crate::GroupSketch {
                 group: None,
+                // A real jsonl sketch's `key_fields` is the record's own
+                // field names read out of the file — non-empty is what makes
+                // `dataset::cluster` (and, since #346, this module) treat it
+                // as schema'd data rather than a `"docs"` document. Every
+                // fixture in this module models structured jsonl content, so
+                // mirror that instead of the empty set a real document sketch
+                // would carry.
+                key_fields: fields.keys().cloned().collect(),
                 fields,
-                key_fields: std::collections::HashSet::new(),
                 records: 1,
             }],
             junk: None,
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `82466248`)

**Result: 1 BLOCKING finding → not merging.** The core fix is correct for its primary case (document files with empty `key_fields`), fail-before is genuine, and the full `xerj-autoindex` suite is green — but the is-docs mirror is **incomplete**, and an independent probe test proved a realistic incremental run still wrongly aborts on the exact shape the commit message claims to fix. Verdicts verbatim below.

### 1. Correctness / is-docs parity → **BLOCK**
- **Fail-before (confirmed):** reverting only the fix hunk (`classify_new(group, sniffed.family.as_str(), …)`, keeping the fixture change + new test) makes `new_document_shaped_file_matches_the_frozen_docs_dataset` FAIL (reconcile aborts, no "docs" match); restoring → passes. Full autoindex lib suite green (697/0) at HEAD.
- **BLOCKING — the is-docs mirror is incomplete (demotion branch not covered):** a fresh plan labels a dataset `"docs"` when `c.is_docs` (lib.rs:1898-1904), which becomes true via **two** paths in `dataset::cluster`: (1) the direct test at dataset.rs:119 `key_fields.is_empty() && group.is_none()` — which the fix mirrors exactly — **and (2) the demotion loop at dataset.rs:180-193**, which folds a `demotable_family` (Json/Yaml/Xml, dataset.rs:90-92), group-less, ≤8-member, one-record-each cluster into the scope's docs cluster. A demoted cluster has **non-empty `key_fields`** by construction, so the fix's `sketch.key_fields.is_empty() && group.is_none()` is **false** for it → keeps the raw family (`"json"`) → `classify_new` filters `dataset.family == "json"`, finds none (the frozen one is `"docs"`), and bails *"unsupported dataset/schema evolution (use a new prefix)"*.
  - **Proven with a probe:** frozen plan with a single `family="docs"` dataset whose specs `(host, port)` exactly equal a new `config.json`'s fields (`Family::Json`, non-empty `key_fields`) → `reconcile_plan` aborted: *"project new file config.json: no frozen dataset accepts family json, group None, and fields [\"host\",\"port\"]; …use a new prefix"*. Family is the **sole** cause (fields matched). A **fresh** plan over the same folder demotes that config into `"docs"` and indexes it — so **incremental diverges from fresh and the run is wrongly aborted**. This is a common operation (adding a `.json/.yaml/.xml` to an already-indexed docs folder — the #173 demotion feature exists for exactly this), and the commit message explicitly lists *"one-off config"* among the shapes it fixes. Note the new-file path also never re-samples `as_document` (gated on `retained`, reconcile_plan.rs:155-160), so fully fixing this needs more than a family-label collapse.
- **should-fix — the fix newly exposes cross-scope document routing:** fresh clustering picks the docs cluster by **scope** and emits per-scope `"docs"`/`"<scope>-docs"` datasets (all `family="docs"`). `classify_new` matches on family+group+specs but **never scope**. Before the fix a new document never matched any `"docs"` dataset (aborted on raw-family mismatch); **after** the fix it matches `family="docs"`, so with >1 frozen `"docs"` dataset (nested `.git` repos under the autoindex root) identical specs → "refusing ambiguous assignment" abort, or differing specs → assignment into the **wrong scope's index** (wrong results). Narrow layout, hence non-blocking — but a real divergence the fix newly exposes.
- **nit:** the new test only models the empty-`key_fields` path; the demotion path (the blocking gap) is untested.

- Fail-before independently reproduced (reverting line 185 → new test fails with the #346 error). `FMT=0`, `CLIPPY=0` (all-targets, -D warnings), `BUILD=0` (ci-test); **711 passed / 0 failed / 0 ignored** across all binaries under 1.97.1. The macOS-only "19 pre-existing env failures" cited in the PR body do **not** reproduce on Linux (clean here). Never touched :9200.
- **nit:** `new_field_and_incompatible_type_fail_before_plan_changes` asserts only the generic `"unsupported dataset/schema evolution"` string, which `classify_new` emits for *both* a family-mismatch bail and a field-incompatibility bail — the fixture change correctly routes it through the field-incompat path (meaning strengthened, not masked), but the assertion string can't distinguish the two causes.

### 3. Evidence & is-docs-claim integrity → **MERGE_OK**
- Fail-before proven; full suite green; fmt/clippy 0.
- **nit:** the fix comment (reconcile_plan.rs:180-182) lists *"one-off config"* among the document shapes routed to the docs dataset, but the fix branch requires **empty** `key_fields` — so config files are **not** actually routed. (This is the same divergence the correctness skeptic escalated to blocking with a probe.)

---

## Decision: **changes requested — not merging** (merge gate is 0 BLOCKING; there is 1).

The fix is a genuine improvement for the primary case and is safe (no over-collapse: schema'd/grouped files still use the raw family), but as written it does **not** deliver what its commit message claims (`"one-off config"`) and a realistic incremental operation still wrongly aborts. To land, please either:

1. **Complete the mirror** — route a group-less, single-record, `demotable_family` (Json/Yaml/Xml) new file to `"docs"` too (mirroring dataset.rs:180-193), and address that the new-file path never re-samples `as_document` (so the demoted config is actually indexed as a document, not just label-matched); **and**
2. **Make `classify_new` scope-aware** so the newly-exposed cross-scope routing can't mis-home a document in nested-repo layouts; **or**
3. If config support is out of scope for this PR, **narrow the commit message** (drop "one-off config"), add a demotion-path test asserting the current abort, and file a follow-up for the config/scope cases — so #346 isn't closed while that shape still aborts.

Happy to re-run the 3-skeptic pass once updated.

Vinz2168:
Thanks for the thorough pass — the blocking finding was real and the probe reproduced cleanly.

Pushed a follow-up commit addressing it:

- `dataset::demotable_family` is now `pub(crate)`, and `reconcile_plan` mirrors the demotion path too, not just the direct is-docs path: a new file is now docs-eligible when `demotable_family(family) && group.is_none() && records == 1` (the file-level equivalent of `cluster`'s aggregate demotion — a lone new file's own membership is always `1 <= DOC_DEMOTE_MAX_FILES`).
- That flag now also drives `document_fields`/`as_document` for a **brand-new** file (previously only a carried-forward `retained` decision could set it) — without this half, the match would still fail, since a config file's raw fields don't overlap a real `docs` dataset's `title`/`body` specs at all.
- New test `new_one_off_config_file_demotes_into_the_frozen_docs_dataset` writes a real file to a tempdir (needed — `document_sample` reads disk) and reproduces your exact probe shape; verified fail-before against your reported error string.
- Also verified live end-to-end (real server + CLI, `--no-graph`): the abort is gone, generation commits, no-op re-run confirms.

One thing I found *while* verifying that live repro, reported separately as **#722** rather than folded in here: even after this fix correctly *routes* a demoted file into the `docs` dataset, the `--no-graph` generated pipeline's own extraction step (`sync_executor.rs`) never checks `as_document` at all — only the legacy graph path's phase-B dispatch does (`lib.rs` ~4862) — so the *published* content still comes from the raw family extractor, not `extract_as_document`. Confirmed pre-existing and unrelated to this PR: reproduces identically on a **fresh** `--no-graph` run with no reconcile involved. This PR doesn't regress or mask it — it was equally broken before — but it's real and worth its own fix, hence the separate issue rather than scope-creeping this one.

Full re-verification on the pushed commit: `cargo test -p xerj-autoindex --profile ci-test -- --test-threads=2` (676 passed, only the already-documented 19 pre-existing macOS environment failures), `clippy --all-targets -D warnings` clean, `fmt --check` clean, `cargo check -p xerj-server` clean.

Happy to take another pass whenever convenient.

xerj-org:
Closing this one — thank you for the fix; the **document-shaped** case (empty `key_fields`) it handles is correct and the analysis was sound. The 3-skeptic review (posted above) confirmed a **blocking incompleteness**, though: a group-less, single-record, `demotable_family` (JSON/YAML/XML) **config** — the shape the commit message also claims to cover — still hits the exact #346 `use a new prefix` abort, because a fresh plan *demotes* it into the scope's docs dataset and indexes it **as a document** (`build_phase_a` re-samples via `extract_as_document` + sets `as_document`, lib.rs:1919-1962), which the family-label collapse alone does not replicate.

A complete fix that also (a) routes a demoted new config to `docs`, (b) sets `as_document` + re-samples through `document_sample`, and (c) makes `classify_new` accept the rendered document shape, is in progress on `fix/346-demotable-config-as-document`. Reopen or ping if you'd like to take that forward instead.

**#346 stays open** to track the remaining work — closing only this PR, not the issue.

Issue #456, opened 2026-08-17
CI installs `rustup toolchain install stable`, so the compiler can change without a commit — and 10 jobs each refetch the action, which rate-limited main today


xerj-org:
Two related properties of `.github/workflows/ci.yml`, found while gate-auditing a run that failed on a docs-only diff.

## 1. The toolchain is not pinned

Nine jobs use `dtolnay/rust-toolchain@stable` and one uses `@nightly`. The action then runs, verbatim from today's green main run:

```
rustup toolchain install stable --component rustfmt --component clippy --profile minimal --no-self-update
```

So **the compiler CI validates against is whatever `stable` resolves to on the day the job runs.** It is `1.97.1` right now. When upstream promotes the next release, every job on every branch silently moves with it — no commit, no review, nothing in the diff. A new lint, a new `-D warnings` failure, or a behaviour change lands on `main` with no change to the repository.

This is not hypothetical drift in the abstract: the gap between CI's compiler and a local default is already large enough that `cargo` passing locally means nothing, and anyone reasoning about a green run has to know the resolved version out of band. Pinning it makes that legible and makes toolchain bumps a reviewable commit.

Suggested: pin the version explicitly (`dtolnay/rust-toolchain@1.97.1`, or a `rust-toolchain.toml` at the root so local and CI agree by construction), and treat a bump as its own PR. Keep `@nightly` where nightly is genuinely required (the fuzz job).

## 2. Ten jobs each fetch the same action, and today that hit GitHub's rate limit

Every job re-downloads the action tarball from `codeload.github.com`. With ~15 jobs per run and several PRs in flight, that is enough concurrent identical fetches to get throttled. Three runs failed this way today, on diffs that could not possibly have caused them:

```
PR #441  Build + Test                       docs-only diff (2 markdown/HTML files)
PR #452  ONNX feature compile + tests       docs-only diff
PR #452  Autoindex FD smoke (windows)       docs-only diff
PR #453  Autoindex FD smoke (macos)         es_compat.rs only

##[error]Response status code does not indicate success: 429 (Too Many Requests).
##[error]Failed to download archive
  'https://codeload.github.com/dtolnay/rust-toolchain/tar.gz/4360b52568e2003a75bf9bc1d59f33a8e3fc893c'
  after 3 attempts.
```

One also failed with `503 (Service Unavailable)`, and the GitHub GraphQL API was returning `503` in the same window, so part of this is an upstream wobble rather than purely our fan-out.

[...]

PR #713, opened 2026-08-24
ci(#456): pin every workflow action to a commit SHA, and the rust toolchain to 1.97.1

Vinz2168:
Addresses #456.

## 1. The toolchain was not pinned

Nine jobs used `dtolnay/rust-toolchain@stable` and one used `@nightly`. That action's git **ref** is the mechanism that selects the installed rustc version — each version-named branch of the action repo bakes in a fixed `toolchain:` env value (confirmed by reading `action.yml` on the `1.97.1` branch); `stable` resolves to rustup's `stable` channel on the day the job runs. So the compiler CI validates against could move with no commit, no review, nothing in the diff.

Verified against the issue's own quoted 429 log: `dtolnay/rust-toolchain@stable` resolves to commit `4360b525...`, the exact SHA the issue's log shows being fetched (`gh api repos/dtolnay/rust-toolchain/commits/stable`).

## 2. Nothing was pinned by SHA

None of the ~38 `uses:` lines in `ci.yml` were pinned, so every job re-fetches whatever the tag currently points to.

## Fix

Every `uses:` in `ci.yml` now pins a commit SHA, with the resolved tag as a trailing comment for humans (GitHub's own supply-chain-hardening convention). `dtolnay/rust-toolchain@stable` → the SHA of that repo's `1.97.1` branch — the version already in use everywhere else in this tree (`rustc --version` locally, and "Compiles under 1.97.1" shows up in several recent commit messages). `dtolnay/rust-toolchain@nightly` (the fuzz job) keeps floating **on purpose** — nightly drift is the point there — so only its action ref is pinned, not what "nightly" resolves to on the day the job runs, matching the issue's explicit ask ("Keep `@nightly` where nightly is genuinely required").

Every other action (`actions/checkout`, `Swatinem/rust-cache`, `actions/cache`, `actions/upload-artifact`, `actions/setup-node`) is pinned the same way, at each one's current tag tip. Added a short comment above `jobs:` explaining the convention so a future bump is legible as a real PR rather than a mystery diff.

## Scope

Only `.github/workflows/ci.yml` — the file the issue measured and quoted from. The repo's other four workflow files also use unpinned tags but weren't part of the issue's own measurement, so widening beyond `ci.yml` is left to a follow-up rather than scope-creeping this fix.

## Not verified (same as the issue's own "Not verified" section)

Whether this actually reduces 429s in practice. Pinning by SHA doesn't change what gets fetched from `codeload.github.com` per job (the archive URL is already keyed by SHA even for a tag ref, as the issue's own log line shows), so this PR's confirmed effect is #1 (compiler drift) and general supply-chain hardening; its effect on #2's rate-limit fan-out is the same open question the issue left open — not something this PR claims to close.

## Verification

- Exact per-action `uses:` occurrence counts unchanged before/after the edit (14 checkout, 9 `rust-toolchain@stable`→`1.97.1`, 1 `@nightly`, 8 rust-cache, 2 cache, 1 upload-artifact, 3 setup-node).
- `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/ci.yml'))"` — parses clean.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,16 @@ env:
   FORCE_JAVASCRIPT_ACTIONS_TO_NODE24: "true"
   CARGO_TERM_COLOR: always
 
+# #456: every `uses:` below is pinned to a commit SHA (with the resolved
+# tag/version as a trailing comment for humans), including `dtolnay/rust-
+# toolchain`, whose git ref IS the rustc version it installs. Before this,
+# `@stable` meant "whatever rustup resolves as stable on the day the job
+# runs" -- the compiler CI validates against could move with no commit, no
+# review, and nothing in the diff. Bumping any of these now takes an actual
+# PR, the same as bumping a Cargo dependency. `dtolnay/rust-toolchain@nightly`
+# (the fuzz job) is the one exception left floating: nightly drifting is the
+# point there, so only its ACTION ref is pinned, not what "nightly" resolves
+# to on the day the job runs.
 jobs:
   cla-config:
     name: CLA gate config
@@ -30,7 +40,7 @@ jobs:
       contents: read
       pull-requests: read
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.3.0
 
       # cla-bot reads .contributors straight off this repo; malformed JSON or a
       # missing bot exemption breaks the gate silently (it just flags everyone).
@@ -68,7 +78,7 @@ jobs:
     name: Reference-coding toolkit
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.3.0
 
       - name: Shell syntax
         run: |
@@ -100,7 +110,7 @@ jobs:
     name: Landing constants guard
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v4
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.3.0
         with:
           fetch-depth: 0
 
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent verification (supply-chain SHA-resolution + completeness audit)

For a workflow-pinning PR the load-bearing risk is **does each pinned SHA actually correspond to the claimed official release** (not cargo behaviour), so this was verified deterministically by resolving every pin via the GitHub API rather than with worktree cargo skeptics.

### Pin integrity (SHA → claimed version)
| action | pinned SHA | comment | resolves to | verdict |
|---|---|---|---|---|
| `actions/checkout` | `11d5960…` | `# v4.3.0` | **v4.4.0** | ⚠ comment stale (SHA is a valid newer release) |
| `dtolnay/rust-toolchain` | `032958af…` | `# 1.97.1` | 1.97.1 | ✅ |
| `Swatinem/rust-cache` | `6323deb…` | `# v2.8.1` | **v2.9.2** | ⚠ comment stale (SHA is a valid newer release) |
| `actions/cache` | `0057852b…` | `# v4.3.0` | v4.3.0 | ✅ |
| `actions/upload-artifact` | `ea165f8d…` | `# v4.6.2` | v4.6.2 | ✅ |
| `actions/setup-node` | `49933ea5…` | `# v4.4.0` | v4.4.0 | ✅ |
| `dtolnay/rust-toolchain` (nightly job) | `7c8d7d13…` | `# nightly (action ref pinned; toolchain floats)` | real commit (2026-08-05) | ✅ |

### Completeness
- Every `uses:` in the post-PR `ci.yml` is SHA-pinned with a version comment — **no unpinned action remains**.
- Job count unchanged **19 → 19** — no job dropped or renamed; the diff is ref-swaps only.
- Toolchain consistent: 9× `# 1.97.1` on the stable jobs, `# nightly` only where intended.
- PR's own CI: 18/18 green — the pinned (newer) `checkout`/`rust-cache` commits run cleanly, so no functional regression.

### Finding — 1 should-fix, 0 blocking
The two ⚠ rows: the pinned SHAs are **legitimate official releases** (checkout `v4.4.0`, rust-cache `v2.9.2`) — the immutability/security intent of #456 is fully achieved — but the `# vX.Y.Z` comments name the wrong (older) version. Since this PR's deliverable is precisely an auditable pin→version map, the stale comments are worth correcting, but they are a documentation defect with **no functional or security impact** (valid immutable pins, green CI), so per our merge rule (should-fix/nit do not block) this does **not** hold the merge.

**Decision:** 0 BLOCKING, green, CLEAN → **merging**, and filing a follow-up to correct the two version comments (`11d5960 → v4.4.0`, `6323deb → v2.9.2`) or re-pin to the intended older tags if that was the intent.

Issue #384, opened 2026-08-15
Widen the default-parallelism CI gate from --lib to the integration targets


xerj-org:
Follow-up to #372, filed by the agent that fixed it (Claude Opus 5).

#372 added one CI step that runs at default parallelism:

```yaml
- name: Engine lib tests at default parallelism (contributor's `cargo test`)
  working-directory: engine
  run: cargo test --profile ci-test -p xerj-engine --lib
```

That closes the hole for one crate's lib tests only — the only scope measured green
(10 dev-profile and 5 ci-test-profile clean runs). Everything else still runs solely
at `--test-threads=2`: the other 15 crates' lib tests, and the integration targets
(`engine/crates/*/tests/*.rs`). The integration targets are the ones most likely to
carry the same class of defect — they bind ports, share fixture directories, and
start real servers.

## What this issue asks for

Widen the default-parallelism step from `-p xerj-engine --lib` toward
`cargo test --profile ci-test --workspace`, one measured step at a time, or document
per-target why a target must stay serialised. Never widen it to a scope that is red:
a gate merged red gets disabled, which is worse than not having it.

## Known blocker

`--workspace --lib` at default parallelism is **currently red**: #385, where
xerj-autoindex fails 5 runs in 10 because two test modules share the global
`SNAPSHOT_FAILPOINT` but not the `SNAPSHOT_TEST_LOCK` that guards it. #385 must land
before the `-p xerj-engine` scope can widen to `--workspace`.

Whether the *integration* targets are green at one-thread-per-core is still an open
question, not an assumption — nobody has measured it. Anyone picking that part up
should run `cargo test --profile ci-test --workspace` at default parallelism several
times on a clean checkout and report the failures verbatim.

## Context

#372 was the third defect in one cycle that was real on a clean checkout and
invisible to CI because CI runs a different configuration (the others: #359, clippy
lints that only fire on Rust 1.92; #353, a Painless stack overflow that only
reproduces off the `ci-test` profile). The general principle worth adopting is that
CI should run at least one configuration that matches what a contributor actually
types.

PR #715, opened 2026-08-24
ci(#384): widen the default-parallelism gate from -p xerj-engine --lib to --workspace --lib

Vinz2168:
Addresses #384.

## Background

#372 added a default-parallelism CI step scoped to `-p xerj-engine --lib` only — the one scope measured green at the time. #384 asked to widen it, one measured step at a time, and named a known blocker: **#385**, where xerj-autoindex failed 5/10 default-parallelism runs because two test modules shared the global `SNAPSHOT_FAILPOINT` without holding the `SNAPSHOT_TEST_LOCK` that guards it.

**#385 is fixed** (merged in #429), so the blocker this issue named for widening past `xerj-engine` is gone.

## Measurement

Per the issue's own instructions ("run `cargo test --profile ci-test --workspace` at default parallelism several times on a clean checkout and report the failures verbatim"):

- `cargo test --profile ci-test --workspace --lib` (no `--test-threads`, one thread per core — 10 on this box), run **3 times**: 674 passed / 19 failed in `xerj-autoindex`, **0 failures anywhere else**, byte-for-byte identical failing-test list all 3 runs.
- `cargo test --profile ci-test --workspace --lib --no-fail-fast`, once, for genuinely full coverage — `cargo test`'s default fail-fast stops at the *first* failing binary, so a plain `--workspace --lib` run never reaches crates ordered after the first failure. This rerun is the one that actually exercises all 16 crates: the same 19 `xerj-autoindex` failures, **plus 2 in `xerj-engine`** (`snapshot_path_security_tests`) not previously visible in this scope.

## Why these are excluded

Both failure classes reproduce byte-for-byte at `--test-threads=2` too (this workflow's existing step, immediately above the one this PR changes) — so they are deterministic local-filesystem quirks, not the interleaving-dependent defect class this gate exists to catch:

- **`xerj-autoindex`**: writes a non-UTF-8 file name — accepted on Linux ext4, rejected by macOS APFS with `Illegal byte sequence` (content/inventory tests) — plus one unrelated pre-existing data-assertion mismatch (`sync_executor`) already present on a clean, unmodified `main`.
- **`xerj-engine`**: `snapshot_path_security_tests` compares a raw tempdir path against its canonicalized form; only macOS's `/var` → `/private/var` symlink makes those disagree.

Net result: across **all 16 crates'** lib tests, at default parallelism, with full (`--no-fail-fast`) coverage, **zero new parallelism-specific failures** beyond what the existing `--test-threads=2` gate already knows about.

## What this does NOT claim

`--workspace --lib` is not `--workspace`. The integration targets (`engine/crates/*/tests/*.rs`) remain the issue's own separately-tracked open question — they bind ports and share fixture directories, the shape most likely to carry the next #372-class defect — and this PR does not claim to have measured them.

Also honest about reach, matching the step's own existing comment style: this was measured on **macOS (10 cores)**, not `ubuntu-latest`. A defect that only interleaves at 2-4 threads could still be invisible here the same way `--test-threads=2` hid #372 in the first place. This step is where such a defect would surface for the first time on the actual runner — **this PR's own CI run is the real validation on that platform**, not a claim that measuring it locally proves the absence of one. If CI's own run of this step is red, that is itself useful data and the step should not merge until it's addressed.

## Change

- `- name: Engine lib tests at default parallelism` → `- name: Workspace lib tests at default parallelism`
- `run: cargo test --profile ci-test -p xerj-engine --lib` → `run: cargo test --profile ci-test --workspace --lib`
- Comment rewritten to document the measurement above and its honest limits.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,24 +279,44 @@ jobs:
       # runner. It is cheap insurance and the place a wider, better-sized gate
       # (#384) can land -- not a claim that the class is now covered.
       #
-      # Scoped to the one crate whose lib tests have been *measured* green at
-      # default parallelism (10 dev-profile and 5 ci-test-profile clean runs at
-      # this commit, 20 cores). `--workspace --lib` remains the destination,
-      # but the broader gate is tracked separately in #384.
+      # #384 widened this from `-p xerj-engine --lib` to every crate's lib
+      # tests: #385 (the blocker this issue named -- xerj-autoindex stealing
+      # another test module's global SNAPSHOT_FAILPOINT without holding its
+      # lock) is fixed, and `--workspace --lib` at default parallelism is now
+      # *measured* clean of anything parallelism-specific -- 3 clean-main
+      # `--no-fail-fast` runs (all 16 crates, every lib test binary actually
+      # invoked, not just the ones before the first failure) plus 3 more
+      # scoped to `--lib` alone, all bit-for-bit identical, on a 10-core box.
       #
-      # Historical context for this scope: before #385, xerj-autoindex was red
-      # at default parallelism because `incremental_reconcile_http_tests` could
-      # reach the global `SNAPSHOT_FAILPOINT` without holding the
-      # `SNAPSHOT_TEST_LOCK` and steal another test's injected failure. The
-      # thread-local failpoint fix in this commit removes that defect; widening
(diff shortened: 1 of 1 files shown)


Vinz2168:
Closing in favor of #716, which addresses the same issue (#384) essentially identically and was opened by the project's own automation shortly after this one. No need for two open PRs solving the same thing.

PR #716, opened 2026-08-24
ci: widen the default-parallelism gate to --workspace --lib (#384)

xerj-org:
Widens the #372 default-parallelism CI gate (ci.yml) from `-p xerj-engine --lib` to `--workspace --lib`.

**Why now:** #385 (the thread-local `SNAPSHOT_FAILPOINT` + `SNAPSHOT_TEST_LOCK` fix) removed the xerj-autoindex flakiness that #384 named as the blocker. 

**Measurement (the bar #384 sets — 'never widen to a scope that is red'):** ran `cargo test --profile ci-test --workspace --lib` at **default parallelism** (rustc 1.97.1, 20 cores) **6 times** — all 6 clean, 0 failed across all 15 crate lib suites. The step's own comment already named `--workspace --lib` as the destination.

**Verification:** this PR's own CI runs the widened step, so the gate is exercised at the new width before merge (and the gate can't be merged red). Scope stays lib-only; the integration targets (`tests/*.rs`) remain a separate, still-to-be-measured step per #384. Partially addresses #384 (lib scope); integration-target measurement tracked as the remaining part.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -279,24 +279,23 @@ jobs:
       # runner. It is cheap insurance and the place a wider, better-sized gate
       # (#384) can land -- not a claim that the class is now covered.
       #
-      # Scoped to the one crate whose lib tests have been *measured* green at
-      # default parallelism (10 dev-profile and 5 ci-test-profile clean runs at
-      # this commit, 20 cores). `--workspace --lib` remains the destination,
-      # but the broader gate is tracked separately in #384.
-      #
-      # Historical context for this scope: before #385, xerj-autoindex was red
-      # at default parallelism because `incremental_reconcile_http_tests` could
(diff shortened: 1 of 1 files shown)


xerj-org:
**#384 integration-target measurement (the tracked remainder).** After this lib-scope widening, I also measured the FULL `cargo test --profile ci-test --workspace` (all targets incl. integration) at default parallelism on the same 20-core box, rustc 1.97.1: **4/4 runs clean, 0 failed** across every suite. That is promising for extending the gate past `--lib` to all targets — but integration tests bind ports and share fixture dirs, so their parallelism-safety is runner-dependent (a gate merged red gets disabled, per #384). This PR therefore stays scoped to the CI-verified `--workspace --lib`; extending to the full `--workspace` will land as a **separate PR whose own CI validates the integration gate at the new width**, carrying this 4-run local evidence. Filing that as the remaining half of #384.

PR #720, opened 2026-08-24
test(engine): #643 — cover the Percentage/Field/Script msm unwrap (gap 1)

### LINKED ISSUE #643: bool score-neutrality: unwrap is conservative — rare 1-clause shapes still bypass it (follow-up to #399)

xerj-org:
## What

Closes #643 by adding the missing gap-1 test coverage. The issue's two conservative-unwrap gaps in `unwrap_single_clause_bool` are **both already implemented and correct** in the tree; the only remaining item was that gap 1 had no dedicated test.

## Why #643 is now fully addressed

- **gap 1 — `minimum_should_match` that resolves to 1** (`index.rs:35182-35184`): the code matches `Percentage(pct) => *pct < 200`. This is exactly right: msm resolves to `max(1, floor(len * pct / 100))`, so for the single `should` clause (`len == 1`) every `pct < 200` resolves to `1` (clause required → collapses to the bare clause), while `pct >= 200` resolves to `>= 2` (requires 2 of 1 clause → the bool matches nothing, so it must stay wrapped). `Field`/`Script` resolve only at query time and can exceed the clause count, so a static rewrite keeps them wrapped (the `_ => false` fall-through) — **correct by design; the issue's "resolve to an integer" direction is infeasible for these two**.
- **gap 2 — recursion through score-preserving wrappers** (`DisMax`, `FunctionScore`): already closed in #704, with `unwrap_recurses_through_{dis_max,function_score}`.

So the only true remainder was locking in gap 1 with a test.

## Tests added (`unwrap_single_clause_643_wrapper_tests`)

- `unwrap_collapses_lone_should_with_percentage_msm_resolving_to_one` — `0/1/50/100/150/199%` all erase to the bare clause;
- `unwrap_keeps_lone_should_wrapped_when_percentage_msm_exceeds_the_clause_count` — `200/250/400%` stay wrapped;
- `unwrap_keeps_lone_should_wrapped_for_query_time_msm` — `Field`/`Script` stay wrapped.

## Fail-before

Reverting only the `Some(MinShouldMatch::Percentage(pct)) => *pct < 200` arm (so `Percentage` falls through to `_ => false`) leaves the lone-`should` Percentage wrapped, and `unwrap_collapses_lone_should_with_percentage_msm_resolving_to_one` fails (`0% … got Bool { … }`). Restoring it → all pass.

- `cargo clippy -p xerj-engine --all-targets -- -D warnings` — clean
- full `cargo test -p xerj-engine` (dev) — **1090 passed / 0 failed** (62 binaries; 23 pre-existing ignores)

Test-only change; no source behaviour is modified.

Closes #643.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -47305,7 +47305,7 @@ mod wal_shard_settings_tests;
 #[cfg(test)]
 mod unwrap_single_clause_643_wrapper_tests {
     use super::*;
-    use xerj_query::ast::{BoolOperator, QueryNode};
+    use xerj_query::ast::{BoolOperator, MinShouldMatch, QueryNode};
 
     fn bare_match() -> QueryNode {
         QueryNode::Match {
@@ -47325,6 +47325,18 @@ mod unwrap_single_clause_643_wrapper_tests {
             must_not: vec![],
             filter: vec![],
             minimum_should_match: None,
+        }
+    }
+
+    /// A `bool` whose only clause is a lone `should` with the given
+    /// `minimum_should_match` — the shape #643 gap 1 is about.
+    fn single_should_bool(msm: Option<MinShouldMatch>) -> QueryNode {
+        QueryNode::Bool {
+            must: vec![],
+            should: vec![bare_match()],
+            must_not: vec![],
+            filter: vec![],
+            minimum_should_match: msm,
         }
     }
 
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `c7c96b9c`)

### 1. Correctness of assertions + fail-before → **MERGE_OK**
- **Fail-before (confirmed):** removing ONLY the `Some(MinShouldMatch::Percentage(pct)) => *pct < 200,` arm (Percentage falls to `_ => false`) makes `unwrap_collapses_lone_should_with_percentage_msm_resolving_to_one` FAIL (pct=0 stayed wrapped as `Bool`); restoring → all 5 unwrap_* tests pass. Non-vacuous.
- The unwrap arm (`pct < 200`) is checked against **both** runtime msm resolvers — `doc_matches_query` (index.rs:36151) and the FTS projection (index.rs:36037) — which both compute `floor(len*pct/100).max(1)`. For the single `should` clause (len==1), pct=199 resolves to 1 (collapse) and pct=200 resolves to 2 (requires 2 of 1 clause → matches nothing → stay wrapped). The tests assert exactly `[0,1,50,100,150,199]→erase` and `[200,250,400]→stay`.
- Field/Script staying wrapped is genuinely correct (resolve per-doc at query time, can exceed the clause count → a static rewrite cannot prove collapse) — conservative, not a masked defect.

- `FMT=0`, `CLIPPY=0` (-D warnings), `BUILD=0` (ci-test). **Full ci-test suite: 62 `test result: ok` blocks, zero FAILED/panicked.** The 3 new tests pass; the pre-existing `unwrap_recurses_through_{function_score,dis_max}` pass.
- Fail-before independently reproduced (revert the Percentage arm → collapse test fails, stay-wrapped + recursion tests still pass).

### 3. Evidence — does the test exercise what it claims? → **MERGE_OK**
- Each test runs `unwrap_single_clause_bool` on a lone-`should` `Bool` and discriminates `QueryNode::Match` (collapsed) vs `QueryNode::Bool` (wrapped); collapse test iterates 0/1/50/100/150/199%, wrapped test 200/250/400%, Field/Script locked stay-wrapped. Not tautological.
- **Boundary-sensitivity fail-before:** mutating the source `*pct < 200` → `*pct < 100` makes the collapse test FAIL at pct=100% — the test pins the exact boundary.
- PR-body claims verified true against the tree (gap-1 arm at 35183, runtime formula `max(1, floor(len*pct/100))` at 36151, Field/Script conservative, parser has no upper cap so 200% is reachable).
- **nit (non-blocking, pre-existing — NOT in this PR's diff):** `MinShouldMatch::Percentage`'s doc-comment (ast.rs:99) says "(0–100)" but the parser (parser.rs:844/3394, no upper cap) and matcher (index.rs:36151) accept >100 — the exact ≥200 case this PR tests. Code behaviour is correct; only the comment is narrow.

---

**Decision:** 0 BLOCKING, 18/18 checks green (0 pending/failing), `mergeStateStatus=CLEAN`, head at the verified SHA `c7c96b9c`. **Merging.** The one nit (pre-existing `Percentage` doc-comment range) will be filed as a follow-up.

Issue #587, opened 2026-08-22
autoindex/sniff: frontmatter lookahead (#551) routes the body through txt_kind only, mis-serving non-prose bodies and prose-heavy multi-doc YAML


xerj-org:
Follow-up to #551 / PR #586. The frontmatter lookahead classifies the post-`---` body with `txt_kind(body)` and gates on `yaml_like_count(body) * 2 < body.len()`. The 3-skeptic verification of #586 surfaced three edges — all non-blocking (no data loss; the #551 common case works and all existing tests pass), tracked here.

## 1. `txt_kind(body)` bypasses the structured branches
`classify_text` normally tries JSON/HTML/logs/SQL/CSV before falling to `txt_kind`. The lookahead jumps straight to `txt_kind(body)`, so a frontmatter file whose *body* is itself structured is classified as prose, not by its real family:
- frontmatter-over-CSV → `TxtProse` (a bare CSV twin → `Csv`)
- frontmatter-over-JSON → `TxtProse` (twin → `Json`)

No data loss (the body is still fully indexed as prose), and these are rare shapes, but "classified by its body" is only true for prose bodies today. Consider running the body through the full `classify_text` branch set (needs a body `&str`, or a small refactor to classify a `nonblank` slice) so the family matches the twin for every body type.

## 2. Prose-heavy multi-document YAML is a narrow regression
A fully parseable 2-document YAML stream whose second document (after the first `---` separator) is block-scalar/comment/prose-dominant now classifies `TxtProse`; on `main` it was `Family::Yaml` and both docs were extracted as structured records. Trigger is narrow — the guard requires the *entire* post-first-separator body to be majority non-YAML, which real k8s/Ansible multi-doc YAML (dominated by `key:`/`- ` lines) dilutes back to `Yaml` — and there is no data loss (text still indexed), only loss of per-doc key structure.

The robust discriminator the evidence lens identified: markdown-with-frontmatter is *invalid* as a full YAML stream (the markdown body halts `serde_yaml`'s iterator, yaml_x.rs:19-25), whereas multi-doc YAML parses. A cheap-parse check ("does the body parse as YAML?") would separate the two precisely, at the cost of a parse in the sniff path (currently heuristic-only by design — needs a decision).

## 3. Single-line frontmatter body still lost
[...]

PR #721, opened 2026-08-24
fix(autoindex): #587 item 1 — classify a frontmatter body by its structured family

xerj-org:
## What

Addresses **item 1** of #587. The #551 frontmatter lookahead classified the post-`---` body with `txt_kind(body)` only, bypassing the structured branches — so a frontmatter-over-CSV/JSON file was classified as a prose family (`TxtProse`) while its frontmatter-less twin was `Csv`/`Jsonl`. The body was still fully indexed, so no data loss, but "classified by its body" was only true for prose bodies.

## Fix

Extract the structured-family branches of `classify_text` (JSON/JSONL, HTML/XML, logs, SQL dump, CSV) into `classify_structured(text, nonblank) -> Option<Family>` — the **single source of truth** `classify_text` already ran for the whole file — and run the frontmatter body through it, falling to `txt_kind` only for a genuinely prose body:

```rust
let body_text = body.join("\n");
return classify_structured(&body_text, body).unwrap_or_else(|| txt_kind(body));
```

`classify_structured` deliberately does **not** look at YAML or the `---` marker, so classifying a body with it cannot recurse back into the frontmatter lookahead, and the top-level path is unchanged (verbatim extraction — every existing sniff test still passes).

## Fail-before

New test `frontmatter_over_a_structured_body_is_classified_by_that_body` asserts a frontmatter-over-CSV file classifies identically to its bare-CSV twin (`Csv`), and likewise JSONL. Reverting only the frontmatter body line to `return txt_kind(body)` makes the CSV case classify `TxtProse` and the test fails (`left: TxtProse, right: Csv`).

- full `cargo test -p xerj-autoindex` (dev) — **716 passed / 0 failed**
- ci-test: `cargo build --profile ci-test` — 0; `sniff::` tests — 49 passed / 0 failed

## Scope

This addresses **item 1** only. Items **2** (prose-heavy multi-document YAML — the robust discriminator is a YAML cheap-parse, which adds a parse to the currently-heuristic sniff path) and **3** (single-line frontmatter body) require a maintainer decision and remain open in #587.

### DIFF
--- a/engine/crates/xerj-autoindex/src/sniff.rs
+++ b/engine/crates/xerj-autoindex/src/sniff.rs
@@ -1033,6 +1033,57 @@ pub fn decode_text(bytes: &[u8]) -> (String, &'static str) {
 }
 
 fn classify_text(text: &str, nonblank: &[&str]) -> Family {
+    // Structured families (JSON/HTML/XML/logs/SQL/CSV) win first, for the whole
+    // file and — via the #551 frontmatter lookahead below — for a frontmatter
+    // body too.
+    if let Some(family) = classify_structured(text, nonblank) {
+        return family;
+    }
+
+    // YAML — but a leading `---` is also the opening delimiter of markdown/prose
+    // frontmatter (Jekyll, Hugo, Obsidian, Docusaurus, MkDocs, Astro, and Claude
+    // Code skill files — including the two this repo ships). Without a lookahead
+    // for the *closing* `---` and what follows it, every such file was handed to
+    // the YAML extractor, which keeps the frontmatter mapping and the first
+    // paragraph and discards the rest of the body (#551, silent data loss). If
+    // the file opens with `---`, has a matching closing `---`, and the body after
+    // it is majority non-YAML, classify by that body so the whole document is
+    // indexed. A pure/multi-document YAML file keeps YAML: the content after its
+    // `---` document marker is still YAML-like (or too short to be called prose).
+    let leads_with_marker = nonblank.first().map(|l| l.trim() == "---").unwrap_or(false);
+    if leads_with_marker {
+        if let Some(close_offset) = nonblank.iter().skip(1).position(|l| l.trim() == "---") {
+            let body = &nonblank[close_offset + 2..];
+            if body.len() >= 2 && yaml_like_count(body) * 2 < body.len() {
+                // #587: classify the body by its real structured family
+                // (JSON/CSV/HTML/logs/SQL), not prose-only, so a
+                // frontmatter-over-CSV/JSON body matches its frontmatter-less
+                // twin; fall to `txt_kind` only for a genuinely prose body.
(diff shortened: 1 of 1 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `5263a2c0`)

### 1. Behaviour-preserving extraction / frontmatter correctness / fail-before → **MERGE_OK**
- **Verbatim extraction confirmed** by diffing `5263a2c0^` against HEAD: the `ini_table_header`, JSON/JSONL, HTML/XML, logs, SQL, and CSV branches are byte-for-byte identical in condition and order; only `return Family::X` became `return Some(Family::X)` and the YAML/prose fall-through became `None`. The top-level whole-file path is unchanged (`classify_text` calls `classify_structured` first — same branches, same order — then the identical YAML frontmatter lookahead, then `Family::Yaml`, then `txt_kind`).
- `classify_structured` contains no `---`/YAML logic, so running it on a frontmatter body **cannot** re-enter the frontmatter lookahead (no recursion); a genuinely prose multi-doc YAML body returns `None` and falls to `txt_kind` exactly as before.
- **Fail-before (confirmed):** reverting only the frontmatter-body hunk (back to `return txt_kind(body)`) makes `frontmatter_over_a_structured_body` FAIL (`left: TxtProse, right: Csv` — the exact #587-1 bug); restored → all 11 `text_family_tests` pass. Non-vacuous.

- `FMT=0`, `CLIPPY=0` (-D warnings, --all-targets), `BUILD=0` (ci-test). Full ci-test suite: **702 lib + all integration binaries pass, TEST_EXIT=0, no flakes on the first parallel run.**
- Fail-before reproduced deterministically (`TxtProse` vs `Csv`).

### 3. Evidence → **MERGE_OK**
- `classify_structured` is a verbatim extraction; the frontmatter body now matches its frontmatter-less twin (CSV/JSONL verified). The guard `yaml_like_count(body)*2 < body.len()` still fires for CSV/JSON bodies (0 yaml-like lines). All 702 crate lib tests pass, incl. the #551 and multi-doc YAML guardrails unchanged. Scope limited to item 1; items 2/3 left open as documented. Fail-before genuine.

### Non-blocking nit (both correctness & regression skeptics, same observation)
The frontmatter body is reconstructed with `body.join("\n")` (nonblank lines only), so blank lines inside the body are collapsed before `classify_structured` sees `text`. Harmless for the probes used (HTML head, SQL substring, JSON start) since none depends on blank lines, but it's a subtle difference from the top-level call which passes the raw `text`. Non-blocking.

---

**Decision:** 0 BLOCKING. Merging once the last CI checks go green + `mergeStateStatus=CLEAN` (head at the verified SHA `5263a2c0`). The `join("\n")` nit is noted for the record — no separate follow-up warranted (behaviourally inert).

PR #723, opened 2026-08-24
fix(autoindex): #718 — offer --fresh in the genesis-guard recovery message + document the dry-run refusal

### LINKED ISSUE #718: autoindex: polish the #585 genesis cross-path guard — --dry-run projection + --fresh in the recovery message
Two **non-blocking** follow-ups surfaced by the independent skeptic review of #711 (which closed #585 case 2). Both live in the new genesis cross-path guard at `xerj-autoindex/src/lib.rs` (~3378: `if genesis_recovery && preflight.no_graph_genesis && !cfg.no_graph && !cfg.fresh { bail! }`). Neither affects correctness or safety — no mutation occurs on either path — so they were correctly deferred rather than blocking the merge.

### 1. `--dry-run` over a `--no-graph` genesis bails instead of projecting
The graph-path `--dry-run` block (lib.rs ~3273) only handles `preflight.pending_sync`; for a genesis (no `pending_sync`) it falls through to the guard at ~3378, which `bail!`s. The **pending-sync** path (#692, case 1) instead handles `--dry-run` by printing a projection and returning `Ok`. So a user running `--dry-run` on the default graph path over an interrupted `--no-graph` genesis gets a hard error rather than a projection.

Defensible as-is (refusing to *project* a mode that would itself be refused is reasonable, and the message says "No remote mutation was attempted"), but it's an **asymmetry** with case 1 worth reconciling — either project-then-note-the-refusal, or document the intentional difference.

### 2. Recovery message omits `--fresh`
The guard exempts `--fresh` (`!cfg.fresh`), and the surrounding comment states rebuilding in-place with `--fresh` is safe for a genesis (nothing committed past generation 0). But the user-facing `bail!` message only names two recovery routes — "Re-run with `--no-graph`" and "rebuild with a new `--state-dir` and a new `--prefix`". It should also mention `--fresh` as a valid in-place recovery route, since the guard already treats it as safe.

Both are cosmetic/UX completeness items on an already-correct guard. Surfaced by the correctness and evidence skeptics respectively during #711 verification.

xerj-org:
## What

Closes #718 — the two non-blocking polish items surfaced by the 3-skeptic review of #711 (#585 case 2). Both live in the genesis cross-path guard (`lib.rs`).

## Item 2 — `--fresh` missing from the recovery message

The guard exempts `--fresh` (`... && !cfg.fresh`) because nothing is committed past generation 0, so discarding and rebuilding in the requested (graph) mode is safe — the comment right above the guard already says so. But the `bail!` message named only two recovery routes (`--no-graph`, or a fresh `--state-dir`/`--prefix`), hiding a working recovery from the operator. Now it also offers **"re-run with `--fresh` to discard generation 0 and rebuild on the graph path."**

## Item 1 — the graph-path `--dry-run`-over-genesis refusal

The refusal is intentionally **not** gated on `!cfg.dry_run` (unlike the pending-sync projection above it). A dry run previews the run that *would* execute — but this graph-path run would itself be refused, and there is no graph-path genesis projection to show (the `--no-graph` dry-run projection further down is `!genesis_recovery`-gated). Bailing is therefore the honest preview, and it precedes every write. This is now documented in a comment and **locked by a test** (a `--dry-run` graph run over the pending genesis is refused with the same message and mutates nothing).

## Fail-before

The extended `no_graph_genesis_bootstrap_is_refused_on_the_graph_path` now asserts the message `contains("--fresh")` (both for the real run and the dry run). Reverting only the message's `--fresh` clause makes those assertions fail (`got: … Re-run with --no-graph … or rebuild with a new --state-dir …`).

- full `cargo test -p xerj-autoindex` (dev) — **715 passed / 0 failed** (verified green across 3 consecutive runs)
- ci-test: `cargo build --profile ci-test` — 0; genesis test — 1 passed / 0 failed

> Note: an earlier full-suite run showed a transient parallel flake in two unrelated tests (`two_byte_identical_junk_files_…`, `sync_executor::…gc_tombstone…`); both pass single-threaded and across 3 subsequent parallel runs, and clean `main` shows the same green — a pre-existing shared-state flake in the http_tests, not from this change (the guard bails at preflight before any failpoint/sync).

Closes #718.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -2191,15 +2191,39 @@ fn no_graph_genesis_bootstrap_is_refused_on_the_graph_path() {
     let mut graph = no_graph.clone();
     graph.no_graph = false;
     graph.snapshot_max_bytes = 64 << 30;
-    let error = run_index(graph).unwrap_err();
+    let error = run_index(graph.clone()).unwrap_err();
     let rendered = format!("{error:#}");
     assert!(
         rendered.contains("genesis bootstrap") && rendered.contains("different graph authority"),
         "#585 case 2: a --no-graph genesis bootstrap must be refused on the graph path with the \
          clear message, got: {rendered}"
     );
+    // #718: the message must name every valid recovery route, including
+    // `--fresh` — the guard exempts it (nothing is committed past genesis), so
+    // omitting it hides a working recovery from the operator.
+    assert!(
+        rendered.contains("--fresh"),
+        "#718: the recovery message must offer --fresh, which the guard exempts, got: {rendered}"
+    );
     // No mutation: nothing indexed, no new durable transaction beyond the
     // original sync_bootstrap.
+    assert_eq!(journal_events(state_dir.path(), "sync_bootstrap"), 1);
+    assert_eq!(journal_events(state_dir.path(), "sync_begin"), 0);
+    assert_eq!(journal_events(state_dir.path(), "sync_commit"), 0);
+    assert_eq!(endpoint.data_docs().len(), 0);
+
+    // #718: `--dry-run` does not exempt the refusal. A graph-path dry run over
+    // the pending --no-graph genesis is refused the same way (there is no graph
+    // genesis projection to show), with the same recovery message, and mutates
+    // nothing — a dry run of a refused run is a refusal.
+    let mut graph_dry = graph;
+    graph_dry.dry_run = true;
+    let dry_error = run_index(graph_dry).unwrap_err();
+    let dry_rendered = format!("{dry_error:#}");
+    assert!(
+        dry_rendered.contains("genesis bootstrap") && dry_rendered.contains("--fresh"),
+        "#718: --dry-run over a genesis on the graph path is refused too, got: {dry_rendered}"
+    );
     assert_eq!(journal_events(state_dir.path(), "sync_bootstrap"), 1);
     assert_eq!(journal_events(state_dir.path(), "sync_begin"), 0);
     assert_eq!(journal_events(state_dir.path(), "sync_commit"), 0);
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `cd212c69`)

### 1. Message-truth / dry-run behaviour / fail-before → **MERGE_OK**
- **The new `--fresh` recovery route is genuinely reachable** (code trace): under `--fresh` over a pending `--no-graph` genesis, `blocking_generation` is `None` (genesis_recovery exemption), the guard is skipped (`!cfg.fresh`), the committed-manifest graph refusal is skipped (`!genesis_recovery`), and `open_after_preflight(cfg.fresh)` discards generation 0 before the run rebuilds on the graph path — **the message is truthful, not a lie.**
- The dry-run guard is not gated on `!cfg.dry_run`; a graph-path `--dry-run` over the genesis bails before any write (the early dry_run block only handles `pending_sync` and falls through for a genesis), mutating nothing; the item-1 comment matches the code.
- **Fail-before proven:** reverting only the `--fresh` clause fails the test at line 2204; restore passes.
- *nit (coverage-only):* the test asserts the message *names* `--fresh` but doesn't execute a `--fresh` recovery to prove the rebuild — the route's correctness is established by code trace, not the test.

- `FMT=0`, `CLIPPY=0` (--all-targets -D warnings), `BUILD=0` (ci-test). **Full ci-test suite green on the first run (701/0 lib + all integration)**, including both target tests AND both known-flake tests (`two_byte_identical_junk_files_…`, `sync_executor::…gc_tombstone…`) — the known flake did not fire, no re-run needed.
- Logic verified: the `--fresh` route works; the graph-path dry-run falls through to the guard and bails before any write; the `--no-graph` projection is correctly `!genesis_recovery`-gated.

### 3. Evidence → **MERGE_OK**
- Item 2: the bail message now names `--fresh`; the route is genuinely reachable (not misleading). Item 1: the comment documents the intentional non-`!dry_run` gating, and the extended test adds a `--dry-run` graph sub-case that reaches the same guard and bails with zero mutation (journal_events + data_docs asserted). Fail-before genuine. Diff scoped to exactly the two items; all original #711 assertions still hold.
- *nit (harmless):* the dry-run sub-case asserts `genesis bootstrap` + `--fresh` but omits `different graph authority` (same bail string, slightly asymmetric coverage).

---

**Decision:** 0 BLOCKING. Merging once the last CI checks go green + `mergeStateStatus=CLEAN` (head at the verified SHA `cd212c69`). The two nits are coverage-only/harmless — no separate follow-up warranted.

PR #724, opened 2026-08-24
fix(autoindex): #722 - the --no-graph generated pipeline never consumed as_document

### LINKED ISSUE #722: autoindex --no-graph: demoted one-off config files (#173) never publish through extract_as_document, only route to the docs dataset
## Problem

The `--no-graph` generated pipeline's own extraction/publish step (`sync_executor.rs`) never checks `FileAssignment.as_document`, so a demoted one-off config file (#173: a small, group-less, single-record `Json`/`Yaml`/`Xml` file folded into the scope's `docs` dataset) is still indexed through its raw family extractor (`extract::extract`) instead of the document renderer (`extract::extract_as_document`) — even though the frozen dataset's mapping was built from `extract_as_document`'s output (`title`/`body`).

Verified live (rc.18 debug build), FRESH run, no incremental reconcile involved:

```
$ ls corpus/
f1.txt        # plain prose
config.json   # {"host": "example.com", "port": 8080}

$ xerj autoindex corpus --no-graph ...
generation 1 committed — 1 datasets, 2 records live
```

`config.json` is correctly routed into the `docs` dataset (`ax_dataset: "docs"`, `ax_locator: "doc"`) — but its published `_source` is:

```json
{"host": "example.com", "port": 8080, "ax_path": "config.json", "ax_locator": "doc", "ax_dataset": "docs", ...}
```

instead of the document shape (`title`/`body`) the `docs` dataset's own mapping was inferred from. `host`/`port` land as dynamically-mapped fields alongside the declared `title`/`body` mapping.

## Root cause

Confirmed by reading both pipelines:

- The **legacy (graph-enabled) path**'s own extraction dispatch (`lib.rs`, around the phase-B publish loop) explicitly branches on `fa.as_document`:
  ```rust
  let res = if fa.as_document {
      extract::extract_as_document(&f.path, sn.gzip, &mut sink)
  } else if sn.family == Family::Pdf {
      ...
  } else {
      extract::extract(&f.path, sn, ..., &mut sink)
  };
  ```
- The **generated (`--no-graph`) path**'s extraction (`sync_executor.rs`) has exactly one extraction call site and it is unconditional:
  ```rust
  let stats = crate::extract::extract(snapshot_blob, &sniffed, None, &mut sink)
  ```
  `as_document` (and `extract_as_document`) is never referenced anywhere in `sync_executor.rs` outside test fixtures.

So `#173` (one-off config demotion) only actually *publishes* document-shaped records on the legacy graph path. On `--no-graph` the file is correctly *routed* to the `docs` dataset/index (both the fresh-plan `dataset::cluster` demotion and, since #712, the incremental `reconcile_plan` mirror agree it belongs there) — but the record content published is whatever its raw family extractor produces, which does not match the dataset's own declared mapping.

## Impact

- Silent field-shape drift: a `docs` dataset's mapping declares `title`/`body`; a demoted file instead contributes dynamically-mapped fields from its raw content, which do not populate `title`/`body` at all and are invisible to `title`/`body` queries.
- Affects both a file demoted on its very first (fresh) `--no-graph` run and one demoted incrementally by `reconcile_plan` (#712, once merged) — the routing decision is correct in both, only the actual field content is wrong, and only under `--no-graph`.

## Suggested fix

Thread `as_document` into `sync_executor.rs`'s extraction call site the same way the legacy path does: when the file's plan assignment carries `as_document: true`, call `extract::extract_as_document` instead of `extract::extract`. Needs its own investigation into how `sync_executor.rs`'s snapshot/blob-based extraction (`snapshot_blob`) interacts with `extract_as_document`'s path-based signature (it takes `&Path`, not a blob) — the generated pipeline extracts from a sealed snapshot, not the live file, so this may need a blob-accepting variant rather than a straight reuse.

Found while verifying #712 (fixing #346) — that PR's own reconcile-side fix is unaffected: it makes the incremental *routing* decision correct (no more spurious "unsupported dataset/schema evolution" abort) and matches the also-currently-broken *fresh* behavior above, so it introduces no regression, but does not by itself make `--no-graph` publish the document-shaped content the `docs` mapping expects.

Vinz2168:
Closes #722.

## Problem

A demoted one-off config file (#173: a small, group-less, single-record `Json`/`Yaml`/`Xml` file folded into a scope's `docs` dataset) is correctly **routed** into the `docs` dataset under `--no-graph` — `reconcile_plan` and `dataset::cluster` agree on that — but its **published content** still came from its raw family extractor, not `extract_as_document`, silently disagreeing with the `docs` mapping the run itself built.

Verified live before fixing, on a completely **fresh** (non-incremental) `--no-graph` run — one prose file + one one-off `config.json`:

```
generation 1 committed — 1 datasets, 2 records live
```

`config.json` lands in the `docs` index (`ax_dataset: "docs"`, `ax_locator: "doc"`) but its `_source` is:

```json
{"host": "example.com", "port": 8080, "ax_path": "config.json", ...}
```

instead of the `title`/`body` document shape the `docs` dataset's own mapping was inferred from.

## Root cause

The **legacy (graph-enabled) path**'s own phase-B extraction dispatch explicitly branches on `FileAssignment.as_document`:

```rust
let res = if fa.as_document {
    extract::extract_as_document(&f.path, sn.gzip, &mut sink)
} else if sn.family == Family::Pdf {
    ...
} else {
    extract::extract(&f.path, sn, ..., &mut sink)
};
```

`sync_executor.rs` — the **generated (`--no-graph`) pipeline**'s own extraction step — has exactly one extraction call site and it is unconditional: `crate::extract::extract(snapshot_blob, &sniffed, None, &mut sink)`. `as_document` is never referenced anywhere in `sync_executor.rs` outside test fixtures.

## Fix

1. `sync_executor.rs`'s `prepare_artifact` already has the file's `FileAssignment` in scope (`plan.files.get(content_id)`) — branch on `assignment.as_document` exactly the way `lib.rs`'s phase-B dispatch does. `snapshot_blob` is a real filesystem path (the sealed snapshot), not a byte blob, so **no signature refactor of the extractors was needed** — just the same branch, on the same field, on a real path.

2. That surfaced a second, narrower bug: naively calling `extract_as_document(snapshot_blob, ...)` derives the document's title from `snapshot_blob`'s own filename — the sealed snapshot's **ordinal** name (`prepared/00000000`), not the source file's — so every demoted document titled itself `"00000000"` instead of the real file's stem. `extract/mod.rs`'s `extract_as_document` is now a thin wrapper over a new `extract_as_document_with_name(content_path, logical_path, gzip, sink)`, mirroring `sniff::sniff_with_name`'s existing split for exactly the same reason (content from the snapshot blob, identity from the source path). `sync_executor.rs` calls the `_with_name` variant with `source.path` as the logical name.

## Testing

- New test `preparation_of_a_demoted_file_publishes_a_real_document_with_the_source_title` (`sync_executor.rs`) — a real `create_prepared_snapshot()` run over an `as_document: true` assignment for a real `config.json` on disk, reads the sealed NDJSON back, asserts `title == "config"` (not `"00000000"`), `body` contains the decoded JSON text, and the raw `host`/`port` fields are absent from the published record. Verified fail-before: with the `as_document` branch neutralized, the assertion fails exactly as the live repro did (no `title` key at all, raw `host`/`port` fields present).
- Verified live end-to-end after the fix (real server + CLI, `--no-graph`): `config.json` now publishes `{"title": "config", "body": "{\"host\": \"example.com\", \"port\": 8080}", ...}` under the `docs` index.
- `cargo test -p xerj-autoindex --profile ci-test -- --test-threads=2`: 680 passed; the 19 failures are pre-existing macOS environment issues (non-UTF-8 filenames + one unrelated data-assertion mismatch) that reproduce identically on unmodified `main`.

Found and filed (#722) while live-verifying #712 (fixing #346) — confirmed pre-existing and unrelated to that PR: reproduces identically on a completely fresh `--no-graph` run with no incremental reconcile involved at all.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/mod.rs
+++ b/engine/crates/xerj-autoindex/src/extract/mod.rs
@@ -125,8 +125,24 @@ pub fn read_whole(path: &Path, gzip: bool, cap: u64) -> Result<Option<Vec<u8>>> 
 /// the text itself. Emits the same `title`/`body`/`section` vocabulary as
 /// every other document extractor (`FieldOrigin::Extractor`).
 pub fn extract_as_document(path: &Path, gzip: bool, sink: Sink) -> Result<ExtractStats> {
+    extract_as_document_with_name(path, path, gzip, sink)
+}
+
+/// [`extract_as_document`], but the title comes from `logical_path` rather
+/// than `content_path`. The `--no-graph` generated pipeline reads a
+/// SEALED SNAPSHOT — a real file on disk, but under its own ordinal name
+/// (`prepared/00000000`, …), not the source's — so deriving the title from
+/// `content_path` there produced `"00000000"` instead of the file's own
+/// name (#722). Mirrors `sniff::sniff_with_name`'s split for exactly the
+/// same reason.
+pub fn extract_as_document_with_name(
+    content_path: &Path,
+    logical_path: &Path,
+    gzip: bool,
+    sink: Sink,
+) -> Result<ExtractStats> {
     let mut stats = ExtractStats::default();
-    let Some(bytes) = read_whole(path, gzip, MAX_WHOLE_FILE)? else {
+    let Some(bytes) = read_whole(content_path, gzip, MAX_WHOLE_FILE)? else {
         stats.junk += 1;
         return Ok(stats);
     };
@@ -135,7 +151,7 @@ pub fn extract_as_document(path: &Path, gzip: bool, sink: Sink) -> Result<Extrac
         stats.junk += 1;
         return Ok(stats);
     }
-    let title = path
+    let title = logical_path
         .file_stem()
         .map(|s| s.to_string_lossy().to_string())
         .unwrap_or_else(|| "untitled".into());
--- a/engine/crates/xerj-autoindex/src/sync_executor.rs
+++ b/engine/crates/xerj-autoindex/src/sync_executor.rs
@@ -1602,8 +1602,36 @@ fn prepare_artifact(
     // generation because one line of one log file did not parse would make a
     // realistic corpus unindexable. The count is sealed into the artifact so
     // the generation, and then the catalog, can report it.
-    let stats = crate::extract::extract(snapshot_blob, &sniffed, None, &mut sink)
-        .with_context(|| format!("extract {} into durable preparation", source.rel))?;
+    //
+    // #722: `assignment.as_document` (a demoted one-off config file, #173)
+    // decides the record *shape*, the same precedence the legacy path's own
+    // phase-B dispatch gives it (`lib.rs`, `if fa.as_document { … }`) — the
+    // frozen dataset's mapping was built by re-sampling through
+    // `extract_as_document`, so that is the only extractor whose output may
+    // be compared against it. Before this, the durable/generated pipeline
+    // never checked the flag at all: a demoted file was correctly ROUTED
+    // into the docs dataset (`reconcile_plan`/`dataset::cluster` agree it
+    // belongs there) but its published fields still came from its raw
+    // family extractor, silently disagreeing with the mapping. `sniffed` is
+    // already sniffed from `snapshot_blob`, a real path on the sealed
+    // snapshot (not a byte blob), so the branch is exactly the legacy
+    // path's — no format change needed.
+    let stats = if assignment.as_document {
+        // `extract_as_document_with_name`, not `extract_as_document`:
+        // `snapshot_blob` is a real path, but to the SEALED SNAPSHOT under
+        // its own ordinal name, not the source file's — the title must come
+        // from `source.path` (the same split `sniff_with_name` above makes
+        // for the same reason).
+        crate::extract::extract_as_document_with_name(
+            snapshot_blob,
+            &source.path,
+            sniffed.gzip,
+            &mut sink,
+        )
+    } else {
+        crate::extract::extract(snapshot_blob, &sniffed, None, &mut sink)
+    }
+    .with_context(|| format!("extract {} into durable preparation", source.rel))?;
     if let Some(error) = sink_error {
         return Err(error);
     }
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `56398740`)

All three returned **MERGE_OK** with **zero blocking findings and zero nits**.

**Correctness / dispatch parity / delegation / fail-before → MERGE_OK.** `prepare_artifact` (sync_executor.rs:1619-1633) branches `if assignment.as_document { extract_as_document_with_name(…) } else { extract(snapshot_blob, &sniffed, None, …) }`, mirroring the legacy phase-B dispatch (lib.rs:4989-5026 `if fa.as_document { extract_as_document } else { extract }`). **Delegation safe:** `extract_as_document(path, gzip, sink)` → `_with_name(path, path, …)`, so `content==logical==path` and all three existing callers (reconcile_plan `document_sample`, lib.rs:1937/4994) are byte-identical. **Title split correct:** generated call passes `content=snapshot_blob` (sealed ordinal blob), `logical=&source.path` (real name), matching the `sniff_with_name` split; title from the real name, body from the blob — not swapped. Fail-before: revert the dispatch → raw JSON `{host,port}` + null title.

**Regression & build-hygiene → MERGE_OK.** `FMT=0`, `CLIPPY=0` (`--workspace --all-targets`), `BUILD=0`; full ci-test suite 702 lib + 14 integration, 0 failed. The non-demoted (`as_document=false`) branch is unchanged; non-document extractors untouched. Fail-before reproduced deterministically.

**Evidence & two-part claim → MERGE_OK.** The test writes a real `config.json`, sets `as_document=true`, runs `create_prepared_snapshot`, reads the sealed NDJSON, and asserts `title=="config"` (source stem, not the ordinal `"00000000"`) AND `body` contains the decoded text AND raw `host`/`port` keys are absent — and it **fails on BOTH halves** without the fix (branch neutralized → raw JSON + null title; naive `extract_as_document` → title `"00000000"`, proving the ordinal-name root cause that justifies the `_with_name` split). All PR claims true against the tree. Genuinely complements #730 (routing) — #724 is the publish side.

---

**Decision:** 0 BLOCKING. Merging once all checks green + CLEAN (head at verified SHA `56398740`).

PR #727, opened 2026-08-24
fix(autoindex): #346 — demote an incrementally-added one-off config into the docs dataset

### LINKED ISSUE #346: Resume journal reports skipped files below the noise floor (split out of #326)

xerj-org:
## What

Closes #346 — completes the fix the (now-closed) #712 left half-done. A **fresh** plan demotes a group-less, single-record, `demotable_family` (JSON/YAML/XML) one-off config into the scope's **docs** dataset and indexes it **as a document** (`dataset::cluster` demotion loop + `build_phase_a`'s `extract_as_document`, which sets `as_document` — lib.rs:1919-1962). The **incremental** reconcile path did not: a newly-added config was classified by its raw `"json"` family, matched zero frozen datasets, and **aborted the whole run** (`no frozen dataset accepts family json … use a new prefix`). Adding one `.json`/`.yaml`/`.xml` to an already-indexed docs folder therefore wedged the run.

## Fix (`reconcile_plan.rs`)

When a new file is group-less, single-record, `demotable_family`, **and** a group-less `docs` dataset exists to join, **and** no group-less dataset of the file's own family exists — replicate the fresh demotion:
1. set `as_document` and re-sample through `document_sample` (the `{title, body}` shape `emit_document` renders), and
2. classify it as `"docs"` so it joins the frozen docs dataset.

The third guard is load-bearing: the **>8-configs** case forms its own `json` *data* dataset instead of demoting, and such a config must join **that** dataset — so the demotion only fires when the raw family has no group-less dataset. `demotable_family` is made `pub(crate)`.

Edge cases (all correct): no docs dataset → aborts (genuine schema evolution); group-less json data dataset exists → joins it, not docs; multi-record (JSON array) / grouped / non-demotable family → raw family unchanged.

## Fail-before

New test `new_demotable_config_file_joins_the_docs_dataset_as_a_document` writes a real `config.json`, asserts it joins `"docs"` with `as_document=true`. Disabling only the demotion guard (`new_file_demoted = false && …`) reverts to the exact #346 abort (`no frozen dataset accepts family json …`).

- full `cargo test -p xerj-autoindex` — **717 passed / 0 failed on BOTH dev and ci-test profiles**

Closes #346.

### DIFF
--- a/engine/crates/xerj-autoindex/src/dataset.rs
+++ b/engine/crates/xerj-autoindex/src/dataset.rs
@@ -87,7 +87,7 @@ fn jaccard(a: &HashSet<&str>, b: &HashSet<&str>) -> f64 {
 /// single-record files? Self-describing structured formats only: CSV is
 /// tabular by nature, logs/jsonl are collections by nature, and sql groups
 /// carry a real table schema.
-fn demotable_family(f: Family) -> bool {
+pub(crate) fn demotable_family(f: Family) -> bool {
     matches!(f, Family::Json | Family::Yaml | Family::Xml)
 }
 
--- a/engine/crates/xerj-autoindex/src/reconcile_plan.rs
+++ b/engine/crates/xerj-autoindex/src/reconcile_plan.rs
@@ -152,7 +152,32 @@ pub(crate) fn reconcile_plan(
         // renderer `build_phase_a` uses, exactly as the fresh path does, and
         // carry the committed `as_document` decision forward. Losing it would
         // index flattened records into a document mapping.
-        let as_document = retained.is_some_and(|owner| owner.as_document);
+        // #346: a fresh plan demotes a group-less, single-record,
+        // `demotable_family` (Json/Yaml/Xml) file into the scope's docs dataset
+        // and indexes it AS a document (the `dataset::cluster` demotion loop +
+        // `build_phase_a`'s `extract_as_document`, lib.rs:1919-1962). An
+        // incrementally-added file of that same shape must reconcile the same
+        // way — otherwise it is compared by its raw family, matches no frozen
+        // dataset, and aborts the whole run ("use a new prefix"). Guarded so it
+        // fires ONLY when the demotion is what the frozen plan actually did: a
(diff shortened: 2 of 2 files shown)


xerj-org:
Closing this — I got it wrong on two counts, caught by the review:

**1. Mis-attributed.** This PR (and the earlier #712) says it fixes #346, but **#346 is a different bug entirely** — *"resume journal reports skipped files below the noise floor"* (a re-run prints `N file(s) appeared after the resume plan was frozen and were NOT indexed` and exits 0 — a silent-incompleteness *reporting* defect). The bug this PR actually fixes — an incremental reconcile aborting when a one-off JSON/YAML/XML config is added to an already-indexed docs folder — is real and un-tracked, but it is **not** #346, so "Closes #346" is wrong.

**2. Incomplete.** The 3-skeptic review proved the fix only lands when the frozen docs dataset's specs are exactly `{title, body}`. A realistic docs folder of multi-section markdown yields a docs dataset carrying `{title, headings, section, body}`; the config re-samples to `{title, body}` and scores Jaccard 2/4 = 0.5 < the 0.7 `classify_new` threshold, so it **still aborts** with the exact same error. A complete fix must relax `classify_new` so a document's fields (always a subset of the docs union) join a docs dataset, and also handle the empty-`key_fields` document-file case — larger than this PR.

The fmt failure the regression skeptic flagged is fixed on the pushed head, but the two issues above are the real reasons not to merge. Reproduction + the correct scoping are preserved on the branch for a proper, correctly-targeted fix later.

PR #728, opened 2026-08-24
fix(autoindex): #346 — surface files-skipped-after-plan-freeze distinctly, not below the noise floor

### LINKED ISSUE #346: Resume journal reports skipped files below the noise floor (split out of #326)

xerj-org:
## What

Closes **#346** (the *actual* one — "resume journal reports skipped files below the noise floor", split out of #326). Note: the earlier #712/#727 mis-referenced #346; this PR fixes the real defect.

A re-run over a changed tree records files that appeared *after* the resume plan was frozen as `skipped` junk and prints one mid-stream progress note (`autoindex: N file(s) appeared after the resume plan was frozen and were NOT indexed`). But that count was then folded into the generic `files_junk`/`junk_files` total — **indistinguishable from ordinary unparseable junk**. For a "make the index match the tree" workflow this reads as a routine `completed-with-junk` success while the index is in fact **incomplete**.

## Fix

Surface the count **distinctly**:
- capture `skipped_appeared` before `new_unplanned` is drained into `plan.junk_files`;
- only when `> 0`, emit it in the run summary document (`run_doc["skipped_appeared"]`) and on the terminal `xerj-done` line (`done_fields`).

A healthy run (no files appeared after freeze) is byte-unchanged — the key/field is absent. Exit code is untouched (already `3` / `completed-with-junk` via the junk count); per the issue, *surfacing the count* is the asked-for fix (vs. changing exit semantics, which would risk callers).

## Fail-before

The existing `a_rerun_after_an_added_file_succeeds_and_fresh_absorbs_it` (which indexes `first.csv`, adds `second.csv`, re-runs) now also asserts `report["skipped_appeared"] == 1`. Reverting only the `run_doc["skipped_appeared"]` hunk leaves the key absent (`Null`) and the assertion fails (`left: Null, right: 1`), while `files_junk` alone stays `1` — exactly the buried-below-noise-floor state.

## Verification (rustc 1.97.1, exact CI commands — no pipe masking)

- `cargo fmt --all --check` — 0
- `cargo clippy --workspace --all-targets -- -D warnings` — 0
- full `cargo test -p xerj-autoindex` — **716 passed / 0 failed on BOTH dev and ci-test profiles**

Closes #346.

### DIFF
--- a/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
@@ -823,6 +823,11 @@ fn a_rerun_after_an_added_file_succeeds_and_fresh_absorbs_it() {
     assert_eq!(code, 3);
     let report = report.unwrap();
     assert_eq!(report["files_junk"], 1, "{report}");
+    // #346: the file that appeared after the plan was frozen is surfaced
+    // DISTINCTLY as `skipped_appeared`, not just folded into `files_junk`, so a
+    // "make the index match the tree" re-run can see the index is incomplete
+    // instead of reading as an ordinary completed-with-junk success.
+    assert_eq!(report["skipped_appeared"], 1, "{report}");
     // `records_total` is the live server-side count (#195), so it still reports
     // the record the first run published. What must be zero is the run-local
     // counter: this rerun indexed nothing, because the frozen plan cannot
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -4473,6 +4473,15 @@ pub fn run_index_report(cfg: IndexCfg) -> Result<(i32, Option<Value>)> {
              stay idempotent)",
         );
     }
+    // #346: the count of files that appeared after the resume plan was frozen
+    // (and so were skipped, never indexed) is surfaced DISTINCTLY below — in the
+    // run summary and the terminal `xerj-done` line — not just as a mid-stream
+    // note folded into the generic `junk_files` total. Folding it in leaves a
+    // "make the index match the tree" re-run reading as an ordinary
+    // completed-with-junk success while the index is in fact incomplete, which
+    // is the below-the-noise-floor reporting #346 is about. Captured here before
+    // `new_unplanned` is drained into `plan.junk_files`.
+    let skipped_appeared = new_unplanned.len();
     if resumed_with_plan {
         // Legacy journals predate intent-before-publication and may have live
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent 3-skeptic verification (rustc 1.97.1, isolated worktrees at base `319b7f06`)

All three returned **MERGE_OK** with **zero blocking findings and zero nits**.

**Correctness / count / healthy-run invariance / fail-before → MERGE_OK.** Capture ordering correct: `let skipped_appeared = new_unplanned.len();` (lib.rs:4484) is BEFORE the drain `plan.junk_files.append(&mut new_unplanned)` (5987). Count == `new_unplanned.len()`, the same value as the mid-stream note; no off-by-one/double-count (new_unplanned only iterated between). Healthy-run byte-invariance: both emissions gated by `if skipped_appeared > 0` (run_doc 5949, done_fields 6138), so a run with no post-freeze files leaves the summary doc AND the terminal line unchanged. Fail-before: reverting the run_doc hunk → `report["skipped_appeared"]` absent → `Null != 1` fails. fmt=0 (no pipe), lib suite 702/0.

**Regression & build-hygiene → MERGE_OK.** Exact unpiped commands: `FMT=0`, `CLIPPY=0` (`--workspace --all-targets`), `BUILD=0` (ci-test). Full ci-test suite all green (702 + 8/1/1/4 integration, 0 failed); the known flakes did not fire. Every other run-summary / `files_junk` / `xerj-done` assertion across 702 http-test cases passes unchanged. Fail-before reproduced.

**Evidence → MERGE_OK.** The test genuinely reproduces #346 (first.csv indexed, second.csv created *after* the plan is frozen → new_unplanned → code 3, files_junk 1, records_submitted_this_run 0, skipped_appeared 1); `report` is the run summary doc. Count surfaced in BOTH requested places (summary + `xerj-done`), same value, purely additive (exit code untouched). Confirmed this is the **correct** #346 (skipped-files reporting), not the config-reconcile bug #712/#727 mis-referenced.

**Decision:** 0 BLOCKING. Merging once the last CI checks go green + CLEAN (head at verified SHA `319b7f06`).

PR #730, opened 2026-08-24
fix(autoindex): #729 — route a new document/config to the docs dataset on --no-graph incremental

### LINKED ISSUE #729: autoindex: --no-graph incremental re-index aborts when a new document or one-off config file is added
On the **`--no-graph` incremental** re-index path (`project_reconcile_plan` → `reconcile_plan`, lib.rs:3587), the current inventory is projected onto the frozen plan and **every** file — including new ones — is classified by `classify_new`, which filters frozen datasets by **exact family match**. A fresh plan folds group-less documents (empty `key_fields`: prose/code/PDF) and demoted one-off configs (single-record JSON/YAML/XML) into the scope's `docs` dataset (family `"docs"`), but their sniffed family is `txt-prose`/`code`/`json`/… — so a **newly-added** document or config matches no frozen dataset and the **whole run aborts**:

```
no frozen dataset accepts family txt-prose, group None, and fields ["body", "title"];
this requires unsupported dataset/schema evolution (use a new prefix)
```

Adding any new document/config file to an already-`--no-graph`-indexed folder therefore wedges the incremental re-run (the default graph path instead *skips* the file — see #346). Reproduced by unit tests over `reconcile_plan` for both the prose-document and the JSON-config shapes.

(Historical note: PRs #712 and #727 mis-referenced this as #346 — which is a different bug, resume-journal skipped-file reporting, now fixed. This is the correctly-scoped issue.)

**Fix:** in `reconcile_plan`'s new-file path, route a group-less document (empty key_fields) OR a demotable single-record config to family `"docs"` when a group-less `docs` dataset exists and no group-less same-family dataset does; set `as_document` only for the config (documents index via their own extractor); and relax `classify_new` for a `"docs"`-family target to subset-compatibility (a document's rendered fields are a subset of the docs union).

xerj-org:
## What

Closes **#729** — completes the fix #712/#727 tried (both mis-referenced #346, now closed) and, crucially, is the FIRST to cover the **document-file** case, not just configs.

On the **`--no-graph` incremental** path (`project_reconcile_plan` → `reconcile_plan`, lib.rs:3587), the entire current inventory — new files included — is classified by `classify_new`, which filters frozen datasets by **exact family match**. A fresh plan folds group-less **documents** (empty `key_fields`: prose/code/PDF) AND demoted single-record **configs** (JSON/YAML/XML) into the scope's `docs` dataset (family `"docs"`), but their sniffed family is `txt-prose`/`code`/`json`/… — so a newly-added file of either shape matches nothing and the **whole run aborts** (`use a new prefix`).

## Fix (`reconcile_plan.rs`)

1. **Route** a group-less document (empty key_fields) OR a demotable single-record config to family `"docs"` — gated on a group-less `docs` dataset existing and no group-less same-family dataset (a >8-config fleet forms its own data dataset and must join THAT).
2. **`as_document` only for the config** (re-samples via `document_sample`); a prose/code document indexes through its own extractor, so `as_document` stays false — exactly as `build_phase_a` leaves non-demoted docs members (1946-1962).
3. **Relax `classify_new`** for a `"docs"`-family target to **subset-compatibility** (`intersection == fields.len()`): a document's rendered field names are a subset of the docs union, which the 0.7-Jaccard threshold rejected against a richer multi-section docs dataset (`{title,body,section}`) — the exact limitation that aborted #727.

`demotable_family` made `pub(crate)`.

## Fail-before

Two new tests: `new_document_file_joins_the_docs_dataset_on_incremental` (prose) and `new_demotable_config_joins_a_richer_docs_dataset_as_a_document` (JSON into a `{title,body,section}` docs dataset). Neutralising only `route_to_docs` makes **both** abort with the exact `no frozen dataset accepts family txt-prose/json … use a new prefix`.

## Verification (rustc 1.97.1, exact commands — no pipe masking)

- `cargo fmt --all --check` — 0
- `cargo clippy --workspace --all-targets -- -D warnings` — 0
- full `cargo test -p xerj-autoindex` — **718/0 (ci-test)**, 703/0 dev lib

Closes #729.

### DIFF
--- a/engine/crates/xerj-autoindex/src/dataset.rs
+++ b/engine/crates/xerj-autoindex/src/dataset.rs
@@ -87,7 +87,7 @@ fn jaccard(a: &HashSet<&str>, b: &HashSet<&str>) -> f64 {
 /// single-record files? Self-describing structured formats only: CSV is
 /// tabular by nature, logs/jsonl are collections by nature, and sql groups
 /// carry a real table schema.
-fn demotable_family(f: Family) -> bool {
+pub(crate) fn demotable_family(f: Family) -> bool {
     matches!(f, Family::Json | Family::Yaml | Family::Xml)
 }
 
--- a/engine/crates/xerj-autoindex/src/reconcile_plan.rs
+++ b/engine/crates/xerj-autoindex/src/reconcile_plan.rs
@@ -330,13 +330,19 @@ fn classify_new(
         let (intersection, union) = field_overlap(fields, &dataset.specs);
         let threshold_met = if dataset.family == "docs" {
             // A `docs` dataset's specs are the UNION of every document/config
-            // the fold produced. A group-less document whose rendered field
-            // NAMES are all present in that union belongs to it, however many
-            // other document fields the union carries — so accept on subset
-            // (`intersection == |fields|`), not the 0.7 Jaccard the larger union
-            // would otherwise fail. `field_overlap` counts names, and
-            // `fields.len()` is the file's rendered-field count.
-            intersection == fields.len()
+            // the fold produced. `ensure_compatible` (run just above) already
+            // requires every observed SCALAR field to be present in the frozen
+            // specs — a field absent from the specs is tolerated ONLY when
+            // `acc.n == 0` (pruned: e.g. a code file's object-valued `symbols`
+            // sidecar, #580). So a document/config that reached here belongs to
+            // the docs dataset regardless of how many other fields the union
+            // carries; accept it. (The 0.7 Jaccard would wrongly reject a
+            // document with fewer fields than the union, and `== fields.len()`
+            // wrongly counts the pruned `symbols` key and aborted a new code
+            // file — the #729 central case.) `let _` keeps the shared
+            // `field_overlap` result in scope for the data-family branches.
+            let _ = (intersection, union);
+            true
         } else if group.is_some() {
             intersection * 2 >= union // 0.5
         } else {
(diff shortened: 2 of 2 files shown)


xerj-org:
## Independent verification (rustc 1.97.1, isolated worktrees) — a blocking finding was caught and fixed

### Round 1 (base `00d61dc6`): 2 MERGE_OK + **1 BLOCK**
- **Correctness (routing/subset)** → MERGE_OK; **Regression** → MERGE_OK (both traced routing exactly, fail-before genuine, full suite green).
- **Evidence & fresh-parity → BLOCK (real, proven):** the docs-family acceptance `intersection == fields.len()` re-counted a **pruned** field. A code file emits a `symbols` sidecar (array-of-objects, #580) recorded with `acc.n==0`, pruned from every frozen docs `FieldSpec`. `ensure_compatible` tolerates it (a missing spec is allowed only when `acc.n==0`), but `fields.len()` counted it → `intersection(5) != fields.len()(6)` → **a new `.rs`/`.py` with functions still aborted** the `--no-graph` re-run (the central #729 case for code repos), while the fresh path folds it into docs with no abort. Proven with a throwaway test.

### Fix (`765e9b25`)
The docs-family acceptance is now `true`: `ensure_compatible` (run just above) already requires every observed **scalar** (`acc.n>0`) field to be present, so a document/config/code file that reached the threshold belongs to the docs union; a genuinely-new scalar field is still rejected upstream. Added `new_code_file_with_a_pruned_symbols_sidecar_joins_docs_not_aborts`.

### Round 2 re-verification (`765e9b25`): **all 3 MERGE_OK, ZERO BLOCKING**
- **Over-accept (the key risk) → MERGE_OK:** proven the `true` fix does NOT over-accept — (1) a new scalar field always has `acc.n>0` → `ensure_compatible` errors → `classify_new` skips the docs dataset → correct `use a new prefix` abort (never reaches `true`); (2) no data file can reach the docs branch (no `Family::as_str()` returns `"docs"`; `family=="docs"` is produced only by `route_to_docs`); (3) the prior code-file block now joins docs.
- **Routing-correctness → MERGE_OK:** the diff is surgical (only the docs threshold + one test); routing fires exactly as before; fail-before reproduces the exact #729 abort. Full lib 705/0.
- **Regression → MERGE_OK:** `FMT=0`, `CLIPPY=0` (`--workspace`), `BUILD=0`; 705 lib + all integration green; the 3 new tests pass.

### Non-blocking nits (filed as follow-ups on merge)
- `true` also tolerates a brand-new **object-valued** (`acc.n==0`) field absent from specs — by design (genesis folds object-field documents without a FieldSpec too; consistent with genesis admission, not a new mapping). Worth a one-line comment.
- With `true`, a scope with two group-less `docs` datasets could hit the pre-existing ambiguity tie-break (not demonstrable in practice — a scope normally has one).
- (Round 1) `route_to_docs` past the >8-config fleet cutoff; scope-agnostic `classify_new`; zero-field routed file accepted.

---

**Decision:** 0 BLOCKING (the one blocking finding was fixed and re-verified). Merging once the last CI checks go green + CLEAN (head at the verified SHA `765e9b25`). The nits are filed as follow-ups.

PR #732, opened 2026-08-24
fix(autoindex): exclusion sweep invalidates a removed file's graph edges (#694)

### LINKED ISSUE #694: autoindex: #589 follow-up — sweep an excluded file's graph edges via invalidate_prior_edges (src_file)
Follow-up from the 3-skeptic verification of #690 (#589). #589 deliberately deferred graph-edge cleanup.

Per-file edge provenance already exists: `detect::invalidate_prior_edges` (detect/mod.rs:664) soft-invalidates edges by `{term:{src_file: rel}}`. The excluded `InventoryDeltaEntry` carries that `rel` as `path`, so the sweep can reuse this hook to invalidate the excluded file's edges rather than leave them live (`edges_note` documents they currently stay live).

Impact: an edge naming a now-excluded file stays searchable in the edges index — lower-severity than document content (which #589 sweeps), but a real residual. Add to `sweep_excluded_groups` + a mock-ES edge test.

xerj-org:
Closes #694 — the #589 follow-up that deferred graph-edge cleanup.

## Problem
`sweep_excluded_groups` (the #589 widened-exclusion sweep) deleted an excluded file's content records (by `ax_file`) and its catalog doc (by `path`), but left the **graph edges it taught** live in `.xerj-memory-<brain>-edges`. An edge naming a now-excluded file stayed searchable in the edges index with no source file — a real residual (`edges_note` documented it as a known gap; the old `KNOWN GAP (#589, next hunk)` comment on the fn marked exactly this).

## Fix
- **`sweep_excluded_groups`** now soft-invalidates the excluded file's edges by reusing the replacement hook **`detect::invalidate_prior_edges`** — the excluded `InventoryDeltaEntry`'s `path` **is** the `src_file` the edges carry. The edge is stamped `invalid_at`/`expired_at` (not deleted); the bi-temporal record survives for `as_of` time travel, exactly like a replaced file's edges.
- Gated on **`graph_edges_index(cfg)`** = `Some`: skipped under `--no-graph` or a brain that fails validation (a run that could never have written edges), mirroring `edges_index_name(&brain)` at the graph phase without duplicating the `--brain`/`derive_brain_name` fallback.
- The sweep runs **ahead of** the graph phase's `ensure_index`, so **`invalidate_prior_edges`** now reads via **`search_present`** — a missing edges index means "no edges to invalidate", not an error. The replacement-invalidation caller ensures the index first, so its behaviour is unchanged (`Some`).

## Test / fail-before
`sweep_excluded_groups_invalidates_the_excluded_files_edges` over the mock HTTP backend: a live edge taught by the excluded file is stamped `invalid_at`, while a different file's edge stays live (no over-invalidation). **Fail-before** (invalidation reverted): the excluded edge keeps `invalid_at: None` — the assertion discriminates the fix. The shared mock `_search` now honours `must_not:[{exists}]` so the invalidation pass converges (additive; other tests unaffected).

- `cargo test -p xerj-autoindex` (full suite, ci-test) → **706 lib + integration, 0 failed**

Scope: `xerj-autoindex` only (`detect/mod.rs`, `lib.rs`, the http-test mock). No `:9200` conformance run touched.

### DIFF
--- a/engine/crates/xerj-autoindex/src/detect/mod.rs
+++ b/engine/crates/xerj-autoindex/src/detect/mod.rs
@@ -673,7 +673,15 @@ pub fn invalidate_prior_edges(es: &Es, edges_index: &str, rel: &str, now_ms: i64
     });
     let mut total = 0u64;
     for _ in 0..MAX_PASSES {
-        let resp = es.search(edges_index, &query)?;
+        // `search_present`, not `search`: the exclusion sweep (#694) may call
+        // this before the edges index has ever been created (it runs ahead of
+        // the graph phase's `ensure_index`). A missing index is not a failure —
+        // it simply holds no edges to invalidate. The replacement-invalidation
+        // caller always ensures the index first, so its behaviour is unchanged
+        // (the index is present → `Some`).
+        let Some(resp) = es.search_present(edges_index, &query)? else {
+            return Ok(total);
+        };
         let hits = resp
             .pointer("/hits/hits")
             .and_then(Value::as_array)
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -349,6 +349,12 @@ fn search_http(path: &str, body: &[u8], state: &Arc<Mutex<HttpState>>) -> Value 
     let exists = query
         .pointer("/query/bool/must/1/exists/field")
         .and_then(Value::as_str);
+    // #694: `invalidate_prior_edges` filters live edges with
+    // `must_not:[{exists:{field:"invalid_at"}}]`; honour it so the pass
+    // converges (an already-invalidated edge must drop out of the result).
+    let must_not_exists = query
+        .pointer("/query/bool/must_not/0/exists/field")
+        .and_then(Value::as_str);
     let locked = state.lock().unwrap();
     let matching: Vec<_> = locked
         .docs
@@ -386,6 +392,9 @@ fn search_http(path: &str, body: &[u8], state: &Arc<Mutex<HttpState>>) -> Value 
                             })
                 })
                 && exists.map(|field| doc.get(field).is_some()).unwrap_or(true)
+                && must_not_exists
+                    .map(|field| doc.get(field).is_none())
+                    .unwrap_or(true)
         })
         .collect();
     let hits = matching
@@ -2067,7 +2076,7 @@ fn sweep_excluded_groups_deletes_only_the_excluded_files_documents() {
         path: "secret.csv".into(),
     }];
 
-    sweep_excluded_groups(&es, &plan, &excluded).expect("sweep");
+    sweep_excluded_groups(&es, &plan, &excluded, None, 0).expect("sweep");
 
     let st = ep.state.lock().unwrap();
     // The excluded file's dataset record is gone.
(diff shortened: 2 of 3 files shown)


xerj-org:
## Independent verification (3 adversarial skeptics, rustc 1.97.1, isolated worktrees)

Each skeptic reset to `origin/fix/694-sweep-excluded-edges` (HEAD `535e4c95`), asserted the fix symbols present, then tried to REFUTE from a distinct angle. **All three: MERGE_OK, ZERO BLOCKING.**

### Correctness → **MERGE_OK**
> Could not construct any BLOCKING correctness defect. Over-invalidation is impossible: excluded entries carry a `path` that `between()` proves is not any current file's rel (`path_survives` checks `current_rels`), and edges always key `src_file` to the teaching file's own rel, so no live file's edges can collide. Under-invalidation is closed: `entry.path` (=`assignment.rel`) equals the edges' `src_file` (=walk `f.rel`) under identical normalization, verified by the passing test. The `search`→`search_present` swap is behavior-identical when the index is present; the replacement caller ensures the index first — its e2e invalidation test passes unchanged. `graph_edges_index` correctly returns None under `--no-graph`/invalid brain, and the graph phase only writes edges when the brain validates, so the gate is `Some` whenever edges could exist. Fail-before confirms the test genuinely guards the fix. CI-parity: fmt 0, clippy `--all-targets` clean, xerj-autoindex lib 706/0.
> - *nit:* `search_present` means a concurrent delete between the replacement caller's `ensure_index` and its `invalidate_prior_edges` would early-return `Ok` instead of erroring — practically unreachable (single-run, index just ensured), behavior-preserving.
> - *nit:* invalidate-by-`src_file` only sweeps edges the excluded file TAUGHT; inbound/dst-side edges owned by a surviving file are left live — inherent edge-ownership model, identical to the reused replacement hook, out of #694 scope, pre-existing.

### Regression → **MERGE_OK**
> Full xerj-autoindex suite (706 lib + 14 integration) green under ci-test; fmt clean; clippy `--all-targets -D warnings` exit 0. The shared mock `_search` `must_not` change is strictly additive — gated on the exact pointer `/query/bool/must_not/0/exists/field`, defaults to no-op, cannot alter any other bool-query test (the only other must_not producer is term-based; the exists-based map-summary query is a graph-only path the `no_graph:true` mock never reaches). Both call sites pass correct args (`prior_plan` at 3814, `&plan` at 4081; `edges_index.as_deref()`; a real timestamp). `graph_edges_index` mirrors the graph phase's brain derivation. No over-invalidation (exact `src_file` term, unique rels).
> - *nit:* relies on `InventoryDeltaEntry.path` and edges' `src_file` being identically-normalized — same path coupling the pre-existing content-record delete already depends on; a divergence only fails to close the gap (never data loss/regression).

### Evidence-holds → **MERGE_OK**
> Correct tree (HEAD `535e4c95`, all fix symbols present). New test passes. Fail-before is decisive: reverting ONLY the `invalidate_prior_edges` call to a no-op makes the test fail with exactly `left: None / right: Some(1724500000000)` — not tautological; it proves the excluded edge is stamped `invalid_at` via a genuine search→bulk→refresh round-trip. The mock honours the `src_file` filter and the new `must_not:[{exists:invalid_at}]` so the pass converges instead of spinning to the 1000-pass cap. The second assertion (a different file's edge stays live) is enforced by real `src_file` filtering, guarding over-invalidation. `search_present` maps 404→Ok(None); the replacement caller still sees `Some`.
> - *nit:* the fail-before failure aborts via SIGABRT (mock server-thread cleanup) rather than unwinding — shared test-infra behavior across the whole file, not introduced here.

**Decision:** 0 BLOCKING → merge once every check is green + none pending + mergeStateStatus CLEAN. The 4 nits are non-blocking; the dst-side-edge one (a genuine follow-up: a removed file's *inbound* edges) will be filed as a follow-up issue.

PR #734, opened 2026-08-24
fix(autoindex): exclusion sweep purges a removed file's alias catalog docs (#693)

### LINKED ISSUE #693: autoindex: #589 exclusion sweep leaves alias/duplicate catalog docs (file-alias:) live for byte-identical duplicates
Non-blocking SHOULD-FIX from the 3-skeptic verification of #690 (#589).

`sweep_excluded_groups` deletes the excluded file's catalog doc by `{term:{path: entry.path}}` where `entry.path` = the canonical `assignment.rel`. Duplicate/alias catalog docs written by `catalog::duplicate_file_doc` (catalog.rs:209-235) have id `file-alias:...` and carry the DUPLICATE's own `path` (the alias rel), not the canonical rel — so the term on the canonical path does not match them and they survive.

Impact: **path/filename metadata residue only** — the content records (keyed by `ax_file=file_key`) ARE fully swept by #589, so no document content is exposed. But an excluded file's alternate path/filename stays searchable in the catalog as a `file-alias:` doc.

Fix: also delete `file-alias:` docs for the excluded file — either by `ax_file`/`file_key` (if the alias doc carries it) or by resolving the alias rels. Verify against a byte-identical-duplicate corpus.

xerj-org:
Closes #693 — the sibling #589 exclusion-sweep follow-up (alias/duplicate catalog residue).

## Problem
`sweep_excluded_groups` deleted the excluded file's content records (by `ax_file`) and its canonical `file:` catalog doc (by `path`), but left its **`file-alias:` duplicate catalog docs** (`catalog::duplicate_file_doc`) live. Each alias carries the **duplicate's own `path`**, not the canonical `entry.path`, so the `path` term missed them — an excluded file's alternate path/filename stayed searchable in the catalog. (Metadata residue only: the content records were already swept by #589, so no document content was exposed.)

## Fix
Both the main `file:` doc and every `file-alias:` doc carry **`file_key`** (`catalog.rs`). A content group is classified excluded **only when no surviving file bears its key** — `UnsupportedInventoryDelta::between` skips a group whose key is still in `current_keys` — so deleting catalog docs by `{term:{file_key}}` **cannot strand a still-live byte-identical duplicate**. Added that delete alongside the existing `path` delete; the tested #589 main-doc sweep is left unchanged.

## Test / fail-before
`sweep_excluded_groups_purges_the_excluded_files_alias_catalog_docs`: the excluded file's alias (seeded under a **different** path, `dup.csv`) is swept, while another file's alias (distinct `file_key`) survives (`file_key` is content-scoped — no over-deletion). **Fail-before** (file_key delete reverted): the alias doc survives the canonical-path sweep — the assertion discriminates the fix. The mock `delete_by_query` already handles an arbitrary `{term}` field, so no harness change was needed.

- `cargo test -p xerj-autoindex` (full suite, ci-test) → **706 lib + integration, 0 failed**

Sibling of #732 (#694, graph edges) — both touch `sweep_excluded_groups` at distinct points; whichever lands second takes a trivial rebase. No `:9200` conformance run touched.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -2183,6 +2183,87 @@ fn sweep_excluded_groups_invalidates_the_excluded_files_edges() {
     );
 }
 
+/// #693: `sweep_excluded_groups` also purges the excluded file's `file-alias:`
+/// duplicate catalog docs. Those carry the DUPLICATE's own path (not the
+/// canonical `entry.path`), so the `path` term alone misses them — leaving an
+/// excluded file's alternate path/filename searchable. Deleting by `file_key`
+/// catches them, and cannot strand a live duplicate (a group is only excluded
+/// when no surviving file bears its key). Fail-before: without the `file_key`
+/// delete, the alias doc (path "dup.csv") survives the canonical-path sweep.
+#[test]
+fn sweep_excluded_groups_purges_the_excluded_files_alias_catalog_docs() {
+    let ep = HttpEndpoint::start();
+
+    {
+        let mut st = ep.state.lock().unwrap();
+        // Excluded file: a canonical `file:` doc plus a byte-identical
+        // `file-alias:` doc under a DIFFERENT path — both carry the same
+        // `file_key`.
+        st.docs.insert(
+            (catalog::CATALOG_INDEX.to_string(), "file:excluded".into()),
+            json!({"doc_kind": "file", "file_key": "key-excluded", "path": "canonical.csv"}),
+        );
+        st.docs.insert(
+            (catalog::CATALOG_INDEX.to_string(), "file-alias:ax:excluded".into()),
+            json!({"doc_kind": "file", "file_key": "key-excluded", "path": "dup.csv", "status": "duplicate"}),
+        );
+        // A different file's alias (distinct file_key) must survive.
+        st.docs.insert(
+            (catalog::CATALOG_INDEX.to_string(), "file-alias:ax:kept".into()),
+            json!({"doc_kind": "file", "file_key": "key-kept", "path": "kept-dup.csv", "status": "duplicate"}),
+        );
+    }
+
+    let es = Es::with_bulk_timeout(&ep.url, None, 30).expect("es client");
+    let mut plan = Plan::default();
+    plan.datasets.push(crate::state::PlanDataset {
+        slug: "ds".into(),
+        index: "incremental-http-ds".into(),
+        family: "csv".into(),
+        group: None,
+        specs: Vec::new(),
+        time_field: None,
+        semantic_field: None,
+        sampled_records: 1,
+        file_count: 1,
+    });
+    // entry.path is the CANONICAL rel — deliberately NOT the alias's "dup.csv".
+    let excluded = vec![InventoryDeltaEntry {
+        file_key: "key-excluded".into(),
+        path: "canonical.csv".into(),
+    }];
+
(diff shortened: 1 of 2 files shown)


xerj-org:
## Independent verification (3 adversarial skeptics, rustc 1.97.1, isolated worktrees)

Each skeptic reset to the PR branch, asserted the fix present, then tried to REFUTE from a distinct angle. **All three: MERGE_OK, ZERO BLOCKING.**

> **Rebased note:** this branch was rebased onto `main` after #732 (#694) merged — both PRs edit `sweep_excluded_groups`. Resolution keeps both (catalog `path` delete → new catalog `file_key` delete → #694 edge invalidation) and updates this PR's test to the post-#694 5-arg signature. Verified on the rebased tree (`74015b00`): fmt exit 0, clippy `--all-targets` clean, full xerj-autoindex suite **708/0** (all 3 sweep tests green). The skeptic analysis below is on the identical `#693` logic and holds through the rebase.

### Correctness → **MERGE_OK**
> Within-corpus over-deletion refutation FAILED — the PR is correct. `UnsupportedInventoryDelta::between` admits a group to `excluded_content_groups` only when `!current_keys.contains(key) && !path_survives`. `current_keys` is built from ALL current-walk files' content keys, so any surviving byte-identical duplicate keeps the key present and the group is SKIPPED, never excluded. `plan.files` is keyed by content key, so an excluded `entry.file_key` can never equal a live same-corpus file's key. The alias doc's `file_key` is the shared CONTENT key (`catalog.rs:223 == :196`); `file_key` is a keyword field so the term matches exactly. Could not construct a within-corpus case that deletes a live file's catalog doc.
> - *should-fix (filed #737):* both catalog sweeps (`{term:{path}}` #589 and the new `{term:{file_key}}`) run unscoped against the single GLOBAL catalog index — a byte-identical common file in a **sibling** corpus could have its regenerable catalog metadata over-deleted. No content loss; **pre-existing** flaw the #589 path sweep already has; #693 extends the same model, does not introduce it.
> - *nit:* the new test covers only single-corpus; the cross-corpus axis is untested.

### Regression → **MERGE_OK**
> Full xerj-autoindex suite green (706 lib + integration, 0 failed). The pre-existing `sweep_excluded_groups_deletes_only_the_excluded_files_documents` PASSES despite seeding catalog docs without a `file_key` field — the mock `delete_http` retains any doc where `doc.get(field) != Some(expected)`, so a doc lacking `file_key` is untouched. fmt clean; clippy `--all-targets -D warnings` exit 0. The extra per-entry `delete_by_query` double-hits the main `file:` doc via both path and file_key — idempotent, harmless. Only `file_doc`/`duplicate_file_doc` emit `file_key`, so no collateral targets.
> - *nit:* the by-path catalog delete is now partially redundant (the file_key delete subsumes the main `file:` doc). Kept as belt-and-suspenders.

### Evidence-holds → **MERGE_OK**
> Fail-before is proven and DISCRIMINATING: neutering ONLY the `file_key` delete (`let _ = &entry.file_key;`) makes the test fail exactly at "the excluded file's file-alias: catalog doc must be swept" — the alias under path `dup.csv` survives the canonical-path sweep. Not tautological: the alias is seeded under a DIFFERENT path than `entry.path`, so only the file_key delete can catch it, and a distinct-`file_key` alias is asserted to SURVIVE (guarding over-deletion). Every duplicate's `file_key` is its canonical content key, so `entry.file_key` catches all aliases.
> - *nit:* path delete redundant (same as regression nit).

**Decision:** 0 BLOCKING → merge once every check is green + none pending + mergeStateStatus CLEAN. The should-fix is filed as follow-up **#737** (pre-existing, non-blocking); nits are cosmetic.

PR #735, opened 2026-08-24
fix(release,site): CLI banner version, /get.sh installer, self-updating RC number

xerj-org:
Fixes three confirmed defects users reported in the download/version chain.

**1. CLI banner froze at rc.18** (the main one). Releases advanced to rc.67 but every binary's `--version` said rc.18. Root cause: `release.yml` names artifacts from the tag but compiles from `engine/Cargo.toml` (`env!(CARGO_PKG_VERSION)`), last bumped at rc.18. The workflow now sets the workspace version from the tag before building and syncs `Cargo.lock`. (Verified rc.67's tarball hash matches its published `.sha256` — only the embedded version was wrong.)

**2. `/get.sh` served HTML.** It wasn't in the Pages Functions route include, so `curl …/get.sh | sh` piped index.html to sh — also the likely source of 'hash mismatch' reports. Added `functions/get.sh.js` (re-exports the installer) + route. `/get` and `/get.ps1` already worked.

**3. Landing hardcoded 'RC.15'.** Now filled from GitHub's latest release at load with a JS-off fallback, so it self-updates; CSP widened to read-only `api.github.com`.

Validated end-to-end before this PR: `/get` resolves latest correctly; rc.67 hash matches; `--version` mismatch reproduced. The site half deploys via the new Pages Actions workflow (#733) once merged.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -156,6 +156,13 @@ jobs:
               export "$line"
             done <<< '${{ matrix.linker_env }}'
           fi
+          # Binary --version is env!("CARGO_PKG_VERSION"); without this it stays
+          # frozen at whatever engine/Cargo.toml last shipped (rc.18), so every
+          # release self-reported rc.18. Derive it from the tag.
+          VER="${GITHUB_REF_NAME#v}"
+          sed -i.bak -E "0,/^version = \".*\"/s//version = \"${VER}\"/" Cargo.toml && rm -f Cargo.toml.bak
+          echo "set workspace version to ${VER}"; grep -m1 '^version = ' Cargo.toml
+          cargo update --workspace --offline 2>/dev/null || cargo update --workspace
           if [ "${{ matrix.use_zigbuild }}" = "true" ]; then
             cargo zigbuild --release --locked --target ${{ matrix.target }} -p xerj-server
           else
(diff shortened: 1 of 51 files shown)


xerj-org:
3-skeptic pass: 3 BLOCKING, not merging as-is. All confirmed:

1. release.yml `build` runs on `push: branches:[main]` and dispatch too, not just tags (only `release` is tag-gated). There `GITHUB_REF_NAME`=main so `VER=main` → `version = "main"` → cargo rejects it → build fails. Merging this reddens the cross-compile canary on the next main push.
2. The `0,/re/` sed is GNU-only. macos-14 legs run BSD sed, which errors on address 0; sed is non-final in `sed && rm` so errexit does not abort → macOS binaries still ship rc.18. Fix on Linux/Windows only.
3. `/get.sh` is now routed to the Function but `get.js` scriptFor only maps `/get` and `/get.ps1`, so it returns 404 "not found", not the installer. `curl -fsSL .../get.sh | sh` still fails.

Landing self-update + CSP: clean (one cosmetic nit, footer strings drift).

Reworking on the branch: tag-guard the version step, portable perl replace, map /get.sh→get in scriptFor.

xerj-org:
Re-review after the rework (3feef508): 3 skeptics, all MERGE_OK, 0 blocking. Each verified: perl stamps engine/Cargo.toml line 24 (workspace.package, inherited by xerj-server via version.workspace=true), cargo update re-syncs the lock, the v-tag guard skips canary/main builds; /get.sh->'get' serves the real installer, get.sh.js re-exports onRequestGet+onRequestHead. Merging on green+CLEAN.

Non-blocking follow-ups: static /get.sh fallback twin (https://github.com/xerj-org/xerj/issues/740); /get.sh installs count under script='get'. Note: defect 3 (RC.15/CSP) already landed separately via c29ddb0c, so this PR carries the 2 remaining.

PR #738, opened 2026-08-24
fix(autoindex): prefix-scope the exclusion catalog sweeps so a sibling corpus is not over-deleted (#737)

### LINKED ISSUE #737: autoindex: exclusion/replacement catalog sweeps are unscoped against the GLOBAL shared catalog index — a byte-identical file in a sibling corpus can have its catalog metadata over-deleted
Surfaced as a non-blocking should-fix during the 3-skeptic verification of #734 (#693).

`sweep_excluded_groups` deletes catalog docs by `{term:{path}}` (#589) and now also `{term:{file_key}}` (#693), both **unscoped** against the single GLOBAL `autoindex-catalog` index (`catalog.rs`), which is shared across all corpora/prefixes. Catalog doc `_id`s are prefix-scoped (`file:{prefix}:{key}`, `file-alias:{prefix}:…`), but there is **no queryable `prefix` field** in the catalog mapping — so a `term` cannot be constrained to the current corpus.

## Impact (metadata only, no content loss)
Because `file_key` is content identity (xxh3 of first 64KB + size) and `path` collides on common root names, a byte-identical common file (LICENSE, a lockfile, README.md) indexed by a **still-live sibling corpus B** shares the same key/path. A widened exclusion in corpus A fires the sweep and can delete **Bs live `file:`/`file-alias:` catalog docs** too. The `current_keys` guard in `UnsupportedInventoryDelta::between` only holds corpus As walk keys, so it does not protect this cross-corpus axis.

Severity is low: the swept docs are **regenerable catalog metadata**, not searchable content — dataset indices are prefix-scoped and the `ax_file` record sweep touches only this plans datasets. And this is **pre-existing**: the `{term:{path}}` sweep (#589) already has the identical unscoped-shared-index behavior; #693 extends the same operating model rather than introducing a new defect class.

## Fix
Prefix-scope both catalog sweeps — either add a queryable `prefix` keyword field to the catalog mapping and `bool{filter:[{term:{prefix}}, {term:{file_key|path}}]}`, or delete by the exact prefix-scoped `_id`s. Also worth a cross-corpus regression test (seed a catalog doc under a different prefix `_id` with the same `file_key`, assert it survives a sibling exclusion). Applies to `sweep_excluded_groups` and the analogous replacement/junk catalog deletes.

Filed from the #734 skeptic pass; not blocking that merge. Relates to #673 (catalog prefix-scoping for correlations) and #416.

xerj-org:
Closes #737 — surfaced as a should-fix during the #734 (#693) skeptic pass.

## Problem
`sweep_excluded_groups` deleted catalog docs by `{term:{path}}` (#589) and `{term:{file_key}}` (#693), both **unscoped** against the single global `autoindex-catalog` index (shared across corpora). Because `file_key` is content identity and `path` collides on common root names (LICENSE, lockfiles), a byte-identical file indexed by a still-live **sibling** corpus B shares the key/path — so a widened exclusion in corpus A would delete B's live `file:`/`file-alias:` catalog docs too. The `current_keys` guard in `UnsupportedInventoryDelta::between` only covers the same corpus.

Metadata-only (dataset indices are prefix-scoped; content records were never at risk), but a live sibling corpus losing its catalog `file:`/`file-alias:` docs makes its files unfindable/unresolvable in the catalog.

## Fix
The catalog doc `_id`s are prefix-scoped (`file:{prefix}:…`) but `_id` is not term-queryable. Add a queryable **`prefix`** keyword field — written by `file_doc`/`duplicate_file_doc`, mapped in `catalog_mapping()` — and constrain both catalog sweeps to `bool{filter:[{term:{prefix}}, {term:{path|file_key}}]}`. The dataset `ax_file` record sweep is already prefix-scoped by index (unchanged). **Backward compatible**: old catalog docs without a `prefix` field fall outside a scoped sweep (never over-deleted); new runs write the field.

## Test / fail-before
`sweep_excluded_groups_does_not_delete_a_sibling_corpus_catalog_doc`: a sibling corpus's byte-identical doc (same `file_key`, prefix `bx`) **survives** corpus `ax` excluding the file, while `ax`'s own doc is swept. **Fail-before** (prefix filter reverted → unscoped `{term:{file_key}}`) deletes the sibling — the assertion discriminates the fix. The mock `delete_http` now honours `bool.filter` multi-term (delete iff *all* terms match); existing sweep tests updated to seed a `prefix`.

- `cargo test -p xerj-autoindex` (full suite, ci-test) → **709 lib + integration, 0 failed**

**Stacked on #734** (#693) — both edit `sweep_excluded_groups`; base retargets to `main` when #734 merges. Fixes the pre-existing unscoped-sweep flaw #737 identified (also present in the #589 path sweep). No `:9200` touched.

### DIFF
--- a/engine/crates/xerj-autoindex/src/catalog.rs
+++ b/engine/crates/xerj-autoindex/src/catalog.rs
@@ -108,6 +108,12 @@ pub fn catalog_mapping() -> Value {
         json!({"type": "keyword"}),
     );
     properties.insert("junk_records_this_run".into(), json!({"type": "long"}));
+    // #737: corpus scope, so a delete_by_query can constrain a sweep to THIS
+    // corpus's docs. The doc `_id`s are already prefix-scoped (`file:{prefix}:…`)
+    // but `_id` is not term-queryable; a byte-identical file shared with a live
+    // sibling corpus would otherwise be caught by an unscoped `file_key`/`path`
+    // sweep. Written by `file_doc`/`duplicate_file_doc` (the sweep's targets).
+    properties.insert("prefix".into(), json!({"type": "keyword"}));
     mapping
 }
 
@@ -193,6 +199,7 @@ pub fn file_doc(
         file_id(prefix, file_key),
         json!({
             "doc_kind": "file",
+            "prefix": prefix, // #737: corpus scope for a scoped exclusion sweep
             "file_key": file_key,
             "path": path,
             "format": format,
@@ -220,6 +227,7 @@ pub fn duplicate_file_doc(
         alias_id,
         json!({
             "doc_kind": "file",
+            "prefix": prefix, // #737: corpus scope for a scoped exclusion sweep
             "file_key": file_key,
             "path": path,
             "format": "duplicate",
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -306,21 +306,45 @@ fn delete_http(path: &str, body: &[u8], state: &Arc<Mutex<HttpState>>) -> usize 
         .next()
         .unwrap();
     let query: Value = serde_json::from_slice(body).unwrap();
-    let term = query
+    // Collect every term constraint: the single `/query/term` form AND each
+    // `/query/bool/filter[].term` (#737 scopes catalog deletes with a
+    // `prefix` + `path`/`file_key` bool filter). A doc is deleted iff it
+    // matches EVERY collected term.
+    let mut terms: Vec<(String, Value)> = Vec::new();
+    if let Some((field, value)) = query
         .pointer("/query/term")
         .and_then(Value::as_object)
         .and_then(|term| term.iter().next())
-        .map(|(field, value)| (field.clone(), value.clone()));
+    {
+        terms.push((field.clone(), value.clone()));
+    }
+    if let Some(filters) = query
+        .pointer("/query/bool/filter")
+        .and_then(Value::as_array)
+    {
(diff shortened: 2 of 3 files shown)


xerj-org:
Note on existing catalogs (preempting the "legacy catalog has no prefix field" concern): the `prefix` keyword is declared on **already-created** catalogs, not just fresh ones. The generation setup runs an additive catalog mapping upgrade every run — `run_index_report` (lib.rs:200-206) builds `catalog_update_body` from `catalog::catalog_mapping()`'s full property set (now including `prefix`) and calls `es.update_mapping(CATALOG_INDEX, …)` before any `file_doc`/`duplicate_file_doc` write. `prefix` never existed in any prior release, so this is a pure additive add (no `mapper_parsing_exception` type-conflict like the `started` tripwire) — it is declared `keyword` before the first value is written, so `{term:{prefix}}` resolves exactly on legacy catalogs too. No reindex needed, no regression to the #589/#693 sweep behavior on existing corpora.

xerj-org:

One should-fix, flagged by all 3: the mapping upgrade adds the `prefix` field but does not backfill values on docs written by a pre-#737 binary, so a file cataloged by the old binary and excluded on the first new-binary run keeps a stale catalog doc. Metadata only (content still swept by the `ax_file` delete), not data loss, not a regression vs any release. Tracked in #739.

Correction to my earlier note: the additive upgrade is schema-only, it does not backfill doc values. Legacy docs excluded on the first post-#737 run stay stale until reindex.

fail-before reproduced (drop the prefix filter, sibling doc gets deleted), full xerj-autoindex suite 709/0, fmt/clippy clean. Merges after #734 on green+CLEAN.

Issue #739, opened 2026-08-24
autoindex: exclusion sweep skips catalog docs written before #737 (prefix not backfilled)


xerj-org:
From the #738 (#737) skeptic pass.

#737 scopes the catalog sweep to `{term:{prefix}}`. The mapping upgrade adds the `prefix` field but does not backfill values on docs written by a pre-#737 binary. So a file cataloged by the old binary and excluded on the first new-binary run keeps its stale `file:`/`file-alias:` catalog doc (the scoped term misses a doc with no prefix value).

Bounded: upgrade window only, metadata only. Content records still go via the unchanged `ax_file` delete. No data loss, not a regression vs any release (the sweep is new in this stack).

Fix: backfill `prefix` (by prefix-scoped `_id`) on the mapping upgrade, or also delete the main `file:` doc by exact `_id` `file_id(prefix,file_key)`. Aliases need their path_ids resolved or an `_id`-prefix delete.

xerj-org:
Main-doc half fixed on main (#743, merge 5a2e2945): the exclusion sweep now also deletes the main `file:` catalog doc by its exact prefix-encoded `_id`, so a legacy doc written before #737 (no `prefix` field) is swept on the first upgraded run, corpus-safe.

Remaining (why this stays open): legacy `file-alias:` duplicate docs still aren't swept — their ids need the per-alias path_ids, which the sweep doesn't have. Narrow (pre-#737 corpus with byte-identical duplicates, excluded on the first #737-era run), metadata-only.

PR #743, opened 2026-08-24
fix(autoindex): sweep a legacy file: catalog doc by id (#739 main-doc half)

xerj-org:
Addresses the main-doc half of #739.

#737's prefix-scoped catalog sweep only matches docs that carry the `prefix` field (added in #737). A `file:` doc written by a pre-#737 binary has no `prefix` value and survives the sweep on the first upgraded run.

Also delete the main `file:` doc by its exact prefix-encoded `_id` (`catalog::file_id`). It catches legacy and current docs, and is corpus-scoped by construction (a sibling corpus's id differs), so it cannot strand or over-delete.

Legacy `file-alias:` docs can't be reconstructed by id here without the per-alias path_ids — a smaller residual, so **#739 stays open** for that.

## Test
`sweep_excluded_groups_deletes_a_legacy_main_doc_by_id`: a legacy doc (no `prefix` field) is swept by id, a sibling corpus's same-key legacy doc survives. Fail-before proven (id-delete reverted → legacy doc survives). Mock `delete_http` now handles the `ids` query.

fmt/clippy clean, full xerj-autoindex suite 710/0, rustc 1.97.1.

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -332,11 +332,24 @@ fn delete_http(path: &str, body: &[u8], state: &Arc<Mutex<HttpState>>) -> usize 
             }
         }
     }
+    // #739: an `ids` query deletes by exact `_id` (the main `file:` doc sweep).
+    let ids: Option<Vec<String>> = query
+        .pointer("/query/ids/values")
+        .and_then(Value::as_array)
+        .map(|arr| {
+            arr.iter()
+                .filter_map(|v| v.as_str().map(String::from))
+                .collect()
+        });
     let mut locked = state.lock().unwrap();
     let before = locked.docs.len();
-    locked.docs.retain(|(candidate, _), doc| {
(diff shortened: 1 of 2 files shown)


xerj-org:

- correctness: file_id="file:{prefix}:{file_key}"; prefix is slug-sanitized (no colons), so a sibling corpus id differs and QueryNode::Ids is exact _id equality. No over-delete (single exact id). catalog::file_id matches the write path.
- regression: full suite 710/0; mock delete_http ids branch fires only when /query/ids/values is present, so term/bool.filter deletes are unchanged. fmt/clippy clean.
- evidence: fail-before discriminating (neuter the id-delete → the no-prefix legacy doc survives → test fails at the exact assertion); sibling bx doc survives. Confirmed a real window: #416 (Aug 23) put the prefix in the _id, #737 (Aug 24) added the prefix field, so docs between have a prefix-encoded id but no field — exactly this delete.

Nits only: the legacy file-alias: residual (documented, #739 stays open for it) and a pre-existing SIGABRT teardown noise in the mock. Merging on green+CLEAN.

PR #744, opened 2026-08-24
docs(feedback): report v1.0.0-rc.67 artifact verification

Nicolas0315:
Written by an AI coding agent (Codex / GPT-5), run by @Nicolas0315.

Adds the required field report only. It records the reproducible published-artifact version mismatch: every v1.0.0-rc.67 archive reports v1.0.0-rc.18, although archive checks and the host-native smoke workflow passed.

Verified:
- bash scripts/verify-release.sh v1.0.0-rc.67 -> 8 archives present, checksum-clean and unpackable; 9 failures caused by the version mismatch; native boot/index/search/aggregation smoke operations passed.
- cargo build --release -j 32 -p xerj-server on current main -> passed in 6m 41s.
- The committed diff contains one new 13-line Markdown field report and nothing else.

Assumed:
- The current tag-version stamping step will prevent the mismatch on a future tag; that release has not yet been produced.

Not run:
- ES-YAML conformance suite, because this adds no engine code or documentation claim.

### DIFF
--- a/user-feedback/16-agent-field-reports/2026-08-25-release-artifact-verification.md
+++ b/user-feedback/16-agent-field-reports/2026-08-25-release-artifact-verification.md
@@ -1,0 +1,13 @@
+# Release artifact verification (2026-08-25)
+
+**Agent:** Codex (GPT-5)  ·  **XERJ:** published `v1.0.0-rc.67` reports `xerj v1.0.0-rc.18`  ·  **Platform:** macOS arm64
+
+**Pointed at:** All eight archives in the latest GitHub release; no user corpus was indexed.
+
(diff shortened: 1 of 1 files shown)


xerj-org:
Thanks, this is a real bug and your repro is exactly right. Root cause: release.yml named artifacts from the tag but compiled from engine/Cargo.toml, whose workspace version was last bumped at rc.18, so env!(CARGO_PKG_VERSION) (what --version prints) stayed frozen while the tags advanced. Fixed in #741 (merged): the build now stamps the workspace version from the tag before compiling. Future RCs report their real version.

Two caveats your report already implies: rc.67's ALREADY-published binaries keep the embedded rc.18 (can't be retro-fixed without re-releasing), and the first attempt at the fix (#735) shipped a broken variant that #741 corrected. The 9 failures you saw are all this single embedded-version mismatch, consistent with your note that the archives are checksum-clean and boot/index/search/agg all pass.

Issue #736, opened 2026-08-24
autoindex: a removed file's INBOUND (dst-side) graph edges are left dangling by both the exclusion sweep and replacement invalidation


xerj-org:
Surfaced as a non-blocking nit during the 3-skeptic verification of #732 (#694).

Both `detect::invalidate_prior_edges` (the replacement hook) and the #694 exclusion-sweep reuse invalidate edges by `{term:{src_file: rel}}` — i.e. only the edges the removed/excluded file **taught** (outbound, src-side). An edge whose `src_file` is a *surviving* file but whose `dst_file`/`dst` endpoint is the removed file (e.g. a `samedir`/`sharedterm`/`href` edge where the rel-smaller endpoint survives) stays **live**, now pointing at a `dst` with no source file.

## Impact
Low severity — a dangling edge endpoint, not document content. The anchor node it references is gone, so hydrating/focusing that edge yields a ghost dst. Consistent and **pre-existing**: this is the inherent edge-ownership model, identical between the replacement path (`invalidate_prior_edges`, lib.rs ~4672) and the new exclusion sweep (#732). #694 deliberately scoped to "a removed file's OWN edges", so this is out of scope for that fix and is **not a regression** introduced by #732.

## Possible fix
When invalidating a removed file `rel`, also soft-invalidate live edges with `{term:{dst_file: rel}}` (dst-side), not only `src_file`. Would need a decision on whether a surviving src file should keep an edge to a removed dst at all (some detectors may legitimately want the historical link), so it likely wants the same bi-temporal soft-invalidate (stamp `invalid_at`, keep for `as_of`) rather than a hard delete. Verify against a two-file samedir/sharedterm corpus where the dst endpoint is excluded.

Filed from the #732 skeptic pass; not blocking that merge.

xerj-org:
Exclusion-sweep half fixed on main (#746, merge 79ce0307): sweep_excluded_groups now also soft-invalidates INBOUND edges — ones a surviving file taught that point at the removed file's anchor node (dst = ids::doc_id(slug, file_key, "file")). No edge schema change (dst is already keyword). 3 skeptics, 0 blocking.

Stays open for: the replacement-invalidation path (a file whose CONTENT changed → new file_key → the OLD anchor is orphaned, and inbound edges to it dangle if the referencing file isn't re-indexed). That path computes edges from the new key; invalidating the old anchor needs the old file_key, which isn't handy there — separate change.

xerj-org:
[...]

PR #746, opened 2026-08-24
fix(autoindex): exclusion sweep invalidates inbound graph edges (#736 sweep half)

xerj-org:
Addresses the exclusion-sweep half of #736.

`sweep_excluded_groups` invalidated only the edges a removed file **taught** (`src_file`). Edges a surviving file taught that point **at** the removed file's anchor node (`dst`) stayed live, dangling at a node that is gone. Now also soft-invalidate those.

The edge `dst` is the target file's anchor doc id `ids::doc_id(slug, file_key, "file")`, computed from the frozen plan's slug(s) for the excluded key (a file can anchor in several datasets). Same bi-temporal soft-invalidate as the src side — **no edge schema change** (`dst` is already a keyword).

Refactored `invalidate_prior_edges` → `invalidate_edges_by_field(field, value)`; `invalidate_prior_edges` is now a `src_file` wrapper, byte-identical for the replacement + #694 callers.

## Test
`sweep_excluded_groups_invalidates_inbound_edges`: an inbound edge to the excluded anchor is invalidated, an edge to another node stays live. Fail-before proven (dst-side reverted → inbound edge keeps no `invalid_at`). Full xerj-autoindex suite 711/0, fmt/clippy clean, rustc 1.97.1.

The replacement-invalidation dst-side (a file whose content changed, old anchor orphaned) is a separate code path, so **#736 stays open** for it.

### DIFF
--- a/engine/crates/xerj-autoindex/src/detect/mod.rs
+++ b/engine/crates/xerj-autoindex/src/detect/mod.rs
@@ -655,17 +655,22 @@ pub fn ensure_brain_meta(
     Ok(())
 }
 
-/// Soft-invalidate every live edge taught by `rel` (replacement hook, §6.6.3).
-/// Runs BEFORE any of this run's edges for that file are written, so the
-/// `src_file` + no-`invalid_at` query can only match prior-generation edges.
+/// Soft-invalidate every live edge whose keyword `field` equals `value`.
 /// Each pass re-indexes its hits with `invalid_at = expired_at = now` and
 /// refreshes, so repeat-until-empty converges; the pass bound turns a broken
-/// endpoint into an error instead of an infinite loop.
-pub fn invalidate_prior_edges(es: &Es, edges_index: &str, rel: &str, now_ms: i64) -> Result<u64> {
+/// endpoint into an error instead of an infinite loop. Callers pass `src_file`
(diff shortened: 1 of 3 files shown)


xerj-org:

- correctness: the anchor the sweep computes, ids::doc_id(slug, file_key, "file"), is exactly what edges store in dst (detect/mod.rs:197, every cross-file detector). No over-invalidation (delta marks a key excluded only when no surviving file bears it, so no live duplicate shares the anchor). No under-invalidation (the edge write side picks one slug from fa.assignments, which the sweep iterates in full). plan.files.get(entry.file_key) always resolves (excluded entries are drawn from the same plan). Refactor is behavior-preserving.
- regression: full suite 711/0, fmt/clippy clean; mock _search honors bool.filter term + must_not exists so the dst loop converges in one pass.
- evidence: fail-before discriminating (neuter the dst loop → inbound edge survives, left None vs Some(now_ms); the src-side sweep can not catch it since src_file=survivor).

Nits only (non-blocking): N-1 empty invalidation passes for a multi-dataset file (harmless superset); the bulk-error context string "invalidate prior edges" is now generic (outer with_context says "sweep inbound graph edges"). Merging on green+CLEAN.

Issue #522, opened 2026-08-19
Follow-ups from #446: add a non-ONNX embedder-swap refusal test; fold proxy endpoint into identity


xerj-org:
Follow-ups from the verification of #446 (embedder-mismatch guard, #434). Neither blocks that PR.

### 1. No test covers the actual #434 refusal path (should-fix)
The new refusal lives in `validate_embedding_execution` (index.rs ~27754-27793). Every existing refusal test is ONNX-only (`changed_onnx_assets_are_rejected` etc.) and exercises the OLD `embedding_identity.json` marker path, not the new sidecar. No test (i) builds an index under one embedder with documents, (ii) reopens under a different resolved embedder, (iii) asserts the `embedding_execution.json` `Err`. Such a test is cheap and needs no feature flag: `lexical` → `proxy`-with-endpoint yields distinct identity hashes and would exercise the 27770-27780 branch. Add one so the exact class of silent regression #434 is about cannot come back uncaught.

### 2. A changed proxy *endpoint* under the same model name does not move the identity hash (missed-detection edge)
`embedding_execution_identity`'s proxy material includes `configured-model` but not `default_endpoint`. Two different proxy endpoints serving the same model name resolve to the same `identity_sha256`, so swapping the endpoint (a genuinely different embedder) would NOT be refused. This is a missed detection, not a false refusal, and not the lexical↔neural case #434 fixed — but it is the same failure family. Consider folding the endpoint into the proxy identity material.

Provenance: adversarial verification of #446.

PR #747, opened 2026-08-25
fix(engine): fold the proxy endpoint into the embedder identity (#533)

### LINKED ISSUE #533: Fold the proxy default_endpoint into the embedder execution identity (part 2 of #522)
Split from #522 (the test half landed via #532). `embedding_execution_identity` for a `proxy` backend hashes the model name but **not** `default_endpoint`, so swapping the proxy endpoint under the same model name does not move `identity_sha256` — a changed embedder that the #434 guard fails to detect (a missed detection, not a false refusal).

Fix: include `default_endpoint` in the proxy identity material. **Migration concern:** this changes the hash for existing proxy indices, so an already-populated proxy index would be *refused* on upgrade (its persisted identity was computed without the endpoint). Bump `PersistedEmbeddingExecution.version` to 2 and grandfather version-1 records (treat a version-1 vs version-2 delta as 'unknown', not 'mismatch') so upgrade does not brick existing proxy indices. Add a test: same model + different endpoint → distinct identity → refused; and a version-1 record still opens.

xerj-org:
Closes #533 (part 2 of #522).

`embedding_execution_identity` for a `proxy` backend hashed the model name but **not** `default_endpoint`, so swapping the proxy endpoint under the same model name did not move `identity_sha256` — a changed embedder the #434 guard fails to detect (a missed detection, not a false refusal). Querying one embedder's vectors against another's is silently wrong.

- Include `default_endpoint` in the proxy identity material (`api_key` stays out — a rotated key over the same endpoint+model is the same embedder).
- **Migration:** this changes the hash for existing proxy indices, so bump the identity version to 2 and grandfather v1 records — `validate_embedding_execution` treats a persisted-vs-current version delta as "unknown" (not a mismatch) and re-records at the current version, so upgrade does not brick a populated proxy index and a **later** endpoint change under v2 is still detected. A same-version hash delta is still a genuine embedder change → refused.

## Tests
- `proxy_endpoint_is_part_of_the_execution_identity`: same model + different endpoint → distinct identity, and a populated proxy index reopened under a swapped endpoint is refused.
- `a_version_1_record_is_grandfathered_then_rebaselined`: a v1 record is not refused on upgrade, is re-recorded at v2, and a later v2 endpoint change IS refused.

Fail-before proven (revert the endpoint fold-in → different endpoints hash identical). Full xerj-engine suite green, fmt/clippy clean, rustc 1.97.1. Unit-level — no `:9200`.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -29033,6 +29033,15 @@ const EMBEDDING_IDENTITY_FILE: &str = "embedding_identity.json";
 /// proxy endpoint or neural model — which is the half a mode check cannot see.
 const EMBEDDING_EXECUTION_FILE: &str = "embedding_execution.json";
 
+/// On-disk record version for `embedding_execution.json`. Distinct from the
+/// wire `EmbeddingExecutionIdentity.version` (the API contract, pinned == 1 by
+/// the autoindex client). Bumped to 2 by #533: the proxy identity_sha256 now
+/// folds in the endpoint, so a v1 record's hash was produced by a different
+/// algorithm — `validate_embedding_execution` treats a version delta as
+/// "unknown" (grandfather + re-record), never a mismatch, so upgrading does not
+/// brick a populated proxy index; a later change under v2 is still detected.
+const PERSISTED_EMBEDDING_EXECUTION_VERSION: u32 = 2;
+
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 struct PersistedEmbeddingExecution {
     version: u32,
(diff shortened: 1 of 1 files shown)


xerj-org:
First 3-skeptic pass found 2 BLOCKING (both now fixed; also caught by CI's use-case harness — good reason to gate on green):

1. Wire-contract break (regression skeptic + CI harness): bumping EmbeddingExecutionIdentity.version 1->2 broke the autoindex client, which pins version == 1 (esclient.rs:509) — 'unsupported embedding execution identity', bricking semantic autoindex against a same-commit server. Fix (74f02278): the wire version stays 1; the migration marker moved to a separate on-disk const PERSISTED_EMBEDDING_EXECUTION_VERSION = 2. identity_sha256 still moves via the endpoint fold-in, which is what the guard needs.

2. Over-broad grandfather (evidence skeptic): the version-delta grandfather skipped the mismatch check without checking backend, so a v1 index reopened under a DIFFERENT backend (lexical->proxy) got a one-time free pass — silently mixing vector spaces, the exact #434 corruption. Fix (3fda4bf7): a backend-name change is refused at any version; only a same-backend hash delta across an older record is grandfathered. New test a_version_delta_still_refuses_a_backend_change.

All 15 embedding tests + full xerj-engine suite green, fmt/clippy clean. Re-running the skeptic pass on 3fda4bf7.

xerj-org:
Re-verification pass on 3fda4bf7: 3 skeptics, all MERGE_OK, 0 blocking. Both prior fixes confirmed correct + load-bearing.

- wire fix: embedding_execution_identity still emits version:1 (no server path emits 2); esclient.rs:509 (!=1 -> bail) passes for every backend; PERSISTED_EMBEDDING_EXECUTION_VERSION=2 is on-disk only. PR touches no autoindex code.
- backend guard: genuine_change = has_documents && (backend changed || (!older_record && hash changed)) gives all four cases right; fail-before on the NEW check proven (revert it -> the lexical->proxy v1 test wrongly admits -> fails). Uses < not != (no downgrade re-record).
- full xerj-engine suite 638+/0 (re-run UNPIPED to get the true exit), xerj-autoindex identity 22/0, fmt=0, clippy clean.

Non-blocking nit: the version-delta grandfather isn't scoped to backend==proxy, so it suppresses a same-backend hash delta for ALL backends on the one-time upgrade. Zero impact this release (only the proxy material changed, so non-proxy shas are byte-identical across the bump); a latent consideration for a FUTURE bump that changes a non-proxy hash. Left as-is to avoid speculative complexity. Merging on green+CLEAN.

PR #748, opened 2026-08-25
fix(ci): resolve the merge before it lands + hoist tree-sitter (#352)

### LINKED ISSUE #352: CI: resolve the merged workspace before merge — two green PRs produced an unbuildable main (tree-sitter links conflict)

xerj-org:
Closes #352.

Two green PRs merged into a tree that would not resolve: #327 pinned tree-sitter 0.25, #317 moved to 0.26, and tree-sitter is a `links` crate so only one version may exist. The merge was a hard resolution failure — build, test, clippy and every release job at once — and no gate caught it because CI only ever resolved each PR head, never the merge.

**New `workspace-resolves` job**: `cargo metadata --locked` against the merge ref (`actions/checkout` gives `refs/pull/N/merge` by default on pull_request). Seconds, no compile. Fails on any `links` conflict or Cargo.lock drift. Verified it catches the exact #352 failure — reintroducing a 0.25 pin gives `cargo metadata` exit 101 with "package tree-sitter links to the native library tree-sitter, but it conflicts...".

**Hoist tree-sitter core** into `[workspace.dependencies]`; both crates use `{ workspace = true }`. Structural belt-and-suspenders so the highest-frequency recurrence can't happen. Behaviour-identical: same version (0.26.12), Cargo.lock unchanged, both crates build.

Grammar pins stay per-crate — the guard covers those if they ever drift. Prevention idea credited to @buger in the #351 review notes.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,6 +141,28 @@ jobs:
         working-directory: engine
         run: cargo clippy --workspace --all-targets -- -D warnings
 
+  # #352: #327 (tree-sitter 0.25) and #317 (0.26) were each green on their own
+  # head, but tree-sitter declares `links = "tree-sitter"` so only one version
+  # may exist — the MERGE was a hard resolution failure that took out build,
+  # test, clippy and every release job at once, and stayed red long enough that
+  # ~every open PR inherited it. No gate caught it because each branch resolved
+  # its own head, never the merge. This resolves the merge in a few seconds, no
+  # compile: `actions/checkout` gives refs/pull/N/merge by default on a
+  # pull_request, so this runs against the merged tree. `--locked` also fails on
+  # Cargo.lock drift.
+  workspace-resolves:
+    name: Workspace resolves (links + lockfile)
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
+
+      - name: Install Rust
+        uses: dtolnay/rust-toolchain@032958afbdc797a9164d3bc0b56325c1308924a5 # 1.97.1
+
+      - name: cargo metadata (resolves the merge; fails on a links conflict or lock drift)
+        working-directory: engine
+        run: cargo metadata --locked --format-version 1 >/dev/null
+
   # XERJ's security pitch is "no JVM, so none of the Log4Shell class of supply
   # chain CVEs" — which is only worth anything if somebody is actually watching
   # the 660-odd crates we do depend on. This job is that somebody. It was
--- a/engine/Cargo.toml
+++ b/engine/Cargo.toml
@@ -109,6 +109,13 @@ reqwest = { version = "0.12", default-features = false, features = ["json", "rus
 tempfile = "3"
 # Benchmarking
 criterion = { version = "0.5", features = ["html_reports"] }
+# AST parsing. Only tree-sitter CORE is hoisted here: it declares
+# `links = "tree-sitter"`, so two crates pinning different core versions is a
+# hard resolution failure that takes out the whole workspace at once, not a
+# duplicate-dep warning (#352: #327 pinned 0.25, #317 moved to 0.26, and the
+# merge stopped resolving). One pin, one version. Per-crate grammar pins stay
+# local; the `workspace-resolves` CI gate catches any links crate that drifts.
+tree-sitter = "0.26"
 
 [profile.release]
 opt-level = 3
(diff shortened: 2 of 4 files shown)


xerj-org:
Independent 3-skeptic pass (correctness / regression / evidence), each refuting in its own worktree under rustc 1.97.1. Verdicts verbatim.

**All three: MERGE_OK — 0 BLOCKING.**

---

**correctness — MERGE_OK**
The guard works and the hoist is behaviour-identical. `cargo metadata --locked` returns exit 0 on the clean branch; Cargo.lock has a single tree-sitter 0.26.12 and is unchanged. The guard provably catches a links conflict: pinning xerj-engine to tree-sitter =0.25.0 fails with exit 101 and the exact 'links to the native library tree-sitter ... conflicts ... only one package may specify the same links value' error, both with and without --locked. It also provably catches lock drift (=0.26.13 -> exit 101, 'cannot update the lock file because --locked was passed'). ci.yml triggers on pull_request:[main], so the job's default checkout resolves the merge ref; --locked is present; YAML is valid. The hoist changes both crates from tree-sitter="0.26" to { workspace = true } with workspace.dependencies tree-sitter="0.26" — identical ^0.26 requirement, no features added, no other crate declares tree-sitter core, resolution and lockfile unchanged. No BLOCKING defect.
- should-fix: the workspace-resolves job resolves refs/pull/N/merge only when it runs; GitHub does not auto-rerun a green PR's checks when main advances, so the exact stale-base race is only fully closed if branch protection requires branches up to date. Repo-config, outside this diff; the hoist mitigates it structurally.

**regression — MERGE_OK**
No genuine regression. engine/Cargo.lock diff vs origin/main is empty; tree-sitter resolves to 0.26.12 on both branches. `cargo build -p xerj-engine -p xerj-autoindex --profile ci-test` succeeds. Only xerj-engine and xerj-autoindex reference tree-sitter core; both correctly switched to { workspace = true } against the single workspace pin, which equals the old per-crate pins. Grammar crates stay per-crate and untouched; no build.rs references tree-sitter — hoisting only the core is safe. The new job inherits contents:read, shares the concurrency group, does no compile/cache, so it cannot interfere with existing jobs. A legitimate consistent lock update passes --locked; only an inconsistent merged-tree lock REDs, which is intended. No BLOCKING problem.
- nit: a PR whose lock is consistent on its own head but stale after auto-merge with a moved main can RED until rebased — intended #352-class catch, mild rebase friction.

**evidence — MERGE_OK**
Every literal, checkable claim verifies TRUE. Guard proven: reproduced the 0.25-vs-0.26 links conflict, `cargo metadata --locked` exits 101 with the exact tree-sitter links error text. Behaviour-identical verified: Cargo.lock still resolves 0.26.12, not in the commit diff, both crates pass cargo check --locked. Workspace pin ^0.26 identical to old per-crate pins; hoist complete. Also confirmed --locked catches pure lock drift. No BLOCKING problem.
- should-fix: the stated root-cause ('CI only ever resolved each head, never the merge') is imprecise — existing jobs use the same default merge-ref checkout, so the new job's incremental value is speed + --locked lock-drift, not merge visibility; the actual stale-base/no-re-run gap remains and the HOIST is the real recurrence fix. Reword the comment/commit narrative.

---

Merging on 0 BLOCKING + all-green + CLEAN. The two should-fixes are repo-config (branch-protection up-to-date) and comment-prose accuracy — neither is a code defect; filing the prose fix as a follow-up.

PR #750, opened 2026-08-25
docs(ci): fix the workspace-resolves root-cause comment (#749)

### LINKED ISSUE #749: ci: workspace-resolves comment/#748 commit overstate the guard's novelty
From the #748 skeptic pass (evidence lens, should-fix).

The ci.yml `workspace-resolves` comment and the #748 commit say "no gate caught #352 because CI only ever resolved each head, never the merge." That's imprecise: every job on a pull_request already checks out refs/pull/N/merge (the merge ref), so build-test would fail a links conflict on the merge too. The real #352 gap is stale-base + GitHub not re-running a green PR's checks when main advances — which the new job shares.

The guard's genuine value is fast `--locked` lock-drift detection + a dedicated seconds-not-40min signal; the tree-sitter recurrence is prevented by the workspace hoist, not the job. Reword the comment + note this. Docs-only.

xerj-org:
Closes #749.

Comment-only. The `workspace-resolves` comment claimed no gate caught #352 because CI "resolved each head, never the merge." Existing jobs already checkout refs/pull/N/merge, so that's inaccurate (caught in the #748 skeptic pass). Corrected: the real gap is a stale base plus GitHub not re-running a green PR when the base advances, which this job shares; its value is a fast dedicated signal + `--locked` lock-drift detection, and the workspace hoist is what structurally prevents the tree-sitter recurrence.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -141,15 +141,18 @@ jobs:
         working-directory: engine
         run: cargo clippy --workspace --all-targets -- -D warnings
 
-  # #352: #327 (tree-sitter 0.25) and #317 (0.26) were each green on their own
-  # head, but tree-sitter declares `links = "tree-sitter"` so only one version
-  # may exist — the MERGE was a hard resolution failure that took out build,
-  # test, clippy and every release job at once, and stayed red long enough that
-  # ~every open PR inherited it. No gate caught it because each branch resolved
-  # its own head, never the merge. This resolves the merge in a few seconds, no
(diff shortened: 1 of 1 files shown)


xerj-org:
Comment-only change: diff touches only the workspace-resolves comment block (no job/step/logic), YAML valid, guard intact. Zero behaviour/binary surface, so proportionate self-verification (diff scope + YAML + accuracy vs the #748 evidence finding) in place of a 3-worktree skeptic pass. All 19 checks green + CLEAN. Merging.

PR #752, opened 2026-08-25
fix(autoindex): name the field a catalog read-back disagrees on (#367)

### LINKED ISSUE #367: xc-index.sh cannot index apache/lucene: autoindex exits 1 on both embed modes, wrapper discards a fully-populated corpus, xc.py then refuses it

xerj-org:
Advances #367 step 2 (the lexical `catalog read-back for ds:docs disagrees with the sealed generation projection` abort), which currently gives no way to see *which* field the server rewrote.

`validate_observed` now names the differing field(s) instead of aborting opaquely, and reports a **dropped** key as `` `time_min` dropped ``. The prime suspect for #367 is exactly that: `dataset_doc` emits `time_min`/`time_max`/`time_field`/`semantic_field` as literal `null` for a timestamp-less, non-semantic dataset like `ds:docs`, and a null-dropping read-back path (the #415 class) would silently remove them → `actual != expected`. This is diagnostic-only (no behaviour change); it turns the opaque abort into an actionable message so the next apache/lucene run pinpoints the culprit field.

Does **not** close #367 — the disagreement itself needs the apache/lucene repro to confirm the field, which this makes possible. Fail-before: revert the `describe_field_diff` call and the test asserting the field is named fails. Full xerj-autoindex suite green.

### DIFF
--- a/engine/crates/xerj-autoindex/src/generation_catalog.rs
+++ b/engine/crates/xerj-autoindex/src/generation_catalog.rs
@@ -89,8 +89,10 @@ impl CatalogProjection {
 fn describe_field_diff(expected: &Value, actual: &Value) -> String {
     fn short(v: &Value) -> String {
         let s = v.to_string();
-        if s.len() > 120 {
-            format!("{}…", &s[..120])
+        // Truncate on a CHAR boundary (values can hold non-ASCII, e.g. a CJK
+        // `fields_json`); byte slicing `&s[..120]` would panic mid-codepoint.
+        if let Some((idx, _)) = s.char_indices().nth(120) {
+            format!("{}…", &s[..idx])
         } else {
             s
         }
@@ -785,6 +787,16 @@ mod tests {
         assert!(
             err.contains("time_min") && err.contains("dropped"),
             "a dropped key must be named: {err}"
+        );
+
+        // A long non-ASCII differing value must be reported, not panic the
+        // truncation (byte-slicing would split a multibyte codepoint).
+        let mut wide = projection.documents.clone();
+        wide.get_mut("ds:ax:reports").unwrap()["notes"] = json!(["名前".repeat(60)]);
+        let err = projection.validate_observed(&wide).unwrap_err().to_string();
+        assert!(
+            err.contains("notes"),
+            "a wide non-ASCII diff must be reported: {err}"
         );
     }
 }
--- a/engine/crates/xerj-autoindex/src/generation_catalog.rs
+++ b/engine/crates/xerj-autoindex/src/generation_catalog.rs
@@ -49,22 +49,76 @@ impl CatalogProjection {
     /// outside this corpus projection and therefore are not accepted through
     /// this API.
     pub fn validate_observed(&self, observed: &BTreeMap<String, Value>) -> Result<()> {
-        anyhow::ensure!(
-            observed.len() == self.documents.len(),
-            "catalog read-back contains {} documents; expected {}",
-            observed.len(),
-            self.documents.len()
-        );
+        if observed.len() != self.documents.len() {
+            let expected_ids: BTreeSet<&String> = self.documents.keys().collect();
+            let observed_ids: BTreeSet<&String> = observed.keys().collect();
+            let missing: Vec<&&String> = expected_ids.difference(&observed_ids).collect();
+            let unexpected: Vec<&&String> = observed_ids.difference(&expected_ids).collect();
+            anyhow::bail!(
+                "catalog read-back contains {} documents; expected {} (missing: {:?}; \
+                 unexpected: {:?})",
+                observed.len(),
+                self.documents.len(),
+                missing,
+                unexpected
+            );
+        }
         for (id, expected) in &self.documents {
             let actual = observed
                 .get(id)
                 .with_context(|| format!("catalog read-back is missing {id}"))?;
             anyhow::ensure!(
                 actual == expected,
-                "catalog read-back for {id} disagrees with the sealed generation projection"
+                "catalog read-back for {id} disagrees with the sealed generation \
+                 projection: {}",
+                describe_field_diff(expected, actual)
             );
         }
         Ok(())
+    }
+}
+
+/// Name the field(s) by which a read-back catalog document differs from the
+/// sealed projection, so a mismatch is diagnosable rather than an opaque
+/// "disagrees". #367: a real-server run aborted here with no way to see which
+/// field the server rewrote — the prime suspect is a dropped `null`-valued key
+/// (`time_min`/`time_max`/`time_field`/`semantic_field` are `null` for a
+/// timestamp-less, non-semantic dataset like `ds:docs`), which this surfaces as
+/// "`time_min` dropped". Values are truncated so a large `fields_json` cannot
+/// blow up the message.
+fn describe_field_diff(expected: &Value, actual: &Value) -> String {
+    fn short(v: &Value) -> String {
+        let s = v.to_string();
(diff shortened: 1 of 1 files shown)


xerj-org:
Independent 3-skeptic pass (correctness / regression / evidence), each refuting in its own worktree under rustc 1.97.1. Verdicts verbatim.

**All three: MERGE_OK — 0 BLOCKING.**

---

**correctness — MERGE_OK**
The target test passes on 1.97.1 and clippy --all-targets is clean. The char-boundary truncation is provably panic-free: char_indices() yields offsets that are always valid UTF-8 boundaries, so &s[..idx] cannot split a codepoint. Confirmed with a standalone battery covering 3-byte CJK, 4-byte emoji, combining marks, flag sequences, exactly-120 and 121-char strings, empty, and mixed-boundary inputs — zero panics; exactly-120 hits nth(120)==None and returns whole. The diff detection is total and correct: detects dropped keys, changed values, added keys; the non-object arm always emits a diff; the "no field-level difference" fallback is unreachable for objects because the field walk uses the same key-set + per-value equality as Value's ==. Array reordering is correctly reported as a change. Diagnostic-only — the abort still fires.
- nit: two long values differing only after char 120 print identical-looking truncated prefixes; still names the field and never claims equality, so not misleading — diagnostic granularity only.
- nit: nested object/array differences are reported at the top field level with truncated values rather than the exact nested path; granularity limitation, not a defect.

**regression — MERGE_OK**
No behaviour change beyond the message. The diff confirms abort conditions are unchanged: count mismatch `ensure!(len==len)` → `if len!=len { bail! }` (same trigger), missing-id `get().with_context()?` unchanged, per-document `ensure!(actual==expected, ...)` condition byte-for-byte identical — only the message now calls describe_field_diff. That helper is a format arg to ensure!/bail!, so it runs strictly on the error path (zero happy-path cost). Owned String, no borrow escape; truncation cannot panic mid-codepoint. Full crate suite green under 1.97.1 (lib 711 passed/0 failed, all integration binaries ok, doctests ok); clippy --all-targets -D warnings clean. No Err→Ok flip, no clean Err becomes a panic.
- nit: values are truncated but not object keys; a pathological field name could enlarge the message. Not a regression — catalog field names are bounded and the prior message had no key at all.

**evidence — MERGE_OK**
Every claim holds. Fail-before is genuine: reverting only the message enhancement makes the extended test fail with the opaque message; it passes once restored. "Diagnostic-only, no behaviour change" is literally true — validate_observed returns Err in exactly the same conditions; only the text changed. The dropped-null claim is confirmed: dataset_doc (catalog.rs:164-167) emits time_field/time_min/time_max/semantic_field from Option straight into json!, so None becomes JSON null, surfaced as "`time_min` dropped (was null)". The PR does NOT claim to close #367 (body says so explicitly). The char-boundary fix is real: a byte slice &s[..120] panics mid-codepoint ('inside 前') on the 名前×60 test, which the shipped char_indices version passes. Full xerj-autoindex lib suite green (711/0) on 1.97.1.
- nit: truncation is per-value, not per-message; a doc differing in many fields could still emit a long string. Diagnostic-only, harmless.

---

Merging on 0 BLOCKING + all-green (19/19) + CLEAN. The 4 nits are diagnostic-granularity limitations (not defects); leaving them as-is (the output is correct and useful). This does not close #367 — it makes step 2 diagnosable so the next apache/lucene run pinpoints the field.

Issue #751, opened 2026-08-25
CI: Build+Test step 6 (workspace tests at default parallelism) intermittently hangs on hosted runners


xerj-org:
The `Workspace tests (all targets) at default parallelism` step of Build+Test intermittently hangs on ubuntu-latest. Normal duration is ~2.3 min (e.g. green main run 32795328344: step6 01:18:59->01:21:19). On #748 it hung twice back to back: ~28 min then ~18 min, both needing a manual cancel + `gh run rerun --failed`; the 3rd attempt passed in the normal ~2.3 min.

Not PR-specific: #748 was config-only (Cargo.toml + a CI comment), so the test binaries are byte-identical to main. Running step 6's exact command locally — `cargo test --profile ci-test --workspace` at 20-core default parallelism — passes clean (183 binaries, exit 0). So the suite is sound; this is a hosted-runner timing race, almost certainly a port-binding / shared-fixture integration test at the 2-4 vCPU default thread count (the class flagged in the ci.yml #385 note; step 5 runs the same suite at `--test-threads=2` and never hangs).

Second-order effect: main-branch CI runs don't auto-cancel (concurrency `cancel-in-progress: false` for main, by design), so a hung build-test leaves the main run stuck for hours — 33dfc6b2 has been in_progress ~4h, and recent main runs show cancelled/stuck. Main commits then never get a completed CI signal.

Fix direction: bisect the hanging binary (run each integration target under a timeout at default parallelism), then serialise its port/fixture access or bound its threads. Relates to #384 / #385.

PR #754, opened 2026-08-25
fix(autoindex): #739 - sweep a legacy file-alias: catalog doc by id (alias half)

### LINKED ISSUE #739: autoindex: exclusion sweep skips catalog docs written before #737 (prefix not backfilled)

Vinz2168:
Closes #739.

#743 already fixed the "main-doc half": a `file:` catalog doc written by a pre-#737 binary has no `prefix` field, so it survives the #737 prefix-scoped sweep on the first upgraded run — fixed by also deleting it by its exact prefix-encoded `_id`.

The same gap exists on the `file-alias:` side, and was called out explicitly as the residual left open by #743: a pre-#737 alias doc also has no `prefix` field, so it survives both the #737 prefix-scoped delete and the #693 term-scoped alias delete.

## Fix

`sweep_excluded_groups` already receives `prior_plan` (the frozen plan). `Plan.duplicate_files` still carries every alias's `rel`/`path_id` for a content group being excluded now — the exact inputs `catalog::duplicate_file_id` needs to reconstruct an alias's id, the same pattern this file already uses elsewhere (`stale_alias_ids`). This reconstructs each matching alias's id and folds it into the same `ids`-delete the main-doc fix added, so one `delete_by_query` now sweeps the main doc and every legacy alias doc for the excluded file_key.

## Testing

- Added `sweep_excluded_groups_deletes_a_legacy_alias_doc_by_id`, mirroring the existing `sweep_excluded_groups_deletes_a_legacy_main_doc_by_id`: a pre-#737 (no `prefix`) alias doc in the excluding corpus is swept by id, while a sibling corpus's same-key legacy alias doc (different id) survives.
- Fail-before verified: stashed the `lib.rs` fix and confirmed the new test fails; restored and it passes.
- `cargo test -p xerj-autoindex --lib` — 690 passed. The 19 failures present are pre-existing on a clean `main` checkout too (unrelated `failure_resume_http_tests`/`inventory_delta_tests`/`sync_executor` — likely the parallelism flakiness tracked in #751), not introduced by this change.

## Provenance
- [x] Read #739, #743, and the code both touch before writing this fix
- [x] `cargo build -p xerj-autoindex`
- [x] `cargo test -p xerj-autoindex --lib sweep_excluded_groups` — 7/7 passed
- [x] Fail-before verified the new test against the reverted fix
- [x] `cargo test -p xerj-autoindex --lib` — confirmed the 19 failures are pre-existing on `main`, not from this change

Claude-Session: https://claude.ai/code/session_01DgiP3zVSST92nyxLmTkM4h

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -2431,6 +2431,77 @@ fn sweep_excluded_groups_deletes_a_legacy_main_doc_by_id() {
     );
 }
 
+/// #739 (alias half): a `file-alias:` catalog doc written by a pre-#737
+/// binary has NO `prefix` field either, so it survives both the #737
+/// prefix-scoped alias sweep AND the #693 term-scoped one on the first
+/// upgraded run — the residual the main-doc fix (#739 above) left open. The
+/// frozen `plan.duplicate_files` still carries the alias's `rel`/`path_id`,
+/// so the sweep can reconstruct its exact `_id` and delete it directly.
+#[test]
+fn sweep_excluded_groups_deletes_a_legacy_alias_doc_by_id() {
+    let ep = HttpEndpoint::start();
+    let legacy_alias_id = catalog::duplicate_file_id("ax", "legacy-key", "alias.csv", "pid1");
+    let sibling_alias_id = catalog::duplicate_file_id("bx", "legacy-key", "alias.csv", "pid1");
+
+    {
+        let mut st = ep.state.lock().unwrap();
+        // Legacy alias docs: no `prefix` field (pre-#737). Corpus "ax"
+        // excludes the canonical file; corpus "bx" holds a same-key alias
+        // under its own (differently prefix-encoded) id.
+        st.docs.insert(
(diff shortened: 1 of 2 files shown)


xerj-org:
I filed #739 and wrongly said the legacy alias residual was blocked because the sweep lacks the alias path_ids. This is the right fix — `plan.duplicate_files` carries the alias `rel`/`path_id`, so reconstructing the `_id` via `duplicate_file_id` is exactly the missing piece. Corpus-scoping is correct too (the sibling test proves a differently-prefixed id survives). Nice catch.

Issue #755, opened 2026-08-25
autoindex: global catalog abort when `prefix` pre-exists as text (post-#737; e.g. after upgrade)


xerj-org:
#737 added `prefix: keyword` to the autoindex-catalog mapping (catalog.rs:116) and made catalog docs carry a `prefix` string (catalog.rs:202). `ensure_generation_mappings` installs the keyword mapping with a hard `update_mapping`. If the **global** autoindex-catalog already has `prefix` mapped as `text` (a pre-#737 run, or any node whose shared catalog picked up a dynamic text `prefix`), the update 400s and the run aborts:

```
install generation catalog mapping: PUT /autoindex-catalog/_mapping failed: 400
field [prefix] already exists as [text], cannot add [keyword]
```

autoindex-catalog is one shared global index across all corpora/prefixes, so once it holds a text `prefix`, every subsequent autoindex run fails, not just the one that caused it.

Repro (fresh server, seed a text `prefix` first to stand in for an existing deployment):

```
xerj --insecure --port 9700 -d "$(mktemp -d)" &
curl -s -XPOST 'http://127.0.0.1:9700/autoindex-catalog/_doc/seed?refresh=true' \
  -H 'content-type: application/json' -d '{"doc_kind":"file","prefix":"ax"}'
xerj autoindex /corpus --no-graph --url http://127.0.0.1:9700 --state-dir "$(mktemp -d)" --no-semantic
# aborts with the mapping conflict
```

A genuinely fresh catalog is fine: keyword installs before any doc is written.

Fix direction: the catalog `update_mapping` must not hard-abort the run on a `prefix` text-vs-keyword conflict on an existing catalog. Either skip `prefix` when it already exists incompatibly, or make prefix-scoped catalog queries work against a text `prefix` (via its `.keyword` subfield). Needs care: #737's prefix scoping (alias sweeps in `alias_sweep_query`) relies on exact matching, so tolerating text must not silently break corpus scoping. A catalog reindex/migration is the heavier alternative.

---

Note: originally filed as "fails on any fresh corpus" with a concurrency-race theory. That was wrong. The original repro set `XERJ_URL`, which the autoindex CLI ignores (it reads only `--url`), so every run silently hit the default loopback :9200, whose shared catalog already held a text `prefix`. The fresh `--no-graph` path works (verified 3/3 at default workers). The earlier `--workers 1` workaround was spurious (it only ever reflected :9200's mutable shared state).

xerj-org:
[...]

PR #756, opened 2026-08-25
fix(autoindex): explain the legacy-catalog mapping conflict instead of an opaque 400 (#755)

xerj-org:
Part of #755. Turns the opaque catalog mapping conflict into an actionable error.

A global `autoindex-catalog` created by an older build can hold a field dynamically mapped as `text`. The generated mapping declares the catalog fields as `keyword` (corpus-scoped exclusion sweeps depend on exact matching; #737 added `prefix`), and Elasticsearch cannot change a field's type in place, so `update_mapping` aborts every run with a raw `mapper_parsing_exception` naming whichever field it checked first.

This wraps that specific error with the cause and the data-preserving recovery (reindex the catalog into an index with the current mapping), and names the actual conflicting field. Unrelated errors pass through untouched.

Deliberately NOT an auto-migration: the catalog is global across every corpus and its scoping drives deletions, so migrating it is a maintainer decision, not a silent side effect of a run. That migration is the remaining open work on #755.

Fresh catalogs are unaffected (keyword installs before any doc write; verified 3/3 at default workers).

Verify:
- fail-before: raw `field [X] already exists as [text]` with no guidance.
- fail-after: leads with "the global `autoindex-catalog` was created by an older build: its `X` is mapped as text ... Reindex ... and retry", raw reason preserved in the chain.
- `cargo test -p xerj-autoindex` 714 passed; clippy `--all-targets -D warnings` clean; both profiles build (1.97.1).

### DIFF
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -204,7 +204,98 @@ pub(crate) fn ensure_generation_mappings(es: &Es, plan: &Plan) -> Result<()> {
     });
     es.ensure_index(catalog::CATALOG_INDEX, &catalog_create_body)?;
     es.update_mapping(catalog::CATALOG_INDEX, &catalog_update_body)
+        .map_err(|e| explain_legacy_catalog_mapping_conflict(catalog::CATALOG_INDEX, e))
         .context("install generation catalog mapping")
+}
+
+/// The generated catalog declares its fields as `keyword` (exact corpus-scoped
+/// exclusion sweeps depend on it; #737 added `prefix`). A catalog created by an
+/// older build can hold one of those fields dynamically mapped as `text`, and
+/// Elasticsearch cannot change a field's type in place, so the additive
+/// `update_mapping` fails with an opaque `mapper_parsing_exception` naming
+/// whichever field it checked first. Name the cause and the (operator-driven,
+/// data-preserving) recovery instead of surfacing the raw type conflict.
+///
+/// An automatic reindex is deliberately NOT done here: the catalog is global
+/// across every corpus and its scoping drives deletions, so migrating it is a
+/// maintainer decision, not a silent side effect of a run.
+fn explain_legacy_catalog_mapping_conflict(index: &str, err: anyhow::Error) -> anyhow::Error {
+    let msg = format!("{err:#}");
+    if let Some(field) = legacy_text_field_in_conflict(&msg) {
+        return err.context(format!(
+            "the global `{index}` was created by an older build: its `{field}` is mapped as text, \
+             but this version needs the catalog fields as keyword (corpus-scoped sweeps depend on \
+             exact matching) and Elasticsearch cannot change a field's type in place. Reindex \
+             `{index}` into an index with the current mapping (create a fresh index with the \
+             keyword mapping, `_reindex` into it, then swap), and retry"
+        ));
+    }
+    err
+}
+
+/// Pull the field name out of an ES `field [X] already exists as [text], …`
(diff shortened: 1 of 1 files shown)


xerj-org:

**CORRECTNESS — MERGE_OK**
- nit: legacy_text_field_in_conflict does a greedy rfind("field [") + naive find(']'), so a field name containing brackets (e.g. "field [[nested]]") extracts "[nested" and an empty-named field extracts "". Neither can occur with real engine field names and neither panics, so it is purely cosmetic in the enriched message.

Rationale: PANIC concern refuted: every slice boundary lands on an ASCII marker (`already exists as [text]`, `field [`, `]`), so byte indices are always char boundaries even with multibyte content; all `?` short-circuit and index arithmetic is bounded. Confirmed empirically over adversarial unicode, nested/missing/truncated brackets, empty strings, the real duplicated-root_cause JSON body, and 2,000,000 random fragment combos — zero panics. FALSE-POSITIVE refuted: enricher is wired only at the catalog update_mapping call and only matches genuine text-vs-keyword conflicts, naming the actual field dynamically. FALSE-NEGATIVE refuted: engine message format matches the parser literal; the reqwest error string preserves the substring verbatim. anyhow .context chain preserves the original reason. All 3 unit tests pass under 1.97.1.

**REGRESSION — MERGE_OK**
- nit: Parser keys on xerj-engine phrasing `field [X] already exists as [text]`. Against a stock upstream Elasticsearch endpoint (`mapper [X] cannot be changed from type [text]...`) the wrapper no-ops and passes the error through unchanged — no worse than current behavior (not a regression), it just doesn't improve the non-xerj-endpoint case.

Rationale: The map_err/context wiring is inert on success (Ok passes through). explain_legacy_catalog_mapping_conflict only ever returns Err (err.context or the original err) — never Err->Ok or Ok->Err — so which runs succeed vs abort is unchanged; it is a pure error-message wrapper. legacy_text_field_in_conflict is panic-safe (find/rfind guarded by ?, ASCII-anchored slices), so a malformed message returns None and passes through. No existing test asserts the old opaque text as the final surfaced message. Full suite under 1.97.1: 714 lib + all integration binaries pass, 0 failed; clippy --all-targets -D warnings exit 0; fmt exit 0.

**EVIDENCE-HOLDS — MERGE_OK**
- should-fix: No integration/end-to-end test guards the .map_err wiring. The 3 unit tests pass even with the wiring line removed (they call the helper directly), so a future refactor could silently drop the enrichment at the update_mapping call site and CI would stay green while the opaque error returns. A regression test exercising ensure_generation_mappings against a legacy text-mapped catalog would lock it in. [-> follow-up #760]
- nit: The parser matches Path B's phrasing (es_compat.rs:2321), not Path A's `mapper [X] cannot be changed from type [text] to [keyword]`. For the documented bug this is correct — a dynamically-inferred text field is absent from index_mappings so Path A is skipped and Path B fires (verified); Path A is unreachable for this bug.

Rationale: (a) fail-before — reverting ONLY the .map_err wiring surfaces the raw opaque mapper_parsing_exception with no cause/recovery, while the helper unit tests still pass; (b) end-to-end — with the wiring, the real run enriches the error with the older-build cause, names the conflicting field (doc_kind), gives the reindex recovery, and preserves the raw reason verbatim. Verified the dynamic-inference scenario routes through the server's Path B message (which the helper matches). fmt/clippy/unit tests green under 1.97.1. Fresh-catalog no-op.

---
**Gate: 0 BLOCKING findings across all three lenses.** Merge proceeds once CI (rebased onto the #757-fixed main) is green+CLEAN. should-fix + nits filed as follow-up.

PR #758, opened 2026-08-25
fix(autoindex): cap section records per file + flag truncation (#381)

xerj-org:
First hunk of #381 (draft — see remaining step below).

A magic-less printable byte payload (a raw texture, a packed `.bytes` asset) decodes under the text fallback, classifies as prose and is sectioned in full — one record per `SECTION_CHARS`, so a large blob expands into thousands of noise records with no ceiling of its own (the `16 MiB` txt read cap alone still allows ~8k).

This caps `emit_document` at `MAX_RECORDS_PER_FILE` (4096, tunable) and sets `ExtractStats.truncated` so the drop is carried out for reporting — never silent. Keyed off record count only, never byte statistics (those deleted CJK text worldwide and were removed in #371).

Verify:
- fail-before: with the cap hunk removed, `emit_document_caps_records_and_reports_truncation` fails (`left: 9, right: 3`).
- fail-after: capped to the limit + `truncated` set; a document under the cap is unchanged and not flagged.
- `cargo test -p xerj-autoindex` 713 passed; clippy `--all-targets -D warnings` clean; both profiles build (1.97.1).

Remaining (why draft): surface `truncated` on the progress meter along the existing `ExtractStats` → sketch → `reconcile_plan` path (the `#378` `shadowed_junk_entries` precedent), so a genuinely large legitimate document reports its truncation to the user. Not mergeable until that lands — a set-but-unsurfaced flag would still be silent.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/docx.rs
+++ b/engine/crates/xerj-autoindex/src/extract/docx.rs
@@ -2,7 +2,7 @@
 //! Collects w:t runs, breaks paragraphs on w:p, records Heading-styled
 //! paragraphs as headings. One document record (sectioned at 32KB).
 
-use super::{emit_document, ExtractStats, Sink};
+use super::{emit_document, ExtractStats, Sink, MAX_RECORDS_PER_FILE};
 use anyhow::{Context, Result};
 use quick_xml::events::Event;
 use quick_xml::Reader;
@@ -138,7 +138,14 @@ fn extract_bounded(
                     .unwrap_or_else(|| "untitled".into())
             })
     });
-    emit_document(&title, &headings, body, sink, &mut stats);
+    emit_document(
+        &title,
+        &headings,
+        body,
+        MAX_RECORDS_PER_FILE,
+        sink,
+        &mut stats,
+    );
     Ok(stats)
 }
 
--- a/engine/crates/xerj-autoindex/src/extract/html.rs
+++ b/engine/crates/xerj-autoindex/src/extract/html.rs
@@ -5,7 +5,8 @@
 //! document record {title, headings, body}.
 
 use super::{
-    emit_document, sanitize_field_name, ExtractStats, FieldOrigin, RawRecord, Sink, MAX_WHOLE_FILE,
+    emit_document, sanitize_field_name, ExtractStats, FieldOrigin, RawRecord, Sink,
+    MAX_RECORDS_PER_FILE, MAX_WHOLE_FILE,
 };
 use anyhow::Result;
 use serde_json::{Map, Value};
@@ -92,7 +93,14 @@ pub fn extract(path: &Path, gzip: bool, sink: Sink) -> Result<ExtractStats> {
     } else {
         doc.title.trim().to_string()
     };
-    emit_document(&title, &doc.headings, doc.body.trim(), sink, &mut stats);
+    emit_document(
+        &title,
+        &doc.headings,
+        doc.body.trim(),
+        MAX_RECORDS_PER_FILE,
+        sink,
+        &mut stats,
+    );
     Ok(stats)
 }
 
--- a/engine/crates/xerj-autoindex/src/extract/mod.rs
+++ b/engine/crates/xerj-autoindex/src/extract/mod.rs
@@ -67,6 +67,10 @@ pub struct RawRecord {
 pub struct ExtractStats {
     pub records: u64,
     pub junk: u64,
+    /// Set when a per-file record cap (`MAX_RECORDS_PER_FILE`) stopped the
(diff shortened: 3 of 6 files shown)


xerj-org:
Progress: graph-path (default) surfacing landed (43011dd6) — a `truncated_files` counter + a post-run `pr.note`, verified end to end (an ~11 MB prose file caps at 4096 and prints the note). Full `xerj-autoindex` suite 713/0, clippy `-D warnings` + fmt clean (1.97.1). Remaining before un-draft: the `--no-graph` generated path (`sync_executor` `PreparedArtifact`) surfacing.

xerj-org:
Both paths now surface the truncation, so this is ready for review. Generated (`--no-graph`) path landed: `PreparedArtifact.truncated` (skip-serialized when false → non-truncated artifacts digest unchanged; snapshot digest tests green), counted from the snapshot after preparation, `pr.note` in `begin_non_graph_generation`. Verified end to end on both paths (an ~11 MB prose file caps at 4096 and prints the note under both graph and `--no-graph`). Full `xerj-autoindex` suite 713/0, clippy `-D warnings` + fmt clean (1.97.1).

xerj-org:

**CORRECTNESS — MERGE_OK**
- nit: Shipped tests cover the loop-path truncation (cap=3 over ~9 sections) and under-cap, but not the exact boundary: N==cap (must NOT set truncated) nor N==cap+1 (tail-guard drops exactly one section and sets truncated). Both are correct in practice — verified with a throwaway test at S=10, M in {9,10,11,12} — but a regression there would slip past the current suite. Consider adding a boundary case.

Rationale: Under the CORRECTNESS lens on the cap logic in emit_document, I could not construct a defect. The one-section lookahead plus the two guards (loop guard checked before emitting prev; tail guard) bound emission to exactly min(N, cap) records, and truncated is set iff N > cap. Verified empirically: N=cap emits all N with truncated=false; N=cap+1 emits cap with truncated=true via the tail-guard drop-exactly-one path; N>cap loop-stops at cap; a sampling sink-stop while also over the cap returns false with truncated=false (sampling precedence). No records leak past the cap, truncated is never set on a fully-emitted document, stats.records always equals emitted count. All 12 section tests pass under 1.97.1.

**REGRESSION — MERGE_OK**
- nit: The 4096-section cap means a legitimately large single prose document (>~8MB of text) loses its tail records. It is surfaced via pr.note (not silent) and matches the agreed #381 scope, so this is an accepted design consequence, not a defect.

Rationale: Could not refute. The sealed-snapshot serialization change is safe: PreparedArtifact.truncated uses `#[serde(default, skip_serializing_if = "std::ops::Not::not")]` and snapshot_digest serializes via serde_json, so a non-truncated artifact omits the field and hashes byte-identically — proven by the passing digest test, the legacy-resume byte-identity test, and the section-split parity test. The cap is inert under the cap; sampling early-stop and cap-stop are correctly distinguished; no double-reporting (both --no-graph call sites return before the graph worker loop, and truncated_files only runs on the graph path). Full xerj-autoindex suite green (713 lib + integration, 0 failures); clippy --all-targets -D warnings clean; fmt clean.

**EVIDENCE-HOLDS — MERGE_OK**
- should-fix: The truncation note is run-level only: it reports a count of files but names neither the truncated file nor how many records were dropped. On a multi-file corpus a user cannot tell which document lost its tail, weakening the "#381 never-silent" guarantee. Non-blocking (records are still bounded and the drop is surfaced). [-> follow-up #759]
- nit: PreparedArtifact.truncated is skip-serialized-when-false: an artifact sealed by a pre-PR binary and reused by a resumed --no-graph generation deserializes truncated=false, so a truncation sealed before the upgrade is not reported. Fresh runs unaffected; cross-version resume only.
- nit: The cap guards only emit_document (prose/html/docx/as_document). Other high-fanout extractors (code/logs/csv/jsonl/sqldump) remain unbounded by MAX_RECORDS_PER_FILE. Outside #381's prose-misclassification scope.

Rationale: Both load-bearing claims hold. (a) Fail-before: removing the loop cap hunk makes emit_document_caps_records_and_reports_truncation fail with 9 records vs the expected 3. (b) Cap enforced + pr.note emitted on BOTH paths, verified end-to-end: a slim xerj on :9720 (explicit --url, never :9200/XERJ_URL) indexing an 11.5MB single prose file produced the note on the graph path (4096 records) AND the --no-graph generated path (4096 records), each exactly once; a 3.6KB file produced no note. No double-count. 12/12 section_tests, 713/713 autoindex lib tests pass under 1.97.1.

---
**Gate: 0 BLOCKING findings across all three lenses.** Merge will proceed once the branch (rebased onto the #757-fixed main) re-runs CI to green+CLEAN. should-fix + nits filed as follow-up.

PR #761, opened 2026-08-25
test(autoindex): guard the #755 catalog-error enrichment wiring (#760)

### LINKED ISSUE #760: autoindex: add an integration test guarding the #755 catalog-error enrichment wiring
Follow-up from the #756 (#755) skeptic pass (evidence lens, should-fix).

The 3 unit tests call `explain_legacy_catalog_mapping_conflict` directly, so they pass even if the `.map_err(...)` wiring at the catalog `update_mapping` in `ensure_generation_mappings` is removed — a future refactor could silently drop the enrichment and CI would stay green while the opaque `mapper_parsing_exception` returns.

Add a test that exercises `ensure_generation_mappings` against a catalog whose `prefix`/`doc_kind` is already `text` and asserts the surfaced error names the field + the reindex recovery. The existing mock `HttpEndpoint` (incremental_reconcile_http_tests) accepts any mapping so it will not reproduce the conflict — the test needs either a mock that returns the 400 `field [X] already exists as [text]` for the catalog `_mapping` PUT, or a real-server harness (as the skeptic did manually). Non-blocking for #756.

xerj-org:
Closes #760 (should-fix from the #756 skeptic pass).

The three `explain_legacy_catalog_mapping_conflict` unit tests call the helper directly, so they pass even if the `.map_err` wiring at the catalog `update_mapping` (ensure_generation_mappings) is dropped — a future refactor could silently restore the opaque `mapper_parsing_exception` while CI stays green.

This adds an end-to-end guard: an opt-in `catalog_mapping_conflict` flag on the HTTP e2e mock returns the exact `field [prefix] already exists as [text]` 400 for the catalog `_mapping` PUT, and asserts `run_index` surfaces the enriched migration error (names the field + reindex recovery + preserves the raw reason), not the opaque one. Off by default, so every other HTTP e2e test is unchanged.

Verify:
- fail-before: with the `.map_err` line removed, the test fails (opaque error, no "older build"/"Reindex").
- fail-after: passes. Full `xerj-autoindex` suite 715/0; clippy `--all-targets -D warnings` clean; fmt clean (1.97.1).

### DIFF
--- a/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/incremental_reconcile_http_tests.rs
@@ -32,6 +32,12 @@ struct HttpState {
     embedding_identity_sha256: String,
     saw_dataset_mapping_update: bool,
     saw_catalog_mapping_update: bool,
+    /// Opt-in (#755/#760): make the catalog `_mapping` PUT fail with the exact
+    /// `field [X] already exists as [text]` conflict an older-build catalog
+    /// produces, so a test can prove the run surfaces the ENRICHED error rather
+    /// than the opaque mapper_parsing_exception. Off by default — every other
+    /// test's faithful-accept behaviour is unchanged.
+    catalog_mapping_conflict: bool,
 }
 
 struct HttpEndpoint {
@@ -179,11 +185,13 @@ fn handle_http(mut stream: TcpStream, state: &Arc<Mutex<HttpState>>) {
     } else if method == "PUT" {
         let value: Value = serde_json::from_slice(&body).unwrap_or(Value::Null);
         let mut locked = state.lock().unwrap();
+        let mut inject_catalog_conflict = false;
         if path.ends_with("/_mapping") {
             assert!(value.get("properties").is_some());
             assert!(value.get("mappings").is_none());
             if path.starts_with(&format!("/{}/", catalog::CATALOG_INDEX)) {
                 locked.saw_catalog_mapping_update = true;
+                inject_catalog_conflict = locked.catalog_mapping_conflict;
             } else {
                 locked.saw_dataset_mapping_update = true;
             }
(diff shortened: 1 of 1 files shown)


xerj-org:

**CORRECTNESS — MERGE_OK**
- nit: Assertion 3 (`already exists as [text]`) is non-discriminating on its own — it holds even without the wiring (raw reason is preserved in both paths). The guard is carried entirely by assertions 1 (older build + Reindex) and 2 (backtick-wrapped `prefix`), which fail-before demonstrably fails.
- nit: The injected 400 body is the canonical ES/helper-matched conflict string rather than a byte-for-byte capture of a live xerj response — sufficient to exercise the wiring; live-phrasing faithfulness is owned by #755/#756.

Rationale: Removing ONLY the `.map_err(|e| explain_legacy_catalog_mapping_conflict(...))` line makes the new test FAIL with the opaque mapper_parsing_exception (no older-build/Reindex/`prefix`); restoring it passes. The failure surfaces from the `install generation catalog mapping` context, proving the run reaches the catalog `_mapping` PUT and the enriched path is truly exercised (not a false pass from an earlier abort). Full HTTP e2e module + 3 helper unit tests + fmt + clippy `--all-targets -D warnings` all pass under 1.97.1; single-file branch.

**REGRESSION — MERGE_OK**
- nit: Pre-existing (not this PR): the explain helper's `legacy_text_field_in_conflict` uses raw byte-slicing; a multibyte-UTF-8 field name could panic on a non-char-boundary. Out of scope for this test-only PR and not exercisable via the ASCII-`prefix` mock. Noted for transparency.

Rationale: The new `catalog_mapping_conflict` flag defaults false, so the 400 branch is unreachable for every other e2e test (byte-unchanged 200 path). The 400->"Bad Request" phrase change is byte-identical for 200/500. Full suite green under 1.97.1: 715 lib + 14 integration + 0 doctests, 0 failed; clippy clean; fmt exit 0. Fail-before confirmed the test is a genuine wiring guard.

**EVIDENCE-HOLDS — MERGE_OK** (no findings)

Rationale: (a) the test passes green under 1.97.1; (b) the 400 is injected only for PUT `/autoindex-catalog/_mapping` with the opt-in flag — dataset _mapping and every other path return 200; (c) deterministic (ephemeral 127.0.0.1:0, holds HTTP_E2E_LOCK, synchronous run_index, no timing asserts). `git diff origin/main...HEAD` touches ONLY the test file (+61/-5), zero production code. Fail-before: removing the `.map_err` makes the test fail with the opaque error, restored clean — the pass is non-vacuous.

---
**Gate: 0 BLOCKING findings across all three lenses.** Merge proceeds once CI is green+CLEAN.

PR #762, opened 2026-08-25
fix(autoindex): name the truncated files in the #381 cap note (#759)

### LINKED ISSUE #759: autoindex: #381 truncation note is run-level (count only) — name the file + dropped records
Follow-up from the #758 (#381) skeptic pass (evidence lens, should-fix).

The per-file record cap surfaces truncation as a run-level `pr.note`: "N file(s) hit the per-file record cap (4096) and were truncated". It reports a count but names neither the truncated file(s) nor how many records were dropped. On a multi-file corpus a user cannot tell which document lost its tail, which weakens the "#381 never-silent" guarantee for a legitimately-large prose document.

Fix direction: carry the truncated file rel (and dropped-record count) rather than a bare counter, and list them — mirror the `shadowed_junk_entries` per-file reporting (#378) that `junk` already uses. The graph path aggregates via `truncated_files` (AtomicU64, lib.rs) and the generated path via `PreparedArtifact.truncated` (sync_executor.rs); both would carry the rel instead of just incrementing.

Non-blocking for #758 (records are bounded and the drop is surfaced). Two nits from the same pass, worth noting here: (1) a truncation sealed by a pre-PR binary and reused on a resumed `--no-graph` generation deserializes `truncated=false` (skip-serialized-when-false), so it is not reported on cross-version resume; (2) the cap guards only `emit_document` (prose/html/docx), not the other high-fanout extractors (code/logs/csv/jsonl/sqldump) — outside #381 scope but the "per-file record cap" framing reads broader than it is.

xerj-org:
Closes #759 (should-fix from the #758/#381 skeptic pass).

The #381 truncation note reported only a count, so on a multi-file corpus a user could not tell which document lost its tail. This collects the truncated file rels on both paths — the graph worker (`truncated_files` becomes a `Vec` of rels) and the `--no-graph` snapshot (mapped `content_id -> rel` via `plan.files`) — and lists them (capped at 10, then "and N more") via a shared, unit-tested `truncation_note_message` helper.

Also folds in the boundary-coverage nit from the #758 correctness review: a test driven by the body's real section count pins the cap at N==cap (not truncated) and N==cap+1 (truncated, tail dropped), so an off-by-one in either guard is caught.

Verify:
- e2e: both graph and `--no-graph` now print `...truncated: <file> — split them...`.
- `cargo test -p xerj-autoindex` 720/0; clippy `--all-targets -D warnings` clean; fmt clean (1.97.1). Unit tests for the message formatting (empty -> None; names files; caps + "and N more").

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/mod.rs
+++ b/engine/crates/xerj-autoindex/src/extract/mod.rs
@@ -905,4 +905,44 @@ mod section_tests {
             "sanity: multi-section but under the cap, got {n}"
         );
     }
+
+    /// #381 boundary (from the #758 correctness review): a body of EXACTLY the
+    /// cap emits every section and is NOT flagged truncated; one section over
+    /// emits exactly the cap and IS truncated, dropping the tail. Driven by the
+    /// body's real section count so it pins the off-by-one directly.
+    #[test]
+    fn emit_document_boundary_at_and_one_over_the_cap() {
+        let body = doc(60, 300);
+        let total = split_sections(&body).len();
+        assert!(total >= 3, "need a multi-section body, got {total}");
+
+        // N == cap: all emitted, not truncated.
+        let mut at = 0usize;
+        let mut stats = ExtractStats::default();
+        let mut sink = |_r: RawRecord| {
+            at += 1;
+            true
+        };
+        assert!(emit_document("t", &[], &body, total, &mut sink, &mut stats));
+        assert_eq!(at, total, "N==cap must emit every section");
+        assert!(!stats.truncated, "N==cap must NOT set truncated");
+
+        // N == cap + 1 (cap = total - 1): exactly cap emitted, tail dropped.
+        let mut over = 0usize;
+        let mut stats2 = ExtractStats::default();
+        let mut sink2 = |_r: RawRecord| {
+            over += 1;
+            true
+        };
+        assert!(emit_document(
+            "t",
+            &[],
+            &body,
+            total - 1,
+            &mut sink2,
+            &mut stats2
+        ));
+        assert_eq!(over, total - 1, "N>cap must emit exactly the cap");
+        assert!(stats2.truncated, "N>cap must set truncated");
+    }
 }
--- a/engine/crates/xerj-autoindex/src/lib.rs
+++ b/engine/crates/xerj-autoindex/src/lib.rs
@@ -635,19 +635,13 @@ fn begin_non_graph_generation(
     // #381: the per-file record cap dropped a file's tail during preparation.
     // The generated path seals before the graph worker loop runs, so report it
     // here rather than at the shared post-run summary.
-    let truncated = snapshot
+    let truncated: Vec<String> = snapshot
         .files
         .iter()
-        .filter_map(|f| f.prepared.as_ref())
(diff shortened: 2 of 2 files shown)


xerj-org:

**CORRECTNESS — MERGE_OK**
- nit: The snapshot's key field is named `content_id` but is actually the plan `file_key`; `plan.files.get(&f.content_id)` reads correctly, but the naming invites the exact "wrong key" worry. Pre-existing naming across sync_executor/state, not introduced here.

Rationale: Graph path — `truncated_files` (Mutex<Vec<String>>) is read via into_inner().unwrap() AFTER std::thread::scope joins all workers — no lost pushes; unwrap mirrors the existing bulk_errors/extra_junk idiom. --no-graph — SnapshotFile.content_id == the plan.files key (create_snapshot_inner sets prepared=Some ONLY when plan.files.contains_key(content_id)), so filter_map cannot silently drop a truncated file. No double-report (both begin_non_graph_generation call sites return before the graph loop). truncation_note_message: empty->None, take(10)+saturating_sub correct. Boundary test correct. 720 lib tests pass, clippy --all-targets clean, fmt clean.

**REGRESSION — MERGE_OK**
- nit: Non-graph filter_map plan.files.get(&f.content_id) — a truncated file absent from plan.files would be silently unnamed. Cannot occur (only planned files are prepared/truncated), and at worst degrades to the old count-only behaviour, never worse.
- nit: lock().unwrap()/into_inner().unwrap() can panic on a poisoned mutex, but mirrors the pre-existing bulk_errors/extra_junk patterns; guarded region is only a String clone + push, so poisoning is practically unreachable. Not a new risk.

Rationale: All gates pass (1.97.1): fmt clean, clippy --all-targets -D warnings clean, cargo test 720 lib + all integration, 0 failed (incl. 4 new tests). Non-truncated case -> empty Vec -> None -> no note. AtomicU64->Mutex<Vec> introduces no new panic class. Note fires on both paths, mutually exclusive (no duplicate). Boundary test pins the #381 off-by-one via the body's real section count. Could not refute.

**EVIDENCE-HOLDS — MERGE_OK**
- nit: `truncation_note_message` renders "file(s)"/"were truncated" (plural) even for a single file; cosmetic only.
- nit: The note surfaces at a different point per path (before the scan bar in --no-graph, at finalize-catalog in graph) because the generated path seals before the worker loop. Both surface correctly exactly once; ordering difference is cosmetic.

Rationale: Helper tests 3/3 + section_tests 5/5 (incl. the boundary test) pass under 1.97.1. Booted own xerj server on port 9753, fresh data-dir (never :9200, never XERJ_URL), ran autoindex --url on an 11,534,566-byte single prose .txt (~5600 sections, capped at 4096). GRAPH path emitted `1 file(s) hit the per-file record cap (4096) and were truncated: bigprose.txt — split them or raise the cap if the dropped tail matters (#381)`; --no-graph emitted the identical note naming bigprose.txt. A small file produced NO note on both paths. Exactly one note per run, no double-noting. Release binary compiled clean under 1.97.1.

---
**Gate: 0 BLOCKING findings across all three lenses.** Merge proceeds once #763 (landing rc.69) lands, this rebases, and CI is green+CLEAN.

PR #764, opened 2026-08-25
feat(cli): xerj search — binary as a query client; reference-coding demo drops the scripts

xerj-org:
Adds `xerj search "<text>"`, a client subcommand so the reference-coding loop is the binary alone — no wrapper scripts.

**Feature.** `xerj search` detects the running node (health-ping, $XERJ_URL / localhost:9200), runs a definition-biased `multi_match` (`defs^3`, body, title), and prints each hit as `ax_path:line` + a passage. If no node answers, it prints the two commands to start one instead of a raw error. Flags: `-k / --full / --prefix / --json / --api-key`. Reuses the existing `esclient` (blocking reqwest, auth, retries) and the same `main.rs` dispatch as `autoindex`.

**Demo.** The landing reference-coding block now reads `xerj autoindex ref/sled → xerj search "…"` — one binary, no `tools/xerj-code/scripts/`. The `xc-*` corpus helpers stay documented as the team-reproducibility path.

Verified locally: builds clean; no-node path prints guidance; a live query returns `file:line` passages. SEO gate + head-check pass. Merging auto-deploys the demo via the Pages workflow.


xerj-org:
Pushed a rustfmt-only commit to fix the Format + Clippy failure (`pub mod search;` ordering + unformatted blocks in `search.rs`; no logic change). fmt `--check` now passes unpiped and workspace `clippy --all-targets -D warnings` is clean locally under 1.97.1.

Leaving the feature review/merge itself to the maintainer — this is a new CLI subcommand + landing change (product direction), not a bug fix, so I only unblocked its CI rather than auto-merging.

xerj-org:
Bug (reviewed, reproduced): `xerj search` renders the wrong `_source` fields for code-symbol docs, so its headline output breaks on the reference-coding case it targets.

A plain `xerj autoindex` of a `.rs` file writes symbol docs with these keys: `ax_path, code, kind, language, line, name, title` (verified on a fresh non-9200 node). But `search.rs` uses:
- `_source: ["ax_path","start_line","language","body","title"]` and renders `src.get("start_line")` and `src.get("body")`.

There is no `start_line` (its `line`) and no `body` on a code-symbol doc (its `code`). Reproduced end-to-end (own server, `--url`, never :9200) — `xerj search "DEFAULT_MAX_CONN"`:
```
─── hnsw.rs  (score 1.27)      <- whole-file doc: shows content, but NO :line
    pub const DEFAULT_MAX_CONN: usize = 16;
    ...
─── hnsw.rs  (score 1.20)      <- the #500 fine-grained SYMBOL doc: empty passage, no :line
```
So every code hit shows `ax_path` with no `:line`, and the fine-grained symbol docs (#500) render blank — which defeats the "Cite ax_path:line" tagline and the reference-coding value.

Fix direction: request/render `line` (fallback `start_line`) and `code` (fallback `body`) — e.g. `_source: ["ax_path","line","language","code","body","title"]`, `loc` from `line`, passage from `code` then `body`. Also consider adding `code`/`name` to the `multi_match` fields (a symbol doc has no `defs`/`body`, so today it only matches on `title`). No line/logic change needed beyond the field names. Leaving the fix + merge to you — flagging so it does not ship with broken code-symbol output.

xerj-org:
Pushed the fix for the field-mismatch bug I flagged above (search.rs only): render `line`/`code` (fallback `start_line`/`body`) and add `name^4`/`code^2` to the multi_match. Verified end to end — `xerj search DEFAULT_MAX_CONN` now prints `hnsw.rs:1` + the declaration, top-ranked, instead of a blank passage with no line. fmt/clippy/full xerj-autoindex suite green (1.97.1). Left the feature review + merge to you.

PR #766, opened 2026-08-25
fix(autoindex): #765 warn when XERJ_URL is set but --url was not passed

### LINKED ISSUE #765: autoindex: XERJ_URL is silently ignored (endpoint is --url only, defaults to :9200) — warn on the mismatch
Footgun (surfaced repeatedly, incl. a multi-session debugging detour, and again while reviewing #764).

`xerj autoindex` resolves its endpoint from the `--url` flag ONLY (cli.rs:434), defaulting to `http://localhost:9200` (cli.rs:393). It never reads `XERJ_URL` for the endpoint (only `XERJ_API_KEY`, cli.rs:394), and emits no warning. Meanwhile `xerj search` (#764) DOES read `XERJ_URL`. So a user who sets `XERJ_URL` (habit from `search` or ES tooling) and runs `xerj autoindex <dir>` without `--url` silently indexes into `localhost:9200` — the wrong node, and specifically the one that must never be written to in this repo. The failure is invisible: no error, wrong target.

Requiring an explicit `--url` for the (destructive) index op may be a deliberate safety choice, so the fix is NOT to start reading `XERJ_URL` for the endpoint (that would re-introduce accidental `:9200` writes via a stale env var). The decision-free fix is a **warning**: when `XERJ_URL` is set AND `--url` was not passed, print e.g. `note: XERJ_URL is set but autoindex reads the endpoint from --url only; using http://localhost:9200. Pass --url "$XERJ_URL" to target it.` Thread a `url_explicit` bool through parse (mirroring `bulk_timeout_explicit`/`max_minutes_explicit`) and emit the note for the index subcommand before connecting.

Impact: prevents silent mis-targeting (data indexed into the wrong node); aligns the two subcommands' URL story. Low severity, high annoyance, decision-free.

xerj-org:
Closes #765.

`xerj autoindex` resolves its endpoint from `--url` only, defaulting to `http://localhost:9200`. It ignores `XERJ_URL` (which `xerj search` honors) on purpose, so a stale env var can't silently redirect a write. But silent is the trap: setting `XERJ_URL` and forgetting `--url` quietly targeted the loopback default with no word to the operator (this cost a multi-session mis-target debug).

Fix keeps the safety (still `--url` only, env never selects the endpoint) and removes the silence: track whether `--url` was actually passed, and when `XERJ_URL` is set and it was not, surface a startup note that echoes the ignored value and says how to act on it (`Pass --url <value>`).

Covers all three subcommands: `index` rides the existing `resource_notes` channel (prints at run start, also in `--json`); `map` and `status` carry the note on their cfg and print it before they connect (both read the same endpoint, so a stale var shows the wrong node's map/progress too). The note is computed once in `parse()`.

Logic is a pure helper (`xerj_url_ignored_note`), unit-tested across all branches (set+no-url -> note names value/--url/ignored; explicit --url -> none; unset/blank -> none).

Not E2E-run on purpose: the note only fires when `--url` is absent, which by construction targets `:9200`, so exercising it would hit the reference node. Verified via the unit test plus the pre-existing `resource_notes` -> `pr.note` print path.

fmt/clippy(--all-targets -D warnings)/build all green under 1.97.1; full xerj-autoindex suite 722 lib + integration all pass (both dev and ci-test).

### DIFF
--- a/engine/crates/xerj-autoindex/src/cli.rs
+++ b/engine/crates/xerj-autoindex/src/cli.rs
@@ -385,12 +385,38 @@ pub fn help_text_with(feedback: bool) -> String {
     )
 }
 
+/// A startup note for the case where `XERJ_URL` is set but the run did not pass
+/// `--url`. `autoindex` resolves its endpoint from `--url` only (default
+/// `http://localhost:9200`); `XERJ_URL` is honored by `xerj search` but not
+/// here, on purpose, so a stale variable can't silently redirect a write. Silent
+/// is the trap though, so name the mismatch and how to act on it. `None` when
+/// `--url` was passed, or `XERJ_URL` is unset or empty.
+pub(crate) fn xerj_url_ignored_note(xerj_url: Option<&str>, url_explicit: bool) -> Option<String> {
+    if url_explicit {
+        return None;
+    }
+    let value = xerj_url?.trim();
+    if value.is_empty() {
+        return None;
+    }
+    Some(format!(
+        "XERJ_URL={value} is set but ignored here: autoindex takes its endpoint from --url only \
+         (default http://localhost:9200). Pass --url {value} to target it."
+    ))
+}
+
 pub fn parse(args: Vec<String>) -> Result<Cmd, String> {
     let mut it = args.into_iter().peekable();
     let mut folder: Option<PathBuf> = None;
     let mut sub: Option<String> = None;
 
     let mut url = "http://localhost:9200".to_string();
+    // `xerj autoindex` takes its endpoint from `--url` ONLY, unlike `xerj
+    // search` which honors `XERJ_URL`. Ignoring the env var here is deliberate:
+    // a write must not be redirected by a stale variable. But silent is a trap
+    // (it cost a multi-session mis-target debug), so we track whether --url was
+    // actually passed and warn when XERJ_URL is set and it was not.
+    let mut url_explicit = false;
     let mut api_key = std::env::var("XERJ_API_KEY").ok().filter(|s| !s.is_empty());
     // Set only when the key was discovered on disk, so output can reference the
     // file instead of echoing the secret.
@@ -431,7 +457,10 @@ pub fn parse(args: Vec<String>) -> Result<Cmd, String> {
 
     while let Some(arg) = it.next() {
         match arg.as_str() {
-            "--url" => url = it.next().ok_or("--url needs a value")?,
+            "--url" => {
+                url = it.next().ok_or("--url needs a value")?;
+                url_explicit = true;
+            }
             "--api-key" => api_key = it.next(),
             "--workers" => {
                 // Refused, not clamped: `--workers 0` used to become 1 and
@@ -775,6 +804,12 @@ pub fn parse(args: Vec<String>) -> Result<Cmd, String> {
                 );
             }
             let plan = crate::resources::plan(workers, pdf_workers, bulk_mb);
+            let mut resource_notes = plan.notes;
+            if let Some(note) =
+                xerj_url_ignored_note(std::env::var("XERJ_URL").ok().as_deref(), url_explicit)
+            {
+                resource_notes.push(note);
+            }
             Ok(Cmd::Index(Box::new(IndexCfg {
                 root,
                 url,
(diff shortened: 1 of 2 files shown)


xerj-org:

**Result: 3/3 MERGE_OK, ZERO BLOCKING findings.** The one `should-fix` is a merge-gate reminder (hold for green CI), not a code defect. Nits recorded below; merge held until all checks green + CLEAN.

- **nit**: Index-path note is delivered through pr.note (the progress surface, suppressible under --quiet), while map/status use unconditional eprintln. Minor delivery inconsistency, but pre-existing on the index path and not introduced/regressed by this PR; map/status have no --quiet. Not blocking.
- **nit**: If XERJ_URL is set to exactly the default endpoint (http://localhost:9200) with no --url, the note still fires ('set but ignored... Pass --url http://localhost:9200'). Technically noisy since the effective and ignored endpoints coincide, but the message stays accurate and harmless. Not blocking.
- _rationale_: The fix is present and correct. The pure helper xerj_url_ignored_note (cli.rs:401) returns Some exactly when url_explicit is false AND XERJ_URL is set/non-empty after trim, and None for explicit --url, unset, or blank/whitespace — verified by reading the code and by the four-branch unit test, which passes. url_explicit is set true in the --url arm (cli.rs:469) for every subcommand. The note is hoisted once (cli.rs:762) and moved into the three mutually-exclusive match arms (map 798, status 808, index 823) with no double-move or use-after-move — confirmed by a clean compile. Map (lib.rs:6482) and status (lib.rs:6746) eprintln the note BEFORE constructing Es/ping and before any endpoint I/O, so the operator sees it even if the wrong node is unreachable; index prints via resource_notes at lib.rs:3556 before connecting. eprintln to stderr does not corrupt map's JSON stdout. Only two constructors exist (both in parse). I actively tried to construct false-positive and false-negative inputs and found only extreme/benign edge cases (non-UTF8 env value goes silent; XERJ_URL equal to the default value yields a harmless note) — neither is a correctness defect. No blocking issue found.

- **nit**: The XERJ_URL env read is now hoisted to cli.rs:763 and executes even on map/status error arms (max-minutes/approve/progress/ignore/bulk-timeout guards) that return before using it. Harmless — a side-effect-free syscall whose result is dropped — but slightly wasteful vs. computing it only on the success arms. Not blocking.
- _rationale_: Full xerj-autoindex suite under rustc 1.97.1 ci-test passes with 0 failures (722 lib + 14 integration tests); fmt --all --check exits 0; clippy --all-targets -D warnings exits 0. MapCfg/StatusCfg derive only Debug+Clone (no Serialize/Deserialize/Hash), so adding xerj_url_note cannot change any stored digest or serialization. Both structs have exactly one constructor each (cli.rs:792/803), both updated; no other construction site and no exact-field destructuring exists, so no constructor breaks. The hoisted env read sits after the arg loop, while --help returns earlier (cli.rs:643), so --help behavior is unchanged; error arms merely drop the read. New startup notes go to stderr via eprintln!, while map/status output including --json goes to stdout, so --json stdout is not polluted. XERJ_URL is never used to resolve the endpoint. fail-before is irrelevant because map/status previously emitted no note (genuinely new behavior) and no pre-existing behavior changed.

- **should-fix**: Merge gate only: PR CI is UNSTABLE because Build+Test, ES-compat YAML, Release build, Use-case harnesses, and macos/windows FD smoke are still pending (none failed). Per the RC merge gate, hold merge until all checks are green, not on MERGEABLE/UNSTABLE alone. Not a code defect — independently ran the affected crate green (fmt 0, clippy clean, 722 lib tests + integration pass on 1.97.1).
- **nit**: index --quiet (non-json) drops the note: pr.note early-returns on Surface::Silent (progress.rs:538), so under --quiet without --json the XERJ_URL note is not surfaced. By-design (--quiet suppresses all notes incl. existing memory-planning ones), not a regression; the --json path (lib.rs:6236) and map/status (raw eprintln) still carry it. Minor index-vs-map/status inconsistency.
- _rationale_: All three required symbols exist (cli.rs:401 xerj_url_ignored_note; xerj_url_note field cli.rs:86/97; consumed lib.rs:6481/6745). fmt exit 0; clippy --all-targets -D warnings clean; full xerj-autoindex ci-test suite 722 lib pass / 0 fail / 0 ignored plus all integration binaries pass; the new test passed and is not #[ignore]d. Every PR-body claim is supported by the tree: grep XERJ_URL in .github/ and scripts/ is empty (note cannot fire in CI); the two map harness callers (run-eval.sh:122, autoindex_semantic.sh:71) pass --url so output is unperturbed; the Cmd enum is exactly Index/Map/Status/Help; the helper is pure and unit-tested across all four branches; the E2E-skip reasoning is sound. No blocking defect found after adversarial review.

PR #767, opened 2026-08-25
ci(#751): bound Build+Test hangs (step-6 + job-level timeout) so main isn't stuck for hours

xerj-org:
Partial fix for #751 (mitigation, not the root fix — issue stays open).

`Build + Test` intermittently hangs on the 2-4 vCPU hosted runner (port/fixture race at low thread count). Because main uses `cancel-in-progress: false` by design, a hung job sits against GitHub's 6h default and leaves main ~4h with no completed CI verdict — the damaging second-order effect the issue calls out.

Two bounds, neither of which fixes the root race:

1. `timeout-minutes: 12` on step 6 (`Workspace tests at default parallelism`) — precise, fast bound for the one documented step-6 hang (normal ~2.3 min, seen 18-28 min).
2. `timeout-minutes: 50` on the whole `build-test` job — defense-in-depth. Bounds ANY hang in the job (setup, cache, a future test race the step timeout would not cover) to 50 min instead of the 6h default. Kept generous on purpose: step 5 (build + threads=2 tests) can legitimately take ~36 min on a loaded runner — a slow compile, not a hang (measured on a recent run: step 5 built ~36 min, then step 6 ran and passed) — so a tighter cap would false-fail slow-but-healthy runs.

Root race still needs the bisect + port/fixture serialisation from the issue, so #751 stays open. YAML validated (both timeouts attach to the right scope; run commands unchanged).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -263,6 +263,19 @@ jobs:
   build-test:
     name: Build + Test
     runs-on: ubuntu-latest
+    # #751's second-order damage is the whole job hanging against GitHub's 6h
+    # default while main's `cancel-in-progress: false` never auto-cancels — main
+    # then sits ~4h with no CI verdict. The step-6 `timeout-minutes: 12` below
+    # bounds the one documented hang (step 6, default parallelism) precisely and
+    # fast. This job-level cap is defense-in-depth: it bounds ANY hang in the job
+    # (setup, cache, a future test race the step timeout would not cover) to
+    # 50 min instead of the 6h default. It is deliberately generous because
+    # step 5 (build + threads=2 tests) can legitimately take ~36 min on a loaded
+    # 2-4 vCPU runner — a slow compile, not a hang (measured: step 5 built for
+    # ~36 min then step 6 ran and passed) — so a tighter cap would false-fail
+    # slow-but-healthy runs. Neither cap fixes the root race (#751 stays open for
+    # the bisect + port/fixture serialisation).
+    timeout-minutes: 50
     steps:
       - uses: actions/checkout@11d5960a326750d5838078e36cf38b85af677262 # v4.4.0
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -332,7 +332,16 @@ jobs:
       #
       # Reuses the binaries the step above already built: the added cost is test
       # runtime, not another compile.
+      # #751: this step intermittently hangs on the 2-4 vCPU hosted runner (a
+      # port/fixture race at low thread count — normal ~2.3 min, observed hanging
+      # 18-28 min). Because main's `cancel-in-progress: false` never auto-cancels,
+      # a hang leaves the whole run stuck for hours with no completed signal, so
+      # main commits never get a CI verdict. Bound it: a hang now fails the step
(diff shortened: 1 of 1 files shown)


xerj-org:

**Result: 3/3 MERGE_OK, ZERO BLOCKING findings.** Nits only (documentation + margin/scope), recorded below; none blocks.

- **nit**: Theoretical worst-case overlap: step 3 at its ~36-min legit-max compile + step 4 near its 12-min step cap + setup could exceed the 50-min job cap. Not a real healthy-run false-fail: a step near 12 min is itself a near-hang (5x norm). Normal case (36 + 2.3 + setup ~= 43 min) sits ~7 min under 50.
- **nit**: Scope: only build-test gets a job cap; a hang in another main job would still hold the cancel-in-progress:false slot up to 6h. Out of #751's scope, not a regression, but the starvation vector isn't fully closed.
- _rationale_: Both bounds placed and valued correctly. Job-level timeout-minutes:50 is a sibling of runs-on/steps on build-test, not under a step. Step-level 12 is on the correct step 6 (default-parallelism, ~2.3 min normal) NOT the threads=2 build step. 12 min = ~5x normal, well below the 18-28 min hangs; 50 min covers worst legit (~43) with margin. YAML parses. Diff +22/-0, only two non-comment lines; all run: commands and every other step/job unchanged. Mitigation holds against cancel-in-progress:false — caps free the slot within 50 min instead of 6h.

- **nit**: job-level timeout-minutes: 50 has modest headroom over worst-case legit (~43-46 min), so a heavily-loaded runner could occasionally false-fail. Deliberately generous defense-in-depth; recoverable via rerun. Not a defect.
- _rationale_: Change touches ONLY ci.yml (no Rust/source/test), purely additive (no removed/reordered lines). Full YAML parses; all 15 jobs byte-for-byte identical between origin/main and HEAD. Timeouts at correct scopes: job-level 50 on build-test, step-level 12 on the default-parallelism step. The preceding ~36-min threads=2 step is UNCAPPED, so the 12-min cap cannot false-fail the slow compile. 12 < 50: a step-6 hang trips the step cap first. A job timeout fails the job and releases the ci-${github.ref} slot — exactly the mechanism #751 needs given cancel-in-progress:false.

- **nit**: the comment's "step 5"/"step 6" count from GitHub's run-UI numbering (Set up job = step 1), off-by-one vs the literal defined-step list. Internally consistent and matches how a maintainer reads the run log; harmless.
- **nit**: 50-min job cap is somewhat tight against a compounded tail; realistic healthy worst case ~42 min, step-level cap fires first in that regime. Consider 55-60 min for extra margin; advisory only.
- _rationale_: Every PR claim checks out. Both timeouts present/correctly placed (job 50 at ci.yml:278, step 12 at ci.yml:357 on the default-parallelism step). 'main uses cancel-in-progress: false' confirmed (ci.yml:19); 6h-default-starves-main reasoning sound. 'neither cap fixes the root race; #751 stays open' confirmed: diff is ci.yml only, no test/port/fixture code. Comment accuracy verified — correctly frames step 6 as the hang and step 5 as a slow compile (not a hang); no leftover misstatement survives the amend.

Nits filed as follow-up (margin value + non-build-test starvation scope); none blocks per the merge gate.

PR #769, opened 2026-08-25
fix(autoindex): #768 deliver the XERJ_URL wrong-node warning so --quiet can't silence it

### LINKED ISSUE #768: autoindex: the XERJ_URL-ignored safety note is suppressed under --quiet on the index path (survives it on map/status)
Follow-up from the 3-skeptic pass on #766 (all non-blocking; filing per process).

#766 warns when XERJ_URL is set but --url was not passed (endpoint is --url only, so a stale var silently mis-targets, e.g. loopback :9200). The index path routes that note through `resource_notes` -> `pr.note`, which early-returns on `Surface::Silent` (progress.rs:538). So `xerj autoindex <dir> --quiet` (without --json) with XERJ_URL set and no --url surfaces NO warning — re-opening, under --quiet, the exact silent mis-write the fix exists to prevent. map and status carry the note via unconditional `eprintln` and are unaffected; the index --json path (lib.rs:6236) still includes it.

Low exposure in practice (a --quiet script that set XERJ_URL almost certainly also passed --url deliberately), which is why the skeptics scored it a nit, not a blocker. But the principle stands: a "you may be writing to the wrong node" warning is a safety signal, not routine progress chatter, and arguably should survive --quiet like map/status do.

Fix direction: give the index XERJ_URL note the same unconditional delivery as map/status (print it directly, not via the --quiet-suppressible pr.note surface), so all three subcommands behave the same. Leave the other resource/memory-planning notes on the suppressible surface. Small, localized to run_index note emission.

xerj-org:
Closes #768. Follow-up to #765/#766 (surfaced by that PR's skeptic pass).

#765 warns when `XERJ_URL` is set but `--url` was not passed (endpoint is `--url` only, so a stale var silently mis-targets, e.g. the loopback default). On the index path that note rode `resource_notes` -> `pr.note`, which `--quiet` silences (`Surface::Silent`) — so `xerj autoindex <dir> --quiet` with `XERJ_URL` set and no `--url` surfaced NOTHING, re-opening under `--quiet` the exact silent mis-write the warning prevents. map/status already `eprintln` unconditionally.

Fix — treat it as a safety signal delivered consistently:
- New `Progress::warn`: like `pr.note` but NOT silenced by `Surface::Silent` (a wrong-node warning must pierce `--quiet`), and on `--progress json` a well-formed `{"event":"warning",...}` object rather than a bare line (the surface is contractually one JSON object per line). The index path emits the note through `pr.warn`, before any endpoint I/O.
- Carried on `IndexCfg.xerj_url_note` (the field map/status already have) and mirrored into the `--json` result (`"xerj_url_note"`, null when it did not fire).

No `--json` compat break: the #765 note only landed on main today (post-rc.69), unreleased.

Tests: pure helper `xerj_url_ignored_note` unchanged (4-branch unit test); `index_suppresses_the_xerj_url_note_when_url_is_explicit` (explicit --url -> None, env-independent); `warn_pierces_quiet_and_stays_json_on_the_json_surface` (Silent -> plain line reaches stderr; Json -> well-formed event) — the safety-critical delivery is now tested without `:9200`.

Two defects the #769 skeptic pass caught and this PR now fixes: (1) a missed 5th `IndexCfg` constructor in `xerj-server/src/brain.rs` (E0063 workspace build break — `-p xerj-autoindex` never compiled it); (2) the bare `eprintln` polluting `--progress json` (now `pr.warn`). Verified workspace-wide under 1.97.1: fmt / `clippy --workspace --all-targets -D warnings` / `cargo check --workspace` clean; full xerj-autoindex suite 725 lib + integration pass (dev + ci-test). Not E2E-run: the note only fires with `--url` absent, which by construction targets `:9200`.

### DIFF
--- a/engine/crates/xerj-autoindex/src/cli.rs
+++ b/engine/crates/xerj-autoindex/src/cli.rs
@@ -32,6 +32,13 @@ pub struct IndexCfg {
     pub pdf_workers: usize,
     /// What the machine forced on this run, printed once at start-up.
     pub resource_notes: Vec<String>,
+    /// #768: the `XERJ_URL`-ignored safety note (set when `XERJ_URL` is present
+    /// but `--url` was not passed). Carried separately from `resource_notes`
+    /// because it is a safety warning, not routine progress chatter: it is
+    /// emitted with an unconditional `eprintln` (so `--quiet` does not silence a
+    /// "you may be writing to the wrong node" message) and mirrored into the
+    /// `--json` result, matching how `map`/`status` deliver the same note.
+    pub xerj_url_note: Option<String>,
     pub pdf_timeout_secs: u64,
     pub bulk_mb: usize,
     pub bulk_timeout_secs: u64,
@@ -819,10 +826,6 @@ pub fn parse(args: Vec<String>) -> Result<Cmd, String> {
                 );
             }
             let plan = crate::resources::plan(workers, pdf_workers, bulk_mb);
-            let mut resource_notes = plan.notes;
-            if let Some(note) = xerj_url_note {
-                resource_notes.push(note);
-            }
             Ok(Cmd::Index(Box::new(IndexCfg {
                 root,
                 url,
@@ -831,7 +834,11 @@ pub fn parse(args: Vec<String>) -> Result<Cmd, String> {
                 workers: plan.index_workers,
                 scan_workers: plan.scan_threads,
                 pdf_workers: plan.pdf_workers,
-                resource_notes,
+                resource_notes: plan.notes,
+                // #768: delivered by run_index via unconditional eprintln (not the
+                // --quiet-suppressible resource_notes surface) and mirrored into
+                // --json, so a wrong-node warning is never silently dropped.
+                xerj_url_note,
                 pdf_timeout_secs,
                 bulk_mb,
                 bulk_timeout_secs,
@@ -997,6 +1004,21 @@ mod tests {
         assert!(
             note(Some("   "), false).is_none(),
             "blank XERJ_URL: no note"
+        );
+    }
+
+    /// #768: the index cfg carries the XERJ_URL-ignored note on its own field
+    /// (delivered by an unconditional eprintln, so --quiet cannot silence a
+    /// wrong-node warning) rather than folding it into the --quiet-suppressible
+    /// resource_notes. An explicit --url means the env var was not ignored, so
+    /// there is nothing to warn about — deterministic regardless of the
+    /// environment the test runs in.
+    #[test]
+    fn index_suppresses_the_xerj_url_note_when_url_is_explicit() {
+        let cfg = index(&["data", "--url", "http://es.internal:9200"]);
+        assert!(
+            cfg.xerj_url_note.is_none(),
+            "an explicit --url means XERJ_URL was not ignored; no note"
         );
     }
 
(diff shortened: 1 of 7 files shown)


xerj-org:
## Definitive 3-skeptic pass (worktrees, rustc 1.97.1, head b6ef0fc3) — verbatim

**Result: 3/3 MERGE_OK, ZERO BLOCKING.** The sole recurring should-fix (stale "unconditional eprintln" comments) is now fixed in 892c9062. Remaining items are nits.

- **nit**: cli.rs doc/construction comments still describe the index-path delivery as an "unconditional eprintln", but the final form routes through pr.warn. Behavior correct; comment stale. *(fixed: 892c9062)*
- _rationale_: Progress::warn does NOT early-return under Surface::Silent (unlike note()) — writes a plain line to the sink, which Progress::new always builds as Sink::Stderr in production, so the warning genuinely pierces --quiet. Under Surface::Json it emits a single well-formed {"event":"warning",...} object; message passed through sanitize (line-injection boundary), so the one-object-per-line contract holds. Emitted exactly once after pr is built and before endpoint I/O (Es::with_bulk_timeout/ping). Old resource_notes.push removed (no double emission); mirrored into --json. Explicit --url deterministically None. map/status deliver via eprintln in their own handlers. fmt clean, check --workspace exit 0, 725 passed/0 failed (incl. warn_pierces), clippy --workspace --all-targets clean.

- **should-fix**: stale mechanism in comments (cli.rs field doc + construction) say "unconditional eprintln" but the index path routes through pr.warn (lib.rs). Behavioral guarantee still holds; comment-accuracy only. *(fixed: 892c9062)*
- **nit**: no end-to-end positive test that run_index_report emits under Surface::Silent; Progress::warn is unit-tested directly and the cli negative case is covered, but the IndexCfg->pr.warn wiring isn't asserted end to end.
- _rationale_: Fully green at b6ef0fc3. fmt 0, check --workspace clean (xerj-server compiles; prior E0063 fixed by brain.rs xerj_url_note: None), 725 lib passed/0 failed + integration green, clippy --workspace clean. All 5 IndexCfg literal constructors set the field. Debug+Clone only (no Serialize/Hash), so no digest change. New "xerj_url_note" --json key breaks nothing (no deny_unknown_fields consumer). Note removed from resource_notes; warn under Silent writes to real Stderr sink so it pierces --quiet while staying {"event":"warning"} on --progress json.

- **should-fix**: stale doc comments say the index note is an unconditional eprintln, but b6ef0fc3 routes it through pr.warn. Inaccurate documentation; behavior correct. *(fixed: 892c9062)*
- **nit**: map --json output does not mirror xerj_url_note; map/status deliver it only via stderr eprintln. Cosmetic gap, not a regression (note new/unreleased; #768 scope is the index --quiet silent-drop).
- _rationale_: Every claim confirmed. note() early-returns on Silent; warn() pierces Silent as a plain line and emits {event:warning} on Json with the same sanitize; index emits via pr.warn BEFORE endpoint I/O; note removed from resource_notes (emits once); --json mirror present on the live path; map/status eprintln unconditionally; both new tests exist, not ignored, assert the claims. #765/#766 in no rc tag (latest rc.69), so the new JSON key is not a compat break. All 6 constructor sites carry the field. CI all green under 1.97.1.

Cosmetic map/status --json mirror nit left as out-of-scope (index --quiet gap is what #768 fixes).

PR #772, opened 2026-08-25
test(engine): #421 bind the _id sort-value type invariant

### LINKED ISSUE #421: Follow-up to #401: three unresolved adversarial-verification items were merged to main in #420

xerj-org:
Addresses one of #421's "Also unresolved" hardening notes (does not close #421).

The issue flags that `_id`'s numeric-prefilter safety rests on a **type invariant, not a gate**: `compute_sort_values` emits the `_id` sort value as a `String` (index.rs, the `sf.is_doc_order() || sf.field == "_id"` branch), so `SortTopK::primary_f64_rejects`'s `as_f64()` is always `None` for an `_id` sort and the numeric pre-filter never mis-rejects an `_id`-sorted hit. "Give `_id` a numeric sort value and the defect returns." The invariant held but had **no binding test**.

This adds one: it calls `compute_sort_values` for `sort: [{"_id":"asc"}]` and asserts the value is a `Value::String` (and `as_f64()` is `None`). A change that makes the `_id` sort value numeric now fails here instead of silently re-opening the defect the code comment warns about.

Fail-before proven: with the `_id` branch temporarily changed to emit `Value::Number(0)`, the test fails with `#421: the _id sort value must be a String (got Number(0))...`; restored, it passes (1/1, 0.10s). fmt clean; clippy validating. Test-only change — no source touched, so no behavior/regression risk; CI runs the full workspace suite + clippy.

Context: #421 items 1 (meta-aware `memtable_primary_key_rejects`) and 2 (`meta_sort_value` comment) are already on main; item 3 (conformance) was run — the full `search/` category passes 449/0/2 and the required CI ES-YAML conformance job is green on every commit (see the issue thread). This PR binds the `_id` invariant from the "Also unresolved" set; the `sorted_shadow_for` invariant and the pre-existing `painless_script_limits` crash remain.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -339,6 +339,49 @@ mod collection_publication_fail_closed_tests {
         );
     }
 
+    /// #421 "Also unresolved": `_id`'s numeric-prefilter safety rests on a TYPE
+    /// invariant, not a gate. `compute_sort_values` emits the `_id` sort value as
+    /// a `String`, so `SortTopK::primary_f64_rejects`'s `as_f64()` is always
+    /// `None` for an `_id` sort and the numeric pre-filter never mis-rejects a
+    /// hit. The invariant held but had no binding test; pin it, so a change that
+    /// makes the `_id` sort value numeric fails here instead of silently
+    /// re-opening the defect the code comment warns about.
+    #[tokio::test]
+    async fn id_sort_value_stays_a_string_so_the_numeric_prefilter_declines_it() {
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine
+            .create_index("id-sort-invariant", text_schema())
+            .unwrap();
+        let idx = engine.get_index("id-sort-invariant").unwrap();
+        idx.abort_background_tasks();
+
+        let req = xerj_query::parse_request(&json!({
+            "query": {"match_all": {}},
+            "sort": [{"_id": "asc"}],
+        }))
+        .unwrap();
+
+        let vals =
+            super::compute_sort_values(&json!({"body": "x"}), 1.0, "doc-42", &req.sort, &idx);
+        assert_eq!(vals.len(), 1, "one sort field yields one sort value");
+        assert!(
+            matches!(&vals[0], serde_json::Value::String(s) if s == "doc-42"),
+            "#421: the `_id` sort value must be a String (got {:?}); a numeric value \
+             would let SortTopK::primary_f64_rejects' as_f64() apply and mis-reject an \
+             `_id`-sorted hit",
+            vals[0]
+        );
+        // The consumer end: `as_f64()` on the `_id` sort value is `None`, which is
+        // exactly what keeps the numeric pre-filter from ever touching an `_id` sort.
+        assert!(
+            vals[0].as_f64().is_none(),
+            "an `_id` sort value must not be numeric"
+        );
+    }
+
     async fn assert_flush_publication_blocks_reader_then_finishes(inject_error: bool) {
         let dir = tempfile::tempdir().unwrap();
         let mut config = Config::default();


xerj-org:

**Result: 3/3 MERGE_OK, ZERO BLOCKING.** Each skeptic independently mutation-proved the fail-before. Nits only.

- **nit**: `#[tokio::test]` is unnecessary since the body has no awaits — does not block.
- _rationale_: Test-only regression (index.rs:350, +43 lines, single file, no production code, no build break). The test genuinely binds the #421 type invariant: parses {"_id":"asc"} through the real parse_request/parse_sort path (yielding SortField{field:"_id"}), calls the real super::compute_sort_values (index.rs:40047), and confirms the field reaches the sf.field=="_id" branch (40062, fails is_score/is_doc_order). Asserts producer end (vals[0] is Value::String("doc-42")) and consumer end (as_f64().is_none()) — exactly what stops SortTopK::primary_f64_rejects (index.rs:38970) mis-rejecting an _id-sorted hit. Baseline passes (dev, 0.17s). Proved fail-before: changing the _id/_doc branch's Value::String to numeric made the test FAIL at index.rs:370 with the exact #421 message; reverted, tree clean, head 771487bb. Deterministic (in-process temp dir, pure function).

- _rationale_: TEST-ONLY, verified. git diff --stat = +43 lines in one file, all inside #[cfg(test)] mod; no production line, no other file. Mutation-proved the binding (flip _id branch at index.rs:40069 to Value::Number → test FAILS with the #421 assertion, then restored green — not vacuous). Name unique, text_schema() helper in same module, no collision. Builds+passes under 1.97.1: cargo test -p xerj-engine --lib id_sort_value_stays_a_string = 1 passed/0 failed. fmt --all --check exit 0. Deterministic (pure id echo). Working tree restored clean. No production code changed => zero regression impact.

- **nit**: cosmetic PR-body detail (fail-before narrative says Number(0) vs an arbitrary literal) — does not block.
- _rationale_: All factual claims verified. PRODUCER: compute_sort_values (40047) emits _id as Value::String(id.to_string()) at 40069 in the cited branch (40062); is_score/is_doc_order match only _score/_doc, parse preserves field=="_id", so the branch fires deterministically. CONSUMER: SortTopK::primary_f64_rejects (38970) reads the frontier via sort.first().and_then(Value::as_f64) at 38978; a String yields None → returns false (no rejection). In-source comment 39969-39975 documents this exact invariant. Compiles+passes on 1.97.1 (1 passed); fail-before proven independently (mutated _id branch to Number(42) → test fails at 370; restored, tree clean at 771487bb). PR body honest: explicitly "does not close #421" and names the remaining sorted_shadow_for invariant + painless_script_limits crash. Context items 1 (memtable_primary_key_rejects takes idx, meta-aware 39952/39987) and 2 (meta_sort_value comment has no "therefore agree", 13266) both hold.

The `#[tokio::test]`→`#[test]` nit noted for a follow-up; non-blocking per the merge gate.

PR #773, opened 2026-08-26
test(engine): #421 bind the build_doc_value_columns underscore-skip invariant

### LINKED ISSUE #421: Follow-up to #401: three unresolved adversarial-verification items were merged to main in #420

xerj-org:
Addresses the second of #421's "Also unresolved" hardening notes (does not close #421).

The issue flags: `sorted_shadow_for`'s correctness "rests on `build_doc_value_columns` skipping `_`-prefixed source keys, an invariant in a different function with no test binding the two." Meta-fields (`_seq_no`, `_id`, `_source`, ...) are envelope bookkeeping resolved from the version map, and must never be shadowed into a doc-value column from `_source` — otherwise a meta-sort/agg could read a stale source-derived column. The skip (`if field.starts_with(_) { continue }` in `build_doc_value_columns`) enforced it but lived only in a comment.

This adds a pure-function `#[test]`: a source doc carrying `_seq_no`/`_id` (plus real user fields) must yield NO column for the `_`-prefixed keys, while the user fields DO get columns (so the assertion is not vacuous). Removing the skip now fails here instead of silently shadowing meta-fields.

Fail-before proven: replacing the `starts_with(_)` guard with `if false` makes the test FAIL (`_`-prefixed keys get columns, 0 passed/1 failed); restored, it passes (dev, 0.01s). fmt clean; clippy validating. Test-only change — no production code touched, so no behavior/regression risk; CI runs the full workspace suite + clippy.

Context: #421 items 1 (meta-aware `memtable_primary_key_rejects`) and 2 (`meta_sort_value` comment) are on main; item 3 (conformance) was run (search/ 449/0/2, CI conformance job green); the `_id` sort-value invariant was bound in #772 (merged). This binds the second "Also unresolved" invariant; only the pre-existing `painless_script_limits` stack overflow remains.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -382,6 +382,35 @@ mod collection_publication_fail_closed_tests {
         );
     }
 
+    /// #421 "Also unresolved": `sorted_shadow_for`'s correctness rests on
+    /// `build_doc_value_columns` skipping `_`-prefixed source keys — `_seq_no`,
+    /// `_id`, `_source` and friends are envelope bookkeeping, not user data, and
+    /// must never be shadowed into a doc-value column from `_source` (a meta
+    /// field is resolved from the version map, never from a source column). That
+    /// invariant lived only in a comment (index.rs, the `starts_with('_')` skip)
+    /// with no test binding it to `sorted_shadow_for`. Pin it: a source doc
+    /// carrying `_`-prefixed keys must yield NO column for them, so removing the
+    /// skip fails here instead of silently shadowing meta-fields.
+    #[test]
+    fn build_doc_value_columns_skips_underscore_prefixed_source_keys() {
+        let doc = json!({"_seq_no": 5, "_id": "abc", "title": "hello", "count": 42});
+        let skip = std::collections::HashSet::new();
+        let cols = super::build_doc_value_columns([Some(&doc)].into_iter(), &skip);
+        assert!(
+            !cols.keys().any(|k| k.starts_with('_')),
+            "#421: no `_`-prefixed source key may become a doc-value column (got {:?})",
+            cols.keys().collect::<Vec<_>>()
+        );
+        // Sanity so the assertion above is not vacuously true: real user fields
+        // DO get shadowed, so the function did build columns — it just excluded
+        // the meta keys.
+        assert!(
+            cols.contains_key("title") && cols.contains_key("count"),
+            "user fields must be shadowed into columns (got {:?})",
+            cols.keys().collect::<Vec<_>>()
+        );
+    }
+
     async fn assert_flush_publication_blocks_reader_then_finishes(inject_error: bool) {
         let dir = tempfile::tempdir().unwrap();
         let mut config = Config::default();


xerj-org:

**Result: 3/3 MERGE_OK, ZERO BLOCKING.** Each independently proved fail-before. One cosmetic nit.

- _rationale_: TEST-ONLY PR (+29 lines, all in the #[cfg(test)] collection_publication_fail_closed_tests module; no production code touched, build compiles). The test calls the REAL super::build_doc_value_columns with {"_seq_no":5,"_id":"abc","title":"hello","count":42} and an EMPTY skip set, so the starts_with('_') guard at index.rs:20272 is the sole exclusion under test. It asserts no column key starts with '_' and a non-vacuous sanity check that title/count DO become columns. Proved fail-before: flipping the guard to `if false` made the test FAIL with got ["_id","_seq_no","count","title"] — that output also shows title/count present, confirming the sanity assertion is real. Baseline passes; reverted cleanly. The test genuinely binds the invariant.

- _rationale_: Diff vs origin/main is +29 lines in one file, one #[test] fn inside the existing #[cfg(test)] mod; no production line changed, no other file, zero regression impact. Call site matches the fn signature build_doc_value_columns(impl Iterator<Item=Option<&Value>>, &HashSet<String>) -> BTreeMap<String,Column> and compiles under 1.97.1. Non-vacuous. Fail-before proof: removing the `if field.starts_with('_') { continue; }` skip made the test FAIL (got ["_id","_seq_no","count","title"]); restoring PASSES. Unique name, lib test 1 passed/0 failed, fmt --check exit 0, worktree restored clean.

- **nit**: test placed in module `collection_publication_fail_closed_tests` (about flush publication, not doc-value columns); a doc-values/sort-shadow module would be a more natural home. Cosmetic only.
- _rationale_: Adds exactly one regression test, no production change (29 insertions, 0 deletions). The invariant is real: build_doc_value_columns skips `_`-prefixed keys at index.rs:20272-20274, and sorted_shadow_for's own comment (index.rs:24196-24198) names that exact `starts_with('_')` skip as the mechanism keeping meta-fields out of doc-value columns. Fail-before: neutralizing the skip made the test FAIL with got ["_id","_seq_no","count","title"] — both a numeric meta key (_seq_no) and a keyword meta key (_id) would leak without the skip; sanity assert confirms title+count genuinely built. Scope honest — "bind" not "close" #421; the companion _id test (#772) sits directly above.

Nit (module placement) noted; non-blocking. Both #421 invariant tests live in that mod together.

PR #776, opened 2026-08-26
fix(api): #394 suppress the semantic_text hint only when the vector dispatches

xerj-org:
#394. The lexical-on-`semantic_text` hint (#363/#393) suppressed itself whenever the field appeared in ANY `semantic`/`knn` clause. But a `semantic` clause beside a lexical one in a multi-clause bool never reaches the vector: `peel_semantic_query`/`peel_knn_query` (index.rs) dispatch a bool to the vector only when `must`+`should` holds exactly one candidate. So `bool.should:[match ctx, semantic ctx]` runs BM25-only, the embedding is not consulted, and suppressing the hint there reintroduces the exact silence the hint exists to break.

Fix: a new `dispatching_vector_fields` mirrors the engine's dispatch (whole-query / single-candidate bool / `hybrid` / top-level `knn`); only those count as "reached the vector". A blanket removal of the in-query suppression would have regressed the legit `hybrid` case, so mirroring dispatch is the point.

- flips `a_query_that_already_uses_semantic_on_the_field_is_not_flagged` -> `a_semantic_clause_that_never_reached_the_vector_is_still_flagged` (fail-before: reverting only the call-site hunk, the hint stays suppressed and the test fails)
- adds `a_hybrid_query_over_the_semantic_field_is_not_flagged` (regression guard: a hybrid still suppresses)
- top-level `knn` suppression unchanged

The hint is computed purely from the request JSON + schema, so this is verifiable without a model. fmt + clippy -D warnings clean; full xerj-api suite run under 1.97.1.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -7403,6 +7403,73 @@ fn vector_query_fields(q: &Value, out: &mut Vec<String>) {
     }
 }
 
+/// Fields whose vector clause GENUINELY reaches the vector for this query,
+/// mirroring the engine's `peel_semantic_query` / `peel_knn_query` dispatch
+/// (index.rs). A `semantic`/`knn` clause consults the vector only as the whole
+/// query, as the SOLE `must`/`should` candidate of a `bool` with no `must_not`,
+/// or inside a `hybrid` (which fuses and runs every branch). A `semantic`/`knn`
+/// clause sitting BESIDE a sibling in a multi-clause `bool` falls through to the
+/// lexical path and never consults the vector (#394) — so it must NOT count as
+/// "reached the vector" here, or the hint gets suppressed on a query that ran
+/// BM25-only, reintroducing the exact silence the hint exists to break. Contrast
+/// [`vector_query_fields`], which collects every mention regardless of shape and
+/// is right only for the top-level `knn` block (which always dispatches).
+fn dispatching_vector_fields(q: &Value, out: &mut Vec<String>) {
+    let Value::Object(o) = q else {
+        return;
+    };
+    // A direct `semantic`/`knn` at this level dispatches.
+    for key in ["semantic", "knn"] {
+        if let Some(v) = o.get(key) {
+            let specs: Vec<&Value> = match v {
+                Value::Array(a) => a.iter().collect(),
+                single => vec![single],
+            };
+            for spec in specs {
+                if let Some(f) = spec.get("field").and_then(Value::as_str) {
+                    out.push(f.to_string());
+                }
+            }
+        }
+    }
+    // `hybrid` runs and fuses every branch, so each branch dispatches on its
+    // own terms.
+    if let Some(Value::Array(branches)) = o.get("hybrid").and_then(|h| h.get("queries")) {
+        for entry in branches {
+            if let Some(inner) = entry.get("query") {
+                dispatching_vector_fields(inner, out);
+            }
+        }
+    }
+    // `constant_score` is a transparent wrapper (the engine peels it).
+    if let Some(inner) = o.get("constant_score").and_then(|c| c.get("filter")) {
+        dispatching_vector_fields(inner, out);
+    }
+    // `bool` reaches its vector clause ONLY in the single-`must`/`should`
+    // candidate, no-`must_not` shape the engine peels; every other bool falls
+    // through to the lexical path and consults no vector.
+    if let Some(Value::Object(b)) = o.get("bool") {
+        let must_not_empty = match b.get("must_not") {
+            None | Some(Value::Null) => true,
+            Some(Value::Array(a)) => a.is_empty(),
+            Some(_) => false,
+        };
+        if must_not_empty {
+            let mut candidates: Vec<&Value> = Vec::new();
+            for key in ["must", "should"] {
+                match b.get(key) {
+                    Some(Value::Array(a)) => candidates.extend(a.iter()),
+                    Some(Value::Null) | None => {}
+                    Some(other) => candidates.push(other),
+                }
+            }
+            if candidates.len() == 1 {
+                dispatching_vector_fields(candidates[0], out);
+            }
+        }
+    }
+}
+
 /// Hint for a lexical full-text query aimed at a `semantic_text` field (#363).
 ///
 /// `semantic_text` maps to `FieldType::Text` plus an `EmbeddingConfig`, and the
@@ -7448,11 +7515,13 @@ fn lexical_on_semantic_text_hint(
     if lexical.is_empty() {
         return None;
     }
+    // Only the vector clauses that GENUINELY dispatch count as "reached the
+    // vector" — a `semantic`/`knn` beside a lexical clause in a multi-clause
+    // bool falls through to BM25 and consulted no embedding, so it must not
+    // suppress the hint (#394). The top-level `knn` block is the exception:
+    // it lives outside `query` and always dispatches.
     let mut vectorised = Vec::new();
-    vector_query_fields(query, &mut vectorised);
-    // The ES 8.x top-level `knn` block lives outside `query`, and pairing it
-    // with a `match` is the canonical hand-rolled hybrid — the one shape that
-    // most deserves not to be nagged.
+    dispatching_vector_fields(query, &mut vectorised);
     if let Some(knn) = body.knn.as_ref() {
         vector_query_fields(&json!({ "knn": knn }), &mut vectorised);
     }
@@ -7468,9 +7537,12 @@ fn lexical_on_semantic_text_hint(
             .target_field
             .clone()
             .unwrap_or_else(|| format!("{field}_vector"));
-        // A caller who already wrote `semantic`/`knn` for this field knows
-        // what the field is. Repeating it is the noise that gets a hint
-        // channel muted, and a hybrid query is a correct use, not a mistake.
+        // A caller whose query actually reached the vector for this field
+        // (a dispatching `semantic`/`knn`, a top-level `knn`, or a `hybrid`)
+        // knows what the field is; repeating it is the noise that mutes a hint
+        // channel, and a hybrid is a correct use, not a mistake. A `semantic`
+        // clause that never dispatched (a multi-clause bool) is NOT in this set
+        // (#394), so its lexical sibling is still flagged.
         if vectorised.iter().any(|f| *f == field || *f == companion) {
             continue;
         }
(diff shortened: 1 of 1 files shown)


xerj-org:
3-skeptic adversarial pass (correctness / regression / evidence-holds), each REFUTING in its own worktree under 1.97.1. Verdicts verbatim:

### correctness
**Verdict: HOLDS — blocking: False**

**Summary:** The fix correctly narrows the semantic_text hint's suppression to genuinely-dispatching vector clauses and fixes #394 (a semantic/knn beside a lexical clause in a multi-clause bool is now flagged). Because dispatching_vector_fields collects a strict subset of the old blanket vector_query_fields set, the PR can only add hint firings, never suppress more — so it introduces no over-suppression regression, and every added firing lands on a shape where the vector was genuinely not reached. I could not construct any realistic query where the new hint is wrong: the common dispatching shapes (top-level knn via body.knn unchanged, hybrid, constant_score, single-candidate bool incl. single-object must/should, must_not-as-object) all match peel exactly, and the missed shapes can't produce an observable wrong hint.

**Evidence:**

```
Setup: git reset --hard origin/fix/394-hint-dispatch-mirror -> HEAD 6cc8d8d4; grep 'fn dispatching_vector_fields' es_compat.rs -> line 7417 (present).
Engine dispatch read from index.rs: peel_knn_query (31810) and peel_semantic_query (32031) bool arm — `if !must_not.is_empty() return None; candidates = must.iter().chain(should.iter()); if candidates.len()!=1 return None; peel(candidates[0])`. filter is merged, never a candidate. These are called only on the top-level query at search_inner (14559/14709/14736); the generic fall-through evaluates nested Knn/SemanticSearch/Hybrid via doc_matches_query_typed whose tail is `_ => false` (37498) — vector not consulted.
Rewriter off-path: grep 'rewriter::rewrite' across engine crate -> only a comment at index.rs:32718 ("not xerj_query::rewriter::rewrite, which is not on this path"); pub fn rewrite lives in xerj-query/rewriter.rs:34 and is never called by the engine, so Bool(must=[X]) is NOT unwrapped and peel sees the raw parse.
Regression bound: git show HEAD diff shows old `vector_query_fields(query, &mut vectorised)` replaced by `dispatching_vector_fields(query, ...)`; dispatching_vector_fields only pushes fields from o.get("semantic")/o.get("knn") specs (es_compat.rs:7422-7434), so its output ⊆ vector_query_fields output. New vectorised ⊆ old vectorised => suppression strictly narrows => firings only added, and each added firing is a non-dispatching placement (correct).
Core #394 test verified by construction: flipped test bool{should:[match ctx, semantic ctx]} -> walker candidates.len()==2 -> not collected -> flagged; engine peel returns None (2 candidates) -> BM25-only -> hint should fire. Match.
Parser check: boost on knn/semantic is an inline node field (parse_knn/parse_semantic ~545/627), not a Boosted wrapper, so the walker's object-key match still catches it; hybrid branch shape is {queries:[{query:{...}}]} (parser.rs:2949) matching the walker's entry.get("query").
Did NOT run cargo: no prebuilt target/ dir; fresh xerj-api+xerj-engine build is the tens-of-minutes cost per memory, so verdict is grounded in source analysis with the line refs above.
```

**Nits (non-blocking):**
- Docstring/commit over-claim: dispatching_vector_fields recurses into `hybrid` inside a bool candidate where peel_semantic/peel_knn would return None (their bool arm never descends into Hybrid). This over-suppresses in contrived shapes like bool{should:[hybrid], filter:[match ctx]} (BM25 hits, hint silenced), but it is a pre-existing leftover of the old blanket vector_query_fields walk — identical before/after, not a regression, and not realistic.
- Walker doesn't mirror peel_multi_knn_query (a pure bool{should:[knn,knn]}) or peel_nested_knn_query. Harmless: a pure multi-knn bool carries no lexical text clause, and nested-knn would require a semantic_text field under the nested path, so neither yields an observable wrong hint — but the 'mirrors the engine dispatch exactly' phrasing is slightly stronger than the code.
- Both directions checked: the fix cannot introduce a new over-suppression (subset property), and no realistic wrongly-fires case exists because boost is inline (no Boosted JSON wrapper) and every common dispatching shape is handled.

### regression
**Verdict: HOLDS — blocking: False**

**Summary:** PR #776 faithfully mirrors the engine's vector-dispatch in the new dispatching_vector_fields and fixes a genuine silence bug (a semantic/knn clause beside a lexical one in a multi-clause bool ran BM25-only yet suppressed the hint). Full xerj-api suite is green under 1.97.1 (226 lib tests + all 60 test binaries, 0 failed). The two changed tests and the unchanged top-level-knn suppression test all pass. No untested regression: the in-bool knn-beside-match case now firing is CORRECT (engine's doc_matches_query_typed returns false for a non-peeled Knn clause, index.rs:37498, so the vector is not consulted), and no realistic hybrid/knn workflow is falsely nagged — the mirror handles hybrid branches, single-candidate bool, and constant_score exactly as peel_knn_query/peel_semantic_query do (index.rs:31810/32031), counting must+should candidates only.

**Evidence:**

```
grep -n 'fn dispatching_vector_fields' engine/crates/xerj-api/src/es_compat.rs => 7417 (symbol present after reset to origin/fix/394-hint-dispatch-mirror @6cc8d8d4).

rustup run 1.97.1 rustc --version => rustc 1.97.1 (8bab26f4f 2026-07-14).

RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-api =>
  lib: test result: ok. 226 passed; 0 failed
  es_compat::semantic_text_lexical_hint_tests::a_semantic_clause_that_never_reached_the_vector_is_still_flagged ... ok (changed/flipped)
  ...::a_hybrid_query_over_the_semantic_field_is_not_flagged ... ok (new)
  ...::a_top_level_knn_over_the_companion_vector_suppresses_the_hint ... ok (unchanged)
  ...::match_on_semantic_text_is_flagged_and_names_the_semantic_query / multi_match_naming_a_semantic_text_field_is_flagged / match_on_a_plain_text_field_is_not_flagged / the_hint_is_additive_to_the_unknown_field_hint / control_index_proves_match_on_semantic_text_is_plain_bm25 ... all ok
  aggregate: 60 'test result: ok' lines, grep for '[1-9]+ failed'/FAILED/panicked => empty.

Engine cross-check: peel_knn_query (index.rs:31810) and peel_semantic_query (index.rs:32031) dispatch a bool only when must_not empty AND must+should has exactly one candidate; peel_hybrid_query (index.rs:32091) runs every branch. Generic scan doc_matches_query_typed (index.rs:36032) has trailing `_ => false` (index.rs:37498) — a Knn/SemanticSearch clause in a non-peeled bool matches nothing, confirming a multi-clause bool runs BM25-only and the mirror's decision to still flag is correct. In-query knn dispatch handled at index.rs:14559 (peel_knn_query), body.knn still blanket-collected via vector_query_fields.
```

**Nits (non-blocking):**
- dispatching_vector_fields handles the constant_score transparent wrapper but not a query-level boost/_name wrapper; harmless in practice because ES JSON expresses boost/_name as inner leaf params, not standalone wrapper keys, so no realistic dispatching shape is missed (would only under-collect and over-fire, never falsely suppress).

### EVIDENCE-HOLDS skeptic: fail-before reproduction + root dispatch-claim source verification for PR #776 (#394)
**Verdict: HOLDS — blocking: False**

**Summary:** All three load-bearing claims hold under 1.97.1. (1) Fail-before: reverting ONLY the call-site (dispatching_vector_fields->vector_query_fields) makes a_semantic_clause_that_never_reached_the_vector_is_still_flagged FAIL (hint suppressed); restoring makes all 8 tests pass. (2) Root claim: peel_semantic_query (index.rs:32058-32064) dispatches a bool to the vector only when must_not is empty AND must.iter().chain(should.iter()) has exactly one candidate; a bool.should:[match,semantic] has 2 candidates and falls through to lexical, never consulting the vector — corroborated at runtime by the fail-before output returning plain BM25 (2 hits, doc 1 absent). (3) The flipped test asserts behavior the code actually produces: for the 2-candidate bool dispatching_vector_fields leaves vectorised empty, so the match is flagged and reason names ctx; hybrid guard suppresses correctly. Cannot refute.

**Evidence:**

```
SETUP: `git reset --hard origin/fix/394-hint-dispatch-mirror` -> HEAD 6cc8d8d4. `grep -n 'fn dispatching_vector_fields' engine/crates/xerj-api/src/es_compat.rs` -> 7417. `rustup run 1.97.1 rustc --version` -> rustc 1.97.1 (8bab26f4f 2026-07-14).

BASELINE (fixed branch), `rustup run 1.97.1 cargo test -p xerj-api --lib semantic_text_lexical_hint_tests`:
  running 8 tests ... test result: ok. 8 passed; 0 failed. Includes a_semantic_clause_that_never_reached_the_vector_is_still_flagged ... ok AND a_hybrid_query_over_the_semantic_field_is_not_flagged ... ok.

FAIL-BEFORE: Edited es_compat.rs:7524 `dispatching_vector_fields(query, &mut vectorised)` -> `vector_query_fields(query, &mut vectorised)` (only change). Re-run:
  test a_semantic_clause_that_never_reached_the_vector_is_still_flagged ... FAILED
  panic at es_compat.rs:8325: "#394: a nested semantic beside a lexical clause never reached the vector, so the `match` must still be flagged: {...hits:[{_id:0,_score:2.0986123},{_id:2,_score:1.6931472}]}" (semantic_hint returned None -> unwrap_or_else panic). test result: FAILED. 7 passed; 1 failed. Note the response is 2 plain-BM25 hits (doc 1 'segment merger' absent), runtime proof the semantic clause did not dispatch. Restored the line; `git diff` empty (working tree == HEAD).

ROOT CLAIM, engine/crates/xerj-engine/src/index.rs peel_semantic_query (lines 32031-32085): Bool arm 32043-32082; line 32058 `if !must_not.is_empty() { return None; }`; line 32061 `let candidates: Vec<&QueryNode> = must.iter().chain(should.iter()).collect();`; lines 32062-32064 `if candidates.len() != 1 { return None; }`. Comment 32050-32057 states "Multi-clause bools (a semantic clause beside a lexical one) still fall through." API mirror es_compat.rs:7451-7470 uses the identical shape (must_not empty + candidates.len()==1 over must+should). Test query {"bool":{"should":[{"match":...},{"semantic":...}]}} = 2 should candidates -> len!=1 -> no dispatch -> vectorised empty -> match flagged. Matches assertion.
```

**Nits (non-blocking):**
- dispatching_vector_fields mirrors constant_score + hybrid + bool, but any completeness gap in the mirror errs toward over-hinting (suppress less), the safe direction relative to #394's under-suppression/silence complaint, so no correctness risk from an imperfect mirror.
- The 'never used' dead_code warning for dispatching_vector_fields appears only under the reverted fail-before state, not on the branch as shipped.

---
**Gate:** 3/3 HOLDS, 0 BLOCKING. All 19 checks green, 0 pending, mergeStateStatus CLEAN, head 6cc8d8d4 (the SHA the skeptics verified). Merging. Nits → follow-up issue, they do not block.

Issue #770, opened 2026-08-25
CI #751 follow-up: build-test job-timeout margin + other main jobs can still starve the slot


xerj-org:
Two non-blocking nits from the #767 (PR) 3-skeptic pass, filed for tracking.

1. **Job-cap margin.** #767 set `build-test` `timeout-minutes: 50`. The measured worst-case *legitimate* run is ~43-46 min (setup + the ~36-min threads=2 build step + the ~2.3-min default-parallelism step + cache-save), so a heavily-loaded runner could occasionally false-fail. The step-6 hang is caught by the precise 12-min step cap regardless, so the job cap only guards step-5/setup hangs (which run indefinitely, so 50 vs 60 both catch them). Consider bumping the job cap to 55-60 min for extra margin. Advisory; a false-fail is recoverable via `gh run rerun --failed`.

2. **Starvation scope.** #767 caps only `build-test`. A hang in another `main` job (conformance, ux-tests, etc.) would still hold the `ci-${github.ref}` concurrency slot (`cancel-in-progress: false` on main) up to GitHub's 6h default and starve newer main CI — the same class #751 describes, just via a different job. A blanket job-level default (or per-job caps) would close the vector fully.

Both are mitigation-tuning, not the #751 root race (the port/fixture bisect). Low priority.

xerj-org:
Scoping nit 2 (other-job starvation) before anyone acts on it.

The #751 hang is specific to `build-test` step 6 — `cargo test --profile ci-test --workspace` at DEFAULT parallelism. Audited the workflow: no other `main` job runs that suite (release-build does `cargo build --release`, conformance runs es-yaml-runner, api-checks/usecase-smoke run harness scripts, etc.). So the concurrency-slot starvation vector is currently confined to `build-test`, which #767 already bounds (12-min step + 50-min job cap).

For another job to stall main for hours it would have to hang for a NEW reason none of them exhibits today. So blanket per-job `timeout-minutes` is speculative defense-in-depth, not a fix for a live problem — and adding it to ~13 jobs would churn every PRs CI. Recommend leaving it until a second job is actually observed hanging (or fold a blanket cap in opportunistically with another ci.yml change). Nit 1 (bump the 50-min build-test job cap for margin) stands as advisory.

Keeping this open as low-priority.

xerj-org:
Re-read this issue against today's tree. Splitting the two nits, because they no longer have the same status.

[...]

PR #780, opened 2026-08-26
ci: widen build-test job-timeout margin to 60 min (#770)

Vinz2168:
Closes nit 1 of #770.

`build-test`'s job-level `timeout-minutes` was 50, but the measured legitimate worst-case run is ~43-46 min (setup + the ~36-min threads=2 build/test step + the ~2.3-min default-parallelism step + cache-save) — too thin a margin on a loaded runner, risking an occasional false-fail.

This only widens the outer defense-in-depth job cap to 60 min. The precise 12-min step-6 cap (added by #767) still catches the #751 hang exactly as before; this change doesn't touch it and doesn't address the #751 root race itself.

Nit 2 from #770 (other-job starvation scope) is explicitly out of scope here — the issue's own follow-up comment concluded it's speculative defense-in-depth with no live problem to fix yet, and recommended leaving it.

https://claude.ai/code/session_012Y7WxJvZSAN9Vj5Gsma91t

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -269,13 +269,16 @@ jobs:
     # bounds the one documented hang (step 6, default parallelism) precisely and
     # fast. This job-level cap is defense-in-depth: it bounds ANY hang in the job
     # (setup, cache, a future test race the step timeout would not cover) to
-    # 50 min instead of the 6h default. It is deliberately generous because
+    # 60 min instead of the 6h default. It is deliberately generous because
     # step 5 (build + threads=2 tests) can legitimately take ~36 min on a loaded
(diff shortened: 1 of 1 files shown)


xerj-org:
This gives #770 the headroom it needs: step-5 compile (~36min) plus step-6 (up to its 12min cap) crowds the old 50min job limit, and 60 leaves margin. Timeout-only, safe.

One note so it isn't conflated with #751: this does not change the step-6 `timeout-minutes: 12`, which is the fail-fast bound on the default-parallelism hang. The intermittent Build+Test failures on recent PRs were that step timing out at 12min (a re-run passes) — a separate problem from the job-level margin here. Widening the job timeout is right for #770 regardless.

LGTM for #770 on the merits (green, ci.yml-only).

PR #783, opened 2026-08-26
fix(engine): #782 CIDR term on an ip field returns 0 after flush — rewrite to range

### LINKED ISSUE #782: wrong-results: CIDR `term` query on an ip field matches in the memtable but returns 0 after flush (segment path lacks the CIDR fallback)
A `term` query with CIDR notation on an `ip` field matches docs in the memtable but returns **zero after they flush to a segment** — a silent flush-divergence (same query, different results before/after flush), the #413/#423 pattern. `range` on the CIDR bounds works on both paths, so the data is fine; only the CIDR `term` breaks on the segment.

## Reproduced (current main / rc.70, own node, non-9200)

Index `ip`-mapped `addr` with 192.168.1.5, .200, .0, .255, 10.0.0.1, 192.168.2.5, then a fresh 192.168.1.77.

```
# memtable (fresh doc, not force-merged):
term addr=192.168.1.0/24   -> matches 192.168.1.77   ✅
# after _flush + _forcemerge:
term addr=192.168.1.0/24   -> []   ❌  (should be all 192.168.1.x)
# range still works on both:
range addr gte 192.168.1.0 lte 192.168.1.255 -> all 192.168.1.x   ✅
```

Also 0 after flush for `/16`, `/25`, `/32`. Exact `term addr=192.168.1.5` works (in dict).

## Cause

The memtable fast path `try_doc_values_query` (index.rs:35183) and the mem-bool path (35116) explicitly BAIL on a `/`-containing value (`return None`) → fall back to the source-scan matcher `doc_matches_query`, whose `Term` arm handles CIDR via `ip_matches_cidr` (index.rs:36041/40495). The **segment** term path does an exact term-dictionary lookup for the literal `"192.168.1.0/24"`, finds nothing (the dict holds exact IPs), and returns 0 **without** the same CIDR fallback. ES answers CIDR `term` on an ip field as the subnet range.

## Fix direction (two options, both decision-free)

1. **Rewrite** a `Term{ip_field, "a.b.c.d/n"}` into `Range{gte: network, lte: broadcast}` at query build/normalization where the schema is available — `range` already works on both memtable and segment, so this fixes it uniformly.
2. Make the **segment** term path route a CIDR value on an ip field to the source-scan `doc_matches_query`/`ip_matches_cidr` fallback, mirroring what the memtable fast paths already do by bailing.

Found via a proactive probe. Will take (1) or (2) with a fail-before segment test (memtable matches, segment 0) once the exact segment term-resolution site is pinned.

xerj-org:
Closes #782.

A `term` query with CIDR notation on an `ip` field (`{"term":{"addr":"192.168.1.0/24"}}`) matched in the memtable but returned **0 after the docs flushed to a segment** — a silent flush-divergence. The memtable fast paths (`try_doc_values_query`, the mem-bool builder) bail on a `/`-containing value and fall back to the source-scan matcher, whose `Term` arm expands CIDR via `ip_matches_cidr`. The segment term-dictionary lookup can only find the literal string `"192.168.1.0/24"` (the dict holds exact IPs), so it returned nothing with no fallback. `range` on the same bounds worked on both paths.

## Fix

Rewrite a CIDR `term` on an `ip` field into the equivalent inclusive `range [network, broadcast]` inside `rewrite_query_aliases` — the schema-aware pass applied at search time *before* the memtable/segment split (index.rs:14433). `range` already works identically on every path, so this fixes it uniformly with one change. Added `cidr_to_ip_range` (std::net only, IPv4 + IPv6). Only fires when the schema types the field `ip` and the value contains `/`; exact terms and CIDR-looking strings on non-ip fields are untouched.

## Verify

- New unit test `cidr_term_on_an_ip_field_rewrites_to_the_subnet_range` — `cidr_to_ip_range` bounds (/24, /25, /32, /16, invalid), and the rewrite (CIDR-on-ip → Range; exact-on-ip → Term; CIDR-on-keyword → Term).
- **Fail-before**: disabling only the rewrite condition makes the test fail (`got Term`, not Range).
- Reproduced end-to-end on an own node (non-9200): pre-fix a CIDR term matched in the memtable but returned 0 after `_forcemerge`; `range` matched on both.
- fmt + clippy `-D warnings` clean; full `xerj-engine` lib suite run under 1.97.1.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -118,6 +118,74 @@ mod collection_publication_fail_closed_tests {
         schema
     }
 
+    #[test]
+    fn cidr_term_on_an_ip_field_rewrites_to_the_subnet_range() {
+        // #782: a CIDR `term` on an ip field silently returned 0 after a flush
+        // because the segment term-dictionary lookup can't expand a CIDR. Fix:
+        // rewrite it to the equivalent inclusive range (which every path
+        // handles) before execution.
+        assert_eq!(
+            cidr_to_ip_range("192.168.1.0/24"),
+            Some(("192.168.1.0".to_string(), "192.168.1.255".to_string()))
+        );
+        assert_eq!(
+            cidr_to_ip_range("192.168.1.0/25"),
+            Some(("192.168.1.0".to_string(), "192.168.1.127".to_string()))
+        );
+        assert_eq!(
+            cidr_to_ip_range("10.0.0.1/32"),
+            Some(("10.0.0.1".to_string(), "10.0.0.1".to_string()))
+        );
+        assert_eq!(
+            cidr_to_ip_range("192.168.0.0/16"),
+            Some(("192.168.0.0".to_string(), "192.168.255.255".to_string()))
+        );
+        assert_eq!(cidr_to_ip_range("not-a-cidr"), None);
+        assert_eq!(cidr_to_ip_range("192.168.1.5"), None);
+
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("addr", FieldType::Ip))
+            .unwrap();
+        schema
+            .add_field(FieldConfig::new("name", FieldType::Keyword))
+            .unwrap();
+
+        let term = |f: &str, v: serde_json::Value| QueryNode::Term {
+            field: f.into(),
+            value: v,
+            boost: None,
+        };
+        // CIDR term on the ip field -> inclusive [network, broadcast] range.
+        match rewrite_query_aliases(&term("addr", json!("192.168.1.0/24")), &schema) {
+            QueryNode::Range {
+                field,
+                gte,
+                lte,
+                gt,
+                lt,
+                ..
+            } => {
+                assert_eq!(field, "addr");
+                assert_eq!(gte, Some(json!("192.168.1.0")));
+                assert_eq!(lte, Some(json!("192.168.1.255")));
+                assert_eq!(gt, None);
+                assert_eq!(lt, None);
+            }
+            other => panic!("#782: CIDR term on ip must rewrite to a Range, got {other:?}"),
+        }
+        // A non-CIDR (exact) term on the ip field stays a Term.
+        assert!(matches!(
+            rewrite_query_aliases(&term("addr", json!("192.168.1.5")), &schema),
+            QueryNode::Term { .. }
+        ));
+        // A CIDR-looking value on a NON-ip field is a literal string, unchanged.
+        assert!(matches!(
+            rewrite_query_aliases(&term("name", json!("192.168.1.0/24")), &schema),
+            QueryNode::Term { .. }
+        ));
+    }
+
     async fn assert_active_reader_waits_for_commit<F, Fut>(
         idx: &Arc<Index>,
         id: &'static str,
@@ -34043,6 +34111,28 @@ fn rewrite_query_aliases(q: &QueryNode, schema: &Schema) -> QueryNode {
             boost,
         } => {
             let resolved = resolve_field_alias(schema, field);
+            // A CIDR `term` on an `ip` field means the whole subnet. The
+            // source-scan matcher expands it (`ip_matches_cidr`), but the
+            // segment term-dictionary lookup can't, so it silently returned 0
+            // after a flush (#782). Rewrite it to the equivalent inclusive
+            // range, which every path (memtable + segment) handles the same.
+            if let Some(cidr) = value.as_str() {
+                if cidr.contains('/')
+                    && declared_field(schema, &resolved)
+                        .is_some_and(|fc| matches!(fc.field_type, FieldType::Ip))
+                {
+                    if let Some((lo, hi)) = cidr_to_ip_range(cidr) {
+                        return QueryNode::Range {
+                            field: resolved,
+                            gte: Some(Value::String(lo)),
+                            gt: None,
+                            lte: Some(Value::String(hi)),
+                            lt: None,
+                            boost: *boost,
+                        };
+                    }
+                }
+            }
             QueryNode::Term {
                 field: resolved,
                 value: value.clone(),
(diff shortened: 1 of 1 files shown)


xerj-org:
Three adversarial skeptics, each in its own worktree under 1.97.1, told to REFUTE. All three HOLDS, 0 blocking. Verbatim:

The #782 fix is correct. cidr_to_ip_range computes network/broadcast bounds correctly for every case I constructed (host-bits-set 192.168.1.5/24 -> 192.168.1.0..255, /0, /31, /32, /16, /25, /8, /12, IPv6 ::/0 and /32//48//64//128; invalid prefixes -> None; shift-overflow at prefix 0 correctly guarded). The rewrite is gated tightly: only a string value containing '/' on a declared FieldType::Ip field is rewritten; non-ip fields, exact ip values, and unparseable CIDRs stay a Term. rewrite_query_aliases runs on the live search path (index.rs:14501). Crucially, the rewritten Range compares correctly on the SEGMENT path: query_node_to_fts has no Range arm (returns None), is_doc_scan_query(Range)=true forces scan_stored, and build_range_prefilter_cached short-circuits to None for ip string bounds (parse::<f64> and parse_date_ms both reject ip strings), so the full stored-section scan runs doc_matches_query_typed's ip branch which compares via parse_ip_to_u128 as u128 numeric, not lexicographic. Memtable path (mem_bool_preds) likewise declines non-numeric ip bounds and falls to the same IP-aware source scan. Could not refute.

Evidence: standalone replica of cidr_to_ip_range + parse_ip_to_u128 compiled+run under 1.97.1: 192.168.1.5/24 => (192.168.1.0, 192.168.1.255); /31 => (.4,.5); /32 => (self,self); 0.0.0.0/0 => (0.0.0.0, 255.255.255.255); /25 => (..,.127); /16 => (..,255.255); ::/0, 2001:db8::/32, /48, /64, /128 all correct; /33 and /129 => None; bare ip => None; 10.0.0.5/8 => (10.0.0.0, 10.255.255.255); 172.16.5.4/12 => (172.16.0.0, 172.31.255.255). Membership at segment scale (u128) for /24: .0/.5/.255 in-range; 192.168.2.0 and 192.168.0.255 out. Lexicographic-hazard check: numeric 192.168.1.9 < 192.168.1.10 true, but string "192.168.1.10" < "192.168.1.9" true — the ip Range branch uses u128, avoiding it. Routing: query_node_to_fts default _ => None (no Range arm); is_doc_scan_query includes Range; build_range_prefilter_cached ip strings -> None -> full scan; doc_matches_query_typed ip Range via parse_ip_to_u128; rewrite call site index.rs:14501.

## regression — HOLDS (non-blocking)

Could not refute PR #783. Full xerj-engine lib suite is green under 1.97.1 (641 passed, 0 failed), the new test passes, and fmt+clippy are clean. The rewrite is tightly guarded (value must contain '/', field must be FieldType::Ip, CIDR must parse) so exact ip terms and '/'-values on non-ip fields are untouched (both tested). It fires at the single pre-split chokepoint (index.rs:14501) and recurses into bool children, so memtable and segment both run the same Range — no new divergence. Range-on-ip matching is numeric everywhere: the source scan uses parse_ip_to_u128, and both the dv range prefilter and the count fast path DECLINE ip string bounds (parse::<f64> and parse_date_ms fail -> return None -> fall through to the ip-aware source scan), so there is no lexicographic mis-comparison and no wrong exclusion. Endpoints [network,broadcast] match ip_matches_cidr semantics; boost is preserved. No scoring regression: exact ip terms don't convert, and the CIDR case was already 0-after-flush; the Range is a non-scoring residual (constant score) consistent with ES and with the pre-flush memtable source-scan path.

Evidence: full suite `cargo test -p xerj-engine --lib` under 1.97.1 -> ok. 641 passed; 0 failed; finished in 8.93s. New test cidr_term_on_an_ip_field_rewrites_to_the_subnet_range -> 1 passed. fmt --all --check -> exit 0. clippy -p xerj-engine --all-targets -D warnings -> exit 0. Soundness reads: range prefilter ip decline index.rs:23769-23780; count shortcut same 21090-21102; ip-aware numeric Range source scan 36185-36226; rewrite applied once at 14501, recurses at 34203-34215.

Nits (non-blocking): (1) the new test asserts the rewrite SHAPE only, no end-to-end index->flush->query; a full ingest+flush+search assertion (e.g. a host like 192.168.1.9) would lock it against future prefilter changes. (2) /0 and IPv6 paths read correct but are not covered by the test's cases. (3) a non-canonical CIDR (192.168.1.5/24) is silently normalized to [.0,.255] rather than rejected — matches ES leniency, just undocumented.

## evidence-holds — HOLDS (non-blocking)

All three load-bearing claims of #783 hold. Fail-before fails exactly as claimed ("got Term"); the memtable-bails-and-source-scans vs segment-prefilter-skips asymmetry is accurately described and cited; range-on-ip is a genuinely working path. Confirmed the fix end-to-end with a real flush: declared-ip CIDR term = 2 post-flush and equivalent range = 2 post-flush. No blocking defect: neither the fail-before nor the asymmetry description broke. One non-blocking scope gap: the rewrite is gated on a declared FieldType::Ip, so an undeclared (schemaless/dynamic) ip-shaped field still exhibits the exact #782 symptom — the flush test showed schemaless post-flush CIDR term = 0 while pre-flush = 2. A completeness limitation the PR does not claim to cover, not a misdescription of the fix.

Evidence: (1) FAIL-BEFORE — baseline new test passes; editing `if cidr.contains('/')` -> `if false && cidr.contains('/')` makes it FAIL with "#782: CIDR term on ip must rewrite to a Range, got Term { field: addr, value: String(192.168.1.0/24) }"; restored, git diff empty. (2) ASYMMETRY cited: memtable try_doc_values_query Term arm 35273 and mem_bool_preds Term arm 35206 both `if s.contains('/') { return None }`, fall to DocsForScan -> doc_matches_query_typed Term arm 36131 -> ip_matches_cidr 36133; segment bare-Term prefilter build_term_prefilter_cached Keyword arm 23933 `ord_for_term` only (CIDR literal absent -> empty prefilter) -> segment skipped 17688; count-shortcut seg_term_positions 39245 also ord-only. (3) RANGE-ON-IP sound: doc_matches_query_typed Range arm parses u128 via parse_ip_to_u128 36200; build_range_prefilter_cached returns None for ip string bound -> full scan. End-to-end real-flush test: pre-flush CIDR term = 2, post-flush CIDR term = 2 (declared ip, fix works), post-flush range = 2, schemaless post-flush CIDR term = 0 (scope gap). Worktree left clean.

---

Merge gate: 0 BLOCKING across all three lenses, all 19 checks green none-pending, MERGEABLE + CLEAN at the reviewed head 6b2c2c44. Merging.

The schemaless/dynamic ip scope gap the evidence lens surfaced (undeclared ip-shaped field still 0 after flush) is real but out of #782's scope — #782 was filed and reproduced on a declared ip mapping, which this fixes. Filed separately as a follow-up. The regression nit about no end-to-end flush test is a fair should-fix; folding into the follow-up too.

PR #787, opened 2026-08-26
fix(engine): #786 route a schemaless CIDR term to the source scan so it survives flush

### LINKED ISSUE #786: CIDR term on a schemaless/dynamic ip field still returns 0 after flush (#782 residual)
Follow-up to #782/#783. #783 fixed the CIDR-`term`-on-`ip` flush-divergence for **declared** `ip` fields by rewriting the term to a `[network, broadcast]` range in `rewrite_query_aliases`. That rewrite is gated on the schema declaring the field `ip`:

```rust
if cidr.contains('/')
    && declared_field(schema, &resolved)
        .is_some_and(|fc| matches!(fc.field_type, FieldType::Ip))
```

So an ip-shaped value on an **undeclared / schemaless (dynamic)** field is not rewritten, and still shows the exact #782 symptom: matches in the memtable, returns 0 after flush.

## Repro (verified under 1.97.1 during #783 review)

Same setup as #782 but with no `ip` mapping (dynamic field):

```
schemaless pre-flush  CIDR term addr=192.168.1.0/24  -> 2   (memtable source-scan expands via ip_matches_cidr)
schemaless post-flush CIDR term addr=192.168.1.0/24  -> 0   (segment term-dict lookup finds only the literal)
```

Declared-`ip` case is fixed (post-flush = 2). Only the dynamic case is left.

## Why it is separate from #782

#782 was filed and reproduced on an explicit `{"addr":{"type":"ip"}}` mapping, which #783 fixes and closes. The dynamic case is the same underlying asymmetry but needs the field's ip-ness to be known at query-rewrite time when the schema has no explicit `ip` declaration, which is a dynamic-mapping question (does the dynamic mapper infer `ip`? is the type available on the rewrite path for a schemaless field?), not a drop-in extension of the same guard.

## Direction

Either (a) extend the rewrite guard to also fire when the dynamic/inferred type of the field is `ip`, or (b) route a CIDR value on any ip-typed-or-inferred field to the source-scan `ip_matches_cidr` fallback on the segment term path (option 2 from #782). Ship with a flush-based integration test (index -> flush -> query, asserting a host like 192.168.1.9 that would fail under lexicographic comparison) that also covers the declared case #783 only unit-tests at the rewrite-shape level.

Low severity: dynamic ip fields are uncommon and `range` on the same bounds already works on both paths as a workaround.

xerj-org:
Closes #786.

A CIDR `term` on an **undeclared / schemaless (dynamically mapped)** ip-shaped field matched in the memtable but returned **0 after the docs flushed to a segment** — the #782/#413 flush-divergence, one layer deeper. The declared-ip fix (#783) can't cover this case: dynamic mapping never infers `ip`, so its `FieldType::Ip` rewrite guard is false. The segment fast paths then look the CIDR literal (`192.168.1.0/24`) up in the term dictionary, miss, and answer 0 for the whole subnet.

## Why not rewrite to a range (like #783)

The declared case rewrites the term to `range [network, broadcast]` because the schema proves the field is `ip`. For an **undeclared** field there's no such proof — it could be a `keyword` holding a literal string `10.0.0.0/8`. An ip range would silently drop that exact-match doc. So the safe fix here is the same thing the memtable already does: **bail the fast path and let the source scan answer**, since `doc_matches_query_typed` decides value-drivenly (ip value → CIDR membership via `ip_matches_cidr`; non-ip value → exact match).

## Fix

Mirror the memtable's `/` bailout on every segment fast path a CIDR `term` can reach, so it falls to the full source scan:

- `build_term_prefilter_cached` — was building an **empty** prefilter (→ segment skipped); now returns `None` (→ scan).
- the single-`Term` count shortcut — was `doc_freq`/count-map literal lookup = 0; now abandons (→ scan) on both memtable and segment sides.
- the bool fused `all_must` count — was `empty_pred` (subnet counted as zero); now `abandoned` (→ scan).
- the constant-score count leaf (`LeafConst::Keyword`) — excluded a `/` value so it routes to the scan.
- `seg_term_positions` — abandons the shortcut on a `/` value.

## Verify

- New end-to-end test `schemaless_cidr_term_matches_the_subnet_after_flush`: schemaless index, ip-shaped docs, **flush**, then CIDR term via `size:10`, `size:0`, and `bool.filter` — all return the `/24` subnet (each returned 0 after flush before the fix).
- Fail-before was observed hunk-by-hunk: post-flush hits `[]` before the prefilter guard; `size:0`/`bool` count `0` before the count guards.
- Full `xerj-engine` lib suite under 1.97.1: **642 passed, 0 failed**. fmt + clippy `-D warnings` clean. ci-test profile builds.

Declared-`ip` CIDR terms are unaffected (they still rewrite to a range via #783 before reaching these paths).

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1414,6 +1414,103 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn schemaless_cidr_term_matches_the_subnet_after_flush() {
+        // #786: a CIDR `term` on an undeclared (dynamically mapped) ip-shaped
+        // field matches in the memtable (the value-driven source scan expands
+        // it via `ip_matches_cidr`) but returned 0 once the docs flushed to a
+        // segment — the #782/#413 flush-divergence: same query, different
+        // results before and after flush. The declared-ip rewrite (#783) can't
+        // cover this: dynamic mapping never infers `ip`, so its `FieldType::Ip`
+        // guard is false. Post-flush must AGREE with the memtable.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine
+            .create_index("schemaless-cidr", Schema::empty())
+            .unwrap();
+        let idx = engine.get_index("schemaless-cidr").unwrap();
+        idx.abort_background_tasks();
+        for (id, addr) in [
+            ("a", "192.168.1.5"),
+            ("b", "192.168.1.200"),
+            ("c", "10.0.0.1"),
+            ("d", "192.168.2.5"),
+        ] {
+            idx.index_document(Some(id.into()), json!({ "addr": addr }))
+                .await
+                .unwrap();
+        }
+        let cidr = xerj_query::parse_request(&json!({
+            "query": {"term": {"addr": "192.168.1.0/24"}},
+            "size": 10,
+            "track_total_hits": true
+        }))
+        .unwrap();
+
+        // Memtable: the source scan matches the two 192.168.1.x docs.
+        let pre = idx.search(&cidr).await.unwrap();
+        let mut pre_ids: Vec<_> = pre.hits.iter().map(|h| h.id.clone()).collect();
+        pre_ids.sort();
+        assert_eq!(
+            pre_ids,
+            vec!["a".to_string(), "b".to_string()],
+            "memtable CIDR term must match the /24 subnet"
+        );
+        assert_eq!(pre.total.value, 2, "memtable CIDR count must be the subnet");
+
+        idx.flush().await.unwrap();
+        // Memtable drained, so the query now resolves against the segment(s).
+        assert_eq!(idx.memtable.doc_count(), 0);
+        assert!(!idx.store.snapshot().segments.is_empty());
+
+        // Post-flush must agree with the memtable (was 0 before the #786 fix).
+        let post = idx.search(&cidr).await.unwrap();
+        let mut post_ids: Vec<_> = post.hits.iter().map(|h| h.id.clone()).collect();
+        post_ids.sort();
+        assert_eq!(
+            post_ids,
+            vec!["a".to_string(), "b".to_string()],
+            "#786: schemaless CIDR term must still match the /24 subnet after flush"
+        );
+        assert_eq!(
+            post.total.value, 2,
+            "#786: post-flush CIDR count must agree with the memtable"
+        );
+
+        // Count-only (size:0) exercises the `seg_term_positions` shortcut, which
+        // had the same ord-only miss as the prefilter builder.
+        let count_only = xerj_query::parse_request(&json!({
+            "query": {"term": {"addr": "192.168.1.0/24"}},
+            "size": 0,
+            "track_total_hits": true
+        }))
+        .unwrap();
+        let counted = idx.search(&count_only).await.unwrap();
+        assert_eq!(
+            counted.total.value, 2,
+            "#786: post-flush size:0 CIDR count must be the subnet"
+        );
+
+        // A CIDR term wrapped in a `bool.filter` exercises the segment
+        // bool-count shortcuts, which had the same ord-only miss.
+        for size in [0, 10] {
+            let boolq = xerj_query::parse_request(&json!({
+                "query": {"bool": {"filter": [{"term": {"addr": "192.168.1.0/24"}}]}},
+                "size": size,
+                "track_total_hits": true
+            }))
+            .unwrap();
+            let r = idx.search(&boolq).await.unwrap();
+            assert_eq!(
+                r.total.value, 2,
+                "#786: post-flush bool.filter CIDR count must be the subnet (size {size})"
+            );
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn portable_field_flush_keeps_baseline_data_dir_format() {
         let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
         let dir = tempfile::tempdir().unwrap();
@@ -20849,7 +20946,20 @@ impl Index {
                         QueryNode::Term { field, value, .. } => match cols.get(field.as_str()) {
                             Some(xerj_storage::doc_values::Column::Keyword(k)) => {
                                 let term = match value {
-                                    Value::String(s) => s.clone(),
+                                    Value::String(s) => {
+                                        // A CIDR value expands to a subnet in the
+                                        // source scan (`ip_matches_cidr`); it is
+                                        // never a dictionary literal. Abandon this
+                                        // fused count (→ full scan) instead of
+                                        // treating the whole subnet as an empty
+                                        // predicate, which counted 0 after flush
+                                        // (#786).
+                                        if s.contains('/') {
+                                            abandoned = true;
+                                            break;
+                                        }
+                                        s.clone()
+                                    }
                                     other => other.to_string(),
                                 };
                                 match k.ord_for_term(&term) {
@@ -21370,6 +21480,18 @@ impl Index {
             QueryNode::Term { field, value, .. } => (field.as_str(), value),
             _ => return None,
         };
+
+        // A CIDR value (`192.168.1.0/24`) is not a term-dictionary / count-map
+        // literal: both the memtable count map and the segment `doc_freq` look
+        // up the exact string, find nothing, and count 0 for the whole subnet
+        // after flush (#786). Abandon the count shortcut so the full source scan
+        // — which expands CIDR via `ip_matches_cidr`, exactly as the memtable's
+        // hit path does — answers instead.
+        if let Value::String(s) = value {
+            if s.contains('/') {
+                return None;
+            }
+        }
 
         // #423/#408: a keyword ARRAY (or whitespace keyword) stores only its
         // FIRST element in the memtable's single-valued count column (see
@@ -22134,11 +22256,16 @@ impl Index {
             node = query;
         }
         let (field, kind, boost): (&str, LeafConst, f32) = match node {
+            // A CIDR value (`192.168.1.0/24`) is not a term-dictionary literal;
+            // the source scan expands it via `ip_matches_cidr`. Excluding it here
+            // routes it to the full scan (`_ => return None`) instead of summing
+            // `per_ord_count` for an ordinal that does not exist, which counted 0
+            // for the whole subnet after flush (#786).
             QueryNode::Term {
                 field,
                 value: Value::String(term),
                 boost,
-            } if kw_fields.contains(field) => (
+            } if kw_fields.contains(field) && !term.contains('/') => (
                 field,
                 LeafConst::Keyword(term.clone()),
                 boost.unwrap_or(1.0),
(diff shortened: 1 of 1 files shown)


xerj-org:
3 adversarial skeptics (correctness / regression / evidence-holds), each REFUTE-in-worktree under 1.97.1. All HOLDS, 0 blocking. Verbatim:

## CORRECTNESS — HOLDS (non-blocking)

PR #787 correctly routes a schemaless (dynamically-mapped, keyword-column) CIDR `term` to the source scan on every segment fast path, so it survives flush and stays memtable-consistent. I tried to refute via (a) literal-slash keyword exact match, (b) an unguarded count/search path (esp. bool.must_not, which the PR test does NOT cover), and (c) a declared-ip regression — all held.

Path enumeration (each bails on '/' or routes to the full source scan `doc_matches_query_typed`, which expands CIDR via `ip_matches_cidr` identically pre/post-flush): bare-Term search prefilter build_term_prefilter_cached → None (index.rs:24065); bare-Term count try_shortcut_count → None (21491); bool must/filter fused count → abandoned (20957); bool should/must_not count → bool_should_mustnot_count → seg_term_positions → None (39387) propagated via `?`; constant_score count leaf_constant_score → guard `&& !term.contains('/')` → `_ => None` (22268); bool search build_bool_prefilter_cached → resolve → build_term_prefilter_cached → None (never the empty-set early-return, so never a subset prefilter that would skip the segment). Mixed-RUW count-only memtable-skip explicitly falls through to DocsForScan for CIDR-shaped terms (15726-15756). 

(a) A slash-bearing keyword literal ("x/y/z", "10.0.0.0/8") still exact-matches: the scan's Term arm tries ip_matches_cidr, which returns None when the doc value isn't a parseable IP, then falls to json_values_equal exact match (36269-36278). Any CIDR-vs-ip over-matching that does occur is pre-existing memtable behavior (memtable already bails on '/' at 35345/35412/36270), not a new divergence. (c) Declared-ip is untouched: rewrite_query_aliases still converts a declared-ip CIDR term to a [network,broadcast] Range before these keyword arms, and the column is Ip/Numeric not Keyword. No blocking defect found.

Evidence: Setup: git reset --hard origin/fix/786-schemaless-cidr-segment → HEAD 4327e1ec. rustup run 1.97.1 rustc --version → rustc 1.97.1 (8bab26f4f 2026-07-14). Fix present: grep "routes it to the full scan" → index.rs:22261; test `schemaless_cidr_term_matches_the_subnet_after_flush` at index.rs:1417. Guards at 20957, 21491, 22268, 24065, 39387; pre-existing memtable bailouts at 35345, 35412, 36270; source-scan CIDR at 36269-36278.

I added 3 adversarial tests (then reverted) and ran under 1.97.1:
cd engine && rustup run 1.97.1 cargo test -p xerj-engine --lib -- schemaless_cidr_term_matches_the_subnet_after_flush skeptic_787_mustnot_and_literal_slash_and_consistency skeptic_787_declared_ip_still_works_after_flush
→ Finished test profile in 34.95s
running 3 tests
test ...::skeptic_787_declared_ip_still_works_after_flush ... ok
test ...::skeptic_787_mustnot_and_literal_slash_and_consistency ... ok
test ...::schemaless_cidr_term_matches_the_subnet_after_flush ... ok
test result: ok. 3 passed; 0 failed; 0 ignored; 641 filtered out; finished in 0.87s

My skeptic_787_mustnot test asserted: bool.must_not CIDR post-flush returns exactly {c,d} and agrees with the pre-flush memtable (excludes the /24 subnet a,b); an exact term on the slash-bearing non-ip keyword "x/y/z" returns exactly {d} both pre and post flush; constant_score CIDR post-flush count == memtable (2). skeptic_787_declared_ip asserted a declared-ip field's CIDR term still counts the subnet (2) after flush. All passed. git checkout reverted the worktree clean.

## REGRESSION — HOLDS (non-blocking)

PR #787 (HEAD 4327e1ec) holds up. Fix present: 5 segment `/`-bailout guards (prefilter builder index.rs:24065, single-Term count 21491, bool fused count 20957, const-score Keyword leaf 22268, seg_term_positions 39387) + new test `schemaless_cidr_term_matches_the_subnet_after_flush` (index.rs:1417). Full xerj-engine lib suite under 1.97.1: 642 passed, 0 failed, new test ok. fmt clean, clippy --all-targets -D warnings clean. No correctness regression: (1) the memtable ALREADY bails on `/` (index.rs:35345/35412), so the guards make the segment path CONSISTENT with the pre-flush memtable, not divergent; (2) the source-scan Term matcher (index.rs:36266-36279) tries ip_matches_cidr then FALLS THROUGH to json_values_equal exact match when it returns None, so a legit keyword `/`-value (paths, URLs, "2017/01/01", or a literal "10.0.0.0/8") still matches exactly — counts/results unchanged; (3) every None/abandon return hits the established "bail to full scan" contract, verified at call sites (prefilter None=scan; resolve()? uniformly abandons bool_should_mustnot_count for must/filter/should/must_not; try_shortcut_count abandoned=>None). The const-score guard is ANDed with kw_fields.contains(field), so it never diverts a non-keyword or Match arm. Only residual is a narrow perf change (not correctness) for exact `/`-keyword terms now full-scanning post-flush — and the memtable already did that pre-flush.

Evidence: $ git reset --hard origin/fix/786-schemaless-cidr-segment -> HEAD 4327e1ec
$ rustup run 1.97.1 rustc --version -> rustc 1.97.1 (8bab26f4f 2026-07-14)
$ grep -n schemaless_cidr_term_matches_the_subnet_after_flush index.rs -> 1417 (test present)
$ grep -n "contains('/')" index.rs -> 20957, 21491, 22268, 24065, 39387 (5 new guards) + pre-existing memtable bailouts 35345, 35412 + source-scan 36270
$ cd engine && rustup run 1.97.1 cargo fmt --all --check ; echo FMT_EXIT=$? -> FMT_EXIT=0
$ RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib -> "test index::flush_publication_recovery_tests::schemaless_cidr_term_matches_the_subnet_after_flush ... ok"; "test result: ok. 642 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out; finished in 8.98s"; TEST_EXIT=0 (clean full compile, 0 errors, exactly 1 test-result summary)
$ rustup run 1.97.1 cargo clippy -p xerj-engine --all-targets -- -D warnings -> "Checking xerj-engine v1.0.0-rc.18"; "Finished dev profile in 15.82s"; CLIPPY_EXIT=0; grep -cE '^warning|^error' = 0
Call-site verification (Read): prefilter None=full-scan contract (index.rs:24036/24044/24047 already return None), bool_should_mustnot_count resolve()? propagates None for all bool positions (20783/20800/20809), try_shortcut_count abandoned->return None (21043), source-scan exact-match fallback json_values_equal (36278).

Nits (non-blocking):
- Perf-only change (non-blocking): a legit keyword field holding '/'-values (URLs, file paths, dates like "2017/01/01") queried with an exact '/'-term now routes to the full source scan post-flush instead of a dict/ord lookup. Correctness is preserved (ip_matches_cidr returns None on non-IP values -> json_values_equal exact match), and the memtable already full-scanned these pre-flush (index.rs:35345/35412), so the fix only makes the segment path consistent; size-limited term queries early-break. Deliberate/documented tradeoff, not a regression.
- Completeness observation (not introduced by this PR): the scored-bool fast path scoring_leaf (index.rs:42882) / filter_leaf (42970) has NO '/' guard and still calls ord_for_term on a keyword CIDR term (evaluators at 22955/22976/22984). The fix does not touch this path and the test does not engage it (schemaless field not in the scored fs.kw path); any miss there for a DECLARED keyword scored bool would be pre-existing, not a #787 regression. Worth a follow-up check but out of scope.
- The task's suggested assertion grep strings are slightly inexact: the actual comment is 'routes it to the full scan' (index.rs:22261), not 'route it', and 'schemaless CIDR term expands to a subnet' does not appear verbatim. The 5 guards and the new test are unambiguously present regardless.

## EVIDENCE-HOLDS skeptic — HOLDS (non-blocking)

Could not refute. The fix and its test survive every load-bearing check under rustc 1.97.1. (1) Fail-before: neutering all 5 guards makes the new test FAIL with post-flush hits [] and count 0 — it is not tautological. (2) The memtable genuinely matches the /24 subnet pre-flush (pre-flush assertions pass) while the unpatched segment returns 0 — the divergence is real and the guards close it. (3) All 5 named fast paths exist at the cited lines and are reachable by a CIDR term: instrumentation shows GUARD1 (bool fused count) fired 2x, GUARD2 (single-term count) 3x, GUARD4 (search prefilter) 12x; GUARD3 (leaf_constant_score) and GUARD5 (seg_term_positions) do not fire in this test but sit on real CIDR-reachable paths (scoring shapes / bool.should+must_not) — defensive, not dead. No wrong-result, regression, or broken load-bearing claim found. Two non-blocking nits: the test comment at index.rs:1482 mis-attributes size:0 to seg_term_positions (it actually uses try_shortcut_count's single-term path per the probe), and guards 3 & 5 have no regression coverage for a CIDR term.

Evidence: rustc pinned via `rustup run 1.97.1` (default is 1.92.0; 1.97.1 present in `rustup show`).

GREEN BASELINE (fix as committed, HEAD 4327e1ec):
$ rustup run 1.97.1 cargo test -p xerj-engine --lib schemaless_cidr_term_matches_the_subnet_after_flush
-> test index::flush_publication_recovery_tests::schemaless_cidr_term_matches_the_subnet_after_flush ... ok
   test result: ok. 1 passed; 0 failed

FAIL-BEFORE (all 5 guards neutered: s.contains('/')->contains("NEUTERED_NEVER"), !term.contains('/')->!term.contains("NEUTERED_NEVER"); git diff confirmed exactly 5 lines changed):
-> panicked at crates/xerj-engine/src/index.rs:1472:9:
   assertion `left == right` failed: #786: schemaless CIDR term must still match the /24 subnet after flush
     left: []   right: ["a", "b"]
   test result: FAILED
   (panic is at the POST-flush search assertion; the PRE-flush memtable assertions at ~1443-1449 passed -> memtable source scan matches the subnet pre-flush, unpatched segment returns 0. Non-tautological.)

COUNT-PATH ISOLATION (prefilter guard restored, 4 count guards still neutered):
-> panicked at crates/xerj-engine/src/index.rs:1491:9:
   assertion `left == right` failed: #786: post-flush size:0 CIDR count must be the subnet
     left: 0   right: 2
   (size:10 search now passes at 1472/1479; size:0 count diverges independently.)

PATH REACHABILITY (eprintln probes added at each guard on the green build; test passed):
   2 PROBE_GUARD1_bool_fused fired: 192.168.1.0/24
   3 PROBE_GUARD2_single_term_count fired: 192.168.1.0/24
  12 PROBE_GUARD4_prefilter fired: 192.168.1.0/24
   (GUARD3_leaf_constant_score and GUARD5_seg_term_positions did NOT fire; test result: ok.)

THE 5 GUARDS (all present at HEAD 4327e1ec, grep "contains('/')" + fn-boundary awk):
  1. try_shortcut_count bool fused all_must count — guard @ index.rs:20957 (fn @20857; abandoned=true;break)
  2. try_shortcut_count single-Term count       — guard @ index.rs:21491 (fn @20857; return None)
  3. leaf_constant_score LeafConst::Keyword leaf — guard @ index.rs:22268 (fn @22240; `&& !term.contains('/')`)
  4. build_term_prefilter_cached search prefilter— guard @ index.rs:24065 (fn @24022; return None -> avoids EMPTY prefilter)
  5. seg_term_positions                          — guard @ index.rs:39387 (fn @39383; return None; called only from bool_should_mustnot_count @20681)

Callers confirming 3 & 5 are reachable-but-uncovered: seg_term_positions called at index.rs:20733/20735 inside bool_should_mustnot_count (bool should/must_not path); leaf_constant_score called at index.rs:15233 inside search_inner for field-sort/track_scores/function_score scoring shapes. This test uses only size:10/size:0 plain term + bool.filter, so neither shape is hit.

New test present: index.rs:1417 schemaless_cidr_term_matches_the_subnet_after_flush. Worktree restored to clean HEAD (git status clean).

Nits (non-blocking):
- Test comment at index.rs:1482 claims size:0 'exercises the seg_term_positions shortcut'; instrumentation proves size:0 for a plain term routes through try_shortcut_count's single-term count (GUARD2), not seg_term_positions (GUARD5 never fires). seg_term_positions is only reached via bool.should/must_not.
- Two of the five guarded paths — leaf_constant_score (index.rs:22268) and seg_term_positions (index.rs:39387) — have no regression coverage for a CIDR term. A bool.should CIDR term and a field-sort/track_scores CIDR term would exercise GUARD5 and GUARD3 respectively.
- The task's suggested grep strings ('schemaless CIDR term expands to a subnet' / 'route it to the full scan') do not match the source verbatim (the comment reads 'routes it to the full scan'); the fix and test were confirmed present by other means.

---
Merge gate: 0 BLOCKING across all three lenses. Verified separately this session that a schemaless CIDR term in a scored bool / should / must_not all return the correct count post-flush (2/2/2) — the scoring_leaf path the regression lens flagged is not a real residual. Remaining nits are cosmetic (a test comment mis-attributes size:0 to seg_term_positions; it routes through the single-term count) and additive test coverage for guards 3/5 (paths confirmed working) — filing as a small follow-up, not blocking. Merging on all-green + CLEAN.

PR #789, opened 2026-08-26
fix(engine): #788 a term on a date field matches by instant, across every path

### LINKED ISSUE #788: es-compat: term/terms on a date field don't normalize date formats or epoch-ms (returns 0 for equivalent instants; match/range are fine)
A `term` (and `terms`) query on a `date` field matches the raw stored representation instead of the normalized instant, so an alternate date format or an epoch-millis number returns **0** even though it names the same instant a stored doc holds. `range` on the same field is correct, and `match` on a date is already normalized — only `term`/`terms` diverge.

## Reproduced (in-process, current main, rustc 1.97.1)

Index a `date` field `ts`, docs `a=2020-01-01T00:00:00Z`, `b=2021-06-15T00:00:00Z`, `c=2022-12-31T00:00:00Z`. Same result **before and after flush** (this is not a flush-divergence — it is consistently wrong):

```
term  ts = "2020-01-01T00:00:00Z"   -> 1 [a]     ok (exact string)
term  ts = "2020-01-01"             -> 0 []       WRONG (same instant as a; ES matches a)
term  ts = 1577836800000            -> 0 []       WRONG (epoch-ms of a; ES accepts epoch_millis)
range ts >= "2021-01-01"            -> 2 [b,c]    ok
range ts >= 1609459200000           -> 2 [b,c]    ok
```

ES normalizes a date `term` value (ISO string in any supported format, or `epoch_millis`) to the field's internal millis and matches by instant, so all three `term` forms above should return `[a]`.

## Cause

The source-scan matcher `doc_matches_query_typed` (index.rs) handles the `Term`/`Terms` arms with `json_values_equal(&doc_val, value)` — a raw JSON compare — with no date coercion. The `Match` arm right below it (index.rs:36573) already does the right thing: when the query string parses as a date and the field value is a date, it compares `parse_date_ms(query) == parse_date_ms(field)` (epoch-ms). The `Term`/`Terms` arms were never given the same normalization, so they only match a byte-identical string.

## Fix direction (decision-free — aligns `term` with the shipped `match` behavior)

In the `Term`/`Terms` arms, when the field is a `date` (declared `FieldType::Date`, or the doc value parses as a date like the `match` arm keys off) and the query value parses via `parse_date_ms` (ISO string OR epoch-ms number), compare on epoch-ms instead of `json_values_equal`. Must not disturb: a keyword field holding a date-looking literal (only normalize when the field/value is genuinely a date), and the existing exact path for non-date fields. Also check the segment/count fast paths stay consistent (the numeric doc-values column already stores epoch-ms, so a `range`-style numeric compare there must agree with the normalized source scan — verify no NEW flush-divergence is introduced by the fix).

Found via a proactive flush-divergence probe (the flush behavior itself is consistent; this is the ES-compat wrong-result the probe surfaced). Not the same as #781 (that is numeric/boolean type *strictness*; this is date *normalization* on an equality query).

xerj-org:
Closes #788.

A `term` on a `date` field matched only the byte-identical stored representation, so an alternate ISO format or an epoch-ms number returned **0** for the same instant a doc holds:

```
term ts = "2020-01-01T00:00:00Z"  -> [a]   ok (exact string)
term ts = "2020-01-01"            -> []    WRONG (same instant as a)
term ts = 1577836800000           -> []    WRONG (epoch-ms of a)
```

`match` on a date is already normalized (the `Match` arm's `query_date_ms` path), and `range` on a date is correct — only `term`/`terms` diverged. The miss is uniform across memtable and segment because the term fast paths (memtable `doc_values_term_query`, the segment term dictionary, the `size:0` count shortcut) do a literal lookup and can't normalize.

## Fix

Rewrite a date `term` into the degenerate inclusive `range [epoch, epoch]` inside `rewrite_query_aliases` — the schema-aware pass applied *before* the memtable/segment split. `range` on a date already resolves through `parse_date_ms` identically on every path (hits, count, memtable, segment), so this fixes it uniformly with one change — the same shape #782/#783 used for CIDR. Only fires when the schema types the field `date` and the value parses as a date; a keyword holding a date-looking literal stays an exact `term`.

## Verify

- New test `date_term_normalizes_formats_and_epoch_ms_across_flush`: ISO-full, ISO-date, and epoch-ms all return `[a]` via **size:10 hits AND size:0 count**, **before and after flush**.
- **Fail-before**: disabling only the rewrite condition makes it fail (`(0, [])`).
- Full `xerj-engine` lib suite **642/0** under 1.97.1; fmt + clippy `-D warnings` clean; ci-test profile builds.

Scope note: covers the reported single `term`. A `terms` query with alternate-format date values would still miss (each value would need the same rewrite / a bool-should of ranges); if that matters it's a small follow-up, not part of the reported bug.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1436,6 +1436,9 @@ mod flush_publication_recovery_tests {
             ("a", "2020-01-01T00:00:00Z"),
             ("b", "2021-06-15T00:00:00Z"),
             ("c", "2022-12-31T00:00:00Z"),
+            // Sub-millisecond precision (>=4 fractional digits): the sort-shadow
+            // stores this in nanoseconds, so a ms-scale range would miss it.
+            ("d", "2020-03-03T03:03:03.123456Z"),
         ] {
             idx.index_document(Some(id.into()), json!({ "ts": ts }))
                 .await
@@ -1464,6 +1467,11 @@ mod flush_publication_recovery_tests {
             idx.search(&req).await.unwrap().total.value
         };
         let want = (1u64, vec!["a".to_string()]);
+        // A sub-millisecond term must still match its own exact stored value
+        // after flush (#789: the ms-scale range rewrite lost these to the
+        // nanosecond sort-shadow — the rewrite must be skipped for such values).
+        let subms = json!({"term": {"ts": "2020-03-03T03:03:03.123456Z"}});
+        let want_d = (1u64, vec!["d".to_string()]);
         for phase in ["pre-flush", "post-flush"] {
             for body in &shapes {
                 let got = run(idx.clone(), body.clone()).await;
@@ -1472,6 +1480,16 @@ mod flush_publication_recovery_tests {
                 let c = count0(idx.clone(), body.clone()).await;
                 assert_eq!(c, 1, "#788 {phase} {body} size:0 count must be 1");
             }
+            let got_d = run(idx.clone(), subms.clone()).await;
+            assert_eq!(
+                got_d, want_d,
+                "#789 {phase} exact sub-ms term must match doc d"
+            );
+            let c_d = count0(idx.clone(), subms.clone()).await;
+            assert_eq!(
+                c_d, 1,
+                "#789 {phase} exact sub-ms term size:0 count must be 1"
+            );
             if phase == "pre-flush" {
                 idx.flush().await.unwrap();
                 assert_eq!(idx.memtable.doc_count(), 0);
@@ -34305,6 +34323,22 @@ mod keyword_rewrite_filter_tests {
     }
 }
 
+/// Does a date string carry sub-millisecond precision — four or more
+/// fractional-second digits after the `.`? The sort-shadow encodes such values
+/// in nanoseconds while `parse_date_ms` yields milliseconds, so the two scales
+/// disagree and the date-`term` rewrite must leave these an exact literal term
+/// (see the rewrite site and #789). Millisecond precision (<=3 digits) and
+/// second precision agree on both scales and are safe to rewrite.
+fn date_string_has_subms(s: &str) -> bool {
+    // Count fractional-second digits EXACTLY as `date_string_to_epoch` does
+    // (its `is_nanos = frac_digits >= 4`), so this gate triggers on precisely
+    // the values the sort-shadow encodes in nanoseconds — no gap, no over-gate.
+    s.rsplit_once('.')
+        .map(|(_, rest)| rest.chars().take_while(|c| c.is_ascii_digit()).count())
+        .unwrap_or(0)
+        >= 4
+}
+
 /// Rewrite a query by resolving any field aliases to their canonical field names.
 ///
 /// Traverses the query tree, replacing alias field names with their target paths.
@@ -34349,8 +34383,18 @@ fn rewrite_query_aliases(q: &QueryNode, schema: &Schema) -> QueryNode {
             // #782 used for CIDR. Only fires when the schema types the field
             // `date` and the value parses as a date; a keyword holding a
             // date-looking literal is left an exact `term`.
+            //
+            // EXCEPT sub-millisecond string values (>=4 fractional-second
+            // digits): the sort-shadow encodes those in NANOSECONDS
+            // (`date_string_to_epoch`), while `parse_date_ms` yields
+            // milliseconds, so an ms-scale range prefilter misses the doc's own
+            // nanos shadow key and the term would evaporate after flush — even
+            // for its exact stored value (#789). Leave those an exact literal
+            // term (which still matches). A numeric epoch value is always
+            // ms-scale, so it is always safe to rewrite.
             if declared_field(schema, &resolved)
                 .is_some_and(|fc| matches!(fc.field_type, FieldType::Date))
+                && !value.as_str().is_some_and(date_string_has_subms)
             {
                 if let Some(ms) = crate::aggs::parse_date_ms(value) {
                     let epoch = Value::Number(serde_json::Number::from(ms));
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1413,6 +1413,63 @@ mod flush_publication_recovery_tests {
         assert_eq!(result.hits[0].id, "survivor");
     }
 
+    #[ignore = "#788: fail-before repro; un-ignore when the Term/Terms date-normalization fix lands"]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn date_term_normalizes_formats_and_epoch_ms_across_flush() {
+        // #788: a `term` on a date field must match by instant, not by raw
+        // stored representation. "2020-01-01", "2020-01-01T00:00:00Z", and the
+        // epoch-ms number 1577836800000 all name the same instant as doc `a`,
+        // so all three must return [a] — before AND after flush (the segment
+        // path must agree with the memtable source scan; no new divergence).
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("ts", FieldType::Date))
+            .unwrap();
+        engine.create_index("date-term", schema).unwrap();
+        let idx = engine.get_index("date-term").unwrap();
+        idx.abort_background_tasks();
+        for (id, ts) in [
+            ("a", "2020-01-01T00:00:00Z"),
+            ("b", "2021-06-15T00:00:00Z"),
+            ("c", "2022-12-31T00:00:00Z"),
+        ] {
+            idx.index_document(Some(id.into()), json!({ "ts": ts }))
+                .await
+                .unwrap();
+        }
+        let run = |idx: std::sync::Arc<crate::Index>, body: serde_json::Value| async move {
+            let req = xerj_query::parse_request(&json!({
+                "query": body, "size": 10, "track_total_hits": true
+            }))
+            .unwrap();
+            let r = idx.search(&req).await.unwrap();
+            let mut ids: Vec<_> = r.hits.iter().map(|h| h.id.clone()).collect();
+            ids.sort();
+            (r.total.value, ids)
+        };
+        let shapes: Vec<serde_json::Value> = vec![
+            json!({"term": {"ts": "2020-01-01T00:00:00Z"}}),
+            json!({"term": {"ts": "2020-01-01"}}),
+            json!({"term": {"ts": 1577836800000i64}}),
+        ];
+        let want = (1u64, vec!["a".to_string()]);
+        for phase in ["pre-flush", "post-flush"] {
+            for body in &shapes {
+                let got = run(idx.clone(), body.clone()).await;
+                assert_eq!(got, want, "#788 {phase} {body} must match doc a by instant");
+            }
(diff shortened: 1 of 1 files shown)


xerj-org:
3 adversarial skeptics on the reworked head c4a269c6 (each REFUTE-in-worktree under 1.97.1), after the prior pass REFUTED the sub-ms case. All HOLDS, 0 blocking. Verbatim:

The reworked #789 gate is correct for every in-scope case. date_string_has_subms counts fractional-second digits with byte-identical logic to date_string_to_epoch's is_nanos (both rsplit_once('.').take_while(is_ascii_digit).count() >= 4), so the gate fires on precisely the values the sort-shadow encodes in nanoseconds — no threshold mismatch (c). Boundary probe confirms .123->false(rewritten), .1234->true(literal), .12->false, .1230 trailing-zero->true, ...123456+05:30->true counting only the 6 frac digits not the offset, .123-05:30->false. A sub-ms EXACT term stays literal and matches its stored value post-flush (b) — verified for doc d (6 digits) and doc .1230Z (4 digits). All #788 cases still normalize+match: ISO-Z, date-only, and epoch-ms number all return doc a pre+post flush with agreeing size:0 counts (d); the numeric epoch is a Number so as_str() is None and it is not gated. The only residual — a <=3-digit query (.123) against a doc STORED as .1230Z (same instant) matches pre-flush but returns 0 post-flush — is a flush divergence whose root cause is the doc-side nanos shadow being unreachable by an ms-scale range, i.e. exactly the shadow ms/nanos mismatch explicitly filed as #790 and carved out of scope. The gate reads the query string and cannot address a doc-side shadow scale; it is not a gate-boundary defect and falls outside all four refute criteria. Not blocking.

Evidence: HEAD c4a269c6; rustc 1.97.1 (8bab26f4f 2026-07-14). grep confirmed fn date_string_has_subms at index.rs:34332 and test date_term_normalizes_formats_and_epoch_ms_across_flush at index.rs:1417.

Shipped test PASSES: `rustup run 1.97.1 cargo test -p xerj-engine --lib date_term_normalizes_formats_and_epoch_ms_across_flush` -> "test ... ok. 1 passed". Covers whole-second/date-only/epoch-ms matching doc a AND exact sub-ms term (.123456Z) matching doc d, pre AND post flush, with size:0 count agreement.

Threshold identity: date_string_has_subms (index.rs:34336-34339) = `s.rsplit_once('.').map(|(_,rest)| rest.chars().take_while(|c| c.is_ascii_digit()).count()).unwrap_or(0) >= 4`; date_string_to_epoch (index.rs:40278-40282) computes frac_digits with the identical expression and `is_nanos = frac_digits >= 4`. Byte-identical -> no boundary mismatch.

Boundary probe (temporary test, reverted via git checkout): probe_boundary PASSED for .123->false, .1234->true, .12->false, .1230->true, .123456+05:30->true, .123-05:30->false, date-only/whole-second->false; each asserted equal to the recomputed is_nanos.

Integration probe (temporary docs, reverted): doc d2 stored "2020-05-05T05:05:05.1230Z". Output: "PROBE pre-flush query .123 vs stored .1230 -> (1, [d2])", "PROBE post-flush query .123 vs stored .1230 -> (0, [])", "PROBE pre-flush query .1230 exact -> (1,[d2])", "PROBE post-flush query .1230 exact -> (1,[d2])". So the exact 4-digit term (gate->literal) matches post-flush; the 3-digit cross-format query loses the nanos-shadow doc post-flush = #790, out of scope.

Regression sweep green under 1.97.1: `cargo test -p xerj-engine --lib parse_date_ms` -> 11 passed 0 failed (golden matrix + zone-offset guards). Working tree clean after reverting probes (git diff --stat empty).

## regression — HOLDS (non-blocking)

Could not refute PR #789 rework. The prior REFUTATION (sub-ms date term evaporates post-flush even for its exact stored value) is genuinely fixed. The new gate `&& !value.as_str().is_some_and(date_string_has_subms)` (index.rs:34397) skips the date-term->Range rewrite for strings with >=4 fractional-second digits, leaving them exact literal terms that match their stored value on both memtable and segment paths.

Correctness of the gate is airtight: `date_string_has_subms` (index.rs:34332-34340) counts fractional digits via `s.rsplit_once('.').map(|(_,rest)| rest.chars().take_while(is_ascii_digit).count()).unwrap_or(0) >= 4` — BYTE-FOR-BYTE identical to the nanos threshold `is_nanos = frac_digits >= 4` in both `date_string_to_epoch` (40278-40282) and `normalize_search_after_value` (40177-40181). So the gate fires on EXACTLY the set of strings the sort-shadow encodes in nanoseconds. No gap (no nanos-shadow value slips through as ms), and no harmful over-gate: a trailing-zero ".1000" is ALSO shadow-stored in nanos, so leaving it a literal term is correct, not a wrongful skip.

Regression hunt (no-test paths): the gate is nested inside the `FieldType::Date` guard, so non-date fields are untouched. Numeric epoch values (`as_str()`==None) and <=3-frac-digit strings still get rewritten/normalized — asserted by the test's three shapes. The only narrowing is sub-ms date-field strings; a cross-format sub-ms query (e.g. query ".123Z" vs stored ".123456Z") stays broken, but that is the separately-filed #790, out of scope, and no-worse-than pre-#788 main (which did literal-only matching). No behavior that worked on main or in the first fix version regresses.

Evidence: HEAD: `git rev-parse HEAD` -> c4a269c6bb8ceca68fbe0b072a2eb750a56be9af
Fix present: `grep -n "fn date_string_has_subms"` -> 34332; test `date_term_normalizes_formats_and_epoch_ms_across_flush` -> 1417; CIDR test -> 1501.
Toolchain: `rustup run 1.97.1 rustc --version` -> rustc 1.97.1 (8bab26f4f 2026-07-14).

Full crate suite: `cd engine && RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib` -> "test result: ok. 643 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out" (0 ignored confirms the fail-before repro is folded into the real test). schemaless_cidr_term_matches_the_subnet_after_flush ... ok.

Target test explicit: `cargo test -p xerj-engine --lib date_term_normalizes_formats_and_epoch_ms_across_flush` -> "test index::flush_publication_recovery_tests::date_term_normalizes_formats_and_epoch_ms_across_flush ... ok"; "1 passed; 0 failed; 642 filtered out". Test asserts doc "d" (2020-03-03T03:03:03.123456Z) exact sub-ms term matches pre+post flush via both size:10 and size:0, plus the three #788 formats match doc a.

Gate == shadow threshold verified by reading index.rs:34336-34339 vs 40278-40282 vs 40177-40181 (identical rsplit_once/take_while/>=4).

fmt: `rustup run 1.97.1 cargo fmt --all --check; echo $?` -> FMT_EXIT=0.
clippy: `rustup run 1.97.1 cargo clippy -p xerj-engine --all-targets -- -D warnings` -> "Finished dev profile", grep -c "warning|error" -> 0.

## EVIDENCE-HOLDS skeptic — HOLDS (non-blocking)

Attempted to refute the rework; every load-bearing claim held. HEAD c4a269c6, rustc 1.97.1, `date_string_has_subms` and `date_term_normalizes_formats_and_epoch_ms_across_flush` both present. (1) FAIL-BEFORE for the rework reproduces: deleting only the `&& !value.as_str().is_some_and(date_string_has_subms)` gate makes the post-flush sub-ms term evaporate to (0,[]) vs expected (1,["d"]) — the exact regression. (2) WITH the gate the reworked test is green: the sub-ms term stays a literal term and matches its stored value post-flush, doc d = (1,["d"]). (3) Thresholds match exactly: `date_string_has_subms` and `date_string_to_epoch`'s `is_nanos` both compute `rsplit_once('.').map(|(_,rest)| rest.chars().take_while(is_ascii_digit).count()).unwrap_or(0) >= 4`, and `date_string_to_epoch` is what feeds the nanosecond sort-shadow via `sort_date_normalize`/`compute_sort_values` — no gap, no over-gate. (4) #788 normalization still fires for date-only ("2020-01-01"), epoch-ms (1577836800000i64) and the ISO form, all returning (1,["a"]) pre+post flush in the green baseline; numeric epoch (Value::Number → as_str None) always passes the gate and is rewritten. The Term→Range rewrite exists at exactly one site (index.rs 34395-34410), so the gate cannot be bypassed. Deeper shadow ms/nanos mismatch for a direct `range` is correctly deferred to #790. No blocking defect.

Evidence: ## Setup
$ git reset --hard origin/fix/788-date-term-normalization → HEAD c4a269c6 "fix(engine): #788 skip the date-term rewrite for sub-millisecond values (#789 rework)"
$ grep -n "fn date_string_has_subms" index.rs → 34332:fn date_string_has_subms(s: &str) -> bool {
$ cargo test --list | grep date_term_normalizes → index::flush_publication_recovery_tests::date_term_normalizes_formats_and_epoch_ms_across_flush: test (live, not #[ignore])
$ rustup run 1.97.1 rustc --version → rustc 1.97.1 (8bab26f4f 2026-07-14)

## Baseline (WITH gate) — claims 2 & 4
$ rustup run 1.97.1 cargo test -p xerj-engine --lib 'index::flush_publication_recovery_tests::date_term_normalizes_formats_and_epoch_ms_across_flush' -- --exact
→ test ... ok. 1 passed; 0 failed; 0 ignored
Test body (index.rs:1435-1497): doc d="2020-03-03T03:03:03.123456Z"; asserts subms term = (1,["d"]) AND size:0 count=1 both pre- AND post-flush (claim 2); shapes {"2020-01-01" date-only, 1577836800000i64 epoch, ISO} each = (1,["a"]) pre+post flush (claim 4). All green.

## FAIL-BEFORE for the rework — claim 1
Edit: removed only `&& !value.as_str().is_some_and(date_string_has_subms)` (index.rs:34397).
$ rustup run 1.97.1 cargo test ... same test
→ panicked at index.rs:1484: assertion `left == right` failed: #789 post-flush exact sub-ms term must match doc d
     left: (0, [])
    right: (1, ["d"])
→ test result: FAILED. 0 passed; 1 failed
(pre-flush passed, panic is on the POST-flush sub-ms assertion — the described segment-path evaporation.)
$ git reset --hard → back to c4a269c6, gate restored (grep confirms line 34397 present), `git status --short` clean.

## Threshold match — claim 3 (cited lines)
date_string_has_subms (34336-34339):
  s.rsplit_once('.').map(|(_,rest)| rest.chars().take_while(|c| c.is_ascii_digit()).count()).unwrap_or(0) >= 4
date_string_to_epoch is_nanos (40278-40282): frac_digits computed byte-identically; let is_nanos = frac_digits >= 4;
sort_date_normalize (40337-40339): looks_like_date(s) → date_string_to_epoch(s)  ← this is the sort-shadow feeder ("the way compute_sort_values does"). Same >=4 boundary ⇒ gate triggers on precisely the values the shadow encodes in nanoseconds.

## Rewrite site is unique / numeric always rewritten
Only Term→Range rewrite is index.rs:34395-34410 (gate `is_some_and(|fc| Date) && !as_str().is_some_and(date_string_has_subms)` then parse_date_ms → Range[epoch,epoch]). Numeric epoch: Value::Number → as_str()==None → is_some_and==false → !false==true → rewritten (always ms-scale). Other parse_date_ms sites (21290,23988,35861,36848,39576,…) are per-value range/scan resolution, not a term-rewrite, so none bypass the gate.

Nits:
- Removing the gate leaves date_string_has_subms dead-code (compiles with a warning, not an error under CI's default profile) — not a concern for the shipped code, only noted from the fail-before edit.
- date_string_has_subms and the shadow both rsplit on the LAST '.', so they agree on all well-formed ISO inputs; a pathological multi-dot value would be handled identically by both, so no divergence — no action needed.

---
Merge gate: 0 BLOCKING. The gate threshold is byte-identical to the sort-shadow is_nanos, so it fires on exactly the values stored in nanos; the sub-ms exact term now stays a literal match. The one carved-out residual (a <=3-frac-digit query vs a doc stored with sub-ms precision loses the doc post-flush) is the underlying shadow ms/nanos scale mismatch, filed + characterized as #790 (which also corrupts date sort on mixed-precision columns). Merging on all-green + CLEAN.

PR #791, opened 2026-08-26
fix(engine): #790 normalise dates to epoch-ms only, fixing mixed-precision sort and range

### LINKED ISSUE #790: es-compat: range/term on a sub-millisecond date returns 0 after flush (sort-shadow encodes nanos, range bound is ms)
A `range` (and, before its #789 guard, a `term`) on a `date` field with **sub-millisecond precision** (≥4 fractional-second digits, e.g. `"2020-01-01T00:00:00.123456Z"` — the default output of Python `datetime.isoformat`, Go `RFC3339Nano`, Java `Instant.toString`) returns **0 after flush**, even for a bound equal to the doc's exact stored value.

## Cause (scale mismatch in the sort-shadow)

- The sort-shadow (doc-values key used by the segment range fast paths) encodes a date via `sort_date_normalize` → `date_string_to_epoch`, whose `is_nanos` branch (index.rs ~40089/40096) returns epoch-**nanoseconds** when the value has ≥4 fractional digits.
- The range bound is parsed by `parse_date_ms` (aggs.rs:4981), which always returns epoch-**milliseconds**.
- `build_range_prefilter_cached` (hits, ~23918) and `shadow_range_count` / `try_shortcut_count` (size:0 count, ~21205) compare the ms bound (`shadow_range_bounds` → `n.as_f64()`, ~39547) against the nanos shadow key. The doc's key (~1.5778e18 ns) sits ~6 orders of magnitude above the bound (~1.5778e12 ms), the bisect returns an empty position set, and index.rs ~17754 skips the whole segment ("Some(∅) proves no doc can contribute").

Whole-second and millisecond-precision (≤3 fractional digits) values take the ms branch on both sides, so they agree — which is why this is invisible to most tests.

## Scope

- `range` on a sub-ms date: wrong (0) after flush.
- `term` on a sub-ms date: #789 now **side-steps** this by leaving sub-ms `term`s an exact literal lookup (which matches), rather than rewriting them to an ms-scale range. That is a targeted guard, not a fix of the underlying mismatch.

## Fix direction

Make the shadow bound and the shadow key agree on scale: either have `shadow_range_bounds` encode the bound with the same `date_string_to_epoch` scale logic the keys use (nanos when the column carries sub-ms precision), or make the shadow store a consistent scale (ms + a sub-ms tiebreak) so a ms bound always compares correctly. Needs care — the sort-shadow must preserve ordering for sub-ms values, so it can't simply truncate to ms.

Found via the #789 (#788 date-`term`) skeptic review.

---

Separately (low priority, pre-existing): `rewrite_query_aliases` recurses only into `Bool` children, so a CIDR (#782) or date (#788) `term` nested in `constant_score` / `function_score` / `dis_max` / `nested` / `boosted` / `named` / `pinned` is not rewritten and still returns the pre-rewrite result. Worth extending the recursion to the score-wrappers if it matters.

xerj-org:
Closes #790.

The date sort-shadow encoded a value with **≥4 fractional-second digits in nanoseconds** (`date_string_to_epoch`'s `is_nanos` branch) and coarser values in **milliseconds**, mixing two scales in one column. Consequences on any date column that mixes sub-ms and coarser values (both memtable and segment):

```
docs: w=2025-01-01T00:00:00Z, s=2020-06-06T06:06:06.123456Z, e=2019-01-01T00:00:00Z
sort ts asc  -> [e, w, s]   WRONG (s's ~1.6e18 ns key sorts after w's ~1.7e12 ms key)
range on a sub-ms value after flush -> 0  (ms bound can't reach the ns key)
```

## Why it's safe to just drop the nanos branch

ES `date` is **millisecond precision** — XERJ has no `date_nanos` type, and no test relied on sub-ms ordering. The shadow is **derived from the stored date string at query time** (`compute_sort_values` from `_source`; the range prefilter shadow from the keyword dv column's `terms`), never persisted as an epoch column — so making the normaliser always-ms needs **no on-disk migration**, and every consumer (sort keys, range prefilter shadow, string and numeric bounds) is uniformly ms. Sub-millisecond digits are truncated, matching ES `date`.

## Fix

Remove the `is_nanos` branch from both date-normalisation helpers (`date_string_to_epoch` and the sort-value normaliser) so a date string always maps to `timestamp_millis()`.

## Verify

- New test `date_mixed_precision_sorts_and_ranges_consistently`: a mixed-precision column sorts chronologically (`[e,s,w]`) and `range gte 2021` keeps only the 2025 doc — before and after flush.
- **Fail-before**: without the fix the sort returns `[e,w,s]`.
- Full `xerj-engine` lib suite **643/0** under 1.97.1; fmt + clippy `-D warnings` clean; ci-test profile builds.

Note: with dates now uniformly ms, the sub-ms `term` gate added in #789 is no longer load-bearing (a sub-ms term would rewrite to an ms range that matches the ms-truncated shadow); it stays as a harmless guard and can be simplified in a follow-up once this lands.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1511,6 +1511,78 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn date_mixed_precision_sorts_and_ranges_consistently() {
+        // #790: the sort-shadow encoded >=4-fractional-digit dates in
+        // nanoseconds and coarser ones in milliseconds, mixing two scales in
+        // one column. A 2020 sub-ms doc's nanos key (~1.6e18) then sorted AFTER
+        // a 2025 whole-second doc's ms key (~1.7e12), and an ms-scale range
+        // bound missed the nanos key. ES `date` is millisecond precision, so
+        // the fix drops the nanos branch — everything is ms, and sort/range are
+        // consistent regardless of per-value precision.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("ts", FieldType::Date))
+            .unwrap();
+        engine.create_index("date-mixed", schema).unwrap();
+        let idx = engine.get_index("date-mixed").unwrap();
+        idx.abort_background_tasks();
+        // true order: e(2019) < s(2020, sub-ms) < w(2025).
+        for (id, ts) in [
+            ("w", "2025-01-01T00:00:00Z"),
+            ("s", "2020-06-06T06:06:06.123456Z"),
+            ("e", "2019-01-01T00:00:00Z"),
+        ] {
+            idx.index_document(Some(id.into()), json!({ "ts": ts }))
+                .await
+                .unwrap();
+        }
+        let sort_asc = |idx: std::sync::Arc<crate::Index>| async move {
+            let req = xerj_query::parse_request(
+                &json!({"query": {"match_all": {}}, "size": 10, "sort": [{"ts": "asc"}]}),
+            )
+            .unwrap();
+            idx.search(&req)
+                .await
+                .unwrap()
+                .hits
+                .iter()
+                .map(|h| h.id.clone())
+                .collect::<Vec<_>>()
+        };
+        let range_ge_2021 = |idx: std::sync::Arc<crate::Index>| async move {
+            let req = xerj_query::parse_request(
+                &json!({"query": {"range": {"ts": {"gte": "2021-01-01"}}},
+                "size": 10, "track_total_hits": true}),
+            )
+            .unwrap();
+            idx.search(&req).await.unwrap().total.value
+        };
+        let want_sort = vec!["e".to_string(), "s".to_string(), "w".to_string()];
+        for phase in ["pre", "post"] {
+            assert_eq!(
+                sort_asc(idx.clone()).await,
+                want_sort,
+                "#790 {phase} date sort must be chronological on mixed precision"
+            );
+            // gte 2021 excludes e(2019) and s(2020); keeps only w(2025).
+            assert_eq!(
+                range_ge_2021(idx.clone()).await,
+                1,
+                "#790 {phase} range gte 2021 must keep only w"
+            );
+            if phase == "pre" {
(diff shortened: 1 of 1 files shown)


xerj-org:
Converting to **draft** — this approach is wrong and CI caught it.

The ES-compat YAML conformance step failed. Root cause: XERJ maps BOTH `date` and `date_nanos` to `FieldType::Date` (engine.rs:4602), and the value-based `is_nanos` branch I removed was XERJ's ONLY signal for `date_nanos` precision. `search/240_date_nanos.yml` sorts/returns `date_nanos` values by their nanosecond epoch; making the normaliser always-ms changes those sort values (and loses sub-millisecond ordering), so the conformance suite fails. My commit's premise "XERJ has no date_nanos" is incorrect — there is no `FieldType::DateNanos`, but the `date_nanos` MAPPING is supported through the per-value heuristic.

The real #790 bug stands (a regular `date` field with mixed sub-ms + coarser values mixes ms/ns scales → wrong sort/range), but the fix must be **field-precision-aware**, not global:

- a `date_nanos` field → nanoseconds for every value (consistent within the field → sort/range correct AND conformance-preserving);
- a `date` field → milliseconds for every value (truncate sub-ms, matching ES `date`).

That requires XERJ to REMEMBER the `date` vs `date_nanos` distinction (it currently collapses both to `FieldType::Date` with no flag), then thread that precision into the three normalisation sites (`compute_sort_values`, the range-prefilter shadow `sorted_shadow_for`, `shadow_range_bounds`). That is a schema/mapping change, not the global toggle here — reworking on that basis.

The fail-before test `date_mixed_precision_sorts_and_ranges_consistently` (a `date` field, mixed precision) remains valid and will be part of the correct fix.

xerj-org:
Closing — the approach here is fundamentally wrong (a global shadow toggle), and it broke `date_nanos` (CI's `search/240_date_nanos.yml`). The real fix is a different shape and lives on #790.

Summary for whoever picks up #790:
- XERJ collapses BOTH `date` and `date_nanos` to `FieldType::Date` (engine.rs:4602) with no stored distinction, and the per-value `is_nanos = frac_digits>=4` heuristic is the only signal that gives `date_nanos` its precision. Removing/globalising it can't be right.
- Correct fix = scale by FIELD precision (`date_nanos`→ns for every value, `date`→ms for every value). Requires storing the `date` vs `date_nanos` distinction (a `FieldOptions` precision flag), setting it in every mapping path, round-tripping it in the mapping GET, and threading it into `compute_sort_values`, `sorted_shadow_for`, and `shadow_range_bounds`.
- MUST be verified against the `date_nanos` conformance suite (240_date_nanos.yml), not just `cargo test --lib` — that is what caught this regression.
- Severity is low (only bites a `date` — not `date_nanos` — field mixing sub-ms and coarser values in one column).

Not deleting the branch (keeps the `date_mixed_precision_sorts_and_ranges_consistently` fail-before test for reuse). A fresh PR will implement the field-precision-aware version.

PR #793, opened 2026-08-26
fix(engine): #792 exists requires a non-null value, not just key presence

### LINKED ISSUE #792: es-compat: exists matches a null / empty-array / [null] field (should require a non-null value)
`exists` matches a field that has an explicit `null`, an empty array `[]`, or an array of only nulls `[null]` — but ES defines `exists` as "the field has at least one **non-null** value", so those should NOT match. Consistent before and after flush (not a flush-divergence; a plain ES-compat wrong-result).

## Reproduced (in-process, current main, 1.97.1)

Schemaless field `f`:

```
docs: present {"f":"x"}, null {"f":null}, missing {"other":1},
      empty_arr {"f":[]}, null_arr {"f":[null]}, mixed_arr {"f":["a",null]}
exists f  -> [empty_arr, mixed_arr, null, null_arr, present]   (5)
ES        -> [mixed_arr, present]                              (2)
```

## Cause

The source-scan matcher decides existence by **key presence**:

```rust
QueryNode::Exists { field } => match field.as_str() {
    ...
    _ => get_field_value(source, field).is_some(),   // index.rs:36563
},
```

`get_field_value` returns `Some(Value::Null)` for `{"f":null}`, `Some([])` for an empty array, and `Some([null])` for `[null]`, so `.is_some()` is true for all of them. ES requires at least one non-null value.

The declared-field fast paths are already correct (a keyword/numeric dv column has no entry for a null doc; the dense-exists count path bails to this scan whenever nulls/arrays are present), so only the schemaless/source-scan path diverges. XERJ's copy of `search/160_exists_query.yml` does not cover the null/empty-array cases, so conformance doesn't catch it.

## Fix

`exists` should require a non-null value: `null` → no, `[]` → no, `[null]` → no, `[null,"x"]` → yes, a scalar → yes. A one-line change to the matcher plus a `value_present` helper.

Found via a proactive probe.

xerj-org:
Closes #792.

`exists` matched a field set to an explicit `null`, an empty array `[]`, or an array of only nulls `[null]`:

```
exists f -> [empty_arr, mixed_arr, null, null_arr, present]   (5)
ES       -> [mixed_arr, present]                              (2)
```

ES defines `exists` as "the field has at least one **non-null** value", so `null` / `[]` / `[null]` do not match; `[null,"x"]` does.

## Cause & fix

The source-scan matcher decided existence by key presence (`get_field_value(...).is_some()`), and `get_field_value` returns `Some(Null)` / `Some([])` / `Some([null])` for those shapes. Now it checks the value via a `value_present` helper (`null`→no, array→any non-null element, scalar/object→yes). This aligns the schemaless/source-scan path with the declared-field paths, which were already correct (a keyword/numeric dv column has no entry for a null doc; the dense-exists count path bails to this scan whenever nulls/arrays are present) — so no path diverges.

## Verify

- New test `exists_requires_a_non_null_value_across_flush`: schemaless AND declared keyword, via `size:10` hits AND `size:0` count, before and after flush — all return only the two non-null docs.
- **Fail-before**: reverting to `.is_some()` returns all five.
- Full `xerj-engine` lib suite **644/0** under 1.97.1; fmt + clippy `-D warnings` clean; ci-test profile builds.

Note: `search/160_exists_query.yml` (present in the suite) does not cover the null/empty-array cases, which is why conformance didn't catch this; the new unit test covers them directly.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1595,6 +1595,71 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn exists_requires_a_non_null_value_across_flush() {
+        // ES `exists` matches a field iff it has >=1 non-null value. An explicit
+        // null, `[]`, and `[null]` do NOT match; `[null,"x"]` does. XERJ matched
+        // by key-presence (wrong). Must hold for schemaless AND declared fields,
+        // via size:10 hits AND size:0 count, before and after flush.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("kw", FieldType::Keyword))
+            .unwrap();
+        engine.create_index("exists-nn", schema).unwrap();
+        let idx = engine.get_index("exists-nn").unwrap();
+        idx.abort_background_tasks();
+        // Each doc sets BOTH a schemaless `f` and a declared keyword `kw` to the
+        // same shape, so both field kinds are exercised identically.
+        let docs = vec![
+            ("present", json!("x")),
+            ("null", Value::Null),
+            ("empty_arr", json!([])),
+            ("null_arr", json!([null])),
+            ("mixed_arr", json!(["a", null])),
+        ];
+        for (id, v) in &docs {
+            idx.index_document(Some((*id).into()), json!({ "f": v, "kw": v }))
+                .await
+                .unwrap();
+        }
+        // "missing" sets neither field.
+        idx.index_document(Some("missing".into()), json!({ "other": 1 }))
+            .await
+            .unwrap();
+        let want = vec!["mixed_arr".to_string(), "present".to_string()];
+        let ex = |idx: std::sync::Arc<crate::Index>, field: &'static str, size: u64| async move {
+            let req = xerj_query::parse_request(&json!({
+                "query": {"exists": {"field": field}}, "size": size, "track_total_hits": true
+            }))
+            .unwrap();
+            let r = idx.search(&req).await.unwrap();
+            let mut ids: Vec<_> = r.hits.iter().map(|h| h.id.clone()).collect();
+            ids.sort();
+            (r.total.value, ids)
+        };
+        for phase in ["pre", "post"] {
+            for field in ["f", "kw"] {
+                let (total, ids) = ex(idx.clone(), field, 10).await;
+                assert_eq!(
+                    ids, want,
+                    "{phase} exists on {field}: only non-null values match"
+                );
+                assert_eq!(total, 2, "{phase} exists on {field} count must be 2");
+                let (c0, _) = ex(idx.clone(), field, 0).await;
+                assert_eq!(c0, 2, "{phase} exists on {field} size:0 count must be 2");
+            }
+            if phase == "pre" {
+                idx.flush().await.unwrap();
+                assert_eq!(idx.memtable.doc_count(), 0);
+            }
+        }
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn portable_field_flush_keeps_baseline_data_dir_format() {
         let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
         let dir = tempfile::tempdir().unwrap();
@@ -36394,6 +36459,19 @@ fn doc_matches_query(q: &QueryNode, source: &Value) -> bool {
     doc_matches_query_typed(q, source, &EMPTY_SCHEMA)
 }
 
+/// ES `exists` semantics: a field is present iff it has at least one non-null
+/// value. An explicit `null`, an empty array `[]`, and an array of only nulls
+/// `[null]` are NOT present; `[null, "x"]` is (it has a non-null element).
+/// Scalars and objects are present (object sub-field recursion is a separate
+/// concern the dense/declared paths already handle).
+fn value_present(v: &Value) -> bool {
+    match v {
+        Value::Null => false,
+        Value::Array(a) => a.iter().any(value_present),
+        _ => true,
+    }
+}
+
 fn doc_matches_query_typed(q: &QueryNode, source: &Value, schema: &Schema) -> bool {
     match q {
         QueryNode::MatchAll => true,
@@ -36519,7 +36597,12 @@ fn doc_matches_query_typed(q: &QueryNode, source: &Value, schema: &Schema) -> bo
         QueryNode::Exists { field } => match field.as_str() {
             "_id" | "_index" | "_seq_no" | "_version" | "_primary_term" => true,
             "_routing" => source.get("_routing").is_some(),
-            _ => get_field_value(source, field).is_some(),
+            // ES `exists` matches a field iff it has at least one NON-NULL value.
+            // Plain key-presence wrongly matched an explicit `null`, an empty
+            // array `[]`, and an array of only nulls `[null]`. This aligns the
+            // schemaless source scan with the declared-field paths (a keyword /
+            // numeric dv column has no entry for a null doc).
+            _ => get_field_value(source, field).is_some_and(|v| value_present(&v)),
         },
 
         QueryNode::Prefix { field, value, .. } => {


xerj-org:
3 adversarial skeptics (correctness / regression / evidence-holds), each REFUTE-in-worktree under 1.97.1. All HOLDS, 0 blocking. Verbatim:

PR #793 correctly narrows source-scan `exists` from key-presence to "has >=1 non-null value" via the new `value_present` helper. Verified fix + test present, rustc 1.97.1. value_present is ES-correct for every shape I reasoned through and empirically probed: null/[]/[null]->false, nested [[null]]/[[]]->false (matches ES array flattening), [null,0]/[false]/[null,"x"]->true, scalars->true. Empty-object {}->true is UNCHANGED from the old is_some() (not PR-introduced; docstring scopes objects out). Meta arms (_id.. -> true, _routing -> key-presence) untouched. No divergence between the fixed source scan and the declared keyword path: I probed the null-FIRST array [null,"a"] — the exact shape the commit claims matches but the shipped test omits (it uses value-first ["a",null]) — and it matches identically on schemaless `f` AND declared keyword `kw`, pre and post flush, with hits==size:0 count==2 (memtable==segment). No wrong result introduced or left. Non-blocking nit: the shipped test uses value-first ["a",null] for its mixed-array case, so it doesn't directly cover the null-first [null,"x"] the commit message explicitly claims; I verified that case passes separately.

Evidence: $ grep -n "fn value_present" engine/crates/xerj-engine/src/index.rs
36467:fn value_present(v: &Value) -> bool {
$ grep -rn "exists_requires_a_non_null_value_across_flush" engine/
engine/crates/xerj-engine/src/index.rs:1598:    async fn exists_requires_a_non_null_value_across_flush() {
$ rustup run 1.97.1 rustc --version
rustc 1.97.1 (8bab26f4f 2026-07-14)

Diff Exists arm (index.rs:36646): meta arms unchanged; `_ => get_field_value(source, field).is_some_and(|v| value_present(&v))`. value_present: Null->false, Array->any(value_present), _->true.

Added a probe test `probe_exists_null_first_array` covering the null-FIRST array [null,"a"] (commit-claimed shape omitted by shipped test) alongside value-first ["a",null], on schemaless `f` and declared keyword `kw`, pre+post flush.

$ cd engine && rustup run 1.97.1 cargo test -p xerj-engine --profile ci-test --lib exists -- --nocapture --test-threads=1
test index::flush_publication_recovery_tests::exists_requires_a_non_null_value_across_flush ... ok
test index::flush_publication_recovery_tests::probe_exists_null_first_array ...
PROBE pre  field=f  total=2 ids=["nullfirst", "valuefirst"]   size0_count=2
PROBE pre  field=kw total=2 ids=["nullfirst", "valuefirst"]   size0_count=2
PROBE post field=f  total=2 ids=["nullfirst", "valuefirst"]   size0_count=2
PROBE post field=kw total=2 ids=["nullfirst", "valuefirst"]   size0_count=2
test result: ok. 4 passed; 0 failed; 0 ignored; 641 filtered out; finished in 0.66s

=> declared keyword path agrees with fixed source scan on the null-first array edge; hits==count==2; memtable(pre)==segment(post). No correctness defect found.

Nits:
- Shipped test `mixed_arr` uses value-first ["a", null]; it does not directly exercise the null-first [null,"x"] the commit message specifically claims should match. Verified separately to pass, but a null-first array case would harden the regression guard.

## REGRESSION — HOLDS (non-blocking)

Could not refute. PR #793 changes the source-scan `exists` arm from key-presence (`.is_some()`) to non-null (`.is_some_and(|v| value_present(&v))`), matching ES semantics. The only behavior change is null/[]/[null] no longer match; scalars (incl. "", 0, false), objects, and arrays with >=1 non-null element still match, so no legitimate exists match is lost. No existing test or conformance case indexes a null/[]/[null] value into an exists-tested field, so nothing existing changes result. Full lib suite 644/0 under 1.97.1; the new test is a genuine guard (fail-before returns all 5, fix returns 2) and confirms the declared-keyword and size:0-count paths agree with the source scan pre- and post-flush. fmt + clippy clean. No regression.

Evidence: git reset --hard origin/fix/792-exists-nonnull -> HEAD f6b07774. rustup run 1.97.1 rustc --version -> 1.97.1. grep "fn value_present" index.rs -> 36467; arm at 36605; test at index.rs:1598.
Full suite: `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib` -> "test result: ok. 644 passed; 0 failed; 0 ignored".
New test: `cargo test -p xerj-engine --lib exists_requires_a_non_null_value` -> "test index::flush_publication_recovery_tests::exists_requires_a_non_null_value_across_flush ... ok".
Fail-before: edited arm to `.is_some()`, reran -> panicked at index.rs:1647 "pre exists on f: only non-null values match" left ["empty_arr","mixed_arr","null","null_arr","present"] right ["mixed_arr","present"]; test FAILED. Restored via `git checkout --`; arm back to is_some_and at 36605, tree clean.
fmt: `rustup run 1.97.1 cargo fmt --all --check` -> FMT_EXIT=0. clippy: `rustup run 1.97.1 cargo clippy -p xerj-engine --all-targets -- -D warnings` -> Finished, no warnings.
Conformance: search/160_exists_query.yml docs 1-3 have non-null values, doc 4 body:{} (missing), separate empty_text sets text:"" (scalar, still matches, expects 1); no null/[]/[null] doc. 161/340/341/200 exists-tested fields hold only non-null scalars/objects or are missing. Integration in_stock=4 (true/true/false/true), tags.key array of non-null scalars, emb non-null vectors, nonexistent=0 -- none index null/[]/[null].

## exists non-null semantics — HOLDS (non-blocking)

PR #793's fix is correct and the test genuinely guards it. Fix asserted present (value_present at index.rs:36467, arm at 36605, test at index.rs:1598) on origin/fix/792-exists-nonnull @ f6b07774; rustc 1.97.1. (1) Fail-before CONFIRMED: reverting only the arm to .is_some() makes the test FAIL, returning all 5 docs. (2) The PR's rationale that declared-field paths were "already correct" is INACCURATE (non-blocking): with the arm reverted, a DECLARED keyword field's exists HITS query also returns all 5 docs both pre- and post-flush, so it too flows through the source scan and is fixed by the same change rather than aligning to an already-correct path. No path DIVERGES with the fix applied (full test green for f and kw, pre/post, size:10/size:0). (3) Dense-count path bails to scan on nulls/arrays — confirmed in code. (4) 160_exists_query.yml never tests null/[]/[null] — confirmed. No blocking defect found.

Evidence: FIX PRESENT + toolchain:
$ grep -n "fn value_present" engine/crates/xerj-engine/src/index.rs -> 36467:fn value_present(v: &Value) -> bool {
$ grep -rn exists_requires_a_non_null_value_across_flush -> index.rs:1598
$ rustup run 1.97.1 rustc --version -> rustc 1.97.1 (8bab26f4f 2026-07-14)
arm at 36605: _ => get_field_value(source, field).is_some_and(|v| value_present(&v)),

WITH FIX (unchanged branch):
$ rustup run 1.97.1 cargo test -p xerj-engine --lib exists_requires_a_non_null_value_across_flush
-> test ... ok. 1 passed; 0 failed (finished in 0.61s). Re-run after restore: ok, 1 passed (0.48s).

(1) FAIL-BEFORE — arm reverted to `.is_some()`:
-> test ... FAILED. panicked at index.rs:1647: "pre exists on f: only non-null values match"
   left: ["empty_arr","mixed_arr","null","null_arr","present"]  right: ["mixed_arr","present"]
Fix necessary; confirmed.

(2) DECLARED PATH probe — arm still reverted, field loop narrowed to ["kw"]:
 pre-flush: FAILED at index.rs:1647 "pre exists on kw" left=all 5, right=[mixed_arr,present]
 post-flush (flush() before loop, phase=["post"], field=["kw"]): FAILED at index.rs:1649 "post exists on kw" left=all 5, right=[mixed_arr,present]
=> declared keyword exists HITS is resolved by the source-scan arm, NOT the KeywordPrefix{prefix:""} scored plan (index.rs:43449). PR's "declared paths were already correct" is inaccurate. But fix covers it (full test green with fix on kw pre+post), so no divergence-from-fix.

(3) COUNT PATH (code): index.rs:21597-21599 `if self.store.version_map.ghost_events() > 0 || self.memtable.doc_count() > 0 { return None; }`; 21619-21621 `if !null_empty || dcount != meta.doc_count { return None; // sparse / partial -> scan }`. Bails to source scan when nulls/arrays present. Confirmed.

(4) CONFORMANCE: engine/tests/es-compat-yaml/yaml/search/160_exists_query.yml
$ grep -cin null 160_exists_query.yml -> 0
$ grep -in ": null|: \[\]|\[null\]" -> (no matches)
Only "empty" case is `text: ""` (line 1273), an empty STRING (non-null scalar) with `hits.total: 1` — correct, not the null/[]/[null] case. Conformance never covered the bug. Confirmed.

Worktree restored: `git reset --hard origin/fix/792-exists-nonnull` -> HEAD f6b07774; fix lines present; final test green.

Nits:
- PR commit message claim 'the declared-field paths, which were already correct' is inaccurate: with the arm reverted, a declared keyword field's exists HITS query returns all 5 docs both pre- and post-flush, so it flows through the same source scan and is corrected by this same fix rather than aligning to an already-correct KeywordPrefix{prefix:''} plan. Non-blocking (fix covers it).
- Stale comment at index.rs ~21575 still says the scan 'resolves get_field_value(...).is_some() exactly'; post-fix it resolves is_some_and(value_present). Cosmetic.

---
Merge gate: 0 BLOCKING; ES-compat YAML conformance is green (no null/[]/[null] case in 160_exists_query.yml, confirmed). Two non-blocking nits noted: (1) the shipped test uses a value-first mixed array ["a",null]; the null-first [null,"x"] case the commit cites was verified to pass separately. (2) CORRECTION to the PR description: a DECLARED keyword field's exists ALSO flowed through the source scan and was ALSO buggy pre-fix (not "already correct" as I wrote) — the same one-line change fixes both schemaless and declared, and no path diverges. Neither blocks. Merging on all-green + CLEAN.

PR #795, opened 2026-08-26
fix(engine): #794 prefix/wildcard in bool.must_not stays case-sensitive after flush

### LINKED ISSUE #794: es-compat: prefix/wildcard in bool.must_not case-folds after flush, over-excluding docs (#423 schemaless site)
A `prefix` (or `wildcard`) clause inside `bool.must_not` **case-folds after flush**, over-excluding documents. On a declared `keyword` field, `prefix` is case-sensitive, so `must_not [prefix "ap"]` must keep `"Application"` (capital A) — but on the segment path it is wrongly excluded. A concrete #423 instance (a schemaless `doc_matches_query` reaching a hit-affecting path).

## Reproduced (in-process, current main, 1.97.1)

Declared `keyword` field `k`; docs `1=apple`, `2=Application`, `3=banana`.

```
                                          pre-flush    post-flush   ES-correct
bool.must_not [term "apple"]              [2,3]        [2,3]  ok    [2,3]
bool.must_not [prefix "ap"]               [2,3]        []     WRONG [2,3]
bool.must_not [wildcard "ap*"]            [2,3]        []     WRONG [2,3]
bool{must:[match_all], must_not:[prefix "ap"]}  [2,3]  [3]    WRONG [2,3]  (drops "Application")
```

`term` in `must_not` is fine on both paths; only `prefix`/`wildcard` diverge, and only after flush. Direct (non-must_not) `prefix "ap"` is correct on every path ([apple] only) — so the case-sensitive keyword-prefix logic itself works; it's the segment `must_not` evaluation that loses it.

## Cause

The segment `must_not` / scoring evaluation (`score_query_against_doc`, index.rs ~38772, and the sibling filter closures) calls the **schemaless** `doc_matches_query(q, source)` — the `EMPTY_SCHEMA` shim (index.rs:36457) — instead of `doc_matches_query_typed(q, source, schema)`. The `Prefix`/`Wildcard` arms key off `declared_field(schema, field)` to pick keyword (case-sensitive) vs the unknown-field fold-both fallback; with `EMPTY_SCHEMA` the field is always "unknown", so a keyword field's prefix folds case and matches `"Application"` for `"ap"`, which `must_not` then excludes.

This is one of the ~10 remaining schemaless `doc_matches_query` call sites #423 tracks. Pre-flush (memtable) uses a schema-aware path, so it's a flush-divergence too.

## Fix direction

Thread the live `Schema` into the segment `must_not`/prefix evaluation so it calls `doc_matches_query_typed(q, source, &schema)`. `score_query_against_doc` and the filter closures take no `Schema` today; the fix threads it from the search entry (which holds `self.schema`) down to these hit-affecting sites — the exact "thread Schema through hit-affecting ones" step #423 names. Care: keep the genuinely-schemaless shim callers (non-hit-affecting) unchanged.

Found via a proactive probe of the #423 schemaless call sites.

xerj-org:
Closes #794.

A keyword `prefix`/`wildcard` inside `bool.must_not` folded case on the segment after flush, over-excluding docs whose value differed only by case (e.g. `must_not: prefix "ap"` wrongly dropped `Application`).

Root cause: two post-flush paths — the count shortcut (`try_shortcut_count`) and the stored-section scan (`scan_stored_section_into`) — still called the schemaless `doc_matches_query` (EMPTY_SCHEMA), so the Prefix/Wildcard arms took the unknown-field fold-both fallback instead of the case-sensitive keyword path. This is one more #423 schemaless-callsite instance. Threads the owned `dmq_schema` through both (plus scoring/hydration paths in earlier commits of this branch).

Fail-before (test un-ignored here): `bool{must:[match_all], must_not:[prefix "ap"]}` over keyword `k` = apple/Application/banana returns `(1, ["3"])` (Application wrongly excluded, count wrong) pre-fix, `(2, ["2","3"])` post-fix, both pre- and post-flush.

- Full `xerj-engine` lib suite 645/0
- `cargo fmt --all --check` clean
- `clippy -p xerj-engine --all-targets -D warnings` clean
- rustc 1.97.1

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1659,6 +1659,52 @@ mod flush_publication_recovery_tests {
         }
     }
 
+    #[ignore = "#794 fail-before repro; un-ignore when schema is threaded into score_query_against_doc/apply_rescore"]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn mustnot_prefix_on_keyword_stays_case_sensitive_after_flush() {
+        // #794: a prefix/wildcard inside bool.must_not case-folded on the segment
+        // path (schemaless doc_matches_query), wrongly excluding "Application"
+        // for prefix "ap" on a declared keyword field. Must be case-sensitive,
+        // pre AND post flush.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("k", FieldType::Keyword))
+            .unwrap();
+        engine.create_index("mn794", schema).unwrap();
+        let idx = engine.get_index("mn794").unwrap();
+        idx.abort_background_tasks();
+        for (id, k) in [("1", "apple"), ("2", "Application"), ("3", "banana")] {
+            idx.index_document(Some(id.into()), json!({ "k": k })).await.unwrap();
+        }
+        let run = |idx: std::sync::Arc<crate::Index>, q: serde_json::Value| async move {
+            let req = xerj_query::parse_request(&json!({"query": q, "size": 10, "track_total_hits": true})).unwrap();
+            let r = idx.search(&req).await.unwrap();
+            let mut ids: Vec<_> = r.hits.iter().map(|h| h.id.clone()).collect();
+            ids.sort();
+            (r.total.value, ids)
+        };
+        // must_not [prefix ap] -> keep Application(2) + banana(3); with match_all base too.
+        let want = (2u64, vec!["2".to_string(), "3".to_string()]);
+        for phase in ["pre", "post"] {
+            for q in [
+                json!({"bool": {"must_not": [{"prefix": {"k": "ap"}}]}}),
+                json!({"bool": {"must": [{"match_all": {}}], "must_not": [{"prefix": {"k": "ap"}}]}}),
+                json!({"bool": {"must_not": [{"wildcard": {"k": "ap*"}}]}}),
+            ] {
+                assert_eq!(run(idx.clone(), q.clone()).await, want, "#794 {phase} {q}: keyword prefix/wildcard in must_not is case-sensitive");
+            }
+            if phase == "pre" {
+                idx.flush().await.unwrap();
+                assert_eq!(idx.memtable.doc_count(), 0);
+            }
+        }
+    }
+
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn portable_field_flush_keeps_baseline_data_dir_format() {
         let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1659,7 +1659,6 @@ mod flush_publication_recovery_tests {
         }
     }
 
-    #[ignore = "#794 WIP: scoring+hydration threaded; exclusion is elsewhere (prefix prefilter?) — TRACE next"]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn mustnot_prefix_on_keyword_stays_case_sensitive_after_flush() {
         // #794: a prefix/wildcard inside bool.must_not case-folded on the segment
@@ -16430,7 +16429,7 @@ impl Index {
         let shortcut_count: Option<u64> = if scored_fast_ready {
             None
         } else {
-            self.try_shortcut_count(query, &snap, is_match_all, mem_matches_known)
+            self.try_shortcut_count(query, &snap, is_match_all, mem_matches_known, &dmq_schema)
                 .await
         };
 
@@ -18065,6 +18064,7 @@ impl Index {
                             &mut dbg_walked,
                             &mut dbg_admitted,
                             score_doc,
+                            &dmq_schema,
                         );
                         dbg_scan_ms += t_scan.elapsed().as_millis() as u64;
                     } else if let Some(warm_slices) = (!sorted_candidates_path)
@@ -18132,6 +18132,7 @@ impl Index {
                                     &mut dbg_walked,
                                     &mut dbg_admitted,
                                     score_doc,
+                                    &dmq_schema,
                                 );
                             }
                         }
@@ -18215,6 +18216,7 @@ impl Index {
                                     &mut dbg_walked,
                                     &mut dbg_admitted,
                                     score_doc,
+                                    &dmq_schema,
                                 );
                                 dbg_scan_ms += t_scan.elapsed().as_millis() as u64;
                                 // Cache only a COMPLETE offsets map (a
@@ -21064,6 +21066,7 @@ impl Index {
         snap: &xerj_storage::index_store::IndexSnapshot,
         is_match_all: bool,
         mem_matches_known: Option<u64>,
+        schema: &Schema,
     ) -> Option<u64> {
         // Unwrap wrapper queries so constant_score and boosted
         // queries benefit from the same fast paths.
@@ -21362,7 +21365,7 @@ impl Index {
             } else {
                 let docs = self.memtable.all_docs_with_sources_arc();
                 docs.iter()
-                    .filter(|(_, src)| doc_matches_query(query, src))
+                    .filter(|(_, src)| doc_matches_query_typed(query, src, schema))
                     .count() as u64
             };
 
@@ -21587,7 +21590,7 @@ impl Index {
             } else {
                 let docs = self.memtable.all_docs_with_sources();
                 docs.iter()
-                    .filter(|(_, src)| doc_matches_query(query, src))
+                    .filter(|(_, src)| doc_matches_query_typed(query, src, schema))
                     .count() as u64
             };
             return Some(seg_matches + mem_matches);
@@ -25585,6 +25588,7 @@ impl Index {
         // Per-doc scorer from `search_inner` (`score_doc`): single-applies
         // any peeled function_score and resolves `_id` for ids clauses.
         scorer: &(dyn Fn(&Value, &str) -> f32 + Send + Sync),
+        schema: &Schema,
     ) {
         // Fast path: scan one doc at a time with a hand-rolled array splitter.
         // The stored section is always shaped as `[{...}, {...}, ...]` (a
@@ -25802,9 +25806,9 @@ impl Index {
                     } else {
                         source_ref.clone()
                     };
-                    doc_matches_query(query, &source_with_id)
+                    doc_matches_query_typed(query, &source_with_id, schema)
                 } else {
-                    doc_matches_query(query, source_ref)
+                    doc_matches_query_typed(query, source_ref, schema)
                 }
             };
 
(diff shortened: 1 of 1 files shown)


xerj-org:

### Correctness

**Summary:** PR #795 correctly fixes #794. Threading the search's owned dmq_schema into scan_stored_section_into / hydrate_prefiltered_unsorted / score_query_against_doc / apply_rescore makes the post-flush segment path case-SENSITIVE for keyword prefix/wildcard in must_not, matching the memtable path and ES. I could not construct any query where the change turns a correct answer wrong or leaves #794 wrong. An 8-shape matrix (keyword/text/dynamic x prefix/wildcard/term/positive) verified pre-flush(memtable)==post-flush(segment) AND count(size:0)==hits(size:10) for every shape, all ES-correct. Text prefix stays case-insensitive (uppercase text prefix now correctly matches nothing on the segment path too — an improvement); dynamic/undeclared fields stay flush-invariant; must_not[term], positive prefix, and count/hit agreement all correct. The two try_shortcut_count changes (21368 Bool-conjunction, 21593 Regexp) are provable no-ops (21368 gated by memtable_count_covered_by_shortcut to schema-independent Term/Range arms; 21593 Regexp uses compiled_anchored_regex; must_not[prefix] count routes via bool_should_mustnot_count which returns None on a prefix leaf and falls to the same schema-aware hit path). Fail-before proven: reverting the segment scan to schemaless reproduces #794 ((1,["3"]) instead of (2,["2","3"])). No regressions across 32 neighbor tests.

**Evidence:**
```
rustc 1.97.1 (8bab26f4f 2026-07-14). At d422b17d. grep confirmed the 4 doc_matches_query_typed(query, .., schema) sites (scan 25965/25967, prefilter 25413, count 21368/21593) and test not #[ignore]d.

PR test (fix in place): `cargo test -p xerj-engine --lib mustnot_prefix_on_keyword_stays_case_sensitive_after_flush` => test result: ok. 1 passed.

Skeptic 8-shape matrix `skeptic794_mustnot_schema_regression_matrix -- --nocapture` (asserts pre==post, count(size0).total==hits(size10).total, ES answers): ok. 1 passed. Probe output pre AND post identical:
kw must_not prefix ap => (2,[2,3]); kw must_not wildcard ap* => (2,[2,3]); kw must_not prefix Ap => (2,[1,3]); kw must_not term apple => (2,[2,3]); kw positive prefix ap => (1,[1]); text must_not prefix ap => (1,[3]); text must_not prefix Ap => (3,[1,2,3]); dyn must_not prefix ap => (1,[3]).

FAIL-BEFORE: perl-reverted scan_stored_section_into+hydrate_prefiltered_unsorted to schemaless doc_matches_query, recompiled 1.97.1 => PR test FAILED: post {prefix k:"ap"} left=(1,["3"]) right=(2,["2","3"]); matrix FAILED at pre!=post [kw must_not prefix ap] left=(1,["3"]) right=(2,["2","3"]). Restored via git checkout.

Regression sweep `cargo test -p xerj-engine --lib -- prefix wildcard must_not mustnot rescore function_score score boost constant dis_max boosting flush_publication` => 32 passed; 0 failed (incl keyword_match_in_rescore_query_is_rewritten exercising threaded apply_rescore). Worktree restored clean (git status empty at d422b17d).
```

**Nits (non-blocking):**
- Line 44302 apply_function_score's per-function filter still calls schemaless doc_matches_query (fold-both) — a function_score filter:{prefix:{keyword_field}} would score case-insensitively, diverging from the case-sensitive main-query semantics. Out of #794 scope, pre-existing #423-family, and flush-invariant (identical on memtable and segment), so not a wrong-result regression from this PR; worth a follow-up under #423.
- The dynamic/undeclared string field ('d') is dynamically mapped to text, so its must_not prefix is case-insensitive rather than the fold-both unknown-field branch — but the observable answer is identical for lowercase prefixes and, more importantly, pre==post, so no correctness impact.

### Regression

**Summary:** PR #795 holds. The fix threads the search's owned dmq_schema (self.schema().await, owned Schema) into both post-flush schemaless matcher paths: try_shortcut_count's two memtable count filters now call doc_matches_query_typed(query, src, schema) (index.rs:21368, 21593), and scan_stored_section_into's stored-scan calls route through _typed (25806/25808), plus hydrate_prefiltered_unsorted (25254), apply_rescore (38733) and score_query_against_doc. Every call site of the signature-changed functions passes a real &dmq_schema (16432 for try_shortcut_count; 18049/18117/18201 for scan_stored_section_into; 18101 for hydrate; 18670 for apply_rescore; 15474/15482/15485 for score_query_against_doc) — no stray EMPTY_SCHEMA leaked in (the only EMPTY_SCHEMA is the legacy doc_matches_query wrapper). The new test mustnot_prefix_on_keyword_stays_case_sensitive_after_flush exists, is not #[ignore]d, and passes pre and post flush (total=2, ids=[2,3]). fmt and clippy --all-targets -D warnings both clean. Full lib suite: 645 passed / 0 failed on a clean re-run. The single transient failure (cancelling_flush_waiter_after_worker_start_still_restores, an Elapsed 5s-timeout) is a pre-existing timing flake untouched by the diff, present on main, involving no queries — it passes in isolation and on suite re-run. Not a regression; no blocking defect found.

**Evidence:**
```
git reset --hard origin/fix/794-mustnot-prefix-schema -> HEAD d422b17d.
grep "doc_matches_query_typed(query, src, schema)" index.rs -> 21368, 21593 (both inside try_shortcut_count, def at 21063).
grep test -> index.rs:1663 async fn mustnot_prefix_on_keyword_stays_case_sensitive_after_flush (no #[ignore]).
self.schema() -> `self.schema.read().await.schema.clone()` (owned Schema); dmq_schema defined index.rs:14709.
Call-site grep: try_shortcut_count only real call 16432 (&dmq_schema); scan_stored_section_into calls 18049/18117/18201 (&dmq_schema); hydrate_prefiltered_unsorted 18101; apply_rescore 18670; score_query_against_doc 15474/15482/15485 — all &dmq_schema/forwarded schema.
`cd engine && rustup run 1.97.1 cargo fmt --all --check; echo $?` -> FMT_EXIT=0.
`RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo clippy -p xerj-engine --all-targets -- -D warnings` -> Finished; CLIPPY_EXIT=0.
`RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib` run 1 -> "644 passed; 1 failed" (cancelling_flush_waiter_after_worker_start_still_restores panicked index.rs:1131 'detached finalizer must publish after caller cancellation: Elapsed(())'); new test mustnot_prefix... printed ok.
Same test in isolation -> "1 passed" in 0.28s.
Full suite run 2 -> "645 passed; 0 failed".
git diff origin/main..HEAD -- index.rs | grep -c cancelling_flush_waiter... -> 0 (untouched); git show origin/main:engine/.../index.rs | grep -c -> 1 (pre-exists on main). Panic site index.rs:1123-1131 is a tokio::time::timeout(5s) waiting on detached finalizer, no queries involved.
rustc 1.97.1 (8bab26f4f 2026-07-14).
```

**Nits (non-blocking):**
- cancelling_flush_waiter_after_worker_start_still_restores uses a hard 5s tokio::time::timeout for a detached background finalizer; under full-suite parallel load on a worker_threads=2 runtime it can starve and flake (Elapsed). Pre-existing on main, not this PR's concern, but a candidate for a longer budget or explicit runtime yield.

### Evidence-holds

**Summary:** Could not refute. At d422b17d (rustc 1.97.1): the fix threads the search's owned schema into both post-flush schemaless matchers — the count shortcut (index.rs:21368/21593) and the stored-section scan (25809/25811) — so a keyword prefix/wildcard inside bool.must_not stays case-sensitive. FIXED test mustnot_prefix_on_keyword_stays_case_sensitive_after_flush passes (1 passed; 0 failed; 0 ignored), returning (2, ["2","3"]) pre AND post flush. Fail-before: reverting ONLY the 4 internal-call swaps back to schemaless doc_matches_query makes the test FAIL with left (1, ["3"]) on the POST (post-flush segment) phase — exactly the claimed signature (Application over-excluded). Test is not #[ignore]d and asserts memtable.doc_count()==0 between phases, so post truly exercises the segment path, not the memtable. Reverting also produced unused-variable:schema warnings at 21069 (try_shortcut_count) and 25591 (scan_stored_section_into), confirming the swap is the load-bearing change. Nit (non-blocking): this test's post-flush failure is driven by the scan-path pair (25809/25811); the count-shortcut threading (21368/21593) is a safe companion change but its pre-flush count path appears served by the bool doc-values fast path here, so it may not be independently exercised by this single test. Not a defect.

**Evidence:**
```
$ git reset --hard origin/fix/794-mustnot-prefix-schema && git log --oneline -1
HEAD is now at d422b17d fix(engine): #794 thread schema into count+scan matchers so must_not[prefix] stays case-sensitive after flush

$ grep -n "doc_matches_query_typed(query, src, schema)" engine/crates/xerj-engine/src/index.rs
21368:                    .filter(|(_, src)| doc_matches_query_typed(query, src, schema))
21593:                    .filter(|(_, src)| doc_matches_query_typed(query, src, schema))
(scan sites 25809/25811: doc_matches_query_typed(query, &source_with_id, schema) / (query, source_ref, schema))
(test present at index.rs:1663, NOT #[ignore]d; flush assertion at 1710: assert_eq!(idx.memtable.doc_count(), 0))

$ cd engine && rustup run 1.97.1 rustc --version
rustc 1.97.1 (8bab26f4f 2026-07-14)

# (1) FIXED test PASSES
$ RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib mustnot_prefix_on_keyword -- --nocapture
test index::flush_publication_recovery_tests::mustnot_prefix_on_keyword_stays_case_sensitive_after_flush ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 644 filtered out; finished in 10.39s

# (2) FAIL-BEFORE: reverted 4 sites (21368/21593 -> doc_matches_query(query,src); 25809/25811 -> schemaless)
$ RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib mustnot_prefix_on_keyword -- --nocapture
warning: unused variable: `schema`  --> crates/xerj-engine/src/index.rs:21069:9   (try_shortcut_count)
warning: unused variable: `schema`  --> crates/xerj-engine/src/index.rs:25591:9   (scan_stored_section_into)
panicked at crates/xerj-engine/src/index.rs:1702:17:
assertion `left == right` failed: #794 post {"bool":{"must":[{"match_all":{}}],"must_not":[{"prefix":{"k":"ap"}}]}}: keyword prefix/wildcard in must_not is case-sensitive
  left: (1, ["3"])
 right: (2, ["2", "3"])
test result: FAILED. 0 passed; 1 failed; 0 ignored; 0 measured; 644 filtered out

# restored
$ git reset --hard origin/fix/794-mustnot-prefix-s
...(truncated)
```

**Nits (non-blocking):**
- This test's post-flush failure is driven by the scan-path pair (25809/25811); the count-shortcut threading at 21368/21593 is a safe companion change but its pre-flush count path is served by the bool doc-values fast path here, so it may not be independently exercised by this single test.
- Reverting leaves `schema` params unused in try_shortcut_count and scan_stored_section_into (warnings only, not deny-level) — confirms the swap is load-bearing but also that no other logic in those fns consumes the threaded schema.


### Independent re-run (correctness lens, fresh worktree)
A second, independently-spawned correctness skeptic ran a 12-shape pre-vs-post + size:0-vs-size:10 matrix (keyword/text/dynamic × prefix/wildcard/term/positive) and also returned **HOLDS / 0 BLOCKING**; fail-before reproduced #794 `(1,["3"])`→`(2,["2","3"])`.

**Gate:** 0 blocking findings across all skeptics + all checks green + mergeStateStatus CLEAN → merging. Non-blocking nits filed as a #423-family follow-up.

PR #798, opened 2026-08-26
fix(engine): #797 must_not-only prefix/wildcard drops all docs after flush (pure-negative FTS projection)

### LINKED ISSUE #797: es-compat: prefix/wildcard in a must_not-ONLY bool returns zero hits after flush (base enumeration scoped to the leaf's prefilter)
A `bool` with **only** a `must_not` (no `must`/`should`/`filter`) containing a `prefix` or `wildcard` silently returns **zero** results after flush. The pre-flush answer is correct; the doc set is not empty. `term` in the same position is fine on both paths, so this is specific to prefix/wildcard leaves.

Distinct from #794 (that was the `must:[match_all]+must_not[prefix]` *case-fold*; here every doc is dropped, count included).

## Repro
Declared keyword field `k`, docs `1=apple, 2=Application, 3=banana`. Query:
```json
{"bool": {"must_not": [{"prefix": {"k": "ap"}}]}}
```
Expected (exclude case-sensitive prefix "ap" = apple only): `total=2, hits=[2,3]`.

Observed (verified on main `2155784b`, rustc 1.97.1, unit harness — no :9200):
```
mustnot-only pre : term=(2,[2,3]) prefix=(2,[2,3]) wildcard=(2,[2,3])   # correct
mustnot-only post: term=(2,[2,3]) prefix=(0,[])    wildcard=(0,[])       # WRONG
```
Same for `wildcard {"k":"app*"}`. `size:0`/`track_total_hits` also report `total=0`.

## Direction
With no positive clause the base set must be **all docs**, with `must_not` subtracting matches. `term` does this correctly. For `prefix`/`wildcard` the segment planner appears to build a **candidate prefilter from the leaf** (the prefix/wildcard term-dictionary expansion) and scope enumeration to it, then `must_not` subtracts every candidate → empty, and the non-matching docs (2,3) never enter the candidate set. Fix = when a prefix/wildcard leaf appears only under `must_not` (negative), it must NOT drive candidate generation; the base enumeration stays all-docs (like the term path / a `match_all` positive driver). Segment path only (pre-flush memtable is correct).

Severity: wrong-results / silent data omission (a valid query returns nothing after flush). Part of the #423 family but a planner/prefilter bug, not the schemaless-matcher one.

xerj-org:
Closes #797.

A `bool` with **only** a `must_not` (no `must`/`should`/`filter`) and a projectable keyword leaf (`prefix`/`wildcard`) returned zero hits after flush; the doc set was non-empty and pre-flush was correct.

## Root cause
`query_node_to_fts_with_keyword_fields` projected the bool to a **pure-negative** FTS bool (`must_not:[prefix]`, no positive clause). FTS — like Lucene — needs a positive clause to enumerate a base set, so a pure-negative bool matches **nothing** → `seg_hits` empty. And because the leaf *did* project, `bool_has_nonprojectable_nonscoring` was false, so `residual_gate` was off and the empty FTS result was treated as authoritative. `term` must_not-only was unaffected only incidentally. Traced: `needs_fts=true, residual_gate=false`, post-flush `seg_total=0 seg_hits_len=0`.

## Fix
Track whether a **positive** (`must`/`should`/`filter`) clause projected. If none did, decline the projection (return `None`) so the stored-doc scan answers over the full doc set — the same path a must_not-only keyword `term` (which never projects) already takes. Positive bools with a non-projectable `must_not` are unchanged (they keep the FTS path + the existing `residual_gate` re-test).

## Verification (rustc 1.97.1, unit harness — no :9200)
Test data is case-fold-invariant (only "apple" starts with "ap"), so it is independent of the separate #794 keyword-prefix case-sensitivity fix.
- Fail-before: `bool{must_not:[prefix "ap"]}` over `apple/grape/banana` → `(0,[])` post-flush; with the fix `(2,["2","3"])`, pre- and post-flush (same for `wildcard "app*"`).
- Full `xerj-engine` lib suite 645/0
- `cargo fmt --all --check` clean; `clippy -p xerj-engine --all-targets -D warnings` clean

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1659,6 +1659,62 @@ mod flush_publication_recovery_tests {
         }
     }
 
+    #[ignore = "#797 WIP: fail-before repro — must_not-only prefix/wildcard drops all docs after flush"]
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn mustnot_only_prefix_keeps_nonmatching_docs_after_flush() {
+        // #797: a bool with ONLY a must_not (no must/should/filter) containing a
+        // prefix/wildcard leaf returns zero hits after flush — the segment
+        // planner lets the negative leaf drive candidate generation, so the
+        // non-matching docs never enter the set. `term` is unaffected. Base set
+        // must be all docs, must_not subtracting only its matches.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("k", FieldType::Keyword))
+            .unwrap();
+        engine.create_index("mn797", schema).unwrap();
+        let idx = engine.get_index("mn797").unwrap();
+        idx.abort_background_tasks();
+        for (id, k) in [("1", "apple"), ("2", "Application"), ("3", "banana")] {
+            idx.index_document(Some(id.into()), json!({ "k": k }))
+                .await
+                .unwrap();
+        }
+        let run = |idx: std::sync::Arc<crate::Index>, q: serde_json::Value| async move {
+            let req = xerj_query::parse_request(
+                &json!({"query": q, "size": 10, "track_total_hits": true}),
+            )
+            .unwrap();
+            let r = idx.search(&req).await.unwrap();
+            let mut ids: Vec<_> = r.hits.iter().map(|h| h.id.clone()).collect();
+            ids.sort();
+            (r.total.value, ids)
+        };
+        // must_not-ONLY [prefix "ap"] (case-sensitive keyword) excludes apple(1);
+        // Application(2) + banana(3) must remain, pre AND post flush.
+        let want = (2u64, vec!["2".to_string(), "3".to_string()]);
+        for phase in ["pre", "post"] {
+            for q in [
+                json!({"bool": {"must_not": [{"prefix": {"k": "ap"}}]}}),
+                json!({"bool": {"must_not": [{"wildcard": {"k": "app*"}}]}}),
+            ] {
+                assert_eq!(
+                    run(idx.clone(), q.clone()).await,
+                    want,
+                    "#797 {phase} {q}: must_not-only prefix/wildcard keeps non-matching docs"
+                );
+            }
+            if phase == "pre" {
+                idx.flush().await.unwrap();
+                assert_eq!(idx.memtable.doc_count(), 0);
+            }
+        }
+    }
+
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn mustnot_prefix_on_keyword_stays_case_sensitive_after_flush() {
         // #794: a prefix/wildcard inside bool.must_not case-folded on the segment
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1659,7 +1659,6 @@ mod flush_publication_recovery_tests {
         }
     }
 
-    #[ignore = "#797 WIP: fail-before repro — must_not-only prefix/wildcard drops all docs after flush"]
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn mustnot_only_prefix_keeps_nonmatching_docs_after_flush() {
         // #797: a bool with ONLY a must_not (no must/should/filter) containing a
@@ -1679,7 +1678,10 @@ mod flush_publication_recovery_tests {
         engine.create_index("mn797", schema).unwrap();
         let idx = engine.get_index("mn797").unwrap();
         idx.abort_background_tasks();
-        for (id, k) in [("1", "apple"), ("2", "Application"), ("3", "banana")] {
+        // Data is case-fold-invariant for prefix "ap"/"app*": only "apple"
+        // starts with it under ANY casing, so the correct answer does not
+        // depend on the separate #794 keyword-prefix case-sensitivity fix.
+        for (id, k) in [("1", "apple"), ("2", "grape"), ("3", "banana")] {
             idx.index_document(Some(id.into()), json!({ "k": k }))
                 .await
                 .unwrap();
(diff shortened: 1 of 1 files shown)


xerj-org:

### Correctness

**Summary:** Could not refute. Fix present at index.rs:41900/41914/41945/41961/41979 (projected_positive set ONLY in must/should/filter loops, NOT must_not; gate `if !projected_positive { return None; }`). Diff of b46818fb shows the sole change is renaming projected_any→projected_positive and dropping the must_not assignment, so a bool that projects no positive clause now DECLINES → stored-doc scan (base=all docs minus must_not), exactly as a must_not-only keyword term already did; positive bools are unchanged. Test mustnot_only_prefix_keeps_nonmatching_docs_after_flush exists, is NOT #[ignore]d, asserts memtable.doc_count()==0 between phases (lines 1657,1716), and uses case-fold-invariant data (only "apple" starts with "ap"/"app*"). (1) FAIL-BEFORE: re-adding `projected_positive = true;` in the must_not loop makes the test FAIL with left:(0,[]) vs right:(2,["2","3"]) on the post-flush phase. (2) Clean fix: test PASSES for both prefix "ap" and wildcard "app*", pre and post flush. (3+4) flush/doc_count assertion and case-fold-invariant data confirmed by reading the test. No regression: full flush_publication_recovery_tests module (16 tests incl. the #794 must_not-only case, which now takes the same changed decline path) all pass.

**Evidence:**
```
$ git fetch origin fix/797-mustnot-only-prefilter && git reset --hard origin/fix/797-mustnot-only-prefilter && git log --oneline -3
HEAD is now at 058041a7 chore: re-trigger CI (runner-saturation trigger miss)
058041a7 chore: re-trigger CI ...
b46818fb fix(engine): #797 must_not-only bool must not project to a pure-negative FTS bool
234e9f33 test(engine): #797 fail-before repro (ignored) ...

$ grep -n "projected_positive" engine/crates/xerj-engine/src/index.rs
41900:            let mut projected_positive = false;
41914:                projected_positive = true;   (must loop)
41945:                    projected_positive = true;   (filter loop)
41961:                projected_positive = true;   (should loop)
41979:            if !projected_positive {
# must_not loop (41965-41977) does NOT set projected_positive. Test at 1663, no #[ignore]; asserts idx.memtable.doc_count()==0 at 1657 & 1716; data [("1","apple"),("2","grape"),("3","banana")], want=(2,["2","3"]).

$ git show b46818fb -- ...index.rs | grep projected
-            let mut projected_any = false;
+            let mut projected_positive = false;
   must:  -projected_any = true; +projected_positive = true;
   filter:-projected_any = true; +projected_positive = true;
   should:-projected_any = true; +projected_positive = true;
   must_not: -projected_any = true;   (line REMOVED, no replacement)
-            if !projected_any {
+            if !projected_positive {

(2) CLEAN FIX PASSES:
$ RUST_MIN_STACK=8388608 rustup ru
...(truncated)
```

- nit: Fail-before reproduces (0,[]) only on the post-flush phase, not pre — consistent with the root cause being the segment/FTS pure-negative projection; the pre-flush memtable path was already correct. Test ordering (assert then flush) exercises both, so this is covered.
- nit: The fix also silently improves a related shape bool{filter:[non-projectable], must_not:[projectable]}: before-fix it too built a pure-negative FTS bool (must_not sets the old projected_any), after-fix it declines to the stored scan. Not a regression (an improvement), and not covered by a dedicated test, but worth noting as extra scope beyond the pure must_not-only case.

### Regression

**Summary:** Could not refute PR #798. The fix is correct and load-bearing. It replaces `projected_any` (set by all four bool loops incl. must_not) with `projected_positive` (set ONLY in must/should/filter), gating the FTS-bool projection on a positive clause: `if !projected_positive { return None; }`. Effect is surgically narrow — the ONLY shape whose behavior changes is a bool that projects a must_not but no positive clause (previously Some(pure-negative FTS bool) → matched nothing post-flush; now None → stored scan over the full doc set = "all docs minus must_not"). Every positive-bearing bool is unchanged (projected_positive=true → FTS path + existing residual_gate). Verified all required shapes (a)-(f) plus the residual-gate path: correct, pre==post, count(size:0)==hits(size:10). Confirmed fail-before by re-introducing the old must_not→flag behavior (must_not-only cases regress to (0,[]) while positive cases stay green, isolating the fix exactly). No query gets a wrong result and no shape breaks.

- nit: Correctness-neutral perf note (not blocking, already fixed a wrong-result): declining projection routes must_not-only bools to the O(N) stored-doc scan instead of the FST postings path. That matches the pre-existing must_not-only keyword-term behavior and is strictly better than the old wrong (empty) result, so it is acceptable, but a future follow-up could enumerate a base set (match_all) + must_not on the FTS path to keep the fast path.

### Evidence-holds

**Summary:** Could not refute PR #798. The fix is correctly scoped: `projected_positive` is set ONLY in the must/filter/should loops (index.rs:41914/41945/41961), never in the must_not loop, and the final gate is `if !projected_positive { return None }` (41979). The only behavioral change is that a bool projecting NO positive clause but a projecting must_not now declines the projection (returns None -> authoritative stored-doc scan) instead of emitting a pure-negative FTS bool that matched nothing. Returning None is always correctness-safe (the stored scan handles every shape and was already the path for must_not-only `term`). filter-only and should-only bools still project (their loops set the flag); positive bools with a non-projecting must_not are unchanged (projected_positive=true keeps the FTS path + residual_gate). No test relied on the old pure-negative projection. Full crate suite green, fmt/clippy clean, merges cleanly into origin/main. No regression.

**Evidence:**
```
Branch: reset --hard origin/fix/797-mustnot-only-prefilter -> HEAD 058041a7 (b46818fb fix + 234e9f33 fail-before). 3 ahead / 1 behind origin/main; no conflict markers in index.rs.

Fix present: grep "projected_positive" -> 41900 (decl false), 41914 (must), 41945 (filter), 41961 (should), 41979 (gate `if !projected_positive`). Confirmed must_not loop (41965-41977) does NOT set the flag.

Test present & not ignored: mustnot_only_prefix_keeps_nonmatching_docs_after_flush at index.rs:1663, plain #[tokio::test], `grep #[ignore]` over the file returns NOTHING.

Toolchain: rustc 1.97.1 (8bab26f4f 2026-07-14).

Targeted test: `RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-engine --lib mustnot_only_prefix_keeps_nonmatching_docs_after_flush` -> "test result: ok. 1 passed; 0 failed".

FULL crate suite (#529 lesson): `... cargo test -p xerj-engine --lib` -> "test result: ok. 646 passed; 0 failed; 0 ignored".

Projection module: `... cargo test -p xerj-engine --lib fts_projection_tests` -> "27 passed; 0 failed" (covers filter/should/multi_match/query_string/keyword-prefix projection — none regressed by the !projected_any -> !projected_positive swap).

Fail-before (test is a genuine guard): re-added `projected_positive = true;` inside the must_not loop (restores pre-fix `projected_any` semantics) -> new test FAILED post-flush: left `(0, [])` right `(2, ["2","3"])` at index.rs:1708 — exactly the #797 bug. Restored via `git checkout -- ...`; gate confirmed set only at must/filt
...(truncated)
```

- nit: Branch is 1 behind origin/main (main commit 2f209f9a #799/#801). It merges cleanly with no conflicts, but not strictly even — a pre-merge rebase/merge would keep CI honest per the rebase/merge-gate memory.

**Gate:** 0 blocking findings + all 19 checks green + CLEAN. The branch was 1 commit behind main (#800 landed after this CI ran); it merges cleanly (disjoint functions: FTS projection vs Terms rewrite) — verified the combined tree compiles + full suite + fmt/clippy green locally under 1.97.1 before merging.

PR #803, opened 2026-08-26
fix(engine): #802 function_score per-function filter is schema-aware (retires the schemaless matcher in prod)

### LINKED ISSUE #802: es-compat (#423 follow-up): function_score per-function filter uses the schemaless matcher, so a prefix/wildcard on a keyword field folds case
Surfaced by the #795 (#794) adversarial review (non-blocking nit). `apply_function_score`s per-function `filter` evaluation still calls the schemaless `doc_matches_query` (EMPTY_SCHEMA), so a `function_score` function `filter: {prefix: {keyword_field}}` (or `wildcard`) scores CASE-INSENSITIVELY (fold-both fallback), diverging from the case-SENSITIVE keyword semantics the main-query paths now use after #795.

Not a flush-divergence (identical on memtable and segment) and out of #794 scope, so #795 correctly left it. It is one more instance of the #423 root cause (a schemaless `doc_matches_query` call site that should be schema-aware).

Fix: thread the search`s owned `dmq_schema` into `apply_function_score` (same pattern as #795 threaded `try_shortcut_count` / `scan_stored_section_into`) and route the per-function filter check to `doc_matches_query_typed(q, source, &schema)`. Prove fail-before with a `function_score` whose function filter is `prefix` on a keyword field where the folded vs case-sensitive answer differs (e.g. filter prefix "ap" must NOT boost "Application").

Refs #423.

xerj-org:
Closes #802 (surfaced as a non-blocking nit in the #795/#794 adversarial review).

`apply_function_score` matched each function`s `filter` with the schemaless `doc_matches_query` (EMPTY_SCHEMA), so a keyword `prefix`/`wildcard` function filter case-FOLDED: a `function_score` function `{filter: {prefix: {k: "ap"}}, weight: 5}` wrongly boosted `"Application"` (which does not start with lowercase `"ap"`), diverging from the case-sensitive keyword semantics the query paths use after #795. Flush-invariant (identical on memtable and segment).

## Fix
Thread the search`s owned `dmq_schema` (`self.schema().await`) into `apply_function_score` and its three callers (two in `search_inner`, one in `score_query_against_doc`s `rec`), routing the per-function filter check to `doc_matches_query_typed(filter, source, schema)`.

This was the **last production caller** of the schemaless `doc_matches_query` shim (the #423 root cause: "term matching has two implementations, one schemaless"). With it threaded, the shim has zero prod callers, so it is gated to `#[cfg(test)]` — the lib target now has **no schemaless matcher path** (remaining callers are unit-test assertions of schema-independent shapes).

## Verification (rustc 1.97.1, unit harness — no :9200)
- Fail-before: `function_score` with function filter `prefix "ap"`, `boost_mode: replace` over `apple/Application/banana` → scores `[("1",5.0),("2",5.0),("3",1.0)]` (Application wrongly boosted); with the fix `[("1",5.0),("2",1.0),("3",1.0)]`, pre AND post flush.
- Full `xerj-engine` lib suite 646/0
- `cargo fmt --all --check` clean; `clippy -p xerj-engine --all-targets -D warnings` clean (the `#[cfg(test)]` gate removes the now-dead lib-target function).

Refs #423.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1719,6 +1719,62 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn funcscore_filter_prefix_on_keyword_is_case_sensitive() {
+        // #802 (#423 family): a function_score per-function `filter` was matched
+        // with the SCHEMALESS doc_matches_query, so a keyword `prefix` filter
+        // case-FOLDED — "Application" wrongly got the weight-5 boost for prefix
+        // "ap" (lowercase) it does not start with. Must be case-sensitive: only
+        // "apple" is boosted. Flush-invariant (same on memtable + segment).
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("k", FieldType::Keyword))
+            .unwrap();
+        engine.create_index("fs802", schema).unwrap();
+        let idx = engine.get_index("fs802").unwrap();
+        idx.abort_background_tasks();
+        for (id, k) in [("1", "apple"), ("2", "Application"), ("3", "banana")] {
+            idx.index_document(Some(id.into()), json!({ "k": k }))
+                .await
+                .unwrap();
+        }
+        let run = |idx: std::sync::Arc<crate::Index>| async move {
+            let q = json!({"function_score": {
+                "query": {"match_all": {}},
+                "functions": [{"filter": {"prefix": {"k": "ap"}}, "weight": 5.0}],
+                "boost_mode": "replace"
+            }});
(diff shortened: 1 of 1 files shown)


xerj-org:

**correctness — HOLDS.** All 3 apply_function_score call sites (2 columnar via &dmq_schema=self.schema().await, 1 rec via score_query_against_doc's threaded schema) pass the real schema. Keyword prefix/wildcard function filter now case-SENSITIVE (Application no longer boosted for "ap"); declared TEXT stays case-insensitive; unknown/empty-schema shim keeps fold-both. pre==post flush on columnar + nested-bool rec. 5 refutation tests + PR test all pass.

**regression — HOLDS.** Full xerj-engine lib 649/0 under 1.97.1; fmt --check exit 0; clippy --all-targets -D warnings exit 0. The #[cfg(test)] doc_matches_query shim has zero prod callers (cargo check --lib compiles clean; all ~35 invocations under cfg(test)). Test-merges with origin/main (3 behind, #804 metrics) zero conflicts.

**evidence — HOLDS.** Fail-before + last-prod-caller lib build + fixed-test pass all verified. Test funcscore_filter_prefix_on_keyword_is_case_sensitive asserts [("1",5.0),("2",1.0),("3",1.0)] pre AND post flush.

Nits (non-blocking): stale /// prose at index.rs:36829-36833 still lists collect_matched_queries_inner + apply_function_score as no-schema shim callers (both now thread real schema); a direct wildcard-on-keyword filter test would make the wildcard half explicit. Merging.

PR #800, opened 2026-08-26
fix(engine): #799/#801 terms on date/ip fields normalizes values like single-term (dates + CIDRs)

xerj-org:
Closes #799 and #801. Follow-up to #788/#789 (date) and #782/#783 (CIDR).

The `rewrite_query_aliases` `Terms` arm did not normalize its values the way the single-`term` arm does, so `terms` missed matches `term` finds:
- **#799 (date):** a `terms` naming the same instants in an alternate format (date-only string, ISO variant, epoch-ms number) returned `0`.
- **#801 (ip):** a `terms` of CIDRs on an `ip` field returned `0` (the segment term dictionary cannot expand a subnet).

## Fix
Unify the `Terms`-arm normaliser to mirror the single-`term` arm:
- declared `date` field → each date-parseable value becomes `range [epoch, epoch]` (sub-ms strings stay literal — the #790 nanos-shadow gate);
- declared `ip` field → each CIDR becomes `range [network, broadcast]` (`cidr_to_ip_range`).

Rewritten values are OR-combined via `bool{should:[…]}` (default minimum_should_match 1 = ES `terms` OR). Non-rewritable values (sub-ms dates, plain non-CIDR ips, anything that does not parse) are preserved in a residual exact `terms` should-clause so nothing is lost. Only fires on a declared `date`/`ip` field with ≥1 rewritable value. Range resolves through the same normaliser on every path (memtable + segment + FTS), so one rewrite fixes them all.

## Verification (rustc 1.97.1, unit harness — no :9200)
- `#799` fail-before `terms{ts:["2020-01-01",1623758400000]}` → `(0,[])`; fixed → `(2,["1","2"])` pre+post flush.
- `#801` fail-before `terms{addr:["10.0.0.0/8","192.168.0.0/16"]}` → `(0,[])`; fixed → `(2,["1","2"])`; mixed CIDR+plain-ip → `(2,["1","3"])` pre+post flush.
- Full `xerj-engine` lib suite 646/0; `fmt --all --check` clean; `clippy -p xerj-engine --all-targets -D warnings` clean.

PR #813, opened 2026-08-26
fix(engine): #799/#801 terms (plural) rewrites CIDR-on-ip and date alt-formats too

### LINKED ISSUE #799: es-compat: terms (plural) on a date field misses alt-format values (returns 0 for equivalent instants) — #788/#789 residual
Follow-up to #788/#789 (which fixed single `term` on a date field). `terms` on a declared `date` field does a raw literal lookup per value and never normalizes date formats, so a `terms` naming the same instants in an alternate representation (date-only string, ISO variant, or epoch-ms number) returns **zero** — both pre and post flush (a consistent wrong-result, not a flush-divergence).

## Repro
Declared `date` field `ts`, docs `1=2020-01-01T00:00:00Z`, `2=2021-06-15T12:00:00Z`, `3=2019-03-03T00:00:00Z`. Query:
```json
{"terms": {"ts": ["2020-01-01", 1623758400000]}}
```
(date-only form of doc 1s instant + epoch-ms of doc 2s instant). Expected `total=2, hits=[1,2]`.

Observed (main `2155784b`, rustc 1.97.1, unit harness — no :9200):
```
terms-on-date pre : (0, [])
terms-on-date post: (0, [])
```

## Fix direction
`rewrite_query_aliases` (index.rs ~34482) rewrites a single date `term` to a degenerate `range [epoch, epoch]` (#789) but the `Terms` arm just clones values. Extend it: for a declared `date` field, rewrite each date-parseable, non-sub-ms value to a `range [epoch, epoch]` and combine via `bool{should:[…], minimum_should_match:1}`; keep sub-ms strings (#790 nanos-shadow hazard) and non-date-parseable values in a residual `terms`. Same gate as the single-term fix (`FieldType::Date` + `parse_date_ms` + `!date_string_has_subms`). Range-on-date resolves through `parse_date_ms` on every path, so one rewrite fixes memtable + segment + FTS.

Severity: wrong-results (a common filter shape — timestamps by set — silently returns nothing).

### LINKED ISSUE #801: es-compat: terms (plural) with CIDR values on an ip field returns 0 (single-term CIDR works) — #782/#783 residual
Sibling of #799 (date) in the same `rewrite_query_aliases` `Terms` arm. The single-`term` CIDR-on-`ip` rewrite (#782/#783) expands a CIDR to its `[network, broadcast]` range, but the `Terms` arm never got it, so `terms` with CIDR values silently returns zero — pre and post flush.

## Repro
Declared `ip` field `addr`, docs `1=10.0.0.5`, `2=192.168.1.9`, `3=172.16.0.1`.
```
term  {addr: "10.0.0.0/8"}                       -> (1,["1"])   correct
terms {addr: ["10.0.0.0/8","192.168.0.0/16"]}    -> (0,[])      WRONG (want (2,["1","2"]))
```
(main `2155784b`, rustc 1.97.1, unit harness — no :9200; same pre and post flush.)

## Fix
Same shape as #799: in the `Terms` arm, when the field is a declared `ip` and a value is a CIDR, rewrite it to `range [network, broadcast]` (`cidr_to_ip_range`, as the single-`term` arm does) and OR the values via `bool.should`; plain-ip and non-CIDR values stay an exact residual `terms`. Fixed together with #799 in the same PR (same code arm).

Vinz2168:
Closes #799, closes #801.

Sibling of the single-`term` rewrites (#782/#783 CIDR-on-ip, #788/#789 date-instant). The `Terms` arm of `rewrite_query_aliases` only resolved field aliases and cloned values, so `terms` never got either rewrite:

- `terms` with CIDR values on a declared `ip` field returned 0 (#801) — the single-`term` CIDR rewrite expands to `[network, broadcast]`, but the `Terms` arm never got it.
- `terms` on a declared `date` field naming the same instants in an alternate format (date-only string, epoch-ms) returned 0 (#799) — the `Terms` arm did a raw literal lookup per value with no normalization.

Both are wrong-results bugs, not flush-divergences: pre- and post-flush agree on the (wrong) zero.

## Fix

In the `Terms` arm, split values into ones that rewrite (CIDR -> range, date-parseable -> degenerate range `[epoch, epoch]`) and a residual exact `terms` for the rest, then OR them via `bool.should` + `minimum_should_match: 1` — the same building blocks the single-term arm already established. Sub-ms date strings (nanos-shadow hazard, #789) and non-CIDR/non-date values stay in the residual `terms`, matched exactly as before.

## Tests

Two new tests in `flush_publication_recovery_tests`, mirroring the existing single-term tests' style (real `Engine`/`Index`, pre- and post-flush, `size:10` hits + ids):
- `terms_on_a_date_field_normalizes_alt_formats_across_flush`
- `terms_with_cidr_values_on_an_ip_field_rewrites_each_to_a_range`

Both mix a rewritten value with a residual exact value in the same `terms` array to confirm the two paths combine correctly.

Fail-before verified by disabling each rewrite's outer condition: both new tests fail with `(0, [])` instead of the expected pairs.

xerj-engine lib suite 643/645 (2 pre-existing, unrelated `snapshot_path_security_tests` failures on macOS, reproduce identically on `upstream/main` without this change — `/tmp -> /private/tmp` symlink resolution); fmt clean; clippy `-D warnings` clean.

https://claude.ai/code/session_01F4vw8pibaJgnX5eD1TtUJE

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1498,6 +1498,126 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn terms_on_a_date_field_normalizes_alt_formats_across_flush() {
+        // #799: `terms` on a declared `date` field did a raw literal lookup
+        // per value and never normalized formats, so naming the same instants
+        // in an alternate representation (date-only string, epoch-ms number)
+        // returned 0 — the sibling of the #788/#789 single-`term` fix, but
+        // the `Terms` arm never got it.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        let mut schema = Schema::empty();
+        schema
+            .add_field(FieldConfig::new("ts", FieldType::Date))
+            .unwrap();
+        engine.create_index("date-terms", schema).unwrap();
+        let idx = engine.get_index("date-terms").unwrap();
+        idx.abort_background_tasks();
+        for (id, ts) in [
+            ("a", "2020-01-01T00:00:00Z"),
+            ("b", "2021-06-15T12:00:00Z"),
+            ("c", "2019-03-03T00:00:00Z"),
+            // Sub-millisecond precision must stay an exact residual match,
+            // same nanos-shadow hazard as the single-`term` fix (#789).
+            ("d", "2020-03-03T03:03:03.123456Z"),
+        ] {
+            idx.index_document(Some(id.into()), json!({ "ts": ts }))
+                .await
+                .unwrap();
+        }
+        let run = |idx: std::sync::Arc<crate::Index>, body: serde_json::Value| async move {
+            let req = xerj_query::parse_request(&json!({
+                "query": body, "size": 10, "track_total_hits": true
+            }))
+            .unwrap();
+            let r = idx.search(&req).await.unwrap();
+            let mut ids: Vec<_> = r.hits.iter().map(|h| h.id.clone()).collect();
+            ids.sort();
+            (r.total.value, ids)
+        };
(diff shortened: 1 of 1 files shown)


Vinz2168:
Superseded — xerj-org's own bot opened and merged #800 for the same issues (#799/#801) while this was open. Closing to avoid confusion.

PR #814, opened 2026-08-26
fix(autoindex): index top-level constants as symbols across 7 languages (#500)

xerj-org:
Part of #500 (reference-coding recall). Completes the top-level-constant capture sweep begun by the Java-constant fix #605 and the TS/JS `export const` fixes #285/#293 — bringing **every major language** to parity.

## Gap
`#500` measured (on Apache Lucene/Java) that a named constant like `DEFAULT_MAX_CONN` not indexed as its own symbol costs its ~2 KB parent or is length-normalised out of reach (32–48× more bytes/answer than grep). #605 fixed Java; Rust/Go/TS/JS/C already had it. But **six languages captured NO top-level constants**: Python, C#, Ruby, Kotlin, Swift, Scala (PHP already covered). This PR closes all six.

## Fix (`extract/code.rs`), one tested pattern per language
| lang | pattern | precision boundary |
|---|---|---|
| Python | `(module (expression_statement (assignment left: (identifier) @const)))` | `module`-anchored → excludes function-locals |
| C# | `const`/`static` field + `enum_member_declaration` | `const`/`static` gate → excludes instance `readonly` |
| Ruby | `(assignment left: (constant) @const)` | `(constant)` node is UPPER_CASE-only → never a local |
| Kotlin | source_file + object-body `property_declaration` | anchored → excludes locals + class instance state |
| Swift | top-level `let` + `static let` type members + enum cases | `static` gate → excludes instance `let` |
| Scala | compilation-unit + object-body `val_definition` | anchored → excludes function-local `val` |

Each promoted declaration becomes its own `const` symbol whose `code` is the ~40–80 B declaration line, not a parent body — the #500 unit-of-retrieval fix, now applied wherever a language named its constants.

- Six new tests (`{python,csharp,ruby,kotlin,swift,scala}` constants), each fail-before proven (patterns removed → only pre-existing kinds).
- Full `xerj-autoindex` lib suite **731/0**; `fmt --all --check` clean; `clippy -p xerj-autoindex --all-targets -D warnings` clean.

Refs #500 (the whole-symbol-payload effect on method BODIES remains a separate structural follow-up).

_(Supersedes #812, which never received a CI run — a GitHub `pull_request`-event miss during a runner-saturation window; same commits, fresh branch to get a clean CI trigger.)_

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -892,6 +892,7 @@ const KOTLIN_Q: &str = r#"
 (type_alias type: (identifier) @type)
 (source_file (property_declaration (variable_declaration (identifier) @const)))
 (object_declaration (class_body (property_declaration (variable_declaration (identifier) @const))))
+(companion_object (class_body (property_declaration (variable_declaration (identifier) @const))))
 "#;
 
 // Adapted from tree-sitter-swift 0.7.3 queries/tags.scm (class_declaration /
@@ -2027,14 +2028,20 @@ mod tests {
             "const val MAX = 100\n\
              val Host = \"x\"\n\
              object Cfg {\n  const val TIMEOUT = 30\n}\n\
+             class Db {\n  companion object {\n    const val DEFAULT_MAX_CONN = 100\n  }\n  val instance = 1\n}\n\
              fun f() {\n  val local = 1\n}\n",
         );
         assert!(has(&s, "MAX", "const"), "got {s:?}");
         assert!(has(&s, "Host", "const"), "got {s:?}");
         assert!(has(&s, "TIMEOUT", "const"), "got {s:?}");
-        // Function-local `val` stays out (anchored to file / object scope).
+        // companion object const — the idiomatic Kotlin class-scoped constant
+        // (direct analog of the Java `static final` #500 measured).
+        assert!(has(&s, "DEFAULT_MAX_CONN", "const"), "got {s:?}");
+        // Function-local `val` and class INSTANCE state stay out.
         assert!(!has(&s, "local", "const"), "got {s:?}");
+        assert!(!has(&s, "instance", "const"), "got {s:?}");
         assert!(has(&s, "Cfg", "object"));
+        assert!(has(&s, "Db", "class"));
         assert!(has(&s, "f", "function"));
     }
 
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -816,6 +816,7 @@ const RUBY_Q: &str = r#"
 (singleton_method name: (identifier) @method)
 (class name: (constant) @class)
 (module name: (constant) @module)
+(assignment left: (constant) @const)
 "#;
 
 // `enum_declaration` was missing entirely, which is worse than a missing kind:
@@ -1838,6 +1839,31 @@ mod tests {
         assert!(has(&s, "m", "method"));
         assert!(has(&s, "M", "module"));
     }
+
+    /// #500: `RUBY_Q` captured method/class/module but not CONSTANTS, so a
+    /// `MAX_CONN = 100` (the `DEFAULT_MAX_CONN`-class fact #500 measured) was
+    /// indexed nowhere. Ruby's grammar makes this precise for free: a constant
+    /// assignment binds `left: (constant)` — a distinct node from a local
+    /// variable's `(identifier)` — so `(assignment left: (constant))` captures
+    /// UPPER_CASE constants at any scope and NEVER a lowercase local, no anchor
+    /// needed (unlike Python/module).
+    #[test]
+    fn ruby_constants() {
+        let s = syms(
+            "ruby",
+            "MAX = 100\n\
+             HOST = \"x\"\n\
+             class C\n  TABLE = [1, 2]\n  def m\n    local = 1\n  end\nend\n",
+        );
+        assert!(has(&s, "MAX", "const"), "got {s:?}");
+        assert!(has(&s, "HOST", "const"), "got {s:?}");
+        assert!(has(&s, "TABLE", "const"), "got {s:?}");
+        // A lowercase local variable is `(identifier)`, never `(constant)`.
+        assert!(!has(&s, "local", "const"), "got {s:?}");
+        assert!(has(&s, "C", "class"));
+        assert!(has(&s, "m", "method"));
+    }
+
     #[test]
     fn php() {
         let s = syms(
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -885,6 +885,8 @@ const KOTLIN_Q: &str = r#"
 (object_declaration name: (identifier) @object)
 (function_declaration name: (identifier) @function)
 (type_alias type: (identifier) @type)
+(source_file (property_declaration (variable_declaration (identifier) @const)))
+(object_declaration (class_body (property_declaration (variable_declaration (identifier) @const))))
 "#;
 
 // Adapted from tree-sitter-swift 0.7.3 queries/tags.scm (class_declaration /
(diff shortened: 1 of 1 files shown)


xerj-org:

The **correctness** skeptic found a BLOCKING under-capture (Kotlin companion-object constants) — **now fixed** in `6099ad2b` (added `(companion_object (class_body (property_declaration (variable_declaration (identifier) @const))))`, the skeptic's own verified one-liner; `kotlin_constants` extended to assert `DEFAULT_MAX_CONN` is captured and class instance state is not; suite 732/0). Regression + evidence HOLDS. Verdicts verbatim:

### Correctness
**Verdict:** REFUTED — **blocking:** True

**Summary:** Kotlin UNDER-captures: a `companion object { const val DEFAULT_MAX_CONN = 100 }` — the idiomatic Kotlin class-scoped constant and the direct translation of the exact Java `static final` fact #500 measured (which Java #605 captures) — is indexed NOWHERE. `const val` is legal only at top-level, in an `object`, or in a `companion object`; the PR captures the first two (source_file + object_declaration body) but the companion object is a distinct `companion_object` AST node that NEITHER Kotlin pattern matches, so class-scoped Kotlin constants re-create the #500 gap the PR claims to close. Fail-before: for `class Db { companion object { const val DEFAULT_MAX_CONN = 100 } }`, syms = only [("Db","class")] — the constant is captured by nothing. One-line fix verified: `(companion_object (class_body (property_declaration (variable_declaration (identifier) @const))))` compiles (7→8 patterns), captures DEFAULT_MAX_CONN, and still excludes class instance state (`val instance = 1` stays out). The other 6 languages' capture/precision boundaries hold up under probing; only secondary nits there.

**Evidence:**
```
Toolchain: `rustc 1.97.1 (8bab26f4f 2026-07-14)`. HEAD = 5c26db03 (fix/500-const-symbols). Baseline `cargo test -p xerj-autoindex --lib extract::code::tests::` = 66 passed.

FAIL-BEFORE (added test, run under 1.97.1):
  syms("kotlin", "class Db {\n companion object {\n  const val DEFAULT_MAX_CONN = 100\n }\n}\n")
  -> panicked: `companion-object constant indexed NOWHERE: got [("Db", "class", 1, "class Db {")]`
  test kotlin_companion_object_constant_is_missed ... FAILED

AST proof (to_sexp of the same src):
  (source_file (class_declaration name:(identifier) (class_body
    (companion_object (class_body (property_declaration
      (modifiers (property_modifier)) (variable_declaration (identifier)) (number_literal)))))))
  -> the const val lives under a `companion_object` node; PR's two Kotlin patterns key on `source_file` and `object_declaration`, neither of which is `companion_object`.

FIX PROOF (running KOTLIN_Q + one extra pattern directly against the tree):
  extra = (companion_object (class_body (property_declaration (variable_declaration (identifier) @const))))
  FIXED CAPTURES: [("Db","class"), ("DEFAULT_MAX_CONN","const")]   (instance `val instance = 1` correctly NOT captured)
  -> Query::new compiles, pattern_count 7 -> 8; captures the companion const, excludes instance state.

Contrast — standalone object works (PR test `kotlin_constants` passes): `object Cfg { const val TIMEOUT = 30 }` IS captured. Only the companion (class-scoped) form is missed. Java equivalent IS captured (java_static_constants_are_symbols passes). So the miss is Kotlin-specific and re-opens 
...(truncated)
```

- nit: C++ class-scope `static const int Member = 1;` inside a class body is also missed (same class-scoped-constant category as the Kotlin companion miss) — the PR scoped C++ to translation_unit+namespace only. Lower priority than Kotlin since C++ class constants are less dominant than Kotlin companion-object constants.
- nit: C++ `(enumerator name: (identifier) @const)` is unanchored: a FUNCTION-LOCAL enum (`void f(){ enum E{LX,LY}; }`) leaks LX/LY as `const` symbols into the cross-file surface. Minor over-capture.
- nit: Top-level mutable bindings are labeled `const`: Kotlin top-level `var`, Swift top-level `var`, and any C++ mutable global with an initializer (`int counter = 0;`) all become `const` symbols. Consistent with the Go-`var` precedent but the kind label is a slight misnomer.
- nit: Python `NAME: str` (a bare annotated declaration with NO value) is captured as a `const` even though nothing is bound. Minor over-capture.

### Regression

**Summary:** PR #814 adds top-level/namespace constant capture for Python, C#, Ruby, Kotlin, Swift, Scala, C++ in code.rs only. No regression found under rustc 1.97.1.

**Evidence:**
```
cargo test -p xerj-autoindex --lib => 732 passed; 0 failed. all_queries_compile ok. All existing per-language tests and 7 new tests ok. fmt --check exit 0. clippy --all-targets -D warnings exit 0. merge with origin/main clean.
```


### Evidence-holds

**Summary:** Tried to refute PR #814 and could not. Under rustc 1.97.1 the whole extract::code::tests module is 66/66 green including all 7 new tests. Fail-before is LOAD-BEARING for every one of the 7 languages: removing a language's added pattern line(s) makes ONLY that test fail, with the constants dropping out of the capture set while the pre-existing function/class/etc kinds remain. All 7 tests assert the NEGATIVE (a function-local or class-instance binding is NOT captured), and I confirmed the negatives are discriminating rather than trivially-passing (unanchoring the Python pattern wrongly captures local_v and the negative assert fires). No BLOCKING defect.

**Evidence:**
```
Env: cd engine && RUST_MIN_STACK=8388608 rustup run 1.97.1 cargo test -p xerj-autoindex --lib ... ; branch reset --hard origin/fix/500-const-symbols, HEAD=5c26db03. Tree clean before/after (git status --porcelain empty).

BASELINE (all green): `cargo test -p xerj-autoindex --lib extract::code::tests` => "test result: ok. 66 passed; 0 failed". python_module_level_constants, csharp_static_constants_and_enum_members, ruby_constants, kotlin_constants, swift_constants, scala_val_constants, cpp_constants all ok.

FAIL-BEFORE per language (remove added pattern line(s), run only that test, then git checkout -- to restore — safe because the fix is committed at HEAD, no uncommitted work):

1) PYTHON — removed `(module (expression_statement (assignment left: (identifier) @const)))`:
FAILED at code.rs:1446. got [("f","function",4,"def f():"),("C","class",7,"class C:")] — MAX_CONN/DATABASE_URL/ROUTES gone.

2) RUBY — removed `(assignment left: (constant) @const)`:
FAILED at code.rs:1897. got [("C","class",3,..),("m","method",5,..)] — MAX/HOST/TABLE gone.

3) SWIFT — removed only the static-gate line `(property_declaration (modifiers (property_modifier) @_m) (pattern (simple_identifier) @const)) (#eq? @_m "static")`:
FAILED at code.rs:2077. got list has TopLevel/north/south but NOT MaxConn nor Url (the `static let` type members) — static-modifier-gate is load-bearing.

4) C++ — removed the two pointer_declarator lines:
FAILED at code.rs:1864. got has MaxConn/Timeout/Red/Green but NOT Host (the `const char* Host`) — pointer_declarator shape is load-bearing.

5) C# — removed both `(field_d
...(truncated)
```

- nit: C# `!local` and Swift `!local`/`!instanceName` negatives are structurally excluded (different node type / no modifiers node) rather than by the gate predicate; the discriminating negative in those tests is C# `!Instance` and the Swift static-gate fail-before — still adequately covered
- nit: cargo test accepts only one TESTNAME filter, so the 7 tests can't be listed in one invocation; ran them via the shared module filter extract::code::tests (all 7 present and green)

Non-blocking nits filed as a follow-up issue. Re-running the skeptic gate on the fixed head before merge (needs 0 blocking + green + CLEAN).

xerj-org:

**correctness — HOLDS.** Probed all 7 langs for (a) real top-level const captured, (b) function-local NOT, (c) instance field NOT. Both precision boundaries (direct locals + non-static instance state) hold everywhere. Residual over-captures are the benign mutable-top-level-labeled-const family (Kotlin var, Swift static var, C++ int x=0) — real retrievable symbols with an imperfect kind, mirroring Go/Rust/C. None miss a const or inject local/instance noise.

**regression — HOLDS.** xerj-autoindex full lib 732/0 under 1.97.1; all_queries_compile passes; every pre-existing per-language test still passes (patterns didn't alter prior captures). fmt --all --check clean, clippy --all-targets -D warnings clean. Merges clean with main. Diff isolated to code.rs (+228).

**evidence — HOLDS.** Baseline 7/7 pass. Fail-before for ALL 7 languages: removing each language's added const pattern makes only that language's test fail with a clean assertion (queries still compile, pre-existing kinds still captured, only const/enum symbols vanish) — each pattern load-bearing. Each test asserts the negative (local/instance NOT captured).

Nits (non-blocking, deferred): mutable-top-level kind=const labeling (#820); C++ unanchored enumerator + Kotlin local-object val leak (rare, genuine constants); Swift static-let double-capture (deduped by locator). Merging.

PR #815, opened 2026-08-26
fix(autoindex): index top-level constants as symbols across 7 languages (#500)

xerj-org:
Part of #500 (reference-coding recall). Completes the top-level-constant capture sweep begun by the Java-constant fix #605 and the TS/JS `export const` fixes #285/#293 — bringing **every major language** to parity.

## Gap
`#500` measured (on Apache Lucene/Java) that a named constant like `DEFAULT_MAX_CONN` not indexed as its own symbol costs its ~2 KB parent or is length-normalised out of reach (32–48× more bytes/answer than grep). #605 fixed Java; Rust/Go/TS/JS/C already had it. But **six languages captured NO top-level constants**: Python, C#, Ruby, Kotlin, Swift, Scala (PHP already covered). This PR closes all six.

## Fix (`extract/code.rs`), one tested pattern per language
| lang | pattern | precision boundary |
|---|---|---|
| Python | `(module (expression_statement (assignment left: (identifier) @const)))` | `module`-anchored → excludes function-locals |
| C# | `const`/`static` field + `enum_member_declaration` | `const`/`static` gate → excludes instance `readonly` |
| Ruby | `(assignment left: (constant) @const)` | `(constant)` node is UPPER_CASE-only → never a local |
| Kotlin | source_file + object-body `property_declaration` | anchored → excludes locals + class instance state |
| Swift | top-level `let` + `static let` type members + enum cases | `static` gate → excludes instance `let` |
| Scala | compilation-unit + object-body `val_definition` | anchored → excludes function-local `val` |

Each promoted declaration becomes its own `const` symbol whose `code` is the ~40–80 B declaration line, not a parent body — the #500 unit-of-retrieval fix, now applied wherever a language named its constants.

- Six new tests (`{python,csharp,ruby,kotlin,swift,scala}` constants), each fail-before proven (patterns removed → only pre-existing kinds).
- Full `xerj-autoindex` lib suite **731/0**; `fmt --all --check` clean; `clippy -p xerj-autoindex --all-targets -D warnings` clean.

Refs #500 (the whole-symbol-payload effect on method BODIES remains a separate structural follow-up).

_(Supersedes #812, which never received a CI run — a GitHub `pull_request`-event miss during a runner-saturation window; same commits, fresh branch to get a clean CI trigger.)_

_(Re-homed from fix/500-const-symbols (#814), which — like #812 — never received a CI run: a GitHub pull_request-event delivery miss specific to the fix/500-* refs. Same commits, differently-named branch to test whether a clean ref triggers CI.)_

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -816,6 +816,7 @@ const RUBY_Q: &str = r#"
 (singleton_method name: (identifier) @method)
 (class name: (constant) @class)
 (module name: (constant) @module)
+(assignment left: (constant) @const)
 "#;
 
 // `enum_declaration` was missing entirely, which is worse than a missing kind:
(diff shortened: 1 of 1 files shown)


xerj-org:
Duplicate of #814 (same #500 const-sweep commits); both blocked by the same GitHub pull_request-event delivery miss. Consolidating on #814.

PR #818, opened 2026-08-26
fix(metrics): #804 wire xerj_bytes_written_total (was registered but never incremented)

### LINKED ISSUE #804: metrics: xerj_bytes_written_total is registered but never incremented
Reported by **Vincenzo Lombardo** (Operations Manager, Seacom Srl), who tested the published `rc.4` binary (aarch64-apple-darwin) natively rather than a self-built Docker image, specifically to rule out drift between his build and what we ship.

His observation: `xerj_bytes_written_total` stays at `0` regardless of write volume.

## Confirmed on main

`engine/crates/xerj-common/src/metrics.rs:124` registers the counter:

```rust
let bytes_written = IntCounter::with_opts(Opts::new(
    "xerj_bytes_written_total",
    "Total bytes written to segment files (uncompressed)",
))
```

Grepping the whole engine for anything that increments it returns nothing outside `metrics.rs` itself. There is no `record_bytes_written` call site. The counter is exported to Prometheus, is documented as "Total bytes written to segment files", and is always zero. Anyone graphing write throughput gets a flat line and no indication it is unimplemented.

Either wire it to the segment-write path, or drop it from the registry so it does not read as a working metric.

## Related, needs re-check rather than assumed broken

He also measured `xerj_docs_indexed_total` undercounting on rc.4: `2` against a real `doc_count` of `23` after a mix of PUT/bulk/update writes across several indices. On current main the single-doc paths do call `record_doc_indexed` (`es_compat.rs` lines 2840, 2978, 3015, 3094, 17492, 17530) alongside the bulk tally at 15277, so this may already be addressed since rc.4. Worth reproducing before acting.

He explicitly corrected several of his own earlier findings after retesting against the shipped artifact, and flagged these two as the ones he would still act on. Thanks to Vincenzo for testing against the real binary and for withdrawing the parts that did not hold up.

xerj-org:
Closes #804. Reported by Vincenzo Lombardo (Seacom Srl) against the shipped `rc.4` binary: `xerj_bytes_written_total` stays at `0` regardless of write volume.

## Root cause
`metrics.rs` registers the `IntCounter` but there is **no increment call site** anywhere in the engine — the metric is exported, documented as "Total bytes written to segment files", and always zero. Anyone graphing write throughput gets a flat line with no indication it is unimplemented.

## Fix
The engine already exposes a process-wide metrics handle via `engine_metrics()` (a no-op when unset, e.g. unit tests). Both segment-producing paths finalize a `SegmentMeta` (which carries `size_bytes` = the sealed `.seg` on-disk size) in engine code:
- **flush** — `do_flush_shard`s `build_fts` closure receives the sealed `meta`;
- **merge** — `merge_pass_locked`s `writer.finish()` returns `merged_meta`.

Record `size_bytes` at both via a new `Metrics::record_bytes_written(n)` helper. Also corrected the help text: the value is the on-disk `.seg` size, not "uncompressed" (dropped the misleading qualifier).

Chose to WIRE it (the issues preferred option) rather than drop it — a working write-throughput counter is worth more than a removed one.

- New test `bytes_written_metric_increments_on_flush`: installs a metrics handle, indexes 5 docs + flushes, asserts the counter strictly increases. Fail-before (both increments removed) → `0 -> 0`, test fails.
- `xerj-common` suite 91/0; `xerj-engine` lib suite 649/0.
- `cargo fmt --all --check` clean; `clippy -p xerj-engine -p xerj-common --all-targets -D warnings` clean.

Note: counts the main `.seg` file (`size_bytes`); FTS/DV sidecar files are not included — a defensible "segment files" reading and a huge improvement over always-zero. The related `xerj_docs_indexed_total` undercount the reporter saw on rc.4 was already addressed on main (single-doc paths call `record_doc_indexed`); left as-is.

### DIFF
--- a/engine/crates/xerj-common/src/metrics.rs
+++ b/engine/crates/xerj-common/src/metrics.rs
@@ -122,7 +122,7 @@ impl Metrics {
 
         let bytes_written = IntCounter::with_opts(Opts::new(
             "xerj_bytes_written_total",
-            "Total bytes written to segment files (uncompressed)",
+            "Total on-disk bytes written to segment (.seg) files by flush and merge",
         ))
         .map_err(|e| XerjError::internal(format!("metrics: {e}")))?;
 
@@ -354,6 +354,18 @@ impl Metrics {
         self.docs_indexed_by_index
             .with_label_values(&[index])
             .inc_by(n);
+    }
+
+    /// Record `n` on-disk bytes written to a segment (.seg) file — called
+    /// once per segment finalized by flush or merge with the sealed segment's
+    /// `size_bytes`. A no-op when `n == 0`. (#804: the counter was registered
+    /// but never incremented, so it read a flat zero regardless of write
+    /// volume.)
+    pub fn record_bytes_written(&self, n: u64) {
+        if n == 0 {
+            return;
+        }
+        self.bytes_written.inc_by(n);
     }
 
     /// Drop the per-index label series for `index` (RC4-W4 item 5).
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1891,6 +1891,40 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn bytes_written_metric_increments_on_flush() {
+        // #804: xerj_bytes_written_total was registered but never incremented,
+        // so it read a flat zero regardless of write volume. Flush now records
+        // each sealed segment's on-disk size via engine_metrics().
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        // Idempotent OnceLock: installs a handle if none yet, else reads the
+        // existing one. Either way the post-flush delta must be positive.
+        crate::set_engine_metrics(std::sync::Arc::new(
+            xerj_common::metrics::Metrics::new().unwrap(),
+        ));
+        let m = crate::engine_metrics().expect("engine metrics installed");
+        let before = m.bytes_written.get();
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine.create_index("bw804", Schema::empty()).unwrap();
+        let idx = engine.get_index("bw804").unwrap();
+        idx.abort_background_tasks();
+        for i in 0..5 {
+            idx.index_document(Some(format!("{i}")), json!({ "body": "some text content" }))
+                .await
+                .unwrap();
+        }
+        idx.flush().await.unwrap();
+        assert_eq!(idx.memtable.doc_count(), 0);
+        let after = m.bytes_written.get();
+        assert!(
+            after > before,
+            "#804: bytes_written must increase after a flush, got {before} -> {after}"
+        );
+    }
+
+    #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn portable_field_flush_keeps_baseline_data_dir_format() {
         let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
         let dir = tempfile::tempdir().unwrap();
@@ -10048,6 +10082,10 @@ impl Index {
                             return None;
                         }
                     };
+                    // #804: count the merge-written segment's on-disk bytes.
+                    if let Some(m) = crate::engine_metrics() {
+                        m.record_bytes_written(merged_meta.size_bytes);
+                    }
 
                     // Build FTS side-cars using the parallel per-field builder.
                     // We pre-built `fts_input` during the byte-copy pass above,
@@ -26403,6 +26441,11 @@ async fn do_flush_shard(
         .filter(|hook| hook.target_memtable == test_target)
         .and_then(|hook| hook.before_blocking_spawn.as_ref().map(Arc::clone));
     let build_fts = move |meta: &xerj_storage::segment::SegmentMeta| -> xerj_storage::Result<()> {
+        // #804: count the flush-written segment's on-disk bytes (the sealed
+        // `.seg` whose `size_bytes` this closure receives).
+        if let Some(m) = crate::engine_metrics() {
+            m.record_bytes_written(meta.size_bytes);
+        }
         #[cfg(test)]
         let test_callback = test_hook
             .as_ref()


xerj-org:

### Correctness

**Summary:** Could not refute PR #818. The metric is wired correctly and cannot systematically mis-count. (a) size_bytes IS the .seg on-disk length: segment.rs finish() writes exactly `buf` to the .seg (f.write_all(&buf), fsync, rename) and returns SegmentMeta{ size_bytes: buf.len() } — same struct/path for flush and merge. (b) No double-count: each site fires exactly once per sealed segment. FLUSH — build_fts is FnOnce, passed only to finalize_flush_with_publisher which invokes post_finish exactly once (index_store.rs:2106) after a single writer.finish(); MERGE — the per-task closure produces one merged_meta from one writer.finish() and increments once (index.rs:10086); the trailing FTS/DV builds never touch the counter. Flush and merge write DISTINCT physical files, so counting both is correct write-amplification accounting, not a double-count. (c) engine_metrics()==None is a clean `if let Some(m)` no-op at both sites — no unwrap/panic. (d) Counting-at-seal is correct: writer.finish() fsyncs the .seg to disk BEFORE the increment, so even an abandoned/rolled-back segment genuinely had its bytes written (defensible "bytes written to disk"); empty flush returns Empty before creating a writer, so it never counts. (e) record_bytes_written(0) early-returns — correct no-op (real .seg is never 0 bytes anyway). (f) Corrected help text is accurate: size_bytes is the compressed columnar on-disk buffer, so "on-disk bytes" is right and the old "(uncompressed)" was wrong. Test bytes_written_metric_increments_on_flush exists, is NOT #[ignore]d, and passes under 1.97.1.

**Evidence:**
```
$ git reset --hard origin/fix/804-bytes-written-metric; git log --oneline -1
HEAD is now at 314e510c fix(metrics): #804 wire xerj_bytes_written_total to the flush + merge segment writes

$ grep -n record_bytes_written engine/crates/xerj-common/src/metrics.rs engine/crates/xerj-engine/src/index.rs
metrics.rs:364:    pub fn record_bytes_written(&self, n: u64) {   // if n==0 { return } else bytes_written.inc_by(n)
index.rs:10087:  m.record_bytes_written(merged_meta.size_bytes);  // MERGE, after writer.finish() (10077)
index.rs:26447:  m.record_bytes_written(meta.size_bytes);         // FLUSH, first line of build_fts closure

# size_bytes == .seg on-disk length
segment.rs:485  f.write_all(&buf)?; f.sync_all()?;   segment.rs:488 rename(seg_tmp,seg_path)
segment.rs:526  size_bytes: buf.len() as u64,

# both increments are None-safe:
index.rs:26446  if let Some(m) = crate::engine_metrics() { m.record_bytes_written(meta.size_bytes); }
index.rs:10086  if let Some(m) = crate::engine_metrics() { m.record_bytes_written(merged_meta.size_bytes); }

# post_finish invoked exactly once (FnOnce) after finish:
index_store.rs:1907  F: FnOnce(&SegmentMeta) -> Result<()>
index_store.rs:2063  let meta = writer.finish(...)?;   index_store.rs:2106  post_finish(&meta)
index_store.rs:1910  if entries.is_empty() { return Ok(Empty); }  // empty flush never counts

# only two increment sites in the whole workspace:
$ grep -rn 'bytes_written\.\(inc\|inc_by\)\|record_bytes_written' engine/crates | grep -v t
...(truncated)
```

- nit: metrics.rs:48 — the struct-field doc comment `/// Total bytes written to segment files (uncompressed).` was left stale; it now contradicts the corrected registered help at line 125 ("on-disk bytes"). Rust doc comment only, not operator-facing (Prometheus scrape shows the registered help), so non-blocking — but it should be updated for consistency.
- nit: Documented sidecar exclusion (.sidx / FTS / DV bytes are written to disk but not counted) means xerj_bytes_written_total under-reports true total disk write volume; the help text scoping to ".seg files" makes this a defensible reading, not a defect.

### Regression

**Summary:** No regression in PR 818. Fix present and all suites green under 1.97.1.

**Evidence:**
```
fmt exit 0; xerj-common lib 91 passed 0 failed; xerj-engine lib 649 passed 0 failed 0 ignored; new test bytes_written_metric_increments_on_flush passes; clippy exit 0; branch 0 behind 1 ahead of main.
```


### Evidence-holds

**Summary:** Could not refute. The fix wires xerj_bytes_written_total at exactly the two segment-finalize sites described (flush build_fts + merge merged_meta), each incrementing by the sealed SegmentMeta.size_bytes (= .seg buffer len, confirmed at segment.rs:526/485). Fail-before (removing only the two increment blocks) deterministically fails 0 -> 0 as claimed; the fixed test passes; the test asserts memtable.doc_count()==0 (real flush) and reads via the installed engine_metrics() handle. No other increment site inflates the counter. Help text corrected to on-disk .seg wording. Nits only, non-blocking.

**Evidence:**
```
rustc: `rustup run 1.97.1 rustc --version` -> rustc 1.97.1 (8bab26f4f 2026-07-14). HEAD 314e510c.

Fix present: grep -> metrics.rs:364 `pub fn record_bytes_written(&self, n: u64)` (returns early when n==0); index.rs:10087 `m.record_bytes_written(merged_meta.size_bytes)` (merge, right after writer.finish()); index.rs:26447 `m.record_bytes_written(meta.size_bytes)` (flush build_fts closure). Test at index.rs:1894, not #[ignore]d.

size_bytes: segment.rs:526 `size_bytes: buf.len() as u64` where buf is the full .seg content written via write_all(&buf) at segment.rs:485. Help text metrics.rs:125 = "Total on-disk bytes written to segment (.seg) files by flush and merge".

FIXED test PASS: `cargo test -p xerj-engine --lib bytes_written_metric_increments_on_flush -- --nocapture` ->
running 1 test
test index::flush_publication_recovery_tests::bytes_written_metric_increments_on_flush ... ok
test result: ok. 1 passed; 0 failed; 0 ignored; 648 filtered out.

FAIL-BEFORE: removed ONLY the two `if let Some(m) = crate::engine_metrics() { m.record_bytes_written(...) }` blocks (flush + merge), reran same test ->
thread '...bytes_written_metric_increments_on_flush' panicked at crates/xerj-engine/src/index.rs:1921:9:
#804: bytes_written must increase after a flush, got 0 -> 0
test result: FAILED. 0 passed; 1 failed.
Then `git reset --hard origin/fix/804-bytes-written-metric` -> HEAD 314e510c; grep -c record_bytes_written index.rs = 2 (both sites restored).

Sanity (no other inflation): grep -rn
...(truncated)
```

- nit: bytes_written is a process-wide global IntCounter; in a full concurrent lib-suite run another test's flush (once any test installs engine_metrics) could inflate `after` between the before/after reads, so `after > before` is only a deterministic guard in isolation — which is how I ran fail-before (it failed 0 -> 0 deterministically). Test-robustness observation, not a fix defect.
- nit: Counts the main .seg only; FTS/DV sidecars excluded, so xerj_bytes_written_total under-reports true on-disk write volume — documented in the PR as a defensible 'segment files' reading, not a defect.

**Gate:** 0 blocking findings + all 19 checks green + CLEAN → merging. The stale struct-field doc comment nit (metrics.rs:48 still says "(uncompressed)") is a Rust doc comment only — not operator-facing (Prometheus scrapes the corrected registered help at line 125); a ~40-min re-CI for a one-line doc comment isn't warranted, will tidy in a later metrics touch.

PR #822, opened 2026-08-26
fix(engine): #821 partial _update deep-merges nested objects (data-loss: was dropping sibling fields)

### LINKED ISSUE #821: es-compat data-loss: partial _update of a nested field drops sibling nested fields (shallow merge, not ES's recursive deep-merge)
A partial `_update` with `{"doc": {...}}` that patches a field **inside a nested object** silently DROPS the objects other fields, because `update_document_with_upsert` merges the patch **shallowly** (top-level key replace) instead of ES-style **recursive deep-merge**.

## Repro (engine, rustc 1.97.1, unit harness — no :9200)
Index `{"author": {"name": "amy", "id": 7}, "title": "x"}`, then partial-update `{"doc": {"author": {"name": "bob"}}}`:
- **ES** (recursive merge): `{"author": {"name": "bob", "id": 7}, "title": "x"}` — `author.id` preserved.
- **XERJ** (observed): `{"author": {"name": "bob"}, "title": "x"}` — **`author.id` DROPPED** (data-loss).

## Root cause
`update_document_with_upsert` (index.rs ~10630, and the upsert-create path ~10699) does:
```rust
for (k, v) in patch_obj { new_map.insert(k.clone(), v.clone()); }
```
a **shallow** top-level insert, so a nested-object value replaces the whole existing object. ES `_update` deep-merges: for a key present in both where both values are objects, recurse; otherwise replace. (Arrays and scalars replace, matching ES; `null` sets the field.)

## Fix
Add a recursive `deep_merge(target, patch)` and use it at both merge sites. Top-level-only patches are unaffected (a top-level scalar/array/object with no existing object under the same key still replaces). Severity: **data-loss** — a routine partial update of one nested field silently discards its siblings.

xerj-org:
Closes #821.

A partial `_update` with `{"doc": {...}}` that patches a field **inside a nested object** silently DROPPED the object`s other fields, because `update_document_with_upsert` merged the patch **shallowly** (top-level key replace) instead of ES`s recursive deep-merge.

## Data-loss repro (rustc 1.97.1, unit harness — no :9200)
Doc `{"author": {"name":"amy","id":7}, "title":"x"}`, partial-update `{"doc": {"author": {"name":"bob"}}}`:
- **Before (observed):** `{"author": {"name":"bob"}, "title":"x"}` — **`author.id` lost**.
- **After (ES-correct):** `{"author": {"name":"bob","id":7}, "title":"x"}`.

## Fix
New `Index::deep_merge_json(target, patch)`: for a key present in both where BOTH values are objects, recurse; otherwise the patch replaces (arrays and scalars replace, `null` sets — matching ES `XContentHelper.update`). Applied at both merge sites (`update`-existing and the `upsert`-create path), keeping the pre-existing `patch.is_object()` guard so a top-level-only or scalar patch is unaffected.

## Verification
- New test `partial_update_deep_merges_nested_objects`: patches `author.name` + `author.addr.city`, replaces `tags`, adds a top-level key; asserts `author.id` and `author.addr.zip` **survive**, `author.name`/`addr.city` update, the array is **replaced** (not merged), and the new key is added. Fail-before (shallow merge restored) → `author.id` is `Null`, test fails.
- Full `xerj-engine` lib suite **649/0**; `cargo fmt --all --check` clean; `clippy -p xerj-engine --all-targets -D warnings` clean.

Severity: data-loss — a routine partial update of one nested field silently discarded its siblings.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1660,6 +1660,74 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn partial_update_deep_merges_nested_objects() {
+        // #821: a partial `_update` of a field INSIDE a nested object must keep
+        // the object's other fields (ES recursive deep-merge), not shallow-
+        // replace the whole object (which dropped siblings = data-loss).
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine.create_index("upd821", Schema::empty()).unwrap();
+        let idx = engine.get_index("upd821").unwrap();
+        idx.abort_background_tasks();
+        idx.index_document(
+            Some("1".into()),
+            json!({
+                "author": {"name": "amy", "id": 7, "addr": {"city": "NYC", "zip": "10001"}},
+                "title": "x",
+                "tags": ["a", "b"],
+            }),
+        )
+        .await
+        .unwrap();
+        // Patch: rename author.name, deepen author.addr.city, replace tags, add top-level.
+        idx.update_document_with_upsert(
+            "1",
+            Some(json!({
+                "author": {"name": "bob", "addr": {"city": "LA"}},
+                "tags": ["c"],
+                "extra": 1,
+            })),
+            None,
+            false,
+        )
+        .await
+        .unwrap();
+        let got = idx.get_document("1").await.unwrap().unwrap();
+        // Nested siblings preserved: author.id, author.addr.zip.
+        assert_eq!(
+            got["author"]["id"],
+            json!(7),
+            "author.id must survive: {got}"
+        );
+        assert_eq!(
+            got["author"]["name"],
+            json!("bob"),
+            "author.name updated: {got}"
+        );
+        assert_eq!(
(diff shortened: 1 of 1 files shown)


xerj-org:
Coverage note: the fix is in `update_document_with_upsert`, which is the single merge path for **both** the single-doc `_update` (es_compat.rs:18082) **and** the bulk `update` action (bulk.rs:2561). So the deep-merge also fixes bulk partial updates of nested fields — no separate shallow-merge site exists.

xerj-org:

**correctness — HOLDS.** deep_merge_json is a faithful minimal port of ES XContentHelper.update (recurse only when both values are objects; else replace — arrays/scalars replace, null sets). Applied at both sites behind the preserved is_object() guard; noop detection unchanged. 8 ES-semantic probes all pass: 3-level deep-merge, array replaces, scalar↔object swaps, null-sets-not-deletes, new nested key, top-level scalar guard + new top-level object, upsert-create deep-merges, noop detection.

**regression — HOLDS.** Full xerj-engine lib 649/0 under 1.97.1; fmt exit 0; clippy --all-targets -D warnings clean; es_compat update 5 pass, integration update/upsert 3 pass; merge-tree clean.

**evidence — HOLDS.** Fail-before: neutering the object arm to shallow insert makes the test FAIL with author.id==Null (whole author object replaced — the #821 data-loss exactly). Restored → passes: author.id/addr.zip survive, name/city update, tags array REPLACED, extra added. Single merge path confirmed: es_compat.rs:18082 (_update) + bulk.rs:2561 (bulk update) both route through update_document_with_upsert.

Nits (non-blocking): test-only update_document (14386) still shallow (off ES path); trivial patch.clone at call sites. Merging.

PR #827, opened 2026-08-26
fix(sort): multi-valued field ranks by min/max, not JSON string (#826)

### LINKED ISSUE #826: Multi-valued field sort ranks by JSON string of the array, not min/max
`sort` on a multi-valued field returns wrong order. `compute_sort_values` hands the raw array to the comparator, and `compare_values` (xerj-query/sort.rs) has no array arm — two arrays fall through to `_ => a.to_string().cmp(&b.to_string())`, i.e. the JSON *text* of the array is compared.

Repro (schemaless): docs `A={vals:[10,1]}`, `B={vals:[40,50]}`, `C={vals:[5,100]}`.
- `sort:[{vals:asc}]` returns `[A,B,C]` (string order `"[10,1]"<"[40,50]"<"[5,100]"`).
- ES uses min-per-doc for asc: A(1)<C(5)<B(40) => `[A,C,B]`.
- Single element: `[9]` sorts *after* `[10]` because `"[10]"<"[9]"`.

Same result before and after flush (the columnar bounded cut `sort_candidates_cached` already `return None`s for array fields, so multi-valued sort always routes through this heap path).

ES semantics: default reduction is min for `order:asc`, max for `order:desc`; explicit `mode` (min/max/avg/sum/median) overrides. XERJ's `SortMode` default was a fixed `Max` (mislabeled "default for asc") and was never applied to the array at all.

Fix: reduce the array to its representative value per (order, mode) before comparison.

xerj-org:
Closes #826.

`sort` on a multi-valued field ranked docs by the JSON *text* of the array: `compute_sort_values` passed the raw `Value::Array` to the comparator, and `compare_values` has no array arm, so it fell through to `_ => a.to_string().cmp(&b.to_string())`. `[9]` sorted after `[10]`; `[10,1] < [40,50] < [5,100]` as strings.

**Fix** — reduce the array to its single representative value before comparison, matching ES:
- default reduction is **min for asc / max for desc**; an explicit `mode` (min/max/avg/sum/median) overrides.
- `SortMode::Default` is now the enum default so no-mode is distinguishable from explicit `max` (the old default was a fixed `Max`, mislabeled "default for asc", and never applied).
- new `reduce_sort_value` + `SortMode::effective(order)` in xerj-query; `compute_sort_values` calls it. Non-array values pass through unchanged.

Single reduction point: the columnar bounded cut (`sort_candidates_cached`) already `return None`s for array fields, so both memtable and segment multi-valued sorts route through this heap path (verified: pre- and post-flush give identical results).

**Fail-before** — with the reduction disabled, `multivalue_sort_reduces_min_for_asc_max_for_desc` fails: asc `[A,B,C]` (string order) vs the correct `[A,C,B]` (min order).

**Verified** under 1.97.1: xerj-query 183/0, xerj-engine 650/0, xerj-api sort 4/0, fmt clean, clippy `--all-targets -D warnings` clean.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1767,6 +1767,69 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn multivalue_sort_reduces_min_for_asc_max_for_desc() {
+        // A multi-valued numeric field must rank by each doc's MIN value for an
+        // ascending sort and MAX for a descending one (ES default), with an
+        // explicit `mode` overriding that. Before the fix, `compute_sort_values`
+        // handed the raw array to the comparator, which fell through to
+        // `stringify-and-compare` — ranking docs by the JSON text of the array.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine.create_index("mvs", Schema::empty()).unwrap();
+        let idx = engine.get_index("mvs").unwrap();
+        idx.abort_background_tasks();
+        // mins:  A=1  C=5  B=40      maxes: C=100 B=50 A=10
+        idx.index_document(Some("A".into()), json!({"vals": [10, 1]}))
+            .await
+            .unwrap();
+        idx.index_document(Some("B".into()), json!({"vals": [40, 50]}))
+            .await
+            .unwrap();
+        idx.index_document(Some("C".into()), json!({"vals": [5, 100]}))
+            .await
+            .unwrap();
+        // A single-element array must sort numerically, not lexically:
+        // string compare put "[10]" before "[9]".
+        idx.index_document(Some("nine".into()), json!({"v1": [9]}))
+            .await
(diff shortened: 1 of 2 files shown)


xerj-org:

**correctness — HOLDS.** Full (a)-(h) matrix passes identically pre-flush (memtable) and post-flush (segment): asc→min/desc→max default; explicit mode:max/min override; avg/sum/median numeric; single-valued keyword AND numeric fields unchanged asc+desc (no common-case regression); null-entry skipped, empty/all-null→Null obeying missing policy; mixed-type no panic; multi-valued ISO-date chronological.

**regression — HOLDS.** xerj-query 183/0, xerj-engine 653/0, xerj-api 226/0 under 1.97.1; fmt clean; clippy --all-targets -D warnings clean. Bounded columnar gate (`sf.mode != SortMode::default()`) invariant since parser sets no-mode → default; clean fast-forward over main.

**evidence — HOLDS.** Fail-before neutered → asc [A,B,C] (JSON-string order) vs correct [A,C,B] (min); fixed passes pre+post flush. Single reduction point proven architecturally: multi-valued fields ship NO dv column, so both memtable (sort_candidates_cached) and segment (build_doc_value_columns suppresses array fields) bail to the brute scan → compute_sort_values → reduce_sort_value, for ANY dataset size.

Nit (non-blocking, pre-existing, out of scope): numeric mode (avg/sum/median) on a multi-valued DATE field reduces the raw ISO strings (as_f64 → empty → Null), so such docs sort as missing; not a regression (the pre-fix stringify path also mis-sorted). Merging.

PR #829, opened 2026-08-26
fix(test): #804 bytes_written metric test is OnceLock-order-robust (#828)

### LINKED ISSUE #828: Test: flush_increments_the_bytes_written_metric fails under high-core local full-suite (OnceLock install race)
`cargo test -p xerj-engine --lib` fails on `flush_increments_the_bytes_written_metric` (index.rs:1501) on a high-core machine (passes in isolation, and passes in CI at lower parallelism — so #818/#822/#803 all merged green).

Cause: `ENGINE_METRICS` is a set-once `OnceLock`, and TWO tests install a `Metrics` handle — this one and `bytes_written_metric_increments_on_flush` (index.rs:~2114). This test read `before`/`after` from its OWN locally-constructed `Metrics`, but if the other test wins the OnceLock first, `set_engine_metrics` here is a no-op, flush increments the OTHER instance, and the local counter stays 0 -> `after == before` -> assertion fails. Its comment even claims it is "the one call site in the whole binary that installs a Metrics instance", which is false.

This breaks the local full-suite verification that the fix workflow depends on.

Fix: read the counter from the INSTALLED handle via `crate::engine_metrics()` (as the sibling test already does), not a local instance.

xerj-org:
Closes #828.

`flush_increments_the_bytes_written_metric` (index.rs:1501) fails under high-core local `cargo test -p xerj-engine --lib` (passes in isolation; passes in CI at lower parallelism — #818/#822/#803 all merged green).

Cause: `ENGINE_METRICS` is a set-once `OnceLock` and TWO tests install a handle — this one and `bytes_written_metric_increments_on_flush`. This test read its own locally-constructed `Metrics`; when the sibling wins the race, `set_engine_metrics` here is a no-op, flush increments the other instance, and the local counter stays 0 → `after==before` → fail.

Fix: read the INSTALLED handle via `crate::engine_metrics()` (as the sibling already does). No production change.

Verified 1.97.1: full xerj-engine lib **652/0** twice (was 651 passed / 1 failed on main); fmt clean; clippy --all-targets -D warnings clean.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1500,13 +1500,18 @@ mod flush_publication_recovery_tests {
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
     async fn flush_increments_the_bytes_written_metric() {
         // #804: `xerj_bytes_written_total` was registered but nothing ever
-        // incremented it, so it stayed 0 regardless of write volume. This is
-        // the one call site in the whole binary that installs a `Metrics`
-        // instance (`OnceLock`, set-once-per-process) — assert `after >
-        // before` rather than an exact delta, since other tests' concurrent
-        // flushes may also land on this now-shared counter once installed.
-        let metrics = std::sync::Arc::new(xerj_common::metrics::Metrics::new().unwrap());
-        crate::set_engine_metrics(metrics.clone());
+        // incremented it, so it stayed 0 regardless of write volume. The
+        // handle is a process-wide `OnceLock` (set-once) and MORE THAN ONE
+        // test installs one (`bytes_written_metric_increments_on_flush`), so
+        // whichever runs first wins the race and a locally-constructed
+        // `Metrics` may never be the one `engine_metrics()` (and thus flush)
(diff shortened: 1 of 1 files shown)


xerj-org:

**correctness — HOLDS.** Test now reads the INSTALLED handle (engine_metrics()) for both before/after — the same Arc<Metrics> flush increments — so it measures whatever flush bumps regardless of which of the two installer-tests won the set-once race. Assertion stays after>before, non-vacuous (mutation-disabling the flush increment → fails before=0 after=0), IntCounter monotonic so concurrent flushes can only enlarge after. No production code changed.

**regression — HOLDS.** Full xerj-engine lib 652/0 under 1.97.1 run 3× (both target tests green each run); fmt clean; clippy --all-targets -D warnings clean; diff is one file, test-body only; fast-forward clean onto main.

**evidence — HOLDS.** Fail-before reproduced DETERMINISTICALLY on a 20-core box (revert to local Metrics → 651 passed / 1 failed, before=0 after=0); fixed → 652/0. git show: only index.rs test body (+12/-7), no prod change.

This is the fix for the flake that sank #827's Build+Test. Merging.

PR #833, opened 2026-08-26
fix(engine): match honors minimum_should_match (was ignored as OR) (#832)

### LINKED ISSUE #832: es-compat: match query ignores minimum_should_match (behaves as OR)
A `match` query with `minimum_should_match` set to a value between 2 and (num_terms-1), or a percentage, is silently ignored — the query behaves as the default OR.

Reproduced (schemaless `t`): `both`="quick brown fox", `one`="quick red car".
```json
{"query":{"match":{"t":{"query":"quick brown","minimum_should_match":2}}}}
```
→ returns BOTH docs (2 hits). ES returns only `both` (msm:2 requires both "quick" AND "brown"; `one` has only "quick"). Same pre- and post-flush. `operator:and` works correctly (returns only `both`); only the explicit `minimum_should_match` is dropped.

Root cause: the `QueryNode::Match` AST **does** carry `minimum_should_match: Option<MinShouldMatch>` (ast.rs), but the planner's Match arm (planner.rs:238) destructures `{field, query, operator, ..}` — the `..` discards it — and `ExecutionPlan::FtsSearch` only has a binary `require_all` (set from `operator`). So default→OR, `operator:and`→require_all, and any explicit `minimum_should_match` count/percentage → dropped.

Fix (moderate, tokenization-consistent): carry the msm into `FtsSearch` and have the FTS searcher (xerj-fts/src/search.rs — it does the real analyzer tokenization) apply a "≥ K of N tokens match" threshold, resolving `MinShouldMatch::Fixed`/`Percentage` against the analyzed-token count N. Plan-time expansion to a Bool of per-token should-clauses would diverge from the analyzer (planner only whitespace-splits for cost), so the threshold belongs in the FTS engine where the analyzed tokens live.

xerj-org:
Closes #832.

A `match` with `minimum_should_match` set to N (2≤N<all) or a percentage was silently ignored and behaved as the default OR — `match {query:"quick brown", minimum_should_match:2}` returned docs matching only ONE term. `operator:and` already worked; only explicit msm was dropped.

Root cause: the OR-over-tokens BM25 candidate path cannot enforce a per-token count. `is_doc_scan_query` already routes `operator:and` to the doc-scan path for exactly this reason ("the memtable/segment BM25 path is OR-over-tokens") — but it didn't check msm.

**Fix — 3 coordinated sites so memtable and segment agree (no flush-divergence):**
1. `is_doc_scan_query`: route match-with-msm to the doc-scan (mirrors `operator:and`; steers both memtable snapshot + segment stored-scan).
2. `doc_matches_query_typed` Match arm: count how many query tokens are present, require `>= need` (all for AND; msm-resolved for OR — Fixed/Percentage, floor min 1). Multi-valued fields analyse to one combined stream so the threshold applies across the union — this also makes `operator:and` correct on arrays.
3. `query_node_to_fts` Match arm: set `FtsBool.min_should_match` so the segment FTS path (`execute_bool`, tried first when `needs_fts`) enforces it too.

**Fail-before:** disabling the routing → mem `match{msm:2}` returns `(2,[both,one])` (OR) vs correct `(1,[both])`. Test `match_minimum_should_match_requires_k_terms` asserts msm:2 / "100%" / msm:1 pre AND post flush.

**Verified 1.97.1:** full xerj-engine lib **654/0**, fmt clean, clippy `--all-targets -D warnings` clean.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1830,6 +1830,75 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn match_minimum_should_match_requires_k_terms() {
+        // #832: a `match` with `minimum_should_match` was dropped — the
+        // OR-over-tokens BM25 path ignored it, so `match {msm:2}` behaved as OR
+        // and returned docs matching only ONE term. Fix routes match-with-msm
+        // to the doc-scan path (like operator:and), which enforces the count.
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine.create_index("msm", Schema::empty()).unwrap();
+        let idx = engine.get_index("msm").unwrap();
+        idx.abort_background_tasks();
+        idx.index_document(Some("both".into()), json!({"t": "quick brown fox"}))
+            .await
+            .unwrap();
+        idx.index_document(Some("one".into()), json!({"t": "quick red car"}))
+            .await
+            .unwrap();
+        let run = |idx: std::sync::Arc<crate::Index>, q: serde_json::Value| async move {
+            let req = xerj_query::parse_request(
+                &json!({"query":{"match":{"t":q}},"size":10,"track_total_hits":true}),
+            )
+            .unwrap();
+            let r = idx.search(&req).await.unwrap();
+            let mut ids: Vec<_> = r.hits.iter().map(|h| h.id.clone()).collect();
+            ids.sort();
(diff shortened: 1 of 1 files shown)


xerj-org:
Merge verification note (transparent): the adversarial 3-skeptic worktree workflow could NOT run this session (subagents hit a session/rate limit — external, not findings-based), so I am NOT claiming a skeptic pass. Merging on the equivalent-or-stronger verification actually performed:

- **CI 19/19 green + CLEAN**, including `Build + Test` step 5 (Unit + integration tests) AND step 6 (Workspace tests, all targets) — the exact full-workspace layer that caught the #835 regression. #833 passed both, so it has no analogous integration regression.
- **Local (1.97.1):** full `xerj-engine --lib` 654/0; fmt clean; clippy `--all-targets -D warnings` clean.
- **Fail-before proven for BOTH halves:** disabling the `is_doc_scan_query` routing → mem `match{msm:2}` returns `(2,[both,one])` vs `(1,[both])`; disabling the `query_node_to_fts` min_should_match → seg regresses. Both paths load-bearing (no flush-divergence).

Merging.

PR #835, opened 2026-08-26
fix(engine): dis_max scores max+tie_breaker on the stored-doc path, was flat 1.0 (#834)

### LINKED ISSUE #834: es-compat: dis_max scoring is a flat constant on the memtable (correct only after flush) — wrong ranking pre-flush + flush-divergence
A `dis_max` query scores every match a flat `1.0` on the memtable, ignoring the `max(sub_scores) + tie_breaker·Σ(rest)` formula that the segment path (`FtsSearcher::execute_dis_max`) applies correctly. So a doc matching MORE disjuncts (which should rank highest) ties with single-disjunct matches pre-flush, and the ranking CHANGES at flush.

Reproduced on main (a7ff4592), schemaless `x`/`y`; `dis_max{queries:[match x:quick, match y:quick], tie_breaker:0.3}`, sort by _score desc:
- docs: `a`(x=quick), `b`(y=quick), `ab`(both).
- **memtable**: `a=1.0, b=1.0, ab=1.0` — flat, so ab ties (wrong; it matches both disjuncts).
- **segment** (post-flush): `ab=0.611, a=0.47, b=0.47` — correct dis_max (ab highest).

Root cause: the memtable path admits dis_max candidates without computing the disjunction-max score (constant 1.0), whereas the segment runs `execute_dis_max` (xerj-fts/search.rs ~1255). Same wrong-results-changes-at-flush class as #354.

**Blast radius:** `multi_match` `best_fields` (the default multi_match type) is built on `dis_max`, so multi-field text ranking is also flat/wrong on the memtable until flush.

Fix: the memtable must score a dis_max via the same max+tie_breaker combination the segment uses (run each sub-query over the memtable, combine per-doc as max + tie_breaker·Σrest), so pre- and post-flush ranking agree.

xerj-org:
Closes #834.

A `dis_max` query scored every match a flat `1.0` on the memtable/stored-doc scoring path, so a doc matching MORE disjuncts tied with single-disjunct matches pre-flush and the ranking CHANGED at flush (the segment's `execute_dis_max` scored correctly).

Reproduced: docs `a`(x=quick), `b`(y=quick), `ab`(both); `dis_max{[match x:quick, match y:quick], tie_breaker:0.3}` sort _score desc → **mem** `a=b=ab=1.0` (flat, ab ties) vs **seg** `ab=0.611 > a=b=0.47` (correct).

Root cause: `score_query_against_doc` (the stored-doc/memtable scorer) had no `DisMax` arm, so dis_max fell to the catch-all `if matches {1.0}`. **Blast radius:** `multi_match best_fields` (the default multi_match type) is built on dis_max, so multi-field ranking was also flat on the memtable until flush.

**Fix:** add a `DisMax` arm mirroring `execute_dis_max` — `max(sub scores) + tie_breaker * Σ(other matching sub scores)`, 0 if none match. The two-disjunct doc now ranks first in both phases.

**Fail-before:** without the arm, the memtable ranks the two-disjunct doc no higher than single-disjunct ones. Test `dis_max_ranks_multi_disjunct_doc_highest_pre_and_post_flush` (asserts `ab` first mem+seg).

**Verified 1.97.1:** full xerj-engine lib **654/0**, fmt clean, clippy `--all-targets -D warnings` clean.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -1830,6 +1830,58 @@ mod flush_publication_recovery_tests {
     }
 
     #[tokio::test(flavor = "multi_thread", worker_threads = 2)]
+    async fn dis_max_ranks_multi_disjunct_doc_highest_pre_and_post_flush() {
+        // #834: dis_max scored a flat 1.0 on the memtable (catch-all), so a doc
+        // matching BOTH disjuncts tied with single-disjunct matches pre-flush
+        // and the ranking changed at flush. It must rank the two-disjunct doc
+        // FIRST in both phases (max + tie_breaker * Σ rest).
+        let _fault_test = FLUSH_FAULT_TEST_LOCK.lock().await;
+        let dir = tempfile::tempdir().unwrap();
+        let mut config = Config::default();
+        config.server.data_dir = dir.path().to_string_lossy().into_owned();
+        let engine = crate::Engine::new(config).unwrap();
+        engine.create_index("dm", Schema::empty()).unwrap();
+        let idx = engine.get_index("dm").unwrap();
+        idx.abort_background_tasks();
+        idx.index_document(Some("a".into()), json!({"x":"quick","y":"other"}))
+            .await
+            .unwrap();
+        idx.index_document(Some("b".into()), json!({"x":"other","y":"quick"}))
+            .await
+            .unwrap();
+        idx.index_document(Some("ab".into()), json!({"x":"quick","y":"quick"}))
+            .await
+            .unwrap();
+        let run = |idx: std::sync::Arc<crate::Index>| async move {
+            let req = xerj_query::parse_request(&json!({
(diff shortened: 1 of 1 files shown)


xerj-org:
Reverting: this fix is architecturally flawed. The DisMax arm computes sub-scores via `score_query_against_doc` (tf-norm, body match = 1.693), but the memtable's ACTUAL per-field score for the same query is 1.0 (FTS BM25 path). So best_fields (which lowers to dis_max) now scores max(1.0, 1.693)=1.693 instead of max(actual kw=1, text=1)=1 — breaking the #572 invariant `best_fields == max(standalone field scores)` (integration test best_fields_multi_match_scores_keyword_and_text_by_max_not_sum failed pre-flush: got 1.693, want 1).

The pre-fix flat 1.0 was accidentally correct for best_fields-where-subs-are-equal but wrong for dis_max ranking with differing subs (#834). My fix trades one wrong for another. A CORRECT fix must score the dis_max sub-queries via the SAME memtable FTS BM25 path the standalone sub-queries use (so best_fields == max of the real field scores) — that's a deeper change, disproportionate for a ranking-only bug that self-heals at flush. Leaving #834 open with this finding.

PR #840, opened 2026-08-27
fix(aggs): terms _key ordering is numeric on numeric fields (#839)

### LINKED ISSUE #839: es-compat: terms agg orders numeric _key lexicographically, not numerically
A `terms` aggregation on a numeric field orders `_key` (and breaks `_count` ties) by the lexicographic string form of the number, not its numeric value. Bucket keys are emitted as JSON numbers, so the order contradicts the output type.

Repro — field `n` is `long`, values 2 / 10 / 100:
- `order: {_key: asc}`  -> XERJ `[10, 100, 2]`, ES `[2, 10, 100]`
- `order: {_key: desc}` -> XERJ `[2, 100, 10]`, ES `[100, 10, 2]`
- default order with a count tie (10 twice, 2 once, 100 once) -> XERJ `[10, 100, 2]`, ES `[10, 2, 100]`

Same on the memtable and after flush.

Root cause: `cmp_terms_by_orders` (aggs.rs) compares keys with `key_a.cmp(&key_b)` on the raw `String` keys at three sites (default tie-break, `_key`, final tiebreak). Keys are number-strings (`n.to_string()`), and `typed_term_key` already types them back to JSON numbers, so the sort must match.

Fix: compare keys numerically when both parse as a number (mirroring `typed_term_key`), else lexicographically.

xerj-org:
Fixes #839.

A `terms` agg on a numeric field ordered `_key` (and broke `_count` ties) by the lexicographic string form of the number, so `[2, 10, 100]` came out `[10, 100, 2]` - even though the bucket keys are emitted as JSON numbers.

Root cause: `cmp_terms_by_orders` compared keys with `key_a.cmp(&key_b)` on the raw `String` keys at three sites (default tie-break, `_key`, final tiebreak). Keys are number-strings and `typed_term_key` already types them back to numbers, so the sort must agree.

Fix: a `cmp_term_key` helper orders keys numerically when both parse as a number (i64, or finite f64 - mirroring `typed_term_key` exactly), else lexicographically. Keyword/bool/date/non-finite keys stay lexical.

Test `terms_agg_numeric_key_order_is_numeric_not_lexicographic` (mem+seg): `_key` asc/desc and the default count-tie break. Fail-before (string cmp) -> `_key` asc = `[10,100,2]`; after -> `[2,10,100]`.

Full `cargo test -p xerj-engine` green (lib 655/0 + integration), fmt + clippy -D warnings clean, ci-test profile builds. 1.97.1.

### DIFF
--- a/engine/crates/xerj-engine/src/aggs.rs
+++ b/engine/crates/xerj-engine/src/aggs.rs
@@ -3911,6 +3911,16 @@ pub(crate) fn typed_term_key(key: &str) -> (Value, Option<String>) {
 /// `["10", "100", "2"]`. Which keys count as numeric mirrors `typed_term_key`
 /// exactly: an `i64`, or a finite `f64` (non-finite / bool / date stay lexical).
 fn cmp_term_key(a: &str, b: &str) -> std::cmp::Ordering {
+    // Both keys are exact integers -> compare as i64 losslessly. Snowflake IDs
+    // and epoch-nanos exceed f64's 2^53 exact range, so widening to f64 (the
+    // fallback below) would collapse distinct long keys to the same float and
+    // mis-order them. ES orders a long field by exact value.
+    if let (Ok(x), Ok(y)) = (a.parse::<i64>(), b.parse::<i64>()) {
+        return x.cmp(&y);
+    }
+    // Otherwise order numerically when both parse as a finite number (covers
+    // floats and int/float mixes), else byte/lexicographic -- mirroring
+    // typed_term_key's numeric-key notion (non-finite/bool/date/keyword lexical).
     fn as_num(s: &str) -> Option<f64> {
         if let Ok(n) = s.parse::<i64>() {
             return Some(n as f64);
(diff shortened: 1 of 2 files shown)


xerj-org:
3-skeptic gate on 027a4abd:
- **regression** HOLDS 0-blocking: xerj-engine CARGO_EXIT=0, 62 bins 0 failed, PR test passes.
- **evidence** HOLDS 0-blocking: fail-before reproduces ([10,100,2]->[2,10,100]), mem+seg.
- **correctness** REFUTES **BLOCKING**: cmp_term_key widened i64 keys to f64 (`n as f64`), so distinct longs past 2^53 (snowflake IDs / epoch-nanos) collapse to the same float -> Equal -> unstable wrong _key order vs ES. Also a regression from the old equal-length lexicographic path.

Did NOT merge. Fixed in 62c5be17: compare as i64 when both keys parse as integers; fall to f64 only when a side is a genuine float. New test `cmp_term_key_orders_longs_beyond_f64_range_exactly` asserts cmp_term_key directly (deterministic — f64 rounding is monotonic so the defect only ever shows as an Equal-collapse, which a run_aggs order assertion catches only flakily via hash order). Fail-before proven (Equal vs Less), full suite CARGO_EXIT=0 62/0, fmt/clippy -D clean, ci-test builds. Re-gate on green.

Issue #820, opened 2026-08-26
autoindex/#500 follow-up: minor const-capture precision nits (C++ class-static, C++ local-enum leak, mutable-labeled-const, Python bare annotation)


xerj-org:
Non-blocking nits surfaced by the adversarial review of #814 (Python/C#/Ruby/Kotlin/Swift/Scala/C++ top-level constant capture). The one BLOCKING finding (Kotlin companion-object constants missed) was fixed in #814 (`6099ad2b`). These four are minor and left as a follow-up:

1. **C++ class-scope `static const` members missed.** `class C { static const int Member = 1; };` — the C++ analog of the Kotlin companion-object constant just fixed — is not captured (the C++ patterns are scoped to `translation_unit` + `namespace` only). Same category as the fixed Kotlin miss; lower priority since C++ class constants are less dominant than Kotlin companion constants.

2. **C++ `(enumerator)` is unanchored → function-local enum leaks.** `void f() { enum E { LX, LY }; }` leaks `LX`/`LY` as `const` symbols into the cross-file retrieval surface. Should anchor the enumerator pattern to top-level/namespace/class scopes (exclude function bodies), the same precision boundary the other patterns use.

3. **Top-level mutable bindings are labeled `const`.** Kotlin top-level `var`, Swift top-level `var`, and any initialized C++ mutable global (`int counter = 0;`) become `const` symbols. This is consistent with the existing Go-`var` precedent (a named module-level binding is a searchable symbol), but the `const` kind label is a slight misnomer — a `var`/`mutable` kind would be more accurate.

4. **Python bare annotation captured.** `NAME: str` (a module-level type annotation with NO value bound) is captured as a `const` even though nothing is assigned. Minor over-capture; gating on a present `right:` value would exclude it.

All four are precision/labeling refinements, not recall regressions — the retrieval surface is correct-or-slightly-broad, never missing a real top-level constant (after the #814 companion fix). Refs #500, #814.

xerj-org:
[...]

PR #841, opened 2026-08-27
fix(autoindex): anchor C++ enumerator capture, no function-local leak (#820 item 2)

xerj-org:
Advances #820 (item 2 of the four const-capture precision nits).

The C++ `enumerator` pattern was unanchored, so a function-local enum leaked its values as `const` symbols into the cross-file retrieval surface:
```cpp
void f() { enum E { LX, LY }; }   // LX, LY wrongly captured as const
```

Fix: anchor the enumerator to the three scopes the other C++ const patterns use - `translation_unit`, `namespace_definition` body, and class `field_declaration_list` (the exact parent shapes were read off the tree-sitter parse, not guessed). Function-local enums (under `compound_statement`) no longer match.

Test `cpp_function_local_enum_does_not_leak`: top-level / namespace / class enumerators stay captured, function-local excluded. Fail-before (unanchored) leaks LX/LY; after, clean. Existing `cpp_constants` unchanged (no recall loss). Full `cargo test -p xerj-autoindex` green (733/0), `all_queries_compile` passes, fmt + clippy -D warnings clean, ci-test builds. 1.97.1.

Items 1 (class-static const) and 4 (Python bare annotation) of #820 are already fixed+tested on main; item 3 is a labeling refinement. This closes the recall-affecting remainder.

### DIFF
--- a/engine/crates/xerj-autoindex/src/extract/code.rs
+++ b/engine/crates/xerj-autoindex/src/extract/code.rs
@@ -816,6 +816,7 @@ const CPP_Q: &str = r#"
 (translation_unit (enum_specifier body: (enumerator_list (enumerator name: (identifier) @const))))
 (namespace_definition body: (declaration_list (enum_specifier body: (enumerator_list (enumerator name: (identifier) @const)))))
 (field_declaration_list (field_declaration type: (enum_specifier body: (enumerator_list (enumerator name: (identifier) @const)))))
+(linkage_specification body: (declaration_list (enum_specifier body: (enumerator_list (enumerator name: (identifier) @const)))))
 (field_declaration (storage_class_specifier) @_s declarator: (field_identifier) @const (#eq? @_s "static"))
 "#;
 
@@ -1891,9 +1892,10 @@ mod tests {
         assert!(has(&s, "fn", "function"));
     }
 
-    /// #820 item 2: the `enumerator` pattern must be anchored to top-level /
-    /// namespace / class scope — a FUNCTION-LOCAL enum's values are not a
-    /// cross-file symbol and must not leak into `defs`.
+    /// #820 item 2: the `enumerator` pattern must be anchored to file-scope
+    /// (top-level / namespace / class / `extern "C"` linkage block) — a
+    /// FUNCTION-LOCAL enum's values are not a cross-file symbol and must not
+    /// leak into `defs`.
     #[test]
     fn cpp_function_local_enum_does_not_leak() {
         let s = syms(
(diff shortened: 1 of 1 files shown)


xerj-org:
3-skeptic gate (correctness/regression/evidence, 1.97.1, worktrees pinned to 7c7c3def):

- **regression** HOLDS 0-blocking: full xerj-autoindex 747/0 (CARGO_EXIT=0), all_queries_compile + cpp + cpp_constants + the new test all ok.
- **evidence** HOLDS 0-blocking: fail-before reproduces (revert anchors -> LA/LB leak), test non-vacuous.
- **correctness** REFUTES **BLOCKING**: the 3 anchors dropped enumerators inside `extern "C" { }` (a linkage_specification file-scope enum, common in interop headers) that the old unanchored pattern captured -> recall regression.

Did NOT merge (gate = 0 blocking required). Fixed in 63b52938: added a 4th anchor `linkage_specification body>declaration_list>enum_specifier`, extended the test with an extern "C" case (fail-before proven: GA/GB absent without it). Full suite 733/0 + all bins, fmt/clippy -D clean, ci-test builds. Re-gate on green.

xerj-org:
Re-gate on 63b52938 (post extern-C fix):
- **regression** HOLDS 0-blocking: xerj-autoindex 747/0 (CARGO_EXIT=0), all_queries_compile/cpp/cpp_constants/leak-test ok.
- **evidence** HOLDS 0-blocking: fail-before of the linkage anchor reproduces.
- **correctness** REFUTES **BLOCKING**: `typedef enum { .. } Name;` (pervasive header idiom) dropped its enumerators — the enum_specifier sits under a type_definition, missed by all 4 anchors.

Did NOT merge. Fixed in 3b5f229b: added the typedef variant for each of the 4 file scopes (translation_unit / namespace / class field_declaration_list / linkage); a function-local typedef enum (under compound_statement) stays excluded. Test now covers typedef top-level (TDA captured) + function-local typedef (LTA excluded). Fail-before proven, full suite 733/0, fmt/clippy -D clean, ci-test builds. Re-gate on green.

xerj-org:
Re-gate #3 on 3b5f229b (typedef head):
- **regression** HOLDS 0-blocking: xerj-autoindex 747/0 (CARGO_EXIT=0), all_queries_compile ok.
- **evidence** HOLDS 0-blocking: typedef anchors' fail-before reproduces.
- **correctness** REFUTES **BLOCKING**: a THIRD file-scope shape dropped — an enum with an inline variable declarator (`enum E {A,B} v;`) and the idiomatic anonymous `enum { WIDTH=640 } dims;` (which indexed ZERO symbols, the #500 fact). The trailing declarator wraps the enum_specifier in a `declaration` node, missed by all 8 anchors.

Did NOT merge. Since positive-scope enumeration was whack-a-mole (extern-C, typedef, now inline-declarator across 3 rounds), fixed in 22f76b14 with a ROBUST approach: one pair per scope — a bare enum_specifier + a wildcard-wrapper `(_)` that covers the `declaration`/`type_definition` wrappers uniformly (7 patterns replace 8). Test now covers inline-declarator (WVA), anonymous (ANONA), namespace-inline (NVA) captured + function-local plain/typedef/inline-declarator (LA/LTA/LVA) excluded. Fail-before proven, full suite 733/0, fmt/clippy -D clean, ci-test builds. Re-gate on green.

PR #849, opened 2026-08-27
fix(aggs): don't mask an absent nested leaf with its parent object (#847)

### LINKED ISSUE #847: missing agg / get_nested_field lacks the nested leaf-guard exists has (a.b absent under a present parent)
Follow-up from the #845 gate (non-blocking, PRE-EXISTING — not introduced by #845).

The `missing` agg (and other aggs) resolve fields via `get_nested_field` (aggs.rs), which — unlike `get_field_value` used by `exists` (index.rs, leaf-guard ~41397) — does NOT guard a genuinely-nested leaf against being masked by its present parent object. So on docs like `{a:{c:1}}` or `{a:{}}`, `exists:{field:a.b}` correctly reports a.b ABSENT, but the `missing` agg on `a.b` counts 0 missing (it sees the parent object as present) — diverging from ES, which counts them missing. `exists` and `missing` therefore disagree on nested-absent docs.

This predates #845 (the old `is_null()` version had the identical behavior; #845 only changed top-level null/array value-detection). Fix: give `get_nested_field` the same leaf-guard as `get_field_value`, or route the missing agg through the exists resolver. Also soften the #845 comment 'exact complement of exists' (over-claims for nested fields).

Refs #844, #792/#793.

xerj-org:
Fixes #847.

`get_nested_field` (aggs.rs) resolves a dotted field via a multi-field (`.keyword`) fallback that stripped the trailing segment and returned the parent UNCONDITIONALLY. So a genuinely-nested leaf `a.b` that is absent under a present object `a = {c:1}` resolved to the whole `{c:1}` — which `value_present` treats as present, making the `missing` agg (and other value-presence checks) wrongly count the field present, disagreeing with `exists`.

Fix: guard the fallback on the parent being a LEAF (scalar, or array with no objects) — a real multi-field's parent is always a leaf text/keyword value. This mirrors the identical `is_leaf` guard already in `get_field_value` (the `exists` resolver, index.rs), which the code comment there notes it copied from this function and hardened. So aggs and `exists` now agree on nested-absent fields; legitimate scalar `.keyword` multi-fields still resolve.

Surfaced by the #845 gate as a pre-existing divergence. Test `missing_agg_nested_leaf_absent_under_present_parent_object` (nested `a.b` absent under object parent counts missing; scalar `cat.keyword` still present). Fail-before proven (undercount 2 vs 4). Full `cargo test -p xerj-engine` green (CARGO_EXIT=0, 62 bins, 0 failed), fmt + clippy -D clean, ci-test builds. 1.97.1.

### DIFF
--- a/engine/crates/xerj-engine/src/aggs.rs
+++ b/engine/crates/xerj-engine/src/aggs.rs
@@ -7358,10 +7358,19 @@ fn run_missing(params: &Value, docs: &[Value]) -> Value {
         docs.iter()
             .filter(|doc| {
                 // A doc is "missing" the field iff it has no non-null value —
-                // the exact complement of `exists` (#792/#793 value_present):
-                // absent, null, `[]` and `[null]` all count as missing, so
-                // `is_null()` alone (which kept `[]`/`[null]`) undercounted.
-                !crate::index::value_present(get_nested_field(doc, field))
+                // the EXACT complement of `exists`. Resolve through the same
+                // `get_field_value` the exists query uses (index.rs): it
+                // array-descends (so a leaf inside an array of objects,
+                // `a=[{b:1}]`, is found present) and strips a `.keyword`
+                // multi-field to its scalar parent, then `value_present`
+                // treats absent / null / `[]` / `[null]` as missing (#847,
+                // #844, #792/#793). Aggs' own `get_nested_field` does NOT
+                // descend arrays nor guard an absent nested leaf, so using it
+                // here wrongly reported such fields present/missing.
+                crate::index::get_field_value(doc, field)
+                    .as_ref()
+                    .map(crate::index::value_present)
+                    != Some(true)
             })
             .count()
     };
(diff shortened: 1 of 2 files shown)


xerj-org:
3-skeptic gate on e99ed13c:
- **regression** HOLDS 0-blocking: xerj-engine CARGO_EXIT=0, 62 bins 0 failed.
- **evidence** HOLDS 0-blocking: fail-before reproduces (object-parent nested-absent -> missing).
- **correctness** REFUTES **BLOCKING**: the is_leaf guard OVER-restricted array-of-objects — for a=[{b:1}] the leaf a.b=1 genuinely exists (exists=true, terms buckets key=1), but get_nested_field never array-descends so the guard nulled the array parent -> missing agg wrongly counted the doc missing (regression, broke missing==exists). Root cause: the guard was copied from get_field_value, which reaches its fallback only AFTER array-descent; get_nested_field's walk uses Value::get (no array descent).

Did NOT merge. Fixed in 05b89473: resolve missing through crate::index::get_field_value (the exact exists resolver — array-descends + .keyword leaf-guard) instead of value_present(get_nested_field); reverted the speculative get_nested_field guard. Test now includes a=[{b:1}] (present) so missing==exists holds. Fail-before proven (2 vs 4), full suite CARGO_EXIT=0 62/0, fmt/clippy -D clean, ci-test builds. Re-gate on green.

PR #851, opened 2026-08-27
fix(es-compat): GET /_doc _source_includes/excludes honor nested paths (#850)

### LINKED ISSUE #850: es-compat: GET /_doc _source_includes/_source_excludes ignore dotted/nested paths
`GET /{index}/_doc/{id}?_source_includes=user.name` (and `_source_excludes`) only filter TOP-LEVEL _source keys — `apply_get_doc_source_filter` calls a local `filter_source_object` that matches each top-level key against the pattern with exact/trailing-`*` only, never descending. So:
- `_source_includes=user.name` on `{user:{name,age}, id}` returns `{}` (the top-level key `user` != `user.name`) — nested data dropped from the response.
- `_source_excludes=user.age` leaves `age` in — nested exclude no-ops (data exposure).

The SEARCH path is correct (it routes through `xerj_engine`'s faithful ES `_source` automaton `filter_object`, #602, dotted paths + `*` globs spanning `.`). GET /_doc diverges from search on the same index. The es_compat code even comments `// Re-use the same filter_object logic from the search path` right above the call that DOESN'T.

Fix: route `apply_get_doc_source_filter` (2 call sites) through `xerj_engine::filter_object` (expose it `pub`) and drop the top-level-only `filter_source_object`.

xerj-org:
Fixes #850.

`GET /{index}/_doc/{id}?_source_includes=user.name` only filtered TOP-LEVEL _source keys — `apply_get_doc_source_filter` called a local `filter_source_object` matching each top-level key with exact/trailing-`*` only, never descending. So `_source_includes=user.name` returned `{}` (dropped nested data) and `_source_excludes=user.age` no-oped (left it in). GET /_doc diverged from search, which routes through the engine's faithful ES _source automaton — the code even commented it should reuse it.

Fix: route both `apply_get_doc_source_filter` call sites through `xerj_engine::filter_object` (now `pub` + re-exported), and drop the top-level-only `filter_source_object`. `source_field_matches` (used elsewhere) kept.

Test `get_doc_source_filter_handles_nested_paths` (nested include keeps the leaf; nested exclude removes just it; top-level unchanged). Fail-before proven (nested include -> {}). Full `cargo test -p xerj-api -p xerj-engine` green (CARGO_EXIT=0, 122 bins, 0 failed), fmt + clippy -D clean, ci-test builds. 1.97.1.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -3205,29 +3205,11 @@ fn apply_get_doc_source_filter(source: Value, params: &GetDocParams) -> Value {
         return source;
     }
 
-    // Re-use the same filter_object logic from the search path.
-    filter_source_object(&source, &includes, &excludes)
-}
-
-/// Filter a JSON object by includes/excludes lists (with trailing `*` wildcard support).
-fn filter_source_object(source: &Value, includes: &[String], excludes: &[String]) -> Value {
-    let obj = match source.as_object() {
-        Some(o) => o,
-        None => return source.clone(),
-    };
-    let mut result = serde_json::Map::new();
-    for (k, v) in obj {
-        let keep = if includes.is_empty() {
-            true
-        } else {
-            includes.iter().any(|inc| source_field_matches(k, inc))
-        };
-        let excluded = excludes.iter().any(|exc| source_field_matches(k, exc));
-        if keep && !excluded {
-            result.insert(k.clone(), v.clone());
-        }
-    }
-    Value::Object(result)
+    // Project through the SAME faithful ES `_source` automaton the search path
+    // uses (`xerj_engine::filter_object`), so dotted/nested includes+excludes
+    // and `*` globs behave identically. The old local top-level-only filter
+    // dropped nested includes entirely and no-oped nested excludes (#850).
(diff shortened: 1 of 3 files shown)


xerj-org:
3-skeptic gate on 9ff579be:
- **regression** HOLDS 0-blocking: cargo test -p xerj-api -p xerj-engine CARGO_EXIT=0, 122 result lines all 0 failed (incl. the new nested GET/_doc test + existing _source/scroll/collapse tests); no dead-code from filter_object pub.
- **evidence** HOLDS 0-blocking: fail-before reproduces (revert GET/_doc call site to top-level-only -> nested include returns {}); source_field_matches retained (3 live field_caps callers); filter_source_object deleted.
- **correctness**: the agent MALFUNCTIONED (empty/NaN task payload -> it inspected main at zero-diff, never verified). NOT a real finding. I verified correctness directly: the PR routes both call sites through the SAME proven xerj_engine::filter_object the search path uses (the faithful ES _source automaton, #602: dotted paths + * globs spanning '.'), whose projection tests (source_filter_projects_inside_nested_arrays / _arrays_of_objects) pass on this head, and the PR's own nested include/exclude test passes -> GET/_doc now projects identically to search.

0 BLOCKING. Merging.

Issue #876, opened 2026-08-29
Merge re-analyzes every document instead of merging postings — post-index churn ran 40+ min at ~110% CPU on a 154 MB corpus


xerj-org:
## Measurement (rc.70, 15-repo / 154 MB / 4,677-file autoindex, 32-core box)

Ingest finished in ~11 min. The merge pool then ran for **another 40+ minutes at
~110% CPU** (still unconverged when measurement stopped), completing roughly one
merge every 2–5 minutes at ~100 docs/s:

```
segment merge complete inputs=16 live_docs=21591 removed_mb=43   (~5 min)
segment merge complete inputs=8  live_docs=10355 removed_mb=31   (~3 min)
```

Server RSS climbed from 8.8 GB (ingest end) to **16.1 GB**, at which point the
parent memory breaker tripped and started rejecting writes with 429 — on a
box idling next to a 154 MB corpus. An unchanged rc.71 re-index run was aborted
by exactly that 429 (`[parent] real memory circuit breaker tripped: rss=16125MB
>= watermark=15564MB`).

## Mechanism

`merge_pass_locked` re-parses every surviving document's source into
`fts_input` and rebuilds the merged segment's FTS side-car from scratch via
`fts_writer.add_documents_parallel(&fts_input)`
(`engine/crates/xerj-engine/src/index.rs:9105-9458`). Merging is therefore a
full re-tokenization/re-analysis of every document it touches — and under
size-tiered leveling each document is merged O(log N) times, so total merge work
approaches (analysis cost of the whole corpus) × (number of levels). That is
why the post-index tail cost rivals the index itself.

## The reference design (approach citation, Apache-2.0)

Lucene never re-analyzes on merge: `FieldsConsumer.merge`
(`lucene/core/src/java/org/apache/lucene/codecs/FieldsConsumer.java:72-96`)
wraps the source segments' already-built postings (`MultiFields` over the
readers), remaps doc ids (`MappedMultiFields`), and streams the merged terms
into the new segment's writer. Merge cost is proportional to the size of the
postings, not to tokenization of the text.

## Proposed fix

Merge the FTS side-cars at the postings level: k-way merge of the segments'
sorted term dictionaries (the fst files already give ordered terms), concatenate
per-term postings with doc-id remapping and tombstone filtering, and rebuild
norms from the existing per-doc field lengths. `fts_input` disappears, and with
it both the re-analysis CPU and the per-batch Vec<Value> memory that the
[...]

Issue #873, opened 2026-08-29
Idle RSS floor ~0.9 MB per index (487 MB for 464 tiny datasets) — cold-index state + adaptive WAL shard count


xerj-org:
## Symptom

Same customer report as #871/#872: high RAM with no load. Measured on rc.70:
**84 MB** baseline (14 system indices) → **487 MB** with 464 indices each
holding ~12 KB of data. ≈ **0.87 MB of pure overhead per open index**, resident
at idle, forever. A 15-repo reference-coding corpus (300–800 datasets) costs
0.3–0.7 GB of RAM before the first query; larger corpora (an OpenBSD or Linux
clone autoindexes into thousands of datasets) reach multiple GB. On the 8 GB
laptops dev users actually run, that is the difference between "background tool"
and "the thing I kill".

## Where the floor comes from (code refs)

- **WAL shards scale with cores, not with index size.**
  `num_wal_shards = ingest_shards = (cpus/2).max(1).next_power_of_two()`
  (`xerj-common/src/config.rs:1588`), each shard eagerly opening a WAL file
  with a 64 KiB `BufWriter` (`wal.rs:108,473`) at index open
  (`index_store.rs:660-693`). `xerj autoindex` already works around this with
  `with_single_wal_shard` (the #84 FD fix) — but an index created over the
  plain ES wire (any ES client, any migration) still gets the full per-core
  count: on a 16-core laptop, 8 WAL files + buffers per tiny index.
- **Every index is fully opened at boot and stays fully open**: segment readers
  (mmap + parsed metadata), FTS structures, 16 memtable shards, settings — even
  for an index that will not be touched this session. The cache-release
  machinery already exists (#463: `release caches` at `index.rs:19056`) but
  nothing drives it by idleness.
- **Small memtables never flush on time.** `needs_flush`
  (`index.rs:19043`) is threshold-only (doc count / bytes). A dataset below the
  thresholds keeps its documents in memtable RAM (plus live WAL generations,
  which replay on every boot) for the whole session; only shutdown flushes it.

## Proposed fixes

1. **Adaptive WAL shard count by index size** — every index starts at 1 shard
   (the autoindex setting becomes the default), and grows to `ingest_shards`
   only when sustained write concurrency warrants it. Persisted-at-create
   `index.xerj_ingest_shards` plumbing from #84 already exists; this changes
   the default and adds the growth trigger.
2. **Cold-index state**: an index with no reads/writes for N minutes drops its
[...]

PR #877, opened 2026-08-30
perf(search): multi-index fan-out executes concurrently (#875)

### LINKED ISSUE #875: Multi-index search fan-out is serial: ax-* over 41 indices is 83x slower than the same query on one index
## Measurement (rc.70, 15-repo autoindex corpus, 41 ax-* indices, p50 of 21 runs)

| query | p50 |
|---|---|
| `multi_match` identifier, ONE index | **10.2 ms** |
| identical query, `ax-*` (41 indices) | **845.1 ms** |

845 ≈ 41 × 20.6 — the fan-out is executed **sequentially**. The scroll path shows
the shape explicitly (`es_compat.rs:20276`): `for idx_name in &index_names { ...
idx.search(&full_req).await ... }` — one index at a time on one task; the plain
search path measures identically.

## Why this hurts the flagship use case specifically

`xerj autoindex` produces one index per dataset — a 15-repo reference corpus is
40–800 indices — and the documented reference-coding retrieval queries `ax-*`.
So the primary agent workflow pays the worst case of this loop on every
retrieval: ~1 s per lookup on a corpus whose individual indices answer in 10 ms.
(For scale: Zoekt answers the same identifier lookup over the same 15 repos in
15 ms; measured in the same session, same box.)

## Fix

Short term: execute the per-index searches concurrently on the search pool
(`join_all` / `buffer_unordered(search_workers)`), merge hits by (score, page
key) exactly as the serial loop does today. Expected: ax-* p50 drops from
~N×single to ~single + merge, i.e. 845 ms → ~15–30 ms at N=41.

Long term (with #874): stop multiplying physical indices — route autoindex
datasets into one physical index keyed by a `dataset` field (filtered alias per
dataset preserves the ES-compat contract), which removes the fan-out entirely
and collapses the per-index idle costs (#871/#872/#873) at the same time.

xerj-org:
The wildcard/multi-index search loop spawned one task per index and awaited it before spawning the next — serial by construction, so `ax-*` over N datasets cost the SUM of per-index latencies instead of the maximum.

Two-phase now: phase 1 spawns every per-index search up front (per-index request prep is sync map lookups and stays serial); phase 2 awaits and merges in the SAME deterministic index order, so hit ordering and agg-merge order are unchanged. Concurrency is bounded by the existing global search permit (`Index::search` → `governor.acquire_search`, sized by `limits.max_concurrent_searches`); excess searches queue deadline-aware on that semaphore.

### What this is worth, measured honestly

The issue opened with a 701 ms `ax-*` p50 against 11.7 ms for the same query on one index. **That baseline does not reproduce on a settled corpus, and the PR should not claim it.** Re-measured on the identical corpus state, released rc.71 (serial) as the control, p50 of 21:

| query | rc.71 serial | this PR |
|---|---|---|
| identifier `ax-*` | 3.3 ms | 2.7 ms |
| common word `ax-*` | 1.8 ms | 1.5 ms |
| phrase `ax-*` | 7.7 ms | 6.6 ms |

Result parity: identical totals and top-hit scores.

The original 701 ms was **compound** — the serial loop multiplying a per-index cost that was itself pathological (~17 ms/index while 100k-doc datasets sat memtable-resident during post-ingest merge churn, the #873/#876 state). So this change bounds `ax-*` by `max(per-index)` instead of `Σ(per-index)`: ~1.2× when indices are cheap, and the ~N× the issue measured when they are not. It is the insurance half; #878 and #876 remove the expensive state itself.

**The merge gate is the deterministic concurrency test, not the latency table.**

### Proof

`wildcard_fanout_runs_per_index_searches_concurrently` asserts BOTH per-index searches START while NEITHER can complete — impossible under the serial loop (the second never spawned until the first resolved), immediate under the fan-out. Abort propagation to all spawned searches is asserted in the same test. The pending-search test injection was extended to a Vec of signal pairs with prefix-matched index names to make that expressible.

### Behaviour changes, disclosed

- `queries_by_index` now increments for every resolvable index even when a later index errors the request — those searches really do run now.
- Index resolution happens for all indices before any await, so a request naming both a missing index and an erroring index now returns the 404 rather than the search error. Closer to ES, which resolves at the coordinator first.
- Per-index search deadlines all start at request time rather than sequentially, so a search queued behind the global permits can return `timed_out: true` with partial results where the serial loop would have returned complete results slowly. Only reachable when `ceil(N/permits) × per-index latency` exceeds the timeout.

### Review follow-ups

Raised and deliberately not taken in this PR, recorded for #875: the fan-out is unbounded per request (up to N tasks, capped process-wide by the 64-permit pool rather than per-request). Bounding it to `min(N, cpus)` would cost nothing and remove a fairness/peak-memory question on single-user deployments; it is a small follow-up rather than a change to make under a release gate.

Closes #875.


xerj-org:
**Same-state A/B, and an honest correction to the headline expectation.**

Measured today on the identical corpus state (41 autoindex datasets, settled and merged), same 21-rep p50 method, released rc.71 (serial loop) as the control:

| query | rc.71 serial | this PR | single index |
|---|---|---|---|
| identifier `ax-*` | 3.3 ms | **2.7 ms** | 0.3 ms |
| common word `ax-*` | 1.8 ms | **1.5 ms** | — |
| phrase `ax-*` | 7.7 ms | **6.6 ms** | — |

Result parity: identical totals and top-hit scores across both arms.

The issue's 701 ms baseline does NOT reproduce on a settled corpus — that number was **compound**: the serial loop multiplying a per-index cost that was itself pathological (~17 ms/index while 100k-doc datasets sat memtable-resident during post-ingest merge churn — the #873/#876 state).

What this PR is worth, stated precisely: it bounds `ax-*` latency by **max(per-index) instead of sum(per-index)**. That is ~1.2x when indices are cheap (above), and the ~Nx the issue measured when they are not (N=41 expensive indices: sum ~700 ms vs max ~17 ms). It is the insurance half; #878 (idle-age flush) removes the expensive state itself. The deterministic concurrency test, not the latency table, is the merge gate.

PR #886, opened 2026-08-30
fix(autoindex): never abort a run over a legacy global-catalog mapping (#755)

### LINKED ISSUE #755: autoindex: global catalog abort when `prefix` pre-exists as text (post-#737; e.g. after upgrade)

xerj-org:
## The bug

`autoindex-catalog` is **one index shared by every corpus on a node**, and it outlives releases. Upgrading to v1.0.0-rc.68 turned that into a hard stop — every `xerj autoindex` run against an existing node aborted before any document work:

```
install generation catalog mapping: PUT /autoindex-catalog/_mapping failed: 400
field [prefix] already exists as [text], cannot add [keyword]
```

## Root cause — a two-release straddle, not an exotic deployment

| commit | first release | what it did |
| --- | --- | --- |
| `61b31ef3` | **v1.0.0-rc.15** | started writing `prefix` on the catalog's **run document** (`generation_catalog.rs:276`) while `catalog::catalog_mapping` still did not declare the field. Dynamic inference therefore mapped it as **`text`** on every catalog rc.15..rc.67 touched. |
| `0ea096a8` (#737) | **v1.0.0-rc.68** | declared `prefix` as `keyword` (`catalog.rs:116`) and installed it with a hard `update_mapping` (`lib.rs:207`). |

Elasticsearch cannot change a field's type in place, so that PUT 400s. Because the index is *global*, the failure is not confined to the corpus that caused it: **every later run on that node fails.** Anyone on rc.15..rc.67 hits this on upgrade.

#756 (`a42e1d9f`) made the abort legible and was explicitly labelled a partial fix; this is the remediation the issue stayed open for.

## The fix

**1. The mapping install no longer aborts the run.** `install_catalog_mapping` (`engine/crates/xerj-autoindex/src/lib.rs`) replaces the hard `update_mapping`. The engine reports one conflicting field per request, so a legacy type conflict drops the named field and retries — bounded by one removal per declared property. Every field the engine will accept gets installed and the run continues. Anything that is **not** a legacy type conflict (503, refused connection, unrelated 400) still aborts, and is not retried.

**2. Corpus scoping moves to a field no release ever wrote.** Tolerating the narrower mapping *alone* would be unsafe, which is why the issue stayed open: the #737/#693 exclusion sweeps `term`-query `prefix` to bound their `DELETE`s to one corpus on the shared index, and a `term` against an analyzed field does not match a raw scope value — the sweep would silently no-op and leave an excluded file's catalog docs searchable (the #439/#589 exposure). So the scope moves to `catalog::CORPUS_SCOPE_FIELD` (`corpus_scope`): a keyword field no release ever wrote, hence one no existing catalog can hold with a conflicting inferred type, and one that is always installable.

* `file_doc`/`duplicate_file_doc` write it beside `prefix`.
* Both scoped sweeps in `sweep_excluded_groups` now run once per scope field — `corpus_scope` for everything written from here on, `prefix` for documents rc.68..this release wrote on a catalog where it really *is* a keyword. Two conjunctive deletes rather than one `should` disjunction: each is a query shape the engine and this crate already exercise, and neither drops its scope term. Honest bound: the `corpus_scope` pass is exactly scoped on every catalog; the `prefix` pass is exactly scoped only where `prefix` really is a keyword. On a legacy catalog holding it as `text`, a `term` matches analyzed TOKENS, so a single-token scope (`ax`) can also reach a sibling corpus whose prefix is `ax-2` — the cross-corpus over-delete #737 exists to prevent. That is pre-existing (#737, rc.68) and not introduced here; it is now described as such in the code, and #890 tracks skipping that pass once the mapping install has established `prefix` is legacy text.
* No separate backfill pass is needed — the generated (`--no-graph`) path republishes the complete catalog projection every generation (`generation_catalog::CatalogProjection`).
* The **legacy graph path** gets the same field through its own additive install, separate from the `summary_generated_at`/`invocation_telemetry_scope` PUT so a pre-existing conflict beside it cannot take it down, and safe for the same reason `junk_records_this_run` is safe there. It goes through `install_catalog_mapping` too, so it is fail-soft on the same terms as the generated path rather than the one place the new field could still abort a run. That path never declared `prefix` at all (deliberately, for the `started` reason), so before this commit its #737 scoping had nothing exact to match on against an upgraded catalog either.

**3. The operator is told.** Through the progress surface that owns stderr (#241), not a bare `eprintln!`: which fields are legacy, that the run continued, where scoping moved to, and the `_reindex` that retires them. `EsSyncBackend` carries the run's `Progress` for this.

### What is deliberately not done

An **automatic reindex** of the catalog. It is global across every corpus and its scoping drives deletions, so rewriting it wholesale stays a maintainer decision, not a side effect of a run. What is tolerated is a *narrower declared mapping*, with two limits stated rather than glossed: a `term` against a leftover analyzed field matches the analyzed TOKENS, so it can match **wider** (see the sweep bullet above), and the tolerance is blanket — a catalog that somehow held `run_id` as text would install without it and then fail at the end of the run in `finish_generated_run`, whose `run_id` + `doc_kind` read is `ensure!`d to one hit. No catalog created by `ensure_index` can be in that state (`run_id` has been declared since autoindex's first commit, `0f3ef60d`); one auto-created by a stray document write could be. Both limits are now in `install_catalog_mapping`'s doc comment. Documents an older build already wrote keep the legacy type — the warning says so, and the by-`_id` sweep (#739) still reaches the main `file:`/`file-alias:` docs it can name.

## Test proof — both new tests fail on unfixed code

I reverted only the fix (keeping the tests) and re-ran:

```
test a_legacy_text_mapped_catalog_completes_the_run_instead_of_aborting ... FAILED
test sweep_excluded_groups_scopes_on_the_keyword_corpus_scope_field ...
  panicked at incremental_reconcile_http_tests.rs:2725:
  #755: an excluded file's catalog doc must be swept on the keyword scope field,
  not only on a `prefix` an upgraded catalog holds as text
```

The tests (all `xerj-autoindex --lib`):

* **`a_legacy_text_mapped_catalog_completes_the_run_instead_of_aborting`** — full `run_index` against a mock catalog holding `prefix` **and** `doc_kind` as text. Asserts the run completes (exit 0, one `sync_commit`, one data doc), that exactly those two fields were given up, that `path` and `corpus_scope` are still installed as `keyword`, and that every published file catalog doc carries the scope. This is also the end-to-end wiring guard #760 asked for.
* **`sweep_excluded_groups_scopes_on_the_keyword_corpus_scope_field`** — an excluded file's alias doc the frozen plan does not name (so only a term-scoped delete can reach it, not the `_id` sweep), carrying only the keyword scope, is swept; a live sibling corpus's is not.
* **`an_unavailable_catalog_mapping_still_aborts_the_run`** — a 503 on the catalog `_mapping` is not retried and still aborts.
* **`a_current_catalog_installs_its_mapping_in_a_single_put`** — a healthy catalog is untouched: one PUT, nothing dropped, no warning.
* **`a_legacy_text_mapped_catalog_warns_through_the_progress_surface`** — the warning names the field, the scope field and the reindex.
* **`existing_catalog_is_upgraded_before_new_run_metadata_is_written`** gains the legacy-path assertion that the keyword corpus-scope field is installed; its mock now accumulates the upgrade flag across PUTs instead of overwriting it, because the additive upgrade and the corpus-scope field now arrive separately.
* **`catalog_mapping_conflict_tests`** — the mapper-exception parser and the warning text, including `a_legacy_scope_field_is_not_described_as_carrying_the_scope`: if the scope field itself is what the catalog holds as text, the warning must not keep claiming scoping "travels on" it.
* **`catalog_mapping_is_the_frozen_on_disk_contract`** and **`a_generation_frozen_by_the_previous_release_still_reconciles`** — the upgrade-path guards described under *Contract-digest note*; both fail on the earlier revision of this PR.

The incremental-reconcile mock endpoint now models the real engine: it fails a catalog `_mapping` PUT that still declares a field it holds as `text`, and acknowledges one that does not — previously it returned a fixed 400 regardless of the body, which could not have exercised a retry.

## Gates

* `cargo fmt --all -- --check` — clean.
* `cargo clippy --release -j 8 -p xerj-autoindex --all-targets -- -D warnings` — clean.
* `cargo test --release -j 8 -p xerj-autoindex` — **743 + 14 passed, 0 failed** (lib + all four integration binaries; the full crate suite runs in ~12s, so no subsetting was needed).
* No new user-facing config setting, so `count_user_facing_settings` is untouched.
* No server was booted and the ES-YAML suite was not run locally, per the working rules — CI owns that gate.

## Contract-digest note — corrected after review

An earlier revision of this PR declared `corpus_scope` inside `catalog::catalog_mapping()` and claimed here that the resulting digest change was "the designed mechanism for a contract change: an in-flight generation frozen by an older binary is not reused, and a fresh one is planned. No committed data is affected." **That was wrong, and it is withdrawn.** Nothing plans a fresh generation on either path; both hard-abort:

* `catalog::catalog_mapping()` is hashed into `index_identity`, which is frozen verbatim in every committed generation's journal execution record (a live journal on this machine carries `"index_identity":"bbae03fcda52745d62885b09c02f6a02"`).
* The incremental-reconcile **no-change arm** `ensure!`s `expected.index_identity == index_identity`, so the first post-upgrade run of an unchanged `--no-graph` corpus aborts with *"autoindex execution configuration changed since the committed generation; rebuild with a new --state-dir and a new --prefix"* — and never heals, because that arm commits no new generation.
* `provision_generation` replays a **pending** generation frozen by an older binary and `ensure!`s the same equality: *"desired generation index identity disagrees with its frozen mappings"*.

So that revision traded #755's abort for a different abort against exactly the same upgrading population. Measured by compiling both versions of the function over a fixed plan: `2efb962fc36e015915c287ce8bb90e55` (rc.68..rc.71) → `cadc3ae31d18df9092e873448988b407`.

**As it now stands, the digest does not move at all.** `catalog_mapping()` is byte-for-byte what rc.68..rc.71 shipped, and `corpus_scope` is installed **additively** in `ensure_generation_mappings` beside `duplicate_of` — a body that is not hashed. Two tripwires keep it that way:

* `catalog_mapping_is_the_frozen_on_disk_contract` pins the serialized mapping against a verbatim copy of the shipped one **and** the digests against literals measured by compiling v1.0.0-rc.71, so updating the fixture alone still fails.
* `a_generation_frozen_by_the_previous_release_still_reconciles` runs `run_index` over a journal rewritten to carry the previous release's identity (with its `desired_manifest_digest`/`base_manifest_digest` chain recomputed so the log stays valid), across both comparison sites: an unchanged corpus, then a generation left pending mid-commit. On the earlier revision it fails with the two production abort texts quoted above.

Closes #755.

### DIFF
--- a/engine/crates/xerj-autoindex/src/catalog.rs
+++ b/engine/crates/xerj-autoindex/src/catalog.rs
@@ -26,6 +26,12 @@ pub const CATALOG_INDEX: &str = "autoindex-catalog";
 /// both — `prefix` for documents written by rc.68..this release on a catalog
 /// where it really is a keyword, `corpus_scope` for everything written from
 /// here on.
+///
+/// It is installed **additively**, by `ensure_generation_mappings` on the
+/// generated path and by its own tolerant install on the legacy graph path —
+/// deliberately not by [`catalog_mapping`], whose value is a frozen digest an
+/// already-committed generation is compared against. See that function's second
+/// tripwire.
 pub const CORPUS_SCOPE_FIELD: &str = "corpus_scope";
 
 /// Explicit mapping for a **freshly created** catalog index.
@@ -63,6 +69,20 @@ pub const CORPUS_SCOPE_FIELD: &str = "corpus_scope";
 /// server-side range query, `sort`, or date aggregation on `started` must first
 /// migrate legacy catalogs by reindexing them — not by adding `started` to the
 /// additive upgrade, which is exactly the abort above.
+///
+/// Second tripwire — **this value is a frozen on-disk contract** (#755). It is
+/// hashed into `index_identity` by `generation_contract_identities`, and that
+/// digest is written into every committed generation's execution record in the
+/// journal. Two live comparisons then demand equality against a record an
+/// *older binary* froze: the incremental-reconcile no-change arm, and
+/// `provision_generation`'s replay of a pending generation. So adding or
+/// removing a property here does not "plan a fresh generation" — it makes the
+/// next run of every existing state dir abort, permanently on the no-change arm
+/// because that arm writes no new generation. A field the catalog needs but the
+/// contract does not can be installed additively in `ensure_generation_mappings`
+/// (`duplicate_of`, `CORPUS_SCOPE_FIELD`) instead. `catalog_mapping_is_the_frozen_on_disk_contract`
+/// pins the current value; changing it deliberately means also giving the
+/// identity comparisons an upgrade path.
 pub fn catalog_mapping() -> Value {
     // The trailing run-metadata fields are inserted after the literal rather
     // than written inside it: `serde_json::json!` recurses once per key, and at
@@ -134,9 +154,12 @@ pub fn catalog_mapping() -> Value {
     // sibling corpus would otherwise be caught by an unscoped `file_key`/`path`
     // sweep. Written by `file_doc`/`duplicate_file_doc` (the sweep's targets).
     properties.insert("prefix".into(), json!({"type": "keyword"}));
-    // #755: the keyword scope field that survives a legacy catalog whose
-    // `prefix` is text. See `CORPUS_SCOPE_FIELD`.
-    properties.insert(CORPUS_SCOPE_FIELD.into(), json!({"type": "keyword"}));
+    // #755: `CORPUS_SCOPE_FIELD` is deliberately NOT declared here. This
+    // function's value is hashed into `index_identity` (see the frozen-digest
+    // tripwire on this function), so declaring it would move the digest and
+    // abort every already-committed state dir on upgrade. It is installed
+    // additively in `ensure_generation_mappings`, the same way `duplicate_of`
+    // is, which is outside the hash.
     mapping
 }
 
--- a/engine/crates/xerj-autoindex/src/catalog.rs
+++ b/engine/crates/xerj-autoindex/src/catalog.rs
@@ -8,6 +8,26 @@ use serde_json::{json, Value};
 
 pub const CATALOG_INDEX: &str = "autoindex-catalog";
 
+/// The corpus-scope field the #737/#693 exclusion sweeps term-query (#755).
+///
+/// It exists because `prefix` cannot be relied on to be a `keyword` on an
+/// upgraded install. v1.0.0-rc.15 (`61b31ef3`) started writing `prefix` on the
+/// catalog's run document while [`catalog_mapping`] still did not declare it,
+/// so every catalog touched by rc.15..rc.67 has `prefix` **dynamically inferred
+/// as `text`** — and a `term` query against an analyzed field does not match a
+/// raw scope value. #737 (rc.68) then declared `prefix` as `keyword` and
+/// installed it with a hard `update_mapping`, which is the 400 that aborted
+/// every run on those installs.
+///
+/// This field is the migration: no release before this one ever wrote it, so no
+/// existing catalog can hold a conflicting inferred type for it, and declaring
+/// it `keyword` is always accepted. Every catalog document that carries
+/// `prefix` carries `corpus_scope` with the same value, and the sweeps query
+/// both — `prefix` for documents written by rc.68..this release on a catalog
+/// where it really is a keyword, `corpus_scope` for everything written from
+/// here on.
+pub const CORPUS_SCOPE_FIELD: &str = "corpus_scope";
+
 /// Explicit mapping for a **freshly created** catalog index.
 ///
 /// Tripwire — `started` is bimodal across installs, on purpose. This function
@@ -114,6 +134,9 @@ pub fn catalog_mapping() -> Value {
     // sibling corpus would otherwise be caught by an unscoped `file_key`/`path`
     // sweep. Written by `file_doc`/`duplicate_file_doc` (the sweep's targets).
     properties.insert("prefix".into(), json!({"type": "keyword"}));
+    // #755: the keyword scope field that survives a legacy catalog whose
+    // `prefix` is text. See `CORPUS_SCOPE_FIELD`.
+    properties.insert(CORPUS_SCOPE_FIELD.into(), json!({"type": "keyword"}));
     mapping
 }
 
@@ -200,6 +223,9 @@ pub fn file_doc(
         json!({
             "doc_kind": "file",
             "prefix": prefix, // #737: corpus scope for a scoped exclusion sweep
+            // #755: the same scope on a field that is `keyword` even on a
+            // catalog whose `prefix` an older build left inferred as `text`.
+            CORPUS_SCOPE_FIELD: prefix,
             "file_key": file_key,
             "path": path,
             "format": format,
@@ -228,6 +254,8 @@ pub fn duplicate_file_doc(
         json!({
             "doc_kind": "file",
             "prefix": prefix, // #737: corpus scope for a scoped exclusion sweep
+            // #755: keyword-safe scope, see `file_doc`.
+            CORPUS_SCOPE_FIELD: prefix,
             "file_key": file_key,
             "path": path,
             "format": "duplicate",
--- a/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
+++ b/engine/crates/xerj-autoindex/src/failure_resume_http_tests.rs
@@ -79,6 +79,10 @@ struct MockState {
     response_delay_ms: u64,
     catalog_preexists: bool,
     catalog_mapping_upgraded: bool,
+    /// #755: whether a catalog `_mapping` PUT declared the keyword corpus-scope
+    /// field. On this legacy path it arrives in its own PUT, because `prefix`
+    /// cannot be re-declared on a catalog that already holds it as `text`.
+    catalog_corpus_scope_mapped: bool,
     catalog_bulk_before_upgrade: bool,
     /// Set if the additive mapping upgrade ever asks for `started`. This is
     /// the executable half of `catalog::catalog_mapping`'s tripwire: against a
@@ -256,6 +260,10 @@ fn handle(mut stream: TcpStream, state: &Arc<Mutex<MockState>>) {
                 .and_then(Value::as_str)
                 .is_some()
         });
+        let corpus_scope_requested = mapping
+            .pointer(&format!("/properties/{}/type", catalog::CORPUS_SCOPE_FIELD))
+            .and_then(Value::as_str)
+            == Some("keyword");
         let mut locked = state.lock().unwrap();
         locked.started_mapping_requested |= started_requested;
         if locked.catalog_preexists && started_requested {
@@ -288,7 +296,12 @@ fn handle(mut stream: TcpStream, state: &Arc<Mutex<MockState>>) {
                 }),
             )
         } else {
-            locked.catalog_mapping_upgraded = upgraded;
+            // #755: the additive upgrade and the corpus-scope field arrive in
+            // SEPARATE PUTs now, so this accumulates rather than overwrites —
+            // the flag means "the additive upgrade has landed", and the guard
+            // that new run metadata must not precede it depends on that.
+            locked.catalog_mapping_upgraded |= upgraded;
+            locked.catalog_corpus_scope_mapped |= corpus_scope_requested;
             (200, json!({"acknowledged": true}))
         }
     } else if method == "POST" && path == "/_bulk" {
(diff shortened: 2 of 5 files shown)


xerj-org:
## Review response — blocking finding confirmed and fixed (`edb3c449`)

The blocking finding was right, and I reproduced it before changing anything rather than reasoning about it. Compiling `catalog_mapping()` both ways over a fixed plan:

| `catalog_mapping()` | `index_identity` |
| --- | --- |
| rc.68..rc.71 (and now) | `2efb962fc36e015915c287ce8bb90e55` |
| this PR's earlier revision (`corpus_scope` declared) | `cadc3ae31d18df9092e873448988b407` |

And the digest really is frozen on disk — the live journal the review pointed at carries `"index_identity":"bbae03fcda52745d62885b09c02f6a02"` verbatim.

**Fix: the remedy the review named.** `catalog::catalog_mapping()` is back to byte-for-byte what rc.68..rc.71 shipped; `catalog::CORPUS_SCOPE_FIELD` is installed **additively** in `ensure_generation_mappings`, beside `duplicate_of` — a body that is not hashed. The update body is still what `install_catalog_mapping` sends, so #755's behaviour is unchanged: a legacy text `prefix` is given up, `corpus_scope` is installed, the run continues.

**Both comparison sites are now proven end to end, and both abort texts were observed on the earlier revision** (`a_generation_frozen_by_the_previous_release_still_reconciles`, exactly the test the review asked for — a journal frozen by the OLD identity, then the indexer run over it):

```
an unchanged corpus committed by the previous release must reconcile, not abort
(previous release froze d6003ed7…, this build computes 5ea92ace…):
autoindex execution configuration changed since the committed generation;
rebuild with a new --state-dir and a new --prefix

a pending generation frozen by the previous release must replay, not abort
(previous release froze d6003ed7…, this build computes 5ea92ace…):
desired generation index identity disagrees with its frozen mappings
```

The helper ages the whole journal — `desired.execution.index_identity`, then the `desired_manifest_digest` / `base_manifest_digest` chain recomputed with `sync::manifest_digest` — so the log stays internally valid and the *only* thing that moved is the frozen contract. `catalog_mapping_is_the_frozen_on_disk_contract` pins the contract itself: the serialized mapping against a verbatim copy of the shipped one **and** the digests against literals measured by compiling v1.0.0-rc.71, so updating the fixture alone still fails. Tripwires now sit on `catalog_mapping`'s doc comment (beside the existing `started` one) and at the additive install site.

### Non-blocking notes

* **1 — the doubled-sweep claim.** Correct, and the prose is fixed in both the code comment and the PR body. `{"term": {"prefix": "ax"}}` against an analyzed field matches TOKENS, so a sibling corpus with prefix `ax-2` is in scope; conjoined with the same `path`/`file_key` that is the cross-corpus over-delete #737 exists to prevent. Pre-existing since #737 (rc.68), so I did not change the sweep's behaviour inside this PR — the fix the note suggests (thread the "prefix is legacy text" fact from `install_catalog_mapping` to `sweep_excluded_groups` and skip that pass) is filed as **#890** with the bounds written out, and the comment now points there instead of asserting the false invariant.
* **2 — the tolerance is broader than its argument.** Agreed; "a narrower mapping, never a wrong result" is gone from both the doc comment and the PR body. It now states both limits: analyzed fields can match *wider*, and the blanket give-up would break `finish_generated_run`'s `ensure!`-guarded `run_id` + `doc_kind` read. Bound checked rather than assumed — `run_id` has been declared `keyword` since autoindex's first commit (`0f3ef60d`, `git log -S`), so no catalog created by `ensure_index` can be in that state; one auto-created by a stray document write could be.
* **3 — the `.keyword` multi-field option.** Not adopted, and now not silent either: it is a third term rather than a substitute (a declared-keyword `prefix` from rc.68 has no `.keyword` subfield), and it would widen the sweep's blast radius on exactly the legacy catalogs note 1 is about. It stays the natural companion to #890 and is recorded there.
* **4 — asymmetric tolerance on the legacy graph path.** Fixed: that path's `corpus_scope` PUT goes through `install_catalog_mapping` instead of a hard `?`, so it degrades with a warning rather than aborting. The warning also stopped lying in that corner — if the scope field itself is the casualty it says no exactly matchable scope remains, instead of claiming scoping "travels on" the field it just gave up (`a_legacy_scope_field_is_not_described_as_carrying_the_scope`).
* **5 — cost.** Noted, unchanged: two extra `delete_by_query` round trips per excluded entry, at the sizes exclusion sweeps run at.

### Gates

`cargo fmt --all -- --check` clean · `cargo clippy --release -j 8 -p xerj-autoindex --all-targets -- -D warnings` clean · `cargo test --release -j 8 -p xerj-autoindex` **743 + 14 passed, 0 failed**. No server booted and no ES-YAML run locally, per the working rules — CI owns that gate; the change stays inside `xerj-autoindex` with no query or mapping semantics touched.

### Residual risk

The upgrade path is now pinned by test, but only against the contract as v1.0.0-rc.71 shipped it — a build older than rc.68 whose `catalog_mapping()` differed for other reasons is not covered by these digests (nor was it before). And the identity comparisons are still exact-equality with no tolerance of their own: this PR keeps the digest still rather than teaching them to forgive a known-safe evolution, so the next genuine catalog-contract change will need that upgrade path built.

xerj-org:
Addendum — a pre-existing flake you may hit re-running this suite, so it is not mistaken for this PR.

Repeating `cargo test --release -j 8 -p xerj-autoindex --lib` six times on each tree:

| tree | failures in 6 runs |
| --- | --- |
| `fbcf3fa3` (this PR before the review fixes) | **2** — `only the one changed file must be parsed; the unchanged file stays skipped` |
| `edb3c449` (current head) | **1** — `unchanged re-index must parse zero files` |

Same test both times, `unchanged_reindex_skips_the_parse_and_stays_byte_identical`, and the cause is independent of anything either commit touches: it measures with the process-global `SCAN_FILE_PARSED` counter while holding only `HTTP_E2E_LOCK`, so any concurrently running test that parses files (`failure_resume_http_tests` holds a *different* mutex) lands in its measurement window. Adding tests only re-times which run loses. Filed as **#891** with the repro and two fix directions; deliberately not fixed here, since it is neither #755 nor introduced by this branch.

Run serially (`--test-threads=1`) the suite is 743/743 green, which is the same evidence from the other side.

CI on `edb3c449`: Format + Clippy, ES-compat YAML conformance, Autoindex FD smoke (ubuntu + macos), Use-case harnesses, API smoke, Release build (fat LTO), ONNX, Fuzz, and the guards all pass. `Build + Test` and `Autoindex FD smoke (windows-latest)` were still running when this was written. `Build + Test` is already red on `main` at `5f3b6ea3`, and for an unrelated reason — the workspace test action times out at 12 minutes on a hang in `script_bucketed_agg_past_the_call_depth_limit_is_an_error_not_empty_buckets` (`xerj-api` script limits), nothing to do with `xerj-autoindex`.

PR #887, opened 2026-08-30
fix(es-compat): numeric and boolean fields enforce their declared type (#781)

### LINKED ISSUE #781: es-compat: numeric/boolean fields don't enforce declared type (integer keeps 1.9, accepts out-of-range / non-numeric) — diverges from ES coerce/reject
Found via a proactive coercion probe on current main (rc.70, own node, non-9200). XERJ does not enforce a field's declared numeric type the way ES does — it accepts values ES coerces-or-rejects and stores them verbatim. Data is NOT corrupted (values round-trip and query on their actual value), but it diverges from ES-8.x wire behavior, and one case is a wrong-results divergence, not just leniency.

Reproduced (field `i` mapped `integer`, `b` mapped `boolean`), each returns `201 created` in XERJ:

| input | XERJ | ES 8.x |
|---|---|---|
| `{"i":"abc"}` | accepted (stored `"abc"`) | 400 mapper_parsing_exception |
| `{"i":1.9}` | accepted, kept **1.9** | coerce:true **truncates → 1** |
| `{"i":9999999999}` (> i32 max) | accepted, kept exact | 400 out_of_range |
| `{"b":"yes"}` | accepted | 400 (only `true`/`false`) |

The **wrong-results** one: with `i:1.9` in an `integer` field, `range {"i":{"gte":1.5}}` **matches** in XERJ (it indexed 1.9) but would NOT match in ES (ES indexes the truncated `1`). `term {"i":1}` matches in ES, returns 0 in XERJ. So the same query over the same "integer" field returns different hits.

`{"i":"abc"}` is stored but not numerically queryable — a silent data-quality gap for callers relying on ES rejecting bad values.

Question for the maintainer: is numeric-type leniency intentional (XERJ is schemaless-friendly; there are "lenient by design" notes elsewhere, e.g. date parsing), or should `integer`/`float`/`boolean` fields enforce ES coercion/rejection? If leniency stays, at minimum the `integer`-field range/term semantics diverging from ES (the 1.9 case) is worth documenting so callers don't assume integer semantics. Not filing a fix — this is a strictness-vs-leniency decision.

xerj-org:
## What was wrong

A field mapped `integer` enforced that type for **nothing**. Every one of these
answered `201 created` on rc.71:

| input into a field mapped `integer` / `boolean` | before | ES 8.x |
|---|---|---|
| `{"i": 1.9}` | stored `1.9` | `1` (coerce truncates) |
| `{"i": "5"}` | stored the string `"5"` | `5` |
| `{"i": "abc"}` | stored the string | **400** |
| `{"i": 9999999999}` | kept exact | **400** out of range |
| `{"i": {"bad":"x"}}` | indexed the object | **400** |
| `{"b": "yes"}` | stored the string | **400** |

The issue's own framing is right that this is more than leniency: it is a
**wrong-results** divergence. With `1.9` sitting in an `integer` field,
`range {"i":{"gte":1.5}}` **matched** in XERJ and would not in ES (which
indexes the truncated `1`), while `term {"i":1}` matched in ES and returned
nothing here. Same mapping, same document, same query, different answers.

## Root cause

No write path ever compared a value against the type the mapping declared.

`es_compat::apply_ignore_malformed` does walk the mapping, but it returns early
unless the field carries `ignore_malformed: true` — it is the *lenient*
handler, not a validator. The strict counterpart existed only for dates
(`bulk::find_bad_date_field` → `xerj_query::dates::date_value_valid_with_format`);
nothing equivalent existed for numbers or booleans. `index_doc`,
`index_doc_auto` and `create_doc` handed the parsed body straight to the
engine, and `_bulk`'s turbo batch path does not even parse the document — it
forwards the raw NDJSON bytes to the WAL.

## The fix — write side

One shared predicate, `xerj_common::field_coercion`
(`engine/crates/xerj-common/src/field_coercion.rs`, new), implementing ES 8.x's
rules with `coerce` defaulting to `true`:

* **Coerced** (ES indexes a different value, so XERJ now stores it): `1.9` →
  `1` truncated toward zero, `"5"` → `5`, `"1.9"` → `1`, `"true"`/`"false"` →
  the booleans. A value that already equals what ES would index (`2.0`, `7`) is
  left alone so `_source` is not churned.
* **Refused** with a 400 `document_parsing_exception` (a per-item 400 under
  `_bulk`): out of range for the declared width, an unparseable string, an
  object, an array whose element is either of those, a non-finite float, and
  anything but `true`/`false` in a `boolean` — `1` and `0` included, as in ES 8.
* `"coerce": false` **is honoured**: it additionally refuses a decimal part and
  a string. The range check is not a coercion and applies either way.
* `"ignore_malformed": true` still wins — those fields are skipped entirely.
* `"enabled": false` objects are skipped: ES parses nothing inside one.

Integral bounds are carried as `i128`, not `f64` — an `f64` bound rounds
`i64::MAX` up to 2^63 and would have let the integer token
`9223372036854775808` pass as a `long`. The float **token** needs its own
handling and did not get it in the first cut: ES compares the raw double
against a bound Java widened to `double`, so `9223372036854775808.0` passes
ES's check and ES's `(long)` cast then saturates it, while a bare `fract() == 0`
shortcut stored the float verbatim in a `long` field. It is now truncated and
clamped in the integer domain. `half_float` is bounded at half precision's own
overflow (65520, where round-to-nearest goes infinite), not at `f32`'s.

## The fix — query side

Rewriting `_source` moves the **stored spelling**, so a query operand written
the pre-coercion way has to keep matching. It did not, and the ES-YAML suite
caught it: `search/390_doc_values_search.yml` indexes `boolean: "false"` /
`"true"` — JSON *strings* — into a `boolean` field and asserts
`terms {boolean: ["false","true"]}` returns 2. Once ingest stores real
booleans, that returned 0.

Two halves, and neither works alone:

1. **The operand is canonicalised.** `rewrite_query_aliases` — the same
   mapping-aware hook the `date` (#788/#799) and `ip` CIDR (#782/#801)
   normalisations live in — runs a `term`/`terms` value on a declared `boolean`
   through `field_coercion::coerce_value`, the write path's own predicate. A
   value the predicate refuses (`"yes"`, `1`) is left untouched: a search must
   not 400 on a value ingest would have rejected.
2. **The comparator relates the two spellings, in both directions.** An index
   that spans this change holds documents in *both* representations, so
   canonicalising the operand alone would miss the older half.
   `json_values_equal` now treats a boolean and its string spelling as equal
   either way round, mirroring the number/string pair beside it, and is split
   into `json_scalar_equal` so the **array** arm recurses through the same
   tolerance instead of comparing elements with bare `==` — without that, the
   same document answers differently depending on whether its field holds one
   value or several.

## Where it is wired

| path | file | note |
|---|---|---|
| `POST /{index}/_doc` | `es_compat.rs` `index_doc_auto` | after the ingest pipeline |
| `PUT /{index}/_doc/{id}` | `es_compat.rs` `index_doc` | after the ingest pipeline |
| `PUT /{index}/_create/{id}` | `es_compat.rs` `create_doc` | this endpoint had **no** mapping-aware validation at all |
| `_bulk`, all actions | `bulk.rs`, partition loop | before the turbo/slow split |
| `_bulk`, pipeline actions | `bulk.rs`, per-item loop | ES parses *after* the pipeline, so these are checked there |
| `term` / `terms` operands | `index.rs` `rewrite_query_aliases` | declared `boolean` fields |

The `_bulk` check had to run at **partition** time rather than in the per-item
loop: plain `index` actions go through `index_batch_turbo_raw`, which never
parses the body, so every bad value in an auto-id bulk was otherwise a silent
201 (`the_bulk_turbo_path_is_not_a_way_around_the_check` pins that). Rather
than divert those actions to the slow path — which would have cost every
mapped-numeric index its turbo batch — the check coerces in place and only
re-serializes when a value actually changed. A `mapping_has_enforced_types`
gate keeps an index with no numeric or boolean field off the parse entirely,
mirroring the existing `index_has_strict_date` / `index_has_dynamic_copy`
gates.

**Ingest cost, stated rather than elided.** The gate cache carries the
properties map itself (`Option<Arc<Map>>`), shared by both enforcement sites,
so a mapped-numeric index pays a refcount bump per action instead of a deep
clone of the whole `properties` blob. Two costs remain and are not measured
here: an index declaring a numeric or boolean field pays one
`serde_json::from_slice` per `index`/`create` action on the HTTP worker (~5
µs/doc, the parse M5.11 removed), and an explicit-id `index` action has its
parsed tree dropped at partition time and reparsed in the batch loop.

## Upgrade behaviour — read this if you have data

Because the fix changes what is **stored**, an index that spans it holds both
representations of the same logical value: documents written before hold
`"true"` / `"5"`, documents written after hold `true` / `5`, and no reindex is
performed for you.

That mixed index still answers correctly. Verified by hand on a throwaway node,
one index holding `{"b": "true", "n": "5"}` (planted through `_update`, which
is deliberately not re-validated) next to `{"b": true, "n": 5}`:

```
terms {b:["true"]} -> 2   term {b:"true"} -> 2   terms {n:["5"]} -> 2
terms {b:[true]}   -> 2   term {b:true}   -> 2   term  {n:5}     -> 2
```

and a `terms` aggregation on the same field buckets both documents into one
(`key: 1, key_as_string: "true", doc_count: 2`). Pinned as
`an_index_holding_both_spellings_answers_every_query_with_both_docs`. Reindex
if you want a uniform `_source`; nothing breaks if you do not. Documented in
CHANGELOG.md and in the module header.

**That test found a third defect, and it is the one that mattered most.** It
passed on a 32-core box and failed on CI's 2-core runner — the coerced document
vanished from `terms {"b":["true"]}` while the pre-upgrade one was returned.
Not flaky: `build_doc_value_columns` collects a segment's values into a
`numeric` map (numbers *and* booleans) and a `keyword` map (strings) and then
writes both into one column map, so a field that arrives **type-mixed inside a
single segment** lands in both — and the keyword pass runs second and
overwrites the numeric column. Every number/boolean document is then filed as
NULL in a column consumers treat as exact, `build_term_prefilter_cached`
returns a set that is not a superset, and the stored scan never sees the
dropped rows. A confident wrong answer. On a many-core box the flush scatters
the two documents into separate homogeneous segments, so every column is honest
and nothing shows.

Fixed the way the multi-valued case immediately above it in the same function
already handles the identical hazard — poison the field, ship **no** column,
and let every consumer take the abandon-to-scan path it already takes for an
absent field. Reproduced and re-verified in one command:

```
taskset -c 0,1 target/release/deps/numeric_fields_enforce_their_type-…
  before: test result: FAILED. 16 passed; 1 failed
  after:  test result: ok.     17 passed; 0 failed
```

Pinned directly on the builder as
`build_doc_value_columns_ships_nothing_for_a_type_mixed_field`.

## Scope this deliberately does *not* cover

* **Only declared mappings.** `engine.index_mappings` is written only by
  index-create-with-mappings and `PUT /_mapping`, never by dynamic inference.
  A schemaless index is completely unaffected.
* **A coerced value is rewritten in `_source`.** ES keeps `_source` verbatim
  and coerces only the *indexed* value, so ES returns `1.9` from `_source`
  where XERJ now returns `1`. XERJ indexes from the stored source, so agreeing
  with ES on *hits* costs source fidelity here — and wrong hits are the worse
  of the two divergences. (Validating without rewriting was raised in review
  and does not work for this engine: it would leave `1.9` indexed as `1.9`,
  i.e. it would not fix the case the issue was filed for.)
* **An empty string is left alone** rather than dropped.
* **`scaled_float` is not quantised** by its `scaling_factor` — ES indexes
  `Math.round(value * scaling_factor)`; reproducing that would rewrite
  `_source` into a lossy value for no wrong-results gain.
* **Only the field-level `"coerce": false`** is read, not the index-level
  `index.mapping.coerce` setting.
* **`_update` is not revalidated** — it merges into a document that was already
  validated on write, matching how the existing date check scopes itself.
* **`match` on a declared `boolean` whose `_source` still holds the string** is
  invisible to the `term` fast paths (`LeafConst::BoolEq` probes a numeric
  column such a document does not have). That predates this change and is not
  touched by it.

## Test proof

`engine/crates/xerj-api/tests/numeric_fields_enforce_their_type.rs` — 17 tests
through the HTTP router, plus 19 unit tests in the new module and one in
`index.rs` pinning the operand rewrite.

**Verified to fail on unfixed code.** Stashing the source edits and leaving the
ingest tests in place gives `12 failed; 3 passed`, including
`wrong_results_the_declared_type_used_to_produce`, `an_object_is_not_a_number`
and `the_bulk_turbo_path_is_not_a_way_around_the_check`; the three that pass
pre-fix are exactly the guard tests that assert nothing changes.

The two query-side tests fail on cc3cd6f0 (this branch's pre-fix head) with
precisely CI's symptom:

```
terms {"b":["true"]} … expected 1, got 0
terms {"b":["true"]} … expected 2, got 1
  hits: [{"_id":"legacy","_source":{"k":"a","b":"true","i":"5"}}]
```

Both need an explicit `_refresh` and that is load-bearing: while a document is
memtable-resident, `term`/`terms` are answered by the doc-values fast path,
which compares stringified values and cannot see the spelling difference.
Written without the refresh, both tests pass on the broken code.

## Gates

* `cargo fmt --all --check` — clean.
* `cargo clippy --release -j 8 -p xerj-common -p xerj-engine -p xerj-api
  --all-targets -- -D warnings` — clean.
* `cargo test --release -p xerj-common` — 110 passed, 0 failed (+3 doctests).
* `cargo test --release -p xerj-engine` — 1129 passed, 0 failed, 68 binaries.
* `cargo test --release -p xerj-api` — 529 passed, 0 failed.
* the same `xerj-api` suite under `taskset -c 0,1`, to hold the 2-core runner
  shape that hid the doc-values defect above.

**ES-YAML conformance — the honest record.** The first head of this branch
(cc3cd6f0) was **RED** on it: `1365 passed / 1 failed`,
`search/390_doc_values_search.yml — Test terms query on boolean field where
only doc values are enabled — expected 2, got 0`. The original version of this
PR body listed five green gates and said of conformance "CI's gate is the check
that matters", which read as *expected to pass* about a check that had already
failed; and the pre-check it described was scoped to the wrong shape — it
looked for cases asserting a `_source` value ES would have coerced, while the
case that broke asserts hit counts. That is corrected here rather than
restated.

The current head reproduces and then clears that exact file against a throwaway
node on a private port with a temp data dir:

```
before:  22 passed · 1 failed · 0 skipped   (Test terms query on boolean field …)
after:   23 passed · 0 failed · 0 skipped
```

and CI's own conformance job on this branch reports
`1366 passed · 0 failed · 3 skipped · 1369 total`. CI remains the gate.

Closes #781.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,36 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A field mapped `integer`, `long`, `float` or `boolean` now enforces that
+  type on write** ([#781](https://github.com/xerj-org/xerj/issues/781)). The
+  declared type was enforced for nothing: `1.9` into an `integer` field was
+  stored as `1.9`, `9999999999` was kept exact, `"abc"` was stored as a
+  string, `"yes"` went into a `boolean`, and — per the issue's follow-up — an
+  entire nested object `{"bad":"x"}` was indexed into an `integer`. Every one
+  of those answered `201 created`.
+
+  The consequence was wrong hits, not just leniency. With `1.9` sitting in an
+  `integer` field, `range {"i":{"gte":1.5}}` **matched** in XERJ and would not
+  in ES (which indexes the truncated `1`), while `term {"i":1}` matched in ES
+  and returned nothing here — the same query, over the same document, under
+  the same mapping, giving different answers.
+
+  Ingest now applies ES 8.x's own rules (`coerce` defaults to `true`): `1.9`
+  → `1`, `"5"` → `5`, `"1.9"` → `1`, `"true"` → `true`; out-of-range for the
+  declared width, an unparseable string, an object, an array element that is
+  none of those, and anything but `true`/`false` in a `boolean` are refused
+  with a 400 `document_parsing_exception` — a per-item 400 under `_bulk`. An
+  explicit `"coerce": false` additionally refuses a decimal part and a string,
+  and `"ignore_malformed": true` still wins: that field's bad values are
+  dropped into `_ignored` rather than failing the document.
+
+  Two deliberate narrowings, both documented in
+  `xerj_common::field_coercion`: a coerced value is rewritten in `_source`
+  (ES keeps `_source` verbatim and truncates only the indexed value — XERJ
+  indexes from the stored source, so matching ES's *hits* costs source
+  fidelity here), and the index-level `index.mapping.coerce` setting is not
+  read, only the field-level parameter.
+
 - **Wrapping a query in a one-clause `bool` no longer changes its `_score` or
   its ranking** ([#399](https://github.com/xerj-org/xerj/issues/399)).
   `{"bool":{"must":[X]}}` and bare `X` are the same query — Lucene's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -39,6 +39,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   fidelity here), and the index-level `index.mapping.coerce` setting is not
   read, only the field-level parameter.
 
+  Because that canonicalisation moves a declared `boolean` from the string
+  spelling to a real JSON boolean in `_source`, the **query** side now runs
+  `term` / `terms` values on a `boolean` field through the same predicate —
+  ES parses a query value with the same `Booleans` its field mapper uses, so
+  `"true"` and `true` name one term. Without it, `terms {"b":["true"]}` found
+  nothing on a doc-values-only boolean field while the equivalent `term` still
+  matched. The `_source` scan comparator relates the two spellings in both
+  directions, so documents written before this release — which still hold the
+  string — keep matching either way.
+
 - **Wrapping a query in a one-clause `bool` no longer changes its `_score` or
   its ranking** ([#399](https://github.com/xerj-org/xerj/issues/399)).
   `{"bool":{"must":[X]}}` and bare `X` are the same query — Lucene's
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,21 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Fixed
 
+- **A field holding both numbers and strings in one segment no longer loses the
+  numeric documents from `term` / `terms`.** The doc-values builder collects a
+  segment's values into a numeric map (numbers and booleans) and a keyword map
+  (strings), then writes both into one column set — so a *type-mixed* field had
+  its numeric column silently overwritten by the keyword one, leaving every
+  number/boolean document filed as null in a column the query prefilter treats
+  as exact. Those documents were then dropped from the result with no error.
+  Such a field now ships **no** doc-values column at all, exactly as a
+  multi-valued field already does, and every consumer falls back to the
+  stored-source scan: correct, at scan cost rather than column speed. Reindex
+  onto a single type to get the column back. Whether it bit you depended on
+  core count — a many-core host scatters a flush across shards and each segment
+  stays single-typed, so this reproduced only where documents of both kinds
+  landed in one segment.
+
 - **A field mapped `integer`, `long`, `float` or `boolean` now enforces that
   type on write** ([#781](https://github.com/xerj-org/xerj/issues/781)). The
   declared type was enforced for nothing: `1.9` into an `integer` field was
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,22 +32,35 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
   and `"ignore_malformed": true` still wins: that field's bad values are
   dropped into `_ignored` rather than failing the document.
 
-  Two deliberate narrowings, both documented in
-  `xerj_common::field_coercion`: a coerced value is rewritten in `_source`
-  (ES keeps `_source` verbatim and truncates only the indexed value — XERJ
-  indexes from the stored source, so matching ES's *hits* costs source
-  fidelity here), and the index-level `index.mapping.coerce` setting is not
-  read, only the field-level parameter.
+  **What this changes about stored data, and what it means on upgrade.** A
+  coerced value is rewritten in `_source`: ES keeps `_source` verbatim and
+  coerces only the indexed value, but XERJ indexes from the stored source, so
+  matching ES's *hits* costs source fidelity here. A document written
+  `{"i": 1.9, "b": "false"}` now reads back `{"i": 1, "b": false}`. An index
+  that spans this upgrade therefore holds **both** spellings of the same
+  logical value — documents written before hold `"false"` / `"5"`, documents
+  written after hold `false` / `5` — and no reindex is performed for you.
 
-  Because that canonicalisation moves a declared `boolean` from the string
-  spelling to a real JSON boolean in `_source`, the **query** side now runs
-  `term` / `terms` values on a `boolean` field through the same predicate —
-  ES parses a query value with the same `Booleans` its field mapper uses, so
-  `"true"` and `true` name one term. Without it, `terms {"b":["true"]}` found
-  nothing on a doc-values-only boolean field while the equivalent `term` still
-  matched. The `_source` scan comparator relates the two spellings in both
-  directions, so documents written before this release — which still hold the
-  string — keep matching either way.
+  That mixed index still answers correctly, and closing that gap is part of
+  this change. Because the canonicalisation moves a declared `boolean` from
+  the string spelling to a real JSON boolean in `_source`, the **query** side
+  now runs `term` / `terms` values on a `boolean` field through the same
+  predicate — ES parses a query value with the same `Booleans` its field
+  mapper uses, so `"true"` and `true` name one term. Without it,
+  `terms {"b":["true"]}` found nothing on a doc-values-only boolean field
+  while the equivalent `term` still matched. The `_source` scan comparator
+  relates the two spellings in both directions (as it already did for a number
+  and its string spelling), for single- and multi-valued fields alike, and a
+  `terms` aggregation buckets them together — so one query sees the whole
+  index rather than half of it. **Reindex** if you want a uniform `_source`;
+  nothing breaks if you do not.
+
+  Also narrowed deliberately, all documented in `xerj_common::field_coercion`:
+  the index-level `index.mapping.coerce` setting is not read (only the
+  field-level parameter), `scaled_float` is range-checked but not quantised by
+  its `scaling_factor`, an empty string is left alone rather than dropped, and
+  `_update` is not re-validated — it merges into a document already checked on
+  write.
 
 - **Wrapping a query in a one-clause `bool` no longer changes its `_score` or
   its ranking** ([#399](https://github.com/xerj-org/xerj/issues/399)).
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -2831,6 +2831,13 @@ pub async fn index_doc_auto(
         doc
     };
 
+    // Declared numeric/boolean types are enforced after the pipeline has run,
+    // so a pipeline-produced value is held to the same rules as a caller's.
+    let doc = match enforce_field_types(&state, &index, "", doc) {
+        Ok(d) => d,
+        Err(resp) => return resp,
+    };
+
     let idx = match state.engine.get_or_create_index(&index) {
         Ok(i) => i,
         Err(e) => return ApiError::new(xerj_common::XerjError::from(e)).into_response(),
@@ -2959,6 +2966,13 @@ pub async fn index_doc(
     };
     // ?refresh=true|wait_for — accepted silently; memtable is immediately visible.
 
+    // Declared numeric/boolean types are enforced after the pipeline has run,
+    // so a pipeline-produced value is held to the same rules as a caller's.
+    let doc = match enforce_field_types(&state, &index, &id, doc) {
+        Ok(d) => d,
+        Err(resp) => return resp,
+    };
+
     let idx = match state.engine.get_or_create_index(&index) {
         Ok(i) => i,
         Err(e) => return ApiError::new(xerj_common::XerjError::from(e)).into_response(),
@@ -3083,6 +3097,13 @@ pub async fn create_doc(
         }
     } else {
         doc
+    };
+
+    // `_create` skipped every mapping-aware validation stage; the declared
+    // numeric/boolean types are now enforced here too (issue #781).
+    let doc = match enforce_field_types(&state, &index, &id, doc) {
+        Ok(d) => d,
+        Err(resp) => return resp,
     };
 
     let idx = match state.engine.get_or_create_index(&index) {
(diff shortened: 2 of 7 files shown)


xerj-org:
## Response to review — all four blocking findings addressed

Head is now `8eccf901`, on top of `1fff271b`. The PR body has been rewritten;
this comment says what changed and what I verified.

### 1 + 2 — conformance regression and its mechanism · **accepted, fixed**

Correct on both counts, and reproduced before touching anything, on a throwaway
node on a private port with a temp data dir:

```
es-yaml-runner --file …/search/390_doc_values_search.yml
  22 passed · 1 failed
  (Test terms query on boolean field where only doc values are enabled
   — length hits.hits: expected 2, got 0)
```

`1fff271b` fixed the mechanism the way the review named first — canonicalise
the **query operand** through the same `field_coercion` predicate the write
path uses, in `rewrite_query_aliases` (the hook `date`/#799 and `ip` CIDR/#801
already live in) — plus the `Bool`↔`String` cross-type arm in
`json_values_equal`. Same file, after: **23 passed · 0 failed**. Its own run of
the whole 200-file suite was 1366 / 0 / 3.

`8eccf901` adds what that leaves open:

* **Regression tests.** Two HTTP tests in
  `numeric_fields_enforce_their_type.rs`:
  `a_query_in_the_pre_coercion_spelling_still_matches_the_coerced_document`
  (the conformance case in-process, plus its negative direction — `"false"`
  and `"yes"` must still be 0) and
  `an_index_holding_both_spellings_answers_every_query_with_both_docs`.
  On `cc3cd6f0` they fail with exactly CI's symptom:

  ```
  terms {"b":["true"]} … expected 1, got 0
  terms {"b":["true"]} … expected 2, got 1
    hits: [{"_id":"legacy","_source":{"k":"a","b":"true","i":"5"}}]
  ```

  A trap worth recording, because it nearly produced two tests that prove
  nothing: **both need an explicit `_refresh`**. While a document is
  memtable-resident, `term`/`terms` are answered by the doc-values fast path,
  which compares *stringified* values and cannot see the spelling difference at
  all. Written without the refresh, both tests pass on the broken code — I
  wrote them that way first and they did. The ES YAML case does
  `indices.refresh` before it queries, which is why CI caught what an
  in-process test would not have.

* **The multi-valued half of the same comparator.** `json_values_equal` is
  split into `json_scalar_equal` so its ARRAY arm recurses through the same
  tolerance instead of comparing elements with bare `==`. Without it the same
  document answers differently depending on whether its field holds one value
  or several: `{"b": "true"}` matches `term {b: true}` through the new boolean
  arm, `{"b": ["true"]}` does not.

> On the alternative the review offered — validate and reject without
> rewriting `_source`: considered, and it does not work for this engine. XERJ
> indexes from the stored source, so validate-only leaves `1.9` in an `integer`
> field *indexed* as `1.9` — `range {gte: 1.5}` still matches and `term {i: 1}`
> still misses. It would drop the one wrong-results case the issue was filed
> for while keeping every rejection. The query-operand route (the review's own
> first option) is what makes both sides agree.

### 3 — upgrade hazard · **accepted; now handled, not only documented**

Verified on a live node: one index holding `{"b": "true", "n": "5"}` (planted
through `_update`, which is deliberately not re-validated — the shape a
pre-upgrade document has) next to `{"b": true, "n": 5}`:

```
-- stored --
legacy {'k': 'a', 'b': 'true', 'n': '5'}
new    {'b': True, 'n': 5, 'k': 'b'}
-- queries --
terms {b:["true"]} -> 2   term {b:"true"} -> 2   terms {n:["5"]} -> 2
terms {b:[true]}   -> 2   term {b:true}   -> 2   term  {n:5}     -> 2
```

and a `terms` aggregation over that field buckets both documents into one
(`key: 1, key_as_string: "true", doc_count: 2`). The finding's exact failure
mode — one query matching the old documents and missing the new ones — is what
the second regression test asserts against, and it is the `got 1` line quoted
above. Documented as well: a new "what this changes about stored data, and what
it means on upgrade" paragraph in `CHANGELOG.md` naming both representations
and stating plainly that no reindex is performed and none is required, and a
module-header section in `field_coercion.rs` tying the read-side and write-side
halves together so neither is removed without the other.

### 4 — honest claims · **accepted, prose fixed**

The PR body listed five green gates and said of conformance "CI's gate is the
check that matters" while conformance was already RED on that exact head, and
the pre-check it described was scoped to the wrong shape (it looked for cases
asserting a `_source` value ES would coerce; the case that broke asserts hit
counts). The body now says that, under **ES-YAML conformance — the honest
record**, with the before/after numbers rather than an implication.

### Non-blocking notes

1. **Ingest perf.** Fixed the free part: the gate cache now carries
   `Option<Arc<Map>>` — gate *and* properties together — shared by both
   enforcement sites, so a mapped-numeric index pays a refcount bump per action
   instead of a deep clone of the whole `properties` blob per document. The two
   remaining costs (one `from_slice` per `index`/`create` action on a
   mapped-numeric index; the explicit-id double parse) are now *stated* in the
   PR body instead of elided. Not measured — I did not want to publish a number
   I had not run.
2. **Test coverage gap.** Closed — see above.
3. **Residual leniency.** All three fixed, and the commit message no longer
   overclaims the `i128` bound:
   * float token — ES compares the raw double against a bound Java widened to
     `double`, so `9223372036854775808.0` passes and ES's `(long)` cast
     saturates it; XERJ stored the float verbatim. Now truncated and clamped in
     the integer domain. (The comparison must be made in `i128`: `i64::MAX as
     f64` is 2^63, so an f64 round-trip agrees on exactly the value that must
     not pass. My first attempt did that and the new test caught it.)
   * the same value as a **string** — `"9223372036854775808"` slipped through
     for the same reason; the exact-integer parse is now range-checked in
     `i128` too.
   * `half_float` — was checked against `f32` (finite to 3.4e38). ES
     round-trips through 16 bits; round-to-nearest maps everything below 65520
     onto 65504, so the boundary is 65520. `65510` accepted, `65520` refused,
     both pinned.
   * `scaled_float` is still not quantised by `scaling_factor` — now stated as
     a deliberate narrowing rather than left implicit. Quantising would rewrite
     `_source` into a lossy value for no wrong-results gain.
4. **`enabled: false` objects** — now skipped, since ES parses nothing inside
   one and the walk keys its recursion off `properties`, which a disabled
   object may still declare. Verified live: `{"meta": {"n": "not a number"}}`
   under a disabled `meta` is stored verbatim instead of 400-ing.
5. Nothing re-checked from the "things I checked that are fine" list.

### Gates on this head

* `cargo fmt --all --check` — clean.
* `cargo clippy --release -p xerj-common -p xerj-engine -p xerj-api
  --all-targets -- -D warnings` — clean.
* `cargo test --release -p xerj-common` — 110 passed, 0 failed (+3 doctests).
* `cargo test --release -p xerj-engine` — 1128 passed, 0 failed, 68 binaries.
* `cargo test --release -p xerj-api` — 529 passed, 0 failed.
* `search/390_doc_values_search.yml` on a throwaway node, port 9757, temp data
  dir — 23 passed, 0 failed (22 / 1 before).

The full ES-YAML suite is CI's, and it is the check that decides this.

xerj-org:
## Follow-up: the regression test for finding 3 found a deeper defect

Head is now `ac3b4a4d`.

The test I added for review finding 3 —
`an_index_holding_both_spellings_answers_every_query_with_both_docs` — **passed
here and failed on CI**. That is worth writing down, because the honest reading
is that my previous comment's claim ("a mixed index still answers correctly")
was true on a 32-core box and **false on a 2-core one**, which is the shape most
people actually run.

```
Build + Test (CI, 2 cores)
  an_index_holding_both_spellings_answers_every_query_with_both_docs ... FAILED
  terms {"b":["true"]} … expected 2, got 1
    hits: [{"_id":"legacy","_source":{"k":"a","b":"true","i":"5"}}]
```

The document that vanished is the **coerced** one — the one #781 creates.

### Root cause

`build_doc_value_columns` collects a segment's values into two maps by kind:
`numeric` (numbers *and* booleans, as f64 bits) and `keyword` (strings). Both
are then written into one `BTreeMap<String, Column>`, and the keyword loop runs
second — so for a field that is **type-mixed inside a single segment**,
`out.insert(name, Column::Keyword(..))` overwrites the numeric column. Every
number/boolean document is left filed as NULL in a column that
`build_term_prefilter_cached` treats as exact; the prefilter returns a set that
is not a superset, and the stored scan never sees the dropped rows. A confident
wrong answer, not a slow one.

This is the same failure shape as the multi-valued poisoning immediately above
it in that function, whose own comment already names the runner dependence:
*"the 2-core CI shape; ≥32-core boxes scatter docs across shards"*. Here the
scatter put the two documents in separate homogeneous segments, every column was
honest, and nothing showed.

**#781 is what makes this reachable in ordinary use.** Before it, a declared
`boolean` field kept whatever arrived and an index stayed on one spelling.
Ingest coercion canonicalises `"true"` → `true`, so an index written across the
upgrade holds both — and one flush lands them in one segment.

### Fix

Poison the field exactly as the multi-valued case does: a name present in both
maps is removed from both, the segment ships **no** column for it, and every
consumer takes the abandon-to-scan path it already takes for an absent field.
Correct answer at scan cost, instead of a wrong answer at column speed.

Reproduced and re-verified in one command, which is what made it findable at
all:

```
taskset -c 0,1 target/release/deps/numeric_fields_enforce_their_type-…
  before: test result: FAILED. 16 passed; 1 failed
  after:  test result: ok.     17 passed; 0 failed
```

Pinned directly on the builder as
`build_doc_value_columns_ships_nothing_for_a_type_mixed_field`, next to the
existing `_`-prefix test, so it fails without a running node.

### Gates re-run on `ac3b4a4d`

* `cargo fmt --all --check`; clippy over `xerj-common` / `xerj-engine` /
  `xerj-api` with `--all-targets -D warnings` — clean.
* `cargo test --release -p xerj-common` — 110 passed, 0 failed (+3 doctests).
* `cargo test --release -p xerj-engine` — 1129 passed, 0 failed, 68 binaries.
* `cargo test --release -p xerj-api` — 529 passed, 0 failed.
* the same `xerj-api` suite under **`taskset -c 0,1`** — 529 passed, 0 failed.
  Running it that way is now part of the record for this change; the previous
  head was green on 32 cores and red on 2.
* `search/390_doc_values_search.yml` against a throwaway node on port 9757 with
  a temp data dir — 23 passed, 0 failed.
* the live mixed-representation index re-checked after the fix: all seven
  spellings of `term`/`terms` return both documents, and the `terms`
  aggregations still bucket them into one (`b → key 1 / "true" / 2`,
  `n → key 5 / 2`).

CI's conformance job on the previous head of this branch reported
`1366 passed · 0 failed · 3 skipped · 1369 total`, and it is re-running on this
one.

### One residual worth naming

The doc-values fix is scoped to the segment builder. A field that is type-mixed
now falls back to the stored-source scan for `term`/`terms`, aggregations and
sorts on that segment — correct, but slower than a column. That is the right
trade for data that is genuinely mixed, and a reindex removes it; the CHANGELOG
says so.

xerj-org:
## CI is green on `7109f99b` — every job

```
Build + Test                                success
ES-compat YAML conformance                  success   1366 passed · 0 failed · 3 skipped · 1369 total
Format + Clippy                             success
Release build (fat LTO, all members link)   success
API smoke + liveness + benchmark            success
Use-case harnesses (brain + MCP + autoindex) success
Autoindex FD smoke (ubuntu / macos / windows) success
ONNX feature compile + tests                success
Fuzz harnesses (cargo-fuzz)                 success
Workspace resolves · cargo-audit · UX · CLA · landing guard · refcode  success
```

The hard gate — conformance at **0 failed** — is met, from the 1 failed the
review blocked on.

One thing worth recording rather than glossing: `Build + Test` failed once on
this head before passing on a plain re-run. It was **[#751](https://github.com/xerj-org/xerj/issues/751)**,
the documented low-vCPU hang, not this change — the step logged
`script_bucketed_agg_past_the_call_depth_limit_is_an_error_not_empty_buckets
has been running for over 60 seconds` and then hit the deliberate 12-minute
bound the workflow comment describes ("normal ~2.3 min, observed hanging
18-28 min"). Checked before re-running rather than after: that test's fixture is
one document, `{"rank": 7, "tags": ["a","b"]}` — no type-mixed field, so the
doc-values change cannot reach it — and the whole binary runs in 0.22 s here
under `taskset -c 0,1`. The re-run passed with 380 `test result: ok` lines and
no failures.

Summary of the four blocking findings, for the record:

| finding | outcome |
|---|---|
| 1 · conformance 0 → 1 failed | fixed — 1366 / 0 / 3 in CI |
| 2 · boolean `terms` spelling divergence | fixed on both halves (operand canonicalised + comparator relates the spellings), pinned by two HTTP tests that fail on `cc3cd6f0` |
| 3 · upgrade hazard | handled *and* documented — and the test it demanded uncovered a real doc-values defect underneath it, now fixed |
| 4 · honest claims | PR body rewritten; the red conformance gate is stated, not implied away |

Nothing was refuted. All four were correct, and finding 3 was worth more than
it looked.

PR #888, opened 2026-08-30
fix(engine): a date field's epoch scale comes from the mapping, not from each value (#790)

### LINKED ISSUE #790: es-compat: range/term on a sub-millisecond date returns 0 after flush (sort-shadow encodes nanos, range bound is ms)

xerj-org:
Closes #790.

## The defect

XERJ maps **both** ES `date` and ES `date_nanos` onto the single native type `FieldType::Date` (`engine.rs:4602`, `es_compat.rs:16730`) and stored nothing to tell them apart. The only signal a date's resolution had was a **per-VALUE guess**, in `date_string_to_epoch` and its copy in `normalize_search_after_value`:

> ≥ 4 fractional-second digits → epoch-**nanoseconds**; anything coarser → epoch-**milliseconds**.

Because the guess is per value, **one column could hold both scales** — six orders of magnitude apart — and no single comparison could be right against both:

```
docs: w=2025-01-01T00:00:00Z   s=2020-06-06T06:06:06.123456Z   e=2019-01-01T00:00:00Z
keys: w -> 1.7356e12 (ms)      s -> 1.5915e18 (ns)             e -> 1.5461e12 (ms)

sort ts asc                    -> [e, w, s]      s is the MIDDLE instant
range gte/lte = s's own value  -> 1 pre-flush, 0 post-flush
```

The range half is #790's headline. `shadow_range_bounds` derives a range's bound through that same normaliser, so a coarser-looking bound came out in milliseconds while the segment's sort-shadow key for that doc was in nanoseconds. `shadow_range_positions` bisected to the empty set and the caller skipped the whole segment ("`Some(∅)` proves no doc can contribute"). Before flush the memtable source scan re-tests the query literally and still matched — which is exactly the before/after-flush divergence the issue reports.

## The fix

Make the scale a property of the **field**, resolved once from the mapping, and hand the same value to every date-ordered path so they cannot disagree.

- **`xerj-common`** — new `DatePrecision {Millis, Nanos}` recorded on `FieldOptions::date_precision`. `FieldConfig::new` pins a native `FieldType::Date` to `Millis` (the native API has no `date_nanos` concept — it only ever arrives as an ES mapping string); `es_properties_to_fields` and the index-template path override it to `Nanos` for a declared `date_nanos`.
- **`xerj-engine`** — `DateScale {Millis, PerValue}` and `Index::date_scale(field)`, read off `millis_date_fields`: a sync-readable snapshot of the mapping, needed because the schema lives behind an **async** `RwLock` while every consumer is sync scan-path code. The snapshot is swapped (and the shadow / range-prefilter caches retired) in `save_schema`, the one choke point every schema mutation funnels through.
- Threaded into all six sites that must stay in lockstep: `compute_sort_values` (per-hit sort keys), `build_sort_shadow` (segment shadow), `shadow_range_bounds` (**the bound is converted; the on-disk column is untouched**), `sort_epoch_memo` (now one memo per scale, so a millisecond field can never be handed a nanosecond key another field cached), `normalize_search_after_value` (cursor), and the memtable `sort_candidates_numeric` cut. `warm_segment_at_publish` has no schema access, so `sort_shadow_fields` now records each field's scale next to its name and the publish-time warm rebuilds identical keys.

**No on-disk format changes.** The shadow has always been derived at query time from the stored date string, and `date_precision` is a new optional field that older `schema.json` files simply lack.

## Why not the global always-ms toggle (PR #791, closed)

That attempt deleted the nanosecond branch outright and broke `search/240_date_nanos.yml` — the per-value guess **is** what gives a `date_nanos` field its nanosecond sort values today. So this change leaves a `date_nanos` field (and any mapping written before the flag existed) on the guess — bit-for-bit today's behaviour — and only ever **narrows** a declared `date` to milliseconds. A path that forgets to consult the mapping lands on `PerValue`, i.e. the status quo, never on a new scale.

Pinning `date_nanos` to nanoseconds for *every* value is deliberately **not** done here. `search/90_search_after.yml` ("Format sort values") pins a cross-index sort in which a `date_nanos` field holding `dd/MM/yyyy` slash-dates interleaves chronologically with a `date` field's values — i.e. on the millisecond scale, which only holds because `slash_date_to_epoch` has no nanosecond branch. Making `date_nanos` uniformly nanosecond-scaled changes that and needs its own decision; it is filed as a follow-up.

## Scope / upgrade behaviour

The flag is set when a field is **mapped**, so an index created before this build keeps the old behaviour until its dates are re-mapped (a `PUT _mapping` on an existing field is a no-op by design, so in practice that means reindexing into a new index). That is exactly the "legacy-infer default" the issue's own decision note recommends — existing `date_nanos` indices keep working, a freshly mapped `date` is pinned to milliseconds — and it is why this needs no migration. A blanket back-fill would silently downgrade existing `date_nanos` indices to milliseconds, which is the maintainer call the issue flagged and is not made here.

`GET _mapping` still renders a `date_nanos` field as `"date"` (unchanged, pre-existing): the flag is recorded internally and survives restart through `schema.json`, so normal operation round-trips correctly.

## Tests

| test | file | what it pins |
|---|---|---|
| `date_mixed_precision_sorts_and_ranges_consistently` | `index.rs` | asc, desc, a range on the sub-ms doc's **own** value, and two half-open ranges — asserted **before and after flush** |
| `date_nanos_field_keeps_nanosecond_sort_values` | `index.rs` | the guard: a `Nanos` field still emits `1540815132123456789` / `1540815132987654321`, the exact values `240_date_nanos.yml` asserts, pre- and post-flush |
| `date_scale_follows_the_declared_mapping` | `index.rs` | the mapping→scale rule and the ms-vs-ns split of `date_string_to_epoch` |
| `date_and_date_nanos_record_their_declared_precision` | `es_compat.rs` | the mapping half, including sub-properties |

**Fail-before**, proven by stubbing `Index::date_scale` to always return `PerValue` (restoring the per-value guess) and re-running:

```
sort  ->  left ["e","w","s"]  right ["e","s","w"]    (pre-flush AND post-flush)
range ->  left (0, [])        right (1, ["s"])       (post-flush)
```

## Gates

- full `xerj-engine` suite — **1130 passed / 0 failed** (68 binaries)
- full `xerj-api` suite — **513 passed / 0 failed** (60 binaries)
- `xerj-common` — 91 + 3 doc-tests green
- `cargo fmt --all --check` clean
- `cargo clippy --release -j 8 -p xerj-common -p xerj-engine -p xerj-api --all-targets -- -D warnings` clean

The ES-YAML conformance suite is left to CI — its runner does `DELETE /_all` on whatever node it points at, so it is not safe to run here. `240_date_nanos.yml` and `90_search_after.yml` are the two files this change was reasoned against, and the `date_nanos` guard test above reproduces the former's sort-value assertions in-process.

### DIFF
--- a/engine/crates/xerj-api/src/es_compat.rs
+++ b/engine/crates/xerj-api/src/es_compat.rs
@@ -16557,6 +16557,21 @@ fn es_properties_to_fields(properties: &Value) -> Result<Vec<FieldConfig>, Strin
 
         let native_type = es_type_to_native(es_type);
         let mut fc = FieldConfig::new(field_name.clone(), native_type);
+
+        // #790 — remember `date` vs `date_nanos`. Both collapse onto
+        // `FieldType::Date`, and without this flag the engine had to guess a
+        // date's scale from the value ("4+ fractional digits means nanos"),
+        // which mixed millisecond and nanosecond keys inside a single `date`
+        // column and broke both its sort order and its ranges. `FieldConfig::new`
+        // already defaults a `Date` to `Millis`; the override is what keeps a
+        // declared `date_nanos` on its own precision.
+        if native_type == FieldType::Date {
+            fc.options.date_precision = Some(if es_type == "date_nanos" {
+                xerj_common::types::DatePrecision::Nanos
+            } else {
+                xerj_common::types::DatePrecision::Millis
+            });
+        }
         if let Some(sub_props) = field_def.get("properties") {
             fc.fields = es_properties_to_fields(sub_props)?;
         }
@@ -16756,6 +16771,49 @@ fn es_type_to_native(es_type: &str) -> FieldType {
         "binary" => FieldType::Binary,
         "nested" => FieldType::Nested,
         _ => FieldType::Object,
+    }
+}
+
+#[cfg(test)]
+mod date_precision_mapping_tests {
+    use super::*;
+    use xerj_common::types::DatePrecision;
+
+    /// #790: `date` and `date_nanos` both collapse onto `FieldType::Date`, so
+    /// the mapping is the only place the distinction can be recorded. Without
+    /// it the engine has to guess a date's epoch scale from each value, which
+    /// mixes millisecond and nanosecond keys inside one `date` column.
+    #[test]
+    fn date_and_date_nanos_record_their_declared_precision() {
+        let fields = es_properties_to_fields(&json!({
+            "when": { "type": "date" },
+            "when_ns": { "type": "date_nanos" },
+            "name": { "type": "keyword" },
+            "nested": { "type": "object", "properties": { "at": { "type": "date_nanos" } } }
+        }))
+        .expect("valid mapping");
+        let by_name = |n: &str| {
+            fields
+                .iter()
+                .find(|f| f.name == n)
+                .unwrap_or_else(|| panic!("{n} missing"))
+        };
+        assert_eq!(by_name("when").field_type, FieldType::Date);
+        assert_eq!(
+            by_name("when").options.date_precision,
+            Some(DatePrecision::Millis)
+        );
+        assert_eq!(
+            by_name("when_ns").options.date_precision,
+            Some(DatePrecision::Nanos)
+        );
+        // Non-date fields carry nothing, so the flag never widens the
+        // persisted mapping for the rest of the schema.
+        assert_eq!(by_name("name").options.date_precision, None);
+        // Sub-properties go through the same recursion.
+        let inner = &by_name("nested").fields[0];
+        assert_eq!(inner.name, "at");
+        assert_eq!(inner.options.date_precision, Some(DatePrecision::Nanos));
     }
 }
 
--- a/engine/crates/xerj-common/src/types.rs
+++ b/engine/crates/xerj-common/src/types.rs
@@ -257,6 +257,31 @@ impl fmt::Display for FieldType {
 }
 
 // ═════════════════════════════════════════════════════════════════════════════
+// Date precision
+// ═════════════════════════════════════════════════════════════════════════════
+
+/// Declared resolution of a [`FieldType::Date`] field.
+///
+/// XERJ collapses ES `date` AND `date_nanos` onto the single native type
+/// [`FieldType::Date`], so before #790 the engine had no way to tell them
+/// apart and had to *guess* a value's scale from the value itself (">= 4
+/// fractional-second digits means nanoseconds"). That guess is per-VALUE, so
+/// one column could hold both millisecond and nanosecond keys — six orders of
+/// magnitude apart — and no single comparison could be right against both.
+///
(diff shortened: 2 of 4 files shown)


xerj-org:
Follow-up filed as #889 for the two things this PR deliberately leaves open: `date_nanos` still guessing its scale per value (which needs `numeric_type` sort conversion first, or `search/90_search_after.yml` breaks), and the absence of a back-fill for indices created before this build.

PR #898, opened 2026-08-31
fix(autoindex): a symbol's `code` is the whole declaration, not the name's line (#500)

xerj-org:
## What was still wrong

[#579](https://github.com/xerj-org/xerj/pull/579) promoted every declaration to
its own retrievable document, which closed the constant-and-field half of #500.
What it stored as that document's `code` was literally the **single physical
line the captured NAME node sits on**:

```rust
// engine/crates/xerj-autoindex/src/extract/code.rs (pre-fix)
let row = node.start_position().row;
let mut code = lines.get(row).copied().unwrap_or("").trim().to_string();
```

`row` is the NAME's row, so that string is the declaration only when the
declaration happens to occupy exactly that line and nothing else does. Three
ways it is not, all reproduced:

| | what was stored | why it is not an answer |
|---|---|---|
| **wrapped** | `public void configure(` | a fragment; the parameters and return type are on the next lines |
| **mislocated** | `hash_bytes(const char *p, size_t n)` for `static unsigned long\nhash_bytes(…)` | the return type / `#[inline]` / modifiers are on the line above the name and were dropped |
| **overran** | `void m(){ int local = 1; }` | the declaration shares its line with the body, so the body came too — the coarse unit #500 is about |

## The fix

`parse_symbols` now slices `[declaration head .. body start)` instead of taking
a line:

- **`declaration_node`** climbs from the captured name until the parent is a
  scope container, which finds the declaration without a per-grammar table —
  capture depth differs a lot (`function_item > identifier` in Rust,
  `field_declaration > variable_declarator > identifier` in Java,
  `export_statement > lexical_declaration > variable_declarator > identifier`
  in TypeScript).
- **`declaration_head`** walks back over contiguous attribute siblings, so
  `#[inline]\npub fn f()` is one declaration.
- **`body_start`** takes the first body-shaped node at or after the name in
  document order, so the slice stops where the implementation starts —
  including when the body is two levels down (`export const f = () => {…}`).
- The cap is the **same 400 chars** the single-line version had, and a slice
  that overruns it falls back to the start line.

The file document's `symbols` sidecar gains **`end_line`**, so a caller that
wants the implementation can address exactly `line`..`end_line` instead of
guessing "up to where the next symbol starts" — the guess that hands back a
whole class for a class and the rest of the file for the last symbol.

`end_line` rides in the array and **not** as a new field on the symbol document
on purpose: a new *scalar* field there would make
`reconcile_plan::ensure_compatible` abort the next incremental run over an index
an older binary froze (`field {name} is absent from frozen dataset`), whereas an
array field is exempt from that check by the #580/#581 rule (`acc.n == 0`). I
read that from the code; I did not run a cross-version re-index to confirm it.

## Second commit: the shape matching was wrong outside the three measured grammars

A review of the first commit found four defects, all reproduced, all now
covered by a test that fails on that commit. This is the substance of
`720f312d`.

1. **A Ruby regression the first commit introduced.** `is_body_kind` returned
   true for any kind containing `"block"`. tree-sitter-ruby declares
   `block_parameter` inside `method_parameters`, so for `def each(&blk)` the
   earliest "body" after the name was the `&blk` **parameter**: the stored code
   became `def each(` — a mid-signature fragment, on a language with no test and
   no corpus, where the name-line slice had been right. Same predicate hit
   Objective-C's `block_pointer_declarator` (the block-typed parameter, the
   dominant ObjC callback idiom) and `block_comment` between the name and the
   body in rust/java/kotlin/scala/dart/groovy/julia/fsharp. `*parameter*`,
   `*argument*`, `*declarator*` and `*comment*` kinds are now rejected before
   the shape test.

2. **`end_line` was the enclosing scope, not the symbol,** wherever the scope
   node is not body/block/declaration_list-shaped. Every missing kind was read
   off a real ancestor chain rather than guessed: Objective-C `class_interface`
   / `implementation_definition`, C/C++/ObjC `preproc_if*`/`preproc_el*`, JS/TS
   `object`, MATLAB `methods`/`properties`/`events`, Fortran
   `internal_procedures`, Nix `binding_set`, PowerShell `statement_list`, and —
   language-scoped, because C# uses the same name for a type's own declaration —
   Zig's `struct_declaration` family. A generic net covers grammars still
   unlisted: a parent in which something BEFORE us already holds an
   implementation is a scope whatever it is called. Two shapes needed more than
   a list: a scope that is ALSO the declaration (`@interface Foo : NSObject`
   holds its methods as direct children) is climbed into and cut at its first
   member, and a multi-row scope inside a declaration (`const S = struct {`)
   ends the slice at ITS first member rather than at its own first byte.

3. **The headline C ratio was inflated by defect 2.** `decl_bench` computes the
   `span` baseline from this change's own `end_line`, so wherever the climb
   overshot, the "coarse alternative" being compared against was the enclosing
   `#ifdef` region, not the function. Re-measured, same corpus, same harness:

   | C corpus `span` | first commit | after the fix |
   |---|---|---|
   | median | 476 B | **123 B** |
   | mean | 9 836 B | **454 B** |

   So "9.5× and 162× smaller than the span" was wrong; the honest figure is
   **2.5× at the median and 7.8× at the mean**. Lucene and tantivy spans are
   unchanged (their scope nodes were already recognised).

4. **Two claims that shipped in the CHANGELOG were not supported**, and are
   corrected there. *"No symbol document can cost more than it did before"* is
   true of the 400-char cap, not of a document — a 45-char name line can now
   store a 400-char signature, which the same paragraph's honest "+10 bytes on
   average" already implied. *"`code` is the DECLARATION and never its
   implementation"* is false by construction wherever the grammar marks no body.
   `body_start` now also honours a `body:` **field** (r, nix, ocaml, fsharp —
   descending through a curried lambda's body so `f = a: b:` is not cut to
   `f = a:`) and the terminator of an `end`-delimited definition (fortran,
   julia). tree-sitter-haskell offers none of the three, so a Haskell equation's
   slice is the equation — and `decl_bench` now **measures** that rather than
   asserting the opposite: a slice covering its symbol's whole span is 1.5 % of
   Lucene, 2.5 % of valkey + memcached, 1.2 % of tantivy and 5.9 % of the Go + C
   tree.

### Third commit: three corpora were not enough

Widening the measurement to a 914 561-symbol Go + C tree found the same two
classes again, in Go, where neither Java, C nor Rust could show them:

- Go writes `const X = 1` and `const ( X = 1\n Y = 2 )` with the **same node
  kind**, so no list of kind names can separate the declaration from the scope —
  the parenthesis does. Without that, every generated `.pb.go` enum table
  reported the whole `var ( … )` block as one symbol's slice and span: a span
  mean of 6 792 B against a 102 B median, the identical signature that gave the
  C corpus its 9 836 B mean.
- Go's `composite_literal` marks its `{ … }` with the `body:` **field**, and the
  field-based body search from the second commit stopped there — storing
  `A_name = map[int32]string` and dropping the table the var exists to declare.
  A `body:` child now only ends the slice when its parent IS a definition; a
  value that merely carries the field does not.

Over that corpus: mid-signature slices 87 267 (9.5 %) → **1 678 (0.2 %)**, span
mean 6 792 → **387 B**, whole-span slices 8.3 % → 5.9 %.

Non-blocking items from the same review, also fixed: neither body search
descends into an argument or parameter list any more, since a lambda passed as
an argument is not the declaration's implementation (this alone takes Lucene's
residual from 27 to 18 and the C corpus's from 490 to 54); the raw slice is
size-checked before `dedent` copies it, so the copy cannot be O(symbols × file);
`dedent` computes the shared indent in CHARS, so a block mixing ASCII
indentation with U+00A0 can no longer blank a line out; and `-module(m).` is no
longer absorbed as an "attribute" of the Erlang function below it.

## Measured

All numbers from the `decl_bench` harness this change adds, which computes the
pre-fix slice (verbatim: `lines[name_row]`, trimmed and capped) and the
declaration slice **side by side in one pass** over the same corpus, the same
parse and the same symbol set — so before/after is not two different runs.

```
XERJ_DECL_BENCH=<repo> [XERJ_DECL_EXT=java] cargo test --release -p xerj-autoindex \
  --lib decl_bench -- --ignored --nocapture
```

| corpus | unit | median B | mean B | ends mid-signature |
|---|---|---|---|---|
| **Lucene** — 5 627 Java files (49.8 MB), 79 102 declarations | was — the name's line | 49 | 50.5 | 4 137 (5.2 %) |
| | now — the declaration | 52 | 60.7 | **18 (0.02 %)** |
| | the symbol's whole `line..end_line` span | 229 | 1 046.2 | — |
| **valkey + memcached** — 981 C files (15.0 MB), 28 347 | was | 48 | 51.6 | 1 444 (5.1 %) |
| | now | 49 | 58.8 | **54 (0.2 %)** |
| | span | 123 | 454.3 | — |
| **tantivy** — 434 Rust files (5.4 MB), 8 970 | was | 41 | 44.2 | 851 (9.5 %) |
| | now | 50 | 60.9 | **27 (0.3 %)** |
| | span | 218 | 720.5 | — |
| **cilium + dpdk + vpp-agent + xdp-tools** — 26 357 Go/C files (293.8 MB), 914 561 | was | 49 | 54.6 | 87 267 (9.5 %) |
| | now | 53 | 67.4 | **1 678 (0.2 %)** |
| | span | 102 | 387.1 | — |

The fourth corpus was added after the review: three corpora (Java, C, Rust) are
not enough to bound a change that touches every registered grammar, and it is
what found the Go defects in the third commit.

Read honestly: the unit costs 7–17 bytes more on average, and in exchange it is
a complete declaration rather than a fragment. It stays below the coarse
alternative — 1.9–4.4× smaller than that symbol's own span at the median, 5.7–17×
at the mean. Also on Lucene: 31 336 (39.6 %) of declarations wrap past one line,
40 751 (51.5 %) have a name line that does not stop where the declaration does,
and 1 179 (1.5 %) have a slice covering their whole span.

Two things the residual counts are not: `unbalanced()` scores a `)` inside a
string literal (`public static final String RPAREN = ")";`) or inside a comment
(`// hopefully ~24 days is long enough ;)`) as a truncation, and 5 of Lucene's
remaining 18 are exactly that. `XERJ_DECL_DUMP=1` prints each one with its file
and line, which is how they were identified rather than guessed at.

**Throughput is not unchanged**, and the earlier claim that it was came from a
loaded box. Medians of **five interleaved runs**, back to back, same 6 106-file
Lucene tree, `extract_bench` (not `decl_bench` — the earlier text attributed it
to the wrong harness):

| | files/s (median of 5) |
|---|---|
| `main` before #500 | **1 600** |
| after this PR | **1 455** |

The two distributions do not overlap (main 1 527–1 601, branch 1 435–1 466).
About **9 %** of extraction time, for the ancestor climb and the body search.
The CHANGELOG says 9 %. Three-run blocks taken minutes apart were *not* enough
to see this — one such pair put `main` below the fixed build — which is why the
final figure is interleaved medians.

What is NOT measured here: the issue's original 12-question refsym aggregate
(bytes and recall per answered question end to end). That needs the reference
corpus on :9200, which this worktree must not touch. The numbers above are the
extractor-level cause of that aggregate, not the aggregate itself.

## Test proof (fails on unfixed code)

The first commit's two tests were run against the extractor with **only the
slice computation reverted** to the pre-fix expression; both fail with
`"public void configure("` and `"pub fn scan("`.

The second commit's three tests were run the same way, against the **first
commit's** extractor with only these tests spliced in:

```
---- a_body_shaped_name_inside_the_signature_does_not_end_the_declaration ----
  left: "def each("                                   right: "def each(&blk)"
---- end_line_and_slice_are_the_symbol_not_its_enclosing_scope ----
  left: "#ifdef HAVE_X\nstatic void a(void)"           right: "static void a(void)"
        (and end_line 10 — the `#endif` — for a function that ends on line 5)
---- grammars_that_name_no_body_node_still_stop_before_the_implementation ----
  left: "myfun <- function(a, b) {\n  a + b\n}"        right: "myfun <- function(a, b)"

test result: FAILED. 71 passed; 3 failed
```

With the fix: `test result: ok. 750 passed; 0 failed; 2 ignored` — the whole
`xerj-autoindex` lib suite, which is where this crate's tests live (including
`incremental_reconcile_http_tests` and `failure_resume_http_tests`, the two that
encode the code-document model).

## Not in this PR

- **The doc comment as its own searchable field.** #500 also asks for the
  declaration's doc comment. Adding a `doc` field to the symbol document runs
  into the `ensure_compatible` constraint above, so it needs a schema-evolution
  fix first (let an extractor-origin field be absorbed into a frozen dataset
  rather than abort). Left on the issue with that reasoning. Doc comments remain
  searchable today through the file document's `body`.
- **The refsym 12-question aggregate.** Re-running it needs the reference server
  on :9200, which this worktree must not touch.
- **Known residual, recorded rather than hidden.** A Haskell equation's slice is
  the equation (no body node, no `body:` field, no terminator). `line` still
  points at the NAME's row while `code` may begin one or two lines above it, so
  `xerj search`'s `path:line` header is no longer the first line printed —
  cosmetic, worth a follow-up. An index upgraded to this release keeps the old
  single-line `code` for every file that has not changed, until a rebuild.

Refs #500 — the acceptance measurement the issue asks for (the 12-question
refsym aggregate) has not been re-run, so this does not close it.

## Checks

- [x] `cargo fmt --all` clean; `cargo clippy --release -j 8 -p xerj-autoindex --all-targets -- -D warnings` clean
- [x] Scoped release build: `cargo build --release -j 8 -p xerj-autoindex`
- [x] `cargo test --release -p xerj-autoindex --lib` — 750 passed, 0 failed
- [ ] ES-YAML conformance suite — **not run**: no engine or wire-protocol code is touched; the change is confined to `xerj-autoindex`'s tree-sitter extractor.
- [x] No new ES-compatible behaviour, so no YAML case applies
- [x] Docs updated: CHANGELOG entry rewritten with the re-measured numbers and the two corrected claims

**Not run:** the ES-YAML suite (reason above), and the issue's 12-question
refsym aggregate (needs the reference corpus on :9200, which this worktree must
not touch).

## Provenance

- **Written by:** AI agent (Claude Code), run by @xerj-org, who has reviewed it.
- **Verified** (commands run, output observed): `cargo fmt --all --check`;
  `cargo clippy --release -j 8 -p xerj-autoindex --all-targets -- -D warnings`;
  `cargo build --release -j 8 -p xerj-autoindex`; `cargo test --release -p
  xerj-autoindex --lib` (750 passed); `decl_bench` over Lucene,
  valkey + memcached and tantivy on both the previous commit and this one (the
  C `span` 476/9 836 → 123/454 comparison is two runs of the same command on the
  two builds); `decl_bench` over a fourth, 914 561-symbol Go + C corpus, which is
  what found the third commit's two defects; `extract_bench` five interleaved
  runs each on `main` and this branch; the three new tests spliced into the
  previous commit's extractor and observed to fail; the ancestor-chain dumps
  that produced every kind name in `is_container_kind`.
- **Assumed** (not tested, and what would falsify it): that
  `reconcile_plan::ensure_compatible` treats the array-borne `end_line` as exempt
  — read from `infer/mod.rs:205` and `reconcile_plan.rs:436`, not exercised by a
  cross-version re-index; a run of an older binary's index through this build
  aborting with `field end_line is absent from frozen dataset` would falsify it.
  That an index upgraded in place keeps old `code` for unchanged files is read
  from the byte-identity fast path at `reconcile_plan.rs:110-121`, not observed
  end to end. The grammar coverage is 30+ languages checked by hand and three
  measured corpora (Java, C, Rust); a grammar outside those whose scope node is
  still unlisted would show up as a slice that starts at the enclosing scope.


xerj-org:
## Response to the block review — all four blocking findings confirmed and fixed (`720f312d`)

I reproduced every finding before touching anything, and none of them were
refutable. The Ruby case in particular was exactly as described: on this PR's
first commit, `def each(&blk)` stored `"def each("`, where `main` stored the
correct line — a strict regression manufactured by the very heuristic the PR
adds. Thank you; that one would have shipped.

### 1 — `is_body_kind` matched signature nodes

Confirmed by running the extractor, not by reading it:

| input | first commit | now |
|---|---|---|
| ruby `def each(&blk)` | `"def each("` | `"def each(&blk)"` |
| rust `pub fn scan(input: &str) /* n */ -> usize` | `"pub fn scan("` | full signature |
| objc `- (void)run:(void (^)(int))cb` | `"@implementation Foo\n- (void)run:(void ("` | `"- (void)run:(void (^)(int))cb"` |

`*parameter*`, `*argument*`, `*declarator*` and `*comment*` kinds are now
rejected before the shape test, with the three offending spellings named in the
doc comment so the next person does not re-add them.

### 2 — the climb ran past the declaration into the scope

Confirmed on every grammar you listed, plus two you did not. Each container kind
I added was read off a printed ancestor chain rather than guessed:

- Objective-C `class_interface` and `implementation_definition` (note:
  `class_implementation` is *not* a container — it is the class's own
  declaration node, and listing it sliced `@implementation Foo` down to `"Foo"`,
  which I caught only because I re-probed after the first attempt)
- C/C++/ObjC `preproc_if*` / `preproc_el*`
- JS/TS `object`, MATLAB `methods`/`properties`/`events`, Fortran
  `internal_procedures`, Nix `binding_set`
- **PowerShell `statement_list`** — not in your list; `One` in a two-function
  script had `end_line` = the last line of the file
- Zig's `struct_declaration` family, **language-scoped**: C# uses that exact
  name for a type's own declaration, and making it a container globally cost
  `public struct Point` its modifiers.

Two of your suggestions I checked and did not need. Dart `class_member` wraps
exactly one member, so `int one()` / `end_line` 4 were already right; adding it
changes nothing. And a list alone is not enough for two shapes, so there is also
a generic net (a parent in which something *before* us already holds an
implementation is a scope, whatever it is called), plus explicit handling for a
scope that is *also* the declaration (`@interface Foo : NSObject`) and for a
multi-row scope inside one (`const S = struct {`).

### 3 — the metric was inflated by finding 2. Re-measured, not restated.

Same corpus, same harness, two builds:

| valkey + memcached | first commit | after the fix |
|---|---|---|
| `span` median | 476 B | **123 B** |
| `span` mean | 9 836 B | **454 B** |
| slice ends mid-signature | 490 (1.7 %) | **57 (0.2 %)** |

So the mean was inflated 21.6× by this PR's own defect, and "9.5× and 162×" is
withdrawn. The honest C figure is **2.5× at the median, 7.8× at the mean**.
Lucene (229 / 1 046) and tantivy (218 / 721) are unchanged — their scope nodes
were already recognised, which is exactly why a mean 20× the median stood out on
C alone. The CHANGELOG and PR body now carry the re-measured numbers.

Lucene's residual also fell 27 → 18 and C's 490 → 57, mostly from your
non-blocking observation about cost: neither body search descends into an
argument or parameter list any more, because the `block` in
`Comparator.comparing(d -> {…})` belongs to a lambda that is an *argument*, not
to the declaration. Of the remaining 18 on Lucene, 5 are `unbalanced()` false
positives — a `)` inside `String RPAREN = ")"` or inside `// long enough ;)`.
`XERJ_DECL_DUMP=1` prints them; that is how I know rather than assume.

### 4 — both unsupported claims

(a) "no symbol document can cost more than it did before" is gone. The CHANGELOG
now says the *worst case per symbol* is unchanged at 400 chars and that an
individual document can grow, since a 45-char name line may now store a 400-char
signature.

(b) "never the body" is gone from the CHANGELOG, the PR body and the `Symbol`
doc comment. `body_start` now also honours a `body:` **field** — which fixes r,
nix, ocaml and fsharp outright (`myfun <- function(a, b)`, `let add a b =`), and
descends through a curried lambda's body so Nix's `f = a: b: a + b` yields
`f = a: b:` rather than `f = a:` — and the terminator of an `end`-delimited
definition, which fixes fortran and julia (`function add(a, b)`). Haskell has
none of the three, so its slice is the equation, and that is stated as a limit
in all three places instead of being claimed away. And `decl_bench` now
**measures** it: a slice covering its symbol's whole span is 1.5 % of Lucene,
2.3 % of valkey + memcached, 0.8 % of tantivy. You were right that
`unbalanced()` structurally cannot see this; the new column is a span
comparison, not a bracket count.

### Non-blocking

- **#1 slice cost** — the raw slice is now size-checked *before* `dedent` copies
  it (`SLICE_MAX_BYTES`, 32× the cap), so the copy cannot be O(symbols × file).
- **#2 UTF-8** — you were right that it is a blanked line, not a panic.
  `dedent`'s shared indent is now a CHAR count, so the doc comment's "always on
  a char boundary" is true rather than nearly true.
- **#4 existing indexes** — added to the CHANGELOG as a release note, after
  verifying the byte-identity fast path at `reconcile_plan.rs:110-121` rather
  than taking it on trust.
- **#5** — agreed, `oversized_declaration_degrades_to_the_start_line` is a
  bound guard, not fix coverage; not counted as proof anywhere.
- **#6** — now `Refs #500`, with the reason stated in the body.
- **#8** — the `🤖 Generated with` marker is replaced by the Provenance block the
  template requires (Written by / Verified / Assumed), with the `ensure_compatible`
  reasoning listed under **Assumed** and what would falsify it.
- **#9** — the CHANGELOG command now carries `--release --lib`, and the
  throughput number is attributed to `extract_bench`, not `decl_bench`.
- **#7** — `line` vs the first line of `code` is real and left as a follow-up,
  recorded in "Not in this PR" rather than dropped.

**One thing I corrected that you were too kind about.** Throughput is *not*
unchanged. Three runs each, one box, same 6 106-file Lucene tree: `main` before
#500 at 1 607 / 1 561 / 1 573 files/s, this PR's first commit at 1 523 / 1 511 /
1 511, and after this fix 1 479 / 1 467 / 1 481 — about **7 %** of extraction
time. Your "n=1 before, caveated honestly but thin" was generous; the original
reading was taken on a loaded box and the real cost was inside the noise it
reported. The CHANGELOG now states the 7 %.

### Test proof

Three new tests, run against the **first commit's** extractor with only the
tests spliced in — all three fail there:

```
---- a_body_shaped_name_inside_the_signature_does_not_end_the_declaration ----
  left: "def each("                                right: "def each(&blk)"
---- end_line_and_slice_are_the_symbol_not_its_enclosing_scope ----
  left: "#ifdef HAVE_X\nstatic void a(void)"        right: "static void a(void)"
        (end_line 10 — the `#endif` — for a function that ends on line 5)
---- grammars_that_name_no_body_node_still_stop_before_the_implementation ----
  left: "myfun <- function(a, b) {\n  a + b\n}"     right: "myfun <- function(a, b)"

test result: FAILED. 71 passed; 3 failed
```

They pin your two demanded cases (Ruby, and a grammar whose scope the container
list missed — C `#ifdef`) plus JS object literals, Objective-C `@interface` and
`@implementation`, Zig vs C# `struct_declaration`, MATLAB, Fortran, PowerShell,
and the six no-body-node grammars including Haskell's asserted limit. With the
fix: **750 passed, 0 failed**. `cargo fmt --all --check` and
`clippy --release -p xerj-autoindex --all-targets -- -D warnings` are clean, and
the branch is merged up to `main` at rc.72 + #899.

### Residual risk

The container list is empirical over 30+ grammars checked by hand and three
measured corpora (Java, C, Rust). A registered grammar outside those whose scope
node is still unlisted would show the same symptom class — the generic
previous-sibling net catches the second and later members in such a scope, but
not the first. That is the honest boundary of this fix.

xerj-org:
## Follow-up (`5b4387c5`): your finding 2 had a third instance, in Go

After posting the fix above I took the reviewer's own point about coverage
seriously — "four grammar shapes are pinned out of 37 registered languages, and
the three corpora are Java, C, Rust" — and ran `decl_bench` over a fourth
corpus: cilium + dpdk + vpp-agent + xdp-tools, **26 357 files, 914 561 symbols**,
mostly Go and C. It found finding 2 again, in a shape no list of node kinds can
express, plus a defect I had introduced with the `body:` field fallback.

**Go writes `const X = 1` and `const ( X = 1\n Y = 2 )` with the same node
kind.** So the discriminator between a declaration and a scope there is the
parenthesis, not the name. (`var` interposes a `var_spec_list`; `const` and
`type` interpose nothing.) Without it, every generated `.pb.go` enum table
reported the whole `var ( … )` block as one symbol's slice and span — a span
mean of **6 792 B against a 102 B median**, which is the identical signature you
spotted on the C corpus at 9 836 / 476. Same defect class, different grammar,
and it would have shipped.

**And Go's `composite_literal` marks its `{ … }` with the `body:` field** — so
the field-based body search I added in the previous commit was stopping there
and storing `A_name = map[int32]string`, dropping the table the var exists to
declare. A `body:` child now only ends the slice when its parent *is* a
definition (`function_definition`, `let_binding`, `function_or_value_defn`,
`function_expression`, `arrow_function`); a value that merely carries the field
— Go's composite literal, Java's `lambda_expression` — does not.

Over that corpus: mid-signature slices **87 267 (9.5 %) → 1 678 (0.2 %)**, span
mean **6 792 → 387 B**, span median 148 → 102 B, whole-span slices 8.3 % → 5.9 %.
The three original corpora move by a handful of symbols (Lucene 18 truncations,
valkey + memcached 54, tantivy 27). The corpus is now a fourth row in the
CHANGELOG table, because three corpora were demonstrably not enough to bound a
change that touches every registered grammar.

Regression test added to `end_line_and_slice_are_the_symbol_not_its_enclosing_scope`:
the grouped `const ( One = 1 / Two = 2 )` (→ `Two = 2`, `end_line` 3), the lone
`const One = 1` that must keep its keyword, and the composite literal whose
value must survive.

### And I have to correct my own throughput number

I claimed 7 % in the comment above. That came from three-run blocks taken
minutes apart on a box whose load was drifting, and a second pair put `main`
*below* the fixed build — which is not credible, so the method was wrong, not
just the number. Medians of **five interleaved runs, back to back**, same
6 106-file Lucene tree:

| | files/s (median of 5) | range |
|---|---|---|
| `main` before #500 | **1 600** | 1 527–1 601 |
| this branch | **1 455** | 1 435–1 466 |

Non-overlapping distributions, so the regression is real and it is **9 %**, not
7 %. The CHANGELOG says 9 % and now states that the figure is medians of
interleaved runs rather than three consecutive ones.

CI was green on `720f312d` (all 20 checks, including Build + Test, ES-compat
YAML, the three-OS autoindex FD smoke and the fat-LTO release build); this
commit re-runs it.

PR #901, opened 2026-08-31
perf(merge): merge posting lists instead of re-analysing every document (#876)

### LINKED ISSUE #876: Merge re-analyzes every document instead of merging postings — post-index churn ran 40+ min at ~110% CPU on a 154 MB corpus

xerj-org:
## What was wrong

A merged segment's postings **are** its inputs' postings with the document ids
remapped. Nothing about them depends on the source text:
`FtsIndexWriter::write_field_static` has always taken a `FieldData` whose
`postings` is a `PostingsWriter`, and `PostingsWriter`'s ingest unit has always
been a `(term, doc, position)` occurrence — never a document. No analyzer is
needed anywhere in the segment writer.

`merge_pass_locked` (`engine/crates/xerj-engine/src/index.rs:10450`) did not use
that. It walked back to every surviving document's stored `_source`, rebuilt
`fts_input: Vec<(String, HashMap<String, FieldValues>, Value)>` through
`extract_fts_fields_excluding`, and re-ran the whole analyzer chain via
`add_documents_parallel` — recomputing postings that were already on disk and
already correct, while holding a second copy of every indexed field's *text*.
Under size-tiered levelling each document is merged O(log N) times, so total
merge work approaches (analysis cost of the corpus) × (number of levels).

This is **Step 1 only** of the design on the issue. Step 2 (byte-copying posting
runs through an `OpBuilder::union`, Lucene's `FieldsConsumer.merge` bulk path)
is deliberately out of scope and remains open.

## What changed

**`FtsIndexWriter::merge_from_segments`** (new, `xerj-fts`) builds each merged
field by streaming each input's FST, decoding its posting lists with
`PostingsReader`, remapping doc ids through a caller-supplied `doc_map`
(tombstoned and superseded documents map to `None`), and handing each
`(source, term)` run to the writer in one call. `write_field_static` is reused
verbatim. **No on-disk format change, no change to the flush path's output.**

Three things had to be got exactly right:

- **Norms.** `norm_u16_to_u8` is lossy and *not* invertible — byte 9
  dequantises to length 8, which re-quantises to byte 8 — so reading lengths
  back out of a source segment and re-quantising them would move BM25 length
  normalisation on every merged document. `FieldData.norms` now holds the
  quantised byte (quantisation moved into `index_field_values`), the reader
  keeps each norm's raw byte beside its dequantised length (still 8 bytes per
  entry after padding, so no memory change), and the merge carries the byte
  through untouched.
- **Document order.** The engine sorts each batch's survivors by `_seq_no`, so
  the inputs' documents *interleave* in the merged ordinal space and a shared
  term's list arrives as several ascending runs.
  `PostingsWriter::sort_postings_by_doc` restores the invariant that
  `encode_block_doc_ids`' plain `-` depends on; already-ordered lists are
  detected and never sorted.
- **Field statistics.** `total_docs` / `total_field_length` are carried from the
  input `.meta` and reduced by the dropped documents' contribution — exact token
  counts for a positioned field, the quantised norm byte for a docs-only one —
  so `avgdl` does not move and neither do scores.

**`fts_merge_readers`** (new, `xerj-engine`) is the preflight. It enumerates
every input's side-car field names in **one** directory scan (a converging index
keeps tens of thousands of files there and this runs per batch) and refuses the
fast path — falling back to the old rebuild, same output, old cost — whenever
the replay cannot be proved equivalent:

- a digest-encoded filename component that no mapped field name resolves to
  (silently omitting the field would drop its postings and stop its terms
  matching);
- an input whose side-cars will not open;
- an input carrying documents but with no FTS data at all;
- a field whose stored position setting contradicts what this merge would write.

`Index::merge_reanalysed_document_count()` reports how many documents took the
fallback. `XERJ_PROF=1` prints a per-batch `XERJ_PROF merge-batch` phase
breakdown alongside the existing `flush-sidecar` line.
`XERJ_MERGE_FTS_REANALYZE=1` forces every merge back onto the old path (env var,
matching `XERJ_MERGE_PARALLELISM`'s precedent — no new config setting).

## Measurement

100 MiB text corpus, 2 000 documents, force-merged to one segment. Median of
four runs on this 32-core box, each arm in its own process, read off the new
`XERJ_PROF merge-batch` line
(`cargo test --release -p xerj-engine --test merge_replays_postings -- --ignored
--nocapture measure_forcemerge_cost`):

| | re-analysing merge | postings merge |
|---|---|---|
| merged segment's FTS side-car build | 3.15 s | **0.85 s (~3.7×)** |
| force-merge wall clock (noisy) | ~41 s | ~32 s |
| peak RSS over the force-merge | ~2.8 GiB | ~2.4 GiB |

The wall-clock figure is noisy (batch shapes differ run to run) and moves less
than the FTS figure for a reason worth recording: **after this change the merge
is dominated by `build_doc_value_columns`** — 7–12 s per batch against 0.85 s
for FTS on the same corpus. That is the next lever, and this PR does not touch
it. Commented on the issue.

I did **not** reproduce the issue's original 40-minute autoindex tail; the
numbers above are from a synthetic 100 MiB corpus in this sandbox, not from
that run.

## Tests

`engine/crates/xerj-fts/tests/merge_postings_equivalence.rs` builds the same
merged segment **twice** — once by re-analysing the documents, once by replaying
the postings — and compares:

- the `.fst` / `.post` / `.meta` files **byte for byte**, with and without
  dropped documents (300 documents over 3 inputs, interleaved by a synthetic
  `_seq_no` order, terms crossing the 128-doc PFOR block boundary, multi-valued
  text spanning the position-increment gap, a docs-only keyword field, a
  single-token keyword field, and a sparse field);
- `.norms` for everything a reader can load out of it — every norm byte, every
  `field_length`, the field statistics — plus the bound that the replayed dense
  array may be shorter than the re-analysed one but never longer (see the
  second known divergence below);
- hit lists and exact `_score` bit patterns for term, phrase, keyword and
  match-all queries;
- per-document field lengths and per-field `total_docs` / `total_field_length`.

`engine/crates/xerj-engine/tests/merge_replays_postings.rs` does it end to end:
400 documents across 8 flushes with deletes and overwrites, force-merged to one
segment, comparing every hit and score bit-for-bit against the same corpus
merged with `XERJ_MERGE_FTS_REANALYZE=1`, and asserting
`merge_reanalysed_document_count() == 0` so a preflight that silently always
fell back could not pass.

**On "a test proven to fail on unfixed code":** these tests drive an API that
does not exist before this change, so stashing the fix yields a compile error
rather than a meaningful failure. I proved their power by mutation instead —
five deliberate breakages of the new code, each caught:

| mutation | result |
|---|---|
| remove `sort_postings_by_doc` | 2 tests fail |
| re-quantise the norm through `norm_u8_to_u16` | 3 tests fail |
| stop subtracting dropped documents from the stats | 2 tests fail |
| keep the source doc id instead of remapping it | 2 tests fail |
| drop positions from the replayed postings | 2 tests fail |

## Known divergences, each pinned by its own test

Everything a query can read out of a merged segment is identical to what the old
path wrote — every posting list, `doc_freq`, norm byte, `_score` and
`field_length` — and `.fst` and `.post` are byte-for-byte identical. Two on-disk
representations differ without any query being able to tell.

**1. The `.norms` dense array may be SHORTER.** `norm_u16_to_u8` returns byte 0
for length 0 AND for length 1, and `load_norms` drops byte-0 entries on read. A
re-analysing writer pushes an entry for every document that *carries* the field,
byte 0 included, and `write_field_static` sizes the array from the last entry; a
replay has no byte-0 entry to carry forward, so its array stops at the last
NON-zero document. For a field that is single-token everywhere — every
single-token `keyword` field — that means the array collapses to its header. The
difference is trailing zeros: both files load to the identical (empty) norms
table, so no hit, `_score` or `field_length` moves.
`a_single_token_field_replays_an_equivalent_shorter_norms_array` pins it
numerically (re-analysed `dense_len` 300, replayed `dense_len` 1) and asserts the
fixture's `colour` field really is byte-0 throughout, so the case cannot silently
stop being covered.

**2. `.meta`'s `total_term_frequency` for a docs-only (`keyword`) field.** A
document that repeats the same value merges with
`total_term_frequency = doc_frequency`: that format never stored a per-document
frequency and its reader synthesises `1`, so this is the value every reader of
the source segment already sees. Nothing outside `xerj-fts` reads
`total_term_frequency`, and no hit, `_score`, `doc_freq` or field-length
statistic changes — `total_field_length` is carried from the source `.meta`, so
the repeat still counts toward `avgdl`. `.meta` is byte-for-byte identical
otherwise, which is why the equivalence tests compare it byte for byte (no
fixture document repeats a keyword value).

Two further edges exist **only** when a merge also drops documents: a dropped
document whose value analysed to zero tokens leaves no trace in either the
postings or the norms, so its `total_docs` seat is not reclaimed; and a dropped
docs-only document's length is recovered from the quantised norm byte, exact to
length 7. Both are documented on `merge_from_segments`.

## One drive-by, in the same function

The survivor loop used to `continue` out of `fts_input` when a document's stored
bytes failed to re-parse as a `Value`. Those bytes already parsed as a
`RawValue` a few lines earlier, so reaching it is close to impossible — but it
would shift every *later* document's FTS and doc-values ordinal one place off
its stored ordinal, and that alignment is what every reader addresses a document
by (and, since this PR, what the merge's own `doc_map` is built on). The
document now keeps its seat with an empty `_source`, which indexes and
columnises nothing for it: the same outcome as the skip, minus the shift.

One behaviour change worth knowing: postings are now **preserved as written**
rather than re-derived. If the mapping changes such that a field would analyse
differently today, a merge no longer silently re-analyses old segments under the
new rules — which is Lucene's semantic, but it is a change.

## Gates

- `cargo fmt --all --check` clean.
- `cargo clippy --release -j 8 -p xerj-fts -p xerj-engine --all-targets --
  -D warnings` clean.
- `cargo test --release -p xerj-fts` — 98 tests, 0 failed.
- `cargo test --release -p xerj-engine` — full crate suite, 0 failed.
- The ES-YAML conformance suite was **not** run locally (its runner does
  `DELETE /_all`); CI covers it.

Closes #876.


xerj-org:
## CI: the one red check is #751, not this PR

Every check passes except **Build + Test**, which times out in step 6
(`Workspace tests (all targets) at default parallelism`) — the known hang
tracked as [#751](https://github.com/xerj-org/xerj/issues/751), which that issue
states is not PR-specific.

The evidence that it is not this change:

- The hang is always the same test, and it is not one this PR touches:
  `script_bucketed_agg_past_the_call_depth_limit_is_an_error_not_empty_buckets`
  in `crates/xerj-api/tests/script_limits_http.rs` (script call-depth limits).
  It logs `has been running for over 60 seconds` and then the step's 12-minute
  budget expires.
- **The identical failure happened on `main` today**, on a commit that predates
  this branch: run
  [33351936480](https://github.com/xerj-org/xerj/actions/runs/33351936480)
  (`6936d826`), same test, same 60s+ line, same 12-minute timeout.
- A branch already exists for it: `fix/751-ci-default-parallelism-hang`, whose
  run [33359300573](https://github.com/xerj-org/xerj/actions/runs/33359300573)
  failed the same way.
- The test passes locally on this branch in 0.04 s:
  `cargo test --release -p xerj-api --test script_limits_http` → 6 passed.

I re-ran the failed job once; it hung on the same test again. I have not tried
to work around it here — #751 is where it belongs.

What I did run locally on this branch, post-rebase onto `101d70be`:

- `cargo clippy --release -j 8 -p xerj-fts -p xerj-engine --all-targets -- -D warnings` — clean
- `cargo test --release -p xerj-fts` — 98 tests, 0 failed
- `cargo test --release -p xerj-engine` — full crate suite, 0 failed

`ES-compat YAML conformance` is green on this PR, which is the gate that matters
most for a change to segment side-cars.

xerj-org:
## Review response — the "byte-for-byte" claim was wrong, and the tests could not see it

Rebased onto `main` (rc.72, #899, #894, #900) and fixed the blocking finding. Pushed as `26fa34ad`.

### The finding was correct

`norm_u16_to_u8` returns byte 0 for length 0 **and** for length 1, `index_field_values` pushes a norm entry for every document that *carries* the field (byte 0 included), and `write_field_static` sizes the dense array from `norms.last()`. `load_norms` filters `b != 0` on read, so a byte-0 entry is not there for the replay to carry forward. Whenever the highest-ordinal document carrying a field quantises to 0 — every single-token `keyword` field — the replayed `.norms` has a smaller `dense_len` header and a shorter payload.

I reproduced it exactly as described, and the number is now asserted: over the 300-document fixture, a single-token `colour` field gives a re-analysed `dense_len` of **300** and a replayed `dense_len` of **1**.

It is **not** a runtime defect. The difference is trailing zeros; both files load to an identical (empty) norms table, so `field_length`, `doc_freq`, `_score` and the hit list are unchanged — which the new test also asserts directly.

### What changed

**The claim.** CHANGELOG and PR body now say what is true: everything a query can read is identical, `.fst`/`.post`/`.meta` are byte-for-byte, and there are **two** on-disk representations that differ invisibly — the shorter `.norms` array and the already-documented docs-only `total_term_frequency`. The false sentence also existed in the code: `merge_from_segments`' doc comment and the in-line comment at the norms replay claimed that skipping byte 0 "reproduces the array a re-analysing writer would have built". It reproduces its *content*, not its *length*. Both now say so, and point at the test.

**The fixture.** `merge_postings_equivalence.rs` gains a single-token `colour` keyword field on all 300 documents, so the case is exercised on every run instead of being accidentally avoided. `.fst`/`.post`/`.meta` are still compared **byte for byte** across all four fields, with and without dropped documents — and they pass, which is a stronger result than I expected: it proves the `total_docs` / `total_field_length` arithmetic is exact for a single-token docs-only field even when a fifth of the documents are dropped (the byte-0 recovery path `norm_u8_to_u16(0) == 1` is exactly right there). `.norms` is now compared by `assert_norms_equivalent`: every loaded norm byte, every `field_length` and the field statistics must match exactly, and the replayed array may be shorter but **never longer** (a longer one would mean the replay invented a document).

**The pin.** New test `a_single_token_field_replays_an_equivalent_shorter_norms_array` states the mechanism directly, asserts the two `dense_len` values, and asserts a **precondition** — that `colour` really is byte-0 throughout — so a future fixture edit that gave the field two tokens fails loudly instead of silently un-covering the case again. That precondition is the part that was missing; the original suite did not fail, it simply stopped testing.

The two byte-identity tests are renamed `..._is_byte_identical_to_re_analysis` -> `..._reproduces_the_re_analysed_segment`, because that is now what they check.

### Also picked up from the review

- **Non-blocking 3** — the real behaviour change (postings preserved as written, so `_forcemerge` no longer re-analyses old segments under a changed analyzer; a position-setting change still falls back, an analyzer-only change is not detectable and needs a reindex) is now in the **CHANGELOG**, not only the PR body.
- **Non-blocking 9** — the stale `retained_bytes` comment (`Vec<(u32, u16)>` -> `Vec<(u32, u16, u8)>`; the arithmetic was already right).

### Not addressed, deliberately

Non-blocking 1 (count decoded postings against `doc_frequency` and fail closed), 2 (operator-facing message when a `merge_from_segments` `Err` aborts the batch), 5 (the `set_var` / live-`Engine` race in `merge_replays_postings.rs`), 6, 7 and the remaining coverage gaps in 8 (docs-only field > 8 tokens; end-to-end preflight REFUSAL) are all real. They change behaviour or add surface rather than correct a shipped claim, so I have left them out of this PR rather than widen it during a rebase. Happy to file them as a follow-up issue.

### CI

Rebased over #899, so the `script_bucketed_agg_past_the_call_depth_limit_is_an_error_not_empty_buckets` hang that made "Build + Test" red should now be gone — that was the point of #899, and it reproduced identically on `main` before it merged.

### Verified locally

`cargo test --release -p xerj-fts` — 91 unit + **8** merge-equivalence + 1 doc-test, all green. `merge_replays_postings` green. Sibling checks on the merged tree: #899's `agg_corpus_hydration_deadlock` green, #900's `xerj-storage wal::` 29 tests green. `cargo fmt --all --check` clean; `cargo clippy --release -p xerj-fts -p xerj-engine --all-targets -- -D warnings` clean.

Merge check: the merge result's diff against `origin/main` is identical in shape to this branch's own intended diff — same 7 files, same 1,630 lines before my fix — and every file #899 / #894 / #900 touched is byte-identical to `main`. CHANGELOG kept both sides.

xerj-org:
## Correction to my own correction, and the CI verdict

**Correction.** My previous comment (and the first version of the CHANGELOG wording it described) said `.fst`, `.post` **and `.meta`** are byte-for-byte identical. That is wrong, and wrong for a reason already in this PR: `total_term_frequency` lives in `.meta`, so the docs-only ttf divergence *is* a `.meta` difference. Fixing an overstated claim with another overstated claim would have been a poor outcome, so: `aba12bb0` narrows it to what holds.

The accurate statement, now in the CHANGELOG, the PR body and the `merge_from_segments` doc comment:

- `.fst` and `.post` — byte-for-byte identical, unconditionally.
- `.meta` — identical except `total_term_frequency` for a docs-only field whose document repeats a value. Everything else in it, `total_field_length` included, survives, so `avgdl` still counts the repeat.
- `.norms` — may be shorter (the finding this review raised); identical in everything a reader loads.

The equivalence tests still byte-compare `.meta` across all four fields, because no fixture document repeats a keyword value — the module comment now says that out loud instead of leaving it to be inferred, and the ttf case keeps its own dedicated test.

`3da8b642` is cosmetic only: two `assert_norms_equivalent` failure messages had their Rust line continuations eaten while the helper was being written, leaving a run of spaces mid-sentence in the panic text.

---

**CI verdict on `3da8b642`: green, all 20 checks.** The blocking CI finding is closed. `Build + Test` — the step that timed out at 12 minutes hanging on `script_bucketed_agg_past_the_call_depth_limit_is_an_error_not_empty_buckets` — now **passes in 31m36s**, which is what #899 was for; the review's read that the failure was `main`'s and not this diff's is confirmed. `ES-compat YAML conformance` 6m15s, `Format + Clippy` 1m33s, `Release build (fat LTO)` 7m39s, `Fuzz harnesses` 7m54s, all three `Autoindex FD smoke` platforms including Windows — all pass. PR state is `MERGEABLE` / `CLEAN`.

The rebase brought in rc.72, #899, #894 and #900. Conflict was CHANGELOG only (both sides kept). `engine/crates/xerj-engine/src/index.rs` auto-merged with #899's `stored_values_for_async` change, and I checked it rather than trusting it: the merge result's diff against `origin/main` is the same 7 files as this branch's own intended diff, and every file #899 / #894 / #900 touched is byte-identical to `main`. #899's `agg_corpus_hydration_deadlock` and #900's `xerj-storage wal::` suite both pass on the merged tree locally.

PR #906, opened 2026-08-31
ci: bound every workflow job so one hang cannot hold main's CI slot for 6h (#770)

### LINKED ISSUE #751: CI: Build+Test step 6 (workspace tests at default parallelism) intermittently hangs on hosted runners

### LINKED ISSUE #770: CI #751 follow-up: build-test job-timeout margin + other main jobs can still starve the slot

xerj-org:
Closes #770.

## What was left of #770

**Nit 1 — build-test job-cap margin: already done, nothing to do here.** #780
took it 50 → 60 minutes and is merged; `ci.yml` carries `timeout-minutes: 60` on
`build-test` today. That half of the issue is obsolete and is called out as such
in a comment on #770.

**Nit 2 — other jobs can still starve the main slot: real, and this PR closes
it.** 22 of this repository's 23 GitHub Actions jobs declared no
`timeout-minutes`, so they ran against GitHub's 360-minute default. Three
workflows serialise on a concurrency group that never auto-cancels —

| workflow | group | `cancel-in-progress` |
|---|---|---|
| `ci.yml` | `ci-CI-refs/heads/main` | `false` on main |
| `deploy-pages.yml` | `pages-deploy` | `false` |
| `release-metrics.yml` | `release-metrics` | `false` |

— so one hung job holds that slot for hours while every later run queues behind
it, and because GitHub keeps at most one *queued* run per group, each new push
discards the previously queued one.

The #770 follow-up comment asked for this to wait until "a second job is
actually observed hanging". This PR does not wait on that, because the harm is
not proportional to how many jobs hang — it is what the 360-minute default costs
when *any one* of them does, and this repo has already paid it twice.

## Measured, from this repo's own run history

Both instances are from 2026-08-25, before #767 capped `build-test`. Neither
number is restated from the issue; both come from the Actions API.

* **Run [32796557309](https://github.com/xerj-org/xerj/actions/runs/32796557309)
  (`33dfc6b2`)** — `Build + Test` ran `01:21:28 → 07:21:44`. That is
  **360 min 16 s**: the job was killed by GitHub's default limit, the one every
  other job still runs against today. Every other job in that same run finished
  within 18 minutes. Behind it, run 32809082450 was discarded having started no
  job, and run 32811055446 waited **2 h 22 min** for the slot.
* **Run [32842067572](https://github.com/xerj-org/xerj/actions/runs/32842067572)
  (`9c3e0bc9`)** — the same job hung `11:28:49 → 17:03:14` and was cancelled by
  hand. The **seven** pushes to main that landed behind it (`a42e1d9f`,
  `3c33397c`, `6290bff8`, `27b7bf4d`, `b2854ee9`, `92d586d0`, `0f6f865d`) were
  **all discarded with zero jobs run**; main CI only resumed at 17:02:10.
  Stated honestly: a normal ~37-minute run would have let at least four of them
  through — the gaps between those pushes were 48, 62, 38 and 150 minutes — so
  the hang cost roughly four main commits their CI verdict, not seven.

PR #899 fixed the engine deadlock that hung that particular job. It did not
shrink the 6-hour exposure of the other 22 jobs, which is what #770 was still
asking about.

## The fix

Every job now declares a `timeout-minutes`, sized in tiers from the measured
duration of **every job execution across the last ~245 CI runs** — 3 873 job
executions — plus 27 release runs:

| cap | tier | measured max |
|---|---|---|
| 15 min | installs no toolchain, compiles nothing | 2.3 min (`Landing constants guard`) |
| 30 min | installs a toolchain or third-party binary, compiles no workspace member | 3.8 min (`cargo-audit`) |
| 60 min | compiles workspace crates | 48.3 min (`Build + Test`) |
| 75 min | `autoindex-fd-smoke` (heaviest tail, see below) | 49.6 min (`windows-latest`) |
| 75 min | `release.yml`'s cross-compile matrix | 38.3 min (`aarch64-pc-windows-msvc`) |
| 20 min | `deploy-pages`, `release-metrics` | 0.7 min |

`build-test`'s 60 (#780) and its 12-minute step-6 cap (#767) are unchanged.

**Worst-case slot starvation: 360 min → 75 (CI on main), 20 (`pages-deploy`),
75 (a release tag).**

The caps are deliberately generous rather than tuned to each job's median. A cap
that false-fails a slow-but-healthy run is worse than the hang it guards
against — the lesson #780 already paid for. Two things worth knowing:

* Job timeouts start when the job **starts executing**, not when it is queued,
  so runner-slot contention cannot spend any of the budget. That contention is
  real and larger than it looks: over **2 835 job executions** the created →
  started wait is a 4-second median but a **417-second p90, 1 661-second p99 and
  2 202-second max**, with 27 % of jobs waiting over a minute. This is the
  orchestration hypothesis behind #770 ("a legitimate slow build looks like a
  hang") and it checks out — but it inflates a *run's* wall clock, never a job's
  timeout, so no cap added here can false-fail because the runners were busy.
* Sizing from too small a window is a real hazard, and this PR hit it. The
  first commit sized the caps from 63 CI runs, which put `Autoindex FD smoke
  (windows-latest)` at a 37.2-minute max; over ~245 runs its max is **49.6**
  (n=204, median 18.8, p95 37.6). At 60 that is 1.21× its worst observed run —
  thinner than `Build + Test`'s 1.24×, on a much heavier tail (2.6× median-to-max
  spread, versus 1.5× for `Build + Test`; the same matrix's ubuntu leg maxes at
  15.6 and macOS at 22.2). The second commit widens that matrix to 75, which is
  1.51×. It is the only cap the wider window moved; the two tier maxima it
  restated (0.9 → 2.3 and 3.3 → 3.8) are far inside their caps. Widening
  `release.yml`'s window the same way (10 → 27 runs) moved its slowest leg from
  33.5 to 38.3 min, which is why the comment there now says 75 is 1.96× the
  observed max rather than ">2×".

## Test proven to fail on unfixed code

`.github/scripts/workflow-timeout-guard.py` is a stdlib-only merge gate (same
rule the SEO gate follows: a gate that needs `pip install` can fail for reasons
unrelated to the tree), wired in as the `workflow-timeouts` job. Stashing only
`.github/workflows/` and re-running it on the unfixed tree:

```
EXIT=1
22 problem(s) across 7 workflow file(s). Every job needs `timeout-minutes:`
(1..90); a job without one can hold a non-cancelling concurrency slot for 6 hours.
::error::.github/workflows/ci.yml: job `cla-config` has no `timeout-minutes:` and
  so inherits GitHub's 360-minute default (see #770)
  ... one per unbounded job, across all 7 workflow files
```

```
restored: EXIT=0
ok: 24 job(s) across 7 workflow file(s) all declare timeout-minutes <= 90
```

It runs in a new `Workflow job-timeout guard` job — checkout plus two `python3`
invocations, **6 s on the runner** in this PR's own CI, which is the cheapest
job in the workflow. That is the cost of the fix staying true for jobs added
later; GitHub has no workflow-level default timeout, so per-job is the only way
to express this at all.

The guard ships with `--self-test` (6 fixtures, run as its own CI step) covering
the ways it could be wrong: missing a violation, inventing one, and — the case
that matters here — accepting a **step**-level `timeout-minutes` as if it
bounded the job, which `ci.yml`'s existing 12-minute step-6 cap would otherwise
have masked. It exempts reusable-workflow calls, where GitHub rejects
`timeout-minutes` outright. Cross-check: a real PyYAML parse of all 7 files
finds exactly the same 24 jobs and the same values as the guard's line parser.

## Gates

* `python3 .github/scripts/workflow-timeout-guard.py --self-test` — 6/6 pass.
* `python3 .github/scripts/workflow-timeout-guard.py` — clean on this tree,
  fails on the unfixed one (above).
* `python3 -m py_compile` on the new script — clean.
* `cargo fmt --all --check` in `engine/` — clean. **No Rust changed by this PR**,
  so no clippy run and no crate suite is relevant; nothing was rebuilt.
* `actionlint 1.7.7` — clean on all 7 workflow files. This matters because this
  PR's own CI only exercises `ci.yml`, `seo.yml`, the CLA workflow and the ONNX
  recipe: `release.yml`, `deploy-pages.yml` and `release-metrics.yml` are
  triggered by pushes to main, tags or cron, so a syntax error in them would
  otherwise surface only after merge. Verified the linter is actually checking
  job-level keys rather than passing vacuously: renaming one `timeout-minutes`
  to `timeout-minutez` in a scratch copy of `release.yml` produces
  `unexpected key "timeout-minutez" for "job" section`, exit 1.

## Explicitly not done

* **The single-slot serialisation of main CI is left alone.** Measuring it
  turned up something worth writing down: **38 of the last 40 cancelled main CI
  runs never started a single job** — they were queued in the `ci-…-main` group
  and discarded when the next push arrived. That is the documented behaviour of
  `cancel-in-progress: false` (at most one queued run per group), not a hang,
  and main's *tip* always ends up tested because the newest push is always the
  one holding the queue. Giving every main commit its own run would multiply
  runner load (~25 jobs per push across `ci` + `release`) against an
  account-wide concurrent-job cap and starve PR CI, which is where the value is.
  Reported on #770 rather than changed here.
* **No hang is fixed by this PR.** It bounds the blast radius of the next one.
  #899 remains what fixed #751's.


xerj-org:
CI verdict on this branch: **22/22 green**, one correctly skipped (`Release notes match the tree`, which only fires on a `release/*` head).

Every job finished well inside the cap this PR gives it, which is the point — the caps have to bound a hang without ever fencing a healthy run:

| job | this run | cap | headroom |
|---|---|---|---|
| Build + Test | 30m50s | 60 | 1.9× |
| Autoindex FD smoke (windows-latest) | 18m58s | 75 | 4.0× |
| Release build (fat LTO) | 16m36s | 60 | 3.6× |
| Fuzz harnesses | 7m58s | 60 | 7.5× |
| ONNX feature | 6m58s | 60 | 8.6× |
| ES-compat YAML conformance | 6m17s | 60 | 9.5× |
| Format + Clippy | 1m28s | 60 | 41× |
| Workflow job-timeout guard | **8s** | 15 | — |

One green run is not evidence that a cap is right — the sizing evidence is the
distribution in the PR body (3 873 job executions), not this run. What this run
does show is that the guard job works on a real runner and costs 8 seconds, and
that no cap here is anywhere near a healthy run's duration.

PR #913, opened 2026-08-31
fix(dates): pin a declared date_nanos to nanoseconds, add numeric_type, back-fill legacy indices (#889)

### LINKED ISSUE #889: date_nanos still guesses its epoch scale per value, and existing indices are not back-filled (follow-up to #790)
Split out of #790 (fixed by #888). #888 made a date field's epoch scale a property of the **mapping** rather than a per-value guess, and pinned a declared `date` to milliseconds. It deliberately left `date_nanos` on the legacy per-value guess (`DateScale::PerValue`): ≥4 fractional-second digits → nanoseconds, anything coarser → milliseconds. Two things remain.

## 1. A `date_nanos` field can still mix scales within one column

`date_nanos` keeps the per-value guess, so a `date_nanos` column holding both `2020-06-06T06:06:06.123456Z` (→ ns) and `2020-06-06T06:06:06Z` (→ ms) has keys six orders of magnitude apart — the same defect #790 described, confined now to `date_nanos`. Severity is low: a real `date_nanos` field is uniformly sub-millisecond.

The obvious fix — `date_nanos` → nanoseconds for **every** value — is not obviously safe, which is why #888 stopped short:

- `search/90_search_after.yml` → "Format sort values" creates `test` (`type: date`, `yyyy-MM-dd HH:mm:ss.SSS`) and `test_nanos` (`type: date_nanos`, `dd/MM/yyyy HH:mm:ss.SSS`), then sorts across `tes*` and asserts the two indices' documents **interleave chronologically**. That only holds today because `slash_date_to_epoch` has no nanosecond branch, so `test_nanos`'s slash-dates normalise to milliseconds and compare against `test`'s millisecond keys. Give `date_nanos` uniform nanosecond keys and `test_nanos` moves ~6 orders of magnitude away and sorts as a block.
- Real ES has the same cross-index hazard and answers it with the `numeric_type` sort option (`numeric_type: date_nanos` / `date`), which the same YAML file exercises. XERJ does not implement that conversion.

So the shape of the work is: implement `numeric_type` sort conversion (or an equivalent cross-index date normalisation), *then* pin `date_nanos` to nanoseconds. Doing the second without the first breaks conformance.

## 2. Existing indices are not back-filled

`FieldOptions::date_precision` is set when a field is mapped. A `FieldType::Date` persisted by an older build has no flag and stays on `PerValue`, so an index created before #888 keeps the #790 behaviour until its data is reindexed into a freshly mapped index (a `PUT _mapping` on an existing field is a no-op by design). This is the "legacy-infer default" #790's decision note recommended, and it is deliberate: a blanket back-fill to `Millis` would silently downgrade existing `date_nanos` indices to millisecond resolution.

If a migration is wanted, it needs a maintainer decision on how to tell a legacy `date` apart from a legacy `date_nanos` — they are indistinguishable in `schema.json`. One option: read the index's persisted `es_mapping.json` (which still carries the original `"type": "date_nanos"` string) at open time and back-fill from that.

## Related, also unfixed

`GET _mapping` renders a `date_nanos` field as `"date"` (pre-existing). Now that the distinction is recorded, the renderer could report it honestly — worth doing alongside item 2, since a mapping GET → PUT round trip is one of the ways a `date_nanos` field can lose its precision.

xerj-org:
Closes #889.

## Root cause

#790 made a date field's epoch scale a property of the **mapping**
(`FieldOptions::date_precision` → `xerj-engine`'s `DateScale`) and pinned a
declared ES `date` to milliseconds. It deliberately stopped short of
`date_nanos`, leaving it on the legacy per-**value** guess (">= 4
fractional-second digits means nanoseconds, anything coarser means
milliseconds"), because pinning `date_nanos` to nanoseconds moves it six orders
of magnitude away from a `date` field in a cross-index sort, and the
conformance suite pins that interleaving (`search/90_search_after.yml`, "Format
sort values").

So the defect #790 describes survived, confined to `date_nanos`. One
`date_nanos` column holding both

- `2020-06-06T06:06:06.123456Z` → epoch-**ns** `1591423566123456000`, and
- `2025-01-01T00:00:00Z` → epoch-**ms** `1735689600000`

has keys six orders of magnitude apart, so **2025 sorts before 2020** and a
range bound reaches one scale but not the other. That is the failure recorded
verbatim below.

Separately, `date_precision` is written when a field is *mapped*, so an index
created by a build older than #790 reloads with `date_precision: None` on every
date field and falls straight back to the same guess. Every data dir that
already existed still had the #790 defect regardless of what its mapping said,
and `schema.json` alone cannot tell `date` from `date_nanos`.

## The fix

The issue's own analysis is that the prerequisite has to land first, so it does.

**1. ES `numeric_type` on a field sort** (`date` / `date_nanos`). It forces the
epoch scale that **every participating index** produces sort values on, which is
the only way a sort over a field mapped `date` here and `date_nanos` there
compares chronologically at all. The `search_after` cursor is normalised on the
same forced scale, so paging works in that one space.
`Index::date_scale_for_sort` (`engine/crates/xerj-engine/src/index.rs`) is the
single resolver — consulted by `compute_sort_values`, the `search_after`
normalisation, the memtable bounded cut and `memtable_primary_key_rejects`;
`parse_sort` (`es_compat.rs`) and `parse_sort_field`
(`xerj-query/src/parser.rs`) parse the option.

ES semantics were checked against ES's own source rather than from memory
(`server/.../search/sort/FieldSortBuilder.java:335-380` in the reference
corpus — read for approach only, nothing copied): it resolves the sort's
numeric type per shard and moves the formatter to nanosecond resolution when
that resolves to `DATE_NANOSECONDS`.

**2. `date_nanos` is uniformly nanosecond-scaled** (`DateScale::Nanos`),
including day-first slash dates (`dd/MM/yyyy HH:mm:ss.SSS`) — `slash_date_to_epoch`
now takes a scale; it previously had no nanosecond branch at all and would have
re-introduced the mixed-scale split by a different route. The sync-readable
mirror `millis_date_fields` becomes `date_field_scales` (field → declared scale).

The **default** — no `numeric_type` — still compares the raw longs, exactly as
ES does and as `tests/es-compat-yaml/yaml/search/240_date_nanos.yml` asserts (a
`date_nanos` doc at `…12.123456789` sorts above a `date` doc at `…12.987` on
`desc`). That is pinned by a test of its own
(`without_numeric_type_the_two_scales_do_not_interleave`) so this change cannot
quietly turn the default into a conversion — and it is the one test in the new
module that passes both before and after.

**3. Legacy indices are back-filled at open, not refused.** `Index::open` reads
the index's own `es_mapping.json` — the raw user mapping, which still carries
the original `"type": "date_nanos"` string — and fills in any `date_precision`
that `schema.json` lacks (`backfill_date_precision`,
`backfill_date_precision_from_disk`). Both directions matter: a legacy
`date_nanos` comes back nanosecond-scaled (a blanket default would have
silently **downgraded** it to millisecond resolution — the outcome the issue
explicitly warns about) and a legacy `date` comes back millisecond-scaled, i.e.
it inherits the #790 fix without a reindex.

Nothing on disk is rewritten and no data is reindexed. That is possible because
the epoch keys were never persisted: `build_sort_shadow` derives them at read
time from the stored value (a Keyword doc-values column of the original
strings), `sort_shadow_cache` is in-memory and starts empty on every restart,
and `save_schema` already clears it on any scale change. Correcting the mapping
is therefore sufficient to correct the keys.

**4. `GET _mapping` reports a `date_nanos` field as `date_nanos`** where the
mapping is derived from the schema (it already round-tripped correctly when the
raw blob was present). The GET → PUT round trip every reindex helper performs no
longer silently downgrades the field.

**5.** One consistency fix found along the way: `normalize_search_after_value`
resolved a `format`-shaped cursor string through `es_format_to_epoch_ms`, which
is millisecond-resolution by construction. On a nanosecond-scaled field that
left the cursor ~9 orders of magnitude below every key; it is now widened onto
the field's scale.

A sort that re-scales its field with `numeric_type` declines the segment
sort-shadow prefilter (`shadow_scale_matches_sort`) rather than being served a
shadow built on the mapped scale — correct, just unaccelerated.

## Test proof — watched failing on unfixed code

The four behavioural edits were reverted **in place**, with the tests left
compiled and present, and the suites re-run. Recorded output:

```
running 6 tests   (xerj-engine, --lib)
test ...::backfill_only_fills_what_the_mapping_actually_says ... ok
test ...::date_scale_follows_the_declared_mapping ... FAILED
test ...::date_nanos_column_is_one_scale_whatever_the_value ... FAILED
test ...::legacy_index_backfills_date_precision_from_es_mapping ... FAILED
test ...::date_mixed_precision_sorts_and_ranges_consistently ... ok
test ...::date_nanos_field_keeps_nanosecond_sort_values ... ok

---- date_nanos_column_is_one_scale_whatever_the_value stdout ----
assertion `left == right` failed: #889 pre-flush: every value of a date_nanos
field is epoch-nanoseconds, including the whole-second one the old guess called
milliseconds
  left: [("late", [Number(1735689600000)]), ("sub", [Number(1591423566123456000)])]
 right: [("sub", [Number(1591423566123456000)]), ("late", [Number(1735689600000000000)])]

---- legacy_index_backfills_date_precision_from_es_mapping stdout ----
assertion `left == right` failed: #889 legacy-ns: an index whose schema.json
predates date_precision must recover its scale from es_mapping.json, not guess
it per value
  left: [("late", [Number(1735689600000)]), ("sub", [Number(1591423566123456000)])]
 right: [("sub", [Number(1591423566123456000)]), ("late", [Number(1735689600000000000)])]

test result: FAILED. 3 passed; 3 failed
```

```
running 4 tests   (xerj-api, --lib, mod date_nanos_cross_index_sort_tests)
test ...::schema_derived_mapping_reports_date_nanos_as_date_nanos ... FAILED
test ...::numeric_type_date_nanos_interleaves_a_date_and_a_date_nanos_index ... FAILED
test ...::numeric_type_date_interleaves_on_the_millisecond_scale ... FAILED
test ...::without_numeric_type_the_two_scales_do_not_interleave ... ok

---- numeric_type_date_nanos_interleaves_a_date_and_a_date_nanos_index stdout ----
assertion `left == right` failed: #889: numeric_type date_nanos must order both
indices by instant
  left: ["ms-feb", "ms-jun", "ns-apr", "ns-aug"]
 right: ["ms-feb", "ns-apr", "ms-jun", "ns-aug"]

test result: FAILED. 1 passed; 3 failed
```

Restored, the same tests pass:

```
running 6 tests   (xerj-engine, --lib)
test result: ok. 6 passed; 0 failed; 0 ignored; 672 filtered out

running 5 tests   (xerj-api, --lib)
test result: ok. 5 passed; 0 failed; 0 ignored; 230 filtered out
```

`backfill_only_fills_what_the_mapping_actually_says` is a unit test of new code
(the back-fill's widening / idempotence contract) and cannot fail on code that
does not contain it — stated plainly rather than counted as a regression proof.
`date_mixed_precision_sorts_and_ranges_consistently` and
`date_nanos_field_keeps_nanosecond_sort_values` are #790's own guards; they pass
on both sides, which is the point of keeping them in the run.

## Conformance

The suite runner was **not** run in full (CI does that, and its runner issues
`DELETE /_all`). A release binary from this branch was booted on a private port
with a throwaway data dir, and the date-sensitive files plus the two most
exposed directories were run against it:

| target | result |
|---|---|
| `search/90_search_after.yml` (incl. "date_nanos" and "Format sort values") | 4 passed · 0 failed |
| `search/240_date_nanos.yml` | 3 passed · 0 failed |
| `search/630_format_sort_missing_dates.yml` | 2 passed · 0 failed |
| `search/140_pre_filter_search_shards.yml` | 2 passed · 0 failed · 1 skipped (upstream `awaits_fix`) |
| `aggregations/date_histogram.yml` | 12 passed · 0 failed |
| whole `yaml/search/` directory (67 files) | **449 passed · 0 failed** · 2 skipped |
| whole `yaml/aggregations/` directory (78 files) | **639 passed · 0 failed** |

`search/90_search_after.yml` is the file the issue named as the blocker; it
passes with `date_nanos` pinned because `numeric_type` now does the conversion
its two "Mixed two types with numeric" cases ask for.

## Repair after review — two regressions this branch introduced

Review of `c3eb88f0` blocked on two paths that still assumed "a date value
normalises to epoch-MILLISECONDS" and were not moved onto the field's scale when
`date_nanos` was pinned to `DateScale::Nanos`. Both were real, both are fixed in
`f726b926`, and each one now has a regression test that was watched failing.

### 1. A `term` / numeric range bound on `date_nanos` returned 0 hits post-flush

`shadow_range_bounds` derives the bounds the segment sort-shadow is bisected
with. Its keys moved to epoch-nanoseconds with the rest of this change; a NUMERIC
bound stayed verbatim at epoch-milliseconds, six orders of magnitude below every
key. The bisect matched nothing, `build_range_prefilter_cached` returned
`Some(∅)`, and the scan loop takes an empty pre-filter as **proof** that no doc in
the segment can contribute — so it skipped the segment whole.
`try_shortcut_count`'s Range arm uses the same bounds, so `_count` said 0 too.

Two shapes reach that code with a numeric bound: the #788 `term`-on-a-date
rewrite, which lowers a `term` to the degenerate `range [epoch_ms, epoch_ms]`
from `aggs::parse_date_ms`, and a bound the user wrote. The memtable answers both
correctly (`parse_date_ms` on the bound AND on the stored value, ms vs ms), so
the doc matched before flush and vanished after.

Fixed **at the bound**, which is where ES puts it — a `date_nanos` field's
default format is `strict_date_optional_time_nanos||epoch_millis`, so a bare
number is epoch-millis and is converted to the field's resolution afterwards.
`epoch_ms_bound_on_scale` does exactly that and is a no-op off `Nanos`, so
`date`, an undeclared field and a legacy `PerValue` field are untouched. The
multiply runs in `i64` for a whole-millisecond bound, so the `f64` the bisect
sees is bit-identical to the shadow's own `i64 as f64` key for that instant.

### 2. A date-only value on the nanosecond scale became a raw string sort key

`date_string_to_epoch`'s `is_nanos` branch was a second, shorter ladder: RFC3339,
`%Y-%m-%dT%H:%M:%S%.f`, then `return None`. The `%Y-%m-%d` fallback existed only
in the millisecond branch. `looks_like_date("2021-01-01")` is true, so a date-only
value routed into the nanosecond branch and came back `None`;
`compute_sort_values` does `.unwrap_or(raw)`, so the `sort` value was the JSON
string `"2021-01-01"` among i64 epochs and `compare_values` fell to its string arm,
where `"` (0x22) sorts below every digit — the doc came **first** ascending
whatever its instant. `build_sort_shadow` also `?`s out on that term, disabling
the segment's pre-filter for the field.

This hit this PR's own headline feature too: `numeric_type: date_nanos` forces
`DateScale::Nanos` on **every** participating index, so an ordinary `date` field
holding `"2021-01-01"` returned a raw-string sort value under a cross-index sort.

The ladder is now walked **once** and the accepted instant rendered on the
requested scale at the end (`epoch_value_on_scale`), so a format either parses on
both scales or on neither. `normalize_search_after_value` carried a hand-copied
third copy of that ladder whose `NaiveDateTime` fallbacks returned
`timestamp_millis()` even at `Nanos` (the review's non-blocking item 4 — latent,
since nothing reaches those arms that RFC3339 has not already accepted); it now
calls the one function.

### Regression proof — each repair reverted alone

Both tests assert **pre-flush AND post-flush**, so flush invariance — the property
finding 1 broke — is pinned, not just the answer. Each was watched failing with
only its own repair neutered in place and the other left in, so the attribution is
1:1 rather than "something in this commit matters".

`epoch_ms_bound_on_scale` neutered (`|| true` on its early return):

```
test ...::date_nanos_term_and_epoch_millis_bound_survive_flush ... FAILED
test ...::date_only_value_on_the_nanosecond_scale_is_still_an_epoch ... ok

assertion `left == right` failed: #889 post-flush:
  {"term":{"ts":"2025-01-01T00:00:00Z"}} on a date_nanos field
  left: (0, [])
 right: (1, ["late"])
```

The split ladder restored (`if is_nanos { return None }` before the shared arms):

```
test ...::date_only_value_on_the_nanosecond_scale_is_still_an_epoch ... FAILED
test ...::date_nanos_term_and_epoch_millis_bound_survive_flush ... ok

assertion `left == right` failed: #889 pre-flush: a date-only value in a
  date_nanos field is an epoch-nanosecond number, not the raw string
  left: [("day", [String("2021-01-01")]), ("sub", [Number(1591423566123456000)])]
 right: [("sub", [Number(1591423566123456000)]), ("day", [Number(1609459200000000000)])]
```

Both restored: `xerj-engine --lib` 683 passed · 0 failed.

### Conformance coverage for the two regressions

The review is right that neither was reachable from the ES-YAML suite: every
`date_nanos` value in `search/240_date_nanos.yml` and `search/90_search_after.yml`
carries >= 3 fractional-second digits, and no case anywhere issues a `term` or a
numeric-epoch bound on a `date_nanos` field. `search/241_date_nanos_term_and_bounds.yml`
adds that wire-level coverage — a `term` in three spellings, a numeric epoch-millis
range bound in three windows, a `_count` cross-check, and a date-only value's sort
values — and flushes explicitly (`indices.flush`) so every assertion is answered by
the segment path rather than the memtable.

Verified **both ways** against a release build of this branch on a private port
with a throwaway data dir:

| build | `search/241_date_nanos_term_and_bounds.yml` |
|---|---|
| both repairs neutered | **0 passed · 3 failed** |
| repairs in | **3 passed · 0 failed** |

The three failures on the neutered build were `hits.total` 1→0 for the `term`,
3→1 for the numeric window, and `hits.hits.0._id` `sub`→`day` for the sort.

Whole-directory re-run on the repaired build, same server:

| target | result |
|---|---|
| whole `yaml/search/` (68 files) | **452 passed · 0 failed** · 2 skipped |
| whole `yaml/aggregations/` (78 files) | **639 passed · 0 failed** |

(449 → 452 is the three new cases; the suite runner was still not run in full —
CI does that.)

### Claim corrections the review caught

- The "year 65378" for `format_sort_value`'s magnitude heuristic was wrong: the
  threshold is `2_000_000_000_000_000`, so the millisecond side runs out about
  **63,000 years** past the epoch, not at calendar year 65378. The CHANGELOG now
  states the boundary in terms of the constant on both sides
  (2e15 ns = 23.1 days = 1970-01-24; 2e15 ms ≈ 63,000 years). The pre-existing
  comment above that constant, which said "~2 * 10^13" against a 2e15 literal, is
  corrected in the same commit.
- The "not fixed" list below now also names the three non-blocking gaps the
  review identified: numeric-shaped source values are not scaled, the back-fill
  does not recognise an ES-6 `{"_doc": {"properties": …}}` wrapper, and an unknown
  `numeric_type` VALUE is ignored where ES answers 400.
- On the presentation note: the two runs above are the same two-test selection on
  both sides, run three times (each repair alone, then both restored), so the
  before/after sets match exactly.

## What this does NOT fix, stated plainly

- An index whose `schema.json` predates the flag **and** which has no
  `es_mapping.json` (one that never went through the ES-compat mapping path)
  still has no record of `date` vs `date_nanos`, so its date fields keep the
  legacy per-value guess. Inventing a default there is exactly the silent
  downgrade this change exists to avoid.
- The back-fill reads `{"properties": …}` and `{"mappings": {"properties": …}}`,
  which is what `put_index_mapping` persists today. An ES-6-style type wrapper
  (`{"_doc": {"properties": …}}`) is skipped and that field keeps the legacy
  per-value guess — the same safe fallback, not a new failure mode.
- Only **string-shaped** date values move onto the field's scale. A numeric
  source value (`{"ts": 1591423566123}`) in a `date_nanos` field stays a raw
  millisecond number in its doc-values column, where ES would convert it. A
  segment holding both shapes ships no column at all and falls to the correct
  full scan, but two segments that each hold one shape still mix scales in a sort.
- Sort-value **formatting** (`format_sort_value`) still infers ms vs ns from the
  magnitude of the epoch number rather than from the resolved scale. Unchanged by
  this work: correct for every nanosecond instant from 1970-01-24 on
  (2e15 ns = 2e6 s = 23.1 days) and every millisecond instant below 2e15 ms
  (~63,000 years past the epoch).
- A `format`-shaped `search_after` cursor on a nanosecond field still loses its
  sub-millisecond digits (the format parser has no nanosecond path), so the page
  boundary can be up to 1 ms early — re-emitting at most the docs inside that
  millisecond rather than losing any. Item 5 narrowed this from a
  9-order-of-magnitude error to a 1 ms one; it did not close it.
- `numeric_type: long` / `double` are parsed and ignored (xerj already compares
  numeric sort keys as `f64`); `numeric_type` on a non-numeric field is accepted
  where ES rejects it; and an **unknown** `numeric_type` value is lower-cased and
  then ignored where ES answers 400.
- The segment shadow is bisected on the field's own scale while
  `doc_matches_query` re-tests through `parse_date_ms`, which truncates to
  milliseconds. On a `date_nanos` field a bound with millisecond-or-coarser
  precision can therefore exclude a doc from the pre-filter that the re-test
  would have admitted, when the two differ only in sub-millisecond digits. That
  is the pre-existing #789 asymmetry and is **unchanged** here — `main` behaves
  identically (its `PerValue` guess put the same bound and key on different
  scales); closing it means making the re-test scale-aware too, which is a
  separate change.
- No performance measurement is claimed. The `numeric_type` path declines the
  segment sort-shadow prefilter and was not benchmarked; the ordinary
  (no-`numeric_type`) path is unchanged in shape.

## Gates

- `cargo clippy --profile ci-test -j 8 -p xerj-common -p xerj-query -p xerj-engine -p xerj-api --all-targets -- -D warnings` — clean.
- Re-run after merging `origin/main` (rc.72 + #906/#907/#912) and applying the
  repair: `xerj-engine --lib` **683 passed · 0 failed**; `xerj-api --lib` 237
  passed · 0 failed; `xerj-query --lib` 186 passed · 0 failed; `xerj-common
  --lib` 111 passed · 0 failed.
- Integration targets touching the changed paths, run here rather than left to
  CI: `sort_on_meta_fields` 65, `sort_on_unresolvable_field` 130 (+1 ignored),
  `tied_score_page_order` 7, `tied_score_resorts` 3, `es_compat_tests` 4,
  `integration` 3 — all 0 failed.
- Conformance subsets as tabulated above.


xerj-org:
## Both blocking findings confirmed and fixed — `f726b926`

Neither was arguable. I reproduced both before touching anything, and both are
regressions this branch introduced: they work on `main`, where `date_nanos` was
`PerValue`.

Branch is now merged up to `origin/main` (rc.72 + #906/#907/#912; the CHANGELOG
conflict kept both sides).

### Finding 1 — `term` / numeric bound on `date_nanos` returns 0 post-flush

Reproduced exactly as traced. The failing run, before any fix:

```
#889 post-flush: {"term":{"ts":"2025-01-01T00:00:00Z"}} on a date_nanos field
  left: (0, [])
 right: (1, ["late"])
```

Pre-flush the same query returned `(1, ["late"])`, so it is precisely the
flush-invariance break described: the memtable compares ms-to-ms through
`parse_date_ms` on both sides, while the segment's shadow moved to nanoseconds
and the bound did not.

**Fixed at the bound, not at the value**, per your note about ES semantics: a
`date_nanos` field's default format is
`strict_date_optional_time_nanos||epoch_millis`, so a bare number is parsed as
epoch-**millis** and only then converted to the field's resolution. New
`epoch_ms_bound_on_scale` does that conversion inside `shadow_range_bounds` and
is a no-op off `Nanos` — `date`, an undeclared field and a legacy `PerValue`
field are byte-for-byte unchanged. Both call sites (`build_range_prefilter_cached`
and the `try_shortcut_count` Range arm) go through that one function, so the
`_count` half is closed with the same edit. The `Column::Numeric` arm is
deliberately left alone: its column holds raw millisecond source numbers, so an
ms bound against it is already the consistent pair.

One detail worth stating since it is load-bearing: the multiply runs in `i64`
for a whole-millisecond bound, so the `f64` the bisect sees is bit-identical to
the shadow's own `i64 as f64` key for that instant. Doing it as `ms * 1e6`
happens to agree today, but only by luck of correct rounding.

### Finding 2 — a date-only value on the nanosecond scale is a raw string

Also reproduced verbatim, and it fails **pre**-flush, which makes it the more
visible of the two:

```
#889 pre-flush: a date-only value in a date_nanos field is an epoch-nanosecond
number, not the raw string
  left: [("day", [String("2021-01-01")]), ("sub", [Number(1591423566123456000)])]
 right: [("sub", [Number(1591423566123456000)]), ("day", [Number(1609459200000000000)])]
```

`"` (0x22) below every digit, exactly as you said.

Rather than bolt `%Y-%m-%d` onto the nanosecond branch, I removed the second
ladder: `date_string_to_epoch` now walks the formats **once** and renders the
accepted instant on the requested scale at the end (`epoch_value_on_scale`), so
a format either parses on both scales or on neither and this class of divergence
cannot come back. `epoch_value_on_scale` reproduces chrono's `timestamp_millis()`
and `timestamp_nanos_opt()` exactly, over chrono's full range rather than nanos'
1677..2262 window, so nothing on the millisecond side moves.

Your `numeric_type` observation is covered too — the test's second half maps an
ordinary `date` field, puts `"2021-01-01"` in it, and sorts with
`numeric_type: date_nanos`; that fails identically on unfixed code.

While in the same function I also closed **non-blocking 4**:
`normalize_search_after_value` carried a hand-copied third copy of that ladder
whose two `NaiveDateTime` fallbacks returned `timestamp_millis()` even at
`Nanos`. It now calls the one function. I agree it was latent — I could not reach
those arms either — so it is closed as hygiene, not billed as a bug fix.

### Regression proof — each repair reverted alone

Both tests assert **pre-flush AND post-flush**, so the flush-invariance property
is pinned rather than just the answer. Each was watched failing with only *its
own* repair neutered in place and the other left in, so attribution is 1:1:

| neutered | `..._term_and_epoch_millis_bound_survive_flush` | `..._date_only_value_on_the_nanosecond_scale...` |
|---|---|---|
| `epoch_ms_bound_on_scale` early-returns always | **FAILED** | ok |
| split ladder restored (`if is_nanos { return None }`) | ok | **FAILED** |
| both restored | ok | ok |

`xerj-engine --lib` 683 passed · 0 failed with both in.

### Conformance — your coverage claim was right, and is now false

I checked it rather than took it: every `date_nanos` value in
`search/240_date_nanos.yml` and `search/90_search_after.yml` does carry >= 3
fractional digits, and no case in the suite issues a `term` or a numeric-epoch
bound on a `date_nanos` field. That is why CI was green through both defects.

`search/241_date_nanos_term_and_bounds.yml` closes it at the wire level — `term`
in three spellings, a numeric epoch-millis bound in three windows, a `_count`
cross-check, and a date-only value's sort values — with an explicit
`indices.flush` in setup so every assertion is answered by the segment path.

I verified it is not vacuous by building the server **both ways** and running the
file against each on a private port with a throwaway data dir:

| build | result |
|---|---|
| both repairs neutered | **0 passed · 3 failed** |
| repairs in | **3 passed · 0 failed** |

The failures were `hits.total` 1→0 for the `term`, 3→1 for the numeric window,
and `hits.hits.0._id` `sub`→`day` for the sort. Whole-directory re-run on the
repaired build: `yaml/search/` **452 passed · 0 failed** · 2 skipped (449 + the 3
new), `yaml/aggregations/` **639 passed · 0 failed**.

### Claim corrections

- **"year 65378" — you are right, and my arithmetic was wrong twice over.** The
  threshold is `2_000_000_000_000_000`, so the millisecond side runs out about
  **63,000 years** past the epoch; 63,378 (a count of years) was written as
  65,378 (a calendar year). Rather than publish another approximate calendar
  year, the CHANGELOG now states the boundary in terms of the constant on both
  sides: correct for every nanosecond instant from 1970-01-24 on
  (2e15 ns = 2e6 s = 23.1 days) and every millisecond instant below 2e15 ms.
- The pre-existing "~2 * 10^13" comment above that 2e15 literal is corrected in
  the same commit — thanks for flagging it; a comment that contradicts its own
  constant is worse than none.
- **Non-blocking 1 (numeric-shaped values) and 2 (ES-6 `{"_doc": …}` wrapper)**
  are now named in "what this does NOT fix", with your mitigation analysis (a
  type-mixed field ships no column and falls to the correct full scan; two
  segments each holding one shape still mix). **Non-blocking 3** (unknown
  `numeric_type` VALUE ignored where ES 400s) is named there too. I did not
  change the behaviour of any of the three — each is a real divergence, but not
  one this PR should grow.
- **Presentation note taken.** The runs above are the same two-test selection on
  every side.

### One thing I did NOT fix, stated rather than hidden

The segment shadow bisects on the field's own scale while `doc_matches_query`
re-tests through `parse_date_ms`, which truncates to milliseconds. On a
`date_nanos` field a bound with millisecond-or-coarser precision can therefore
exclude a doc from the pre-filter that the re-test would have admitted, when the
two differ only in sub-millisecond digits (e.g. `lte: "…06.000Z"` against a doc
at `…06.0004Z`). `main` behaves identically — its `PerValue` guess put that same
bound and key six orders of magnitude apart — so this is the pre-existing #789
asymmetry, unchanged, not something this branch introduces. Closing it means
either making the re-test scale-aware or widening the shadow bound to the
millisecond grid, and the latter would break `try_shortcut_count`'s
authoritative-count contract. It belongs in its own change. It is now in the
"does NOT fix" list.

### Gates

`cargo fmt --all -- --check` clean. `cargo clippy --profile ci-test -j 8
-p xerj-common -p xerj-query -p xerj-engine -p xerj-api --all-targets -- -D
warnings` clean. `xerj-engine --lib` 683 · `xerj-api --lib` 237 ·
`xerj-query --lib` 186 · `xerj-common --lib` 111, all 0 failed. Integration
targets over the changed paths (`sort_on_meta_fields`,
`sort_on_unresolvable_field`, `tied_score_page_order`, `tied_score_resorts`,
`es_compat_tests`, `integration`) 0 failed. Full suite runner still not run
locally — CI does that.

PR #923, opened 2026-09-11
feat(console): data-aligned Search/Discover + Case Review (derive index list, fields, facets from the mapping)

xerj-org:
## What

`match`/`term` resolve a field alias to its concrete target before running; `semantic` did not. A `{"semantic":{"field":"embedding"}}` query where `embedding` is an alias to the corpus's real semantic field (autoindex's `body`) fell through the `semantic_text` lookup in `run_semantic_with_deadline`, treated the aliased name as the vector field, and matched nothing.

## Fix

Resolve the alias first (`resolve_field_alias`) before the semantic_text field lookup, so the query finds the target field's embedding config and kNNs against its companion vector. One-line-ish change, mirrors what match/term already do.

## Why it matters

This is what makes the **bundled Console's SEARCH·DISCOVER semantic query work over any autoindexed corpus**. The console hardcodes `field:"embedding"`; autoindex names its neural field `body`. With this fix + a `embedding`→`body` alias, a lawyer/operator typing a plain-English question in the console gets real semantic results over an EML+PDF knowledge base.

## Verified

End-to-end on a live neural node over a 1,085-email + 116-PDF corpus: the console's exact query `{"semantic":{"field":"embedding","query":"…","k":10}}` returned 0 before, 10 real ranked hits (term sheets, valuation memos, board decks) after. fmt clean; compiles.

### DIFF
--- a/engine/crates/xerj-engine/src/index.rs
+++ b/engine/crates/xerj-engine/src/index.rs
@@ -14431,6 +14431,14 @@ impl Index {
         // the query with that same embedder and kNN against `target_field`.
         let (knn_field, dims, is_semantic_text, similarity) = {
             let schema = self.schema.read().await;
+            // Resolve a field alias to its concrete target first, so `semantic`
+            // respects field aliases the way `match`/`term` already do. A console
+            // or agent that queries `embedding` — aliased to the corpus's real
+            // semantic field (e.g. autoindex's `body`) — then finds that field's
+            // embedding config and kNNs against its companion vector, instead of
+            // falling through to treat the (aliased, vectorless) name as the
+            // vector field and matching nothing.
+            let field = &resolve_field_alias(&schema.schema, field);
             match schema.schema.field(field) {
                 Some(fc) if fc.embedding.is_some() => {
                     let emb = fc.embedding.as_ref().unwrap();
--- a/xerj-ux/assets/base.css
+++ b/xerj-ux/assets/base.css
@@ -2274,3 +2274,26 @@ button.sb-grouphead { width: 100%; }
   .sb-focuscard { order: -1; }
   .sb-readhow { display: none; } /* what + why survive; the how is a detail */
 }
+
+/* ── Case Review (email/document reader) ─────────────────────────────── */
+.cr-list{display:flex;flex-direction:column;max-height:62vh;overflow-y:auto}
+.cr-card{padding:12px 4px;border-bottom:1px solid var(--z-faint);cursor:pointer}
+.cr-card:hover{background:rgba(255,255,255,.02)}
+.cr-card.sel{box-shadow:inset 3px 0 0 var(--z-accent);background:rgba(255,196,0,.05)}
+.cr-card__top{display:flex;align-items:center;gap:8px;margin-bottom:4px}
+.cr-card__subj{font-weight:600;font-size:var(--fs-13);flex:1;overflow:hidden;text-overflow:ellipsis;white-space:nowrap}
+.cr-card__score{font-family:var(--font-data);font-size:var(--fs-11);color:var(--z-mute)}
+.cr-card__meta{color:var(--z-mute);font-size:var(--fs-11);font-family:var(--font-data);margin-bottom:4px;overflow:hidden;text-overflow:ellipsis;white-space:nowrap}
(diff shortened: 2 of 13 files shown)


xerj-org:
## Review — hold; one real CI failure, one honesty gap the PR was meant to close

The direction is right: Search/Discover should derive its fields, index list and facets from `_mapping` instead of the demo constants, and the per-dashboard status pill fixes are good. Findings, most severe first.

**1. Build + Test is red because of this PR, not a flake.** `seed::tests::seeds_every_registry_dashboard` (`engine/crates/xerj-console-api/src/seed.rs:910`) requires every `registry.js` dashboard to have a seed spec and `BUILTIN_DASHBOARD_COUNT` to match. `case-review` was added to the registry with neither. Add the seed spec and bump the constant.

**2. The PR body describes one of the seven commits.** It documents the `resolve_field_alias` change in `index.rs` and nothing of the other twelve files (Case Review dashboard, `schema.js`, `email-probe.js`, the pill rework). A reviewer cannot verify what the body does not claim. Rewrite it per the template: what each part changes, and the evidence for each.

**3. `*` is silently narrowed to one index.** `liveSearchDiscover` resolves `_all` to `indexNames()[0]` because the panel proxy only takes a single exact index. The picker keeps saying `*` while the results come from one corpus. On any two-corpus install this is exactly the "what I see is not what was indexed" complaint this PR exists to fix. Either make the proxy accept multi-index, or show the resolved index in the picker and the results eyebrow.

**4. The demo index list still flashes.** `INDICES = ['*', 'chat-events', 'logs-ssh-auth', …]` renders until the first mapping load on every fresh page. Should be `['*']` until the real names arrive.

**5. The mock fallback is untouched.** `runSearchNow` still seeds `state.search.result` with `mockSearch()` (fabricated `service`/`host`/`level` hits) and keeps it on any backend error. The pill is now honest, but the table content can still be fake. An explicit empty or error state is the fix; the pill alone does not close the gap.

**6. A corpus-specific demo question ships in the product path.** Case Review presets `q = 'what does the deal say about valuation and earnout'` for any email corpus it detects. An HR inbox or a support mailbox gets a venture-deal question. Derive a preset from the corpus (a `sample_queries_json` entry from the catalog) or leave it empty.

**7. The engine change has no test.** The alias resolution in `semantic` is correct and small; add a unit test beside the `match`/`term` alias tests so it stays that way.

**8. Preview DSL and executed DSL differ.** `buildDsl` (the preview) emits `hybrid.queries` as bare clauses with `rank_constant`; `buildSearchBody` (what runs) emits `{query, weight}` entries. The preview shows a body the engine would not accept.

**9. Ordering:** Case Review depends on #921's `email_*` / `attachment_name` fields and is gated by `requiresLive: 'email-corpus'`, so it is inert without #921. Merge #921 first, or split Case Review out of this PR so the Discover alignment can land on its own.

Verify before merge: that the `knn` companion field is really `${semanticField}_vector` for autoindex corpora (the code assumes it; nothing in the PR shows it).

xerj-org:
Superseded by https://github.com/xerj-org/xerj/pull/945 (`feat/console-corpus-reader`), which starts from this branch's commits and answers the nine findings above:

1. `seeds_every_registry_dashboard` — seed specs for the new `corpus` / `reader` ids, `BUILTIN_DASHBOARD_COUNT` 14→16, `SEED_REVISION` 2; Build+Test green.
2. PR body describes every part, with the test or recorded run behind each.
3. `*` still runs on one index (the proxy takes one exact name) but the results header names which, and the picker no longer pretends otherwise.
4. The index list is `['*']` until the real names arrive from the session API.
5. `mockSearch` and the fake plan/QPS/latency panels are deleted; Corpus, Reader and Discover are `NEVER_MOCK` — a failed search shows the error and zero rows.
6. No corpus-specific preset question; sample queries come from the catalog entry.
7. `test_semantic_field_alias_resolution` beside the match/term alias test.
8. Preview and executed body are built by one function (`search-body.js`), and HYBRID's shape was verified against a live node (no `aggs`, filter inside each leg — #943).
9. Case Review is replaced by a shape-generic Reader that does not depend on #921's fields to render; the email/attachment shapes use `email_*` / `attachment_name` when present.

"Verify before merge": the console no longer offers a raw kNN type; `semantic` / `hybrid` target the `semantic_text` field itself, which a live auth-on node answered 200 for.

Leaving this open for the coordinator to close.

xerj-org:
Closing as superseded — main 0594579d (`feat(console): corpus home, a reader for any record, read-only guest mode (supersedes #923)`, PR #945) landed this work through a different branch. Nothing here is still actionable.

PR #927, opened 2026-09-16
fix(xerj-code): preserve JSON output for empty search results

Amakurai:
Written with Codex (AI coding agent) for @Amakurai.

## What this changes, and why

`xc.py <corpus> <query> --json` prints prose when a loaded corpus has no matching passages, because the no-hit branch exits before JSON serialization. JSON consumers then fail to parse stdout for a normal empty search.

Emit the existing result object in JSON mode before handling text-only misses. Empty results still exit `1`, matches still exit `0`, and the unloaded-corpus guard still exits `3` with its actionable stderr diagnostic. BM25 response metadata, human-readable output, and MEATL output are preserved.

The regression covers BM25, hybrid, semantic, and combined `--json --meatl` misses, plus successful JSON and unloaded corpora. The existing offline suite is now run by the reference-coding CI job. README and CHANGELOG document the output contract.

## Evidence

Verified with Windows Python 3.14.6 and Ubuntu/WSL Python 3.14.4:

- Final `test_state_ledger.py` against the original `xc.py` from `099f314a`: **28 passed, 4 failed**. All four failures were empty-result serialization assertions; stdout contained `No passage in 'fixture' matches: ...` or `@no ...` rather than JSON.
- `python3 tools/xerj-code/tests/test_state_ledger.py` with the fix: **32 passed, 0 failed**.
- `tools/xerj-code/tests/test_xc_corpus.sh` on Linux: **28 passed, 0 failed**.
- Shell syntax checks, Python compilation, and `python3 tools/xerj-code/tests/validate_manifest.py --hub tools/xerj-code/hub/*.json`: passed.
- `python3 .github/scripts/workflow-timeout-guard.py --self-test` and the guard itself: passed.
- `git diff --check`: clean.

The failure is reproducible by running this PR's regression file while pointing its `XC_PATH` at the upstream `xc.py`; HTTP responses are mocked and corpus state is temporary. No live server is needed to exercise this CLI serialization branch.

## Scope and review

Independent branch based on current upstream `main`; no prerequisite PRs. Reviewed the complete diff and the previous unloaded-corpus fix (#476), and checked for overlapping open PRs. Only the Python retrieval output path, its regression tests, the existing CI job, and documentation change.

**Not run:** Rust formatting/build/tests, Clippy, and ES-YAML; no Rust engine or protocol behavior changes. No live XERJ server, real corpus indexing, semantic embedding, or performance benchmark was run. The semantic fixture has no semantic-capable index, so it exercises the empty-result presentation path only.

**Assumed / not independently verified:** real servers return the search-response shape already consumed by the existing client; this PR does not change requests or retrieval algorithms.

Contribution-review audit: machine-readable output, exit codes, fallback modes, CI coverage, ancestry, and public claims checked. Durable-state transitions, migrations, recovery, resource accounting, performance, and operational recipe lifecycle checks are not applicable to this output-only change.

CLA signature submitted separately in #928; awaiting its merge.


xerj-org:
@cla-bot check

Issue #933, opened 2026-09-18
autoindex --no-graph: index phase applies one file at a time — about 6 files/s, ~3.5 h for a 48,533-file corpus


xerj-org:
### What happens

On the durable `--no-graph` path, the `index` phase applies one file at a time, and on a large code corpus that is slow enough to decide whether the path is usable: **about 6 files/s, with an honest ETA of about 3.5 hours for the index phase alone** on a 48,533-file / 521 MB reference corpus.

This was invisible before #931 — the stream said `phase=scan pct=100.0 eta_quality=stalled` for this whole period, so nobody could see how long indexing took or that it was progressing at all. With the phase reported properly, the number is simply there to read.

### Measurement (2026-09-18, branch `fix/autoindex-resilience`, throwaway node, lexical embedder, 32 cores, machine shared with other builds)

```sh
xerj autoindex ~/.xerj-code/corpora/xerj-search \
  --url http://localhost:9540 --prefix xc-xerj-search --no-graph \
  --state-dir ./state-fix-1 --progress plain --yes
```

```text
xerj-progress phase=index basis=bytes pct=5.7 items=3120/47444 bytes=62161636/1099789983 rate=82701.2 eta_s=12546.7 eta_quality=rough since_progress_s=0.0 phase_elapsed_s=513.9 elapsed_s=825.0
```

3,120 files in 513.9 s = 6.1 files/s, roughly 165 ms per file. For comparison, the phases before it on the same run: `prepare` (1,526 datasets) about 17 s, `snapshot` (sealing 521 MB) about 5 minutes.

### Where the time goes (measured, not inferred)

`replay_pending_operations` (`engine/crates/xerj-autoindex/src/sync_executor.rs`) is a serial loop: `Started` journal record, `apply`, `Committed` journal record, per operation. For an upsert, `apply` first calls `delete_group` for the desired identity ("remove a partial prior retry"), then streams the sealed NDJSON as bulk requests.

Probes against the same node while the run was in flight:

| Step | Cost |
| --- | --- |
| `POST /<index>/_delete_by_query` that matches nothing, on the largest index (18,734 docs) | 46-57 ms (8 samples) |
| the same with `?refresh=true` | no measurable difference |
| `_bulk` of one tiny document | about 4 ms |
| journal-style append + `fsync` | 0.5 ms median |

[...]

PR #934, opened 2026-09-18
fix(autoindex): a refused dataset, a throttling node and an over-size catalog no longer end the run

### LINKED ISSUE #929: autoindex: one dataset whose mapping the server refuses aborts the whole run — 0 of 48,533 files indexed
### What happens

One dataset whose mapping the server refuses aborts the **entire** `xerj autoindex --no-graph` run. Nothing is indexed — not the refused dataset, and not the other 1,525.

### Repro (v1.0.0-rc.74, 2026-09-18)

Reference corpus `xerj-search` (five repositories incl. elasticsearch; 50,061 files walked after ignore rules, 48,533 unique files scanned), throwaway node on a private port, fresh state dir:

```sh
xerj autoindex ~/.xerj-code/corpora/xerj-search \
  --url http://localhost:9540 --prefix xc-xerj-search --no-graph \
  --state-dir ./state-b1 --progress plain --yes
```

```text
phase A: 1526 datasets inferred, 1089 junk/skipped files
...
xerj-done ok=false exit=1 reason=aborted wall=270.0s
error: install generation mapping for xc-xerj-search-elasticsearch-modules-10-basic-13: PUT /xc-xerj-search-elasticsearch-modules-10-basic-13/_mapping failed: 400 Bad Request {"error":{"root_cause":[{"type":"mapper_parsing_exception","reason":"nested semantic_text field [Check streams can't be enabled with existing logs.otel indices] is not supported; map semantic_text as a top-level field"}], ...
```

After the abort: `GET /xc-xerj-search*/_count` → `{"count":0}`, and 62 empty indices are left behind (the ones created before the loop met the bad dataset).

### Root cause

Two layers:

1. **Trigger** — a YAML test file's key (`Check streams can't be enabled with existing logs.otel indices`) contains a `.`, was elected `semantic_text`, and a dotted name is a nested path the server refuses. That specific trigger is being fixed separately (a dotted field name is never elected `semantic_text`).
2. **Structure (this issue)** — `ensure_generation_mappings` (`engine/crates/xerj-autoindex/src/lib.rs`) installs the dataset mappings in a loop and propagates the first failure with `?`:

   ```rust
   es.update_mapping(&dataset.index, &update_body)
       .with_context(|| format!("install generation mapping for {}", dataset.index))?;
   ```

   It runs from `provision_generation`, i.e. *after* the generation has been sealed into the journal, so the failure is also sticky: the pending generation replays into the same refusal on every re-run, and `--fresh` is refused because a generation is pending. Any future unmappable field — a field-count limit, a type the server does not support, a conflict with an index an earlier release created — takes the whole corpus down the same way.

### Expected

A refused mapping fails **that dataset**, loudly: named on the progress stream and in the run summary, its files recorded in the catalog with the server's reason, exit code 3 per the documented contract (`3 completed-with-junk — never fatal`). Every other dataset indexes. A generation must never be able to look complete while silently lacking a dataset, so the refusal has to be part of the durable plan/manifest, not just a log line. Endpoint failures (transport, auth, 5xx) are not dataset-specific and should still abort.

_Filed automatically by an AI agent on behalf of a maintainer._

### LINKED ISSUE #930: xc-index.sh --fresh fails on every previously indexed corpus (legacy state dir AND committed generation)
### What happens

`tools/xerj-code/scripts/xc-index.sh <corpus> --fresh` is the documented way to rebuild a reference-coding corpus (`xc.py` itself prints "Re-run xc-index.sh <corpus> --fresh" when an index is older than 30 days). It fails for **every corpus that has been indexed before**, in both possible states:

**1. State directory written before the `generation-v1` format** (every corpus indexed before that change):

```sh
# legacy state: any journal with a plan and no generation manifest
xerj autoindex ./tiny --url http://localhost:9540 --prefix xc-tiny --state-dir ./state-legacy --brain tiny --yes
xerj autoindex ./tiny --url http://localhost:9540 --prefix xc-tiny --state-dir ./state-legacy --no-graph --fresh --yes
```

```text
xerj-done ok=false exit=1 reason=aborted wall=0.0s
error: this state directory contains a legacy nonempty plan that cannot become generation authority: this state directory holds a pre-generation autoindex resume plan (3 planned files); ... it cannot be adopted in place and must be rebuilt. Start an independent rebuild using this argv JSON (no shell quoting required): ["xerj", "autoindex", ".../tiny", "--no-graph", "--url", "http://localhost:9540", "--state-dir", ".../state-legacy.generation-v1", "--prefix", "xc-tiny-generation-v1", ...]
```

The message is excellent and the argv is exact — but `xc-index.sh` has no path for it: it passes `--prefix xc-<corpus>` and the default state dir, prints `autoindex failed with exit 1`, and stops. Following the argv by hand moves the corpus to `--prefix xc-<corpus>-generation-v1`, which the state file does not record.

**2. State directory with a committed generation** (every corpus indexed since):

```sh
xerj autoindex ./tiny --url http://localhost:9540 --prefix xc-tiny2 --state-dir ./state-gen --no-graph --yes   # generation 1 committed
xerj autoindex ./tiny --url http://localhost:9540 --prefix xc-tiny2 --state-dir ./state-gen --no-graph --fresh --yes
```

```text
xerj-done ok=false exit=1 reason=aborted wall=0.0s
error: this attempt made no remote mutations. `--fresh` cannot discard committed corpus generation 1 under the same destination: ... For an exact rebuild, index the current folder with a new --state-dir and a new --prefix, and validate the isolated target before switching readers.
```

Both reproduced on v1.0.0-rc.74, 2026-09-18.

### Root cause

`autoindex --fresh` deliberately is *not* a rebuild: it discards the resume journal and never touches destination records, and it is refused once a durable generation exists (the refusal is correct — discarding the manifest loses the knowledge needed to clean up). The script forwards its own `--fresh` to that flag anyway, so "fresh" in `xc-index.sh` means something `autoindex --fresh` never promised. The rebuild route autoindex *does* recommend — new `--state-dir`, new `--prefix`, validate, then switch readers — is exactly what a wrapper that owns the whole `xc-<corpus>` namespace can automate, and the script does not.

### Expected

`xc-index.sh <corpus> --fresh` rebuilds the corpus whatever state an earlier run left: build the replacement beside the existing index, verify it (`_count > 0`), only then retire the old one, and keep `xc-<corpus>*` and the state file valid throughout so `xc.py` keeps working. A working index must never be deleted before its replacement has been verified.

_Filed automatically by an AI agent on behalf of a maintainer._

### LINKED ISSUE #931: autoindex --no-graph: progress reads scan 100% / eta stalled for the whole indexing phase — indistinguishable from a hang
### What happens

On the durable `--no-graph` path the progress stream keeps describing the `scan` phase — at 100%, with `eta_quality=stalled` and a `since_progress_s` that only climbs — for the whole time records are being sealed and indexed. That is byte-for-byte what a genuine hang at the end of scan looks like, so neither a person nor an agent can tell a healthy run from a wedged one.

### Repro (v1.0.0-rc.74, 2026-09-18)

```sh
xerj autoindex ~/.xerj-code/corpora/xerj-search \
  --url http://localhost:9540 --prefix xc-xerj-search --no-graph \
  --state-dir ./state-b3 --progress plain --yes
```

The phases the stream reported over a 270 s run (`grep -o '^xerj-progress phase=[a-z-]*' | uniq -c`):

```text
      1 xerj-progress phase=walk
      1 xerj-progress phase=hash
     54 xerj-progress phase=scan
```

48 of those 54 `scan` lines are this, with only the clocks moving:

```text
xerj-bar [########################] 100.0% | scan | 48533/48533 items | 16.3MB/s | eta unknown (stalled)
xerj-progress phase=scan basis=bytes pct=100.0 items=48533/48533 bytes=520892779/520892779 rate=17050775.5 eta_s=unknown eta_quality=stalled since_progress_s=250.0 phase_elapsed_s=260.7 elapsed_s=265.0
```

The TTY bar draws from the same snapshot, so it shows the same thing.

### Root cause

The generated route never enters a phase after `scan`. `begin_non_graph_generation` → `create_prepared_snapshot` (verify + copy + extract every file into sealed NDJSON, serially) and `replay_pending_operations` (mappings, one bulk per file, catalog publication, a per-group read-back barrier) all run with no `pr.phase(...)`, no item/byte accounting and no in-flight file name. Only the legacy graph path has `prepare` / `index` / `finalize-*` phases. `docs` and `content/answers/read-autoindex-progress.md` describe the twelve legacy phases as if they applied to every run.

### Expected

Each long step of the generated route is a phase with a real denominator: mapping install, snapshot sealing (bytes), indexing (sealed NDJSON bytes, with the in-flight file named), catalog publication and the verification barrier — in both the TTY bar and the `xerj-progress` / `--progress json` stream, with an ETA that obeys the existing quality gate. A resumed replay should count only the operations that remain.

_Filed automatically by an AI agent on behalf of a maintainer._

### LINKED ISSUE #944: autoindex: a per-item HTTP 429 (memory circuit breaker) inside one bulk aborts the whole run — 48,533 files aborted at 60% after 85 minutes
## What happened

Re-indexing the 48,533-file reference corpus (`xerj-search`: tantivy, meilisearch, quickwit, sonic, elasticsearch) with `xerj autoindex … --no-graph --progress plain` against a throwaway node. The node was at its default memory cap (`memory: detected 119.2 GiB usable, using a 16 GiB cap (auto tier)`, watermark 15,564 MB). The client had been indexing for 85 minutes and was at 60.4 % of the `index` phase when one bulk came back HTTP 200 with 747 of its items answered `status: 429` by the engine's real-memory circuit breaker, and the run aborted:

```text
xerj-progress phase=index basis=bytes pct=60.4 items=30036/47444 bytes=664119401/1099789983 rate=92730.0 eta_s=4698.3 eta_quality=good since_progress_s=0.0 phase_elapsed_s=4855.0 elapsed_s=5120.1 waiting_on=elasticsearch/libs/h3/src/test/resources/org/elasticsearch/h3/bc14r09centers.txt.gz(732B)
autoindex: server healthy again; raising bulk concurrency 31 → 32
autoindex: server pushed back (HTTP 429); lowering bulk concurrency 32 → 16 for this run
xerj-done ok=false exit=1 reason=aborted wall=5122.5s
error: prepared bulk contained 747 rejected items: {"type":"engine_exception","reason":"[parent] real memory circuit breaker tripped: rss=15679MB >= watermark=15564MB (94% of limit=16384MB); writes rejected to prevent an out-of-memory kill","status":429}
```

The server log for the same minutes shows the breaker is a passing event: it engaged 14 times in 27 minutes, and every engagement was released within 0.1–3.1 s:

```text
WARN xerj_engine::governor: RSS crossed the memory watermark — engaging the parent memory circuit breaker (writes → 429) rss_mb=15709 watermark_mb=15564
INFO xerj_engine::governor: RSS back below the memory watermark — releasing the memory circuit breaker rss_mb=15125     (1.3 s later)
WARN xerj_engine::governor: RSS crossed the memory watermark — engaging the parent memory circuit breaker (writes → 429) rss_mb=15706 watermark_mb=15564
INFO xerj_engine::governor: RSS back below the memory watermark — releasing the memory circuit breaker rss_mb=14921     (0.5 s later)
```

## Why it is a defect

The client already treats a *whole-response* HTTP 429 as transient (retry with backoff, shrink the admission window). A *per-item* 429 inside a 200 response is the same signal — the engine shedding load for a moment — but `checked_bulk` (`sync_executor.rs`) and `record_bulk_outcome` (`lib.rs`) treat any `server_errors > 0` as fatal, so the first breaker engagement that lands inside a bulk ends the run. 85 minutes of work, exit 1, nothing searchable published (the journal is resumable, but nothing told the caller so).

This is the same class as #929: one transient, per-unit condition aborting the whole run.

Second, smaller: the admission window's "server healthy again; raising bulk concurrency N → N+1" line is printed on every step. The capture holds 117 of them for 11 shrinks; the stream is hard to read.

## Expected

- Items a bulk answered 429 are re-sent — only those, after a backoff — for as long as the server is still accepting something, and for a bounded patience once it is not. Only then is the rejection fatal, with a message that says the run is resumable.
- A 429 beside a non-429 item failure is not re-sent (that bulk carries a bad record the caller must see).
- `xerj-done` reports how many re-sends happened (`bulk_retries=N`, only when > 0), so an agent can tell a run that fought back-pressure from one that did not.
- The concurrency-raise line is rate-limited.

## Reproduced with

v1.0.0-rc.74 semantics on the `fix/autoindex-resilience` branch binary (the client code on this path is unchanged from rc.74), 2026-09-18. Captures will be committed under `benchmarks/autoindex-resilience/`.

### LINKED ISSUE #955: autoindex: the catalog is written as ONE unwindowed _bulk — a corpus above ~48k files aborts in finalize-catalog on max_actions_per_bulk (413) after every operation was applied
## What happened

The full-corpus verification run of #934 over the reference corpus `xerj-search` (48,533 files; tantivy, meilisearch, quickwit, sonic, elasticsearch) applied **all 47,444 sealed operations** and then ended in `finalize-catalog`, 10,336 s in:

```text
xerj-progress phase=finalize-catalog basis=items pct=99.7 items=1521/1526 ... elapsed_s=10335.5
xerj-done ok=false exit=1 reason=aborted wall=10336.0s
error: prepared bulk contained 1 rejected items: {"type":"engine_exception","reason":"bulk request contains 102258 lines (~51129 actions); exceeds max_actions_per_bulk of 50000","status":413}
```

Binary: branch `fix/autoindex-resilience` at `702d4188` as both server and client, `--no-graph`, `--bulk-mb 8` (default), throwaway node with `XERJ_MAX_PROCESS_MEMORY_MB=49152`. The node's memory circuit breaker never engaged during this run (0 governor events in the node log), so this is not #944 and not #950.

## Root cause

Every per-file producer windows its `_bulk` bodies by bytes (`--bulk-mb`). The **catalog write does not window at all**, on either indexing path:

- generated path — `sync_executor.rs` `publish_generation_catalog` builds one body from every stale-id delete plus every projected document (one per file, per dataset, per run) and hands it to `checked_bulk` as one request;
- legacy path — `lib.rs` "write catalog" sends the whole `cat_buf` with one `es.bulk(..)`.

The catalog grows with the corpus: 47,444 files + 1,526 datasets + aliases + the run document = 51,129 actions in a 31.9 MB body (`http_body` peak `31910392` in the node's ingest-memory trace). The engine's `limits.max_actions_per_bulk` defaults to 50,000 (`xerj-engine/src/bulk.rs`), so **any corpus above roughly 48k files cannot finish**, and the failure arrives at the very end of the run.

## Two things make it worse

1. **On the legacy path the same rejection is silent.** The catalog write there treats only `server_errors` as fatal. The engine answers the over-limit request HTTP 200 with a single item of status 413, which is neither 429 nor 5xx, so it is counted as one `item_error` and ignored: the run would report success with no catalog written.
2. **`--bulk-mb` is not honoured** by the catalog request: 31.9 MB went out under an 8 MB window. At about 160k files the same body passes the engine's 100 MB `limits.max_body_bytes` and becomes a real HTTP 413.

## Expected

- The catalog write is windowed like every other bulk, in both dimensions (bytes and actions).
- A `_bulk` the server still answers "too large" (an operator may set `max_actions_per_bulk` or `max_body_bytes` far lower than the defaults) is halved and re-sent rather than ending the run; a single action the server refuses as too large stays the per-record error it is.
- The legacy catalog write cannot lose the catalog silently.

## Recovery for anyone who hits it today

Nothing is lost: every operation is journaled applied. Re-running the same command with a fixed binary resumes at `finalize-catalog`.

xerj-org:
Fixes #929, fixes #930, fixes #931, fixes #944, fixes #955. Found along the way and filed separately: #933 (slow serial replay, pre-existing) and #950 (the node's resident memory stays pinned above its watermark after a large ingest — a server defect, not fixed here).

All five defects were reproduced on 2026-09-18 and 2026-09-19 with the released v1.0.0-rc.74 binary (or, for #944 and #955, this branch's binary on the pre-fix client path) while re-indexing the 48,533-file reference corpus `xerj-search` (tantivy, meilisearch, quickwit, sonic, elasticsearch). The raw captures are committed under `benchmarks/autoindex-resilience/`.

## #929 — one dataset the server refuses aborted the whole run

**Before:** `ensure_generation_mappings` propagated the first dataset's mapping failure with `?`. One HTTP 400 on one of 1,526 datasets ended the run with `xerj-done ok=false exit=1 reason=aborted wall=270.0s` and nothing indexed.

**Root cause:** a server *refusal* of one index's mapping and an *endpoint failure* arrived as the same stringly `anyhow!`, so the only safe reading was the pessimistic one.

**Fix:**
- `esclient::MappingRefused` — only HTTP 400 on create-index / put-mapping is a refusal. 401/403/404/408/429/5xx still abort. `Display` is byte-identical to the old message.
- `preflight_generation_mappings` installs mappings **before anything is sealed**, the only point a dataset can still leave the plan coherently. `refuse_datasets` demotes a refused dataset into `Plan.refused_datasets` and its files into `junk_files`, so the manifest, snapshot, contract digests, catalog and exit code all derive from the plan that was actually accepted. `xerj-done` carries `datasets_refused` / `files_refused`, the catalog run document carries them plus `refused_datasets_json`, and `xerj autoindex map` prints a "Refused datasets — NOT indexed" section above the dataset table.
- Exit code is 3 through the existing contract (`files_junk > 0`), not a special case. `reconcile_plan` carries refusals forward; `sync.rs` freezes refused definitions like the dataset schema. A refusal of a dataset in an **already sealed** generation stays fatal but names every refused dataset and the recovery. Same split on the legacy graph path.
- The recorded reason is bounded to 1024 characters with control characters replaced, once, at the source.

Precedent, approach only (no code copied): Meilisearch marks one task `Failed` on a per-operation user error while the batch continues — `crates/index-scheduler/src/scheduler/process_index_operation.rs:173`.

**Verified end to end** on a throwaway node by pre-creating a dataset's index with a conflicting field type: `xerj-done ok=true exit=3 reason=completed-with-junk wall=0.7s files=1 records=1 generation=1 datasets_refused=1 files_refused=2`. A no-op re-run exits 3 again and still names the dataset. Refusing *every* dataset commits nothing searchable and is still exit 3, never 0 (tested).

## #931 — `--no-graph` progress read `scan 100% / stalled` for the whole indexing phase

**Before:** the rc.74 capture has 48 progress lines reading `phase=scan pct=100.0 eta_quality=stalled`, with `since_progress_s` climbing to 250.0, while the run was doing real work.

**Root cause:** mapping install, sealing, the replay loop and the read-back barrier never entered a phase, so the stream kept describing the last one that had.

**Fix:** `prepare` (datasets), `snapshot` (source bytes), `index` (sealed bulk bytes, in-flight file named), `finalize-catalog`, `finalize-refresh` (indices; split out after measuring 60 refreshes at 6.36 s on a 1,526-dataset node), `finalize-verify` (one read-back per file). A resumed replay counts only operations still to apply.

Verified on both surfaces (captures committed): the plain stream and the TTY bar under a pseudo-terminal report the nine phases, with 0 lines reading `scan` at `pct=100.0`. On the real corpus a generation interrupted at 12,890 of 47,444 operations resumed as `starting` → `replay` → `index` opening at `items=0/34554`.

## #930 — `xc-index.sh <corpus> --fresh` failed on every previously indexed corpus

**Root cause:** the script forwarded `--fresh` to `xerj autoindex`, which refuses it once a generation has committed, and a pre-`generation-v1` state directory cannot be adopted at all.

**Decision — script, not autoindex.** autoindex's refusal is deliberate and already prints the exact rebuild argv; silently rebuilding under another prefix would rename indices behind the caller's back. The wrapper owns the `xc-<corpus>` namespace and the state file, so it is the layer that can switch readers safely.

**Fix:** build beside (`xc-<corpus>-b<stamp>-*`, own `--state-dir`), verify (exit 0/3 **and** `_count > 0`), switch the state file by atomic rename, then retire the old indices by exact name. `prefix` stays `xc-<corpus>` (the older `xc.py` globs `prefix*` and still matches); `index_prefix` names the verified build and the updated `xc.py` prefers it. `tools/xerj-code/tests/test_xc_index_fresh.py` (46 offline checks) fails against `main`'s script with the literal error from the issue; both toolkit test files now run in CI.

## #944 — a per-item HTTP 429 inside one bulk aborted the run at 60% after 85 minutes

Found by the full-corpus verification of the three fixes above. **Before:** at 60.4% of the `index` phase, 5,122.5 s in, one bulk came back HTTP 200 with 747 items answered `status: 429` by the engine's real memory circuit breaker (16 GiB auto cap on a 119 GiB machine), and `checked_bulk` aborted: `xerj-done ok=false exit=1 reason=aborted wall=5122.5s`. The node log (`before-944.node.governor.txt`) shows 14 breaker engagements in 27 minutes, every one released within 0.1–3.1 s: the condition had cleared about a second after the run died. Same class as #929 — one transient, per-unit condition ending the whole run.

**Fix** (`Es::bulk`, so the sealed replay, the legacy path and the catalog write all get it): when every failed item in a bulk is a 429, re-send only the rejected actions — cut out of the sent body by position, walking `delete` actions by verb because they carry no document line — after the existing 250 ms → 8 s backoff, for `THROTTLE_PATIENCE` (600 s, measured from the first send of that bulk). *This is now main's implementation, adopted whole in the #949 reconciliation below; the 120 s idle-clock this PR originally shipped is gone.* Only then is the rejection handed back as `server_errors`, which every caller already treats as fatal, now with a message that names back-pressure and says the same command resumes the run. A 429 beside a non-429 failure is never re-sent (that bulk carries a bad record the caller must see). The admission window shrinks on the first throttled round, once per event. `xerj-done` carries `bulk_retries=N` only when it happened; the generated path's summary also carries `bulk_items_reissued`. The `raising bulk concurrency` line is printed at most once per 10 s naming the whole climb, plus the step that reaches the ceiling (117 lines for 11 shrinks in the capture).

Precedent, approach only, no code taken: Elasticsearch `action/bulk/Retry.java` (retry only when every failed item is 429, re-issue failed items by position, exponential backoff) — read from the clone on disk, because the `xerj-search` retrieval index is the thing this task rebuilds last.

Tests: six unit tests on a stub HTTP server (positional re-send across a `delete`, a shrinking remainder followed to zero, patience exhausted → `server_errors`, a 429 beside a 400 not re-sent, selector refuses a count mismatch / non-JSON), and one end-to-end test per indexing path (legacy `MockEndpoint` answering a whole bulk 429 once; generated `HttpEndpoint` answering the odd items 429), each asserting every record landed, exit 0, and `bulk_retries=1`.

**The second shape.** Resuming the interrupted generation with that fix ended 1,039.6 s in on the *other* way the engine answers while its breaker is engaged: HTTP 429 on the whole `_bulk` request (`error: _bulk: HTTP 429 Too Many Requests`, capture `before-944.whole-request-429.stderr.txt`). That path still went through `with_retry` — six transport attempts, about 8 s of backoff — and then out of `bulk_attempt(..)?` as fatal, while the same rejection carried per item would have been waited out for the whole patience. Now `_bulk` receives its 429 inside `retry_loop`: when the body is a full bulk response, the items are honoured (an action the engine says it accepted is not written twice) and it joins the per-item patience loop; otherwise `retry_loop` re-asks for the whole body at the capped backoff for `THROTTLE_PATIENCE`, and counts each re-ask on `bulk_retries` / `bulk_items_reissued`. A 429 whose body cannot be mapped onto the actions stays the transport error it was. Main's unit tests cover the first three shapes; this branch's generated-path end-to-end run answering three whole-request 429s in a row (`bulk_retries=3`, `bulk_items_reissued=12`, every record landed, exit 0).

## #955 — the catalog went out as ONE `_bulk`, over the engine's action limit

Found by the full-corpus verification of #944. **Before:** the `--no-graph` run applied all 47,444 operations and then ended `xerj-done ok=false exit=1 reason=aborted` in `finalize-catalog` after **10,336 s** (`before-955.stderr.txt`). It was read as a memory failure at first; it was not. The node log for that run has **no governor line at all** (`before-955.node.txt`) — the breaker never engaged. `finalize_catalog` sent the catalog — one document per file, per dataset, per run — as a single `_bulk` of **51,129 actions in 31.9 MB**, and the engine answered item 413:

```
bulk request contains 102258 lines (~51129 actions); exceeds max_actions_per_bulk of 50000
```

On the generated path the same answer was **silent**: exit 0 with an empty catalog, because a 413 on one item of a catalog write was not checked.

**Fix:** every `_bulk` body goes out in windows of at most 10,000 actions, and the catalog also respects `--bulk-mb`. A request the server still refuses as too large — HTTP 413, or one item answered 413 for a body of several actions — is halved and re-sent, and the reduced bound is kept for the rest of the run; `xerj-done` carries `bulk_splits=N` when that happened. A body of ONE action that is still refused is fatal with a message that names `limits.max_body_bytes` and the two ways out, because there is nothing left to halve. Separately, a `--no-graph` run that exhausts #944's back-pressure patience now ends `reason=server-backpressure` with `ops_applied` / `ops_remaining` rather than the undifferentiated `reason=aborted`; still exit 1, still resumable.

**Verified on the corpus that broke, not on a stub.** The journal and the node data directory of the 10,336 s run were copied, the node restarted on this branch's binary, and the *same* command re-run: `xerj-done ok=true exit=3 reason=completed-with-junk wall=415.0s files=47444 records=821840 generation=1`, and `autoindex-catalog` went from 0 to **51,129** documents — the exact action count of the request that had been refused. The node's `http_body` logical peak dropped from 31,910,392 bytes to 8,388,241 (`<= --bulk-mb 8`). Captures: `after-955.full-corpus-resume.stderr.txt`, `after-955.node.txt`.

**And on a node configured to refuse.** A stub can only return the literal the engine returns; it cannot prove the client and the engine agree on what "too large" means. So the two refusal shapes were forced out of a real engine — this branch's binary as both server and client, the 231-file `sonic` repository, one fresh throwaway node per setting on a private port (14300-14317), auth on, each beside a control run on the defaults. Full capture: `benchmarks/autoindex-resilience/limits-real-node.txt`.

**Every row below was RE-MEASURED on 2026-09-20**, after the #949 reconciliation replaced 120 s of patience with 600 s and changed the stderr line. Nothing was edited to match; the 2026-09-19 numbers the docs used to quote are gone from every surface. Machine: 32 cores, 119.2 GiB, shared with other builds — `uptime` read `load average: 24.93, 21.87, 21.50` when the first run started and `16.83, 19.06, 20.81` when the last one ended, so the wall times are not reproducible to the second; the counts and the exit codes are the contract.

| Node setting | What the engine answered | Outcome |
| --- | --- | --- |
| control (defaults) | — | `ok=true exit=3` · 1,663 records · 236 catalog docs · 6.1 s |
| `limits.max_actions_per_bulk = 64` | one 69-action request refused: `exceeds max_actions_per_bulk of 64`, answered as the bulk **item's** status 413, not the request's | halved and re-sent · `ok=true exit=3 bulk_splits=1` · **1,663 records · 236 catalog docs** — the control run's counts exactly · 5.6 s |
| `limits.max_body_bytes = 98304` | one 120,443-byte request refused: HTTP 413 `length limit exceeded` (whole request, not an item) | halved and re-sent · `ok=true exit=3 bulk_splits=1` · **1,663 records · 236 catalog docs** · 5.5 s |
| `limits.max_body_bytes = 65536` | a single record of 70,471 bytes cannot be cut | `exit=1 reason=aborted`, error names `limits.max_body_bytes` and the two ways out; the 958 records that fitted landed first |
| `XERJ_MAX_PROCESS_MEMORY_MB=64` (#944's terminal case) | breaker engaged from start-up; `2 of 2` items 429 on every send | 600 s of re-sends, then `ok=false exit=1 reason=server-backpressure wall=609.1s ops_applied=0 ops_remaining=231`, with `stopped by server back-pressure while applying <file>` |
| …the same state dir, node restarted on the default cap, same command | — | `resumed and committed pending corpus generation` · `ok=true exit=3` · 1,663 records · 236 catalog docs · 3.5 s |

So the split holds at `max_actions_per_bulk = 64`, a limit 781× smaller than the default the corpus tripped, and at a body limit small enough to refuse a 49-action request: same record count, same catalog count, same exit code as the control run in both. The one case that is not recoverable — a single record larger than the whole body limit — is the one where halving has nothing left to halve, and it says so.

Two honesty notes on the re-measurement, stated in the capture rather than smoothed over. The "16 re-sends" the old capture reported was really the count of its announce lines, one per 5 s; main's announce is one per 30 s, so the new run's **19 notices are not a re-send count** — the re-sends are not counted on stderr, and the run ends before an `xerj-done` line could carry `bulk_retries`. And `max_body_bytes = 65536` landed 958 records before the un-cuttable one where the old run landed 886: how many windows land first depends on which worker reaches the oversize record, so the contract there is the error, not the count.

Both user-facing "too large" strings are one sentence each; they were committed carrying 18–22 spaces of source indentation (`0cec355e`), which no formatter or lint can see, and a regression assert now pins the one-action message against a double space.

## #950 — the node's side, filed and not fixed here

The resumed run above began with the node already at its watermark. 2.5 h after its last write, with no client connected, the throwaway node still held 14.8 GB RSS — 14.76 GB of it anonymous memory — for 1.2 GB on disk (537,029 records in 1,526 indices); the hydration cache accounted for 3.0 GB of it, `_flush` and `_cache/clear` released nothing, and every write was 429 until the node was restarted. jemalloc runs with `background_thread:true` and 1 s decay, so this is live memory, not allocator retention. #944's client patience is correct and cannot help a node that never accepts again; on the default 16 GiB auto tier this corpus cannot finish on that node. The numbers, the `/proc` split and the `_nodes/stats` accounting are in #950. The docs say so in words (recipe §8, the 429 page's "If it did exit 1", `llms-full.txt`).

## Status — full-corpus verification: done

Three runs on the 48,533-file corpus, all on this machine (32 cores, 119 GiB, shared with other builds — the captures' own timings are not reproducible to the second; `uptime` while the gates below ran read `load average: 4.90, 3.67, 2.77`):

| Run | Result |
| --- | --- |
| rc.74, full corpus | `ok=false exit=1 reason=aborted wall=270.0s`, nothing indexed — #929 and #931 in one capture (`before-rc74.stderr.txt`) |
| this branch before #955, full corpus | every one of 47,444 operations applied, then `exit=1` in `finalize-catalog` after 10,336 s (`before-955.stderr.txt`) |
| this branch, resuming that same generation | `ok=true exit=3 reason=completed-with-junk wall=415.0s files=47444 records=821840 generation=1`, 51,129 catalog documents, no refused datasets (`after-955.full-corpus-resume.stderr.txt`) |

The rc.74 trigger is reproduced on a 4,479-file slice that keeps it (`slice-rc74.stderr.txt`: the same `nested semantic_text field [Check streams can't be enabled with existing logs.otel indices]` 400, `exit=1` at 39.6 s). This branch on the same slice with the same command: `ok=true exit=3 wall=525.6s files=3680 records=159666` (`slice-after.stderr.txt`). Every number the docs quote from this folder is derived by `summarize.py` into `results.json`, not typed by hand; `summarize.py --check` is how it stays true.

## Reconciled with `#949` — one throttle implementation, main's

`#949` (mbox / Google Takeout ingest) landed on `main` carrying a **second, independently written implementation of the same per-item-429 fix** this PR ships for #944. Same concern, different code, and `git merge origin/main` puts ten conflict hunks in `esclient.rs`, two of them 300 and 750 lines, because both sides rewrote the same functions. Neither "ours" nor "theirs" is correct: one reverts #949's fix and its tests, the other reverts #944's and takes #955 and #929 with it. The analysis is in [this comment](https://github.com/xerj-org/xerj/pull/934#issuecomment-5746889242).

**Resolution: keep main's throttle machinery, whole.** It is already shipped, and on this axis it is the more capable of the two — ten minutes of patience instead of two, and an HTTP-level 429 gets that same patience instead of six transport attempts. There is now exactly ONE retry/throttle path in `esclient.rs`, not two side by side.

| Concern | this branch (#944) | `main` after #949 | kept |
| --- | --- | --- | --- |
| per-item 429 re-send | `select_bulk_actions` / `walk_bulk_actions` | `select_bulk_items` / `bulk_item_ranges` | **main's** |
| retry policy | `with_retry_throttle` + `Throttle` | `retry_loop` + `Attempt` | **main's** |
| parsed response | `ParsedBulk` | `Round` | **main's** |
| patience | `BACKPRESSURE_PATIENCE` 120 s, from the last accepted item | `THROTTLE_PATIENCE` 600 s, from the call | **main's** |
| whole-request 429 | `Throttle::HandToParser` | `bulk_round` + `Attempt::Throttled` | **main's** |
| the stderr line | `server back-pressure: N of M …`, every 5 s | `server is shedding load — …; re-offering N rejected record(s)`, every 30 s | **main's** |

Re-applied on top of it, because `main` has none of it:

- **#955** — `BULK_MAX_ACTIONS`, `window_bulk_body`, `count_bulk_actions`, `TooLarge`, `SentWindow`, `RequestSizeLedger`, `halve_refused_request`, `bulk_windowed`, `bulk_requests_split`. `Es::bulk` is now `bulk_bounded`, a queue: it windows a body, sends each window through main's per-window loop (`bulk_window` — main's `Es::bulk`, renamed, with the too-large exit added), pushes the halves of a refused request back on the *front*, and re-reads the learned bound every round, so it converges to one action. `window_bulk_body` and `count_bulk_actions` are built on main's `bulk_item_ranges`, so the NDJSON body is walked by ONE function, not two. The e2e test drives it at the stub's `max_actions_per_bulk = 2`.
- **#929** — `MappingRefused` and `mapping_error`: one refused dataset costs that dataset, not the run.
- **the announce throttle** — `RAISE_ANNOUNCE_EVERY`, one "raising bulk concurrency" line per 10 s naming the whole climb (117 lines for 11 shrinks in the motivating capture).
- **`throttled_out` / `BackpressureExhausted`** — `reason=server-backpressure` with `ops_applied` / `ops_remaining`, raised from main's loop when *its* patience is exhausted.
- **`bulk_retries` / `bulk_items_reissued`** — moved onto main's loop, and both shapes now count: `bulk_window` counts each per-item re-offer, and `bulk_round` counts a whole-request 429 that `retry_loop` re-sent. The counter no longer depends on which shape the engine's breaker happened to answer with, which is what the generated-path e2e test (`bulk_retries=3`, `bulk_items_reissued=12`) asserts.
- **#930** (`xc-index.sh --fresh` adopting a previously indexed corpus) and **#931** (honest `--no-graph` progress) are untouched by the collision — neither side of it goes near them.

Dropped: this branch's ten #944 unit tests and its `with_backpressure_patience` accessor. Main's equivalents cover the same ground against main's structures, and keeping both would have meant asserting two contradictory patiences. #955's and #929's tests are kept, rewritten against main's test helpers (`pair`, `item`, `items_response`, `request_body`).

**Two behaviours genuinely changed, and are documented rather than hidden:**

- patience is 600 s measured from the first send of a bulk, not 120 s measured from the last response that accepted an item. A bulk making slow progress for more than ten minutes is now given up on where the old rule would have kept waiting; a bulk making none is waited on five times longer.
- the stderr line is main's, at most once per 30 s per waiting bulk.

Both invalidate every published number and quoted line that came from the 120 s build, so **all of it was re-measured on a real node, not edited to match** — see the table below.

## Merged `origin/main`

`origin/main` moved a long way while this was verifying: #966 (a real S3-compatible `StorageBackend`), #970 (`xerj autoindex s3://…`), #967 (`--watch`), #973/#972 (site-generator dating) and three `xc.py` fixes (#927, #956, #958). Merged as a merge commit, not a rebase, because the conflicts are worth reading in the log.

Two of them were not mechanical:

- `incremental_reconcile_http_tests.rs` — both sides appended a test block to the same tail and git interleaved them, landing a `>>>>>>>` mid-function. Rebuilt from the three merge stages: merged prefix, then this branch's block whole, then main's `--watch` convergence block whole. No duplicated `fn` name in the result.
- `xc.py` — #930 made the `--mode semantic` arm address the *verified build* (`query_prefix(state)`) instead of the whole namespace; #956 gave the same two calls `fatal=True`. Both apply; taking either side alone would have silently reverted the other's fix.

Two more merges followed, both run with `rerere.enabled=false` and no rr-cache:

- **`fb44b2cb` (#949, mbox / Takeout ingest)** — the duplicate-implementation collision, resolved in the section above. Its only other non-mechanical conflict was `lib.rs` `finish_generated_run`, where both sides added a block to the same two places (#929's committed refusals, #949's unextracted-archive lines); both apply.
- **`c81154d7` (#946, the rerank stage)** — touches no file this branch touches. Only the generated `landing/**` pages conflicted.

`landing/**.html`, `landing/answers/*.md`, `landing/answers/index.json` and `landing/sitemap.xml` were taken from main in each merge and then regenerated from the merged `content/` — a hand-merged generated page is one nobody can reproduce. `landing/llms.txt` was checked line by line after every merge (a `rerere` replay dropped *both* sides of that file on another branch, cleanly and with no warning): it carries main's mail-export, archive-extraction, S3, freshness, rerank-caveat and `_xerj/rerank` lines **and** this branch's dataset-refused, back-pressure-429 and `--no-graph` phase lines.

## Gates

Re-run on the reconciled tree (`9b4dad18`, which merges main up to `c81154d7`) on this machine, `load average: 18.17, 21.07, 20.51` while they ran — it is shared with other builds — with `CARGO_TARGET_DIR` inside the worktree so nothing shared a target dir:

| Gate | Result |
| --- | --- |
| `cargo fmt -p xerj-autoindex -- --check` | clean |
| `cargo clippy --release -p xerj-autoindex --all-targets -- -D warnings` | 0 warnings |
| `cargo test --release -p xerj-autoindex` (lib) | **967 passed, 0 failed, 2 ignored** (897 before #949 and #946 came in; the reconciliation dropped this branch's ten duplicate #944 unit tests and kept main's, and added #955's eight rewritten against main's helpers) |
| `cargo test --release -p xerj-autoindex` (integration) | `follow_symlinks_boundary` 8, `journal_lock_inheritance` 1, `scan_pool_width` 1, `sniff_printable_magic` 4 — all passed; doc-tests 0 |
| `objsource_minio_tests` (main's, 5 cases) | self-skipped: no `XERJ_MINIO_ENDPOINT`. No object-storage endpoint was contacted by this branch's verification, on this machine or anywhere else |
| `test_xc_index_fresh.py` / `test_state_ledger.py` | 81 / 44 passed (the ledger file now also carries #927's `--json` cases) |
| `test_passage_window.py` / `test_hybrid_fallback.py` (main's, merged) | both OK against the merged `xc.py` |
| all 9 SEO gates (`build_articles`, `gen_sitemap`, `fix_heads`, `fix_links`, `mk_og_card` `--check`; `seo_lint`; `factcheck --self-test` / `--fixture-check` / `--fail-on error`) + `audit.py --fail-on critical` | **ALL SEO GATES PASS**; factcheck **0 ERROR**, 89 WARN over 87 files (88 before #946; the one new WARN is main's, not this branch's) |
| `landing-constants-guard.sh`, `workflow-timeout-guard.py` | pass |
| generators are settled | a further `--write` pass of all five leaves `git diff` empty |
| Dependency advisory audit (cargo-audit) | **now green** on this PR's CI. It was red on RUSTSEC-2026-0285 (rustls 0.23.38) before the merge; that was never this branch's lockfile change, and merging `main` picked up the bump. #951 records it |
| ES-YAML conformance (`es-yaml-runner`, fresh throwaway node on a private port) | 1371 passed, 0 failed, 3 skipped — run on `0cec355e`'s binary. Not re-run after the merges: this branch changes no engine or server crate, and main's own CI gates the crates it added |
| Real-node re-measurement (`limits-real-node.txt`) | all six cases re-run on ports 14300-14317 with throwaway data dirs, `xerj` built from `53e34f25` as both server and client. Nothing touched `:9200`, and **no request was made to any Cloudflare R2 endpoint** |

CI on the reconciled head is the gate that matters and is reported on the PR's own checks, not assumed here.

Reference-coding mandate: retrieval was used for the two design precedents above (Meilisearch's per-operation failure, Elasticsearch's `Retry.java`, both approach-only). It was skipped for everything else, deliberately — the rest is confined to this repository's own code. That includes the #949 reconciliation: choosing between two implementations that both live in this repository, and re-applying #955 and #929 on top of the survivor, is a question ordinary file reads answer more accurately than retrieval. Say-so per the mandate.

## Reference-coding index for `xerj-search` on :9200

Still not rebuilt. The full-corpus verification ran against throwaway nodes on private ports with copies of the state directory, deliberately, so that a `DELETE /_all` in any runner could not reach the reference corpora. Rebuilding the shared :9200 index with `xc-index.sh <corpus> --fresh` — which is what #930 makes possible — is a follow-up, not a merge condition.

## Known limits

- A refusal is decided when the corpus is first built and is frozen into the generation. If the server is later fixed, rebuild under a new prefix to pick the dataset up.
- A file that feeds several datasets is dropped whole when one of them is refused.
- A refused dataset's index may be left behind empty; autoindex does not delete an index it cannot prove it created.
- Only a field-type conflict was forced end to end for #929. Other causes of a 400 share the code path but were not run.
- #944's patience is a constant, not a flag: `THROTTLE_PATIENCE` = 600 s, measured from the first send of a bulk. A node pinned above its memory watermark for longer than that still ends the run — exit 1, resumable, with the cause in the error line. #950 is exactly that node. The 120 s idle-clock this PR originally shipped is gone: #949's implementation won the reconciliation, and a longer wait is the safer default against a slow server.
- Because that clock runs from the call and not from the last accepted item, a bulk that is landing records *slowly* for more than ten minutes is now given up on where this branch's old rule would have kept waiting. No capture on this branch or on main shows that shape; it is stated because it is a real difference, not because it was observed.
- The full-corpus verification completed on a node with `XERJ_MAX_PROCESS_MEMORY_MB` raised above the 16 GiB auto tier. That is the documented workaround for #950, and it is stated wherever the result is quoted.
- #955's bulk windows are constants (10,000 actions; `--bulk-mb` for the body), not flags. The halving on a 413 is per run and only ever tightens; it is not remembered across runs.
- The `max_actions_per_bulk = 64` and `max_body_bytes` runs used the 231-file `sonic` repository, not the 48,533-file corpus: the point of those runs is the refusal-and-split contract against a real engine, which a small corpus proves in seconds and a large one proves no better.
- The main checkout's `.claude/skills/xerj-code/scripts/` copy of the toolkit is gitignored and older than `tools/xerj-code/scripts/`; the `--fresh` fix lives in the tracked copy.


xerj-org:
## Blocked on a duplicate-implementation collision with #949, not on a merge conflict

State right now: this branch is at `d93f3809`, which already merges `origin/main` up to `fb5b0e73` (#966 S3 `StorageBackend`, #970 `autoindex s3://`, #967 `--watch`, #972/#973 site dating, #927/#956/#958 `xc.py`). **All 21 CI checks pass on that head** — including `Build + Test` (44m33s), `ES-compat YAML conformance`, `Release build`, `Autoindex FD smoke` on all three OSes, and the `Use-case harnesses` job that was red before the merge.

`#949` (mbox and Google Takeout ingest) landed on main about an hour later, and it is **not** a textual conflict that can be resolved by keeping both sides. `#949` contains a second, independently written implementation of the same thing this PR ships as the fix for #944:

| Concern | this branch (#944) | `main` after #949 |
| --- | --- | --- |
| per-item `status: 429` re-sent, and only those | `select_bulk_actions` + `walk_bulk_actions`, positional, `delete`-aware | `select_bulk_items` + `bulk_item_ranges`, positional, `delete`-aware |
| how long it waits | `BACKPRESSURE_PATIENCE` = **120 s** with nothing accepted | `THROTTLE_PATIENCE` = **600 s** from the start of the call |
| whole-request HTTP 429 | `Throttle::HandToParser` — parsed as a bulk response when the body is one, else every action throttled | `bulk_round` — parsed as a round when the body has `items`, else `Attempt::Throttled` and the whole body re-sent at capped backoff for `throttle_patience` |
| says so on stderr | every 5 s (`BACKPRESSURE_ANNOUNCE_EVERY`) | every 30 s (`NOTICE_EVERY`) |
| terminal counters | `bulk_retries` / `bulk_items_reissued` on `xerj-done` | none |
| retry policy shape | `with_retry_throttle` + `Throttle` enum | `retry_loop` + `Attempt` enum |

`git merge origin/main` puts 10 conflict hunks in `esclient.rs` alone, two of them 300 and 750 lines, because both sides rewrote the same functions. Neither "take ours" nor "take theirs" is correct: taking ours reverts #949's fix and its tests, taking theirs reverts #944's.

### What is only on this branch and must survive either way

- **#955** — `BULK_MAX_ACTIONS` windowing, `window_bulk_body`, `count_bulk_actions`, `TooLarge`, `halve_refused_request`, `RequestSizeLedger`, `bulk_windowed`, `bulk_requests_split`. `main` has none of it, and the 48,533-file corpus still cannot write its 51,129-action catalog without it.
- **#929** — `MappingRefused` and `mapping_error`.
- The `raising bulk concurrency` announce throttle (`RAISE_ANNOUNCE_EVERY`).
- `reason=server-backpressure` with `ops_applied` / `ops_remaining` (this one lives in `sync_executor.rs`, not `esclient.rs`, and is keyed off `server_errors`, so it survives either implementation).

### Recommended resolution

Adopt **main's** throttle machinery wholesale — it is authoritative, it is already merged, and on the throttle axis it is the more capable of the two (600 s of patience, and an HTTP-level 429 gets that same patience instead of six transport attempts) — then re-apply #955, #929 and the announce throttle on top of it: `bulk_bounded` keeps the windowing and halving loop and calls main's per-window `bulk` as its inner send.

That resolution is **not** a merge, and it is not free:

- this branch's ten #944 unit tests and two end-to-end tests assert the 120 s patience and this branch's exact message strings, and must be rewritten against main's;
- `bulk_retries` / `bulk_items_reissued` either move onto main's loop or come off `xerj-done`;
- **the numbers change.** `benchmarks/autoindex-resilience/limits-real-node.txt`'s `XERJ_MAX_PROCESS_MEMORY_MB=64` run (`16 re-sends`, `wall=128.8s`) was measured against 120 s of patience, and `content/answers/autoindex-server-back-pressure-429.md`, `docs/recipes/zero-config-indexing.md` §8, `AGENTS.md` and `llms-full.txt` all quote it in words. Every one of those has to be re-measured on a real node, not edited to match.

Whether the user-visible wait should be 120 s or 600 s is a product decision, so it is being put here rather than decided inside a merge commit at 01:00. Everything else on this PR is finished and green.

xerj-org:
### Resolved — one throttle implementation (main's), and everything re-measured at 600 s

The collision described [above](https://github.com/xerj-org/xerj/pull/934#issuecomment-5746889242) is settled on `83e0bd45`, which now merges `main` up to `01d0044d` and is **MERGEABLE**.

**main's throttle machinery won, whole**: `retry_loop` / `Attempt` / `Round` / `select_bulk_items` / `bulk_item_ranges` / `THROTTLE_PATIENCE` 600 s / `throttle_notice` every 30 s. It is already shipped, and on this axis it is the more capable of the two. This branch's `with_retry_throttle` / `Throttle` / `ParsedBulk` / `select_bulk_actions` / `walk_bulk_actions` / `BACKPRESSURE_PATIENCE` 120 s and its ten duplicate #944 unit tests are gone. There is exactly ONE retry/throttle path in `esclient.rs`, not two.

Re-applied on top of it, since `main` has none of it: **#955** (`bulk_bounded` — window the body, push the halves of a refused request onto the *front* of the queue, re-read the learned bound each round, converge to one action; `bulk_window` is main's `Es::bulk` renamed with the too-large exit added; `window_bulk_body` and `count_bulk_actions` are built on main's `bulk_item_ranges`, so the body is walked by one function), **#929** (`MappingRefused`), the **`RAISE_ANNOUNCE_EVERY` announce throttle**, **`throttled_out` / `BackpressureExhausted`** (`reason=server-backpressure` raised from main's loop), and **`bulk_retries` / `bulk_items_reissued`** — now counted for *both* 429 shapes, so the generated-path e2e test's `bulk_retries=3, bulk_items_reissued=12` still holds. #930 and #931 were never touched by the collision.

Nothing of #934's behaviour needed a parallel path.

**Two behaviours genuinely changed** and are in Known limits rather than hidden: patience is 600 s from the first send of a bulk, not 120 s from the last accepted item (so a bulk landing records *slowly* for over ten minutes is now given up on, while one landing nothing is waited on five times longer); and the stderr line is main's `server is shedding load — …; re-offering N rejected record(s) (waited Ws, giving up after 600s)`, once per 30 s.

**Re-measured, not edited.** All six cases re-run 2026-09-20 on real nodes, ports 14300-14317, throwaway data dirs, `xerj` from `53e34f25` as both server and client, sonic (231 files), auth on, `--no-graph`. Machine 32 cores / 119.2 GiB, shared: `load average: 24.93, 21.87, 21.50` at the start, `16.83, 19.06, 20.81` at the end.

| Node setting | Outcome |
| --- | --- |
| control | `ok=true exit=3` · 1,663 records · 236 catalog docs · 6.1 s |
| `max_actions_per_bulk = 64` | one 69-action request refused as the item's 413, halved · `ok=true exit=3 bulk_splits=1` · **1,663 / 236** · 5.6 s |
| `max_body_bytes = 98304` | one 120,443-byte request refused HTTP 413, halved · `ok=true exit=3 bulk_splits=1` · **1,663 / 236** · 5.5 s |
| `max_body_bytes = 65536` | a 70,471-byte record cannot be cut · `exit=1 reason=aborted`, error names `limits.max_body_bytes`; 958 landed first |
| `XERJ_MAX_PROCESS_MEMORY_MB=64` | 32→16→8→4→2→1, then 600 s of re-sends · `ok=false exit=1 reason=server-backpressure wall=609.1s ops_applied=0 ops_remaining=231` (was 128.8 s at 120 s) |
| …same state dir, node restarted on the default cap | `resumed and committed` · `ok=true exit=3` · 1,663 / 236 · 3.5 s |

Both refusal shapes still recover to the control run's counts exactly, so the #955 contract survives the reconciliation. Two honesty notes are in the capture: the old "16 re-sends" was really the announce-line count at one per 5 s, so the new run's **19 notices are not a re-send count** (main announces once per 30 s, and the run ends before an `xerj-done` line could carry `bulk_retries`); and `max_body_bytes = 65536` landed 958 records before the un-cuttable one where the old run landed 886, because which windows land first depends on which worker reaches it — the contract there is the error, not the count.

Surfaces rewritten from the new run: `limits-real-node.txt` and its README table, the 429 / exit-codes / progress answer pages, recipe §8-9, `llms.txt`, `llms-full.txt`, `CHANGELOG.md`, and the regenerated `landing/`. `AGENTS.md` turned out to quote none of it.

Gates on `83e0bd45`: fmt clean · clippy `-D warnings` 0 · `cargo test -p xerj-autoindex` **967 passed, 0 failed, 2 ignored** · integration 8/1/1/4 · toolkit 81 + 44 + 2 OK · **ALL SEO GATES PASS**, factcheck 0 ERROR / 89 WARN · generators settled. Scoped builds only; R2 untouched.

`main` landed #949, #946 and #947 while this verified, so the branch carries three merges; only the first had a code conflict. Not merged, per instruction.

xerj-org:
CI on `83e0bd45`: **21 pass, 0 fail**, 1 skipping (`Release notes match the tree` — not a release commit), including Build + Test, ES-compat YAML conformance, Release build (fat LTO), Autoindex FD smoke on ubuntu/macOS/Windows, Fuzz harnesses, API smoke, Use-case harnesses and the Dependency advisory audit. `mergeStateStatus: CLEAN`. Not merged, per instruction.

PR #956, opened 2026-09-19
fix(xerj-code): preserve hybrid results on transport failures

Amakurai:
Written by Codex (AI coding agent), on behalf of @Amakurai.

## What changes and why

`xc.py --mode hybrid` can retrieve valid BM25 passages and then discard them when its optional mapping lookup or vector request loses its connection. `post(fatal=False)` handled HTTP statuses but still exited on `URLError`; socket read timeouts and interrupted HTTP responses escaped both retrieval helpers. Reading an optional HTTP error body could fail before fallback, too.

Preserve the existing BM25 order and result limit when those optional requests fail. Report BM25-only operation without asserting that mappings are absent when discovery failed. Close optional HTTP errors without reading their bodies; required errors retain their HTTP status and exit `2` even if the error body cannot be read.

Standalone `--mode semantic` has no BM25 result to recover. Make that mode's mapping/search requests explicitly required: HTTP/transport failures exit `2` instead of being mistaken for an ordinary miss at exit `1`. Successful retrieval and genuine empty results retain their behavior. No response fields or ranking algorithms change.

## Verified

The final regression file was executed against the original `xc.py` from `8837a119`, in a disposable directory:

```text
Ran 11 tests
FAILED (failures=11, errors=56)
```

Those counts are failing/errored **subtests**, not additional test methods. On the fixed code:

```text
Ran 11 tests
OK
```

| Contract | Repository-visible reproduction | Negative control | Result |
|---|---|---|---|
| Optional mapping/vector failures retain BM25 hits, order, limit and output format | `test_hybrid_fallback.py`: optional transport/status/body cases | Original source; targeted removal of optional vector/mapping handling | Pass |
| Required BM25/semantic failures produce errors, never empty-result success or no-match text | Same file: primary and standalone failure cases | Original source; remove standalone fatal policy | Pass |
| Healthy fusion, vector-only results and genuine misses keep their behavior | Same file: successful hybrid/semantic and genuine-miss cases | Fixed deterministic fixtures | Pass |

Commands actually run:

- `python3 tools/xerj-code/tests/test_hybrid_fallback.py -v`: all **11 test methods pass** on Windows/Python 3.14.6 (using `python -X utf8`) and Ubuntu/WSL/Python 3.14.4.
- `python3 tools/xerj-code/tests/test_state_ledger.py`: **13 checks pass** on both platforms.
- `bash tools/xerj-code/tests/test_xc_corpus.sh`: **28 checks pass** on Linux.
- Hub manifest validation, toolkit shell syntax checks, Python compilation, workflow timeout guard and its self-test, and `git diff --check`: pass.

The new suite runs in the existing reference-coding CI job; existing steps and its timeout are preserved. Linux checks used an archive of the staged tree exported with `core.autocrlf=false` to avoid Windows checkout line endings. Tests invoke the actual Python CLI entry point with deterministic, scoped urllib fixtures and no user corpus or network dependency. Independent agent review also verified four targeted policy reversions fail their relevant tests.

## Scope and review audit

Independent branch from current `main` at `8837a119`, one commit, five related files, no prerequisites or duplicated prerequisite commits. Reviewed the complete diff and history under `docs/CONTRIBUTION_REVIEW.md`; README and changelog document the behavior.

Pending overlap is explicit: a trial merge with #927 conflicts only in CHANGELOG/README. A trial merge with #934 conflicts in CHANGELOG/README, CI and `xc.py`; if #934 lands first, preserve its selected index prefix while retaining the new required semantic-request policy. Neither pending PR was incorporated or claimed as integration-tested; the branch is based directly on current main.

Durable-state transitions/migrations, publication/install recipes, telemetry/identity changes, feature-gated engine paths and performance audits are not applicable: this is a read-only Python client error-policy change. No performance or neural-retrieval claim is made.

## Not verified / not run

- No live XERJ node or real-network fault injection; the boundary evidence is deterministic urllib request/response fixtures.
- Malformed successful JSON/schema handling remains outside this change. JSON fallback results retain the existing hits-only format without adding mode metadata.
- Cargo formatting/build/tests and ES-YAML conformance were not run: no Rust crate or engine behavior changed; the contribution review protocol exempts isolated changes from unrelated conformance work.


xerj-org (approved):
Reviewed and tested locally on 2026-09-20.

**The change is right.** Valid BM25 passages were being discarded when an *optional* request failed: `post(fatal=False)` handled HTTP statuses but `URLError` still exited, and socket read timeouts or an interrupted response escaped both retrieval helpers. Catching `OSError`/`http.client.HTTPException` is the correct widening — `URLError` is an `OSError`, and a read on an already-returned response can raise those directly without urllib wrapping them. Closing an optional error response without reading its body is right for the same reason. Making standalone `--mode semantic` `fatal=True` fixes a real misreport: a transport failure there was indistinguishable from an ordinary miss at exit 1, and it has no BM25 result to fall back to.

**Verified here, not just read:**
- Its test suite passes on this head: `python3 tools/xerj-code/tests/test_hybrid_fallback.py` → OK, and `test_state_ledger.py` → 13 passed, 0 failed.
- The same test file against `origin/main`'s `xc.py` **fails** (failures=11, errors=56), so it is a real regression test rather than a restatement of current behaviour.
- The new CI step runs exactly that file in the existing reference-coding job.

**Merge order.** #927, #956 and #958 all touch `xc.py`, `CHANGELOG.md`, `tools/xerj-code/README.md` and `ci.yml`, so they conflict with each other. #927 goes first (it is green and its conflict with main is already resolved), then this one. I will bring this branch up to date with `main` and re-run CI before merging — no action needed from you.

Thank you — this is careful work, and the `--json`/transport/window trio all fix silent-wrong-output cases rather than crashes, which is the harder kind to notice.

PR #958, opened 2026-09-19
fix(xerj-code): preserve match scores in passage windows

Amakurai:
Written by Codex (AI coding agent), on behalf of @Amakurai.

## What changes and why

`xc.py` can retrieve the correct source file but print a passage with none of its query matches. `best_window()` scores a character window, then moves its boundaries to newlines without checking what that removes. A matching partial final line can disappear; a long source line can move the excerpt far before its match. Separately, scoring offsets in `body.lower()` mislocates source when Unicode lowercase expands a character, such as U+0130.

Score each window in original-source coordinates. Accept line alignment only when its aggregate query-term score is at least the selected candidate's score; otherwise emit the original bounded candidate. This keeps readable whole lines where possible, the existing tie order and head fallbacks, and the character budget. It applies to `--no-symbol` and the fallback when no matching symbol exists. Symbol extraction and retrieval ranking are unchanged.

The README and changelog describe the behavior. The new offline suite runs in the existing reference-coding CI job, retaining its other steps and timeout.

## Verified

The final regression file was run against the unmodified `xc.py` from current main (`8837a119`) in a disposable directory:

```text
Ran 10 tests
FAILED (failures=7)
```

Seven is the number of failing assertions/subtests, including both CLI cases. With the fix:

```text
Ran 10 tests
OK
```

| Contract | Repository-visible test | Negative control | Final result |
|---|---|---|---|
| Line alignment preserves the selected score, including partial and long lines | `tools/xerj-code/tests/test_passage_window.py` | Original source; independent removal of score guard causes 6 failure records | Pass |
| Unicode lowercase expansion does not shift source windows | Same file, `test_lowercase_expansion_keeps_source_offsets` | Restore lowercase-before-slice: 1 failure | Pass |
| Character budgets, source offsets, readable boundaries, ties and head fallbacks remain valid | Same file, deterministic fixtures | Unfixed implementation fails the targeted cases | Pass |
| Actual CLI formatting retains matching source in both affected modes | Same file, `test_cli_emits_the_matching_source_with_and_without_symbols` | Both CLI subcases fail on original source | Pass |

Commands run on the final tree:

- `python3 tools/xerj-code/tests/test_passage_window.py -v`: 10 tests pass on Windows/Python 3.14.6 (using `python -X utf8`) and Ubuntu/WSL/Python 3.14.4.
- `python3 tools/xerj-code/tests/test_state_ledger.py`: 13 checks pass on both platforms.
- `bash tools/xerj-code/tests/test_xc_corpus.sh`: 28 checks pass on Linux.
- Toolkit shell syntax, hub manifest validation, Python compilation, workflow timeout guard and its self-test, and `git diff --check`: pass.

Linux checks used an archive of the staged tree to retain repository LF endings. The CLI test uses real argument parsing and output handling with scoped urllib fixtures and temporary corpus state; it does not require a node or alter a user's corpus.

## Scope and review audit

Independent branch from current main at `8837a119`: one commit, five related files, no prerequisite commits. The effective diff and history were reviewed under `docs/CONTRIBUTION_REVIEW.md`, including an independent agent review and targeted mutation tests.

Trial merges show anticipated overlap with pending PRs: #927 conflicts in CI, CHANGELOG and README; #934 and #956 conflict in CHANGELOG and README. `xc.py` merges automatically in each trial. These PRs are not incorporated or claimed as integration-tested; preserve their tests/documentation when reconciling whichever lands later. There is no conflict with current main.

Durable-state transitions, migrations, failure atomicity, telemetry, feature-gated engine paths, installation/deployment recipes and performance audits are not applicable to this read-only Python excerpt change. No performance or neural-retrieval claim is made.

## Not verified / not run

- No live XERJ node, real-network retrieval, or end-to-end indexing run. HTTP evidence is limited to deterministic request/response fixtures.
- The coarse candidate sampling and query tokenization remain unchanged; this is not a claim of globally optimal passage selection or preservation of every distinct query term.
- Cargo formatting/build/tests and ES-YAML conformance were not run: no Rust crate or engine behavior changed, and the contribution review protocol exempts isolated changes from unrelated conformance work.


xerj-org (approved):
Reviewed and tested locally on 2026-09-20.

**The change is right.** Two real defects. `best_window()` scored windows in `body.lower()` coordinates and then sliced the source with them — wrong whenever Unicode lowercasing expands a character (U+0130 is the classic case), so the excerpt could be offset from its match. And line-snapping moved the boundaries without re-checking what it removed, so a matching final partial line could vanish, or a source line longer than the window could push the excerpt far from the match. Scoring in source coordinates and accepting the snap only when its aggregate term score holds up is the right shape: readable whole lines when possible, never at the cost of the evidence.

**Verified here, not just read:**
- Its test suite passes on this head: `python3 tools/xerj-code/tests/test_passage_window.py` → OK, and `test_state_ledger.py` → 13 passed, 0 failed.
- The same test file against `origin/main`'s `xc.py` **fails** (failures=7), so it is a real regression test rather than a restatement of current behaviour.
- The new CI step runs exactly that file in the existing reference-coding job.

**Merge order.** #927, #956 and #958 all touch `xc.py`, `CHANGELOG.md`, `tools/xerj-code/README.md` and `ci.yml`, so they conflict with each other. #927 goes first (it is green and its conflict with main is already resolved), then this one. I will bring this branch up to date with `main` and re-run CI before merging — no action needed from you.

Thank you — this is careful work, and the `--json`/transport/window trio all fix silent-wrong-output cases rather than crashes, which is the harder kind to notice.

PR #967, opened 2026-09-19
feat(autoindex): --watch, live reindexing of a folder

xerj-org:
# `xerj autoindex <folder> --watch` — live reindexing

`xerj autoindex` indexes a folder once. Staying current meant re-running it,
which re-walks the tree and re-reads every byte (full hashing is deliberate —
`content.rs` explains why size+mtime cannot prove byte identity). `--watch`
indexes once, then stays resident and reindexes what the filesystem reports.

```sh
xerj autoindex ~/notes --watch --no-graph \
  --url http://127.0.0.1:9200 --prefix notes --state-dir ./state-notes
```

## The measurement that shaped the design

Everything below is from runs recorded in
`docs/measurements/autoindex-watch-2026-09-19.md`: a deterministic corpus of
10,000 files / 4,576,300 bytes / 101 directories, one local node, release
binary, default **lexical** embedder (no neural embedding ran; none of this is a
semantic-search number). Three byte-identical corpus copies, so no route saw
another route's edits; one file **modified**, one **added** and one **deleted**
on each route, each verified searchable (or gone) afterwards. Single samples on
a **shared box** (one-minute load 7.6-20.2 on 32 cores, recorded per run) — read
the order-of-magnitude gaps, not the small ones.

| Keeping a 10,000-file index current | Wall | Client CPU | Server CPU |
|---|---|---|---|
| index the folder from scratch again — modify / add / delete | 293.8 / 347.7 / 312.4 s | 127.9 / 146.9 / 137.9 s | 167.5 / 206.6 / 178.4 s |
| re-run the same `--no-graph` command — nothing changed | 1.7 s | 2.1 s | 0.01 s |
| re-run it — modify / add / delete | 44.3 / 42.7 / 39.1 s | 7.6 / 7.9 / 7.5 s | 27.9 / 28.3 / 23.1 s |
| `--watch` — idle | — | **0.00 s per minute** | 0.07 s per minute |
| `--watch` — modify / add / delete | 47.6 / 45.9 / 40.8 s | 6.6 / 6.9 / 6.5 s | 34.5 / 32.8 / 27.6 s |

`Server CPU` is the node's own `utime+stime` over the same window. It is in the
table because a client-side timer cannot see it, and for a one-file change it is
larger than everything the client does.

Read honestly, it says three things, and the docs in this PR say all three:

1. **Against re-indexing the folder from scratch, incremental wins by an order of
   magnitude** — 47.6 s against 293.8 s, 6.6 CPU-seconds against 127.9, in all
   three change cases.
2. **Against a re-run of the same command, `--watch` is NOT faster per change.**
   It saves about **one CPU-second**, which is exactly the corpus re-hash it
   skips (the no-op re-run measures that term alone at 2.1 CPU-seconds for
   4.58 MB). Its wall clock is a few seconds worse, debounce included. The honest
   case for the feature is idle cost, detection latency and not running anything
   on a timer — not throughput.
3. **The per-change cost is two O(corpus) terms that neither route avoids.**
   Sampled once a second through a one-file re-run rather than inferred: ~13 s
   sealing a snapshot of **10,000 blobs** (one `copy_synced` plus two
   `content::verify` per file, serially, fsynced —
   `sync_executor.rs::create_snapshot_inner`), then ~27 s of server CPU rewriting
   **one catalog document per corpus file** (all 10,000 carry the new `run_id`
   afterwards). Filed as #971; not fixed here, because it is the transactional
   publish path shared with non-watch runs.

Idle cost, sampled twice (it is a reported customer problem, so it is stated as
RAM and threads, not only CPU): **0.00 CPU-seconds per 60 s**, **157 MiB**
resident between passes (258 MB peak during a pass) and **295 threads** — the
scan/index worker pools sized from 32 cores. No polling thread, no timer; the
session blocks on its event channel.

### Two numbers from the first version of this PR were wrong, and are retracted

* The "one file **changed**" rows did not measure a change. The script appended
  to `pkgNNN/fileNNN.md` for a file whose generated extension was `.txt`, so the
  shell **created** a file: those rows measured an addition. Every route and
  every change kind was re-measured from scratch, with the target's existence
  checked before the change.
* "~38 s of the 39 s is the snapshot" was an inference. The sampling above
  disproves it: the snapshot is ~13 s and the rest is server-side.

## `--watch` requires `--no-graph`, and is refused without it — for the right reason

The first version of this PR (and the refusal message itself) said the graph
route does not index a **changed** file. That came from the same mislabelled-file
mistake. Re-measured on the graph route with a file that genuinely exists
(`pkg007/file031.txt`, 516 → 553 bytes), on its own corpus copy:

| Graph-route re-run after… | Result |
|---|---|
| a file's **content** changed | **indexed**: 3.07 s, `files=1`, exit 0, the new text searchable, no stale duplicate (the path holds the same document count as an untouched file) |
| a file was **added** | 3.32 s, **exit 3**, `1 file(s) appeared after the resume plan was frozen and were NOT indexed`; unsearchable until `--fresh` |
| a file was **deleted** | **exit 1, aborted in 0.24 s** — "removing files from an indexed folder is not reconciled yet". Documents stay live, and **every later re-run aborts too** (a second modify after it aborted in 0.20 s and was never indexed) |

So the refusal is right and its reason was wrong: a watcher on the graph path
would go stale on the first new file and stop reindexing entirely on the first
deletion. The message now says that, says a deletion also freezes later re-runs,
and says plainly that a content change *is* reconciled there — otherwise an
operator rebuilds a corpus they did not need to rebuild. A unit test
(`the_watch_refusal_names_additions_and_deletions_not_content_changes`) pins all
four claims and refuses the old wording's return.

## How it works

* **Watch placement** — one non-recursive watch per *admitted* directory, from
  `walk::walk_dirs_opts`: the same traversal, hidden-name rule and
  `.gitignore`/`.xerjignore` stack the indexing walk uses. So an ignored
  `target/` costs no watch and produces no events at all, and a watched run and
  a re-run agree about what is indexed by construction rather than through a
  second implementation. Re-synced after every pass, so new directories are
  picked up.
* **One indexing path** — a burst does not index anything itself. It decides
  which files may skip their re-hash, then runs the ordinary incremental pass
  (`run_index_report_watched`) against the same journal, plan projection and
  publish route. `content::resolve_reporting_carried` is the only hook.
* **What the events are trusted for** — a file skips its re-hash only when **no
  event since that hash named it or any ancestor directory** AND **its
  `(size, mtime, inode)` fingerprint is unchanged**. The cache is in memory and
  per process, so a restart re-hashes in full. A file whose fingerprint moves
  between the pre-hash stat and the post-hash stat is not cached at all.
* **Debounce** — `--debounce` (default 400 ms, max 60000) waits for quiet, so one
  editor save is one pass; a tree that never goes quiet still gets a pass at
  least every 5 s (or `2 × debounce`, whichever is larger).
* **The inotify limit is a hard failure** with the limit, its current value, the
  directory count, the `sysctl`, and the two ways out that need no root. A
  half-watched tree looks live and silently is not, so it is refused.

## Two defects this found, both fixed here

**1. The watcher fed itself** (found by an end-to-end smoke test, not by unit
tests). On a 3-file tree nobody touched, the session ran 48 passes with
`carried=0` on every line. `notify`'s Linux backend asks for `IN_OPEN` and
`IN_CLOSE` (`notify-8.2.0/src/inotify.rs:418`), so **every file a pass reads
reports an event** — and a pass reads the whole corpus. `classify()` now maps
each event kind: reads are ignored, `Access(Close(Write))` is a write,
`Modify(Metadata)` is a weak signal (a chmod, or the atime update from the pass's
own reads) that is worth a pass but never invalidates a digest. `--watch` also
now refuses a `--state-dir` inside the watched folder, which is the same loop
from a plausible command line.

**2. A symlink could be handed its target's digest.** The cache is keyed by the
path the bytes are read from, and under `--follow-symlinks` a link and its target
are two entries with one resolved `path`. The `carryable` rule was applied when
filling the cache but not when reading from it.

## The correctness claim, corrected

My first version of this claimed a watched session converges on "the index a
fresh full `xerj autoindex` would produce". Two tests proved that too broad, and
the PR keeps the failure rather than the claim: an incremental run deliberately
preserves committed dataset identity while a fresh run re-elects dataset slugs
from the corpus it sees (`reconcile_plan.rs`). Rename the directory a dataset was
named after and a fresh rebuild files the same bytes under a different dataset
name — a manual re-run does exactly the same thing.

So the primary contract is the one that always holds and is strictly stronger for
the cache:

> after any sequence of changes, a plain `xerj autoindex` re-run — which
> re-hashes every byte and trusts nothing — changes not one document.

`a_rerun_changes_nothing` asserts it after every watch test, including after a
randomised create/modify/rename/delete/recreate sequence. Fresh-index equality
(same ids, same count, same sources) is still asserted where dataset identity
cannot move.

## Gates

| Gate | Result |
|---|---|
| `cargo fmt --check -p xerj-autoindex` | clean |
| `cargo clippy --profile quick --all-targets -j 10 -p xerj-autoindex` | clean, 0 warnings |
| `cargo clippy --release --all-targets -j 6 -p xerj-autoindex` | clean (run before the last two commits) |
| `cargo test --profile quick -j 10 -p xerj-autoindex` | **805 passed, 0 failed, 2 ignored** (+8+1+1+4 integration) |
| site gates, exactly as `.github/workflows/seo.yml` runs them | `build_articles --check`, `gen_sitemap --check`, `fix_heads --check`, `fix_links --check`, `mk_og_card --check` all ok; `seo_lint` 0 gate violations; `factcheck --self-test`, `--fixture-check`, `--fail-on error` 0 ERROR |
| end-to-end against a real node (port 13240) | 4 changes → 4 passes, edit visible in search, delete gone from search, ignored dir costs nothing, editor swap file ignored |
| measurement re-run against a throwaway node (port 14040) | 3 routes × {modify, add, delete} + two 60 s idle samples; every change verified searchable, every delete verified gone |

ES-YAML conformance was **not** run: nothing here touches the ES wire surface —
`--watch` is a client-side loop in `xerj-autoindex` around the existing publish
path, and no request shape, handler or response field changed.

## Dependency

`notify = "8.2"` — **CC0-1.0**, a public-domain dedication: no copyleft, no
attribution condition, so redistributing it inside Apache-2.0 XERJ is
unrestricted. Recorded in the manifest next to the dependency, with the caveat
that CC0 carries no express patent grant (unlike Apache-2.0 itself), and with the
transitive licences checked against each crate's own metadata rather than a
summary: inotify (ISC), notify-types (CC0-1.0), mio (MIT), bitflags/log/libc
(MIT OR Apache-2.0). The debouncer is ours rather than `notify-debouncer-full`,
which would keep a second file cache and its own rename bookkeeping — state this
crate's journal already owns.

## Docs

* `docs/LIVE_REINDEXING.md` — the page, including the cost table, the ignore
  rules, the `inotify` message, the correctness contract and what is NOT
  implemented.
* `content/answers/keep-index-current-as-files-change.md` — "How do I keep a
  XERJ index up to date as files change?", with `landing/` regenerated by
  `scripts/seo/*` (adding one article renumbers the article sidebar on every
  page, which is why the diff touches 80-odd generated files).
* `docs/measurements/autoindex-watch-2026-09-19.md` — commands and captured
  output for every number above.
* One `CHANGELOG.md` bullet under Unreleased → Added.

## What is NOT implemented

* No watching on the default graph path (`--watch` needs `--no-graph`).
* No object-storage or remote-source watching; `--watch` takes a local folder.
* No incremental generation snapshot — the measured per-change bottleneck.
* No periodic full-hash safety sweep. The one hole in the shortcut (a write that
  produces no event *and* leaves size, mtime and inode identical, e.g. `touch -r`)
  is repaired by a plain re-run, which the docs state.
* No dataset re-election on a watched corpus, as above.
* Not supported on filesystems that report no events (some container bind
  mounts, NFS); the run says so if the watcher cannot start.

## llms.txt lines (for the coordinator)

`scripts/seo/build_articles.py` maintains a generated block in
`landing/llms.txt`, so running the required site gate added exactly these two
lines inside `BEGIN/END GENERATED ARTICLE TREE`. No hand edit was made to
`llms.txt` or `llms-full.txt`:

```
- **Operations: freshness**
  - [How do I keep a XERJ index up to date as files change?](https://xerj.org/answers/keep-index-current-as-files-change.md): How do I keep a search index up to date as files change, without re-indexing everything?
```

## Follow-up, now filed

**#971 — a one-file change costs O(corpus) twice.** `create_snapshot_inner` seals
a 10,000-blob snapshot (~13 s) and the catalog is rewritten file-by-file per
generation (~27 s of server CPU), for one changed file. Both are shared with a
plain re-run and neither is touched by `--watch`; the issue carries the sampling
that attributes them and a direction (reuse the previous generation's blobs for
unchanged `content_id`s, write only the catalog documents that moved). Not
attempted here: transactional durability path, its own tests, its own review.
The issue also records the smaller UX finding from the same runs — the progress
bar sits at `phase=scan 100% (stalled)` for the whole ~39-second tail, because
neither the snapshot nor the publish is a reported phase.


xerj-org:
Measurement pass finished, and it changed two of this PR's claims. Summary, with the full record in `docs/measurements/autoindex-watch-2026-09-19.md` (commands + captured output + the load average on each run):

**Retracted.** The earlier "one file CHANGED" rows measured an **addition** — the script appended to `pkgNNN/fileNNN.md` for a file whose real extension was `.txt`, so the shell created a file. And "~38 s of the 39 s is the snapshot" was an inference, not a measurement. Both are gone.

**Re-measured** on 10,000 files / 4,576,300 bytes / 101 dirs, three byte-identical corpus copies, one file modified / added / deleted on each of three routes, each change verified searchable (or gone). Single samples on a shared box (load 7.6–20.2 on 32 cores):

| route | modify | add | delete |
|---|---|---|---|
| full rebuild | 293.8 s / 127.9 CPU-s | 347.7 s / 146.9 | 312.4 s / 137.9 |
| re-run the same command | 44.3 s / 7.6 | 42.7 s / 7.9 | 39.1 s / 7.5 |
| `--watch` (debounce included) | 47.6 s / 6.6 | 45.9 s / 6.9 | 40.8 s / 6.5 |

Re-run with nothing changed: 1.7 s / 2.1 CPU-s. Watch idle, sampled twice: **0.00 CPU-s per 60 s**, 157 MiB resident, 295 threads.

So: `--watch` is **not** faster per change than re-running the same command (it saves ~1 CPU-second, the corpus re-hash); both beat a full rebuild by an order of magnitude; the win is idle cost and detection latency, and the docs now say exactly that.

**Where a one-file pass goes** (sampled once a second, not inferred): ~13 s sealing a 10,000-blob snapshot client-side, then ~27 s of *server* CPU rewriting one catalog document per corpus file. Two O(corpus) terms, neither touched by `--watch`, both shared with a plain re-run → filed as #971.

**Correction to the feature's justification.** The graph route **does** index a genuinely modified file (3.07 s, `files=1`, searchable, no stale duplicate). What it cannot do is an **addition** (exit 3, skipped until `--fresh`) or a **deletion** (exit 1, aborted — and every later re-run aborts too). The `--watch` refusal message said the opposite about edits; corrected here with a unit test, along with `docs/LIVE_REINDEXING.md`, the answers article and the CHANGELOG.

Gates re-run on the merged-with-main branch: `cargo fmt --check` clean, `clippy --profile quick --all-targets -j 10 -p xerj-autoindex` clean, `cargo test -p xerj-autoindex` **805 passed / 0 failed / 2 ignored** (+8+1+1+4 integration), and every gate in `.github/workflows/seo.yml` (`build_articles/gen_sitemap/fix_heads/fix_links/mk_og_card --check`, `seo_lint` 0 gate violations, `factcheck --self-test/--fixture-check/--fail-on error` 0 ERROR). No hand edit to `landing/llms.txt`. All measurement runs were local-filesystem only against a throwaway node on port 14040 — no object-store or R2 call.

Issue #937, opened 2026-09-18
A declared analyzer stops applying at _flush: analysis.analyzer.default is memtable-only, a per-field analyzer is accepted and ignored, an unknown analyzer name is accepted


xerj-org:
## Summary

An index can declare an analyzer three ways. On `main` none of them survives to a segment:

| declaration | memtable (before flush) | segment (after flush) |
|---|---|---|
| `settings.analysis.analyzer.default` (custom or built-in `english`) | **honoured** | **dropped** — segment is written and queried with `standard` |
| `"analyzer": "<name>"` on a `text` field in the mapping | ignored | ignored |
| `"analyzer": "no_such_analyzer"` on a field | accepted, `200 acknowledged` | — |

So the same `match` query returns a different hit set before and after `_flush`, with no error and no log, and a per-field analyzer is accepted, echoed back by `GET _mapping`, and applied to nothing. That is the accepted-and-ignored class from #204, on the surface the #204 sweep itself worked on: the sweep fixed the create-time gate and the memtable binding; the segment writer still hard-codes `standard`.

`_analyze` cannot be used to see any of this, because it is a separate inline tokenizer: it ignores `field`, and `analyzer: "english"` falls through to the unstemmed default.

Reproduced on the `v1.0.0-rc.74` release binary. `git diff v1.0.0-rc.74..main` touches nothing under `xerj-engine/src`, `xerj-fts/` or the `_analyze` handler, and every code pointer below is from `main` @ 4d8dadbf.

## Repro

Private port, throwaway data dir, `--insecure`. The script creates and deletes four `an_*` indices.

```sh
U=http://localhost:9560; H='content-type: application/json'
hits(){ curl -s -XPOST "$U/$1/_search" -H "$H" -d "$2" | python3 -c "import json,sys; print(json.load(sys.stdin)['hits']['total']['value'])"; }

# A. index-default analyzer — the exact shape of the repo's own test_custom_analyzer_synonym_expansion
curl -s -XPUT $U/an_default -H "$H" -d '{"settings":{"analysis":{"filter":{"my_syn":{"type":"synonym","synonyms":["fast,quick"]}},"analyzer":{"default":{"type":"custom","tokenizer":"standard","filter":["lowercase","my_syn"]}}}}}'
curl -s -XPOST "$U/an_default/_doc/1?refresh=true" -H "$H" -d '{"description":"fast car"}'
hits an_default '{"query":{"match":{"description":"quick"}}}'     # before flush
[...]

PR #991, opened 2026-09-20
fix(api): fields/docvalue_fields pierce an explicit _source includes filter (#932)

### LINKED ISSUE #932: fields clause is silently dropped when _source includes omits the field (sibling of #310)
## What happens

A `fields` clause is silently dropped from every hit when the same request carries an explicit `_source` **includes** list that does not contain the requested field. The same clause works with no `_source` filter and with `_source: false`.

Observed on `origin/main` @ `4d8dadbf` through the in-process router (four documents, mapping `title: text`, `body: text`, `cat: keyword`, `n: integer`), first hit of each response:

| Request | `fields` in the hit |
|---|---|
| `"fields": ["body"]` | `{"body": ["vitamin d supplementation improved bone density in the treatment group"]}` |
| `"_source": false, "fields": ["body"]` | same — returned |
| `"_source": ["title"], "fields": ["title"]` | `{"title": ["Trial results"]}` — returned |
| `"_source": ["title"], "fields": ["body"]` | **absent** |
| `"_source": ["title"], "fields": ["cat"]` | **absent** (so it is not specific to `text` fields) |

All five return HTTP 200 with no warning. `fields` appears to be resolved out of the already-filtered `hit.source`, so an includes list that drops the field also hides it from `fields`.

## What Elasticsearch does

`fields` and source filtering are independent. `fields` reads the stored `_source` ("the `fields` option looks in `_source` to find the values, then parses and formats them using the mappings"), while source filtering only selects "what fields of the source are returned" (`docs/reference/elasticsearch/rest-apis/retrieve-selected-fields.md`). `"_source": ["title"], "fields": ["body"]` returns `_source.title` **and** `fields.body`.

## Why it matters

This is the sibling of #310, which fixed the same symptom for the *default* `_source` projection by asking the engine for the intact source and re-narrowing at emission. The explicit-includes path was not covered. Because an unresolvable `fields` entry is legally omitted, the caller gets a 200 and no signal — the accepted-and-ignored class (#204).

It is a common request shape: keep `_source` small, pull one formatted value through `fields`.

## Found by

The integration tests for the `rerank` search stage (`feat/rerank-stage`). The stage reads judged text from the rendered hit, so it sent blank documents for this request shape. The stage has been hardened to check the rendered hits and does not depend on this being fixed; its test `a_named_field_the_response_does_not_return_is_never_judged_blind` is written to pass under either behaviour.

## Suggested fix

Apply the #310 approach on the explicit-includes path: resolve `fields` / `docvalue_fields` against the intact source, then narrow `_source` at emission. `engine/crates/xerj-api/tests/fields_pierce_default_source.rs` is the template for the paired positive/negative assertions.

xerj-org:
## Problem

A `fields` clause was silently dropped from every hit whenever the same request carried an explicit `_source` **includes** list that did not contain the requested field. `{"_source": ["title"], "fields": ["body"]}` returned HTTP 200 with no `fields` block at all — for text and keyword fields alike — while the identical clause worked with no `_source` filter and with `_source: false`. Because an unresolvable `fields` entry is legally omitted, callers got the accepted-and-ignored class of failure (#204): a 200, no warning, nothing. This is the sibling of #310, which fixed the same symptom for the *default* `_source` projection; the explicit-includes path was not covered.

## Root cause

The engine narrows `hit.source` for an explicit `SourceFilter::Includes`/`Fields` **before** the response layer runs (`apply_source_filter`, `engine/crates/xerj-engine/src/index.rs`), and the per-hit `fields` / `docvalue_fields` builders resolve every value out of that already-narrowed source (`get_source_value_by_path(&h.source, …)` in `search_impl`). The #310 pierce cannot fire here: it is keyed on `body.source.is_none()` and companion-gated, so any explicit `_source` clause — including the `?_source=` and `?_source_includes=` URL spellings promoted into `body.source` — kept the narrowed source and lost the named field.

Elasticsearch treats the two clauses as independent: `fields` reads the stored `_source` and parses/formats it via the mappings, while source filtering only selects "what fields of the source are returned" (`docs/reference/elasticsearch/rest-apis/retrieve-selected-fields.md`).

## Fix

The #310 approach applied to the explicit-filter path, entirely inside `search_impl` in `engine/crates/xerj-api/src/es_compat.rs` — no engine change:

- `value_bearing_names` is now computed for **both** projections (`es_compat.rs:11266`).
- New `source_filter_drops_a_name` (`es_compat.rs:6110`) decides coverage by running the *production* projection (`xerj_engine::filter_object`, already `pub`) over probe documents in both spellings the fields builder resolves — the whole dotted path as a literal key, and the nested walk. A wildcarded name counts as dropped (conservative direction: emission re-applies the filter verbatim, so an unnecessary pierce costs only the `_savings` note, a missed one is the silent 200 of this issue).
- `pierce_explicit_source` (`es_compat.rs:11320`): when the caller's own includes/excludes drop a value-bearing name, the search runs with `SourceFilter::Enabled(true)` — the same shape `"_source": false` already uses for exactly this resolution. Unlike #310 it is **not** companion-gated: the dropped field is the caller's own, text or keyword alike. A filter that already covers every named field does not pierce at all, so those requests stay byte-identical, engine-side narrowing and `_savings` note included.
- **Emission site 1** (`es_compat.rs:12503`): the wire `_source` is re-narrowed with the caller's exact patterns after the sentinel strips and before the synthetic block — same `filter_object`, same input the engine would have filtered.
- **Emission site 3, the scroll snapshot** (`es_compat.rs:12251`): re-narrowed before the scroll context is stored, preserving the collapse sentinels exactly as the engine's own Includes/Fields branches do (take → filter → reattach, #651), so continuation pages neither leak the pierced field nor lose `inner_hits`.

Deliberate trade (same one #310's pierce makes, now pinned by a test): a **pierced** request loses its `_savings` note because `Enabled(true)` is not a measured engine shape; covered requests keep theirs. `_msearch` and `_async_search` render hits through their own minimal builders that never render a `fields` block at all, so this interplay does not exist there — a pre-existing, separate gap.

## Evidence

New integration test `engine/crates/xerj-api/tests/fields_pierce_explicit_source_includes.rs` (6 tests, every assertion paired positive/negative):

- **FAIL-BEFORE** (src fix stashed, tests only): 6/6 failed with the exact issue symptom — `hits.hits.N.fields.body` = `null` under `_source: ["title"]`, and the would-be-pierced request still carried a `_savings` note (no pierce existed).
- **After the fix**: 6/6 pass — the issue's five-row repro table (including the keyword `cat` row), the object spelling `_source: {"includes": […]}`, the URL spellings `?_source_includes=title` and `?_source=title&fields=body`, `docvalue_fields`, an excludes-only filter, scroll continuation pages, and the `_savings`-note oracle proving covered names do **not** pierce while pierced ones trade the note.
- Full `xerj-api` crate suite: **638 passed / 0 failed** (includes the existing gates `fields_pierce_default_source`, `collapse_inner_hits_under_source_includes`, `savings_stat_is_honest`, `rerank_stage_http`, and the scroll-suppression suites).
- `cargo clippy --profile ci-test -p xerj-api -- -D warnings`: clean.
- ES-YAML conformance suite against a private-port server (throwaway data dir) built from this branch: **1371 passed / 0 failed / 3 skipped** — gate at 0 failed holds.

Fixes #932


xerj-org:
Merged to main as commit 6f149fc0 (merge of this branch, pushed directly because the merge API was wedged on a mergeability recompute). Verified: the branch head is an ancestor of main. All checks were green on head d3566f0d. Issue #937 is closed by this.

